PR #185, opened 2013-05-28
Added redirect_history property to HTTPResponse instances

cezarsa:
Hey guys, first of all, thank you for your awesome work on urllib3.

One thing I missed though was being able to figure out what were the redirects my request had to jump through before arriving at the final destination, and what was this final destination.

To solve this problem I added a new property to instances of HTTPResponse called `redirect_history`. It contains a list of tuples with each status code and redirect location sent by the servers before the request was completed. I also added some test cases to illustrate this behavior.

Please let me know if you think this useful. I'm open to any changes you might require, just let me know and I'll work on it.

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -121,6 +121,17 @@ class TestingApp(WSGIHandler):
         headers = [('Location', target)]
         return Response(status='303 See Other', headers=headers)
 
+    def multi_redirect(self, request):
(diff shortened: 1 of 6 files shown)


shazow:
@cezarsa Sorry this never got merged! Our latest progress on this feature is https://github.com/shazow/urllib3/pull/848, would love your feedback. :)

PR #284, opened 2013-11-23
SOCKS proxy support

Anorov:
Here's the SOCKS proxy support patch.

Usage is identical to setting HTTP proxies: `ProxyManager("socks4://localhost:1080")` or `ProxyManager("socks5://localhost:1080")` (or `proxy_from_url()`). Default ports for both are 1080.

I added new SOCKS Connection classes in `connection.py`, and added a `_is_socks` attribute to the proxy URL that is passed to `ProxyManager` (on the `Url` object). The `_is_socks` attribute is used by the connection pools so that HTTP proxy negotiation steps aren't taken when the proxy has a SOCKS scheme. This could possibly be refactored (separate `socks_proxy` and `http_proxy` attributes?); tell me how you feel about the `_is_socks` solution.

I also touched up some other parts of the codebase, as follows, which I noticed while working through files and debugging.

* Fixed inconsistent indentation in a few places.
* Fixed some typos in comments.
* Changed `__str__` methods to `__repr__`. `__str__` automatically defers to `__repr__` if it isn't already defined, but the reverse isn't true. So this keeps the same behavior while also allowing easier debugging from the REPL.
* Added a newline in the middle of the proxy connection exception message, to make it a little easier to read.

I believe this branch should currently work fine for regular usage; it seems to have no trouble with any SOCKS proxies.

However, I had many confusing issues and bugs while trying to modify the test suite. The tests are currently broken (I added `raise SkipTest()` in the proxy tests, for the time being). I am using Twisted for the SOCKS4 and SOCKS5 proxy servers, and added that to the test requirements.

The test issues seem to stem from the Tornado HTTP and HTTPS dummy servers. On my computer, they spat out odd SSL errors (like `SSLError: [Errno 1] _ssl.c:504: error:1411B072:SSL routines:SSL3_GET_NEW_SESSION_TICKET:bad message type
` and many others), and would hang seemingly randomly.

I upgraded the Tornado version in `test-requirements.txt` because the version listed before appeared to have a bug that prevented me (or others) from running in IPv6 mode, see here: https://github.com/facebook/tornado/pull/593; I had to upgrade to get IPv6-related tests to pass. I'm not sure if the upgrade has anything to do with the current issues.

I think it has something to do with trying to run Tornado in separate threads. I was not able to get multithreading to work with Twisted, so I spawned a new process for each SOCKS proxy server instead; Twisted seems to work fine, as I can connect to and use the proxy servers while the tests are running.

I would appreciate if someone could modify the tests so the HTTP servers work properly, or give me some advice on how to fix the issue. And of course, I'd appreciate any suggestions or changes for the SOCKS patch itself.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -93,5 +93,8 @@ In chronological order:
 * Peter Waller <p@pwaller.net>
   * HTTPResponse.tell() for determining amount received over the wire
 
+* Pyotr Vorona <anorov.vorona@gmail.com>
+  * Added support for SOCKS proxies
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/dummyserver/httpproxy.py
+++ b/dummyserver/httpproxy.py
@@ -37,7 +37,7 @@ import tornado.httpclient
 __all__ = ['ProxyHandler', 'run_proxy']
 
 
-class ProxyHandler(tornado.web.RequestHandler):
+class HTTPProxyHandler(tornado.web.RequestHandler):
     SUPPORTED_METHODS = ['GET', 'POST', 'CONNECT']
 
     @tornado.web.asynchronous
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -20,8 +20,6 @@ import tornado.web
 
 from dummyserver.handlers import TestingApp
 from dummyserver.httpproxy import HTTPProxyHandler
-from dummyserver.socks4proxy import run_socks4_proxy
-from dummyserver.socks5proxy import run_socks5_proxy
 
 log = logging.getLogger(__name__)
 
@@ -35,7 +33,6 @@ DEFAULT_CA_BAD = os.path.join(CERTS_PATH, 'client_bad.pem')
 
 
 # Different types of servers we have:
-
 
 class SocketServerThread(threading.Thread):
     """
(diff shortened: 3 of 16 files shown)


jbrockmendel:
@a3nm You mentioned a bug in requesocks.  Can you comment on what that bug was and if it's been resolved in the 11 months since you mentioned it?  Does it happen to involve hundreds of unclosed sockets?

a3nm:
@jbrockmendel From my notes about the bug at the time, the problem was that requests performed by requesocks through the proxy included the entire path, e.g., "GET http://example.com/foo HTTP/1.1" rather than "GET /foo HTTP/1.1". This unusual request style was often OK but caused problems with some servers. This does not seem to have anything to do with unclosed sockets.

As requesocks does not seem to have been updated in two years (https://github.com/dvska/requesocks), I see no reason to believe that this bug has been fixed since then.

cattleyavns:
@Anorov: Hi Anorov, can you finish the socks support for urllib3, these days socks support is really needed, because there is socks proxy like JonDo, AdvOr, Bitvise SSH Client that can help users increase privacy, also can improve internet speed..

shazow:
@cattleyavns Thanks for voicing your support but keep in mind that it's not any one contributor's responsibility to supply these features. :) We welcome you to dive in and start working on socks support for urllib3, or otherwise you can provide funding for [urllib3 contributors via bountysource](https://www.bountysource.com/teams/urllib3) or [through sponsorship](https://urllib3.readthedocs.org/en/latest/#sponsorship).

shazow:
Obsoleted by #762

ydlz123:
who can take me sock4,just commuation with tor?

PR #444, opened 2014-08-20
allow performing *only* fingerprint verification

t-8ch:
Before if we verified a fingerprint we also required a valid certificate.
Valid in 'signed by a trusted CA, current time in valid timerange', not in the
sense of matching hostnames.
We should allow people to verify certificates via fingerprints that are totally
crap otherwise.


untitaker:
Not sure how to write code that works with both versions and pins an outdated (or otherwise broken) cert.

* ``verify_fingerprint='...'`` only works for the new versions.
* ``verify=False, verify_fingerprint='...'`` is a security risk on old versions.

PR #468, opened 2014-09-16
Changed default: preload_content=False

### LINKED ISSUE #436: Make `preload_content=False` the new default?
More and more people want to use response objects as file-like objects (or streams). The default pre-consume-and-leave-nothing-to-read is confusing.

This would be a fairly substantial change which could cause leaks to old code using urllib3, so it would need to be a major release bump.

See also: https://stackoverflow.com/questions/25067580/passing-web-data-into-beautiful-soup-empty-list/25068934#25068934
##

shazow:
Fixes #436. Hopefully this will be backwards compatible to nearly all users. Slated for v1.10.

If there are any concerns, please raise them. :) /cc @Lukasa @sigmavirus24

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changes
 dev (master)
 ++++++++++++
 
+* Changed default response reading to not consume the content
+  until it is explicitly read. (Issue #468)
+
 * Add Url.url property to return the composed url string. (Issue #394)
 
 * Fixed PyOpenSSL + gevent ``WantWriteError``. (Issue #412)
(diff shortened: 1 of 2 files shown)


alecz20:
If we get rid of the .data property, how can we still get the contents? The documentation is mentioning the data property, but nothing more. If this property is removed, the documentation should make it clear on what is the new way of accessing that content. Even better, the alternative should be provided before .data is removed.

shazow:
@alecz20 You'll be able to `.read()` instead of `.data`.

alecz20:
I am trying to figure out how to replace the calls to `.data` with `.read()` in my code but I get empty strings. With `.data` I get a bytes object.
The response is raw binary. This is with 1.9.1 btw.

shazow:
See the title of this issue. :) you need preload_content=False.

jonparrott:

PR #477, opened 2014-09-24
disable tunneling for https proxy

### LINKED ISSUE #476: allow http requests through https proxy
this has been brought up a few times for `requests`:
- https://github.com/kennethreitz/requests/issues/1622
- https://github.com/kennethreitz/requests/issues/1903

and elsewhere:
- http://wiki.squid-cache.org/Features/HTTPS#Direct_SSL.2BAC8-TLS_connection
- http://dev.chromium.org/developers/design-documents/secure-web-proxy

case for me is: send http req --ssl--> https proxy (manipulates and fwds) --> origin. but `ProxyManager` tunnels these reqs unnecessarily (?).

would like to disable tunneling for https proxies so behavior is same as for http proxies just over ssl'd cxn. leaves open  proxying https reqs through https proxy (this doesnt appear to work now, but dont specifically need it).
##

bninja:
i dont think https proxying http reqs works now anyway (added test fails without patch). by disabling tunneling for https proxies we can now: https proxy http requests (no http connect tunnel est needed).

closes #476

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -93,6 +93,10 @@ class HTTPDummyProxyTestCase(unittest.TestCase):
     proxy_host = 'localhost'
     proxy_host_alt = '127.0.0.1'
 
+    proxys_host = 'localhost'
+    proxys_host_alt = '127.0.0.1'
+    proxys_certs = DEFAULT_CERTS
+
     @classmethod
     def setUpClass(cls):
         cls.io_loop = ioloop.IOLoop()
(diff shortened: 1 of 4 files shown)


dummyserver/testcase.py
```diff
@@ -109,6 +113,10 @@ def setUpClass(cls):
         cls.proxy_server, cls.proxy_port = run_tornado_app(
             app, cls.io_loop, None, 'http', cls.proxy_host)
 
+        app = web.Application([(r'.*', ProxyHandler)])
```
schlamar: Is there a reason why you are running two proxies in one test case? Would make more sense if you make a separate class as HTTPS proxy test case.

schlamar:
> this patch just lets us use https proxies w/ http requests using normal (tunneling is n/a here) proxy semantics.

I don't think this is an acceptable solution. In this case the proxy can intercept all data from HTTPS connections because the proxy can encrypt all TLS data sent to it.

BTW, some information about encrypted proxy connections can be found [here](http://wiki.squid-cache.org/Features/HTTPS#Encrypted_browser-Squid_connection).

jonparrott:

PR #493, opened 2014-10-28
Raise timeout error instead of SSLError on handshake timeout

### LINKED ISSUE #492: timeout can be raised during _validate_conn
Steps to reproduce:
- set up a port to accept incoming TCP traffic, and never send any data back (for example port 5501 on [hamms](https://github.com/kevinburke/hamms))
- make an HTTPS request to the port with a timeout

``` python
import urllib3, certifi
http = urllib3.PoolManager(cert_reqs='CERT_REQUIRED', ca_certs=certifi.where())
r = http.request('GET', 'https://localhost:5501/', timeout=3)
```

The traceback is roughly this:

``` python
  File "urllib3/request.py", line 68, in request
    **urlopen_kw)
  File "urllib3/request.py", line 81, in request_encode_url
    return self.urlopen(method, url, **urlopen_kw)
  File "urllib3/poolmanager.py", line 153, in urlopen
  File "urllib3/connectionpool.py", line 516, in urlopen
    body=body, headers=headers)
  File "urllib3/connectionpool.py", line 304, in _make_request
    self._validate_conn(conn)
  File "urllib3/connectionpool.py", line 722, in _validate_conn
    conn.connect()
  File "urllib3/connection.py", line 237, in connect
    ssl_version=resolved_ssl_version)
  File "urllib3/util/ssl_.py", line 132, in ssl_wrap_socket
    ssl_version=ssl_version)
  File "/usr/local/Cellar/python/2.7.8_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 392, in wrap_socket
    ciphers=ciphers)
  File "/usr/local/Cellar/python/2.7.8_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 148, in __init__
    self.do_handshake()
  File "/usr/local/Cellar/python/2.7.8_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ssl.py", line 310, in do_handshake
    self._sslobj.do_handshake()
```

The error is a SSLError but it seems like this should be a timeout:

> SSLError: [Errno 0] The handshake operation timed out

We have logic to catch SSLErrors and raise them as timeouts, which would catch this:

```
    except BaseSSLError as e:
        # Catch possible read timeouts thrown as SSL errors. If not the
        # case, rethrow the original. We need to do this because of:
        # http://bugs.python.org/issue10272
        if 'timed out' in str(e) or \
           'did not complete (read)' in str(e):  # Python 2.6
            raise ReadTimeoutError(
                    self, url, "Read timed out. (read timeout=%s)" % read_timeout)

        raise
```

Unfortunately this code block doesn't catch this timeout which is raised during `_validate_conn`.

There are a few options here.
- duplicate the `if "timed out" in ...` logic around `_validate_conn`
- move the `if "timed out" in ...` logic up to urlopen. Existing read timeout errors would get caught in the `except (BaseSSLError, CertificateError) as e` block in urlopen, and not have the connection closed/replaced.
- do nothing and leave "handshake timed out" errors as SSLErrors.

kevinburke:
Fixed #492 . Adds a test to verify the new behaviour.

So, you can make the if statement in `_raise_timeout` one big conditional but it looks super ugly, see https://gist.github.com/kevinburke/36e6403a1f443092f858.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -320,9 +320,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # Trigger any extra validation we need to do.
         try:
             self._validate_conn(conn)
-        # Python2 raises this as a BaseSSLError. Python3 raises it as socket
-        # timeout.
         except (SocketTimeout, BaseSSLError) as e:
+            # Py2 raises this as a BaseSSLError, Py3 raises it as socket timeout.
             self._raise_timeout(err=e, url=url, timeout_value=conn.timeout)
             raise
 


thomascellerier:
I'm having an issue here and I wonder if it relates to this patch.
I have a HTTPSConnection that throws a read time out, then (in a `__del__`) I reuse the same connection (created with `urllib3.connection_from_url`), then in case of error I do some teardown using the same connection and I get this error when reusing the connection from url object.
urllib3.exceptions.SSLError: Underlying socket connection gone (_ssl.c:1568)

reproduce with something like this:
```Python
conn = urllib3.connet_from_url(my_url)
try:
    conn.request('POST', '/mypath', body=mybody timeout=10, retries=False)
except urllib3.ReadTimeoutError:
    conn.request('DELETE', /someobect', timeout=10, retries=False)
```

Issue #459, opened 2014-08-29
urllib3 upload retry fails to seek file-object


joneskoo:
I believe urllib3 fails to seek file object to beginning when it retries the upload. This bug is probably the true reason why I got timeout in issue #442 .

I overwrote read method in file object to demonstrate this. 

https://gist.github.com/joneskoo/3fe7fb68836d9b387273

The retry handling is non-trivial and gets even more complicated as there's file objects and not strictly strings to deal with.

I believe the retry with file objects should:
1. seek file to beginning or position reported by tell() before reading the file object
2. FAIL if the file object does not support seeking

Currently if I understand correctly the current behaviour will corrupt uploads or result in Timeout error as I experienced before.

Thanks @Lukasa for continued help trying to debug this.
##

nateprewitt:
This has been resolved with #1039.

PR #506, opened 2014-11-14
Seek uploaded file-like object to original position before retry (Fix for issue #459)

joneskoo:
Fix candidate for #459. @sigmavirus24 can you check if this is good enough?

It doesn't attempt to do anything for non-seekable files. I'm not sure how to properly deal with those.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -721,6 +721,20 @@ class TestRetryWithTimeout(HTTPDummyServerTestCase):
                                  assert_same_host=False, redirect=False)
         self.assertEqual(resp.status, 200)
 
+    def test_redirect_put_file(self):
+        '''PUT with file object should work with a redirection response'''
+        retry = Retry(total=3, status_forcelist=[418])
+        # httplib reads in 8k chunks; use a larger content length
+        content_length = 65535
+        uploaded_file = io.BytesIO(b'A' * content_length)
+        headers = {'test-name': 'test_redirect_put_fileobj_timeout',
+                   'Content-Length': str(content_length)}
+        resp = self.pool.urlopen('PUT', '/redirect?target=/successful_retry',
+                                 headers=headers,
+                                 retries=retry,
+                                 body=uploaded_file,
+                                 assert_same_host=False, redirect=True)
+        self.assertEqual(resp.status, 200)
 
 if __name__ == '__main__':
     unittest.main()


shazow:
I'd still love to get this merged, any idea on what's outstanding?

nateprewitt:
It looks like this issue has been sitting for a while. I just did some work similar to this in Requests, and I'd be happy to try and replicate some of that over here.

Since this is a much more contained use case (single method) than Requests, would something like this work? It still needs a little more error handling and some of the logic would probably be nicer somewhere like util/request.py.

```python
def urlopen(self, ... body_pos=None, **response_kw):

[...]

is_rewindable = (
    getattr(body, 'tell', None) is not None  and
    getattr(body, 'seek', None) is not None
)

# Make note of current file pointer position if available and
# not previously noted. If it's noted, set body to that position.
if is_rewindable and body_pos is None:
        body_pos = body.tell()
elif is_rewindable:
        body.seek(body_pos)
```

I know this is yet another parameter into an already crammed method, but the alternative solution of a BodyWrapper seems a bit over-engineered for a smaller edge case. 

* If we wrap any body then we're creating a new object for every request, when we'll only need it for a fraction of those. 
* If we create the object conditionally (has read, seek, tell? file-like superclass?) we're still adding the same logic as above and probably more to avoid the `body_pos` param. 

Are there are other use cases that would benefit from extracting a generalized body object out? If so, I may be being a bit hasty. It just seems like a generic body object will either be incredibly overloaded or unpredictable (e.g. sometimes it has attr_x, other times it won't)

SethMichaelLarson:
@nateprewitt This should also be integrated into #1030

nateprewitt:
@SethMichaelLarson sure, I think we could get that worked out. What were you thinking is the best way to make that happen? Is there a need for recreating the generator in filepost, or is it because these are both related to streaming content?

SethMichaelLarson:
@nateprewitt Just linking these two PRs as they're working on the same theme but different areas.  If this PR is merged in first then there may be change that I need to accommodate for in my PR and I want to be able to find it quickly. :)  Maybe it would be nice to integrate changes here into #1030 but I don't think that's necessary.

nateprewitt:
@SethMichaelLarson Alright :) Let me put together a few more commits on this to get things working I'll open a PR. Either way, there isn't much file overlap so I think the conflicts would be minimal, if they even exist. @Lukasa can I get a quick confirmation that the above approach is one we'd be willing to consider?

Lukasa:
It seems reasonable to me. =)

joneskoo:
For the record, you may consider this PR abandoned from my end, sorry. Absolutely amazing that @nateprewitt picked this up! I'll be really happy if a fix lands to urllib3. 🎉

nateprewitt:
Always happy to lend a hand, @joneskoo 😊 Glad we can finally get your work into urllib3, thanks for providing these tests!

nateprewitt:
This can be closed since it was superseded by #1039.

PR #507, opened 2014-11-21
Add ssl_context parameter to Connection classes

sigmavirus24:
After creating a shim SSLContext class in urllib3.util.ssl_, we now want
to start accepting instances of that class in general so users can take
advantage of some features provided only on certain versions of Python.

The first step is to allow users to pass ssl_context to both
HTTPSConnection and VerifiedHTTPSConnection and then pass these down to
the shim ssl_wrap_socket function in urllib3.util.ssl_. This does not
yet account for use with PyOpenSSL.

---

So this basically just works except for the fact that the following will raise an InsecureRequestWarning:

```python
import ssl
import urllib3
import urllib3.util.ssl_ as ussl


context = ussl.SSLContext(ssl.PROTOCOL_SSLv23)
context.load_verify_locations('../requests/requests/cacert.pem')
context.verify_mode = ssl.CERT_REQUIRED

manager = urllib3.PoolManager(ssl_context=context)
resp = manager.urlopen('GET', 'https://api.github.com/users',
                       preload_content=False)
```

Because the way I wrote `urllib3.util.ssl_.ssl_wrap_socket`, any settings on the the context set needs to be confirmed when creating the pool manager, e.g., the following works:

```python
import ssl
import urllib3
import urllib3.util.ssl_ as ussl


context = ussl.SSLContext(ssl.PROTOCOL_SSLv23)
context.load_verify_locations('../requests/requests/cacert.pem')
context.verify_mode = ssl.CERT_REQUIRED

manager = urllib3.PoolManager(ssl_context=context,
                              cert_reqs=ssl.CERT_REQUIRED)
resp = manager.urlopen('GET', 'https://api.github.com/users',
                       preload_content=False)
```

I think if we have `urllib3.util.ssl_.ssl_wrap_socket` return early when given a valid context, we won't break any prior behaviour, but I have yet to test that. Does anyone have any thoughts?

### DIFF
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -26,7 +26,7 @@ pool_classes_by_scheme = {
 log = logging.getLogger(__name__)
 
 SSL_KEYWORDS = ('key_file', 'cert_file', 'cert_reqs', 'ca_certs',
-                'ssl_version', 'ca_cert_dir')
+                'ssl_version', 'ca_cert_dir', 'ssl_context')
 
 
 class PoolManager(RequestMethods):
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -344,12 +344,11 @@ class TestUtil(unittest.TestCase):
         self.assertEqual(timeout.get_connect_duration(), 37)
 
     def test_resolve_cert_reqs(self):
-        self.assertEqual(resolve_cert_reqs(None), ssl.CERT_NONE)
-        self.assertEqual(resolve_cert_reqs(ssl.CERT_NONE), ssl.CERT_NONE)
-
-        self.assertEqual(resolve_cert_reqs(ssl.CERT_REQUIRED), ssl.CERT_REQUIRED)
-        self.assertEqual(resolve_cert_reqs('REQUIRED'), ssl.CERT_REQUIRED)
-        self.assertEqual(resolve_cert_reqs('CERT_REQUIRED'), ssl.CERT_REQUIRED)
+        self.assertEqual(resolve_cert_reqs(None, None), ssl.CERT_NONE)
+        self.assertEqual(resolve_cert_reqs(ssl.CERT_NONE, None), ssl.CERT_NONE)
+        self.assertEqual(resolve_cert_reqs(ssl.CERT_REQUIRED, None), ssl.CERT_REQUIRED)
+        self.assertEqual(resolve_cert_reqs('REQUIRED', None), ssl.CERT_REQUIRED)
+        self.assertEqual(resolve_cert_reqs('CERT_REQUIRED', None), ssl.CERT_REQUIRED)
 
     def test_is_fp_closed_object_supports_closed(self):
         class ClosedFile(object):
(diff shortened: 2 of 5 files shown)


shazow:
How's this looking?

sigmavirus24:
I need to pick this back up soon. Looks like the demand for it is increasing.

tiran:
@Lukasa has pointed me to this pull request. It's exactly what I need. Thanks!

urllib3/poolmanager.py
```diff
@@ -24,8 +24,8 @@
 
 log = logging.getLogger(__name__)
 
-SSL_KEYWORDS = ('key_file', 'cert_file', 'cert_reqs', 'ca_certs',
-                'ssl_version')
+SSL_KEYWORDS = set(['key_file', 'cert_file', 'cert_reqs', 'ca_certs',
+                    'ssl_version', 'ssl_socket'])
```
divergentdave: I think this should be `ssl_context`, right?

sigmavirus24: Good catch

sigmavirus24:
So I resolved a bunch of the review comments last night. That said the tests are still failing and I've yet to dig into why.

divergentdave:
All three tests failures are related to setting the TLS version. It looks like when the ConnectionPool sets the ssl_version on a connection (https://github.com/sigmavirus24/urllib3/blob/5e6a67d3c670c5c8c80a733e75218c133e0bd03e/urllib3/connectionpool.py#L674)[here], nothing happens, because the connection has already read self.ssl_version and created a context in its constructor (https://github.com/sigmavirus24/urllib3/blob/5e6a67d3c670c5c8c80a733e75218c133e0bd03e/urllib3/connection.py#L173)[here]. Plus, I don't see anything that would set self.ssl_version on the connection.

divergentdave:
All three tests failures are related to setting the TLS version. It looks like when the ConnectionPool sets the ssl_version on a connection [here](https://github.com/sigmavirus24/urllib3/blob/5e6a67d3c670c5c8c80a733e75218c133e0bd03e/urllib3/connectionpool.py#L674), nothing happens, because the connection has already read self.ssl_version and created a context in its constructor [here](https://github.com/sigmavirus24/urllib3/blob/5e6a67d3c670c5c8c80a733e75218c133e0bd03e/urllib3/connection.py#L173). Plus, I don't see anything that would set self.ssl_version on the connection before that point in the constructor. Perhaps the best way to preserve HTTPConnection.ssl_version as an API would be to give it a setter/getter and pass it on to the context?

divergentdave:
I have a fix for the tests in my fork, take a look at divergentdave@1f80feb2bf68aed33173f32125cf8191a82877c7

sigmavirus24:
Feel free to send a pr to my branch on my fork
-- 

SirCmpwn:
What outstanding issues are there with this pull request, and how can I help?

sigmavirus24:
Superseded by #835

PR #526, opened 2014-12-19
always disable the builtin hostname verification

### LINKED ISSUE #524: assert_hostname=False seems to be ignored in 1.10
I have some code that is relying on assert_hostname=False to work.
I upgrade urllib3 to version 1.10 and the code fails with SSLError: hostname 'remote-host' doesn't match 'localhost'

I haven't looked through the code to try to determine why this happens.

Did anyone else notice this? If so, what is the fix.

Essentially I am creating a connection pool like this:

```python
import urllib3

http = urllib3.PoolManager(cert_reqs='CERT_REQUIRED',
                           ca_certs="cert.pem",
                           assert_hostname=False)
try:
    r = http.request('GET', 'https://remote-host:4443/')
    print("Certificate verification NO HOSTNAME successful")

    print ("SSL Error:", e)
    return -1
    
return 0
```
<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7114018-assert_hostname-false-seems-to-be-ignored-in-1-10?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

t-8ch:
It conflicts with our own, more flexible functionality

Fixes #524

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -240,6 +240,15 @@ class TestHTTPS(HTTPSDummyServerTestCase):
                                         '7A:F2:8A:D7:1E:07:33:67:DE'
         https_pool.request('GET', '/')
 
+    def test_good_fingerprint_and_hostname_mismatch(self):
+        https_pool = HTTPSConnectionPool('127.0.0.1', self.port,
+                                         cert_reqs='CERT_REQUIRED',
+                                         ca_certs=DEFAULT_CA)
+
+        https_pool.assert_fingerprint = 'CC:45:6A:90:82:F7FF:C0:8218:8e:' \
+                                        '7A:F2:8A:D7:1E:07:33:67:DE'
+        https_pool.request('GET', '/')
+
     @requires_network
     def test_https_timeout(self):
         timeout = Timeout(connect=0.001)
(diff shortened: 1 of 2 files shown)


alecz20:
We should also add a test case for this issue. Even if this might not break in the future, a test case would prevent a release if the functionality is broken somewhere else.

t-8ch:
There comes a patch.

sigmavirus24:
:+1: when the tests pass

t-8ch:
They passed, finally...

Lukasa:
LGTM. :+1: :cake:

Yasumoto:
Howdy all,

Any chance you can cut a new release that includes this? I'm relatively sure that we're running into #543 and I'd be willing to pitch in if needed so we can get pip (and thus virtualenv) upgraded.

Thanks!
Joe

shazow:
@Yasumoto Good call. How about early next week? We have a couple of issues I'd like to get merged and then push out 1.10.1

Yasumoto:
That sounds awesome! We're [pretty far downstream](https://issues.apache.org/jira/browse/AURORA-1114), so much appreciated :)

shazow:
Np, thanks for bringing it up. :) :tropical_fish:

shazow:
@Yasumoto Just pushed out v1.10.1, let me know if you have any problems. :)
https://pypi.python.org/pypi/urllib3/1.10.1

Yasumoto:
Thanks @shazow !! Will do :+1:

alecz20:
I can confirm that the issue is resolved in v1.10.1

shazow:
@alecz20 Glad to hear it, thanks for the followup!

PR #528, opened 2015-01-02
Updated Security documentation for proper capture of security warnings

BenjamenMeyer:
Per the issue I filed in https://github.com/shazow/urllib3/issues/525#issuecomment-68499803, updated the documentation so that people know the warnings can be easily captured instead of having to ignore them.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -133,5 +133,8 @@ In chronological order:
 * Evgeny Kapun <abacabadabacaba@gmail.com>
   * Bugfixes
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
In reStructuredText you have to use two backticks for fixed-width styling. In other words

```
`warnings`
```

Should be 

```rst
``warnings``
```

And some goes for logging. Please update this appropriately

BenjamenMeyer:
@sigmavirus24 - updated as requested

sigmavirus24:
:+1: from me

PR #530, opened 2015-01-08
Example downloading files with shutil.copyfileobj

alecz20:
Added an example on using the response as a file-like object in a context manager to properly handle large files.

Current documentation is not very clear on how to properly download large files in a context manager.

(see http://stackoverflow.com/questions/27387783/how-to-download-a-file-with-urllib3 for more details)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -97,6 +97,20 @@ like buffering:
     >>> secondpart = b.read()
 
 
+The response can be treated as a file-like object.
+A file can be downloaded directory to a local file in a context without
+being saved in memory.
+
+.. doctest ::
+
+    >>> import urllib3
+    >>> import shutil
+    >>> url = 'http://example.com/big.file'
+    >>> http = urllib3.PoolManager()
+    >>> with http.request('GET', url, preload_content=False) as r, open(filename, 'wb') as f_out:
+    >>> ....    shutil.copyfileobj(r, f_out)
+
+
 Components
 ==========
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -105,6 +105,7 @@ being saved in memory.
 
     >>> import urllib3
     >>> import shutil
+    >>> filename = '/path/to/local/file'
     >>> url = 'http://example.com/big.file'
     >>> http = urllib3.PoolManager()
     >>> with http.request('GET', url, preload_content=False) as r, open(filename, 'wb') as f_out:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -98,7 +98,7 @@ like buffering:
 
 
 The response can be treated as a file-like object.
-A file can be downloaded directory to a local file in a context without
+A file can be downloaded directly to a local file in a context without
 being saved in memory.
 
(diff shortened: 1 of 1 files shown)


Lukasa:
This is a good idea! The build breakages appear to be irrelevant as well.

I note that the doctest has no way of passing, because `filename` is undefined (and so `NameError`). That said, AFAIK doctests aren't run for urllib3. @shazow, is that right? If so, we should probably just remove the doctest directives from the block. Otherwise, we should integrate doctest into Travis.

shazow:
@Lukasa Yea you're right. Would be swell if somebody wants to get doctests working, else remove them. :)

docs/index.rst
```diff
@@ -97,6 +97,20 @@ like buffering:
     >>> secondpart = b.read()
 
 
+The response can be treated as a file-like object.
+A file can be downloaded directory to a local file in a context without
+being saved in memory.
+
+.. doctest ::
+
+    >>> import urllib3
```
shazow: For consistency with other examples, let's skip the urllib3 import? (I think the narrative assumed the examples continue from previous contexts.)

shazow:
@alecz20 I really appreciate this, doc contributions are very desirable. :) Thank you!

Let me know when it's ready to merge.

alecz20:
I defined filename, so it should pass the doctest now.

Once that is done, it is ready to merge unless someone else has anything else to change.

Thanks!

shazow:
@alecz20 Nah it still won't pass because example.com and /path/to/file don't exist, etc.

Honestly it's not worth trying to make these pass unless we can do it simply. I'd lean towards making the example simpler (stripping away all the cruft, like defining variables).

Perhaps this would be a better example for simplicity:

```
>>> http = urllib3.PoolManager()
>>> with http.request('GET', url, preload_content=False) as r, open(filename, 'wb') as fp:
>>> ....    shutil.copyfileobj(r, fp)
```

shazow:
Btw is this ready to merge?

alecz20:
I made a small typo correction and I think it's ready to be merged.

docs/index.rst
```diff
@@ -97,6 +97,18 @@ like buffering:
     >>> secondpart = b.read()
 
 
+The response can be treated as a file-like object.
+A file can be downloaded directly to a local file in a context without
+being saved in memory.
+
+.. doctest ::
```
sigmavirus24: Shouldn't this be:

```rst
.. doctest::
```

alecz20: it is `.. doctest ::` in the other sections. I am not sure exactly how it works.

sigmavirus24: `.. doctest ::` is a comment in reStructuredText, not a directive. It should be a directive.

sigmavirus24: In other words, those other sections are *wrong*

Lukasa:
This languished long enough and the notes on it aren't important enough to leave it unmerged. Of course, we had a merge conflict, so merged this by hand. Thanks @alecz20!

PR #532, opened 2015-01-10
Close connections properly

### LINKED ISSUE #529: urllib3 keeps connection open after failed fingerprint validation
1.  Launch a single-threaded HTTPS server, like the one from Werkzeug/Flask, with ad-hoc certificate generation.

2.  Paste this into a file called lol.py:

        from urllib3.poolmanager import PoolManager

        def x():
            # This is GitHub's certificate fingerprint. This function should not exit successfully.
            fingerprint = 'A0:C4:A7:46:00:ED:A7:2D:C0:BE:CB:9A:8C:B6:07:CA:58:EE:74:5E'  
            p = PoolManager(assert_fingerprint=fingerprint)
            r = p.request('GET', 'https://127.0.0.1:5000')
            p.clear()  # doesn't seem to matter

3.  Launch ``python -i lol.py``, in a Python 3 virtualenv with the latest PyPI-release of urllib3 installed.

4.  Execute ``x()`` once, an exception for the mismatching fingerprint is thrown.

5.  Execute ``x()`` again, the function doesn't terminate. Werkzeug is not responding to it because it is busy with another connection, but that one should've been closed.

Notes:

- If one kills the Python process where urllib3 is used, the connections are terminated and the server is usable again.

- This is Python-3-specific.

- This was found during usage with requests, which has the exact same problem. See https://github.com/untitaker/vdirsyncer/pull/161

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7534476-urllib3-keeps-connection-open-after-failed-fingerprint-validation?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

untitaker:
Fix #529

I couldn't set up the testcase locally so I'm going to use Travis.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -521,6 +521,43 @@ class TestSSL(SocketDummyServerTestCase):
         finally:
             timed_out.set()
 
+    def test_ssl_failed_fingerprint_verification(self):
+        def socket_handler(listener):
+            for i in range(2):
+                sock = listener.accept()[0]
+                ssl_sock = ssl.wrap_socket(sock,
+                                           server_side=True,
+                                           keyfile=DEFAULT_CERTS['keyfile'],
+                                           certfile=DEFAULT_CERTS['certfile'],
+                                           ca_certs=DEFAULT_CA)
+
+                ssl_sock.send(b'HTTP/1.1 200 OK\r\n'
+                              b'Content-Type: text/plain\r\n'
+                              b'Content-Length: 5\r\n\r\n'
+                              b'Hello')
+
+                ssl_sock.close()
+                sock.close()
+
+        self._start_server(socket_handler)
+        # GitHub's fingerprint. Valid, but not matching.
+        fingerprint = ('A0:C4:A7:46:00:ED:A7:2D:C0:BE:CB'
+                       ':9A:8C:B6:07:CA:58:EE:74:5E')
+
+        def request():
+            try:
+                pool = HTTPSConnectionPool(self.host, self.port,
+                                           assert_fingerprint=fingerprint)
+                response = pool.urlopen('GET', '/', preload_content=False,
+                                        timeout=Timeout(connect=1, read=0.001))
+                response.read()
+            finally:
+                pool.close()
+
+        self.assertRaises(SSLError, request)
+        # Should not hang, see https://github.com/shazow/urllib3/issues/529
+        self.assertRaises(SSLError, request)
+
 
 def consume_socket(sock, chunks=65536):
     while not sock.recv(chunks).endswith(b'\r\n\r\n'):
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -50,6 +50,7 @@ class SocketServerThread(threading.Thread):
     def __init__(self, socket_handler, host='localhost', port=8081,
                  ready_event=None):
         threading.Thread.__init__(self)
+        self.daemon = True
 
         self.socket_handler = socket_handler
         self.host = host
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -558,6 +558,14 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 conn = None
             raise SSLError(e)
 
+        except SSLError:
+            # Treat SSLError separately from BaseSSLError to preserve
+            # traceback.
+            if conn:
+                conn.close()
+                conn = None
+            raise
+
         except (TimeoutError, HTTPException, SocketError, ConnectionError) as e:
             if conn:
                 # Discard the connection for these exceptions. It will be


untitaker:
Committed separately to confirm that testcase fails without fix. Will rebase later.

t-8ch:
Wouldn't it be cleaner to add another clause to the already existing SSL except block?

untitaker:
My idea was that the user wants to have his connections closed for all kinds of exceptions.

untitaker:
Also it seems that on failing testcases, the build hangs. I added a fix for that too.

t-8ch:
Yeah. I also noticed that if we added it to the existing case we would wrap it again in a `SSLError` which does not sound right.

t-8ch:
There is also [`test_discards_connection_on_sslerror`](https://github.com/untitaker/urllib3/blob/1311dbdb45b27e01fb5ab4528573ade1770a0679/test/with_dummyserver/test_https.py#L377) which seems related.

untitaker:
I wonder why https://github.com/untitaker/urllib3/blob/1311dbdb45b27e01fb5ab4528573ade1770a0679/test/with_dummyserver/test_https.py#L205 doesn't fail though.

untitaker:
Ah, I see now. Maybe remove that other testcase and add just another clause to this one?

untitaker:
I suspect tornado's server is able to handle multiple connections?

sigmavirus24:
Closing every connection on any exception seems horribly backwards incompatible. I can't imagine why someone would want a connection after an exception but I *can* imagine someone, somewhere, expecting to be able to muck with it.

See obligatory [xkcd](https://xkcd.com/1172/).

shazow:
@untitaker Thank you for getting involved and contributing. <3
@t-8ch Thank you for helping figure this out!
@sigmavirus24 Thank you for reviewing this. :)

Ping me when you feel this is ready for merging.

untitaker:
Hmm, I think I'd have to get a proper testing environment first, as on my machine, nothing dummyserver-related works.

sigmavirus24:
@untitaker for what it's worth, I had similar problems and it was resolved by using a static DNS (like Google's). You might want to see if that helps.

untitaker:
Seems unrelated, but I've tried it anyway: No luck...

untitaker:
Funny thing is that ``python -m dummyserver.server`` works.

sigmavirus24:
I wonder if you could share some more details so I can try to reproduce. 

- OS
- Exact python version
- Are you using virtualenv?
- Anything else potentially relevant.

untitaker:
>OS

Archlinux

>Exact python version

Python 2.7, but it happens on all that I've installed.


>Are you using virtualenv?

Yes, through tox or otherwise

untitaker:
Actually only the "socketlevel" tests fail, and one "connectionpool"-one:

    ProtocolError: ('Connection aborted.', gaierror(-9, 'Address family for hostname not supported'))

t-8ch:
I am on the same platform as @untitaker and iirc it worked fine, so it is probably not the platform. I'll recheck later this evening.

untitaker:
Tried various things such as disabling ipv6 with a kernel parameter, and similar IPv6 related stuff.

untitaker:
Just found a PC where the tests work without problems.

@sigmavirus24 Are you sure that the connection should be kept open in other cases? It strikes me as blatantly wrong.

sigmavirus24:
> Closing every connection on any exception seems horribly backwards incompatible. I can't imagine why someone would want a connection after an exception but I can imagine someone, somewhere, expecting to be able to muck with it.

In other words, closing every failed connection and not making this a feature for 2.0.0 (in which we could make backwards incompatible changes) *may* lead to surprising behaviour for users. Granted this behaviour is already mildly surprising so it's a bit of a Catch-22. Mind you I didn't say it was the correct behaviour (as you seem to have understood).

untitaker:
I don't think I misunderstood you, but I can't imagine where people would actually rely on such behavior. Maybe it'd help to just close the socket, and not set the attribute to ``None``?

I wouldn't have a problem with postponing this bugfix to 2.0.

untitaker:
BTW please review, not sure how to write a sockethandler that would handle two requests.

sigmavirus24:
> I can't imagine where people would actually rely on such behavior. 

I can't imagine good reasons to do it either, but I can imagine people doing it because programmers never bother to consider if what they're doing is a bad idea so long as it just works.

shazow:
> Are you sure that the connection should be kept open in other cases?

I feel more comfortable only closing the connection in scenarios where it makes semantic sense. That is, if the failure is unrelated to this specific request and if further requests on this socket are likely to fail in the same way.

> BTW please review, not sure how to write a sockethandler that would handle two requests. The tests fail, but the test script I gave in #529 works.

Your approach looks reasonable to me. Not sure why it's failing. :/ Unfortunately I don't have the time to debug this right now, would appreciate it if anyone else can take a look. @kevinburke @Lukasa up for a challenge? :D

untitaker:
As suggested by @sigmavirus24, this patch now only catches SSLError. However, the tests still fail for no obvious reason.

sigmavirus24:
The tests fail on [python 2.6](https://travis-ci.org/shazow/urllib3/jobs/48185072) it seems because the `SSLError` is not raised as expected. If I remember correctly, `fingerprint` verification isn't available on 2.7 so you'll want to skip that test there. You may also have to disable retries for the `SSLError` to actually make it through to the the test suite (in those tests).

untitaker:
2.6 seems to fail differently because the assertRaises helper can't act as a contextmanager there... I fixed that. However, the other tests are just timing out...

t-8ch:
@sigmavirus24 afaik fingerprint verification should work everywhere

test/with_dummyserver/test_socketlevel.py
```diff
@@ -521,6 +521,47 @@ def socket_handler(listener):
...
+                sock.close()
+
+        self._start_server(socket_handler)
+        # GitHub's fingerprint. Valid, but not matching.
+        fingerprint = ('A0:C4:A7:46:00:ED:A7:2D:C0:BE:CB'
+                       ':9A:8C:B6:07:CA:58:EE:74:5E')
+
+        def request():
+            try:
+                pool = HTTPSConnectionPool(self.host, self.port,
```
t-8ch: Should we not use the same pools for both runs?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -521,6 +521,47 @@ def socket_handler(listener):
...
+                sock = listener.accept()[0]
+                ssl_sock = ssl.wrap_socket(sock,
+                                           server_side=True,
+                                           keyfile=DEFAULT_CERTS['keyfile'],
+                                           certfile=DEFAULT_CERTS['certfile'],
+                                           ca_certs=DEFAULT_CA)
+
+                buf = b''
+                while not buf.endswith(b'\r\n\r\n'):
+                    buf += ssl_sock.recv(65536)
```
t-8ch: This buffer does not get any data while running this test. I am not sure why. I assume fixing this will also fix the test.

sigmavirus24: Does endswith even work on bytes on Python 3?
-- 

t-8ch: I think it should. However I put in a `print()` in the loop an `buf` is just always empty.

t-8ch: Maybe a problem with TLS buffering data?

test/with_dummyserver/test_socketlevel.py
untitaker: In the linked issue the problem occurs even though the pool is not reused.

test/with_dummyserver/test_socketlevel.py
untitaker: >This buffer does not get any data while running this test

d'oh, that was it. The tests should pass now!

untitaker:
@shazow This is ready for review.

shazow:
@untitaker Wonderful, thank you! :)

@t-8ch @sigmavirus24 Thanks so much for your constructive feedback and helping push it through! <3

PR #534, opened 2015-01-12
Replaced concat/split by actual collection stuff

seocam:
Addresses #533

PR #544, opened 2015-02-05
Faster headers implementation

ml31415:
I had been trying to speed up urllib3 by monkeypatching with geventhttpclient. Unfortunately there is not too much gain, as lots of copying of response and header objects is done, which could largely be skipped, when the header and response objects would be sufficiently compatible and directly recognized and accepted as compatible, instead of being converted by lots of copying around.

So for the sake of better compatibility and speedups, I would have liked to directly create a header class - the standard urllib3.HTTPHeaderDict - by geventhttpclients c-parser when monkeypatching, which won't need any further processing from urllib3s response object. When I compared both header container implementations, turned out that the current urllib3 version isn't that fast as it could be. Below some timings comparing the old vs the new version, HTTPHeaderDict_ being the old version:

```python
print "Empty initializations"
%timeit h = HTTPHeaderDict_()
%timeit x = HTTPHeaderDict()

print "Initializations with data"
%timeit h = HTTPHeaderDict_(asdf='ddd', abc='100', ddd='200', efg='300')
%timeit x = HTTPHeaderDict(asdf='ddd', abc='100', ddd='200', efg='300')

from random import choice
from string import ascii_lowercase as asciis
hdrs = dict((''.join(choice(asciis) for _ in range(9)),
             ''.join(choice(asciis) for _ in range(9))) for _ in range(200))

print "Initializations with more data"
%timeit h = HTTPHeaderDict_(hdrs)
%timeit x = HTTPHeaderDict(hdrs)

h = HTTPHeaderDict_(asdf='ddd', abc='100', ddd='200', efg='300')
x = HTTPHeaderDict(asdf='ddd', abc='100', ddd='200', efg='300')

print "Fetching items"
%timeit h.__getitem__('abc')
%timeit x.__getitem__('abc')

print "Copying"
%timeit h.copy()
%timeit x.copy()

print "Adding headers"
%timeit h = HTTPHeaderDict_(); [h.add(k,v) for k, v in hdrs.iteritems()];
%timeit x = HTTPHeaderDict(); [x.add(k,v) for k, v in hdrs.iteritems()];

h = HTTPHeaderDict_(hdrs)
x = HTTPHeaderDict(hdrs)

print "Iteration of keys"
%timeit [k for k in h]
%timeit [k for k in x]

print "Iteration of values"
%timeit [h[k] for k in h]
%timeit [x[k] for k in x]

print "Getting keys"
%timeit h.keys()
%timeit x.keys()

print "Getting values"
%timeit h.values()
%timeit x.values()
```

For my machine, I get the following timings:
```
Empty initializations
100000 loops, best of 3: 4.38 µs per loop
1000000 loops, best of 3: 858 ns per loop
Initializations with data
100000 loops, best of 3: 11.3 µs per loop
100000 loops, best of 3: 12 µs per loop
Initializations with more data
1000 loops, best of 3: 252 µs per loop
1000 loops, best of 3: 235 µs per loop
Fetching items
100000 loops, best of 3: 2.34 µs per loop
1000000 loops, best of 3: 1 µs per loop
Copying
100000 loops, best of 3: 11.7 µs per loop
100000 loops, best of 3: 5.36 µs per loop
Adding headers
1000 loops, best of 3: 267 µs per loop
1000 loops, best of 3: 242 µs per loop
Iteration of keys
10000 loops, best of 3: 58.3 µs per loop
100000 loops, best of 3: 17.9 µs per loop
Iteration of values
1000 loops, best of 3: 706 µs per loop
1000 loops, best of 3: 271 µs per loop
Getting keys
10000 loops, best of 3: 44.6 µs per loop
100000 loops, best of 3: 2.25 µs per loop
Getting values
1000 loops, best of 3: 671 µs per loop
1000 loops, best of 3: 267 µs per loop
```
The new version passes all current tests, though it comes with one downside, which is worth discussing: It's not storing the original case of the header items. While the current version stores that, and therefore can restore the case if desired e.g. when iterating over the keys, this implementation is consequently using lower case, except when pretty printing the object. The standards require case insensitivity, so not preserving the case seems to me like an acceptable, maybe even desireable behaviour. Furthermore, also the current implementation has to drop the case information, when joining same header items together.

In terms of compatibility, I hope this should work with all python versions, though I mainly tested it on 2.7.

```
python test/test_collections.py
................
Ran 16 tests in 0.004s

OK
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 dev (master)
 ++++++++++++
 
+* Don't re-use connections which experienced an SSLError. (Issue #529)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ dev (master)
 
 * Add sha256 support for fingerprint verification. (Issue #540)
 
+* Fixed handling of header values containing commas. (Issue #533)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,8 @@ dev (master)
 
 * Don't re-use connections which experienced an SSLError. (Issue #529)
 
+* Don't fail when gzip decoding an empty stream. (Issue #535)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 dev (master)
 ++++++++++++
 
+* Pools can used as context managers. (Issue #545)
+
 * Don't re-use connections which experienced an SSLError. (Issue #529)
 
 * Don't fail when gzip decoding an empty stream. (Issue #535)
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ dev (master)
 
 * Don't fail when gzip decoding an empty stream. (Issue #535)
 
+* Add sha256 support for fingerprint verification. (Issue #540)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -239,6 +239,14 @@ class TestHTTPHeaderDict(unittest.TestCase):
         rep = "HTTPHeaderDict({'Cookie': 'foo, bar'})"
         self.assertEqual(repr(self.d), rep)
 
+    def test_items(self):
+        items = self.d.items()
+        self.assertEqual(len(items), 2)
+        self.assertEqual(items[0][0], 'Cookie')
+        self.assertEqual(items[0][1], 'foo')
+        self.assertEqual(items[1][0], 'Cookie')
+        self.assertEqual(items[1][1], 'bar')
+
     def test_items_preserving_case(self):
         # Should not be tested only in connectionpool
         HEADERS = {'Content-Length': '0', 'Content-type': 'text/plain',
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -222,7 +222,7 @@ class TestHTTPHeaderDict(unittest.TestCase):
         a = self.d[key]
         b = self.d.pop(key)
         self.assertEqual(a, b)
-        self.assertNotIn(key, self.d)
+        self.assertFalse(key in self.d)
         with self.assertRaises(KeyError):
             self.d.pop(key)
         dummy = object()
@@ -230,7 +230,7 @@ class TestHTTPHeaderDict(unittest.TestCase):
 
     def test_discard(self):
         self.d.discard('cookie')
-        self.assertNotIn('cookie', self.d)
+        self.assertFalse('cookie' in self.d)
         self.d.discard('cookie')
 
     def test_len(self):
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -223,10 +223,9 @@ class TestHTTPHeaderDict(unittest.TestCase):
         b = self.d.pop(key)
         self.assertEqual(a, b)
         self.assertFalse(key in self.d)
-        with self.assertRaises(KeyError):
-            self.d.pop(key)
+        self.assertRaises(KeyError, self.d.pop, key)
         dummy = object()
-        self.assertIs(dummy, self.d.pop(key, dummy))
+        self.assertTrue(dummy is self.d.pop(key, dummy))
 
     def test_discard(self):
         self.d.discard('cookie')
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -213,5 +213,12 @@ class TestHTTPHeaderDict(unittest.TestCase):
         rep = "HTTPHeaderDict({'A': 'foo, bar'})"
         self.assertEqual(repr(self.d), rep)
 
+    def test_items_preserving_case(self):
+        # Should not be tested only in connectionpool
+        HEADERS = {'Content-Length': '0', 'Content-type': 'text/plain',
+                    'Server': 'TornadoServer/1.2.3'}
+        h = dict(HTTPHeaderDict(HEADERS).items())
+        self.assertEqual(HEADERS, h) # to preserve case sensitivity        
+
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -156,11 +156,11 @@ class TestHTTPHeaderDict(unittest.TestCase):
         d.add('a', 'asdf')
         self.assertEqual(d['a'], 'foo, bar, asdf')
         
-    def test_update_add(self):
-        self.d.update_add([('c', '100'), ('c', '200'), ('c', '300')], c='400')
+    def test_extend(self):
+        self.d.extend([('c', '100'), ('c', '200'), ('c', '300')], c='400')
         self.assertEqual(self.d['c'], '100, 200, 300, 400')
 
-        self.d.update_add(dict(a='asdf'))
+        self.d.extend(dict(a='asdf'))
         self.assertEqual(self.d['a'], 'foo, bar, asdf')
         
         class NonMappingHeaderContainer(object):
@@ -175,7 +175,7 @@ class TestHTTPHeaderDict(unittest.TestCase):
                 return self._data[key]
 
         header_object = NonMappingHeaderContainer(e='foofoo')
-        self.d.update_add(header_object)
+        self.d.extend(header_object)
         self.assertEqual(self.d['e'], 'foofoo')
 
     def test_getlist(self):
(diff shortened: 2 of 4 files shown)


wernerspiegelpup:
:+1:

sigmavirus24:
> The new version passes all current tests, though it comes with one downside, which is worth discussing: It's not storing the original case of the header items.

I for one think this is unacceptable. We're all aware of the specification, but preserving the data as it is received is important and something I'm certain I've observed people relying upon.

sigmavirus24:
Further, in its current state, this PR isn't compatible with Python 3 (according to Travis CI) and doesn't achieve 100% test coverage. Those are 2 things that are going to block this more significantly in the near term than anything else.

ml31415:
Just extended the tests for full coverage. Have some issues with the full test run locally, so I haven't uploaded it right now. In case the suggested version would otherwise be fine, I'll fix it for Python3, too.

sigmavirus24:
@ml31415 I'd rather you fix it and push for Python 3 so I can pull it and verify your benchmarks.

ml31415:
Fixing the case preservation? I suppose that would kill most of the speed gains.

sigmavirus24:
@ml31415 Fixing the code on Python 3: https://travis-ci.org/shazow/urllib3/jobs/49604628

ml31415:
Excellent, now it's failing the coverage check due to the version fork. Any idea how to fix this?

ml31415:
I modified the changes, so that the original case information is preserved. After some tuning, the speed improvements are similar as the previous version.

```
Empty initializations
1000000 loops, best of 3: 1.01 µs per loop
Initializations with data
100000 loops, best of 3: 12.8 µs per loop
Initializations with more data
1000 loops, best of 3: 258 µs per loop
Fetching items
1000000 loops, best of 3: 940 ns per loop
Copying
100000 loops, best of 3: 6.29 µs per loop
Adding headers
1000 loops, best of 3: 237 µs per loop
Iteration of keys
100000 loops, best of 3: 17 µs per loop
Iteration of values
1000 loops, best of 3: 239 µs per loop
Getting keys
100000 loops, best of 3: 2.07 µs per loop
Getting values
1000 loops, best of 3: 251 µs per loop
```

urllib3/_collections.py
```diff
@@ -156,43 +183,95 @@ def add(self, key, value):
...
+        if new_vals is not vals:
+            # The item was not inserted, as there was a previous one
+            if isinstance(vals, list):
+                # If already several items got inserted, we have a list
+                vals.append(val)
+            else:
+                # Only one item so far, need to convert the tuple to list
+                _dict_setitem(self, key_lower, [vals[0], vals[1], val])
+
+    def update_add(*args, **kwds):
```
shazow: Hmm, `update_add` is a weird name. Maybe `load`?

shazow:
Looks good overall. `update_add` is the messiest part, but I'm not sure how to clean it up. Renaming would be appreciated if you feel it makes sense.

Otherwise if everyone is onboard, I'm happy to merge. :)

ml31415:
I named it like that on purpose, in order not to mess with the previous update implementation. Previous update replaces items when hitting double occurances, update_add extends entries when required. It's meant for replacing for-loops containing headerobj.add(key, value), like seen in the request module, or as a general import function of arbitrary other header objects. I also added the from_httplib constructor for that reason.

shazow:
Yup, I got all that. :) I just mean that the name "update_add" is confusing. I understand why it exists, and I agree it's necessary.

ml31415:
I suppose I'm quite uncreative with namings :) Better suggestions cordially welcome!

shazow:
@Lukasa @sigmavirus24 Thoughts?

test/test_collections.py
```diff
@@ -168,6 +195,10 @@ def test_equal(self):
         self.assertEqual(self.d, b)
         c = [('a', 'foo, bar')]
         self.assertNotEqual(self.d, c)
+        
+        # For some reason, the test above doesn't run into the __eq__ function,
+        # lacking test coverage for line 165, so comparing directly here
+        self.assertFalse(self.d == 2)
```
sigmavirus24: Did you mean

```py
self.assertNotEqual(len(self.d), 2)
```

?

Lukasa:
I think extend works for me.

test/test_collections.py
ml31415: Without that line, the testrun reported missing coverage for _collections#163, and I had expected, this not equal comparison the line above would also call __eq__ finally, but for some reason it didn't.

ml31415:
What remaind a bit undefined is, when to return case-preserved headers and when not. E.g. `keys()` now returns the actual lower case headers, `items()` restores it. To me it would have been preferable, to have separate case-preserving functions, but the others are already expected to preserve case in some cases, e.g. `test_connectionpool`, line 621.

urllib3/_collections.py
shazow: Let's go with `extend`. :)

ml31415:
Here some recent timings:

New:
```
Empty initializations
1000000 loops, best of 3: 847 ns per loop
Initializations with data
100000 loops, best of 3: 10.6 µs per loop
Initializations with more data
1000 loops, best of 3: 213 µs per loop
Fetching items
1000000 loops, best of 3: 938 ns per loop
Getting list hit
1000000 loops, best of 3: 1.18 µs per loop
Getting list miss
1000000 loops, best of 3: 1.16 µs per loop
Copying
100000 loops, best of 3: 5.45 µs per loop
Adding headers
1000 loops, best of 3: 209 µs per loop
Iteration of keys
100000 loops, best of 3: 17.2 µs per loop
Iteration of values
1000 loops, best of 3: 227 µs per loop
Getting keys
100000 loops, best of 3: 2.1 µs per loop
Getting values
1000 loops, best of 3: 231 µs per loop
```
Current:
```
100000 loops, best of 3: 4.46 µs per loop
Initializations with data
100000 loops, best of 3: 11.3 µs per loop
Initializations with more data
1000 loops, best of 3: 249 µs per loop
Fetching items
100000 loops, best of 3: 2.5 µs per loop
Getting list hit
100000 loops, best of 3: 7.11 µs per loop
Getting list miss
100000 loops, best of 3: 7.05 µs per loop
Copying
100000 loops, best of 3: 12 µs per loop
Adding headers
1000 loops, best of 3: 273 µs per loop
Iteration of keys
10000 loops, best of 3: 58.9 µs per loop
Iteration of values
1000 loops, best of 3: 648 µs per loop
Getting keys
10000 loops, best of 3: 44.7 µs per loop
Getting values
1000 loops, best of 3: 639 µs per loop
```

ml31415:
In the header implementation for geventhttpclient I had also added a check for the header field name like
```python
MULTIPLE_HEADERS_ALLOWED = set(['cookie', 'set-cookie', 'set-cookie2'])
```
in order to only allow duplicate fields for certain fields. That's  probably closest to what specs say. The performance degradation for the check would be neglectable. What's your thoughts on that?

shazow:
@ml31415 +1, I like the idea of pulling those out into a constant so that it's more sensible and self-documenting for why the code is branching.

shazow:
FYI we just merged #534, could you rebase against the latest master and see if you can make this PR work with the fixes? :) Thank you!

shazow:
Thank you for rebasing. :) FYI some tests are failing due to the new header splitting logic.

ml31415:
Yes, saw that. Fix is on the way.

shazow:
:D :fish_cake:

ml31415:
Did I already mention that py2.6 sucks?

ml31415:
Finally. I added the `from_httplib` constructor again for python 2. Python 3 returns all headers including duplicate ones simply when calling `items()` on the message object, so no special factory required for conversion.

Python 2 `httplib.HTTPMessage` also offers `items()`, but as we painfully know there the duplicates are missing. The `getheaders()` method of the message object is quite slow, as it iterates over all the preserved raw data and puzzles everything together again. So it would be possible, to rewrite the parsing of the raw lines from the message object, so all duplicate headers would have been preserved, not only the cookies. But this would repeat nearly all parsing efforts done already, when the dictionary got filled by the message object. So it would be slow again.

That's why I suppose it's actually the best way is just to copy the duplicate header items and ignore possible other duplicate headers for python 2. This is mainly the same solution, that was already implemented in #533, except that the logic moved to the header module.

CHANGES.rst
```diff
@@ -4,12 +4,16 @@ Changes
 dev (master)
 ++++++++++++
 
+* Pools can used as context managers. (Issue #545)
```
shazow: Hmm I think these shouldn't be here, maybe try the latest master?

shazow:
Minor merge error in CHANGES.rst, but otherwise looks good. Happy to merge after that's fixed (please ping), if nobody objects. :)

ml31415:
Merged master again. Hope that shall do now.

shazow:
Yay thank you! Also, you're welcome to add yourself to the CONTRIBUTORS.txt for this lovely improvement. :)

ml31415:
Ah, too much honor ;)

sigmavirus24:
@ml31415 please take a look at #561

PR #549, opened 2015-02-14
Fix file descriptor leak due to stack frame reference

### LINKED ISSUE #548: File descriptor leak with HTTPS proxies via CONNECT
It looks like when making HTTPS requests through an HTTP proxy, there is a file descriptor leak if an error occurs (e.g. socket timeout or 404 error) during httplib.HTTPConnection._tunnel. Tested in versions 1.9, 1.9.1, 1.10.

The way I've been testing is to use netcat to listen for connections but not respond to CONNECT, eg.:

    $ nc -l localhost 8000

Then making a request with that "proxy":

    In [1]: import urllib3
    In [2]: pm = urllib3.ProxyManager('http://localhost:8000/')
    In [3]: r = pm.request('GET', 'https://httpbin.org/ip', timeout=1.0)
    ---------------------------------------------------------------------------
    MaxRetryError                             Traceback (most recent call last)

At that point, `lsof` shows multiple connections (one per retry) in CLOSE_WAIT state, which stay that way permanently.

It looks like the issue is related to a circular reference preventing a socket from being cleaned up properly. For example, this commit fixes the issue: https://github.com/ianpreston/urllib3/commit/a0953a93c877468d9b60a6785c767c5ac31d0dc2 , as does wrapping the `raise` statements in `Retry.increment` with a "finally: del _stacktrace".

The workaround above is working for now but I'd like to find the root cause, or at least a workaround that doesn't break stack traces. Any idea where to start looking? Thanks

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8433025-file-descriptor-leak-with-https-proxies-via-connect?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ianpreston:
Fixes #548

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -588,14 +588,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 conn.close()
                 conn = None
 
-            stacktrace = sys.exc_info()[2]
             if isinstance(e, SocketError) and self.proxy:
                 e = ProxyError('Cannot connect to proxy.', e)
             elif isinstance(e, (SocketError, HTTPException)):
                 e = ProtocolError('Connection aborted.', e)
 
-            retries = retries.increment(method, url, error=e,
-                                        _pool=self, _stacktrace=stacktrace)
+            retries = retries.increment(method, url, error=e, _pool=self,
+                                        _stacktrace=sys.exc_info()[2])
             retries.sleep()
 
             # Keep track of the error for the retry warning.
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -82,6 +82,7 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
             err = _
             if sock is not None:
                 sock.close()
+                sock = None
 
     if err is not None:
         raise err


sigmavirus24:
So it'd be excellent if there were a test for this, but *I* don't know how you would begin to test this in the context of urllib3's test suite. That said, @shazow or @Lukasa should definitely be more helpful in that regard. The one thing I'd request is that you leave a comment near the `sock = None` line explaining the relation to issue #548 

Also, I think those test failures are purely Travis' fault but if they continue, we may need to look further into why they're failing.

Lukasa:
I suspect the way to test this is to test the cause of what we're fixing. Actually demonstrating a reference cycle in a test is a bit tricky, but you can test that the frame local variables don't have the stacktrace. Bit of a tricky test to write, but if we think it's important it's totally do-able.

shazow:
Looks safe to me. :) While a test is always appreciated, I'm not sure exactly how to approach it in a way that scales across Python implementations. Happy to merge it as-is, but if you find a test you'd like to add then it's very welcome!

PR #553, opened 2015-02-23
Failure to raise max retry error in pool manager.

wtolson:
When retries is set to something like `Retry(total=None, redirect=1)`, a `MaxRetryError` is never raised, even in the case of a infinite redirect loop. This PR makes `PoolManager` act consistently with `HTTPConnectionPool` in respect to this behavior.


shazow:
Ah good catch, thank you!

shazow:
@wtolson Btw, thank you for the perfect pull request. :) You're welcome to add yourself to our CONTRIBUTORS.txt, and I hope to see more contributions from you.

Issue #557, opened 2015-03-02
Emit a warning on platforms where we can't configure OpenSSL as we'd like


alex:
Specifically this means:

* Python < 2.7.9 (Probably something for older Python 3s, but I don't know that off hand)
* and no pyOpenSSL + pyasn1 + ngd-httpsclient

On these platform we can't do all the configuration we'd like, including the fix for verification that we've been discussing. It'd be great if users know they should upgrade their Python (or install the appropriate optional dependencies)

Lukasa:
+1. This sounds like a good idea to me.

dstufft:
This should probably actually detect whether there is a ``SSLContext`` object or not instead of inspecting Python versions.

alex:
Great point.

On Mon, Mar 2, 2015 at 7:08 AM, Donald Stufft <notifications@github.com>
wrote:

> This should probably actually detect whether there is a SSLContext object
> or not instead of inspecting Python versions.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shazow/urllib3/issues/557#issuecomment-76727578>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

sigmavirus24:
:+1:

shazow:
Sounds good. Agree about checking for features rather than versions when possible.

Does anyone want to take this task?

Lukasa:
I'll do it.

Lukasa:
Alright, comments requested on #558.

Lukasa:
Resolved in #558. =)

agriffis:
Could someone provide pointers for "On these platform we can't do all the configuration we'd like, including the fix for verification that we've been discussing."  I'd like to fully understand the details of the warning, but I haven't been able to trace through the history beyond this statement.

Lukasa:
Sure. =)

The SSLContext object is the modern way to configure TLS settings in Python. There are a number of things the SSLContext can do that platforms without it cannot do, mostly to do with disabling certain features. For example, you cannot disable TLS compression support without it: disabling TLS compression is necessary to avoid the CRIME attack. We also can't do SNI without an SSLContext.

Generally speaking, platforms without SSLContexts are dramatically less secure than those that have them.

agriffis:
[...]

PR #558, opened 2015-03-02
Emit warnings when SSLContext isn't available.

Lukasa:
This change emits warnings when the `SSLContext` object isn't available.

To restrict the scope of the warning to situations only when it really isn't available, this warning actually gets emitted only if the weird mock object we use is actually being used. In no other circumstances should this warning be emitted.

The test is pretty ghetto, but it turns out to be kinda a pain in the ass to test this really properly. If you'd like me to, I can take another swing at it.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -111,6 +111,8 @@ Once you find your root certificate file::
         ...
 
 
+.. _pyopenssl:
+
 OpenSSL / PyOpenSSL
 -------------------
 
@@ -136,6 +138,8 @@ Now you can continue using urllib3 as you normally would.
 
 For more details, check the :mod:`~urllib3.contrib.pyopenssl` module.
 
+
+.. _insecurerequestwarning:
 
 InsecureRequestWarning
 ----------------------
@@ -166,3 +170,20 @@ warnings to your own log::
 
 Capturing the warnings to your own log is much preferred over simply disabling
 the warnings.
+
+SSLConfigurationWarning
+-----------------------
+
+.. versionadded:: 1.11
+
+Certain Python platforms have limitations in their ``ssl`` module that limits
+the configuration that ``urllib3`` can apply. In particular, this can cause
+HTTPS requests that would succeed on more featureful platforms to fail.
+
+If you encounter this warning, it is strongly recommended you upgrade to a
+newer Python version, or that you use pyOpenSSL as described in the
+:ref:`pyopenssl` section.
+
+If you know what you are doing and would like to disable this and other
+warnings, please consult the :ref:`insecurerequestwarning` section for
+instructions on how to handle the warnings.
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -176,9 +176,10 @@ SSLConfigurationWarning
 
 .. versionadded:: 1.11
 
-Certain Python platforms have limitations in their ``ssl`` module that limits
-the configuration that ``urllib3`` can apply. In particular, this can cause
-HTTPS requests that would succeed on more featureful platforms to fail.
+Certain Python platforms (specifically, versions of Python earlier than 2.7.9)
+have restrictions in their ``ssl`` module that limit the configuration that
+``urllib3`` can apply. In particular, this can cause HTTPS requests that would
+succeed on more featureful platforms to fail.
 
 If you encounter this warning, it is strongly recommended you upgrade to a
 newer Python version, or that you use pyOpenSSL as described in the
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -171,7 +171,7 @@ warnings to your own log::
 Capturing the warnings to your own log is much preferred over simply disabling
 the warnings.
 
-SSLConfigurationWarning
+InsecurePlatformWarning
 -----------------------
 
 .. versionadded:: 1.11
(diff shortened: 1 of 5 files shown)


Lukasa:
Oh, this resolves #557.

docs/security.rst
```diff
@@ -166,3 +170,20 @@ warnings to your own log::
...
+the configuration that ``urllib3`` can apply. In particular, this can cause
+HTTPS requests that would succeed on more featureful platforms to fail.
+
+If you encounter this warning, it is strongly recommended you upgrade to a
+newer Python version, or that you use pyOpenSSL as described in the
+:ref:`pyopenssl` section.
+
+If you know what you are doing and would like to disable this and other
+warnings, please consult the :ref:`insecurerequestwarning` section for
+instructions on how to handle the warnings.
```
alex: Maybe this should specifically mention `2.7.x` where `x < 9`, that's going to be the most common case that hits this I think.

urllib3/util/ssl_.py
```diff
@@ -195,6 +197,16 @@ def create_urllib3_context(ssl_version=None, cert_reqs=ssl.CERT_REQUIRED,
     """
     context = SSLContext(ssl_version or ssl.PROTOCOL_SSLv23)
 
+    if hasattr(context, 'urllib3_wrapper_context'):
```
dstufft: Instead of doing this hasattr for a special attribute, would it not make sense to just move this warning to the mock ``SSLContext().wrap_socket()``?

docs/security.rst
dstufft: I agree that you might want to mention under what cases this is likely to happen, <2.7.9 and <3.2.

urllib3/util/ssl_.py
Lukasa: Yeah, this is likely a bit better, but tricky to actually get to test. To test it I need to get the SSL module to raise an import error on the SSLContext. Any ideas how I go about that?

dstufft:
```python
import ssl
del ssl.SSLContext
from ssl import SSLContext
```

Lukasa: Presumably I'm going to need a new test file for that?

urllib3/util/ssl_.py
```diff
@@ -195,6 +197,16 @@ def create_urllib3_context(ssl_version=None, cert_reqs=ssl.CERT_REQUIRED,
...
+        warnings.warn(
+            'A true SSLContext object is not available. This prevents urllib3 '
+            'from configuring SSL appropriately. This may cause certain SSL '
+            'connections to fail. For more information, see '
+            'http://urllib3.readthedocs.org/en/latest/security.html'
```
kevinburke: This link is available over HTTPS

urllib3/util/ssl_.py
```diff
@@ -195,6 +197,16 @@ def create_urllib3_context(ssl_version=None, cert_reqs=ssl.CERT_REQUIRED,
     """
     context = SSLContext(ssl_version or ssl.PROTOCOL_SSLv23)
 
+    if hasattr(context, 'urllib3_wrapper_context'):
+        warnings.warn(
+            'A true SSLContext object is not available. This prevents urllib3 '
+            'from configuring SSL appropriately. This may cause certain SSL '
```
kevinburke: two "This"'s in a row is a little awkward, how about "from configuring SSL appropriately and may cause certain SSL connections to fail"

urllib3/util/ssl_.py
dstufft: I'm not sure what you mean by test file in this context. You can restore the ``SSLContext`` after the test has been ran, however since this is going to do different things at the module level based on an import error you're also going to need to delete this module from ``sys.modules`` to trigger another import. I'm not sure if it's worth the machinations required to test that.

Another reason to do the warning there, is if this API goes away and urllib3 starts using the ``SSLContext`` directly then this warning has less likely to get lost in the shuffle.

Lukasa: Actually, this is immune from test coverage, and I don't believe there's any real value obtained in including a test for it, so I simply removed it.

urllib3/exceptions.py
```diff
@@ -144,6 +144,11 @@ class ResponseError(HTTPError):
     SPECIFIC_ERROR = 'too many {status_code} error responses'
 
 
+class SSLConfigurationWarning(HTTPWarning):
```
shazow: Should this be a SecurityWarning? (See exceptions below)

Also, maybe call it `InsecurePlatformWarning`?

Lukasa: It's not clear to me that this makes the platform less *secure*, just less functional.

dstufft: I don't think TLS compression is disabled on Pythons without ``SSLContext`` nor is it possible to disable SSLv3 without ``SSLContext``. Python 2.7 (at least some versions) have SSLv2 disabled, I'm not sure if 2.6 does or if all versions of 2.7 does.

I think it's pretty fair to say that without ``SSLContext`` the security of the connection is likely to be less secure unless the platform has disabled these items inside of OpenSSL itself.

Lukasa: Alrighty, updated.

shazow:
Looks good to me. Let me know when you deem it ready for merging. :)

Lukasa:
I'm happy to have it merged, but let @dstufft give his ok before we charge ahead. =)

urllib3/util/ssl_.py
```diff
@@ -40,6 +41,7 @@
 
     class SSLContext(object):  # Platform-specific: Python 2 & 3.1
         supports_set_ciphers = sys.version_info >= (2, 7)
+        urllib3_wrapper_context = True
```
dstufft: I don't think you need this anymore.

dstufft:
One small comment, but otherwise looks good to me.

Lukasa:
Dealt with. =)

dstufft:
txprm

PR #560, opened 2015-03-04
implement generator for chunked responses

### LINKED ISSUE #550: [rfe] provide access to chunks when server responds with `Transfer-Encoding: chunked`
I've opened [an issue](https://github.com/kennethreitz/requests/issues/2449) at [requests](https://github.com/kennethreitz/requests) but was advised to move it here. There it goes:

When server responds with `Transfer-Encoding: chunked`, it would be nice if `urllib3` provided access to chunks sent by server. Therefore upper layers, `requests`, could process those encoded chunks and developers wouldn't need to decode those chunks.

More info:
http://en.wikipedia.org/wiki/Chunked_transfer_encoding
http://mihai.ibanescu.net/chunked-encoding-and-python-requests

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8681938-rfe-provide-access-to-chunks-when-server-responds-with-transfer-encoding-chunked?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

TomasTomecek:
Fixes #550

So, I have figured out first, proof of concept, very hacky, very ugly implementation. I would like to get some feedback.

(Bear in mind that I haven't run a single test yet)

### How to test

I have used this code to generate chunked responses: https://gist.github.com/josiahcarlson/3250376

Client code:

```python
import urllib3
http = urllib3.PoolManager()
r = http.request('GET', 'http://localhost:8080/')
for x in r.stream():
    print x,
```

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -257,6 +257,35 @@ disabled individually.
 See the :class:`~urllib3.util.retry.Retry` definition for more details.
 
 
+Stream
+------
+
+You may also stream your response and get data as they come (e.g. when using
+``transfer-encoding: chunked``). In this case, method
+:func:`~urllib3.response.HTTPResponse.stream` will return generator.
+
+::
+
+    >>> import urllib3, datetime
+    >>> http = urllib3.PoolManager()
+
+    >>> # https://gist.github.com/josiahcarlson/3250376
+    >>> # server which responds with chunked encoding responses and
+    >>> # generates chunks every second
+    >>> r = http.request("GET", "http://localhost:8080/")
+
+    >>> r.getheader("transfer-encoding")
+    'chunked'
+
+    >>> for chunk in http.request("GET", "http://localhost:8080/").stream():
+          print "[%s] %s" % (datetime.datetime.now().strftime("%H:%M:%S.%f"), chunk)
+    [10:57:35.650812] this is chunk: 0
+    [10:57:36.652022] this is chunk: 1
+    [10:57:37.653238] this is chunk: 2
+    [10:57:38.654431] this is chunk: 3
+    [10:57:39.655601] this is chunk: 4
+
+
 Foundation
 ----------
 
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -2,8 +2,12 @@ import unittest
 
 from io import BytesIO, BufferedReader
 
+try:
+    import http.client as httplib
+except ImportError:
+    import httplib
 from urllib3.response import HTTPResponse
-from urllib3.exceptions import DecodeError
+from urllib3.exceptions import DecodeError, ResponseNotChunked
 
 
 from base64 import b64decode
@@ -156,10 +160,6 @@ class TestResponse(unittest.TestCase):
 
     def test_io(self):
         import socket
-        try:
-            from http.client import HTTPResponse as OldHTTPResponse
-        except:
-            from httplib import HTTPResponse as OldHTTPResponse
 
         fp = BytesIO(b'foo')
         resp = HTTPResponse(fp, preload_content=False)
@@ -174,7 +174,7 @@ class TestResponse(unittest.TestCase):
 
         # Try closing with an `httplib.HTTPResponse`, because it has an
         # `isclosed` method.
-        hlr = OldHTTPResponse(socket.socket())
+        hlr = httplib.HTTPResponse(socket.socket())
         resp2 = HTTPResponse(hlr, preload_content=False)
         self.assertEqual(resp2.closed, False)
         resp2.close()
@@ -406,11 +406,200 @@ class TestResponse(unittest.TestCase):
         self.assertEqual(next(stream), b'o')
         self.assertRaises(StopIteration, next, stream)
 
+    def test_mock_transfer_encoding_chunked(self):
+        stream = [b"fo", b"o", b"bar"]
+        fp = MockChunkedEncodingResponse(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+
+        i = 0
+        for c in resp.stream():
+            self.assertEqual(c, stream[i])
+            i += 1
+
+    def test_mock_transfer_encoding_chunked_custom_read(self):
+        stream = [b"foooo", b"bbbbaaaaar"]
+        fp = MockChunkedEncodingResponse(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+        expected_response = [b'fo', b'oo', b'o', b'bb', b'bb', b'aa', b'aa', b'ar']
+        response = list(resp.read_chunked(2))
+        if getattr(self, "assertListEqual", False):
+            self.assertListEqual(expected_response, response)
+        else:
+            for index, item in enumerate(response):
+                v = expected_response[index]
+                self.assertEqual(item, v)
+
+    def test_mock_transfer_encoding_chunked_unlmtd_read(self):
+        stream = [b"foooo", b"bbbbaaaaar"]
+        fp = MockChunkedEncodingResponse(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+        if getattr(self, "assertListEqual", False):
+            self.assertListEqual(stream, list(resp.read_chunked()))
+        else:
+            for index, item in enumerate(resp.read_chunked()):
+                v = stream[index]
+                self.assertEqual(item, v)
+
+    def test_read_not_chunked_response_as_chunks(self):
+        fp = BytesIO(b'foo')
+        resp = HTTPResponse(fp, preload_content=False)
+        r = resp.read_chunked()
+        self.assertRaises(ResponseNotChunked, next, r)
+
+    def test_invalid_chunks(self):
+        stream = [b"foooo", b"bbbbaaaaar"]
+        fp = MockChunkedInvalidEncoding(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+        self.assertRaises(httplib.IncompleteRead, next, resp.read_chunked())
+
+    def test_chunked_response_without_crlf_on_end(self):
+        stream = [b"foo", b"bar", b"baz"]
+        fp = MockChunkedEncodingWithoutCRLFOnEnd(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+        if getattr(self, "assertListEqual", False):
+            self.assertListEqual(stream, list(resp.stream()))
+        else:
+            for index, item in enumerate(resp.stream()):
+                v = stream[index]
+                self.assertEqual(item, v)
+
+    def test_chunked_response_with_extensions(self):
+        stream = [b"foo", b"bar"]
+        fp = MockChunkedEncodingWithExtensions(stream)
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
+        if getattr(self, "assertListEqual", False):
+            self.assertListEqual(stream, list(resp.stream()))
+        else:
+            for index, item in enumerate(resp.stream()):
+                v = stream[index]
+                self.assertEqual(item, v)
+
     def test_get_case_insensitive_headers(self):
         headers = {'host': 'example.com'}
         r = HTTPResponse(headers=headers)
         self.assertEqual(r.headers.get('host'), 'example.com')
         self.assertEqual(r.headers.get('Host'), 'example.com')
+
+
+class MockChunkedEncodingResponse(object):
+
+    def __init__(self, content):
+        """
+        content: collection of str, each str is a chunk in response
+        """
+        self.content = content
+        self.index = 0  # This class iterates over self.content.
+        self.closed = False
+        self.cur_chunk = b''
+        self.chunks_exhausted = False
+
+    @staticmethod
+    def _encode_chunk(chunk):
+        return '%X\r\n%s\r\n' % (len(chunk), chunk.decode())
+
+    def _pop_new_chunk(self):
+        if self.chunks_exhausted:
+            return b""
+        try:
+            chunk = self.content[self.index]
+        except IndexError:
+            chunk = b''
+            self.chunks_exhausted = True
+        else:
+            self.index += 1
+        encoded_chunk = self._encode_chunk(chunk)
+        return encoded_chunk.encode()
+
+    def pop_current_chunk(self, amt=-1, till_crlf=False):
+        if amt > 0 and till_crlf:
+            raise ValueError("Can't specify amt and till_crlf.")
+        if len(self.cur_chunk) <= 0:
+            self.cur_chunk = self._pop_new_chunk()
+        if till_crlf:
+            try:
+                i = self.cur_chunk.index(b"\r\n")
+            except ValueError:
+                # No CRLF in current chunk -- probably caused by encoder.
+                self.cur_chunk = b""
+                return b""
+            else:
+                chunk_part = self.cur_chunk[:i+2]
+                self.cur_chunk = self.cur_chunk[i+2:]
+                return chunk_part
+        elif amt <= -1:
+            chunk_part = self.cur_chunk
+            self.cur_chunk = b''
+            return chunk_part
+        else:
+            try:
+                chunk_part = self.cur_chunk[:amt]
+            except IndexError:
+                chunk_part = self.cur_chunk
+                self.cur_chunk = b''
+            else:
+                self.cur_chunk = self.cur_chunk[amt:]
+            return chunk_part
+
+    def readline(self):
+        return self.pop_current_chunk(till_crlf=True)
+
+    def read(self, amt=-1):
+        return self.pop_current_chunk(amt)
+
+    def flush(self):
+        # Python 3 wants this method.
+        pass
+
+    def close(self):
+        self.closed = True
+
+
+class MockChunkedInvalidEncoding(MockChunkedEncodingResponse):
+
+    def _encode_chunk(self, chunk):
+        return 'ZZZ\r\n%s\r\n' % chunk.decode()
+
+
+class MockChunkedEncodingWithoutCRLFOnEnd(MockChunkedEncodingResponse):
+
+    def _encode_chunk(self, chunk):
+        return '%X\r\n%s%s' % (len(chunk), chunk.decode(),
+            "\r\n" if len(chunk) > 0 else "")
+
+
+class MockChunkedEncodingWithExtensions(MockChunkedEncodingResponse):
+
+    def _encode_chunk(self, chunk):
+        return '%X;asd=qwe\r\n%s\r\n' % (len(chunk), chunk.decode())
+
+
+class MockSock(object):
+    @classmethod
+    def makefile(cls, *args, **kwargs):
+        return
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -1,9 +1,15 @@
+try:
+    import http.client as httplib
+except ImportError:
+    import httplib
 import zlib
 import io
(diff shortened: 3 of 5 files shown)


urllib3/response.py
```diff
@@ -351,3 +363,77 @@ def readinto(self, b):
         else:
             b[:len(temp)] = temp
             return len(temp)
+
+    def _read_chunked(self, amt):
```
Lukasa: `amt` should be an optional parameter, defaulting to `None`.

urllib3/response.py
```diff
@@ -351,3 +363,77 @@ def readinto(self, b):
...
+        def readline():
+            r = ""
+            while True:
+                ch = self._fp._safe_read(1)
+                r += ch
+                if r.endswith("\r\n"):
+                    return r
```
Lukasa: Is this really needed? My suspicion is that httplib/http.client will provide a socket with readline already defined on it (by having previously called `socket.makefile` on it).

Lukasa:
This is nice!

I don't like that iterating over responses will fail if they're not chunked. Should we fall back to using `stream()` with a well-defined read size?

urllib3/response.py
TomasTomecek: I took the prototype directly [from httplib](https://hg.python.org/cpython/file/2.7/Lib/httplib.py#l590). Will fix.

urllib3/response.py
TomasTomecek: You are right! (how could I miss that)

On the other hand, this looks pretty ugly:
```python
self._fp.fp.readline()
```
Is there a better way?

urllib3/response.py
```diff
@@ -117,8 +118,17 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
...
             self._fp = body
 
-        if preload_content and not self._body:
-            self._body = self.read(decode_content=decode_content)
+        # are we using the chunked-style of transfer encoding?
+        tr_enc = self.headers.get('transfer-encoding')
+        if tr_enc and tr_enc.lower() == "chunked":
+            self.chunked = True
+            self.chunk_left = None
+        else:
```
shazow: Style-wise, I'd prefer to start with `self.chunked = False` in the beginning, then override as necessary (I tend to avoid empty `else:` statements when possible).

urllib3/response.py
```diff
@@ -178,6 +188,8 @@ def read(self, amt=None, decode_content=None, cache_content=False):
             after having ``.read()`` the file object. (Overridden if ``amt`` is
             set.)
         """
+        if self.chunked:
+            return
```
shazow: Does .read() simply stop working if there is a chunked header? Isn't that surprising?

shazow:
Perhaps this would work best in conjunction with #436, then we can use a generator interface in general?

urllib3/response.py
```diff
@@ -351,3 +363,71 @@ def readinto(self, b):
...
+                yield ''.join(value)
+            else:
+                yield self._fp._safe_read(chunk_left)
+                amt -= chunk_left
+
+            # we read the whole chunk, get another
+            self._fp._safe_read(2)      # toss the CRLF at the end of the chunk
+            chunk_left = None
+
+        # read and discard trailer up to the CRLF terminator
```
shazow: Another style nit: Please make all comments sentence-like for consistency (starts with a capital, ends with a period).

:)

urllib3/response.py
```diff
@@ -351,3 +363,71 @@ def readinto(self, b):
...
+
+    def __iter__(self):
+        """
+        credit/resources:
+         * https://gist.github.com/ericflo/514507
+         * http://blog.dowski.com/2008/04/02/reading-chunked-http11-responses/
+         * https://hg.python.org/cpython/file/2.7/Lib/httplib.py
+        """
+        if not self.chunked:
+            return stream()
```
shazow: Is there a problem with having a return in a generator? Seems tests are failing for this reason: https://travis-ci.org/shazow/urllib3/jobs/53029337

shazow:
Also, further related to #436, I've been thinking: Maybe it would make sense to make the response object a context manager iff you want it streaming? This way we can avoid issues of manually closing the response if you haven't exhausted it.

Might look something like this:

```python
# Current semantics, preload_content=False default
r = pool.request(...)
data = r.read()
# If you used preload_content=True, you'd need to remember to close:
r.close()

# New semantics, preload_content=True default
with r as pool.request(...):
   data = r.read(10)

# No need to call r.close()
```

@Lukasa What do you think?

(Sorry this is only mildly related to the current PR, but it'd be nice to have consistent streaming semantics.)

urllib3/response.py
Lukasa: There is. It should probably be iterating over `stream()`, which itself returns a generator. Ideally, we'd use `yield from` here, but sadly Python 2.

Lukasa:
@shazow I guess? What happens if you use the object as a context manager when it isn't streaming?

urllib3/response.py
TomasTomecek: okay, makes sense to me (I tried to copy most of the code from httplib; my reasoning was that if it's in httplib, it gotta be perfect) — will update

urllib3/response.py
TomasTomecek: TBH I'm not really sure how to approach this: since the response is chunked and will be continually growing (which may take hours), read may block for a long time. Is this something we really want?

But as I'm thinking about it: `read` may be the blocking call, and `read_chunked` the non-blocking asynchronous call. How about that?

urllib3/response.py
TomasTomecek: Well, I really hate sentence comments. But if that's your wish, no problem updating it.

urllib3/response.py
TomasTomecek: Great catch! (why haven't my linter discovered this?)

TomasTomecek:
It's updated. Didn't rebase, rather made new commit: so many changes, it will be easier to review (realized that specifying amount of `_read_chunked` completely breaks it). Will squash when it's ready.

Also created two tests, please review.

test/test_response.py
```diff
@@ -2,6 +2,7 @@
 
 from io import BytesIO, BufferedReader
 
+from httplib import HTTPResponse as HLHTTPResponse
```
shazow: `httplib_HTTPResponse` might be clearer (had a few double-takes about what an HLHTTP is.) :)

urllib3/response.py
```diff
@@ -351,3 +361,71 @@ def readinto(self, b):
         else:
             b[:len(temp)] = temp
             return len(temp)
+
+    def read_chunked(self, amt=None):
+        # FIXME: could this be handled better?
+        assert self.chunked
```
shazow: Maybe better to just throw a TypeError or something?

urllib3/response.py
```diff
@@ -351,3 +361,71 @@ def readinto(self, b):
...
+        self.chunk_left = self.chunk_left
```
shazow: Does this line do anything?

urllib3/response.py
```diff
@@ -351,3 +361,71 @@ def readinto(self, b):
         else:
             b[:len(temp)] = temp
             return len(temp)
+
+    def read_chunked(self, amt=None):
```
shazow: This method is pretty meaty, would be a good candidate for somekind of helper. Maybe a ChunkedReader class?

Not required for this PR, just something to think about. :)

urllib3/response.py
```diff
@@ -351,3 +361,71 @@ def readinto(self, b):
...
+            if line == '\r\n':
+                break
+
+        # We read everything; close the "file".
+        self.close()
+
+
+    def __iter__(self):
+        """
+        credit/resources:
```
shazow: A formatting suggestion:

```diff
-        credit/resources:
-         * https://gist.github.com/ericflo/514507
-         * http://blog.dowski.com/2008/04/02/reading-chunked-http11-responses/
-         * https://hg.python.org/cpython/file/2.7/Lib/httplib.py
-         # server which produces chunked responses:
-         https://gist.github.com/josiahcarlson/3250376
+        References used for this implementation:
+         * https://gist.github.com/ericflo/514507
+         * http://blog.dowski.com/2008/04/02/reading-chunked-http11-responses/
+         * https://hg.python.org/cpython/file/2.7/Lib/httplib.py
+         * https://gist.github.com/josiahcarlson/3250376 (Server which produces chunked responses)
```

shazow:
@TomasTomecek That looks great, thank you! I didn't reason through the logic of `read_chunked` superclosely, so I'll take the word of your tests for it. :)

@sigmavirus24 Want to take a quick look at this too?

test/test_response.py
sigmavirus24: +1

test/test_response.py
```diff
@@ -406,11 +407,96 @@ def close(self):
...
+        """
+        self.content = content
+        self.index = 0
+        self.closed = False
+        self.exhausted = False
+        self.cur_chunk = ''
+
+    def get_chunk(self):
+        #if self.exhausted:
+        #    self.cur_chunk = ''
```
sigmavirus24: Can this be removed? Should it be uncommented?

test/test_response.py
```diff
@@ -406,11 +407,96 @@ def close(self):
...
+        if len(self.cur_chunk) <= 0:
+            try:
+                chunk = self.content[self.index]
+            except IndexError:
+                self.exhausted = True
+                chunk = ''
```
sigmavirus24: Pretty sure all of these need to be bytes. I'm a bit surprised Python 3 tests haven't exploded on you (although I haven't checked the travis results yet).

test/test_response.py
```diff
@@ -406,11 +407,96 @@ def close(self):
...
+            else:
+                self.index += 1
+            self.cur_chunk = '%X\r\n%s\r\n'%(len(chunk), chunk)
```
sigmavirus24: This won't work when you eventually convert the chunks to bytes, you'll have to use a different approach.

sigmavirus24:
So... I have an API question. Why isn't this `iter_chunked` (or `iterchunked`, or just `chunked`) instead of making the response iterable? If we do that only for chunked responses, people are going to be very confused. I'd rather it simply be:

```py
for chunk in response.iter_chunked():
    # ...
```

shazow:
@sigmavirus24 Good question.

Taking that further, maybe it should just be part of `response.stream()`?

test/test_response.py
TomasTomecek: okay, will `import httplib`

urllib3/response.py
TomasTomecek: I think that `ValueError` will be more appropriate.

urllib3/response.py
TomasTomecek: refactoring leftover, shame on me

urllib3/response.py
TomasTomecek: I'll totally agree. Will put a *FIXME* inside.

urllib3/response.py
TomasTomecek: To be honest, I'm planning to strip all those URLs from the final PR and put it to commit message. What do you think?

test/test_response.py
TomasTomecek: Will remove (had different logic in my head than the one I actually wrote).

urllib3/response.py
Lukasa: I don't. =) From [the docs](https://docs.python.org/2/library/exceptions.html#exceptions.ValueError):

> Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

`TypeError` is closer:

> Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.

But altogether I recommend adding a new urllib3 exception that multiply-inherits from `urllib3.exceptions.ProtocolError` and `TypeError`.

test/test_response.py
TomasTomecek: Yeah, it's really weird, will update.

TomasTomecek:
@shazow @sigmavirus24 
You want me to merge `iter_chunked` into stream? (as I'm thinking about, makes perfect sense to me: to have a method which returns generator, rather than Response being iterable)

Updated this PR with all the issues you guys found.

urllib3/response.py
TomasTomecek: Okay, will create new exception type then.

urllib3/response.py
shazow: Makes sense.

shazow:
@TomasTomecek Yea let's do that. :) Thank you!

TomasTomecek:
Squashed all the commits, created new, single, commit with proper message. From my PoV, it's ready for merge (once travis passes ofc).

TomasTomecek:
Okay, so coverage is failing on:
```python
try:
  # python 2 import
except:  # <-- here
  # python 3 import
```
how to approach this?

Lukasa:
Refactor to this:

```
try:  # Platform-specific: Python 2
  # python 2 import
except:  # Platform-specific: Python 3
  # python 3 import
```

Lukasa:
Sorry, my fault. Instead of the #pragma: no cover, use `except ImportError:`. You should only catch that exception anyway, and that will also automatically get excluded from coverage.

TomasTomecek:
@Lukasa thanks Cory! That worked. Travis is green now, wooooooh!

test/test_response.py
```diff
@@ -2,8 +2,12 @@
 
 from io import BytesIO, BufferedReader
 
+try:
+    import http.client as httplib
+except:
```
Lukasa: Let's do this as `except ImportError` as well. Naked `except` statements are generally pretty bad because they catch all sorts of things you aren't expecting (for example, `KeyboardInterrupt`!).

test/test_response.py
```diff
@@ -406,11 +406,207 @@ def close(self):
...
+        r = resp.read_chunked()
+        next(r)
+        self.assertRaises(ResponseNotChunked, next, r)
+
+    def test_read_not_chunked_response_as_chunks(self):
+        fp = BytesIO(b'foo')
+        resp = HTTPResponse(fp, preload_content=False)
+        r = resp.read_chunked()
+        self.assertRaises(ResponseNotChunked, next, r)
+
```
Lukasa: This test is duplicated. =)

test/test_response.py
TomasTomecek: You're totally right, fixed.

test/test_response.py
TomasTomecek: How could that happen?! :o thanks for noticing

TomasTomecek:
Btw, when I tried to run test suite locally using `make test`, it got stuck on some test. Is that expected, or worth reporting?

urllib3/response.py
```diff
@@ -1,9 +1,14 @@
...
+except ImportError:
+    import httplib
 import zlib
 import io
 from socket import timeout as SocketTimeout
 
 from ._collections import HTTPHeaderDict
-from .exceptions import ProtocolError, DecodeError, ReadTimeoutError
+from .exceptions import ProtocolError, DecodeError, ReadTimeoutError, \
+    ResponseNotChunked
```
Lukasa: Rather than using the explicit line continuation, prefer using python's implicit line continuation between parentheses:

```python
from .exceptions import (
    ProtocolError, DecodeError, ReadTimeoutError, ResponseNotChunked
)
```

urllib3/response.py
```diff
@@ -117,8 +122,17 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
         if hasattr(body, 'read'):
             self._fp = body
 
-        if preload_content and not self._body:
-            self._body = self.read(decode_content=decode_content)
+        # Are we using the chunked-style of transfer encoding?
+        self.chunked = False
+        self.chunk_left = None
+        tr_enc = self.headers.get('transfer-encoding')
```
Lukasa: You can save yourself a conditional check on the next line by returning the empty string in the failure case:

```python
tr_enc = self.headers.get('transfer-encoding', '')
```

Lukasa:
Not sure about the local test suite hang, but I have seen it happen in the past. Python threading isn't always elegant. =(

urllib3/response.py
TomasTomecek: you are really thorough; I like that!

TomasTomecek:
@Lukasa okay, will report then
updated with your latest suggestion and rebased

Lukasa:
Sweet, LGTM. Thanks so much for this @TomasTomecek, it's really excellent work! :cake: :star: :sparkles: :cookie:

shazow:
Looks great! Would you like to add a snippet to the docs to show how to treat a response as a stream? :) Also you're welcome to add yourself to our list of CONTRIBUTORS.txt!

Let me know when you want me to merge it, looks good on my end. :)

sigmavirus24:
> Btw, when I tried to run test suite locally using make test, it got stuck on some test. Is that expected, or worth reporting?

Does it happen every time? If so, you may need to use a better DNS system. I've had that happen to me, and I fixed it by configuring my computer to use Google's DNS. Works every time like a charm.

test/test_response.py
```diff
@@ -406,11 +406,200 @@ def close(self):
...
+        """
+        self.content = content
+        self.index = 0  # This class iterates over self.content.
+        self.closed = False
+        self.cur_chunk = b''
+        self.chunks_exhausted = False
+
+    @staticmethod
+    def _encode_chunk(chunk):
+        return '%X\r\n%s\r\n' % (len(chunk), chunk.decode())
```
sigmavirus24: Should this be

```py
def _encode_chunk(chunk):
    return ('%X\r\n%s\r\n' % (len(chunk), chunk.decode())).encode()
```

So that it returns bytes instead of unicode (which is what we expect to receive)?

urllib3/response.py
```diff
@@ -351,3 +370,59 @@ def readinto(self, b):
...
+        if not self.chunked:
+            raise ResponseNotChunked("Response is not chunked. "
+                "Header 'transfer-encoding: chunked' is missing.")
+        while True:
+            # First, we'll figure out length of a chunk and then
+            # we'll try to read it from socket.
+            if self.chunk_left is None:
+                line = self._fp.fp.readline()
+                line = line.decode()
+                # See rfc2616: Chunked Transfer Coding.
```
sigmavirus24: 2616 is dead. Should this be updated to point at the correct RFC number?

Lukasa: Probably. RFC 7230 is the correct one.

test/test_response.py
TomasTomecek: this actually means transfer-encode-chunk, not unicode-to-byte-encode chunk; that encoding call is on line 533

urllib3/response.py
TomasTomecek: that is correct, will update

TomasTomecek:
@sigmavirus24 issue created: #568 

@shazow done and done: wasn't sure where to put the docs, so I stuff it in index in the end.

Ready to get merged from my point of view.

docs/index.rst
```diff
@@ -257,6 +257,35 @@ disabled individually.
...
+Stream
+------
+
+You may also stream your response and get data as they come (e.g. when using
+``transfer-encoding: chunked``). In this case, method
+:func:`~urllib3.response.HTTPResponse.stream` will return generator.
+
+::
+
+    >>> import urllib3, datetime
```
shazow: In the future, lean towards simpler examples (e.g. no need for datetime).

Also maybe use http://httpbin.org/ rather than a gist/localhost reference. :)

shazow:
@TomasTomecek You did a fantastic job, thank you!

docs/index.rst
TomasTomecek: Wanted to explicitly demonstrate, that chunks are printed as they come; can update it with another PR if you wish. (wasn't really aware of httpbin.org)

TomasTomecek:
@Lukasa @shazow @sigmavirus24 
I'm really glad this went so smoothly. Thanks a lot for your excellent suggestions!

Lukasa:
:cake: :sparkles: Fantastic work @TomasTomecek!

Lukasa:
Hmm, just noticed that the `read_chunked` method closes the connection when the chunked read is complete, rather than releasing it to the pool. Might want to fix that up. =)

shazow:
Mmm good point, I think we want `Response.release_conn()`.

docs/index.rst
shazow: That's fair. It might make sense to add another section to the docs about handling different kinds of responses where we can go into deeper detail, but I feel that it's important to keep examples brief for the landing index page.

urllib3/response.py
```diff
@@ -117,8 +123,17 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
         if hasattr(body, 'read'):
             self._fp = body
 
-        if preload_content and not self._body:
-            self._body = self.read(decode_content=decode_content)
+        # Are we using the chunked-style of transfer encoding?
+        self.chunked = False
+        self.chunk_left = None
+        tr_enc = self.headers.get('transfer-encoding', '')
+        if tr_enc.lower() == "chunked":
```
sigmavirus24: Isn't it possible for there to be multiple Transfer-Encodings specified, e.g., `chunked, gzip`?

TomasTomecek: I don't know, would have to check RFC

Lukasa: Yes.

sigmavirus24: This is taken care of in #597

PR #562, opened 2015-03-04
Fix HTTPHeaderDict implementation for multiple headers

### LINKED ISSUE #561: duplicate/multiple headers are lost
When the server returns multiple headers of the same type, duplicated headers are lost. Example:

WWW-Authenticate: Negotiate
WWW-Authenticate: NTLM

Results in only NTLM in the www-authenticate header.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9193217-duplicate-multiple-headers-are-lost?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
- Allow all headers to have multiple values
- Special-case how cookie, set-cookie, and set-cookie2 are presented/handled
- Trim some extraneous whitespace

Fixes #561

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -20,7 +20,8 @@ from .packages.six import iterkeys, itervalues, PY3
 __all__ = ['RecentlyUsedContainer', 'HTTPHeaderDict']
 
 
-MULTIPLE_HEADERS_ALLOWED = frozenset(['cookie', 'set-cookie', 'set-cookie2'])
+SPECIAL_CASE_MULTIPLE_HEADERS = frozenset(['cookie', 'set-cookie',
+                                           'set-cookie2'])
 
 _Null = object()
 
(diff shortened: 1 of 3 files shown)


sigmavirus24:
Please look very carefully at this pull request while reviewing it.

Specifically, for the three header names that were special cased, I do not return all of the values when you do

```py
header_dict['set-cookie']
```

Only the first one in the list. If all header values are needed, then you must use ```HTTPHeaderDict.getlist``

Lukasa:
What do we think the odds are that anyone is calling `getlist` for those headers?

urllib3/_collections.py
```diff
@@ -148,11 +149,18 @@ def __init__(self, headers=None, **kwargs):
             self.extend(kwargs)
 
     def __setitem__(self, key, val):
-        return _dict_setitem(self, key.lower(), (key, val))
+        if key in self:
+            del self[key]
+        self.add(key, val)
+        return val
```
Lukasa: My suspicion is that this reversion also reverts part of the purpose of the original change.

urllib3/_collections.py
```diff
@@ -148,11 +149,18 @@ def __init__(self, headers=None, **kwargs):
...
 
     def __getitem__(self, key):
-        val = _dict_getitem(self, key.lower())
-        return ', '.join(val[1:])
+        key_lower = key.lower()
+        _, values = _dict_getitem(self, key_lower)
+        if key_lower in SPECIAL_CASE_MULTIPLE_HEADERS:
+            # NOTE(sigmavirus24): Should we return something else?
```
Lukasa: Maybe a list, but in practice I think that's breaky.

sigmavirus24: I think a list is the wrong thing here. We can't join the header values like we do elsewhere.

Lukasa:
Closing in favour of #563.

PR #563, opened 2015-03-05
Fix multiheader regression #561

ml31415:
See #561 for discussion.

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -141,7 +141,7 @@ class HTTPHeaderDict(dict):
     def __init__(self, headers=None, **kwargs):
         dict.__init__(self)
         if headers is not None:
-            if isinstance(headers, type(self)):
+            if isinstance(headers, HTTPHeaderDict):
                 self._copy_from(headers)
             else:
                 self.extend(headers)
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -20,8 +20,6 @@ from .packages.six import iterkeys, itervalues, PY3
 __all__ = ['RecentlyUsedContainer', 'HTTPHeaderDict']
 
 
-MULTIPLE_HEADERS_ALLOWED = frozenset(['cookie', 'set-cookie', 'set-cookie2'])
-
 _Null = object()
 
 
@@ -223,11 +221,8 @@ class HTTPHeaderDict(dict):
                 vals.append(val)
             else:
                 # vals should be a tuple then, i.e. only one item so far
-                if key_lower in MULTIPLE_HEADERS_ALLOWED:
-                    # Need to convert the tuple to list for further extension
-                    _dict_setitem(self, key_lower, [vals[0], vals[1], val])
-                else:
-                    _dict_setitem(self, key_lower, new_vals)
+                # Need to convert the tuple to list for further extension
+                _dict_setitem(self, key_lower, [vals[0], vals[1], val])
 
     def extend(*args, **kwargs):
         """Generic import function for any type of header-like object.
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -141,7 +141,10 @@ class HTTPHeaderDict(dict):
     def __init__(self, headers=None, **kwargs):
         dict.__init__(self)
         if headers is not None:
-            self.extend(headers)
+            if isinstance(headers, type(self)):
+                self._copy_from(headers)
+            else:
+                self.extend(headers)
         if kwargs:
             self.extend(kwargs)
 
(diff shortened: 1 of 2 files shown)


ml31415:
Not sure what went wrong with CI, but the errors seem to be unrelated to my changes.

Lukasa:
Agreed, CI failures are unrelated to this change.

Lukasa:
@ml31415 This needs not just removal of the previous test, but the addition of a new test (or maybe several) that specify exactly how combinations of headers do and don't work, and what effects they have on various methods of retrieving headers. That means seeing how `__setitem__` interacts with `add` and `extend`, and then how `__getitem__`, `get`, and `getlist` interact as well. Try to keep these tests simple and split them up into lots of different single tests so that we can confirm that the tested behaviour matches the expected behaviour.

sigmavirus24:
@ml31415 I'm uncertain why this is better since you're not going to properly handle how `cookie`, `set-cookie` and `set-cookie2` headers should be treated.

Lukasa:
To be clear, no `cookie` header should ever be concatenated for any reason. From [RFC 6265](https://tools.ietf.org/html/rfc6265) section 3:

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field.  The usual mechanism for folding HTTP headers fields (i.e., as defined in [RFC2616]) might change the semantics of the Set-Cookie header field because the %x2C (",") character is used by Set-Cookie in a way that conflicts with such folding.

Strictly this directive applies to origin servers only, but given that we're trying to represent them we shouldn't map incorrectly.

Lukasa:
`Cookie` headers, however, can safely be joined by semicolons.

ml31415:
So instead of joining any of this three special headers with `,` on `__getitem__`, they should be joined with `;` ? `__getitem__` is supposed to return a string, so a merge of any kind has to happen in case of multiheaders.

Lukasa:
No. `Cookie` may be merged, but `set-cookie` and `set-cookie2` headers must not be merged. There is no safe way to merge those header fields.

ml31415:
Well, so what to do on `__getitem__['set-cookie']` ? Return a list? Throw an error instantly? All will probably break previously working code somewhere.

Lukasa:
For back compatibility I think we need to join it with a comma, so for now do that. @sigmavirus24, yours should do that too, or requests will have to worry about the new version.

However, I'm increasingly tempted to throw the whole thing out and come up with a brand new header map that is expressly *not* a dictionary.

ml31415:
The best fit for headers in my eys is actually something like `defaultdict(list)`, and right now it is quite close to that. The problem is the old python message interface, which treats multiple headers as some kind of strange alien, and not as a common case.

Lukasa:
Even defaultdict(list) isn't quite right, because it doesn't have (IMO) the correct iteration semantics. I'll see if I can whip something up that looks closer to accurate.

seocam:
@Lukasa I agree we should really rethink this headers structure. Probably is best to open an issue just to discuss that.

ml31415:
Iteration over items could be modified, to list all duplicate header lines separately, giving a nice reconstruction of the original headers. Also length attribute can be adjusted to that. Another question is, how `keys()` is supposed to behave in terms of duplicates.

Lukasa:
@ml31415 All of that is out of scope for now, we need to un-break what we broke. =) We can worry about doing this *right* a bit later.

urllib3/_collections.py
```diff
@@ -143,7 +141,10 @@ class HTTPHeaderDict(dict):
     def __init__(self, headers=None, **kwargs):
         dict.__init__(self)
         if headers is not None:
-            self.extend(headers)
+            if isinstance(headers, type(self)):
```
seocam: Here I'd use HTTPHeaderDict explicitly since we probably want to keep this behavior on inherited classes and type(self) will return the subclass instead of the parent.

ml31415:
So the only question remaining is, what to do with special case headers in case of `__getitem__`? #562 simply returns the first value, also commenting this as a questionable choice. #563 merges the values with `,` together. Both are imho not nice solutions, though there is probably no best solution.

seocam:
@ml31415 I agree with @Lukasa again on that. On __getitem__ calls I'd merge with commas to preserve backwards compatibility.

urllib3/_collections.py
ml31415: Fixed.

Lukasa:
+1 @seocam

ml31415:
I'm also fine with that. So some more test cases, and we're good to go?

Lukasa:
I hope so, yeah. =) Obviously we'll review the finished product before merging, but hopefully that's all we'll need.

shazow:
@ml31415 Thank you for doing this. Agree with everyone's comments. :) Please ping me when you feel this is ready for final review/merge. Will push out v1.10.3 after that.

sigmavirus24:
@shazow would it be possible for urllib3 to cut 1.10.3 which simply reverts the problematic pull request until @ml31415 can finish this up? requests would like to cut a new release which fixes other issues including the ones introduced in #544

seocam:
@sigmavirus24 unless @ml31415 or someone wants to right some extra test cases or to give an extra review this PR I believe this is ready to merge.

Lukasa:
I'd still like the extra test cases if we can have them.

ml31415:
I added one more test for not well known multiheaders, so that you can put this fixed version out.

ml31415:
The CI errors are unrelated.

sigmavirus24:
> I added one more test for not well known multiheaders, so that you can put this fixed version out.

I think the desire was to have a number more tests, not just one. But that's @Lukasa and @shazow's call if they want to be strict about it

ml31415:
Well, it clearly tests the regression, so for having the fix out it should be fine.

Lukasa:
I'm ok with merging for now. Longer term we should aim to build a suite of header tests.

shazow:
Agree, it'd be nice to have a more robust header suite but just addressing the specific regression is sufficient to push this through. Thank you. :)

PR #565, opened 2015-03-05
Test for set-cookies headers are lost after creating a HTTPHeaderDict using an instance of HTTPHeaderDict

seocam:


Lukasa:
For obvious reasons, this test fails the build. That is, after all, what it's for. ;)

sigmavirus24:
I merged this into #562

Lukasa:
I'll leave this open until it gets merged into #563 as well.

Lukasa:
Done. @seocam thanks for this, both current pull requests now contain your test. =) :cake:

PR #566, opened 2015-03-10
Further tests for HTTPHeaderDict, better python2.7 raw httplib header import

ml31415:
Added full support for python2.7 multiheader import, borrowed from 
https://gist.github.com/wtolson/bf7dada12ddda0482d10
Thanks to @wolston for that code!

Extended test for python2.7 raw header import
Added further header tests borrowed from geventhttpclient

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -306,6 +306,9 @@ class HTTPHeaderDict(dict):
     @classmethod
     def from_httplib(cls, message): # Python 2
         """Read headers from a Python 2 httplib message object."""
+        # python2.7 does not expose a proper API for exporting multiheaders
+        # efficiently. This function re-reads raw lines from the message 
+        # object and extracts the multiheaders properly.
         headers = []
          
         for line in message.headers:
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -227,20 +227,20 @@ class HTTPHeaderDict(dict):
                 # Need to convert the tuple to list for further extension
                 _dict_setitem(self, key_lower, [vals[0], vals[1], val])
 
-    def extend(*args, **kwargs):
+    def extend(self, *args, **kwargs):
         """Generic import function for any type of header-like object.
         Adapted version of MutableMapping.update in order to insert items
         with self.add instead of self.__setitem__
         """
-        if len(args) > 2:
-            raise TypeError("update() takes at most 2 positional "
+        if len(args) > 1:
+            raise TypeError("extend() takes at most 1 positional "
                             "arguments ({} given)".format(len(args)))
-        elif not args:
-            raise TypeError("update() takes at least 1 argument (0 given)")
-        self = args[0]
-        other = args[1] if len(args) >= 2 else ()
+        other = args[0] if len(args) >= 1 else ()
         
-        if isinstance(other, Mapping):
+        if isinstance(other, HTTPHeaderDict):
+            for key, val in other.iteritems():
+                self.add(key, val)
+        elif isinstance(other, Mapping):
             for key in other:
                 self.add(key, other[key])
         elif hasattr(other, "keys"):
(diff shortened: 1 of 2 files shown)


urllib3/_collections.py
```diff
@@ -227,20 +227,20 @@ def add(self, key, val):
...
+            raise TypeError("extend() takes at most 1 positional "
                             "arguments ({} given)".format(len(args)))
-        elif not args:
-            raise TypeError("update() takes at least 1 argument (0 given)")
-        self = args[0]
-        other = args[1] if len(args) >= 2 else ()
+        other = args[0] if len(args) >= 1 else ()
         
-        if isinstance(other, Mapping):
+        if isinstance(other, HTTPHeaderDict):
```
shazow: Would we be better off checking `hasattr(other, 'iteritems')`?

urllib3/_collections.py
```diff
@@ -304,17 +304,17 @@ def items(self):
...
-        # extra code.
-        for key in duplicates:
-            ret.discard(key)
-            for val in message.getheaders(key):
-                ret.add(key, val)
-            return ret
+        headers = []
+         
+        for line in message.headers:
+            if line.startswith((' ', '\t')):
```
shazow: Hmm can you explain what we're doing here? (Maybe a comment)

shazow:
@ml31415 Good changes, thank you! :) Left some comments, please ping when it's ready for merge.

urllib3/_collections.py
ml31415: Mappings should have iteritems available. I specifically added this, so that when a HTTPHeaderDict is put in for extending, `.__getitem__()` will not be called, which would merge header lines together and lose information.

urllib3/_collections.py
ml31415: My previous implementation sucked, as it only copied all single header items, and then added the explicitly requested multiheaders. This obviously left out all not explicitly multiheaders asked for and reduced them to only the last item. See my previous comment.

The new code is re-iterating the raw header information completely, including any multiheader. The code is taken from @wtolson, link see above, thanks for that again! I also extended the header parsing testcase for checking multiple lines. Might be worth, to borrow some further tests from original python2.7 headers ...

ml31415:
@shazow :bell:

urllib3/_collections.py
ml31415: In case I missed something there in the py2 HTTPMessage, which would avoid reinventing the wheel and still being reasonable fast for importing the whole object, i.e. a oneway pass through the headers, I welcome any hints and well deserved bitchslaps!

sigmavirus24: iteritems doesn't exist on Python 3
-- 

shazow: Basically the scenario I want to avoid is if urllib3.HTTPHeaders is used with requests.packages.urllib3. If we use isinstance, they won't be interchangeable.

ml31415: As long as there is no agreement, how the common header API will look like, it won't make sense to try to detect it imho. When it is decided, it should be a minor fix.

urllib3/_collections.py
ml31415: In case I missed something there in the py2 HTTPMessage, which would avoid reinventing the wheel and still being reasonable fast for importing the whole object, i.e. a oneway pass through the headers, I welcome any hints and well deserved bitchslaps!

urllib3/_collections.py
ml31415: I just hat a look at requests CaseInsensitiveDict. Looks like the only thing we would have to add is something to replicate `.lower_items()` and then they might be good to go to just take HTTPHeaderDict as it is.

shazow: > As long as there is no agreement, how the common header API will look like, it won't make sense to try to detect it imho.

What do you mean? Requests vendors urllib3 inside of its own package. Because it's a *copy*, the two class definitions would not be the same, so isinstance would fail despite that they're effectively the same.

I'm proposing we do more duck-typing rather than instance-checking for internal objects.

> Looks like the only thing we would have to add is something to replicate .lower_items() and then they might be good to go to just take HTTPHeaderDict as it is.

+1.

ml31415: I'd agree to prefer duck typing, but on which methods distinct from a Mapping do you want to rely? I'm not sure if iteritems is the definite choice here, though it would work for now. Still, would be nice, if the behaviour of all the iterations would have a better defined behaviour, i.e. extracting multiheaders on iteration or not. keys and items handle this different right now, there are reasons for both and it's still not discussed.

shazow: Fair points.

In that case, what if we just checked for `elif isinstance(other, Mapping)`, and skip the iteritems case? Any downside? It would work in both cases.

shazow: I guess there's a mild performance benefit in that case.

Anyways, I'm fine with keeping it as-is for now, maybe something we should do a bigger refactor for in the future.

ml31415: If we skip this new, first if, then HTTPHeaderDict will be treated like any Mapping. The Mapping will be iterated over all the keys and __getitem__ issued on all headers. This will cause e.g. multiple cookies to be merged together with commas, which of course should be avoided.

shazow: Agree.

https://docs.python.org/2/library/collections.html#collections-abstract-base-classes Here's what we have to work with natively. Seems `mapping` also assumes you have an `items()` method which should do the trick, rather than looking up per-key?

PR #570, opened 2015-03-15
Properly detect Python 3.0-3.1 in SSLContext

akanouras:
This stops pip (through its use of requests) from working under Python 3.1 (and probably 3.0 too).

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -40,7 +40,8 @@ except ImportError:
     import sys
 
     class SSLContext(object):  # Platform-specific: Python 2 & 3.1
-        supports_set_ciphers = sys.version_info >= (2, 7)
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for this @akanouras!

Do you mind putting the check in parentheses, rather than using the explicit line continuation character?

akanouras:
Thanks for the fast response!
Added a commit, please tell me if I should rebase instead.

Lukasa:
No problem! =) I'll let @shazow cast his eye over this before he merges it.

The test failure appears to be irrelevant to your change.

shazow:
Looks great, thank you!

piotr-dobrogost:
> This stops pip (through its use of requests) from working under Python 3.1 (and probably 3.0 too).

According to https://pip.pypa.io/en/latest/installing.html pip does not support Python 3.1 (nor 3.0).

PR #573, opened 2015-03-17
Stop relying on the stdlib's default cipher list

sigmavirus24:
urllib3's release cadence is much faster so it's preferable if
urllib3 handles its own default cipher list as it will undoubtedly
be more quickly updated/released.


shazow:
+1 to @kevinburke, otherwise looks good.

dstufft:
I would actually personally drop the ``HIGH`` ciphers, almost nothing is going to require them and the only reason the stdlib has them is because it was the only way I could get them to agree to a stricter cipher list.

t-8ch:
@sigmavirus24 We have another cipherlist declaration in contrib/pyopenssl. Can you change this, so both use the same?

Lukasa:
Superseded by #588.

PR #571, opened 2015-03-17
Link the security docs from the front page of the docs.

Lukasa:
I keep getting confused by the fact that I can't find the security docs from the index.


Lukasa:
@shazow Ask, and ye shall receive.

PR #575, opened 2015-03-18
read chunked: don't close socket, release connection instead

TomasTomecek:
Fixup for #560.


Lukasa:
LGTM. :cake:

shazow:
Ah thank you for following up!

PR #581, opened 2015-04-06
proof of concept: split ProtocolError into socket.connect, socket.write, socket.read

kevinburkeshyp:
In https://github.com/shazow/urllib3/issues/547 I said we could detect which type of error got raised by httplib. Here is a proof of concept showing how to do this. 

Summary of the changes - 

- The essential `httplib` calls are both in `_make_request`: the first is `conn.request` and the second is `conn.getresponse()`. In order to distinguish these at the `urlopen` level, I had to split `_make_request` into two functions. Otherwise, you have to distinguish between errors between all three of socket connect/write/read; did the `ETIMEDOUT` occur on the connect or the read?
    - This gets ugly since we want to pass a lot of state from the first function to the second function, in particular the `conn` object and the `timeout`. There's a failing test `test_total_timeout` that I couldn't figure out what to do with (besides bail on the feature, which I wouldn't be too against)
    - a lot of the tests needed to be updated to call both `_make_` functions, which shouldn't be too terrible, since it's an underscored function.
    - what is terrible is most of the exception handling has to be duplicated for `_make_connect` and `_make_request`... some of it could probably be killed if we looked harder at what could be thrown by both methods.

So yeah, I'm not amazingly sure about this, and given that I feel bad about code I felt better about when I submitted it, I'm not sure this is worth going down. 

Still, this is evidence that if we *wanted* to do this, we *could* without necessarily rewriting httplib.

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -140,21 +140,21 @@ class TestConnectionPool(unittest.TestCase):
             raise ex()
 
         def _test(exception, expect):
-            pool._make_request = lambda *args, **kwargs: _raise(exception)
+            pool._make_connect = lambda *args, **kwargs: _raise(exception)
             self.assertRaises(expect, pool.request, 'GET', '/')
 
             self.assertEqual(pool.pool.qsize(), POOL_SIZE)
 
         # Make sure that all of the exceptions return the connection to the pool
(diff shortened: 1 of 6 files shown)


kevinburkeshyp:
work account... whoops

PR #585, opened 2015-04-08
Add timeout to ssl_wrap_socket

fermayo:
Our code is getting stuck sometimes on `select.select([sock], [], [])` forever. It seems to me that we need a timeout in that call. Would adding `sock.gettimeout()` to the `select` call make any sense?

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -299,7 +299,9 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         try:
             cnx.do_handshake()
         except OpenSSL.SSL.WantReadError:
-            select.select([sock], [], [])
+            rd, _, _ = select.select([sock], [], [], sock.gettimeout())
+            if not rd:
+                raise timeout('select timed out')
             continue
         except OpenSSL.SSL.Error as e:
             raise ssl.SSLError('bad handshake', e)


shazow:
Hm interesting, does this help in your scenario? Any idea how to reproduce/test it?

(cc @kevinburke, any ideas what's going on here?)

Lukasa:
We should confirm here whether the select ends because it times out or because it returned, and if it timed out we should probably throw a nice big shiny exception (`HandshakeTimeout` or `ConnectionError` or something).

shazow:
+1

fermayo:
I have updated the PR. Would something like that work?

Lukasa:
@fermayo Looks pretty good.

urllib3/contrib/pyopenssl.py
```diff
@@ -299,7 +299,9 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         try:
             cnx.do_handshake()
         except OpenSSL.SSL.WantReadError:
-            select.select([sock], [], [])
+            rd, _, _ = select.select([sock], [], [], sock.gettimeout())
+            if not rd:
+                raise timeout()
```
shazow: Can we add a message here?

fermayo: done

shazow:
Any ideas on how to test this? (What would cause select to hang? unresponsive server?)

fermayo:
I think it happens with very busy servers that have the SSL port open but don't respond when trying to connect to it

Lukasa:
Yeah, you'd have to begin the handshake but not finish it. I really don't know how to simulate that on our end though, I don't think any library gives us low-level enough control.

shazow:
Something along these lines? https://github.com/shazow/urllib3/blob/master/test/with_dummyserver/test_socketlevel.py#L459

Either way, this is a good fix and I don't want to block on tests here but I'd appreciate it if anyone wants to try and reproduce. :)

PR #586, opened 2015-04-09
add a new test-nossl Makefile target

vbraun:
Adds "make test-nossl" to run the unittests  without ssl being importable. Right now urllib3 cannot be imported without ssl, but you can only fix it if you can test it.

Its somewhat tricky to test this because nosetests will have urllib3 already imported by the time the unittest is run. Hence this contraption with blocking ssl from the tip of `test/__init__.py`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ clean:
 test: requirements
 	nosetests
 
+test-nossl: requirements
+	BLOCK_SSL=yes nosetests test/
+
 test-all: requirements
 	tox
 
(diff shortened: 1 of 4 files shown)


shazow:
Thanks @vbraun! Looks a bit painful, but can't think of a better way to do this. HMM.

Lukasa:
I think I'd rather not have the makefile target if possible: see #587.

shazow:
@vbraun I'll close this for now in favour of #587.

Issue #584, opened 2015-04-08
Support Python without SSL (2)


vbraun:
This is essentially the same as issue #320, urllib3 fails to import if there is no ssl module (Tested on Python 2.7.9). There are various reasons one might not have it, e.g. self-complied Python without ssl headers, the steady stream of security vulnerabilities, or objections to the not-really-free openssl license. But right now pip requires ssl because it uses urllib3.
~~~~
In [1]: import urllib3
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-dbe53ea169b1> in <module>()
----> 1 import urllib3

/home/vbraun/Code/sage/local/lib/python2.7/site-packages/urllib3/__init__.py in <module>()
      8 
      9 
---> 10 from .connectionpool import (
     11     HTTPConnectionPool,
     12     HTTPSConnectionPool,

/home/vbraun/Code/sage/local/lib/python2.7/site-packages/urllib3/connectionpool.py in <module>()
     29 from .packages.ssl_match_hostname import CertificateError
     30 from .packages import six
---> 31 from .connection import (
     32     port_by_scheme,
     33     DummyConnection,

/home/vbraun/Code/sage/local/lib/python2.7/site-packages/urllib3/connection.py in <module>()
     43 from .packages.ssl_match_hostname import match_hostname
     44 
---> 45 from .util.ssl_ import (
     46     resolve_cert_reqs,
     47     resolve_ssl_version,

/home/vbraun/Code/sage/local/lib/python2.7/site-packages/urllib3/util/__init__.py in <module>()
      3 from .request import make_headers
      4 from .response import is_fp_closed
----> 5 from .ssl_ import (
      6     SSLContext,
      7     HAS_SNI,

/home/vbraun/Code/sage/local/lib/python2.7/site-packages/urllib3/util/ssl_.py in <module>()
     10 
     11 import errno
---> 12 import ssl
     13 
     14 try:  # Test for SSL features

/home/vbraun/Code/sage/local/lib/python/ssl.py in <module>()
     58 import textwrap
     59 
---> 60 import _ssl             # if we can't import it, let the error propagate
     61 
     62 from _ssl import OPENSSL_VERSION_NUMBER, OPENSSL_VERSION_INFO, OPENSSL_VERSION

ImportError: No module named _ssl
~~~~

Lukasa:
[...]

PR #587, opened 2015-04-09
test import without ssl module

t-8ch:
duplicates #586

### DIFF
--- a/test/test_no_ssl.py
+++ b/test/test_no_ssl.py
@@ -1,0 +1,56 @@
+"""
+Test what happens if Python was built without SSL
+
+* Everything that does not involve HTTPS should still work
+* HTTPS requests must fail with an error that points at the ssl module
+"""
+
+import sys
+import unittest
+
+
+class ImportBlocker(object):
+    """
+    Block Imports
+
+    To be placed on ``sys.meta_path``. This ensures that the modules
+    specified cannot be imported, even if they are a builtin.
+    """
+    def __init__(self, *namestoblock):
+        self.namestoblock = namestoblock
+
+    def find_module(self, fullname, path=None):
+        if fullname in self.namestoblock:
+            return self
+        return None
+
+    def load_module(self, fullname):
+        raise ImportError('import of {0} is blocked'.format(fullname))
+
+
+ssl_blocker = ImportBlocker('ssl', '_ssl')
+
+
+class TestWithoutSSL(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        sys.modules.pop('ssl', None)
+        sys.modules.pop('_ssl', None)
+        sys.meta_path.insert(0, ssl_blocker)
+
+    @classmethod
+    def tearDownClass(cls):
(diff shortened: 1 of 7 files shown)


vbraun:
The problem is that urllib3 has already been imported by the time that the test is run, so "import urllib3" uses the cache. In fact, the Travis tests should fail because of #584 but passes.

I tried imp.reload / unloading urllib3* but couldn't get that to work properly. After reloading urllib3 you end up with a mix of the cached previous import and the new import somewhere in mock leading to test failures.

t-8ch:
@vbraun The new version should be better

t-8ch:
Now it breaks coverage...

test/test_no_ssl.py
```diff
@@ -0,0 +1,93 @@
...
+            return self
+        return None
+
+    def load_module(self, fullname):
+        raise ImportError('import of {0} is blocked'.format(fullname))
+
+
+class ModuleStash(object):
+    """
+    Stashes avail previously imported modules
```
vbraun: Stashes away?

t-8ch: Yes, of course. No idea what my brain did there

test/test_no_ssl.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def stash(self):
+        self._data[self.namespace] = self.modules.pop(self.namespace, None)
+
+        for module in list(self.modules.keys()):
+            if module.startswith(self.namespace + '.'):
+                self._data[module] = self.modules.pop(module)
+
+    def pop(self):
+        for module in list(self.modules.keys()):
```
vbraun: Do you really need this loop? the sys.modules.update at the end overwrites them anyways. Also, you don't pop urllib3 (without dot).

vbraun:
The approach looks good, if it works then thats great!

test/test_no_ssl.py
t-8ch: I'd rather be more explicit. Maybe there are different modules imported while the stash was active.

shazow:
Ah yes, this is a bit better!

shazow:
... Except that tests fail, of course. :P

t-8ch:
@shazow Here :-)
You may want to revert db95 onnce merged, to make the tests active

shazow:
@t-8ch Sorry if I'm being dense, but why is db95ad4 required now but not later? :/

t-8ch:
@shazow it is only there to show, that the test works. In reality it is failing currently

shazow:
Then perhaps this PR should be on the repo that is fixing the problem, or we should hold off until that is merged?

t-8ch:
Yeah. But to be pedantic: we are currently broken :sadface:

shazow:
We're not broken, we're just differently-fixed.

t-8ch:
@shazow @vbraun Care to review again?

vbraun:
Sweet, you fixed it already! +1 from me

shazow:
You're so eli7e.

PR #594, opened 2015-04-23
make test suite more robust and compatible with Tornado 4

matejcik:
the current test suite relies on WSGI compatibility layer from Tornado 3, which was significantly reworked in version 4. As a result, one can't easily create Tornado HTTPRequests from WSGI parameters (I ran afoul of issue https://github.com/tornadoweb/tornado/issues/1118, which has since been fixed, but it is still safer to avoid it entirely).
fwiw, creating HTTPRequest from WSGI values this way doesn't make sense when the whole point of Tornado WSGI wrapper is to convert HTTPRequest *to* WSGI values, so, um.

This is a rather crude but practical way to use Tornado RequestHandler directly instead of the WSGI runaround.

In the second commit, I make the socket-based tests time out instead of hanging forever if for whatever reason the server fails to start. A similar change might be useful for the HTTP tests, but it's not as straightforward, so maybe next time ;)

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -9,7 +9,7 @@ import time
 import zlib
 
 from io import BytesIO
-from tornado.wsgi import HTTPRequest
+from tornado.web import RequestHandler
 
 try:
     from urllib.parse import urlsplit
@@ -28,16 +28,17 @@ class Response(object):
         self.status = status
         self.headers = headers or [("Content-type", "text/plain")]
 
-    def __call__(self, environ, start_response):
-        start_response(self.status, self.headers)
-        return [self.body]
+    def __call__(self, request_handler):
+        status, reason = self.status.split(' ', 1)
+        request_handler.set_status(int(status), reason)
+        for header,value in self.headers:
+            request_handler.add_header(header,value.decode('utf8'))
 
+        request_handler.write(self.body)
 
-class WSGIHandler(object):
-    pass
+RETRY_TEST_NAMES = collections.defaultdict(int)
 
-
-class TestingApp(WSGIHandler):
+class TestingApp(RequestHandler):
     """
     Simple app that performs various operations, useful for testing an HTTP
     library.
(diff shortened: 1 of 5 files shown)


Lukasa:
@shazow This might work? I don't really know enough Tornado to know...

Lukasa:
That said, the server seems to fail a lot, so...

matejcik:
also i apparently broke the python3 case sensitivity test in one way or another, but i don't understand the point of that test so i can't fix it :(

shazow:
That's exciting, thank you @matejcik!

The test_httplib_headers_case_insensitive test you mentioned, the idea was to make sure we're working around httplib's header parsing to maintain case insensitivity relative to what the server returns. It might be more straight-forward to move this into a socket-level test to avoid running into Tornado's own headers.

Also seems this is broken in Py26. Any ideas there?

matejcik:
hm, i originally thought that the "address already in use" thing from the first test run was the cause of the py2.6 failures, so i didn't look into it further... apparently python 2.6 didn't return meaningful values from `event.wait()`

matejcik:
as for the case-insensivity test, so what is the supposed result? the parser should always return headers with specific spelling ("Content-type")?
the test looks somewhat weird to me, too -- the HEADERS constant uses inconsistent spelling, "Content-Length" vs "Content-Type"

shazow:
@matejcik Er right, there was a bug where httplib would mangle the case of headers, and we had to implement a thing which bypassed that, and this is a test to verify it's working (assuming that Tornado was not also mangling the headers, which might not be true anymore).

matejcik:
ah, i see, the handler does specify "Content-type" in lowercase, so apparently Tornado is normalizing it.
i'll see if i can prevent it from doing that

shazow:
Cool. :) Else could migrate it to a socket-level test (these are preferred these days, aside from the fact that they're all cohabitating in one giant file). Some similar tests:

https://github.com/shazow/urllib3/blob/master/test/with_dummyserver/test_socketlevel.py#L337
https://github.com/shazow/urllib3/blob/master/test/with_dummyserver/test_socketlevel.py#L29

matejcik:
and here we go, all green :)

shazow:
@matejcik You're basically the best. Thank you. :)

shazow:
Btw, you're welcome to add yourself to our CONTRIBUTORS.txt file!

PR #595, opened 2015-04-23
Decode information received from read_chunked in stream

### LINKED ISSUE #593: decode_content parameter to Response.stream() is not respected for chunked bodies
Originally kennethreitz/requests#2561.

It seems that if you hit an endpoint that sets both Content-Encoding: gzip and Transfer-Encoding: chunked, and then use `Response.stream()` with `decode_content=True`, we don't decompress the content. That represents an API change, and I'd argue a regression.

Should be a fairly simple fix though.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11842377-decode_content-parameter-to-response-stream-is-not-respected-for-chunked-bodies?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
Previously, when we read chunked data we always decoded it because it
was handled by the read method on the HTTPResponse. Now that it's a
separate method, we need to handle it. This is a quick fix for #593 to
address a regression in behaviour.

Fixes #593 

---

I confirmed this worked by linking my working copy of urllib3 into requests and testing against the site in the original issue on requests, e.g., by doing

```py
import requests
r = requests.get('https://liquidstate.atlassian.net/rest/api/2/serverInfo')
r.json()
```

I simply didn't have time to write a test case. Also IMO a better solution would be to make `read_chunked` handle this instead of `stream` but this was the fastest hotfix I could produce. I'm adding both @shazow and @Lukasa to my fork so they can dev on this if they want from the same branch

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -418,6 +418,29 @@ class TestResponse(unittest.TestCase):
             self.assertEqual(c, stream[i])
             i += 1
 
+    def test_mock_gzipped_transfer_encoding_chunked_decoded(self):
+        """Show that we can decode the gizpped and chunked body."""
+        def stream():
+            # Set up a generator to chunk the gzipped body
+            import zlib
+            compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
+            data = compress.compress(b'foobar')
+            data += compress.flush()
+            for i in range(0, len(data), 2):
+                yield data[i:i+2]
+
+        fp = MockChunkedEncodingResponse(list(stream()))
+        r = httplib.HTTPResponse(MockSock)
+        r.fp = fp
+        headers = {'transfer-encoding': 'chunked', 'content-encoding': 'gzip'}
+        resp = HTTPResponse(r, preload_content=False, headers=headers)
+
+        data = b''
+        for c in resp.stream(decode_content=True):
+            data += c
+
+        self.assertEqual(b'foobar', data)
+
     def test_mock_transfer_encoding_chunked_custom_read(self):
         stream = [b"foooo", b"bbbbaaaaar"]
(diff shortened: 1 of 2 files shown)


Lukasa:
\o/ This looks great! I'll write a test case for this this evening, I'm getting pretty good at them now.

test/test_response.py
```diff
@@ -517,7 +540,9 @@ def __init__(self, content):
 
     @staticmethod
     def _encode_chunk(chunk):
-        return '%X\r\n%s\r\n' % (len(chunk), chunk.decode())
+        # In the general case, we can't decode the chunk to unicdoe
```
Lukasa: Unic, doe?

Lukasa:
One slightly sarcastic comment inline.

sigmavirus24:
Updated to fix the typo

urllib3/response.py
```diff
@@ -284,9 +299,10 @@ def stream(self, amt=2**16, decode_content=None):
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
+        self._init_decoder()
         if self.chunked:
             for line in self.read_chunked(amt):
-                yield line
+                yield self._decode(line, decode_content, True)
```
shazow: In the future, it'd be nice to supply a named var or param for magic values like True.

sigmavirus24: I'll be happy to fix that when I add support to read_chunked to decode the content and move the use of `_decode` into `read_chunked`.

PR #597, opened 2015-04-24
Refactor fix for 593

sigmavirus24:


sigmavirus24:
https://travis-ci.org/shazow/urllib3/jobs/59822656 seems to be a transient failure.

urllib3/response.py
```diff
@@ -387,48 +386,67 @@ def readinto(self, b):
             b[:len(temp)] = temp
             return len(temp)
 
-    def read_chunked(self, amt=None):
+    def _get_next_chunk(self):
+        # First, we'll figure out length of a chunk and then
+        # we'll try to read it from socket.
+        if self.chunk_left is None:
+            line = self._fp.fp.readline()
+            line = line.decode()
```
Lukasa: What's this decode call doing here?

I missed it in the last code review, but it looks wrong.

urllib3/response.py
```diff
@@ -387,48 +386,67 @@ def readinto(self, b):
...
+            # See RFC 7230: Chunked Transfer Coding.
+            i = line.find(';')
+            if i >= 0:
+                line = line[:i]  # Strip chunk-extensions.
```
Lukasa: Might be nicer to do `line.split(';',1)[0]`

urllib3/response.py
```diff
@@ -387,48 +386,67 @@ def readinto(self, b):
             b[:len(temp)] = temp
             return len(temp)
 
-    def read_chunked(self, amt=None):
+    def _get_next_chunk(self):
```
Lukasa: The name of this method is confusing, because it doesn't get the chunk, it gets the chunk length.

urllib3/response.py
sigmavirus24: So I expect this is necessary on Python 3 because Python 3 doesn't have the same methods as strings do. I have an idea to avoid this though. I'll push it up shortly

urllib3/response.py
sigmavirus24: Done in c1317ff

urllib3/response.py
sigmavirus24: Done in c1317ff but using a slightly different way around the fact that `line` will be `bytes`.

sigmavirus24:
In #594 I added the `_decode` method but I renamed it in 7168db7. To me, even though I created it, the name was misleading. The usual implication of things dealing with bytes/strings named `decode` imply we should be receiving `unicode` back from the method but we were actually simply decompressing data so it is now `_decompress`. Let me know if you want me to strip 7168db7 for that reason.

sigmavirus24:
The latest failure was on pypy and looks like it would be intermittent.

sigmavirus24:
Relatedly, https://github.com/kennethreitz/requests/issues/2568 and I'm not happy with the fact that urllib3 is raising exceptions out of httplib at the moment (and violating the httplib state machine with the chunked reads). I'm going to see if we can refactor this some more to do better things.

Lukasa:
@sigmavirus24 I'm nervous about that, it might be better to refactor first and then attempt to fix the bugs in the refactored version, to cut down on the complexity of the diffs.

sigmavirus24:
Fair enough.

urllib3/response.py
```diff
@@ -126,8 +129,8 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
         # Are we using the chunked-style of transfer encoding?
         self.chunked = False
         self.chunk_left = None
-        tr_enc = self.headers.get('transfer-encoding', '')
-        if tr_enc.lower() == "chunked":
+        tr_enc = self.headers.get('transfer-encoding', '').lower()
+        if "chunked" in set(tr_enc.split(", ")):
```
shazow: Why cast it to a set? Seems like a performance penalty of instantiating the set without any benefit.

urllib3/response.py
```diff
@@ -183,7 +186,7 @@ def _init_decoder(self):
             if content_encoding in self.CONTENT_DECODERS:
                 self._decoder = _get_decoder(content_encoding)
 
-    def _decode(self, data, decode_content, flush_decoder):
+    def _decompress(self, data, decode_content, flush_decoder):
```
shazow: We do refer to this as decoding everywhere else, I'm -0.5 on renaming it to decompress for consistency.

urllib3/response.py
```diff
@@ -387,48 +389,63 @@ def readinto(self, b):
             b[:len(temp)] = temp
             return len(temp)
 
-    def read_chunked(self, amt=None):
+    def _update_chunk_length(self):
+        # First, we'll figure out length of a chunk and then
+        # we'll try to read it from socket.
+        if self.chunk_left is None:
```
shazow: Can we invert this?

```
    return
...
```

urllib3/response.py
```diff
@@ -387,48 +389,63 @@ def readinto(self, b):
...
+            line = self._fp.fp.readline()
+            line = CHUNKED_SPLIT_RE.split(line, 1)[0]
```
shazow: What's the benefit of using a regexp split over a normal split?

urllib3/response.py
shazow: >  The usual implication of things dealing with bytes/strings named decode imply we should be receiving unicode back from the method...

I agree that in the general Python context, decode means between bytes/strings, but I think it's fair that in a transport protocol context it can mean converting from some arbitrary encoding like a binary one (e.g. compression) to a native format.

urllib3/response.py
sigmavirus24: For some reason, I was under the impression that bytes objects didn't support `.split` but apparently I'm wrong:

```py
~ python3.4
Python 3.4.2 (default, Oct  8 2014, 19:29:52)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.51)] on darwin
>>> b'foo; bar'.split(b';')
[b'foo', b' bar']
```

Might have been a problem with 3.2 or 3.3 but I'm probably just wrong.

urllib3/response.py
sigmavirus24: Sure can

sigmavirus24: Fair enough

urllib3/response.py
sigmavirus24: Fair enough

urllib3/response.py
sigmavirus24: Sets will make lookup faster in a case where a server sends a bunch of encodings separated by commas. If you don't want to hedge against that case (which I've heard of happening with TomCat servers from a friend), we can remove it.

shazow: Lookups, yes. But set instantiation is way slower (lower bound is the same amount of work as just checking membership in a list). In this case, we'd be instantiating every time we're looking up, so it's not buying us anything. :)

sigmavirus24:
I rebased the branch to remove the commits that renamed `_decode` and added the regular expression. I also removed the call to set and changed the logic in `_update_chunk_length`.

urllib3/response.py
```diff
@@ -387,48 +386,64 @@ def readinto(self, b):
...
+        line = self._fp.fp.readline()
+        line = line.split(b';', 1)[0]
+        try:
+            self.chunk_left = int(line, 16)
+        except ValueError:
+            # Invalid chunked protocol response, abort.
+            self.close()
+            raise httplib.IncompleteRead(line)
+
+    def _handle_chunk(self, amt):
```
shazow: This was renamed back from update_chunk_length, is that intentional?

sigmavirus24: Nope. Must have been caught in the pruning of commits.

sigmavirus24: Fixed in 6c1dcdb

urllib3/response.py
```diff
@@ -126,8 +126,8 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
         # Are we using the chunked-style of transfer encoding?
         self.chunked = False
         self.chunk_left = None
-        tr_enc = self.headers.get('transfer-encoding', '')
-        if tr_enc.lower() == "chunked":
+        tr_enc = self.headers.get('transfer-encoding', '').lower()
+        if "chunked" in tr_enc.split(", "):
```
Lukasa: It's probably better to split on comma, and then strip each element. In practice the space-separator is a bit hit and miss.

urllib3/response.py
```diff
@@ -387,48 +386,64 @@ def readinto(self, b):
...
-                self._fp._safe_read(2)  # Toss the CRLF at the end of the chunk.
-                self.chunk_left = None
-                yield value
-            else:  # amt > self.chunk_left
-                yield self._fp._safe_read(self.chunk_left)
-                self._fp._safe_read(2)  # Toss the CRLF at the end of the chunk.
-                self.chunk_left = None
+            self._update_chunk_length()
+            if self.chunk_left == 0:
+                break
```
Lukasa: I'm worried this break is too early, and in practice will leave us with a trailing '\r\n'. This also doesn't account for trailers (if present). Both of these will bust connection re-use.

Lukasa: To be clear, I'm not proposing we actually expose trailers, just eat them.

sigmavirus24: @Lukasa if you read further down into `read_chunked` there's a while look to consume the rest of the body. Does that ease your concerns?

urllib3/response.py
sigmavirus24: Done in  563f0e8

urllib3/response.py
Lukasa: Ah, yes it does: lost it in the diff. =(

sigmavirus24: Yeah. Sometimes it's easier to clone a PR and read the code in a proper editor. This is one thing I like about working on OpenStack, that's thoroughly encouraged and very much necessary sometimes.

sigmavirus24:
@shazow @Lukasa any other feedback?

shazow:
Looks good to me. If @Lukasa is happy, I'll hit the button. :)

Lukasa:
:cake: :sparkles: LGTM :shipit:

PR #599, opened 2015-04-27
Fix Chunked keepalives

### LINKED ISSUE #598: Keep-alive broken in 1.10.3 with chunked transfer-encoding
Per kennethreitz/requests#2568, starting in urllib3 1.10.3 an exception is thrown if a connection with chunked transfer-encoding is reused.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11976603-keep-alive-broken-in-1-10-3-with-chunked-transfer-encoding?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

t-8ch:
I am not sure if this is the right way. The test should be picked anyways.

Fixes #598.

/cc @shazow @sigmavirus24

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -21,9 +21,6 @@ log = logging.getLogger(__name__)
 
 class Response(object):
     def __init__(self, body='', status='200 OK', headers=None):
-        if not isinstance(body, bytes):
-            body = body.encode('utf8')
-
         self.body = body
         self.status = status
         self.headers = headers or [("Content-type", "text/plain")]
@@ -34,7 +31,20 @@ class Response(object):
         for header,value in self.headers:
             request_handler.add_header(header,value)
 
-        request_handler.write(self.body)
+        # chunked
+        if isinstance(self.body, list):
+            for item in self.body:
+                if not isinstance(item, bytes):
+                    item = item.encode('utf8')
+            request_handler.write(item)
+            request_handler.flush()
+        else:
+            body = self.body
+            if not isinstance(body, bytes):
+                body = body.encode('utf8')
+
+            request_handler.write(body)
+
 
 RETRY_TEST_NAMES = collections.defaultdict(int)
 
@@ -207,6 +217,9 @@ class TestingApp(RequestHandler):
             return Response("Retry successful!")
         else:
             return Response("need to keep retrying!", status="418 I'm A Teapot")
+
+    def chunked(self, request):
+        return Response(['123'] * 4)
 
     def shutdown(self, request):
         sys.exit()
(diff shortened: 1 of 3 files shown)


dummyserver/handlers.py
```diff
@@ -34,7 +31,20 @@ def __call__(self, request_handler):
...
+                    item = item.encode('utf8')
+            request_handler.write(item)
+            request_handler.flush()
+        else:
+            body = self.body
+            if not isinstance(body, bytes):
+                body = body.encode('utf8')
+
+            request_handler.write(body)
+
```
Lukasa: What's all this for? As far as I know, chunked responses never return a list as the body.

Lukasa:
This looks like it fixes the core issue, but there seems to be some unrelated work in here. Let's sort that out before merging. =)

dummyserver/handlers.py
t-8ch: This is meant to make the `chunked()`-endpoint work

Lukasa: Ah, fair enough. I take it you confirmed that this genuinely does write correctly-framed chunked transfer encoding? (As in, the request_handler correctly applies the chunked framing)

t-8ch: At least in triggers our bug :-)
(And I reconfirmed it on my side)

t-8ch: Woah, wait. It seems to discard output (but is chunked...) I have to check again...

t-8ch: I messed up during final cleanup. It would only emit the last chunk. Fixed now

Lukasa: On the one hand, I'm glad I asked! On the other hand, now the tests failed. =(

Lukasa:
Tests failing, otherwise I'm happy. =)

t-8ch:
Sometimes we have resource warnings in py3....

Lukasa:
Dem intermittent test failures.

The resource warnings aren't *necessarily* a surprise (they happen when used connection pools go out of scope) but if they're intermittent then they are a surprise.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -618,6 +618,25 @@ def test_source_address_error(self):
...
+
+        for _ in range(x):
+            response = self.pool.request(
+                    'GET',
+                    '/chunked',
+                    headers={
+                        'Connection': 'keep-alive',
+                        },
+                    preload_content=False,
+                    retries=0,
```
shazow: `retries=False` disables retries (and exception wrapping).

t-8ch: @shazow changed

shazow:
Looks alright to me, lmk when you deem it done. :)

t-8ch:
@shazow If you like it, it is done :-)

PR #603, opened 2015-04-28
Fix #602: don't override existing warning config at import

### LINKED ISSUE #602: Really don't spam InsecurePlatformWarning
urllib3 should configure its warnings using append=True to avoid overriding the user's preferences as specified with python -W or PYTHONWARNINGS.

If this issue were fixed, the user could work around pypa/pip#2681 with

    export PYTHONWARNINGS="ignore:A true SSLContext object is not available"

Additionally, the urllib3 docs are very unclear about why this is considered worth warning the end user about, particularly given that adding this strange message has effectively introduced a bug in hundreds of other projects.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12376031-really-don-t-spam-insecureplatformwarning?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

tlynn:
Filter warnings with append=True to fix issue #602.

No test since it affects import-time behaviour and it's a two-word change, but I've given it a quick try and it seems to do the right thing.

I haven't changed disable_warnings(), since if people are calling that they presumably mean it, and can call the underlying method if not.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -145,6 +145,9 @@ In chronological order:
 * Tomas Tomecek <ttomecek@redhat.com>
   * Implemented generator for getting chunks from chunked responses.
 
+* tlynn <https://github.com/tlynn>
+  * Respect the warning preferences at import.
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
 
--- a/urllib3/__init__.py
+++ b/urllib3/__init__.py
@@ -57,9 +57,10 @@ del NullHandler
 
 import warnings
 # SecurityWarning's always go off by default.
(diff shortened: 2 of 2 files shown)


Lukasa:
\o/ LGTM.

tlynn:
I don't think the Travis test failure is likely to be related to this commit.

shazow:
Excellent, thank you @tlynn! If you're also up to making our docs more clear for people who find themselves in your situation, that'd be very appreciated. :)

PR #604, opened 2015-04-28
SSL session reuse support through pyOpenSSL

xnulinu:
Added feature to support SSL resume through pyOpenSSL. The more discussion can be found at 
https://github.com/shazow/urllib3/issues/590

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -89,6 +89,12 @@ DEFAULT_SSL_CIPHER_LIST = util.ssl_.DEFAULT_CIPHERS
 orig_util_HAS_SNI = util.HAS_SNI
 orig_connection_ssl_wrap_socket = connection.ssl_wrap_socket
 
+class SSLResume:
+    def __init__(self, context, initial_session):
+        self.context = context
+        self.initial_session = initial_session
+
+ssl_session = dict()
 
 def inject_into_urllib3():
     'Monkey-patch urllib3 with PyOpenSSL-backed SSL-support.'
@@ -252,7 +258,12 @@ def _verify_callback(cnx, x509, err_no, err_depth, return_code):
 def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
                     ca_certs=None, server_hostname=None,
                     ssl_version=None):
-    ctx = OpenSSL.SSL.Context(_openssl_versions[ssl_version])
+
+    ssl_resume = ssl_session.get(server_hostname)
+    if ssl_resume:
+        ctx = ssl_resume.context
+    else:
+        ctx = OpenSSL.SSL.Context(_openssl_versions[ssl_version])
     if certfile:
         keyfile = keyfile or certfile  # Match behaviour of the normal python ssl library
         ctx.use_certificate_file(certfile)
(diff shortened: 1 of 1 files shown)


t-8ch:
`Connection.get_session()` has only been introduced in pyopenssl 0.14, so this will break in older versions.

shazow:
Mmm. Would be nice to add support but it must degrade gracefully.

xnulinu:
@shazow  @t-8ch 

Thanks for the feedback. I was thinking about the compatibility issues with older version of pyOpenSSL. Can someone point me the better way to check method existence so that I can integrate in the code.

t-8ch:
@xnulinu I just began to wonder if this really brings us very much, as we already have connection pooling, which also reuses the established connections.

Otherwise we should also respect the used port in the cache.

shazow:
You can do something like...

```python
if hasattr(someobj, 'somenewmethod'):
    do new stuff
```

If you need to do that check in different places, it might be better to just set a variable somewhere, like:

```python
self.has_ssl_resume = hasattr(someobj, 'somenewmethod')
```

shazow:
> I just began to wonder if this really brings us very much, as we already have connection pooling, which also reuses the established connections.

Good point. I guess this lets us re-use SSL sessions across a pool of connections, which is not the default behaviour anyways (default pool size is 1). Maybe this should be something that needs to be enabled explicitly?

xnulinu:
@t-8ch 
SSL resume is more about avoiding full SSL handshake each time a client request a resources. This will save complete round-trip time and avoids the costlier part of full SSL handshake.  This feature with the Connection Pool will increase latency and computation time. 

About the used port on the cache, I also think we should use that. I don't know how to retrieve in the pyOpenSSL ssl_wrap_socket method.

shazow:
Maybe a better place to cache SSL contexts would be within the HTTPSConnectionPool instances?

t-8ch:
@xnulinu You can do `sock.getpeername()`

xnulinu:
@t-8ch 

that will do it.

shazow:
Shouldn't need to get the host or peer if the context is saved in the HTTPSConnectionPool right after the first socket is created.

xnulinu:
@shazow 

If we do on the HTTPSConnectionPool, how we are going to retrieve the context in the ssl_wrap_socket  method. Are you thinking to do the method change or use some setattr magic.

shazow:
The `WrappedSocket` that our pyopenssl module returns has the connection context fromwhich we can get the session. Could add a `ssl_context_session(connection)` helper into `util/ssl_.py` which attempts to extract a re-usable session from a connection, otherwise returning `None` (right now only supporting the pyopenssl thing, but ultimately maybe py3+ will get support too?).

Then in `HTTPSConnectionPool` can attempt to re-use an SSL session if it has one, else just keep doing what it normally does (and save the next SSL session it gets).

t-8ch:
Are connectionpools keyed by DNS names? If so we have to use `getpeername` else we may reuse a session and the DNS name may resolve to a different host with which we can't resume the session initiated by the other host.
This Means we have to query the cache after we created the socket.
Or should we prefer those IP addresses of DNS names we already have sessions for?

shazow:
You're right, the PoolManager is keying on the (scheme, hostname, port) right now. Maybe it shouldn't be, but then we'd have to resolve the host earlier than later. :/

t-8ch:
I think the PoolManager should be allowed to. But the SessionCache can't.

shazow:
While I'd love to re-use our pooling logic whenever possible (rather than re-implementing it in various islands). For now, since this only applies to a sub-version of pyopenssl, we can keep it isolated.

t-8ch:
OTOH, different connectionpools to the same endpoint may very well have different SSL configurations (in our case SSL versions). So caching sessions globally can lead to inconsistencies.

And: Can we reuse a session while it is still in use by another connection? Which will easily, as we keep connections open.

jonparrott:

devshorts:
@theacodes I'd definitely still like this PR to come to fruition. Being able to re-use ssl tickets is a big performance booster when dealing with many requests.

PR #605, opened 2015-04-28
Learn from httplib how to handle chunked HEAD requests

### LINKED ISSUE #601: No tolerance for responses to head requests that indicate chunking.
As shown [here](https://ptpb.pw/r/AFSJnZhspw4EkC13PK8TBHpNqMDH), if we receive a redirect response that has no body, but claims Transfer-Encoding: chunked, we will hang indefinitely waiting for a chunked response we never see.

This has been observed with requests 2.6.2 (containing the new chunked handling) but not with requests 2.6.0 (containing the old broke stuff), which moves along just fine.

Now, it's worth noting that Google has totally screwed up here, because the response is invalid, as this Wireshark trace shows:

```
HEAD /~r/DeveloperChronicles/~3/iauy0NfhAus/hitchhikers-guide-to-modern-android.html HTTP/1.1
Host: feedproxy.google.com
Connection: keep-alive
Accept-Encoding: gzip, deflate
Accept: */*
User-Agent: python-requests/2.6.2 CPython/2.7.8 Windows/7

HTTP/1.1 301 Moved Permanently
Location: http://www.devchronicles.com/2015/03/hitchhikers-guide-to-modern-android.html?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+DeveloperChronicles+%28Developer+Chronicles%29
Content-Type: text/html; charset=UTF-8
Date: Tue, 28 Apr 2015 09:56:23 GMT
Expires: Tue, 28 Apr 2015 09:56:23 GMT
Cache-Control: private, max-age=0
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Server: GSE
Alternate-Protocol: 80:quic,p=1
Transfer-Encoding: chunked
Accept-Ranges: none
Vary: Accept-Encoding

```

Note that Google have sent Transfer-Encoding: chunked, but no body. This is, as far as I can tell, against all the relevant RFCs: a zero length body has a valid chunked representation (`0\r\n\r\n`). 301 is also not defined as having no body.

Regardless, however, curl gets this right, so we should too if at all possible. We should also confirm whether this is a problem for older versions of urllib3 or whether they too function correctly.

Arguably this may want to block a requests release, @sigmavirus24.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12364347-no-tolerance-for-responses-to-head-requests-that-indicate-chunking?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
When we perform a HEAD request and get a response with a chunked
transfer-encoding, there's no point in looking for a body at all. We
should just close the response immediately.

Fixes #601 

cc @Lukasa this needs (probably) socket-level tests

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -432,14 +432,14 @@ class HTTPResponse(io.IOBase):
             'content-encoding' header.
         """
         self._init_decoder()
-        # FIXME: Rewrite this method and make it a class with
-        #        a better structured logic.
+        # FIXME: Rewrite this method and make it a class with a better structured logic.
         if not self.chunked:
             raise ResponseNotChunked("Response is not chunked. "
                 "Header 'transfer-encoding: chunked' is missing.")
 
-        if (self._original_response and
-                self._original_response._method.upper() == 'HEAD'):
+        if self._original_response and self._original_response._method.upper() == 'HEAD':
+            # Don't bother reading the body of a HEAD request.
+            # FIXME: Can we do this somehow without accessing private httplib _method?
             self._original_response.close()
             return
 
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -437,6 +437,12 @@ class HTTPResponse(io.IOBase):
         if not self.chunked:
             raise ResponseNotChunked("Response is not chunked. "
                 "Header 'transfer-encoding: chunked' is missing.")
+
+        if (self._original_response and
+                self._original_response._method.upper() == 'HEAD'):
+            self._original_response.close()
+            return
+
         while True:
             self._update_chunk_length()
             if self.chunk_left == 0:
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -616,3 +616,33 @@ class TestHeaders(SocketDummyServerTestCase):
         HEADERS = {'Content-Length': '0', 'Content-type': 'text/plain'}
         r = pool.request('GET', '/')
         self.assertEqual(HEADERS, dict(r.headers.items())) # to preserve case sensitivity
+
+
+class TestHEAD(SocketDummyServerTestCase):
+    def test_chunked_head_response_does_not_hang(self):
+        handler = create_response_handler(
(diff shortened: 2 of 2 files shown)


shazow:
Edited with some style tweaks.

+1 to a socketlevel test, should be fairly easy.

urllib3/response.py
```diff
@@ -432,11 +432,17 @@ def read_chunked(self, amt=None, decode_content=None):
...
-        # FIXME: Rewrite this method and make it a class with
-        #        a better structured logic.
+        # FIXME: Rewrite this method and make it a class with a better structured logic.
         if not self.chunked:
             raise ResponseNotChunked("Response is not chunked. "
                 "Header 'transfer-encoding: chunked' is missing.")
+
+        if self._original_response and self._original_response._method.upper() == 'HEAD':
+            # Don't bother reading the body of a HEAD request.
+            # FIXME: Can we do this somehow without accessing private httplib _method?
```
sigmavirus24: Wasn't there an issue about attaching request information to a response somehow? Or was that about tracking redirect history somehow? Was there one for each? I forget.

shazow: https://github.com/shazow/urllib3/issues/576 maybe?

Lukasa:
Your testing fairy has visited and provided some tests. I also made sure that I covered some other bits of HEAD method logic, because coverage is not the only metric of testing success.

shazow:
I like our 3-way collab on this PR. Good job all. :)

Lukasa:
Clearly we need @t-8ch to land a commit here to get the full set. =D

t-8ch:
I would love to, but I have to leave :-)
Maybe if you wait till tonight. But better get this out fast :-)

t-8ch:
And ofc after all of you went over this, it is perfect by definition!

PR #608, opened 2015-04-30
[RFC] add preliminary HSTS support

t-8ch:
_Do not merge yet_

TODO:
- [ ] _more_ tests!
- [ ] reread the RFC and check every point
- [ ] find the proper place for the logic
- [x] Make HSTS failures fatal
- [x] Ignore IP addresses
- [x] make storage abstract
- [x] process HSTS also in lower connection classes
- [ ] documentation

See also #607


urllib3/util/hsts.py
```diff
@@ -0,0 +1,167 @@
...
+        return self._records.values()
+
+    def store_record(self, record):
+        self._records[record.domain] = record
+
+    def invalidate_record(self, domain):
+        self._records.pop(domain, None)
+
+
+class HSTSManager(object):
```
shazow: IMO this kind of stuff should live outside of util.

t-8ch: sure

shazow:
FYI I posted some high-level thoughts on the sister PR, mostly relevant to this one too: https://github.com/shazow/urllib3/pull/607#issuecomment-98037005

t-8ch:
@shazow

* The HSTSManager is already at the PoolManager level. To function properly it
  has to, as only there we have access to http and https connections.

* About the location of the code: Currently we need the code in the PoolManager
  itself. If there was a way to inject this functionality we very well could
  make this an external module.

* Naming: sure

t-8ch:
The `util.hsts` module is meant to be moved to the other util modules or a new one.

t-8ch:
So, now we have special exceptions on HSTS violations.
There are still some policy questions open:

* Do we also handle HSTS lower in the abstraction stack?
  * Warnings if we have a plain HTTPConnection with a hsts entry?
  * Warnings/Errors in unverified httpsconnections
* Do we want to add policies from unverified https connections to the
  database? (I'll recheck the RFC)
* Should the HSTS violation be a subclass of SSLErrors?
* Maybe we want some sort of PolicyManager, which will unify a HSTS Manager and
  a HPKP manage

Depending on the answers to those questions I will move the checking from the
current ugly place somewhere else.

/cc @shazow @sigmavirus24 @lukasa

urllib3/poolmanager.py
```diff
@@ -155,10 +166,21 @@ def urlopen(self, method, url, redirect=True, **kw):
...
+                response = conn.urlopen(method, u.request_uri, **kw)
+        except SSLError as e:
+            if hsts_active:
+                raise HSTSViolation(e)
+            else:
+                raise
+
+        if self.hsts_manager:
+            # As only PoolManager.urlopen follows redirects we know that
+            # response.scheme == request.scheme
```
Lukasa: That's not true, `ConnectionPool.urlopen` does too.

Lukasa:
> Warnings if we have a plain HTTPConnection with a hsts entry?

If someone deliberately takes a HTTPConnectionPool and makes a request to a HSTS host, they've quite explicitly done an end-run around the HSTS stuff. That's on them. A warning-level log is probably sufficient.

> Warnings/Errors in unverified httpsconnections

Again, to do it the user has to have done this on purpose, so a warning level log is good enough.

> Do we want to add policies from unverified https connections to the database? (I'll recheck the RFC)

No.

> Should the HSTS violation be a subclass of SSLErrors?

I think so, yes. =)

> Maybe we want some sort of PolicyManager, which will unify a HSTS Manager and a HPKP manage

We will. =)

urllib3/poolmanager.py
t-8ch: Then we should probably refactor this

PR #607, opened 2015-04-30
HPKP

Lukasa:
HPKP is cool, and I think we should be able to do it.

This issue tracks us (me?) spiking out how best to do it. I believe we can do it, and continue to be Python HTTP security wizards.

People who may want to watch this issue: @sigmavirus24 @reaperhulk @dstufft @tiran.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/12612776-hpkp?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>


Lukasa:
Ok, I've turned this into a PR with my initial code attached.

There's no way this code passes UTs yet, and I haven't written any of my own either, so it's definitely a work in progress. However, now is absolutely the time to provide feedback on the user interface I've got proposed here.

There's also a parallel proposal for HSTS over in #608, thanks to @t-8ch, which may or may not want to be merged with this at some point (or vice versa).

Lukasa:
@joelverhagen Indeed, but if we have to go outside of the stdlib anyway (to pyasn1) it'll just be better to use cryptography, I suspect.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,239 @@
+# -*- coding: utf-8 -*-
+import collections
+import time
+
+from urllib3.exceptions import HPKPError
```
t-8ch: relative import?

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,239 @@
...
+            # First, check whether this host is still valid. If it's not, throw
+            # it away and move on.
+            if (host.start_date + host.max_age) > time.time():
+                self.db.invalidate_host(host)
+                continue
+
+            # Now, check whether this KnownPinnedHost applies to this
+            # connection. It does if the domain either exactly matches the KPH
+            # domain, or if the KPH has include_subdomains set to True.
+            if domain == host.domain or host.include_subdomains:
```
t-8ch: Shouldn't this actually check if  `domain` is a subdomain of `host.domain`?

urllib3/util/hpkp.py
Lukasa: I wasn't confident that a relative import beginning with `..` would go well on Python 2.6, and didn't really want to break my flow to check. Will it?

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,239 @@
...
+def parse_public_key_pins(header, domain):
+    """
+    Parses a Public-Key-Pins header, returning a KnownPinnedHost. Invalid
+    headers cause exceptions to be thrown.
+
+    :param header: The PKP header value.
+    :param domain: The domain name used in the request. Must not be an IP
+        address.
+    """
+    # TODO: This method is super complicated, refactor.
```
t-8ch: `cookielib.split_header_words` should make this easier

urllib3/util/hpkp.py
Lukasa: It doesn't need to, because it's handled in another method. Agree that this should be commented though.

urllib3/util/hpkp.py
t-8ch: We do it in many places. It should work :-)

jpf:
In terms of APIs: Based on my limited experience, I would suggest designing the API to work with "pinset" objects, meaning, objects that contain one or more HPKP pins. Requests inspired me to use that pattern in [PublicKeyPinsetConnectionPool](https://github.com/okta/okta-openvpn/blob/master/okta_openvpn.py#L66) and that pattern ended up coming in very handy when writing [tests](https://github.com/okta/okta-openvpn/blob/master/tests/test_ssl_public_key_pinning.py#L63)].

Lukasa:
@jpf At this kind of relatively low level, I'm mostly considering implementing this in a manner that's pretty close to the description the RFC uses. In this case, the 'pinset' notion corresponds roughly to the `KnownPinnedHost` namedtuple the code uses: the difference is that the `KnownPinnedHost` contains the other information associated with a pin, such as domain and max age.

jpf:
Does this code take the the ability to use a "[pre loaded pin list](https://tools.ietf.org/html/rfc7469#section-2.7)" into account? I know that the RFC marks this feature as optional but it's something that's important for the code I maintain.

urllib3/util/hpkp.py
Lukasa: Ok, I clarified in the docstring of this method. Shout if you think that's insufficient.

Lukasa:
@jpf It depends on your HPKP database, but in the general case, sure. Provide a HPKP database that is preloaded with `KnownPinnedHost` objects that have an appropriately set `max_age` and `start_date` and you should be good to go. =) It should be really easy to write a factory function that automatically uses those as well.

jpf:
@Lukasa ah, after reading the code a little bit more, I can see how I would make this work. It looks like I would load up the HPKP database and then run `store_host` for each pin that I wanted to pre-load, right?

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,269 @@
...
+        Processes a HTTPS response for HPKP.
+
+        This method checks whether the PKP header is present on the request,
+        and if it is updates its store of key pins. It also validates the
+        connection against the HPKP header, confirming that it does pass at
+        this stage.
+        """
+        # TODO: Handle ip address domains.
+        # If no PKP header is provided, then no action is required.
+        pkp = response.getheader('public-key-pins')
```
alex: This should only be respected on responses over TLS, right?

Lukasa: Correct. In practice, this method is only called from the `HTTPSConnectionPool` class, which I believe solves this concern, though @shazow should confirm.

Lukasa:
@jpf Indeed. =) For persistent-storage databases, that should also take care of removing any prior pins you had.

urllib3/util/hpkp.py
alex: Ah, sorry, it's a bit hard to tell.

On Thu, Apr 30, 2015 at 5:39 PM, Cory Benfield <notifications@github.com>
wrote:

> In urllib3/util/hpkp.py
> <https://github.com/shazow/urllib3/pull/607#discussion_r29473326>:
>
> > +        # pins.
> > +        # TODO: Actually do this!
> > +        return
> > +
> > +    def process_response(self, domain, response):
> > +        """
> > +        Processes a HTTPS response for HPKP.
> > +
> > +        This method checks whether the PKP header is present on the request,
> > +        and if it is updates its store of key pins. It also validates the
> > +        connection against the HPKP header, confirming that it does pass at
> > +        this stage.
> > +        """
> > +        # TODO: Handle ip address domains.
> > +        # If no PKP header is provided, then no action is required.
> > +        pkp = response.getheader('public-key-pins')
>
> Correct. In practice, this method is only called from the
> HTTPSConnectionPool class, which I believe solves this concern, though
> @shazow <https://github.com/shazow> should confirm.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shazow/urllib3/pull/607/files#r29473326>.
>



-- 
"I disapprove of what you say, but I will defend to the death your right to
say it." -- Evelyn Beatrice Hall (summarizing Voltaire)
"The people's good is the highest law." -- Cicero
GPG Key fingerprint: 125F 5C67 DFE9 4084

Lukasa: Fair enough, it's a big diff (already), and it might be the case that we want to move some of this logic into these classes anyway.

jpf:
@Lukasa perfect. This all looks great to me then!

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,269 @@
...
+        host = self._find_valid_pinned_host(domain, hosts)
+        if host is None:
+            # Again, no previous pinning was done on this connection, so
+            # there's no validation to do.
+            return
+
+        # If we got this far, we have a KnownPinnedHost in hand that applies to
+        # this connection. Get the certificate and check it matches one of the
+        # pins.
+        # TODO: Actually do this!
```
jpf: What's the plan for [reporting pin validation failures](https://tools.ietf.org/html/rfc7469#section-3)?

Lukasa: I explicitly do not plan for urllib3 to do this automatically: we'll throw exceptions with the relevant data on them. urllib3 really shouldn't be making unsolicited POSTs.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,269 @@
...
+        # exit.
+        if not pin.max_age or (pin.max_age + pin.start_date <= time.time()):
+            self.db.invalidate_host(domain)
+            return
+
+        # Otherwise, we can store off this pin.
+        self.db.store_host(pin)
+
+        # TODO: Check if the RFC wants us to immediately validate the
+        # connection: if it does, do it.
```
jpf: I *think* that this answers your question? https://tools.ietf.org/html/rfc7469#section-2.6

alex: I would think we actually need to check both before and after. If you're already in the HPKP you need to check the cert, and then update the pins and revalidate.

urllib3/util/hpkp.py
jpf: Yes, that sounds reasonable. What are your thoughts on providing some code that would make it easier for a developer to handle this? Maybe this could be a pre-written exception handler that isn't used by default? Or maybe it could just be a method that would give you the payload you could POST yourself?

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,269 @@
...
+        if name == 'pin-sha256':
+            value = unquote_string(value)
+            pin_directives.append(value)
+        elif name == 'max-age':
+            # There may only be one max-age directive. Be careful about
+            # policing this, we don't want to accidentally allow more than one.
+            if max_age:
+                raise HPKPError("Multiple max-age directives in PKP header")
+
+            max_age = int(unquote_string(value))
```
jpf: What are your thoughts on the RFC's suggestions for a "maximum" max-age: https://tools.ietf.org/html/rfc7469#section-4.1 ?

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,269 @@
...
+            include_subdomains = True
+        elif name == 'report-uri':
+            if report_uri:
+                raise HPKPError("Multiple report-uri directives in PKP header")
+
+            report_uri = unquote_string(value)
+
+    if not pin_directives:
+        raise HPKPError("No pin directives in PKP header")
+
```
jpf: The RFC requires that UAs require [backup pins](https://tools.ietf.org/html/rfc7469#section-4.3), is this something that would be checked here?

shazow:
Some high-level thoughts before I fall asleep:

- Not a big fan of adding more things to the Connection/ConnectionPool init and set_cert and such. Does it have to be specified during construction? Does it make sense for an HPKPManager to be at the PoolManager level or higher? (Maybe with the appropriate connection-level hooks added... Hrm my thoughts aren't clear here, need to think about this more.)
- HPKP probably doesn't belong in util, those are more for generally-useful helpers that don't require any awareness of Connection/ConnectionPool/Request/Response objects/behaviour. Basically not specific to urllib3. Maybe a urllib3/hpkp/*.py subpackage would be better?
- Can we think of a better name than HPKPStuff? KeyPinDatabase?
- Would it make sense to separate the idea of the key pin database and the nuances of HPKP itself? (At least as far as semantics/naming/module homes go)
- Would it be feasible to introduce this as an external or contrib module first, keeping the core code changes to a minimum?
- s/EphemeralDatabase/MemoryDatabase/ (more standard terminology in this context I think)

urllib3/util/hpkp.py
Lukasa: Yeah, this is the bit I actually needed from the spec:

>  - The TLS connection was authenticated with a certificate chain containing at least one of the SPKI structures indicated by at least one of the given SPKI Fingerprints (see Section 2.6).
> - The given set of Pins contains at least one Pin that does NOT refer to an SPKI in the certificate chain.  (That is, the host must set a Backup Pin; see Section 4.3.)

We need to do both of these checks here.

urllib3/util/hpkp.py
Lukasa: I think the second one is the better approach, at least initially. =)

urllib3/util/hpkp.py
Lukasa: I haven't thought that far ahead, but I'm not inherently opposed to it. It's an easy enhancement to either the HPKP manager or the HPKP database.

urllib3/util/hpkp.py
Lukasa: No, we'll check in `process_response`, while we handle the other PKP header logic.

Lukasa:
> Not a big fan of adding more things to the Connection/ConnectionPool init and set_cert and such. Does it have to be specified during construction? 

No, but it has to be present during the `connect` method, or used shortly after. HPKP only works if you ensure that you do not send/receive any non-TLS data on that connection until you've validated the key against the pin.

> Does it make sense for an HPKPManager to be at the PoolManager level or higher? (Maybe with the appropriate connection-level hooks added... Hrm my thoughts aren't clear here, need to think about this more.)

Yes, I think so, I think my tired mind missed that and put it at ConnectionPool level. I think the PoolManager could probably add it as a transparent kwarg on `urlopen`.

That said, it makes requests' life a little trickier because we play with ConnectionPools directly. Still, I write it, I can work around it. :wink:

I'd like to see a proposal from you on this though, because I feel a bit weird about the architecture too, and you know the design of urllib3 so much better than I do.

> Maybe a urllib3/hpkp/*.py subpackage would be better?

and

> Would it be feasible to introduce this as an external or contrib module first, keeping the core code changes to a minimum?

So, we can definitely move hpkp.py out of util, I have no problems with that.

Right now the core code changes are very small: one method call in `VerifiedHTTPSConnection.connect` and one in `ConnectionPool.urlopen`. We *could* remove the call in `urlopen` and require users to handle that manually, but I don't think we can remove the call in `connect` under any circumstances, as I said above.

I'm happy to move the module to `contrib`, or even put it up on PyPI, but urllib3 will need at minimum those changes. What would you like to do here?

> Can we think of a better name than HPKPStuff? KeyPinDatabase?

Almost certainly. Names are tricky.

> Would it make sense to separate the idea of the key pin database and the nuances of HPKP itself? (At least as far as semantics/naming/module homes go)

Not sure. If nothing else I could create a whole `hpkp` module and have a `database` submodule in it (`from urllib3.hpkp.database import BaseKeyDatabase` or something).

> s/EphemeralDatabase/MemoryDatabase/ (more standard terminology in this context I think)

Yup, that's a good change.

Lukasa:
So the biggest outstanding problem at the moment is that HPKP lets you pin *any* point of the cert chain, including the root. Unfortunately, neither PyOpenSSL nor the standard library appear to let us check that whole chain: the stdlib only lets us check the leaf, and PyOpenSSL lets us check all but the root.

We *can* still ship HPKP in this state, but it's not really fully useful, and potentially a bit dangerous, so I'm going to chat with some PyCA people to see if we can coerce PyOpenSSL into giving us the relevant root cert as well.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,334 @@
...
+            value = unquote_string(value)
+            pin_directives.append(value)
+        elif name == 'max-age':
+            # There may only be one max-age directive. Be careful about
+            # policing this, we don't want to accidentally allow more than one.
+            if max_age:
+                raise HPKPError("Multiple max-age directives in PKP header")
+
+            max_age = int(unquote_string(value))
+        elif name == 'includeSubDomains':
```
t-8ch: names of directives are case insensitive

urllib3/util/hpkp.py
t-8ch: In its current form this accepts negative values, which it shouldn't.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,334 @@
...
+    if not pin_directives:
+        raise HPKPError("No pin directives in PKP header")
+
+    # LUKASA: Is the use of time.time() safe here? Probably not.
+    return KnownPinnedHost(
+        domain, pin_directives, max_age, include_subdomains, report_uri, time.time()
+    )
+
+
+def validate_connection(self, connection, host, shortcut=True):
```
t-8ch: This has the same name, as a method of the manager. And it is called from inside the manager. This may lead to confusion, when people try to follow the code.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,334 @@
...
+        host = self._find_valid_pinned_host(domain, hosts)
+        if host is None:
+            # Again, no previous pinning was done on this connection, so
+            # there's no validation to do.
+            return
+
+        # If we got this far, we have a KnownPinnedHost in hand that applies to
+        # this connection. Get the certificate and check it matches one of the
+        # pins. If it doesn't we'll blow up here.
+        return self._validate_connection(socket, host)
```
t-8ch: This is now called `validate_connection` (no method, no underscore)

urllib3/util/hpkp.py
Lukasa: You're quite right.

urllib3/util/hpkp.py
Lukasa: Good catch @t-8ch, thanks!

urllib3/util/hpkp.py
t-8ch: Shameless plug: Maybe `parse_directives_header` from my HSTS branch hmakes this easier

urllib3/util/hpkp.py
Lukasa: Fixed, thanks!

urllib3/util/hpkp.py
Lukasa: Agreed, changed to `_validate_pin`.

urllib3/util/hpkp.py
Lukasa: It almost certainly does, and depending on the order in which these patches land we should switch to using common code.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,336 @@
...
+        else:
+            non_match = True
+
+    if not match:
+        raise HPKPError("Failed to validate trust chain with HPKP!")
+
+    return match and non_match
+
+
+def _certificate_in_pins(der_certificate, host):
```
t-8ch: Maybe this should be split into `pubkey_fingerprint(der_cert, hash=hashlib.sha256)`
and a function which actually does the verification.

Lukasa:
So after further discussion, it seems that what we'll actually have to do here is set the application verification callback (via `SSL_CTX_set_cert_verify_callback`) to pull the certificate chain out at that point, where we should be able to get the trust anchor as well. We can then call `X509_cert_verify` to maintain the default certificate verification behaviour. This will allow us to do the HPKP verification outside the verify callback, so we can throw handy Python exceptions and generally not hate ourselves. (Well, it's OpenSSL, so more accurately so we can hate ourselves less.)

@tiran, please be aware of this complication, because it may entirely rule out the possibility of doing HPKP with the standard library, or at least complicate matters greatly. @dstufft, this is probably worth being aware of for its effect on pip.

urllib3/util/hpkp.py
```diff
@@ -0,0 +1,336 @@
...
+    cert = x509.load_der_x509_certificate(
+        der_certificate, default_backend()
+    )
+    key = cert.public_key()
+    public_key = key.public_bytes(Encoding.PEM,
+                                  PublicFormat.SubjectPublicKeyInfo)
+    public_key_base64 = ''.join(public_key.split("\n")[1:-2])
+    public_key_raw = base64.b64decode(public_key_base64)
+    public_key_sha256 = hashlib.sha256(public_key_raw).digest()
+    public_key_sha256_base64 = base64.b64encode(public_key_sha256)
```
Lukasa: Next time cryptography releases they'll add `Encoding.DER`, which will allow us to skip this weird trip through PEM (which requires us to do the crazy split-join-b64decode that we do here).

Lukasa:
Getting hold of these callbacks requires work in cryptography and pyopenssl, so this change is now blocked behind those projects.

Lukasa:
Ok, I've managed to reproduce Christian's code (with a bit of work). It requires some more cryptography bindings enhancements, and some PyOpenSSL changes, which makes it kinda a pain in the ass, but we can do it if we just believe.

I'll start pushing these changes either today or tomorrow.

urllib3/util/hpkp.py
alex: This is released now

encodeltd:
Hi, I found this thread while working on sha1/sha256 pubkey digest pinning for https requests for urllib3.

My commits:

https://github.com/encodeltd/urllib3/commit/5f85b6745c2f65f7a4a44b47b9bc93f4184e8eca
https://github.com/encodeltd/urllib3/commit/1f7ed4a8a5e9c278d5c41a0dde0f2d5fa3305ba7

I haven't added proper test to `test/with_dummyserver/test_https.py`, but it's the next thing on my list.

Here's a simple testing script:

```python
import ssl
import logging
import urllib3

from urllib3.contrib import pyopenssl
pyopenssl.inject_into_urllib3()

digest = (
    'IOxdCvvGwOLhlVbFNSs8YHim7ZVVwvqGgkBP21Up060=',
    'JpoZo4gowd1wG6DKLJjbxuFPNz4=',
    'gMxWOrX4PMQesK9qFNbYBxjBfjUvlkn/vN1n+L9lE5E=',
    'R0nfFlf0bIvSjHkbmfufKIEqYOA='
)
test_request = urllib3.PoolManager(assert_pubkey_digests=digest)
test_response = test_request.request('GET', 'https://twitter.com/')
print test_response.status

digest = ('-')
test_request = urllib3.PoolManager(assert_pubkey_digests=digest)
test_response = test_request.request('GET', 'https://twitter.com/')
print test_response.status
```

It this of any use for you?

jpf:
@encodeltd very nice, I like the `assert_pubkey_digests` interface, in a large part because that's very close to the `cert` interface that Requests uses. 

Out of curiosity, I'm wondering why you used *M2Crypto* instead of *cryptography*?

encodeltd:
@jpf thank you! The implementation pretty much follows how `assert_fingerprint` is done.

The reason for using `M2Crypto` is historical. I already have this implementation working with `urllib2` and `M2Crypto`, but there is one massive issue with it, which is that you cannot set timeout on requests. I found it's because of a bug that won't be fixed so I had to look elsewhere. `urrlib3` seems much nicer than `urllib2` and I ported the code to `urllib3`.

I just read your comment above and I see that you have ported similar implementation from `M2Crypto` to `cryptography` so I may follow your lead and do the same in my implementation.

encodeltd:
@jpf I switched the code to use `cryptography` instead of `M2Crypto`:

https://github.com/encodeltd/urllib3/commit/93c7cc49dbe8a6e37a7263f9b60583b5e0e1772b

Lukasa:
So at the moment this has slowed down because I want to extend `cryptography` with an interface that will help clean up some of the mess here.

encodeltd:
@Lukasa any thoughts on my implementation? Shall I make a separate pull request for it?

Lukasa:
> any thoughts on my implementation?

Sure =)

First, it's not HPKP. It's simply not a comparable feature to the one being developed here. It's a *similar* feature, but your implementation has a few notable differences:

1. HPKP only allows SHA256 hashes, your implementation also allows SHA1. For your implementation that's fine, but SHA1 is considered less than ideal at this stage, so if it were me I'd want to remove that anyway.
2. Your implementation is not a Trust On First Use service, it's a manual verification step: the user has to manually provide the hash they want to use. That certainly has utility and might well be a worthy extension to urllib3's existing fingerprint verification, but it's not HPKP. HPKP involves parsing headers and managing a trust store so that users do not need to manually pin certs.
3. You don't actually check every certificate. I bumped into this the hard way, as you can see in a comment above. That's unfortunate, because HPKP allows you to pin the root certificate if you want to: your implementation does not verify that certificate.

I think your code is interesting and potentially useful, but it's not HPKP. Feel free to open a pull request with that in mind, but at some point we'll need to discuss how we want to handle these multiple proposals for key pinning.

Lukasa:
For those keeping an eye on this issue: it's not dead. I've been working with @reaperhulk to come up with a new approach to this that does not rely directly on cryptography itself, which I believe I'm close to achieving. When that's ready, I'll start looking at cleaning this up and preparing it for proper support.

Lukasa:
Specifically, this requires a PyOpenSSL release containing pyca/pyopenssl#382. When that is present, we can do this entirely using PyOpenSSL.

This will limit HPKP support to deployments with PyOpenSSL for the short term, but I'll start a thread on python-ideas to discuss whether we can add the necessary support to the standard library (we need a thread on python-ideas because we have some problems around X.509 certs in the `ssl` module). @tiran, you may want to dive into that thread too.

Lukasa:
Ok, I've updated this again. I've now used this code to successfully validate a HPKP connection to PyPI, which is pretty exciting!

```
INFO:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.python.org
DEBUG:urllib3.util.hpkp:Validating connection to pypi.python.org
DEBUG:urllib3.util.hpkp:No trust information for this host
DEBUG:urllib3.connectionpool:"GET /pypi HTTP/1.1" 200 24007
DEBUG:urllib3.util.hpkp:Searching for PKP header on HTTPS response
DEBUG:urllib3.util.hpkp:Storing host KnownPinnedHost(domain='pypi.python.org', pins=['WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=', '5C8kvU039KouVrl52D0eZSGf4Onjo4Khs8tmyTlV3nU=', '5C8kvU039KouVrl52D0eZSGf4Onjo4Khs8tmyTlV3nU=', 'lCppFqbkrlJ3EcVFAkeip0+44VaoJUymbnOaEUk7tEU=', 'TUDnr0MEoJ3of7+YliBMBVFB4/gJsv5zO7IxD9+YoWI=', 'x4QzPSC810K5/cMjb05Qm4k3Bw5zBn4lTdO/nEW/Td4='], max_age=600, include_subdomains=True, report_uri=None, start_date=1452077905.522401)
INFO:urllib3.connectionpool:Starting new HTTPS connection (1): pypi.python.org
DEBUG:urllib3.util.hpkp:Validating connection to pypi.python.org
DEBUG:urllib3.util.hpkp:Found relevant pin for pypi.python.org
DEBUG:urllib3.util.hpkp:Successfully validated connection to pypi.python.org using HPKP
DEBUG:urllib3.connectionpool:"GET /pypi HTTP/1.1" 200 24007
DEBUG:urllib3.util.hpkp:Searching for PKP header on HTTPS response
DEBUG:urllib3.util.hpkp:Storing host KnownPinnedHost(domain='pypi.python.org', pins=['WoiWRyIOVNa9ihaBciRSC7XHjliYS9VwUGOIud4PB18=', '5C8kvU039KouVrl52D0eZSGf4Onjo4Khs8tmyTlV3nU=', '5C8kvU039KouVrl52D0eZSGf4Onjo4Khs8tmyTlV3nU=', 'lCppFqbkrlJ3EcVFAkeip0+44VaoJUymbnOaEUk7tEU=', 'TUDnr0MEoJ3of7+YliBMBVFB4/gJsv5zO7IxD9+YoWI=', 'x4QzPSC810K5/cMjb05Qm4k3Bw5zBn4lTdO/nEW/Td4='], max_age=600, include_subdomains=True, report_uri=None, start_date=1452077905.72515)
```

There's some work still to do here:

- the HPKP Manager isn't thread-safe but is used by all connections and connection pools at PoolManager scope. We need to lock around it.
- This requires PyOpenSSL, and probably breaks stuff if it's not available. We should fall back to a no-op HPKP implementation if PyOpenSSL isn't being used.
- Obviously I need tests and stuff.
- It'd be good if we shipped a file-based HPKP database either out of the box or as a third-party module.

Regardless, it's exciting that this works!

Lukasa:
Oh, and if you're watching this issue and want to play with it, here's a test file that you can play with:

```python
import certifi
import urllib3
import logging
import time

logging.basicConfig(level=logging.DEBUG)

p = urllib3.PoolManager(cert_reqs='CERT_REQUIRED', ca_certs=certifi.where())
p.request('GET', 'https://pypi.python.org/pypi')
p.clear()
p.request('GET', 'https://pypi.python.org/pypi')
```

shazow:
SUPER COOL! :) Great work, @Lukasa. :custard:

kislyuk:
@Lukasa, are there any news on the status of this and #608? I'm interested in helping out with HPKP/HSTS support. Also, there doesn't seem to be any mention of preload from https://chromium.googlesource.com/chromium/src/+/master/net/http/transport_security_state_static.json.

Lukasa:
@kislyuk This has been put on the back burner as being just not a high enough priority for my time. The previous integration was somewhat working, although it was missing a lot of stuff, but the landscape of urllib3 has change quite a bit and would need some substantial reworking.

As to preload, I would rather ship the preload list as a separate package from urllib3: it would be extremely annoying to get stuck with an out-of-date preload list because you can't upgrade urllib3.

kislyuk:
OK, thanks. Can I help get it merge-ready?

I agree that the preload list should be handled certifi-style, the code to do so is fairly trivial and I have something working. If I ship it, would you consider using it as a dependency?

Lukasa:
You're certainly welcome to try!

I don't think we'd use it as a hard dependency, but certainly as a soft one.

kislyuk:
@Lukasa, could you elaborate on what you think needs changing before this can be merged, aside from your list in https://github.com/shazow/urllib3/pull/607#issuecomment-169297669? You mentioned "the landscape has changed", is there something I should be aware of architecturally? All I see when merging master into this branch is a minor change to how the SSL context gets carried around and a new `_match_hostname` wrapper.

I was hoping to remove the dependency on pyopenssl, but could find no equivalent to `connection.get_peer_cert_chain()` in the standard library. Without that you obviously can't walk the cert chain and examine the pins. So that looks like a hard requirement.

Lukasa:
Since this was last done, I made a *substantial* rewrite of the PyOpenSSL module. That may lead only to pretty minor changes, but I wanted to warn you about it. =)

As to removing PyOpenSSL, that simply cannot happen. The standard library is incapable of giving us the detail we need, because even if it provides APIs to get the peer cert chain that only includes the certificates that the peer *actually sent*. However, HPKP needs to validate the *chain*, which may not include the certificates the peer sent. That's why I have had to use this substantial injection into PyOpenSSL's validation logic in order to get this to work.

kislyuk:
For reference: https://bugs.python.org/issue24107

tiran:
The feature won't land in Python 3.6, maybe in 3.7.

encodeltd:
@Lukasa any further thoughts re static pinning as per my fork of urllib3?

Eg. https://kuziel.nz/notes/2016/08/18/urllib3-fork-ssl-certificate-pinning-sha1-sha256-pubkey-digests.html

You told me at Kiwi PyCon last year that you are not very keen to merge my changes in, so I didn't bother doing a pull request, but to this date I haven't seen anything better in place.

Re HPKP, I find this blog quite insightful https://blog.qualys.com/ssllabs/2016/09/06/is-http-public-key-pinning-dead

Lukasa:
@encodeltd My misgivings now remain as they were then: namely, that static vs dynamic is just not really where the code complexity comes in. The code complexity comes entirely from the fact that you can pin any of the certs in the valid chain. We already have support for pinning the leaf (`assert_fingerprint`), which is the only part we can reliably obtain with minimal code changes. Everything else requires a very substantial code change in pyopenssl to ensure that we can safely extract the *validated* trust chain.

encodeltd:
@Lukasa thank you for the follow up. I do not agree with you, but I respect what you have to say. I shall follow up in the future via a new issue, because my fork/work is not related to HPKP and I do not feel good being constantly off-topic in this thread :-)

Lukasa:
@encodeltd Remind me, please: how does your work differ from HPKP?

jonparrott:

PR #609, opened 2015-04-30
don't allow path in urls which do not begin with /

t-8ch:
It was possible to create such an object, but it would not
serialize -> parse roundtrip.

The alternative would be to reject paths without a leading /


Lukasa:
Seems like an obvious change.

UT failure is unrelated.

Lukasa:
:cake: :+1:

shazow:
Thankiepankie.

Issue #610, opened 2015-05-07
Unit tests fail with exception gaierror


ddriddle:
When I run the unit tests for urllib3 v1.10.2 or the latest version (master) in GitHub I receive the following error:
```
urllib3/test$ nosetests
S.............................E.....................................S..S....................................Exception in thread Thread-11:
Traceback (most recent call last):
  File "/opt/rh/python27/root/usr/lib64/python2.7/threading.py", line 811, in __bootstrap_inner
    self.run()
  File "/services/scratch/ddriddle/src/urllib3/dummyserver/server.py", line 88, in run
    self.server = self._start_server()
  File "/services/scratch/ddriddle/src/urllib3/dummyserver/server.py", line 75, in _start_server
    sock.bind((self.host, 0))
  File "/opt/rh/python27/root/usr/lib64/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
gaierror: [Errno -2] Name or service not known
```
The problem seems to be that my system does not support IPv6 but socket.has_ipv6 returns true.

shazow:
Mmm strange. Any ideas for how to fix this?

ddriddle:
Yes I do. See pull request #611.

ddriddle:
What needs to happen is to catch the gaierror then bind with an IPv4 address. Of course this breaks test_connectionpool.py:601:test_source_address since it also depends on socket.has_ipv6 so it is necessary to communicate to that function that it should not perform IPv6 tests. My solution fixes the problem but their maybe a more elegant solution.

PR #611, opened 2015-05-07
Issue #610 -- Tests fallback to IPv4 if IPv6 fails

ddriddle:
On some systems binding to an IPv6 address will fail with a socket.gaierror exception even though socket.has_ipv6 is set to True. Added code to catch these exceptions and fallback to using IPv4.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -148,6 +148,9 @@ In chronological order:
 * tlynn <https://github.com/tlynn>
   * Respect the warning preferences at import.
 
+* David D. Riddle <ddriddle@illinois.edu>
+  * IPv6 bugfixes in testsuite
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
 
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -38,6 +38,35 @@ DEFAULT_CA = os.path.join(CERTS_PATH, 'cacert.pem')
 DEFAULT_CA_BAD = os.path.join(CERTS_PATH, 'client_bad.pem')
 NO_SAN_CA = os.path.join(CERTS_PATH, 'cacert.no_san.pem')
 
+def _has_ipv6(host):
+    """ Returns True if the system can bind an IPv6 address. """
+    sock = None
+    has_ipv6 = False
+
+    if socket.has_ipv6:
+        # has_ipv6 returns true if cPython was compiled with IPv6 support.
+        # It does not tell us if the system has IPv6 support enabled. To
+        # determine that we must bind to an IPv6 address.
+        # https://github.com/shazow/urllib3/pull/611
+        # https://bugs.python.org/issue658327
+        try:
+            sock = socket.socket(socket.AF_INET6)
+            sock.bind((host, 0))
+            has_ipv6 = True
+        except:
+            pass
+
+    if sock:
+        sock.close()
+    return has_ipv6
+
+# Some systems may have IPv6 support but DNS may not be configured
+# properly. We can not count that localhost will resolve to ::1 on all
+# systems. See https://github.com/shazow/urllib3/pull/611 and
+# https://bugs.python.org/issue18792
+HAS_IPV6_AND_DNS = _has_ipv6('localhost')
+HAS_IPV6 = _has_ipv6('::1')
+
 
 # Different types of servers we have:
 
@@ -64,7 +93,7 @@ class SocketServerThread(threading.Thread):
         self.ready_event = ready_event
 
     def _start_server(self):
-        if socket.has_ipv6:
+        if HAS_IPV6_AND_DNS:
             sock = socket.socket(socket.AF_INET6)
         else:
             warnings.warn("No IPv6 support. Falling back to IPv4.",
@@ -117,7 +146,7 @@ def bind_sockets(port, address=None, family=socket.AF_UNSPEC, backlog=128,
     sockets = []
     if address == "":
         address = None
-    if not socket.has_ipv6 and family == socket.AF_UNSPEC:
+    if not HAS_IPV6 and family == socket.AF_UNSPEC:
         # Python can be compiled with --disable-ipv6, which causes
         # operations on AF_INET6 sockets to fail, but does not
         # automatically exclude those results from getaddrinfo
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -36,7 +36,7 @@ from urllib3.util.timeout import Timeout
 
 import tornado
 from dummyserver.testcase import HTTPDummyServerTestCase
-from dummyserver.server import NoIPv6Warning
+from dummyserver.server import NoIPv6Warning, HAS_IPV6_AND_DNS
 
 from nose.tools import timed
 
@@ -600,7 +600,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
 
     def test_source_address(self):
         for addr, is_ipv6 in VALID_SOURCE_ADDRESSES:
-            if is_ipv6 and not socket.has_ipv6:
+            if is_ipv6 and not HAS_IPV6_AND_DNS:
                 warnings.warn("No IPv6 support: skipping.",
                               NoIPv6Warning)
                 continue
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -1,6 +1,8 @@
 import unittest
 import json
 
+from nose.plugins.skip import SkipTest
+from dummyserver.server import HAS_IPV6
 from dummyserver.testcase import (HTTPDummyServerTestCase,
                                   IPv6HTTPDummyServerTestCase)
 from urllib3.poolmanager import PoolManager
@@ -154,6 +156,9 @@ class TestPoolManager(HTTPDummyServerTestCase):
 
 
 class TestIPv6PoolManager(IPv6HTTPDummyServerTestCase):
+    if not HAS_IPV6:
+        raise SkipTest("IPv6 is not supported on this system.")
+
     def setUp(self):
         self.base_url = 'http://[%s]:%d' % (self.host, self.port)
 


shazow:
@ddriddle Any ideas what conditions are required for this to happen? (ie. has_ipv6 false-positive)

@TomasTomecek Any thoughts on this? (You wrote the ipv6 detection code in our test suite before)

ddriddle:
I am not sure I am receiving a false positive for socket.has_ipv6. I am running on RHEL6. I receive this error with Python 2.6.6 and Python 2.7.5 on the same system.

Lukasa:
@ddriddle Does your system support IPv6 but simply not have the module loaded?

ddriddle:
I talked with my sysadmin and the system supports IPv6. The module is loaded. The problem is that localhost is an unknown host because the IPv6 address for localhost is not listed in /etc/hosts. I don't think it is a good idea to use localhost for IPv6 since even my stock Debian box does not support localhost as an alias for ::1.

If I hardcode ::1 in dummyserver/server.py then the server will start but I receive a bunch of errors from failed unit tests here is the first one.
```
ERROR: test_source_address (test.with_dummyserver.test_connectionpool.TestConnectionPool)
  File "/services/scratch/ddriddle/src/urllib3/test/with_dummyserver/test_connectionpool.py", line 609, in test_source_address
    r = pool.request('GET', '/source_address')
  File "/services/scratch/ddriddle/src/urllib3/urllib3/request.py", line 68, in request 
    **urlopen_kw)
  File "/services/scratch/ddriddle/src/urllib3/urllib3/request.py", line 81, in request_encode_url
    return self.urlopen(method, url, **urlopen_kw)
  File "/services/scratch/ddriddle/src/urllib3/urllib3/connectionpool.py", line 597, in urlopen 
  File "/services/scratch/ddriddle/src/urllib3/urllib3/util/retry.py", line 222, in increment
    raise six.reraise(type(error), error, _stacktrace)
  File "/services/scratch/ddriddle/src/urllib3/urllib3/connectionpool.py", line 544, in urlopen 
    body=body, headers=headers)
  File "/services/scratch/ddriddle/src/urllib3/urllib3/connectionpool.py", line 349, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/opt/rh/python27/root/usr/lib64/python2.7/httplib.py", line 973, in request
    self._send_request(method, url, body, headers)
  File "/opt/rh/python27/root/usr/lib64/python2.7/httplib.py", line 1007, in _send_request
    self.endheaders(body)
  File "/opt/rh/python27/root/usr/lib64/python2.7/httplib.py", line 969, in endheaders
    self._send_output(message_body)
  File "/opt/rh/python27/root/usr/lib64/python2.7/httplib.py", line 829, in _send_output
    self.send(msg)
  File "/opt/rh/python27/root/usr/lib64/python2.7/httplib.py", line 791, in send    self.connect()
  File "/services/scratch/ddriddle/src/urllib3/urllib3/connection.py", line 155, in connect
    conn = self._new_conn()
  File "/services/scratch/ddriddle/src/urllib3/urllib3/connection.py", line 134, in _new_conn
  File "/services/scratch/ddriddle/src/urllib3/urllib3/util/connection.py", line 88, in create_connection
    raise err
ProtocolError: ('Connection aborted.', gaierror(-9, 'Address family for hostname not supported'))
-------------------- >> begin captured logging << --------------------
urllib3.util.retry: DEBUG: Converted retries value: False -> Retry(total=False, connect=None, read=None, redirect=0)
urllib3.connectionpool: INFO: Starting new HTTP connection (1): localhost
--------------------- >> end captured logging << ---------------------
```
I will keep working on this to see if I can figure anything else out. Please tell me what you think.

shazow:
>  I don't think it is a good idea to use localhost for IPv6 since even my stock Debian box does not support localhost as an alias for ::1.

Why is that?

Are you able to supply a different alias for your ::1? Maybe we could override the default hostname with an environment variable for the test suite if there are other people in your situation.

> If I hardcode ::1 in dummyserver/server.py then the server will start but I receive a bunch of errors from failed unit tests here is the first one.

I believe that's because our code is assuming that you're passing in a hostname, rather than an IP, so it attempts to resolve it. @Lukasa @sigmavirus24 I think this is a bug? HTTP should work with an IP as a hostname, no?

sigmavirus24:
> If I hardcode ::1 in dummyserver/server.py then the server will start but I receive a bunch of errors from failed unit tests here is the first one.

I don't think so @shazow. That's because the server is only starting up for IPv6 and if the tests try to use IPv4, then you'll get the gaierror that the address family we're trying to open the socket with is not supported. You would get the same error if you bound using only IPv6 and tried to open another socket connecting to it using IPv4. If the server doesn't support IPv6 or IPv4 and we try to connect using the one it doesn't support, this seems like something that we would expect to see.

> Why is that?

I just spun up a Debian VM and it had 127.0.0.1 and ::1 both in `/etc/hosts`.

Lukasa:
@sigmavirus24 is correct, `/etc/hosts` on a correctly-configured IPv4/6 system will have both 127.0.0.1 and ::1 as aliases for localhost. This will happen on startup *if* your system expects to start up with both IPv4 and IPv6 enabled.

sigmavirus24:
You can even edit it without restarting and it should work. The question should be, why does your sysadmin remove it? 
-- 

ddriddle:
@sigmavirus24 Does the Debian VM have an IPv6 address assigned? My Debian system does not. When I look at my /etc/hosts it has a line with `::1` hashed out. I did not change it. The difference I see is that on my Debian system if run ping6 I receive:
```
$ ping6 ::1
connect: Network is unreachable
```
But on the RHEL6 system I receive
```
$ ping6 ::1
PING ::1(::1) 56 data bytes
64 bytes from ::1: icmp_seq=1 ttl=64 time=0.051 ms
64 bytes from ::1: icmp_seq=2 ttl=64 time=0.067 ms
^C
--- ::1 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1452ms
rtt min/avg/max/mdev = 0.051/0.059/0.067/0.008 ms
```
I ran the urllib3 unit tests on my Debian system and I received the same errors as on the RHEL6 system:
```
$ nosetests
S.............................E.....................................S..S.............E......................Exception in thread Thread-10:
  File "/usr/lib/python2.7/threading.py", line 552, in __bootstrap_inner
    self.run()
  File "/home/ddriddle/src/urllib3/dummyserver/server.py", line 88, in run
    self.server = self._start_server()
  File "/home/ddriddle/src/urllib3/dummyserver/server.py", line 75, in _start_server
    sock.bind((self.host, 0))
  File "/usr/lib/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
gaierror: [Errno -2] Name or service not known

^C
```
Note that on the Debian system `socket.has_ipv6` is set to true,
```
$ python
Python 2.7.3 (default, Mar 14 2014, 11:57:14) 
[GCC 4.7.2] on linux2
>>> import socket
>>> socket.has_ipv6
True
```
even though ifconfig clearly shows I have no IPv6 address:
```
$ ifconfig 
eth0      Link encap:Ethernet  HWaddr f0:4d:a2:dd:a6:ca  
          inet addr:192.17.24.166  Bcast:192.17.27.255  Mask:255.255.252.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:110416728 errors:0 dropped:0 overruns:0 frame:0
          TX packets:11899882 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:2729123451 (2.5 GiB)  TX bytes:2248881360 (2.0 GiB)
          Interrupt:41 Base address:0xa000 

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          UP LOOPBACK RUNNING  MTU:16436  Metric:1
          RX packets:290421 errors:0 dropped:0 overruns:0 frame:0
          TX packets:290421 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:882914165 (842.0 MiB)  TX bytes:882914165 (842.0 MiB)
```
I went into my `/etc/hosts` file and uncommented out the lines related to IPv6. This changes the error messages but of course they still fail because IPv6 is not enabled on my Debian system.
```
$ nosetests
SEEEEEEEEEEException in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 552, in __bootstrap_inner
    self.run()
  File "/home/ddriddle/src/urllib3/dummyserver/server.py", line 88, in run
    self.server = self._start_server()
  File "/home/ddriddle/src/urllib3/dummyserver/server.py", line 75, in _start_server
    sock.bind((self.host, 0))
  File "/usr/lib/python2.7/socket.py", line 224, in meth
    return getattr(self._sock,name)(*args)
error: [Errno 99] Cannot assign requested address
```
@sigmavirus24 Can you retry your VM test but make sure the Debian system does not have an IPv6 address? Either way it is clear that `socket.has_ipv6` is not reliable.

Lukasa:
As to the reliability of socket.has_ipv6, the best thing to do might be to try to create an IPv6 socket and see if it fails.

ddriddle:
I looked at the [documentation](https://docs.python.org/3.5/library/socket.html#socket.has_ipv6) for `has_ipv6` for cPython (dev) and it states:
```
socket.has_ipv6

    This constant contains a boolean value which indicates if IPv6 is supported on this platform.
```
This is a bit cryptic but I would assume by that it means that IPv6 is supported by the system and the cPython compiler but looking at the C code it appears that is not the case. Take a look at (`cpython/Modules/socketmodule.c`):
```c
#ifdef ENABLE_IPV6
    has_ipv6 = Py_True;
#else
    has_ipv6 = Py_False;
#endif
    Py_INCREF(has_ipv6);
    PyModule_AddObject(m, "has_ipv6", has_ipv6);
```
And the headers contain the following relevant code (`cpython/Modules/socketmodule.h`) : 
```c
/* VC6 is shipped with old platform headers, and does not have MSTcpIP.h
 * Separate SDKs have all the functions we want, but older ones don't have
 * any version information.
 * I use SIO_GET_MULTICAST_FILTER to detect a decent SDK.
 */
# ifdef SIO_GET_MULTICAST_FILTER
#  include <MSTcpIP.h> /* for SIO_RCVALL */
#  define HAVE_ADDRINFO
#  define HAVE_SOCKADDR_STORAGE
#  define HAVE_GETADDRINFO
#  define HAVE_GETNAMEINFO
#  define ENABLE_IPV6
# else
```
Looking at the headers and the C code it looks to me that `has_ipv6` is set to True or False at compile time. I see no code in cPython that performs a dynamic check to set `has_ipv6`. What I did find in the tests is the following code (cpython/Lib/test/support/__init__.py):
```c
def _is_ipv6_enabled():
    """Check whether IPv6 is enabled on this host."""
    if socket.has_ipv6:
        sock = None
        try:
            sock = socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
            sock.bind((HOSTv6, 0))
            return True
        except OSError:
            pass
        finally:
            if sock:
                sock.close()
    return False

IPV6_ENABLED = _is_ipv6_enabled()
```
I also found the following bit of information in cPython's docs (`cpython/Misc/HISTORY`):

> There is also a new module variable, has_ipv6, which is True iff the current Python has IPv6 support. See SF patch [#658327](https://bugs.python.org/issue658327).

Looking at the ticket it reads to me that `has_ipv6` is set if cPython was compiled with IPv6 support it does not tell you if the system supports IPv6.

ddriddle:
@Lukasa completely agree. That is what this patches does.

ddriddle:
I made changes to the patch and added comments now that I better understand the problem. When you have a chance please review it and tell me if it needs any further improvements.

shazow:
Thanks @ddriddle!

PR #613, opened 2015-05-09
Handle EPIPE on shutdown from pyopenssl

### LINKED ISSUE #612: Uncaught exception from pyOpenSSL's shutdown call
See https://github.com/pypa/pip/issues/2753 for more details

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13757454-uncaught-exception-from-pyopenssl-s-shutdown-call?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
This should resolve #612 and pypa/pip#2753.

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -54,6 +54,7 @@ import OpenSSL.SSL
 from pyasn1.codec.der import decoder as der_decoder
 from pyasn1.type import univ, constraint
 from socket import _fileobject, timeout
+import errno
 import ssl
 import select
 
@@ -210,7 +211,14 @@ class WrappedSocket(object):
 
     def close(self):
         if self._makefile_refs < 1:
-            return self.connection.shutdown()
+            # Shutdown may need to write data to the socket: if the remote end
+            # has already closed it, this will fail with EPIPE. That's ok,
+            # catch it.
+            try:
+                return self.connection.shutdown()
+            except OpenSSL.SSL.SysCallError as e:
+                if e.args[0] != errno.EPIPE:
+                    raise
         else:
             self._makefile_refs -= 1
 


t-8ch:
Maybe we should not confuse `close()` and `shutdown()`. The stdlib also has a dedicated shutdown. What about. `close = _drop`?

Lukasa:
@t-8ch Not sure I fully understand your point. Right now `_drop` is implemented in terms of `close`, which is implemented in terms of `shutdown`.

t-8ch:
Hm, should have woken up properly first...
I mean, the stdlib does not use `shutdown` on `close`, so our compat wrapper should also not do so.
This means, we don't shut down properly which should be addresses in a separate issue, also covering our non-pyopenssl codepath

Lukasa:
@t-8ch I don't see any `close` logic in pyopenssl, so I take it your plan is that we just directly close the socket and let the `OpenSSL.SSL.Connection` object go out of scope?

t-8ch:
https://pythonhosted.org/pyOpenSSL/api/ssl.html#OpenSSL.SSL.Connection.close

Lukasa:
Ah, I see, sorry: the `OpenSSL.SSL.Connection` object is a proxy to the socket, so it still has socket methods on it.

So, the reason not to do that is the reason we're using `shutdown` in the first place, which is that we do not cleanly shut the connection down: it's possible that OpenSSL has data it wants to write to the socket that it has not been able to write.

I take it that we therefore have similar 'bugs' with the stdlib, that's your assessment?

t-8ch:
@Lukasa exactly

Lukasa:
So let's ask the follow-up question: do we care? The stdlib currently manages closing the socket in the `HTTPConnection` class, and doesn't bother calling `shutdown`. Should we just be calling `close` and being done with it?

t-8ch:
Weirdly enough it does do a shutdown in the `HTTPSConnection` if a handshake fails.
I would say: go for plain close and try to cook up a testcase for the bug.

We should also provider a `shutdown()` method for consistency (and maybe later use)

Lukasa:
Alright, that feels like a good patch. I feel a bit bad if I rewrite this code to do that (because it's totally your idea), so it might be better if you push up that patch and I'll review it. =)

Lukasa:
Closing in favour of #614.

shazow:
Thank you both. :)

smtakeda:
Hi, I came across this issue while I was investigating EPIPE handling. I just found OpenSSL/SSL.py no longer has `Connection.close` method in the code on Github.
https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/SSL.py

I'm not sure what change has happened from this repository: https://pythonhosted.org/pyOpenSSL/api/ssl.html#OpenSSL.SSL.Connection.close

but the doc still says `Connection.close` exists:
https://pyopenssl.readthedocs.org/en/latest/api/ssl.html#OpenSSL.SSL.Connection.close

Maybe I should raise this PyOpenSSL issue tracker or am I missing something? If so, calling `Connection.close` won't work then?

Lukasa:
@smtakeda It's fine. =) The `Connection.close` method still exists, but it exists as a proxy through to the underlying `socket` object. The logic for it is [here](https://github.com/pyca/pyopenssl/blob/master/src/OpenSSL/SSL.py#L1123-L1133).

The code will function correctly. Thanks for your concern!

smtakeda:
oh, sorry about my ignorance and thanks for the quick reply. I'm glad that I leaned something new in Python and I can apply the same patch to my copy of urllib3. Thank you.

Lukasa:
> oh, sorry about my ignorance

There is never any need to apologise for not knowing something. We all start out not knowing, and then we learn. You were polite and considerate, which is all you need to be. Don't worry about not knowing things: you're learning!

(If you want to ask more questions, you can join us in IRC at #python-requests.)

PR #619, opened 2015-05-12
refactor assert_fingerprint for efficiency + style

jab:


urllib3/util/ssl_.py
```diff
@@ -8,6 +8,13 @@
 HAS_SNI = False
 create_default_context = None
 
+# Maps the length of a digest to a possible hash function producing this digest
+HASHFUNC_MAP = {
+    16: md5,
+    20: sha1,
+    32: sha256,
+}
+
```
jab: This is used as a constant, so capitalized and moved from being defined inside the function to module scope.

urllib3/util/ssl_.py
```diff
@@ -112,31 +119,22 @@ def assert_fingerprint(cert, fingerprint):
...
+    hashfunc = HASHFUNC_MAP[digest_length]
 
     cert_digest = hashfunc(cert).digest()
 
-    if not cert_digest == fingerprint_bytes:
+    if cert_digest != fingerprint_bytes:
         raise SSLError('Fingerprints did not match. Expected "{0}", got "{1}".'
-                       .format(hexlify(fingerprint_bytes),
-                               hexlify(cert_digest)))
+                       .format(fingerprint, hexlify(cert_digest)))
```
jab: After `fingerprint = fingerprint.replace(':', '').lower()` above, `fingerprint == hexlify(fingerprint_bytes)` (given `fingerprint_bytes = unhexlify(fingerprint.encode())`), so no need to re-hexlify.

urllib3/util/ssl_.py
shazow: It wasn't pulled out as a module-constant is because it's not shared with any other code. The only reason it was defined in a named dict is for self-documenting purposes and avoiding unnamed magic numbers inline. :)

We can keep this change if you prefer it, I have no strong feelings either way.

shazow:
Thanks for the cleanups, very appreciated! :)

jab:
Thanks for the quick review and merge! And thanks for the great work on urllib3 :)

shazow:
:sparkles: :heart: :sparkles:

Issue #618, opened 2015-05-12
Fix AppEngine support


shazow:
We claim urllib3 works with AppEngine but looks like it's broken today (as of a few months ago). Nobody on our core team uses AppEngine, so maintaining support for it has been a substantial drain.

Ideally, we'd like official support/resources from the Google Cloud/AppEngine team to maintain support for AppEngine.

Related issues:
- https://github.com/shazow/urllib3/issues/583 (urlfetch is not used when it should be)
- https://github.com/kennethreitz/requests/issues/2595 (chunked-encoding streaming is broken)

t-8ch:
Fun fact: The python appengine SDK even ships with requests/urllib3.

t-8ch:
So the first problem with kennethreitz/requests#2595 is:

In `requests/models.py` we have:
```python
try:
      # Special case for urllib3.
      try:
          for chunk in self.raw.stream(chunk_size, decode_content=True):
              yield chunk
      except ProtocolError as e:
          raise ChunkedEncodingError(e)
      except DecodeError as e:
          raise ContentDecodingError(e)
      except ReadTimeoutError as e:
          raise ConnectionError(e)
except AttributeError as e:
      # Standard file-like object.
      while True:
          chunk = self.raw.read(chunk_size)
          if not chunk:
              break
          yield chunk

```

Deeper in the stack, in `urllib3/response.py:HTTPResponse.read_chunked` we
execute:

```py
self._original_response._method.upper() == 'HEAD'
```

On Appengine `httplib.HTTPResponse` is an `int` which does not have an `upper`
method. We get an `AttributeError` which bubbles up and makes requests fall back
to the nondecompressing branch.

So the error is with requests und urllib3.
Requests should do a better check than just waiting for `AttributeError`s and
urllib3 should not access private fields of other classes. (There was even a
FIXME)

But no mention of Appengine without it being broken: Even after fixing this we
run into a different chunked encoding exception. Fun!

agfor:
@t-8ch I've been looking into it as well when I have time. I tracked down the second exception you mention. The problem is that the GAE httplib response object, stored as `_fp` on the response object, has already been un-chunked, despite GAE passing through the chunked header. So when urllib3 tries to determine the chunk length, it gets a line of the response rather than a line giving the chunk length.

It seems like a GAE bug, but one that urllib3 (or, alternatively, requests) should be able to work around.

t-8ch:
[...]

PR #620, opened 2015-05-12
[WIP] partially unbreak chunked gzip responses on appengine

t-8ch:
see #618

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -439,7 +439,7 @@ class HTTPResponse(io.IOBase):
                 "Header 'transfer-encoding: chunked' is missing.")
 
         # Don't bother reading the body of a HEAD request.
-        if is_response_to_head(self._original_response):
+        if self._original_response and is_response_to_head(self._original_response):
             self._original_response.close()
             return
 
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -432,8 +432,7 @@ class HTTPResponse(io.IOBase):
         method = response._method
         if isinstance(method, int):  # Platform-specific: Appengine
             return method == 3
-        else:
-            return method.upper() == 'HEAD'
+        return method.upper() == 'HEAD'
 
     def read_chunked(self, amt=None, decode_content=None):
         """
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -423,6 +423,18 @@ class HTTPResponse(io.IOBase):
             self.chunk_left = None
         return returned_chunk
 
+    @staticmethod
+    def _is_response_to_head(self, response):
+        # FIXME: Can we do this somehow without accessing private httplib _method?[m
+        if response is None:
+            return False
+
+        method = response._method
+        if isinstance(method, int):  # Platform-specific: Appengine
+            return method == 3
+        else:
+            return method.upper() == 'HEAD'
+
     def read_chunked(self, amt=None, decode_content=None):
         """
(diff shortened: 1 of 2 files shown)


shazow:
Added WIP, ping when it's ready for review. :)

urllib3/response.py
```diff
@@ -422,6 +422,18 @@ def _handle_chunk(self, amt):
...
         return returned_chunk
 
+    @staticmethod
+    def _is_response_to_head(response):
+        # FIXME: Can we do this somehow without accessing private httplib _method?
+        if response is None:
+            return False
+
+        method = response._method
+        if isinstance(method, int):  # Platform-specific: Appengine
```
Lukasa: Wait, what?

t-8ch: Don't blame the messenger :-)

urllib3/response.py
```diff
@@ -437,7 +449,7 @@ def read_chunked(self, amt=None, decode_content=None):
             raise ResponseNotChunked("Response is not chunked. "
                 "Header 'transfer-encoding: chunked' is missing.")
 
-        if self._original_response and self._original_response._method.upper() == 'HEAD':
+        if self._is_response_to_head(self._original_response):
             # Don't bother reading the body of a HEAD request.
             # FIXME: Can we do this somehow without accessing private httplib _method?
```
shazow: I think this FIXME is obsolete here, since it got moved into the helper?

urllib3/response.py
```diff
@@ -422,6 +422,18 @@ def _handle_chunk(self, amt):
             self.chunk_left = None
         return returned_chunk
 
+    @staticmethod
```
shazow: Why staticmethod? Seems to make sense for it to be a regular instance method and skip passing response in.

urllib3/response.py
```diff
@@ -422,6 +422,18 @@ def _handle_chunk(self, amt):
...
+    def _is_response_to_head(response):
+        # FIXME: Can we do this somehow without accessing private httplib _method?
+        if response is None:
+            return False
+
+        method = response._method
+        if isinstance(method, int):  # Platform-specific: Appengine
+            return method == 3
+        else:
```
shazow: Style nit: Can we lose the unnecessary `else:`?

urllib3/response.py
t-8ch: I thought about moving it to utils/connection or so, where the other platform specific stuff lives.

shazow: +0.5 to that. I was thinking the same.

agfor:
I do think this is correct, it just doesn't fix things all the way because the data coming back from GAE httplib is still not chunked.

sigmavirus24:
This needs to be rebased. @t-8ch would you like me to do that?

t-8ch:
Sorry for forgetting this...
@sigmavirus24 This would be awesome, it may take till tomorrow for me to get at my PC

sigmavirus24:
@t-8ch I'll take a stab in a couple hours. :custard:

shazow:
:sparkles: :custard: :sparkles:

sigmavirus24:
Also @t-8ch please don't apologize for forgetting this. You did the hard work of figuring out what int GAE returns for a response to a HEAD request. That would have blocked me.

t-8ch:
@sigmavirus24 @shazow I did the rebase, it was actually trivial :-)

shazow:
Thanks @t-8ch! :custard:

PR #625, opened 2015-05-16
refactor assert_fingerprint further for efficiency and style

jab:
It occurred to me that the code in `assert_fingerprint` could be even simpler. Sorry I didn't realize this when I created #619.

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -10,9 +10,9 @@ create_default_context = None
 
 # Maps the length of a digest to a possible hash function producing this digest
 HASHFUNC_MAP = {
-    16: md5,
-    20: sha1,
-    32: sha256,
+    32: md5,
+    40: sha1,
+    64: sha256,
 }
 
 import errno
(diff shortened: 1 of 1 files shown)


urllib3/util/ssl_.py
```diff
@@ -10,9 +10,9 @@
 
 # Maps the length of a digest to a possible hash function producing this digest
 HASHFUNC_MAP = {
-    16: md5,
-    20: sha1,
-    32: sha256,
+    32: md5,
+    40: sha1,
+    64: sha256,
```
jab: make the comment above this actually true ;P

urllib3/util/ssl_.py
```diff
@@ -120,16 +120,14 @@ def assert_fingerprint(cert, fingerprint):
     """
 
     fingerprint = fingerprint.replace(':', '').lower()
-    digest_length, odd = divmod(len(fingerprint), 2)
```
shazow: Never noticed how bizarre this line is, why did we even do this?

shazow:
>  Sorry I didn't realize this when I created #619.

Not a problem at all, iterative improvements are always very appreciated. :)

urllib3/util/ssl_.py
jab: I was wondering that myself. Let me know if you figure it out :)

jab:
Thanks again for the amazingly fast review!

shazow:
@jab Btw, you should add yourself to our CONTRIBUTORS.txt file!

urllib3/util/ssl_.py
shazow: @sigmavirus24 Can you double check that we didn't ruin something important somehow?

jab:
Aw, since this is already merged I'll just wait till my next PR, but thanks for the kind invite! Always such a pleasure when the new contributor experience is as welcoming as you've been making it ;)

shazow:
<3 THE FEELING IS MUTUAL!

jab:
::OSS WARM FUZZIES::
(sorry for the OT but) just came across https://github.com/shazow/unstdlib.py and had to smile -- I used to wonder whether https://github.com/jab/bidict was [missing from stdlib](https://bidict.readthedocs.org/en/master/cut.html#missing-bidicts-in-stdlib) (but @raymondh quickly disabused me of that notion)

shazow:
:D

Issue #412, opened 2014-06-23
gevent ssl and WantWrite error


vitek:
Hi!

Using `gevent`  I got `OpenSSL.SSL.WantWriteError` error.  Here is simple script to reproduce the issue:

``` python
from gevent import monkey
monkey.patch_all()

import requests

payload = 'x' * 1024 * 1000
requests.get('https://example.org/', data=payload, verify=False)
```

Here is possible fix:

``` python
# urllib3/contrib/pyopenssl.py:
class WrappedSocket(object):
....
    def send_until_done(self, data):
        while True:
            try:
                return self.connection.send(data)
            except OpenSSL.SSL.WantWriteError:
                _, wlist, _ = select.select([], [self.socket], [],
                                            self.socket.gettimeout())
                if not wlist:
                    raise
                continue

    def sendall(self, data):
        while len(data):
            sent = self.send_until_done(data)
            data = data[sent:]
```

PR #627, opened 2015-05-16
Fix performance of sendall() on large files

### LINKED ISSUE #626: pyopenssl performance hit from issue#412
The fix for issue #412 (25755abf27790118cbe1df9bd3b967085c039e9d) introduced a significant performance hit to large file uploads.

In my benchmarks testing with a 50 MB file:
* curl can upload the file in 0.5s (just for context)
* requests + urllib3 using the current implementation takes over 20 seconds (20-22s)
* requests + urllib3 with that fix reverted (`sendall` instead invokes `self.connection.sendall(data)`) uploads the file in 1.15s

The cost appears to be a result of a combination of two factors:
* OpenSSL will only write up to 16384 bytes at a time (the limit derives from the maximum amount of plaintext in a single SSL record)
* sendall() makes a copy of `data` (removing what was already sent), which is copying the entire payload

So with a 50 MB file, urllib3 is performing over 3000 allocations with an average size around 25 MB.  It gets worse when I try uploading even larger files...

Something like the following restores the performance.  There are still a large number of string copies, but they're of a limited (fixed) size.  I should add that I thought about using a memoryview here instead of the slices, but pyOpenSSL immediately turns around and casts back to bytes, so I didn't see a need for the complexity.

```python
    def sendall(self, data):
        total_sent = 0
        while total_sent < len(data):
            sent = self._send_until_done(data[total_sent:total_sent+16384])
            total_sent += sent

```

With this implementation, I'm again seeing uploads complete in about 1.15s.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15321819-pyopenssl-performance-hit-from-issue-412?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

darkrain42:
Fixes the significant performance issues I'm seeing as a result of the fix for #412 

Fixed #626

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -204,9 +204,11 @@ class WrappedSocket(object):
                 continue
 
     def sendall(self, data):
-        while len(data):
-            sent = self._send_until_done(data)
-            data = data[sent:]
+        total_sent = 0
+        while total_sent < len(data):
+            # OpenSSL will only write 16K at once
+            sent = self._send_until_done(data[total_sent:total_sent+16384])
+            total_sent += sent
 
     def shutdown(self):
         # FIXME rethrow compatible exceptions should we ever use this


shazow:
Thanks for putting this together so quickly! :)

urllib3/contrib/pyopenssl.py
```diff
@@ -204,9 +204,11 @@ def _send_until_done(self, data):
                 continue
 
     def sendall(self, data):
-        while len(data):
-            sent = self._send_until_done(data)
-            data = data[sent:]
+        total_sent = 0
+        while total_sent < len(data):
+            # OpenSSL will only write 16K at once
+            sent = self._send_until_done(data[total_sent:total_sent+16384])
```
sigmavirus24: Would have been nice for 16384 to not be a magic number, e.g., make it a constant that's named so we ave context for the future. (The comment alleviates this, but it's nice to have the comment and the constant in my opinion. And it's irrelevant if @shazow disagrees ;).)

shazow: +1 that would be better, though I generally accept good comments as substitutes for named vars. :)

PR #628, opened 2015-05-18
Further improvements to pyopenssl sendall()

darkrain42:
Trying to address the feedback from #626 and #627
* no semi-magic number
* use a memoryview, even if PyOpenSSL will still perform a conversion to string (to make cffi happy) today.  Futureproofing!

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -85,6 +85,8 @@ _openssl_verify = {
 
 DEFAULT_SSL_CIPHER_LIST = util.ssl_.DEFAULT_CIPHERS
 
+# OpenSSL will only write 16K at a time
+SSL_WRITE_BLOCKSIZE = 16384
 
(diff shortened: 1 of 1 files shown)


urllib3/contrib/pyopenssl.py
```diff
@@ -204,10 +207,12 @@ def _send_until_done(self, data):
                 continue
 
     def sendall(self, data):
+        if sys.version_info >= (2, 7) and not isinstance(data, memoryview):
```
shazow: Would be better to check for the presence of `memoryview` rather than check `sys.version_info`.

Maybe something like this near the top?

```python
try:
    _ = memoryview
    has_memoryview = True
except NameError:
    has_memoryview = False

shazow:
Thanks for the improvements, I appreciate you following up on this. :) Btw you're welcome to add yourself to our CONTRIBUTORS.txt file!

PR #631, opened 2015-05-23
Rough stab at App Engine environment tests.

jonparrott:
* Added test/appengine, containing some code to step the sandbox version of httplib/socket/urllib
* Added test_urlfetch.py, checking the URLFetch is used when making http and https requests (not sure if this is the desired behavior at the moment).
* Added GAE environment to tox, currently requires you to set your PYTHONPATH=~/google-cloud-sdk/platform/google_appengine. I can configure travis to do this automatically.

Travis will definitely barf on this. To run tests locally, just install the cloud sdk (cloud.google.com/sdk) and set the python path as above, then run tox -e gae.

I can add more tests to cover situations where sockets are enabled, but I just wanted to get something basic out here to open the dialog.


jonparrott:
Seems ``make test`` just runs ``nosetests`` directly instead of using tox. Is there any particular reason for that?

shazow:
> Travis will definitely barf on this. To run tests locally, just install the cloud sdk (cloud.google.com/sdk) and set the python path as above, then run tox -e gae.

Is there a way to make this work on travis too? (I guess install the cloud sdk as a dependency for one of the test targets?)

> Seems `make test` just runs `nosetests` directly instead of using tox. Is there any particular reason for that?

`make test-all` runs tox, which takes a fairly long time to run so it's annoying to do repeatedly during development. :) Otherwise tests run in just a couple of seconds.

test/appengine/test_urlfetch.py
```diff
@@ -0,0 +1,51 @@
+from . import MockResponse, activate_sandbox
+activate_sandbox()
```
shazow: Maybe better to do this on suite setup and skip if the environment is not available?

jonparrott: Let me try, the sandbox is really designed to be activated before anything is imported at all, so that could cause issues.

jonparrott:
> make test-all runs tox, which takes a fairly long time to run so it's annoying to do repeatedly during development. :) Otherwise tests run in just a couple of seconds.

So does travis run ``make test`` or ``make test-all`` ? The latter would be better, as it'll not break every other environement. I can update the make test command to skip app engine tests either way.

test/appengine/test_urlfetch.py
shazow: Hmm, if that's the case then we could have it only import this and not-skip tests iff some environment variable is present.

shazow:
Travis is configured using [.travis.yml](https://github.com/shazow/urllib3/blob/master/.travis.yml). It has its own notion of environments, so tox's configuration is redundant there. I think there's a way to do tox with travisci but I'm not sure, I never really dug too deeply there. @t-8ch might know?

jonparrott:
Cool, we can work out the travis stuff in a bit. I want to make sure I've covering the behavior you guys are expecting.

So far, I have tests that:
* Confirm that URLFetch is used with HTTP
* Confirm that URLFetch is used with HTTPS (this one is currently failing).

Do I need to add tests that confirm that sockets are used? What situations would those be used?

shazow:
> Do I need to add tests that confirm that sockets are used? What situations would those be used?

That's not a source of support requests for us right now so we can put that off I think.

jonparrott:
Alright, I re-worked things a little bit:
* New make targets test-gae and test-travis.
* test-travis has a check for python 2.7, and if it's there, will run the app engine tests.
* Because of the way nosegae works, it's better to isolate its usage to just the GAE tests.
* The GAE tests raise skiptest if GAE SDK or NoseGAE is not available.
* The GAE test will only run on python 2.7.
* Updated travis.yaml to download and install the cloud sdk when running py2.7

Problems I don't know how to fix right now:
* Tests are erring because ``No handlers could be found for logger "nose.plugins.cover"``, not sure what that is about.
* Some unrelated tests are now failing (rounding errors and a pickling error). I need to make sure this isn't due to the sandbox.

@shazow let me know your thoughts.

jonparrott:
Other environments failing was due to a sandbox issue. Should be fixed now. The only remaining issue is the ``No handlers could be found for logger "nose.plugins.cover"`` error that's nerfing the py2.7 tests.

shazow:
> * Tests are erring because ``No handlers could be found for logger "nose.plugins.cover"``, not sure what that is about.

Probably because `coverage` is not installed. `pip install dev-requirements.txt` should do the trick, unless `coverage` does not work with NoseGAE? Also that should be a warning, not a failure.

Regarding the tests arrangement, ideally I'd love to have py27+gae be its own environment on travisci similar to how you set it up on tox. Not sure how that works exactly.

@Lukasa @sigmavirus24 @t-8ch Thoughts?

jonparrott:
Yeah I'm not sure how to setup ``py27+gae`` as a separate travis environment, but that would definitely be ideal.

Coverage works with NoseGAE, and I'm not changing anything around dev-requirements, so I'm unsure what it going on there.

shazow:
Re tox+travisci, looks like we could setup an ENV matrix. Not sure if this is a terrible idea. Some references:

http://jsatt.com/blog/using-tox-with-travis-ci-to-test-django-apps/
https://github.com/eventlet/eventlet/blob/master/.travis.yml
https://github.com/travis-ci/travis-ci/issues/3024

shazow:
> Coverage works with NoseGAE, and I'm not changing anything around dev-requirements, so I'm unsure what it going on there.

Oh I see what it is. We require 100% coverage for tests to pass. Maybe GAE tests should be exempt since they're only testing a small subset.

jonparrott:
@shazow alright, that'll require some surgery as you guys use setup.cfg to force coverage on every run of nosetests. I can delegate all that out to tox, though.

shazow:
@jonparrott Right right, could pass an override flag to the nose command for the GAE environment scenario.

jonparrott:
@shazow just added the build matrix for tox - see [here](https://travis-ci.org/shazow/urllib3/builds/64133000). I'll get the gae environment exempt from coverage next.

shazow:
@jonparrott :+1: Thank you for doing this, very helpful. :)

jonparrott:
Okay, the gae tests are now exempt from coverage and travis should turn green.

The https test is being skipped because it currently fails, I suppose that's now the test to fix? (I'm still unsure what is causing trouble for you guys on App Engine). I can provide code to help detect the sandboxed GAE environment so you guys can make a special case for it.

Let me know what else I can do to help at this point.

shazow:
Excellent, thanks. See this issue for references to other failures: https://github.com/shazow/urllib3/issues/618

Off the top of my head, chunked encoding requests break in AppEngine, as well as HTTPS when sockets aren't enabled. If you can run the full suite in the GAE sandbox, you should get lots of failures. 

Want me to merge this PR so far and continue the fixes in a new PR?

jonparrott:
Hold off on merging, I broke something with the environment variables with the build matrix. Let me fix that first. 

I think it'd be good to go ahead and merge this so the fixes can be done separately, as I don't think I'll be able to personally do the fixes in a timely manner. I can absolutely help and review though.

shazow:
> Hold off on merging, I broke something with the environment variables with the build matrix. Let me fix that first.

Sounds good, let me know when it's ready for final review/merge.

> I think it'd be good to go ahead and merge this so the fixes can be done separately, as I don't think I'll be able to personally do the fixes in a timely manner. I can absolutely help and review though.

Understood, though none of our contributors are using AppEngine at the moment, so these issues are unlikely to get fixed without support from your team. :)

jonparrott:
Travis is (finally) green. Yay!

In terms of helping out with App Engine, I would need some direction on how you guys would like this structured. Should App Engine has its own Connection and HTTPSConnection classes that use URLFetch directly? With sockets enabled, should they use the normal stuff?

shazow:
> In terms of helping out with App Engine, I would need some direction on how you guys would like this structured.

Happy to help with that. :)

> Should App Engine has its own Connection and HTTPSConnection classes that use URLFetch directly? 

Off the top of my head, I believe only one of those will be necessary. Since URLFetch handles its own connection pooling on the backed (is this a correct assumption?) then we can just have an AppEngineConnectionPool class which straight-up calls URLFetch without any pooling.

At that point, we'll need to add an API for PoolManager to tell it which type of ConnectionPool to use. (This might be an opportunity to add adapter-like things, @sigmavirus24)

> With sockets enabled, should they use the normal stuff?

I believe so? I'm not sure how much is different in AppEngine with sockets enabled. Will need some guidance from your team for that.

.travis.yml
```diff
@@ -1,17 +1,35 @@
...
-  - "3.4"
-  - "pypy"
-script: make test
+script: tox
 before_install:
   - openssl version
+  # Install Google Cloud Platform SDK to enable App Engine tests
+  - if [[ $TOXENV == 'gae' && ! -d $HOME/gcloud/google-cloud-sdk ]]; then
+      mkdir -p $HOME/gcloud &&
+      wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=$HOME/gcloud &&
```
sigmavirus24: Can this not be pip installed? If it can be, there's an easier way to do this.

.travis.yml
```diff
@@ -1,17 +1,35 @@
 language: python
-python:
-  - "2.6"
-  - "2.7"
-  - "3.2"
-  - "3.3"
-  - "3.4"
-  - "pypy"
-script: make test
+script: tox
```
sigmavirus24: In general, I really really like this

shazow: Agree, assuming there are no hidden regrets, this will make our local tox runs much more symmetric with travisci. Yay.

dev-requirements.txt
```diff
@@ -1,4 +1,6 @@
 nose==1.3.3
+NoseGAE==0.5.7
```
sigmavirus24: So, can we move this into the tox.ini section for gae and make `test-gae` do

```
    tox -e gae
```

Or maybe,

```
    TOXENV=gae tox
```

test/test_no_ssl.py
```diff
@@ -71,7 +71,7 @@ def setUp(self):
         sys.meta_path.insert(0, ssl_blocker)
 
     def tearDown(self):
-        assert sys.meta_path.pop(0) == ssl_blocker
+        sys.meta_path.remove(ssl_blocker)
```
sigmavirus24: If we always insert it, why do we need to switch this to "remove"? Does the gcloud sdk install stuff on the `meta_path`?

tox.ini
```diff
@@ -1,10 +1,21 @@
...
-   nosetests \
-   []
+    nosetests []
 setenv =
     PYTHONWARNINGS=always::DeprecationWarning
+
+[testenv:gae]
+basepython = python2.7
+setenv=
+    NOSE_IGNORE_CONFIG_FILES = TRUE
```
shazow: Hm maybe it'd be better to pass a `--cover-min-percentage=0` flag to `nosetests` rather than throwing away all the configs?

jonparrott:
Sounds good. I think having a concrete bug to work on will help me allocate time. Can you create a bug for adding an AppEngineConnectionPool implementation using just URLFetch and assign it to me? No promises on timeline as I'll be doing this in my spare time.

tox.ini
jonparrott: Let's circle back to that after we establish the connection pool - so there will actually be something to cover.

test/test_no_ssl.py
jonparrott: Yes, activate_sandbox muddles with the meta_path as well.

tox.ini
shazow: I mean in general, it'd be good to retain the rest of the settings we don't want to override.

.travis.yml
jonparrott: Unfortunately no, this is the same method we're using in our [own tests](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/.travis.yml).

dev-requirements.txt
jonparrott: good call, will do.

tox.ini
jonparrott: Cool, will switch.

shazow:
> Sounds good. I think having a concrete bug to work on will help me allocate time. Can you create a bug for adding an AppEngineConnectionPool implementation using just URLFetch and assign it to me? 

That's this issue: https://github.com/shazow/urllib3/issues/583#issuecomment-101361461

Consider it yours. :) Github won't let me actually assign things to people unless they're maintainers on the repo. :/

> No promises on timeline as I'll be doing this in my spare time.

One of us, one of us! :P

.travis.yml
sigmavirus24: That's sad.

test/test_no_ssl.py
sigmavirus24: Mucking around with `meta_path` is fun ;)

sigmavirus24:
Thanks for answering my questions @jonparrott.

sigmavirus24:
And after one last check, I'm :+1: on this.

.travis.yml
```diff
@@ -1,17 +1,35 @@
...
 before_install:
   - openssl version
+  # Install Google Cloud Platform SDK to enable App Engine tests
+  - if [[ $TOXENV == 'gae' && ! -d $HOME/gcloud/google-cloud-sdk ]]; then
+      mkdir -p $HOME/gcloud &&
+      wget https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz --directory-prefix=$HOME/gcloud &&
+      cd $HOME/gcloud &&
+      tar xzf google-cloud-sdk.tar.gz &&
+      printf '\ny\n\ny\ny\n' | ./google-cloud-sdk/install.sh &&
```
t-8ch: using `yes` may be more future-proof

tox.ini
```diff
@@ -1,10 +1,23 @@
...
-   nosetests \
-   []
+    nosetests []
 setenv =
     PYTHONWARNINGS=always::DeprecationWarning
+
+[testenv:gae]
+basepython = python2.7
+deps= 
+    -r{toxinidir}/dev-requirements.txt
```
t-8ch: Maybe this can be replaced by `{[testenv]deps}`

tox.ini
```diff
@@ -1,10 +1,23 @@
...
+    nosegae
+commands=
+    nosetests \
+        --cover-min-percentage=0 \
+        --with-gae \
+        --gae-application=test/appengine/app.yaml \
+        test/appengine \
+        []
```
t-8ch: Could we put this all into a dedicated config file under `test/appengine`?

tox.ini
```diff
@@ -1,10 +1,23 @@
 [tox]
-envlist = py26, py27, py32, py33, py34, pypy
+envlist = py26, py27, py32, py33, py34, pypy, gae
 
 [testenv]
+passenv = PYTHONPATH TRAVIS*
```
t-8ch: Do we need the `TRAVIS*`?
How about setting a env var `GAE_PYTHONPATH` and then do `setenv = PYTHONPATH={env:GAE_PYTHONPATH}`?
This way we should not pick up anything which our users have set.
(or `URLLIB3_GAE_PYTHONPATH`)

.travis.yml
t-8ch: Can we pin the sdk to a specific release? We do it with all other dependencies too.
Best store the version in an env variable?

We could also move the download to tox. Which would probably annoy developers, but we could then exclude gae from the default envlist. @shazow?

tox.ini
```diff
@@ -1,10 +1,23 @@
...
-   []
+    nosetests []
 setenv =
     PYTHONWARNINGS=always::DeprecationWarning
+
+[testenv:gae]
+basepython = python2.7
+deps= 
+    -r{toxinidir}/dev-requirements.txt
+    nosegae
```
t-8ch: Could we move this to its own requirements file with a version pin?
Including all deps.

.travis.yml
shazow: Sounds sensible.

I'm okay with not pinning if it's not easy, I generally only pin things that have a history of breaking our tests (namely tornado).

t-8ch: The ping was mainly about downloading the sdk from tox. :-)

jonparrott: Our older releases tend to roll off the server rather quickly (this is actually currently breaking tests in my own project because I tried to pin a specific version). Even if you did, the ``gcloud components update`` command right after would update to the latest SDK anyway. In general, though, the GAE SDK pathologically avoids breaking APIs (with the exception of the dev_appserver2 switchover).

tox.ini
jonparrott: another requirements file just for gae?

tox.ini
jonparrott: A separate tox file?

t-8ch: Another config file for nose

test/appengine/test_urlfetch.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestHTTPS(AppEngineSandboxTest):
+    nosegae_urlfetch = True
+
+    def test_urlfetch_called_with_https(self):
+        """
+        Check that URLFetch is used when fetching https resources
+        """
+        raise SkipTest()  # Skipped for now because it fails.
```
t-8ch: Why does this fail? Should this be implemented before the merge or is it blocked somewhere?

test/appengine/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        if sys.version_info[:2] != (2, 7):
+            raise SkipTest("App Engine only tests on py2.7")
+
+        if 'APPLICATION_ID' not in os.environ:
+            raise SkipTest("NoseGAE plugin not used.")
+
+        try:
+            activate_sandbox()
+        except ImportError:
+            raise SkipTest("App Engine SDK not available.")
```
t-8ch: I would prefer blowing completely up here, as we want to execute this code only if we have GAE installed.

test/appengine/test_urlfetch.py
jonparrott: This is the issue behind the bug that spurred this. Urlfetch should be used for https, but it's currently not. This test will validate the fix (whenever I can get that in)

tox.ini
jonparrott: setup.cfg is currently the config file for nose, what do you recommend for this?

test/appengine/__init__.py
jonparrott: Do you mean we should make sure the SDK is available before trying to activate the sandbox?

t-8ch: I think we should not catch the import error. If someone tries to run the GAE tests without the SDK they should know, that this will not work.

tox.ini
t-8ch: `test/appengine/nose.cfg`

test/appengine/__init__.py
shazow: Hm maybe a Skip is a bit too quiet, but raising some pointer to instructions for how to install the SDK is not a bad idea.

tox.ini
jonparrott: Will do.

tox.ini
jonparrott: Will do.

jonparrott:
Updated with some of @t-8ch's feedback:
* separate config for gae
* separate requirements for gae
* Using GAE_PYTHONPATH instead of PYTHONPATH

Let me know what else need to be done here, once this is in I can start working on a GAE connection pool.

test/appengine/__init__.py
jonparrott: If these tests raise errors instead of skipping, the run for non-gae environments will fail.

t-8ch: We should then remove gae from the tox envlist, so it will not be run by default.
Or we download the SDK ourselves.
I would not bother with the automatic download atm.

jonparrott: I don't have any problem with removing gae from the default tox envlist. 

I would recommend leaving in the automatic download for travis so that travis can ensure the GAE tests have been run.

t-8ch: This does happen at the moment, no?

jonparrott: It does. I just pushed a new commit that removes gae from the tox envlist. This way, when running tox locally it'll skip the GAE tests. Travis will still run the GAE tests though. I think this strikes a good balance as it doesn't interrupt the current workflow at all.

If you want to run the GAE tests locally, you can do ``GAE_PYTHONPATH=~/google-cloud-sdk/platform/appengine tox -e gae``.

t-8ch:
For the record: Nose picks up setup.cfg and the new gae config.

@jonparrott Sorry for being so nitpicky!
:+1: from me!

jonparrott:
@t-8ch being picky is sometimes what makes great open-source software. This is my first contributing to urllib3 so I don't mind some back and forth while I learn the ropes. Thanks for taking the time to review.

re: setup.cfg + gae config. I figured, which is why it overrides the coverage percentage.

shazow:
@jonparrott Thank you very very much!! Really appreciate your effort here for adding coverage on the AppEngine platform but also migrating us to the env-based test matrix which will be handy in the future. :)

jonparrott:
No problem. I'll try to start on the connection pool later this week or next.

shazow:
@jonparrott Sounds good, much appreciated!

PR #633, opened 2015-05-28
HTTPHeaderDict in requests

### LINKED ISSUE #632: Urllib3 HTTPHeaderDict and proxy_from_url problem
It is a little bit redundant if I reexplain this problem, here is my post about that problem include code of the script: http://www.prxbx.com/forums/showthread.php?tid=2188&pid=18453#pid18453

The main thing is if I do any proxy_from_url or ProxyManager too, that problem appear: http://pastebin.com/g2mK1CW4
And if I use HTTPHeaderDict, i cannot use proxy, and if I use self.headers from http.server, I CAN use proxy, but ALL other page without proxy cannot even load.
The only and best possible at this time I found is use self.headers from http.server with proxy and HTTPHeaderDict with normal page:

```
            if "ghacks" in self.host:
                self.pool = urllib3.proxy_from_url('http://127.0.0.1:7777/')
                headers = self.headers
```

Code: http://www.prxbx.com/forums/showthread.php?tid=2188&pid=17911#pid17911

Need:
pip install colorama
pip install pyOpenSSL
pip install urllib3

and Python 3.4.2
urllib3 1.10.4

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18297390-urllib3-httpheaderdict-and-proxy_from_url-problem?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

shazow:
Do we want to accept HTTPHeaderDict objects in request header fields interchangeably with dicts/lists of tuples? Any reason why not?

Right now if we do that, `foo=bar` headers get encoded as `foo: ('Foo': 'bar')`

- [x] Tests
- [ ] Fix

Fixes #632

### DIFF
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -302,6 +302,7 @@ class TestHTTPHeaderDict(unittest.TestCase):
         hdict = {'Content-Length': '0', 'Content-type': 'text/plain', 'Server': 'TornadoServer/1.2.3'}
         h = dict(HTTPHeaderDict(hdict).items())
         self.assertEqual(hdict, h)
+        self.assertEqual(hdict, dict(HTTPHeaderDict(hdict)))
 
     def test_string_enforcement(self):
         # This currently throws AttributeError on key.lower(), should probably be something nicer
--- a/test/with_dummyserver/test_proxy_poolmanager.py
+++ b/test/with_dummyserver/test_proxy_poolmanager.py
@@ -9,6 +9,7 @@ from dummyserver.server import (
     DEFAULT_CA, DEFAULT_CA_BAD, get_unreachable_address)
 from .. import TARPIT_HOST
 
+from urllib3._collections import HTTPHeaderDict
 from urllib3.poolmanager import proxy_from_url, ProxyManager
 from urllib3.exceptions import (
     MaxRetryError, SSLError, ProxyError, ConnectTimeoutError)
@@ -48,7 +49,7 @@ class TestHTTPProxyManager(HTTPDummyProxyTestCase):
 
     def test_proxy_conn_fail(self):
         host, port = get_unreachable_address()
-        http = proxy_from_url('http://%s:%s/' % (host, port), retries=1)
+        http = proxy_from_url('http://%s:%s/' % (host, port), retries=1, timeout=0.05)
         self.assertRaises(MaxRetryError, http.request, 'GET',
                           '%s/' % self.https_url)
         self.assertRaises(MaxRetryError, http.request, 'GET',
@@ -222,6 +223,22 @@ class TestHTTPProxyManager(HTTPDummyProxyTestCase):
         self.assertEqual(returned_headers.get('Hickory'), None)
         self.assertEqual(returned_headers.get('Host'),
                 '%s:%s'%(self.https_host,self.https_port))
+
+    def test_headerdict(self):
+        default_headers = HTTPHeaderDict(a='b')
+        proxy_headers = HTTPHeaderDict()
+        proxy_headers.add('foo', 'bar')
+
+        http = proxy_from_url(
+            self.proxy_url,
+            headers=default_headers,
+            proxy_headers=proxy_headers)
+
+        request_headers = HTTPHeaderDict(baz='quux')
+        r = http.request('GET', '%s/headers' % self.http_url, headers=request_headers)
+        returned_headers = json.loads(r.data.decode())
+        self.assertEqual(returned_headers.get('foo'), 'bar')
+        self.assertEqual(returned_headers.get('baz'), 'quux')
 
     def test_proxy_pooling(self):
         http = proxy_from_url(self.proxy_url)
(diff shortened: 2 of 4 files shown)


shazow:
Looks like the problem is that `dict(foo='bar') != dict(HTTPHeaderDict(dict(foo='bar')))`. I think it's because Py27+ uses dict.viewitems() to clone which we're returning the wrong thing for.

```python
>>> h = HTTPHeaderDict({'foo': 'bar'})
>>> h.viewitems()
dict_items([('foo', ('foo', 'bar'))])
```

/cc people who worked on HTTPHeaerDict @schneier @ml31415

sigmavirus24:
I had a fix for this in my fork but we closed that PR in favor of @ml31415's header fixes. I don't think I have it lying around anymore either

shazow:
@sigmavirus24 Hm are you saying that this was a known bug? I had no idea. :/ Any guidance on how you fixed it?

ml31415:
The HTTPHeaderDict is implemented to preserve the case. Don't know where this get's messed up. I don't think it's headerdicts fault.

Lukasa:
I think I agree with @shazow: we need to override `viewitems` but we kinda can't, because it's a C-level function and you just can't really override those. This is why subclassing from `dict` is a bad idea: this object should really subclass `collections.MutableMapping`.

shazow:
@ml31415 It is headerdict's fault because viewitems returns the wrong thing. In order to support both lowercase and originalcase, we're storing things as `self[lowercase_key] = (originalcase_key, value)` which is what viewitems returns but should be returning the equivalent of `self[originalcase_key] = value`.

So we either need to override viewitems somehow (can't figure out how) or stop inheriting from dict and inherit from MutableMapping as before.

... Erm, what @Lukasa just said while I was writing this.

ml31415:
Not 100% sure, but I suppose `.viewitem` could get overriden as well, just as the other methods can. Viewitem was not considered in the current implementation yet, true, but to see it from another point of view: Is there any special reason against simply using `.items`  instead of `.vieitems` there?

shazow:
@ml31415 It's not us deciding whether to use items or viewitems. It's the Python implementation. :)

```python
a = dict(foo='bar')
b = dict(HTTPHeaderDict(a))`
a == b # False, because b is {"foo": ("foo", "bar")}
```

shazow:
This is basically what we'd want: https://github.com/kennethreitz/requests/blob/master/requests/structures.py#L14

ml31415:
Well, creating a dict that way is just not a lossless operation otherwise. Use `.itermerged` instead.

shazow:
@ml31415 Again, this is not something _I'm_ doing. It's what would be required to support passing an HTTPHeaderDict into an httplib request and having it Work(tm).

ml31415:
Not sure if we're understanding each other here, or talk about different things ... :) Any plain dict does not support duplicate header lines. The requests CaseInsensitiveHeaderDict also doesn't do that. I wouldn't expect requests to run with another header implementation without changes. One of them being, to call the necessary conversion function, i.e. `.itermerged`, before throwing it into something multiheader-unaware like httplib. If there should be something else, that would be required from HTTPHeaderDict, I'm sure it could be provided easily.

The ugly thing about it, iterating the whole headerdict for every http request would be quite a time waster, given that it's a quite rare case to send duplicate headers. As it wasn't implemented so far, no one seems to have missed it too much.

sigmavirus24:
Y'all are talking past each other. This was an issue I was at least aware of. If we want to fix it, I'll try to get some time during the night this week to fix it

shazow:
@sigmavirus24 Still up for taking a crack at this? Would love to get this fixed before the next release. (I think #670 is suffering from this too)

atodorov:
@shazow ,
confirmed. #670 is a duplicate of this one and the problem really breaks an important piece of software for me - the python-bugzilla tools and module in particular. 

Any idea if this can be fixed quickly or I need to go with a workaround in python-requests, essentially converting back to dict and then to their CaseInsensitiveDict ?

shazow:
@atodorov Unfortunately it can only be fixed as quickly as it takes for somebody to fix it. You're welcome to take a crack at it if you have the time to spare, but otherwise I'd put in a workaround. I suggest using `.items()`.

atodorov:
@shazow - I'm on Python 2.7 so .items is not an option according to your previous comments. I will take additional look at it though.

shazow:
@atodorov What's wrong with using `.items()` on Py27? The comments above refer to `.viewitems()` which is different.

shazow:
@sigmavirus24 Any chance you're still up for tackling this soon? I want to do a major version release early next week, would love to get this in.

shazow:
Closing in favour of #679.

PR #634, opened 2015-05-28
do not use assertions

t-8ch:
`assert` can be disabled at runtime.

### DIFF
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -227,8 +227,8 @@ class ProxyManager(PoolManager):
             port = port_by_scheme.get(proxy.scheme, 80)
             proxy = proxy._replace(port=port)
 
-        assert proxy.scheme in ("http", "https"), \
-            'Not supported proxy scheme %s' % proxy.scheme
+        if proxy.scheme not in ("http", "https"):
+            raise ValueError('Not supported proxy scheme %s' % proxy.scheme)
 
         self.proxy = proxy
         self.proxy_headers = proxy_headers or {}


urllib3/poolmanager.py
```diff
@@ -227,8 +227,8 @@ def __init__(self, proxy_url, num_pools=10, headers=None,
             port = port_by_scheme.get(proxy.scheme, 80)
             proxy = proxy._replace(port=port)
 
-        assert proxy.scheme in ("http", "https"), \
-            'Not supported proxy scheme %s' % proxy.scheme
+        if proxy.scheme not in ("http", "https"):
+            raise ValueError('Not supported proxy scheme %s' % proxy.scheme)
```
sigmavirus24: This is horribly backwards incompatible. `assert` raises an `AssertionError`, this raises a `ValueError`.

shazow: That's a horribly hyperbolic statement. :P Does requests depend on this behaviour?

sigmavirus24: 90% certain we don't, but others might. It's backwards incompatible because if someone has

```py
try:
    proxied_http = ProxyManager(url)
except AssertionError:
    pass
```

To catch the case where this exception is raised, then when they upgrade, the next time they see a proxy url that would have triggered that, they'll have an uncaught exception.

shazow: Yea that's possible but fairly unlikely (hardly horribly). Anyways, maybe we should just save this for a future major release rather than argue. :) I'll revert.

t-8ch:
@shazow How about raising the AssertionError explicitly? Without using `assert`?

shazow:
¯\_(ツ)_/¯ 

Is there a specific problem you're trying to solve by removing the assert, or just a style thing?

t-8ch:
Style and the possibillity that `assert` is optimised out and users will miss the error.

shazow:
Fair. Makes sense. @sigmavirus24 Objections?

sigmavirus24:
FWIW, I agree about the possibility that `assert` could be optimized out (although I don't think I've ever seen anyone run python with `-O`) and that it might be a problem. As for raising AssertionError explicitly, we could also define an exception that inherits from the AssertionError so that the handling I described will catch it but it will also be far more descriptive to users. Something like

```py
class ProxySchemeUnknown(AssertionError, HTTPError):
    # TODO(t-8ch): Stop inheriting from AssertionError in v2.0
    ...
```

But I have no objections to raising the AssertionError directly either.

Lukasa:
I think strictly this should be a `ValueError`, but I like the idea of multiply inheriting from `AssertionError` and `ValueError` until 2.0.

shazow:
Just to dispel any versioning debates early, major updates in urllib3 are 1.x.y, not x.y. Version 2.0 would be a nearly full rewrite/majority API changes/etc.

shazow:
There, wrote up the version scheme, comments welcome: https://github.com/shazow/urllib3/pull/637/

sigmavirus24:
@shazow that works for me, because it's written down somewhere and we can point people towards it if (and when) they complain

shazow:
@t-8ch Feel like doing a fresh PR with some permutation of the agreed variants above? :P

PR #635, opened 2015-05-28
make `python -m urllib3` work

t-8ch:


shazow:
lol cute, trying to figure out if we're going to regret this... @Lukasa @sigmavirus24 thoughts?

t-8ch:
I wanted to download the GAE SDK from our tox initialisation and needed a cross-platform way to do so.
And as we are a http library in the first place :-)

shazow:
@t-8ch No curl/wget/httpie? :p 

Dunno I'm worried people are going to want more and more features if we start with this. Entry points, ability to post, specify headers, next thing you know we're another httpie and half the code is just for the `__main__` scenario.

t-8ch:
@shazow Then all contributors would have to have curl installed (on windows *cough*).
Of course we can also just depend on httpie, but this would pull in pygments.
(If we don't document it, it is practically unsupported, no?)

sigmavirus24:
> If we don't document it, it is practically unsupported, no?

What is "it"? I'm confused.

t-8ch:
python -m urllib3

sigmavirus24:
Except "it's undocumented" is a bad excuse for a poorly supported feature that we don't want people to expand. If @shazow were to accept it, it would have to be documented as a hack for people who don't want to install other things and already have urllib3 installed. Which, when you put it that way, sounds kind of silly. A person is willing/able to install urllib3, but not httpie? That seems a bit far-fetched.

t-8ch:
@sigmavirus24 It was meant to be used by us to set up our own testenvironment, not users of the library.
I'll just move it into a standalone script in the `test` directory.

shazow:
Yea I think it'd be safer to move it into a fetch script in the tests directory. Else we should put a big ugly "please don't use/depend on this in any way" warning which we may as well just not have it exposed at that point.

Lukasa:
I think we should think really hard about going down this route. I feel like we should just be recommending httpie and call it a day.

shazow:
If we really want a pure-Python setup script for GAE, could shove it into a self-contained Python thing rather than a bunch of bash commands. I'm not too worried about Windows GAE devs not knowing how to download a zip file, though.

sigmavirus24:
Something like `tests/download_gae_sdk.py`? That's fine by me. I'd much rather that than `python -m urllib3`.

PR #637, opened 2015-05-31
Docs: urllib3's versioning semantics

shazow:
We don't use semver, but rather a compatibility-based scheme (which I feel is more popular in-practice for open source anyways). This outlines the details.


Lukasa:
:cake: :sparkle:

sigmavirus24:
With this in mind, I'm far less opposed to using a ValueError assuming it's well documented in the changelog.

PR #642, opened 2015-06-02
[RFC] Warn if header parsing failed

t-8ch:
Useful for broken servers like in kennethreitz/requests#2622

I'd like to get feedback for:
* The idea
* `logging` vs. `warnings`
* Loglevel
* Whatever you can come up with

@shazow @sigmavirus24 @Lukasa

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -10,6 +10,7 @@ from urllib3.exceptions import (
         SSLError,
         ProtocolError,
 )
+from urllib3.response import httplib
 from urllib3.util.ssl_ import HAS_SNI
 from urllib3.util.timeout import Timeout
 from urllib3.util.retry import Retry
@@ -18,9 +19,14 @@ from dummyserver.testcase import SocketDummyServerTestCase
 from dummyserver.server import (
     DEFAULT_CERTS, DEFAULT_CA, get_unreachable_address)
 
-from .. import onlyPy3
+from .. import onlyPy3, LogRecorder
 
 from nose.plugins.skip import SkipTest
+try:
+    from mimetools import Message as MimeToolMessage
+except ImportError:
+    class MimeToolMessage(object):
+        pass
 from threading import Event
 import socket
 import ssl
@@ -616,6 +622,57 @@ class TestHeaders(SocketDummyServerTestCase):
         HEADERS = {'Content-Length': '0', 'Content-type': 'text/plain'}
         r = pool.request('GET', '/')
         self.assertEqual(HEADERS, dict(r.headers.items())) # to preserve case sensitivity
+
+
+class TestBrokenHeaders(SocketDummyServerTestCase):
+    def setUp(self):
+        if issubclass(httplib.HTTPMessage, MimeToolMessage):
+            raise SkipTest('Header parsing errors not available')
+
+        super(TestBrokenHeaders, self).setUp()
+
+    def _test_broken_header_parsing(self, headers):
+        handler = create_response_handler((
+           b'HTTP/1.1 200 OK\r\n'
+           b'Content-Length: 0\r\n'
+           b'Content-type: text/plain\r\n'
+           ) + b'\r\n'.join(headers) + b'\r\n'
+        )
+
+        self._start_server(handler)
+        pool = HTTPConnectionPool(self.host, self.port, retries=False)
+
+        with LogRecorder() as logs:
+            pool.request('GET', '/')
+
+        expected_strings = [
+            'Errors while parsing headers',
+            self.host,
+            str(self.port),
+        ]
+
+        if not any([
+            all([s in record.msg for s in expected_strings])
+                for record in logs]):
+            self.fail('Missing log about unparsed headers')
+
+    def test_header_without_name(self):
+        self._test_broken_header_parsing([
+            b': Value\r\n',
+            b'Another: Header\r\n',
+        ])
+
+    def test_header_without_name_or_value(self):
+        self._test_broken_header_parsing([
+            b':\r\n',
+            b'Another: Header\r\n',
+        ])
+
+    def test_header_without_colon_or_value(self):
+        self._test_broken_header_parsing([
+            b'Broken Header',
+            b'Another: Header',
+        ])
 
 
 class TestHEAD(SocketDummyServerTestCase):
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -651,9 +651,7 @@ class TestBrokenHeaders(SocketDummyServerTestCase):
             str(self.port),
         ]
 
-        if not any([
-            all([s in record.msg for s in expected_strings])
-                for record in logs]):
+        if not any([all([s in record.msg for s in expected_strings]) for record in logs]):
             self.fail('Missing log about unparsed headers')
 
     def test_header_without_name(self):
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -645,14 +645,11 @@ class TestBrokenHeaders(SocketDummyServerTestCase):
         with LogRecorder() as logs:
             pool.request('GET', '/')
 
-        expected_strings = [
-            'Errors while parsing headers',
-            self.host,
-            str(self.port),
-        ]
-
-        if not any([all([s in record.msg for s in expected_strings]) for record in logs]):
-            self.fail('Missing log about unparsed headers')
+        for record in logs:
+            if 'Failed to parse headers' in record.msg and \
+                    pool._absolute_url('/').url == record.args[0]:
+                return
+        self.fail('Missing log about unparsed headers')
 
     def test_header_without_name(self):
         self._test_broken_header_parsing([
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -647,7 +647,7 @@ class TestBrokenHeaders(SocketDummyServerTestCase):
 
         for record in logs:
             if 'Failed to parse headers' in record.msg and \
-                    pool._absolute_url('/').url == record.args[0]:
+                    pool._absolute_url('/') == record.args[0]:
                 return
         self.fail('Missing log about unparsed headers')
 
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -389,12 +389,12 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         except HeaderParsingErrors as hpe:  # Platform-specific: Python 3
             log.warning(
                 'Failed to parse headers (url=%s): %s',
-                self._absolute_url(url).url, hpe, exc_info=True)
+                self._absolute_url(url), hpe, exc_info=True)
 
         return httplib_response
 
     def _absolute_url(self, path):
-        return Url(scheme=self.scheme, host=self.host, port=self.port, path=path)
+        return Url(scheme=self.scheme, host=self.host, port=self.port, path=path).url
 
     def close(self):
         """
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -17,6 +17,7 @@ from .exceptions import (
     ClosedPoolError,
     ProtocolError,
     EmptyPoolError,
+    HeaderParsingErrors,
     HostChangedError,
     LocationValueError,
     MaxRetryError,
@@ -38,7 +39,7 @@ from .request import RequestMethods
 from .response import HTTPResponse
 
 from .util.connection import is_connection_dropped
-from .util.response import extract_parsing_errors
+from .util.response import assert_header_parsing
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import get_host, Url
@@ -383,11 +384,12 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                                           httplib_response.status,
                                           httplib_response.length))
 
-        header_parsing_errors = extract_parsing_errors(httplib_response.msg)
-        if header_parsing_errors:  # Platform-specific: Python 3.
-            logging.warning(
-                'Failed to completely parse headers for %s: %s' %
-                (self._absolute_url(url).url, header_parsing_errors))
+        try:
+            assert_header_parsing(httplib_response.msg)
+        except HeaderParsingErrors as hpe:  # Platform-specific: Python 3
+            log.warning(
+                'Failed to parse headers (url=%s): %s',
+                self._absolute_url(url).url, hpe, exc_info=True)
 
         return httplib_response
 
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -388,13 +388,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             logging.warning(
                 'Errors while parsing headers for %s %s: %s. Unparsed data: %s' %
                 (method,
-                    self._full_url(url).url,
+                    self._absolute_url(url).url,
                     header_parsing_errors.defects or 'Unknown',
                     repr(header_parsing_errors.unparsed_data)))
 
         return httplib_response
 
-    def _full_url(self, path):
+    def _absolute_url(self, path):
         return Url(
                 scheme=self.scheme, host=self.host, port=self.port, path=path)
 
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -38,6 +38,7 @@ from .request import RequestMethods
 from .response import HTTPResponse
 
 from .util.connection import is_connection_dropped
+from .util.response import extract_parsing_errors
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import get_host, Url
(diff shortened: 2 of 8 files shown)


urllib3/util/response.py
```diff
@@ -20,3 +25,33 @@ def is_fp_closed(obj):
...
+
+
+def validate_headers(headers):
+    """
+    Checks whether headers have been successfully and completely parsed.
+
+    Only works on Python 3.
+
+    :param headers: Headers to check.
+    :type headers: `httplib.HTTPMessage`.
```
Lukasa: Are these single backticks right?

urllib3/response.py
```diff
@@ -319,6 +319,8 @@ def from_httplib(ResponseCls, r, **response_kw):
         with ``original_response=r``.
         """
         headers = r.msg
+        validate_headers(headers)
```
shazow: Should this do anything with the response value?

Maybe `assert_headers` would be a better name, and have it raise an exception if it fails? Then we can catch/log in the handler with more context rather than in the validator itself.

urllib3/util/response.py
```diff
@@ -20,3 +25,33 @@ def is_fp_closed(obj):
...
+
+    :returns: Whether any errors occured during header parsing.
+    :rtype: bool
```
Lukasa: Why does this method have a return value? We aren't using it anywhere.

urllib3/util/response.py
```diff
@@ -20,3 +25,33 @@ def is_fp_closed(obj):
...
+    result = False
+    defects = getattr(headers, 'defects', None)
+
+    if defects:
+        log.error('Errors while parsing headers: %s', defects)
+        result = True
+
+    get_payload = getattr(headers, 'get_payload', None)
+
+    if get_payload:
```
shazow: Can we invert the logic rather than nesting it?

```python
if not get_payload:
    return result

unparsed_headers...
```

urllib3/util/response.py
t-8ch: Nope :-(

urllib3/util/response.py
t-8ch: I wanted it to be more generic, so other people can use it.

Lukasa: Generally I'd be in favour of going with YAGNI here and abandoning it. However, if you decide to keep it, I'd invert the logic: a method called `validate_headers` should probably return `True` if the headers are valid and `False` if they aren't.

urllib3/response.py
t-8ch: More context is indeed needed for users to be able to handle it properly.

shazow:
I like the idea.

- What happens when headers get mangled? Do they get omitted from our HTTPHeaderDict in the response?
- Re logging vs warning, I lean towards log.warn() rather than log.error(). I feel like log.error() is when something unexpected goes wrong with urllib3 itself that is not recoverable. If we're ignoring the misparsed headers, then it's technically recovered.
- Tests will be needed, ofc. :)

urllib3/response.py
shazow: Note we already have assert_fingerprints which raises on failure. I think that's a good pattern to follow here. :)

Lukasa: I don't know that that's true.

If `assert_fingerprints` fails there is a security risk, and we should explode. It's not quite so clear that failed header parsing is a security risk (it *may* be, but an argument can be made for lenient parsing).

urllib3/util/response.py
shazow: https://github.com/shazow/urllib3/pull/642#discussion_r31504790

urllib3/response.py
shazow: @Lukasa 

>  ... Then we can catch/log in the handler with more context rather than in the validator itself.

t-8ch:
Unfortunately the response does not know the URL it came from, which we would want to log.
I have to look at this again later.

urllib3/response.py
shazow: If we feel strongly it should return a boolean rather than raise, then I'd follow the `is_foo()` pattern we use elsewhere too.

shazow:
> Unfortunately the response does not know the URL it came from, which we would want to log.

https://github.com/shazow/urllib3/issues/576 strikes again.

Lukasa:
Uh...I just checked whether this would help with kennethreitz/requests#2622 and I don't think it will. @t-8ch, are you sure this works as you expect? This:

```python
import http.client as http
c = http.HTTPConnection('www.merrickbank.com', 80)
c.request('GET', '/')
r = c.getresponse()
r.read()

print("Defects: %r" % r.headers.defects)
print("Payload: %r" % r.headers.get_payload())
```

prints this for me:

```                                
Defects: []
Payload: ''
```

Lukasa:
Put another way, a quick look through the code doesn't show me anything that actually populates those fields of the headers object.

t-8ch:
@Lukasa this is not the same `headers`. If you look at the patch the `headers` variable is overridden just after I do the validation.

Lukasa:
@t-8ch Rewriting to this doesn't help:

```python
import http.client as http
c = http.HTTPConnection('www.merrickbank.com', 80)
c.request('GET', '/')
r = c.getresponse()
r.read()

print("Defects: %r" % r.msg.defects)
print("Payload: %r" % r.msg.get_payload())
```

Can you provide me a pure httplib example that shows that this works?

t-8ch:
@Lukasa Please forget my last comment. You have the enable gzip to get the broken headers from them.

Lukasa:
Hmm. Sadly, it seems that Merrick Bank have fixed their service: I can no longer reproduce the requests bug, which makes this difficult or impossible to test.

shazow:
Go go gadget socket-level tests.

Lukasa:
Alright, I did a quick socket level test and this does work! \o/ Victory to @t-8ch!

shazow:
\<meme="share-the-load.gif">Share the sourceee\</meme>

t-8ch:
@Lukasa would you mind sharing your test? I can't manage to write a working one. (probably because it is 4 o'clock in the morning...)

t-8ch:
Ping @shazow @Lukasa

t-8ch:
@Lukasa
I couldn't believe that a bank would actually get their shit together and fix their stuff. And indeed, for their website is still broken.

Lukasa:
Functionally this seems plausible, but I suspect @shazow will want to do some structure/style review.

urllib3/connectionpool.py
```diff
@@ -381,8 +382,22 @@ def _make_request(self, conn, method, url, timeout=_Default,
...
+            logging.warning(
+                'Errors while parsing headers for %s %s: %s. Unparsed data: %s' %
+                (method,
+                    self._full_url(url).url,
+                    header_parsing_errors.defects or 'Unknown',
+                    repr(header_parsing_errors.unparsed_data)))
+
         return httplib_response
 
+    def _full_url(self, path):
```
shazow: `_absolute_url` may be a better name.

urllib3/connectionpool.py
```diff
@@ -381,8 +382,22 @@ def _make_request(self, conn, method, url, timeout=_Default,
         log.debug("\"%s %s %s\" %s %s" % (method, url, http_version,
                                           httplib_response.status,
                                           httplib_response.length))
+
+        header_parsing_errors = extract_parsing_errors(httplib_response.msg)
+        if header_parsing_errors:  # Platform-specific: Implementation.
```
shazow: What does Implementation mean? "Python 3 and PyPy 2" might be clearer.

urllib3/util/response.py
```diff
@@ -20,3 +37,34 @@ def is_fp_closed(obj):
...
+    # This will fail silently if we pass in the wrong kind of parameter.
+    # To make debugging easier add an explicit check.
+    if not isinstance(headers, httplib.HTTPMessage):
+        raise TypeError('expected httplib.Message, got {}.'.format(
+            type(headers)))
+
+    defects = getattr(headers, 'defects', None)
+    get_payload = getattr(headers, 'get_payload', None)
+
+    if get_payload:  # Platform-specific: Implementation dependent.
```
shazow: Some details about which platform this is known to work with is helpful for future revisions.

urllib3/util/response.py
```diff
@@ -20,3 +37,34 @@ def is_fp_closed(obj):
...
+        unparsed_data = get_payload()
+    else:  # Platform-specific
```
shazow: Can we set `unparsed_data = None` before the if-statement and just let the default pass-through? (Not a fan of else's that don't do anything)

urllib3/connectionpool.py
```diff
@@ -381,8 +382,22 @@ def _make_request(self, conn, method, url, timeout=_Default,
         log.debug("\"%s %s %s\" %s %s" % (method, url, http_version,
                                           httplib_response.status,
                                           httplib_response.length))
+
+        header_parsing_errors = extract_parsing_errors(httplib_response.msg)
```
shazow: - Why do we prefer returning a value rather than raising an exception/catch-logging?
- Why return a namedtuple thing? This also feels like it should be an exception with a couple extra properties.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -618,6 +624,57 @@ def test_httplib_headers_case_insensitive(self):
...
+           b'HTTP/1.1 200 OK\r\n'
+           b'Content-Length: 0\r\n'
+           b'Content-type: text/plain\r\n'
+           ) + b'\r\n'.join(headers) + b'\r\n'
+        )
+
+        self._start_server(handler)
+        pool = HTTPConnectionPool(self.host, self.port, retries=False)
+
+        with LogRecorder() as logs:
```
shazow: Can you remind me why this is better than `with warnings.catch_warnings(record=True) as w:`?

I recall somebody mentioned some shortcomings but I can't find what they were.

urllib3/connectionpool.py
t-8ch: Sure

urllib3/connectionpool.py
```diff
@@ -381,8 +382,22 @@ def _make_request(self, conn, method, url, timeout=_Default,
...
+                'Errors while parsing headers for %s %s: %s. Unparsed data: %s' %
+                (method,
+                    self._full_url(url).url,
+                    header_parsing_errors.defects or 'Unknown',
+                    repr(header_parsing_errors.unparsed_data)))
+
         return httplib_response
 
+    def _full_url(self, path):
+        return Url(
```
shazow: We're not doing col80 anymore, feel free to merge the line.

urllib3/connectionpool.py
```diff
@@ -381,8 +382,22 @@ def _make_request(self, conn, method, url, timeout=_Default,
         log.debug("\"%s %s %s\" %s %s" % (method, url, http_version,
                                           httplib_response.status,
                                           httplib_response.length))
+
+        header_parsing_errors = extract_parsing_errors(httplib_response.msg)
+        if header_parsing_errors:  # Platform-specific: Implementation.
+            logging.warning(
+                'Errors while parsing headers for %s %s: %s. Unparsed data: %s' %
```
shazow: May as well give the error a sensible `__repr__` and use `%r`.

test/with_dummyserver/test_socketlevel.py
t-8ch: Aren't `logging` and `warnings` different things?

shazow: Oh right you are!

urllib3/connectionpool.py
t-8ch: Why not `__str__`?

shazow: Whatever makes sense to you. I just saw you do `repr(...)` below. :)

shazow: I mean rather than formatting the error in the warning string, would be better to just have it formatted by the error itself and the warning string be mostly-static (just add the url for context).

t-8ch: Yes, added this. The `repr` is to have newlines in the data escaped

shazow: Btw, rephrasing suggestion: It's confusing that a _warning_ is mentioning _errors_. Maybe something like: `"Failed to parse headers ({0}) for url: {1}".format(err, url)"`

urllib3/connectionpool.py
t-8ch: For me exceptions are something even more special than `logging.error`. So if we will not abort we should not be using them.

urllib3/connectionpool.py
t-8ch: I'd prefer to have the URL first, because the unparsed data may be big:
"Failed to parse headers for url '{0}': '{1}'".format(err, url)"?

urllib3/connectionpool.py
shazow: As far as the header parsing layer is concerned (urllib3.util.request), it has failed to perform its job given the input. It raises an exception.

Then the layer above is like "oh well, I don't mind if some headers are missing, I'll handle it by emitting a warning and continuing." That's how we're using exceptions everywhere, see all the socket error related exceptions that we attempt to recover from rather than outright fail.

As long as the respective layer of abstraction considers what happened a failure, exceptions are appropriate IMO.

t-8ch: ok

urllib3/connectionpool.py
sigmavirus24: Also, nit: logging best practice is to not interpolate the values yourself.

urllib3/connectionpool.py
sigmavirus24: We're doing col9001, right? ;)

urllib3/connectionpool.py
shazow: > I'd prefer to have the URL first, because the unparsed data may be big

Fair, sure.

Take a look at other logs for consistent formatting. As a rule of thumb, I usually write the explanation, then colon, then value. No quotes around the value (colon is easy enough to parse on if necessary).

Here's a similar example where we do url -> repr, here: `log.debug("Incremented Retry for (url='%s'): %r" % (url, new_retry))`  (IMO the "for" is extraneous here)

So how about: `"Failed to parse headers (url={0}): {1}".format(url, err)`

urllib3/connectionpool.py
shazow: We got 99 characters but a new-line ain't one.

sigmavirus24:
So whenever I see something implement `__bool__`/`__nonzero__` I cringe a little. I agree with @shazow that the exception workflow is better and more consistent across the rest of urllib3.

t-8ch:
@shazow @Lukasa Here we go again.

urllib3/connectionpool.py
```diff
@@ -381,8 +383,19 @@ def _make_request(self, conn, method, url, timeout=_Default,
...
+            assert_header_parsing(httplib_response.msg)
+        except HeaderParsingErrors as hpe:  # Platform-specific: Python 3
+            log.warning(
+                'Failed to parse headers (url=%s): %s',
+                self._absolute_url(url).url, hpe, exc_info=True)
+
         return httplib_response
 
+    def _absolute_url(self, path):
+        return Url(scheme=self.scheme, host=self.host, port=self.port, path=path)
```
shazow: Seems we're just using the stringified url everywhere, not the original Url object. Maybe `self._absolute_url(url).url` should just be `self._absolute_url(url)`.

urllib3/exceptions.py
```diff
@@ -176,3 +176,20 @@ class ProxySchemeUnknown(AssertionError, ValueError):
...
+
+class HeaderParsingErrors(HTTPError):
+    def __init__(self, defects, unparsed_data):
+        super(HeaderParsingErrors, self).__init__(defects, unparsed_data)
+
+    def __str__(self):
+        return '%s, unparsed data: %r' % (
+                    self.defects or 'Unknown', self.unparsed_data)
+
+    @property
```
shazow: Is this better than just assigning them on init?

shazow:
Looks much better, couple of small nits then good to merge. :)

urllib3/exceptions.py
t-8ch: It seemed to me, exceptions could only hold one attribute. CPython implementation magic. And blow up, as soon as the second attribute is assigned. This was my workaround.

urllib3/connectionpool.py
t-8ch: Will do

urllib3/exceptions.py
shazow: Umm that should not be the case. See our other exceptions.

```python
class FooException(Exception):
     def __init__(self, foo, bar):
         super(FooException, self).__init__()
         self.foo = foo
         self.bar = bar
```

... works fine.

t-8ch:
ping @shazow

urllib3/exceptions.py
```diff
@@ -176,3 +176,10 @@ class ProxySchemeUnknown(AssertionError, ValueError):
     def __init__(self, scheme):
         message = "Not supported proxy scheme %s" % scheme
         super(ProxySchemeUnknown, self).__init__(message)
+
+
+class HeaderParsingErrors(HTTPError):
+    def __init__(self, defects, unparsed_data):
+        message = '%s, unparsed data: %r' % (
```
shazow: Still weren't able to assign the properties? This is fine though.

Also such col80, very wow.

urllib3/exceptions.py
```diff
@@ -176,3 +176,10 @@ class ProxySchemeUnknown(AssertionError, ValueError):
     def __init__(self, scheme):
         message = "Not supported proxy scheme %s" % scheme
         super(ProxySchemeUnknown, self).__init__(message)
+
+
+class HeaderParsingErrors(HTTPError):
```
shazow: Could you make Errors singular, for consistency?

shazow:
:+1:

Issue #644, opened 2015-06-04
Connection pool exhausted when connection failures occur, should refill with empty connections


jlatherfold:
If preload_content is not specified on a request, connections are implicitly returned to the pool after the request is read. However, when a timeout error occurs or the connection is dropped the connection is closed but not returned in to the pool (since the response is None). This problem is compounded when retries are turned on since the next attempt will grab a new connection from the pool thus depleting it further. With non-blocking pools this is not a problem since we will create a new connection if the pool is empty but when the pool is blocking we have found that eventually the pool size becomes zero (after no.of connections timeout errors) which causes the calling application to hang on it's next request.
This can be fixed in connectionpool.urlopen in the exception handlers by explicitly returning the closed connection to the pool via _put_conn (if release_conn is false), since subsequent calls to _get_conn will check if the connection has been dropped and return a new one.

Lukasa:
Hello again!

I think we need to be sure that we close the connection before we return them to the pool. Otherwise we run the risk of attempting to re-use a live connection that has timed out, which will end extremely poorly for us.

shazow:
We should be putting a `None` into the pool if we want to discard the connection. That will be replaced with a fresh connection when we try to get it out of the pool. That does sound like a bug.

Lukasa:
@shazow's idea is even better than mine.

shazow:
@jlatherfold Is this something you'd be interested in working on?

Producing a failing test would be the first step. Make a ConnectionPool with a small pool (1-2), trigger some connection exceptions, see if the pool gets exhausted.

jlatherfold:
Hi,

I've already done this, hence this bug report. Locally I've fixed it by putting the [closed] connection back on to the pool in the exception handlers (which do close the connections). Those Nones or closed connections are discarded and a new one is returned in subsequent calls to _get_conn. Sadly I don't know when I will get time to do any more on this right now.

shazow:
@jlatherfold Would be handy if you can share your code for reproducing the issue. :)

Also if you can confirm that putting `None` rather than the closed connection still works as expected, that would be great.

jlatherfold:
I can confirm that _put_conn(None) works as expected.

Here's some basic code, which also produces this error. Setting the read-timeout to a low value will cause read timeout exceptions. You'll need to hit a url with some large content - doesn't have to be massive - 500KB or so although you can probably replicate this with much less):


import os
import sys
import requests

def get(req, url):
[...]

PR #646, opened 2015-06-10
Cleanup properly after timeout or connection-based error.

jlatherfold:
If you run the test I've added without the changes it will fail. It should pass with the updates to connectionpool.py.
I've added cleanup code for BaseSSLError, CertificateError, and SSLError as well (which looks like it should be handled in the same way) but no test for this.

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -17,6 +17,8 @@ from urllib3.exceptions import (
     SSLError,
 )
 
+from dummyserver.testcase import HTTPDummyServerTestCase
+
 from socket import error as SocketError
 from ssl import SSLError as BaseSSLError
 
@@ -226,40 +228,51 @@ class TestConnectionPool(unittest.TestCase):
 
         self.assertRaises(Empty, old_pool_queue.get, block=False)
 
-    def test_connection_recycling(self):
+
+class TestConnectionPoolConnectionCleanup(HTTPDummyServerTestCase):
+    '''
+    Tests in this suite are concerned with managing the pool
+    connections, specifically in the event of errors where
+    connection recycling must be handled internally.
+    '''
+    def test_cleanup_on_connection_error(self):
         '''
         Test that connections are recycled to the pool on 
-        connection/readtimeout errors where no http response 
-        is received.
+        connection errors where no http response is received.
         '''
         poolsize = 2
-        with HTTPConnectionPool('google.com', maxsize=poolsize, 
-                                  retries=1, block=True) as pool:
-
-            self.assertEqual(pool.pool.qsize(), poolsize)
-
-            # set a silly read timeout to force timeout exception. 
-            # We won't get a response for this and so the conn won't 
-            # be implicitly returned to the pool.
-            pool.timeout._read = 0
-            self.assertRaises(MaxRetryError, pool.urlopen, 'GET', 
-                              'http://google.com',
-                              release_conn=False)
-
-            # the pool should still contain poolsize elements
-            self.assertEqual(pool.pool.qsize(), poolsize)
-
         with HTTPConnectionPool('localhost', maxsize=poolsize, 
                                   retries=1, block=True) as pool:
 
             self.assertEqual(pool.pool.qsize(), poolsize)
 
-            # force a connection error by supplying a 
-            # non-existent url. We won't get a response for this 
-            # and so the conn won't be implicitly returned to 
-            # the pool.
+            # force a connection error by supplying a non-existent 
+            # url. We won't get a response for this  and so the 
+            # conn won't be implicitly returned to the pool.
             self.assertRaises(MaxRetryError, pool.urlopen, 'GET', 
                               'localhost/nosuchurl',
+                              release_conn=False)
+
+            # the pool should still contain poolsize elements
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
+    def test_cleanup_on_read_timeout(self):
+        '''
+        Test that connections are recycled to the pool on 
+        connection errors where no http response is received.
+        '''
+        poolsize = 2
+        with HTTPConnectionPool('localhost', maxsize=poolsize, 
+                                  retries=1, block=True) as pool:
+
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
+            # set a silly read timeout to force timeout exception.
+            # We won't get a response for this and so the conn won't
+            # be implicitly returned to the pool.
+            pool.timeout._read = 0
+            self.assertRaises(MaxRetryError, pool.urlopen, 'GET', 
+                              'localhost',
                               release_conn=False)
 
             # the pool should still contain poolsize elements
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -226,6 +226,45 @@ class TestConnectionPool(unittest.TestCase):
 
         self.assertRaises(Empty, old_pool_queue.get, block=False)
 
+    def test_connection_recycling(self):
+        '''
+        Test that connections are recycled to the pool on 
+        connection/readtimeout errors where no http response 
+        is received.
+        '''
+        poolsize = 2
+        with HTTPConnectionPool('google.com', maxsize=poolsize, 
+                                  retries=1, block=True) as pool:
+
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
+            # set a silly read timeout to force timeout exception. 
+            # We won't get a response for this and so the conn won't 
+            # be implicitly returned to the pool.
+            pool.timeout._read = 0
+            self.assertRaises(MaxRetryError, pool.urlopen, 'GET', 
+                              'http://google.com',
+                              release_conn=False)
+
+            # the pool should still contain poolsize elements
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
+        with HTTPConnectionPool('localhost', maxsize=poolsize, 
+                                  retries=1, block=True) as pool:
+
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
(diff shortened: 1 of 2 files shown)


test/test_connectionpool.py
```diff
@@ -226,6 +226,45 @@ def test_contextmanager(self):
...
+            # force a connection error by supplying a 
+            # non-existent url. We won't get a response for this 
+            # and so the conn won't be implicitly returned to 
+            # the pool.
+            self.assertRaises(MaxRetryError, pool.urlopen, 'GET', 
+                              'localhost/nosuchurl',
+                              release_conn=False)
+
+            # the pool should still contain poolsize elements
+            self.assertEqual(pool.pool.qsize(), poolsize)
```
Lukasa: Why is this test doubled-up? Is it just for the different types of exceptions? If it is, I think it'd probably be nicer to have it be either separate tests, or a loop in the same test.

urllib3/connectionpool.py
```diff
@@ -409,7 +431,7 @@ def is_same_host(self, url):
 
         # TODO: Add optional support for socket.gethostbyname checking.
         scheme, host, port = get_host(url)
-
+
```
Lukasa: Please remove this whitespace.

urllib3/connectionpool.py
```diff
@@ -568,25 +590,19 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
             # Close the connection. If a connection is reused on which there
             # was a Certificate error, the next request will certainly raise
             # another Certificate error.
-            if conn:
-                conn.close()
-                conn = None
+            conn = self._cleanup_conn(conn, not release_conn)
```
Lukasa: It's a little weird to me that the `release_conn` argument to `cleanup_conn` is being passed a value `not release_conn`. Can we get the two arguments around the same way somehow, or at least give the argument to `cleanup_conn` a clearer name?

Lukasa:
This looks entirely reasonable to me. For those keeping track (and so we can build up a nice connected web of issues), this is a fix for #644.

A few notes that I'd like addressed, but otherwise it's a :+1: from me.

test/test_connectionpool.py
jlatherfold: Yes they force different exceptions, but actually they are handled in the same handler (for now) so we could just remove one of these. They are doubled up because they are slightly different and use a different host. You decide - create separate tests for each or remove one (which one - localhost or google)?

test/test_connectionpool.py
```diff
@@ -226,6 +226,45 @@ def test_contextmanager(self):
...
+        with HTTPConnectionPool('google.com', maxsize=poolsize, 
+                                  retries=1, block=True) as pool:
+
+            self.assertEqual(pool.pool.qsize(), poolsize)
+
+            # set a silly read timeout to force timeout exception. 
+            # We won't get a response for this and so the conn won't 
+            # be implicitly returned to the pool.
+            pool.timeout._read = 0
+            self.assertRaises(MaxRetryError, pool.urlopen, 'GET',
```
shazow: Does this make an external request to actual google.com? We don't allow external requests in our test suite to keep it fast and runnable offline.

Could either mock the connection objects or use our with_dummyserver suite.

On a related note, would be nice if we had somekind of FakeConnection object we could use in cases like this. (Just thinking aloud.)

urllib3/connectionpool.py
```diff
@@ -275,6 +275,28 @@ def _put_conn(self, conn):
         if conn:
             conn.close()
 
+    def _cleanup_conn(self, conn,  release_conn):
```
shazow:
```diff
-    def _cleanup_conn(self, conn,  release_conn):
+    def _cleanup_conn(self, conn, release_conn):
```

urllib3/connectionpool.py
```diff
@@ -275,6 +275,28 @@ def _put_conn(self, conn):
...
+        """
+        Close the connection and return it to the pool if requested. 
+        Returns ``None``.
+
+        :param conn:
+            Connection object for the current host and port as returned by
+            :meth:`._new_conn` or :meth:`._get_conn`.
+       
+        :param  release_conn:
```
shazow:
```diff
-        :param  release_conn:
+        :param release_conn:
```

urllib3/connectionpool.py
jlatherfold: I agree. release_conn though is a good name for what we want to do explicitly in _cleanup_conn, but we only want to do it if we're not releasing it in the calling method. It just so happens that it's also a good name in the calling method since that's what he's going to do. I could change it to something like self._cleanup_conn(release_conn == False), otherwise we could do something like:

def _cleanup_conn(self, conn, auto_release):
   ....
   if not auto_release:
        self._put_conn(None)

...and have the caller do self._cleanup_conn(conn, release_conn).
I'm open to suggestions....

shazow: Agree, the extra negation is confusing.

shazow: I'd do 

```python
def _cleanup_conn(self, conn, replace_conn=False):
    if not conn:
        return

    conn.close()
    if replace_conn:
        self._put_conn(None)
```

shazow: ... Then call it with

```python
conn = self._cleanup_conn(conn, replace_conn=not release_conn)
```

It still has the negation, but at least it's not negating into the same variable name and you get context in the call.

test/test_connectionpool.py
jlatherfold: I think it does, but we can remove that test since it invokes the same exception handler as for ConnectionError or SocketError.

shazow: Whatever works for you. Feel free to piggyback on any existing tests too, if we were already exercising the condition but just weren't asserting the pool size.

But anything that actually opens a socket should go in with_dummyserver subdir.

Note that all the tests in this suite might setup pools but none of them actually make requests.

jlatherfold:
Hmm, what's up with the test - works locally for me. As stated before we can remove the tst for read timeout.

shazow:
It's possible that our tests are just flakey. I triggered a re-run.

Btw, further down we have:

```python
        ...
        finally:
            if release_conn:
                # Put the connection back to be reused. If the connection is
                # expired then it will be None, which will get replaced with a
                # fresh connection during _get_conn.
                self._put_conn(conn)
```

Any idea why that was not sufficient?

Lukasa:
> Any idea why that was not sufficient?

Seems to have inverted logic around `release_conn`, which would be my guess. The problem is with those other exceptions we want to return a `None` *always*, because we cannot possibly return a response with a connection a user could, you know, use.

jlatherfold:
Yes, release_conn is false when stream=True, since we need the connection to hang around while we pull the content. I work with a system where we hit a server serving large content and we use streaming. Since we're not preloading the content we can't put the connection back until we've finished with it. I ran trace through requests and urllib3 to work out what was going on and release_conn was False - I assume that it's set in requests or elsewhere in urllib3.

jlatherfold:
Btw, when streaming and you hit a 404 - you will need to consume the response otherwise we get the same problem. That's not at all obvious and I only found that the other day when accidentally setting one of our pools to blocking. So, if stream=True and we get a HTTPError we need to do a _ = e.response.content otherwise another conn is lost to us. What happens to it since it's not been closed - potential resource leak I assume - I don't know when or how connections are garbage-collected.

shazow:
I may refactor this back out to only release the conn is one place again, hope that's okay. :)

urllib3/connectionpool.py
```diff
@@ -568,25 +590,19 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
 
         except (TimeoutError, HTTPException, SocketError, ConnectionError) as e:
-            if conn:
-                # Discard the connection for these exceptions. It will be
-                # be replaced during the next _get_conn() call.
-                conn.close()
-                conn = None
+            # Discard the connection for these exceptions. It will be
+            # be replaced during the next _get_conn() call.            
+            conn = self._cleanup_conn(conn, replace_conn=not release_conn)
```
shazow: Should this be releasing too? This does not re-raise an exception.

jlatherfold:
As you wish. The important thing is that we don't 'lose' connections and the pool become exhausted during blocked scenarios. The implementation details are up to you....

urllib3/connectionpool.py
jlatherfold: Yes, but only if release_conn is False. The exception is raised further down the line when retries are all used up. If a connection is closed and no response will be consumed, and if release_conn is False, when else would it be release?

shazow: Right you are!

shazow:
@jlatherfold Thank you for doing all the hard work on this one. :) You're welcome to add yourself to our CONTRIBUTORS.txt file!

jlatherfold:
Cool. Happy to help.

jlatherfold:
Hi,

I just took a look at your refactoring and I think that you may have introduced a subtle bug. The problem may occur during retries, consider the following:

* request fails with read timeout - you set release_conn=True and the connection is released.
* the pool has retries=1 so just after you release the conn you then make a recursive call to urlopen, but now with release_conn=True.
* this time the request succeeds, however since release_conn is now true the conn is released - but in the case where we don't want pre-loaded content (eg. stream=True) we will not be able to read since the connection is gone and an exception will be thrown (I assume).

The problem is that if release_conn is False to begin with, you should not set it to true in the event of retries, since you've now changed the default behaviour of the pool. I may be wrong about this but it at least deserves a closer look.

shazow:
@jlatherfold Which line are you referring to in the change? Feel free to ping it with a comment here: https://github.com/shazow/urllib3/pull/647/files

PR #647, opened 2015-06-10
Replenish pool when release_conn=False and error occurred

### LINKED ISSUE #644: Connection pool exhausted when connection failures occur, should refill with empty connections

shazow:
Replaces #646
Fixes #644

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -647,6 +647,27 @@ class TestConnectionPool(HTTPDummyServerTestCase):
 
         self.assertEqual(b'123' * 4, response.read())
 
+    def test_cleanup_on_connection_error(self):
+        '''
+        Test that connections are recycled to the pool on 
+        connection errors where no http response is received.
+        '''
+        poolsize = 3
+        with HTTPConnectionPool(self.host, self.port, maxsize=poolsize, block=True) as http:
+            self.assertEqual(http.pool.qsize(), poolsize)
+
+            # force a connection error by supplying a non-existent 
+            # url. We won't get a response for this  and so the 
+            # conn won't be implicitly returned to the pool.
+            self.assertRaises(MaxRetryError,
+                http.request, 'GET', '/redirect', fields={'target': '/'}, release_conn=False, retries=0)
+
+            r = http.request('GET', '/redirect', fields={'target': '/'}, release_conn=False, retries=1)
+            r.release_conn()
+
+            # the pool should still contain poolsize elements
+            self.assertEqual(http.pool.qsize(), http.pool.maxsize)
+
 
 class TestRetry(HTTPDummyServerTestCase):
     def setUp(self):
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -275,6 +275,28 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         if conn:
             conn.close()
 
+    def _cleanup_conn(self, conn,  release_conn):
+        """
+        Close the connection and return it to the pool if requested. 
+        Returns ``None``.
+
+        :param conn:
+            Connection object for the current host and port as returned by
+            :meth:`._new_conn` or :meth:`._get_conn`.
+       
+        :param  release_conn:
+            If ``True`` replenish the pool with a ``None`` which will be 
+            recycled as a new connection. If ``False`` the connection
+            should be returned to the pool by the caller.
+        """
+        if conn:
+            conn.close()
+
+            if release_conn:
+                self._put_conn(None)
+
+        return None
+
     def _validate_conn(self, conn):
         """
         Called right before a request is made, after the socket is created.
@@ -409,7 +431,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         # TODO: Add optional support for socket.gethostbyname checking.
         scheme, host, port = get_host(url)
-
+        
         # Use explicit default port for comparison when none is given
         if self.port and not port:
             port = port_by_scheme.get(scheme)
@@ -568,25 +590,19 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # Close the connection. If a connection is reused on which there
             # was a Certificate error, the next request will certainly raise
             # another Certificate error.
-            if conn:
-                conn.close()
-                conn = None
+            conn = self._cleanup_conn(conn, not release_conn)
             raise SSLError(e)
 
         except SSLError:
             # Treat SSLError separately from BaseSSLError to preserve
             # traceback.
-            if conn:
-                conn.close()
-                conn = None
+            conn = self._cleanup_conn(conn, not release_conn)
             raise
 
         except (TimeoutError, HTTPException, SocketError, ConnectionError) as e:
-            if conn:
-                # Discard the connection for these exceptions. It will be
-                # be replaced during the next _get_conn() call.
-                conn.close()
-                conn = None
+            # Discard the connection for these exceptions. It will be
+            # be replaced during the next _get_conn() call.            
+            conn = self._cleanup_conn(conn, not release_conn)
 
             if isinstance(e, SocketError) and self.proxy:
                 e = ProxyError('Cannot connect to proxy.', e)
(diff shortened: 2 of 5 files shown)


shazow:
@jlatherfold @Lukasa Please review.

Also need some more tests, I think socket-level. @Lukasa up for helping? <3 <3 <3

test/with_dummyserver/test_connectionpool.py
```diff
@@ -647,6 +647,27 @@ def test_chunked_gzip(self):
 
         self.assertEqual(b'123' * 4, response.read())
 
+    def test_cleanup_on_connection_error(self):
```
shazow: Is this test even testing what we want to test?

shazow:
Alright, I snuck in some extra assertions into existing tests, good enough! Please review, then merging.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -647,6 +647,27 @@ def test_chunked_gzip(self):
...
+        Test that connections are recycled to the pool on 
+        connection errors where no http response is received.
+        '''
+        poolsize = 3
+        with HTTPConnectionPool(self.host, self.port, maxsize=poolsize, block=True) as http:
+            self.assertEqual(http.pool.qsize(), poolsize)
+
+            # force a connection error by supplying a non-existent 
+            # url. We won't get a response for this  and so the 
+            # conn won't be implicitly returned to the pool.
```
Lukasa: s/won't/will/

urllib3/connectionpool.py
```diff
@@ -409,7 +409,7 @@ def is_same_host(self, url):
 
         # TODO: Add optional support for socket.gethostbyname checking.
         scheme, host, port = get_host(url)
-
+
```
Lukasa: =O Don't you have strip trailing whitespace in neovim?

Lukasa:
Some notes, but otherwise :+1:

urllib3/connectionpool.py
shazow: I do but I like to run it manually (and forget).

test/with_dummyserver/test_connectionpool.py
shazow: Fixed.

urllib3/connectionpool.py
```diff
@@ -568,25 +568,22 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
         except (TimeoutError, HTTPException, SocketError, ConnectionError) as e:
-            if conn:
-                # Discard the connection for these exceptions. It will be
-                # be replaced during the next _get_conn() call.
-                conn.close()
-                conn = None
+            # Discard the connection for these exceptions. It will be
+            # be replaced during the next _get_conn() call.
+            conn = conn and conn.close()
+            release_conn = True
```
jlatherfold: This could result in a read error on retry when we're not preloading content. release_conn should not be changed since it's used in subsequent recursive calls to this method. If you want to do this you should use a separate (local) variable. Retries should use the same value of release_conn as supplied to this call.

urllib3/connectionpool.py
```diff
@@ -568,25 +568,22 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
             raise SSLError(e)
 
         except SSLError:
             # Treat SSLError separately from BaseSSLError to preserve
             # traceback.
-            if conn:
-                conn.close()
-                conn = None
+            conn = conn and conn.close()
+            release_conn = True
```
jlatherfold: This could result in a read error on retry when we're not preloading content. release_conn should not be changed since it's used in subsequent recursive calls to this method. If you want to do this you should use a separate (local) variable. Retries should use the same value of release_conn as supplied to this call.

urllib3/connectionpool.py
```diff
@@ -568,25 +568,22 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
             # Close the connection. If a connection is reused on which there
             # was a Certificate error, the next request will certainly raise
             # another Certificate error.
-            if conn:
-                conn.close()
-                conn = None
+            conn = conn and conn.close()
+            release_conn = True
```
jlatherfold: This could result in a read error on retry when we're not preloading content. release_conn should not be changed since it's used in subsequent recursive calls to this method. If you want to do this you should use a separate (local) variable. Retries should use the same value of release_conn as supplied to this call.

shazow: Hrm I could have sworn I left a reply but I guess it didn't take. —

How would the subsequent recursive call happen if we're raising an exception immediately after?

jlatherfold: In the case of BaseSSLError, CertificateError or SSLError you raise right after setting release_conn to True but where is that exception caught? - I think that it will miss the finally block where you do that actual release with self._put_conn so it looks like those connections will never make it back to the pool.

In the TimeoutError, HTTPException, SocketError, ConnectionError exception handler you do not raise an exception. You set release_conn to True, release it in the finally block and then make a recursive call to urlopen, this time with release_conn set to True. If that succeeds you'll then release the conn back to the pool (because release_conn is now true) before the response is read, meaning that it's possible that when response.read() happens the connection might be in use by someone else.

Apologies if I'm wrong about this cos I haven't tested it yet but just by inspecting the code it looks that way.

urllib3/connectionpool.py
jlatherfold: Apologies, my first comment was incorrect but I think there is still a bug here triggered by the next line. If we raise from here then it looks like the connection will not be returned to the pool at all.

urllib3/connectionpool.py
jlatherfold: Sorry, I've got the lines mixed up.

The BaseSSLError, CertificateError, SSLError exception handlers raise after setting release_conn to true, but that will bypass the finally block and the the conn (or None in this case) is never put back on to the queue. For the (TimeoutError, HTTPException, SocketError, ConnectionError) exception handler you don't raise (presumably because there may be retries) and you will go on to release the conn and make another call to urlopen (if retries). But you call it with release_conn=True (since you've just set it to true) - that happens if that call succeeds? You'll release the conn back to the pool before the response is read. At least that's what it looks like to me.

urllib3/connectionpool.py
shazow: On L600:

```python
         finally:
             if release_conn:
                 # Put the connection back to be reused. If the connection is
                 # expired then it will be None, which will get replaced with a
                 # fresh connection during _get_conn.
                 self._put_conn(conn)
```

This will still get called just before the exception is raised.

urllib3/connectionpool.py
shazow: > The BaseSSLError, CertificateError, SSLError exception handlers raise after setting release_conn to true, but that will bypass the finally block and the the conn (or None in this case) is never put back on to the queue. 

Why will that bypass the finally block?

> For the (TimeoutError, HTTPException, SocketError, ConnectionError) exception handler you don't raise (presumably because there may be retries) and you will go on to release the conn and make another call to urlopen (if retries). But you call it with release_conn=True (since you've just set it to true) - what happens if that call succeeds?

Yea this condition is a little hard to follow, we should try to add a test for it. Hmm.

urllib3/connectionpool.py
jlatherfold: Your are right - not sure what I was thinking - of course the finally block will be executed.

PR #648, opened 2015-06-10
[Docs] Suggest people `pip install cryptography`

talos:
When I tried to follow these instructions to use PyOpenSSL on Ubuntu 14.04, the `import urllib3.contrib.pyopenssl` line failed in Python as the `cryptography` was not already installed in my base Python.  Installing it separately (in the virtualenv in which I was doing this work) made it so these instructions worked.  I think it should be one of the requirements specified here; at worst, it will already be installed in the user's Python and be a no-op.

It was also necessary for me to `sudo apt-get install libffi-dev`, before some of the other `pip` requirements would install.  It may be worth noting that in a platform-agnostic way, or not.

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -126,7 +126,7 @@ are several limitations which are addressed by PyOpenSSL:
 To use the Python OpenSSL bindings instead, you'll need to install the required
 packages::
 
-    $ pip install pyopenssl ndg-httpsclient pyasn1
+    $ pip install pyopenssl ndg-httpsclient pyasn1 cryptography
 
 Once the packages are installed, you can tell urllib3 to switch the ssl backend
 to PyOpenSSL with :func:`~urllib3.contrib.pyopenssl.inject_into_urllib3`::
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -126,7 +126,11 @@ are several limitations which are addressed by PyOpenSSL:
 To use the Python OpenSSL bindings instead, you'll need to install the required
 packages::
 
-    $ pip install pyopenssl ndg-httpsclient pyasn1 cryptography
+    $ pip install pyopenssl ndg-httpsclient pyasn1
+
+If ``cryptography`` fails to install as a dependency, make sure you have `libffi
+<http://sourceware.org/libffi/>`_ available on your system and run
+``pip install cryptography``.
 
 Once the packages are installed, you can tell urllib3 to switch the ssl backend
 to PyOpenSSL with :func:`~urllib3.contrib.pyopenssl.inject_into_urllib3`::


Lukasa:
`cryptography` is a dependency of `pyopenssl`. How did it fail to get installed?

talos:
I see what happened here -- because `libffi-dev` wasn't installed, `pip install` failed after installing these requirements but not their dependencies (in this case, `cryptography`.)

What's problematic is that after installing the necessary package, re-running the original `pip install` did not detect the missing dependency, so `cryptography` remained missing.

I leave it at your discretion how to talk about this in the docs.  I personally think it's worth a sentence or two, as it could take a less experienced user a long time Googling stuff to try and figure out why this doesn't work.  I'm happy to write whatever is appropriate, whether that's a mention that you should install the equivalent of `libffi-dev`, or make sure that `cryptography` is installed manually, or something else.

Lukasa:
It does seem reasonable to link to the cryptography docs, at the very least.

shazow:
What was the error when it failed to install for you? If it's hard to figure out, we could add a mention for what to do when you encounter it—although it would be better if the respective dependency had friendlier errors rather than us trying to document what might go wrong with each dependency. :/

talos:
Yep, it's rough.  It's obviously not the responsibility of these docs to handle crummy error messages from other packages, but it's bad if naively following the instructions leads to a tricky-to-resolve error state without clues to resolution.

I'm not at the computer where I tried to do the install, so I can't get the error message right now.  I will be able to tomorrow.  It was not very informative.  The first error (from `pip` when `libffi` was missing) was a mess of C error messages; the second error (in Python when I tried to run `import urllib3.contrib.pyopenssl` after `libffi-dev` was installed, but when the `cryptography` dependency still wasn't installed despite my running `pip install pyopenssl ndg-httpsclient pyasn1 cryptography` again) was semi-informative, in that I was able to tell that it was trying to install something from a missing `cryptography` library.

Here's one possible way of handling it in the docs:

> If you have problems installing the `pip` dependencies, make sure [libffi](http://sourceware.org/libffi/)
> is available on your system.  You may need to run `pip install cryptography` separately if you tried
> to install the `pip` dependencies without `libffi`.

If that sounds good, I can update the PR (and remove the `cryptography` I initially added.)

shazow:
So did all of the other pip targets install successfully and cryptography did not? Re-running `pip install pyopenssl ndg-httpsclient pyasn1` would not try to install cryptography again?

I'm reluctant to add a lot of messaging for edge cases because people are already complaining that instructions in that page are too buried. Maybe something short like

> "If the install fails, make sure you have [libffi](http://sourceware.org/libffi/) available on your system."

talos:
Yes, somehow the other pip targets installed, even though `cryptography` did not.  Simply re-running the `pip install` line did not install `cryptography` even after `libffi` was installed in the system.

I would definitely note that people should try to `pip install cryptography` in that note, due to pip's sub-optimal behavior in this case.

shazow:
Fair, maybe something like:

> If `cryptography` fails to install as a dependency, make sure you have [libffi](http://sourceware.org/libffi/) available on your system and run `pip install cryptograph`."

talos:
:+1: I'm going to change it to that text.

shazow:
Excellent, thank you for improving our documentation, that's super valuable for our community! You're welcome to add yourself to our CONTRIBUTORS.txt :)

talos:
How generous!  Done: https://github.com/shazow/urllib3/pull/652

PR #653, opened 2015-06-22
Get rid of any trailing whitespace on hostname

laconic75:
Add port.rstrip to get rid of any trialing whitespace on the hostname.
Any trialing whitespace will cause port.isdigit() to fail. This can be a
very hard bug to find. Many hostnames are taken from files and it is a
very simple thing for someone to leave a trailing space when typing in a
hostname.

### DIFF
--- a/urllib3/util/url.py
+++ b/urllib3/util/url.py
@@ -177,6 +177,7 @@ def parse_url(url):
     # Port
     if ':' in url:
         _host, port = url.split(':', 1)
+	port = port.rstrip() # Drop any trailing whitespace
 
         if not host:
             host = _host


shazow:
Wouldn't `port = int(port)` raise an error and you'd notice? What makes this harder to debug than a rogue space in any other part of the url?

laconic75:
Because the port is at the end of the URL. If you have these two URL's in a file
```
http://myserver1:8080
http://myserver2:8080 
```
they look very similar. If you try to read each line and parse it with parse_url as the function is currently, one will work and the other will fail. In my case, I was using the requests library behind a proxy. Then requests was in turn using urllib3 to parse my http_proxy environmental variable I had set 3 months ago. I then get an error
```
Parse Error:  Can't parse 10.10.10.1:8080 
```
 I found it eventually, but I couldn't help but thinking that adding one line stripping out that space would have made my life much simpler today.

Lukasa:
I think this is a reasonable change, but I don't think it's sufficiently broad.

Given that leading and trailing whitespace are both meaningless in URLs, we should just strip all whitespace from the URL before working with it, both leading and trailing. Let's not fix just this instance of the problem, let's fix all instances of the problem.

shazow:
> Parse Error:  Can't parse 10.10.10.1:8080 

Where is that error message coming from? Seems like surrounding the full value in quotes would have been helpful also.

> I think this is a reasonable change, but I don't think it's sufficiently broad.

I'd prefer for stripping strings on input ideally should be done a layer above our parser, but I could be convinced. That is strictly removing functionality, and somebody out there is going to want to force a space somehow, sigh. :P

Lukasa:
Oh sure, we should do it at the `urlopen` layer, or whatever top layer of the urllib3 code handles the input. =)

Alternatively, requests could do it! ;)

laconic75:
>Where is that error message coming from?

It comes from line 187

         if port:
             # If given, ports must be integers.
             if not port.isdigit():
                 raise LocationParseError(url)

The error I gave isn't exactly correct. Just as close as I could remember while typing just before going to bed. 

>I think this is a reasonable change, but I don't think it's sufficiently broad.

You could do 

    url=url.strip()

at the beginning of the function, though I don't think leading spaces would be as common or as hard to find, and may not cause an error. It probably would be good practice just to get rid of any leading and trailing white space though.

shazow:
I'm +1 to putting quotes around the location variable in [LocationParseError](https://github.com/shazow/urllib3/blob/master/urllib3/exceptions.py#L131). That should resolve @laconic75's difficulty in discovering the problem, right?

I'm still reluctant to strip spacing inside of parse_url(...) because that strictly removes functionality that somebody might depend on (or maybe not). I'm open to stripping spaces in helpers layered above, but it sounds like @laconic75 was using it directly to parse environment variables, so that won't help.

Either way, I'd like to add a test that exercises spaces before/after the input string. :)

laconic75:
I was trying to some research on whitespaces in URL's and here are some references on I found. 
* [RFC3986 Appendix A] (https://tools.ietf.org/html/rfc3986#appendix-A)
* [RFC3986 Appendix C] (https://tools.ietf.org/html/rfc3986#appendix-C)
* [Stack Overflow -- Spaces in URL's] (http://stackoverflow.com/questions/5442658/spaces-in-urls)
* [Stack Overflow -- Is a URL allowed to contain a space?] (http://stackoverflow.com/questions/497908/is-a-url-allowed-to-contain-a-space)

To summarize, if I understand it correctly, a URL should not contain a space, and you can use whitespace to delimit a URL.

shazow:
@sigmavirus24 Hey Mr RFC, thoughts on this?

sigmavirus24:
@laconic75 is correct that URIs aren't allowed to contain a space. That said, I would expect some percentage of users to attempt to do something like

```py
import urllib3

http = urllib3.PoolManager()
resp = http.request('GET', 'https://httpbin.org/get?foo=bar baz')
```

Expecting that space to become either a `+` or `%20`.

@laconic75 is also correct that the RFC cites examples of how people delimit URIs in contexts that don't have defined mark-up for them. The first two examples are `"http://example.com"` and `<http://example.com>` and the last discusses using whitespace ` http://example.com `. That said, those contexts are not urllib3's responsibility to discern. In other words, we shouldn't start accepting

```py
import urllib3

http = urllib3.PoolManager()
resp = http.request('GET', '"https://httpbin.org/get?foo=bar baz"')
resp = http.request('GET', '<https://httpbin.org/get?foo=bar baz>')
```

So why are we talking about accepting

```py
resp = http.request('GET', ' https://httpbin.org/get?foo=bar baz )
resp = http.request('GET', 'https://httpbin.org/get?foo=bar baz ')
resp = http.request('GET', ' https://httpbin.org/get?foo=bar baz ')
```

Those should all be exceptions because while we can *guess* what the user wants us to do, I don't think we'll ever be 100% certain. And yes, I think quoting the parts that are causing us to error is the most user-friendly way of doing it.

shazow:
Yea, I'm not comfortable doing .strip() on input here. There's no good reason for why we should do it here over every other string input we accept all over urllib3. Let's stick to just improving the exception message. @laconic75 Does that sound sensible for now?

sigmavirus24:
This seems reasonable to me since the crux of the issue was:

> I found it eventually, but I couldn't help but thinking that adding one line stripping out that space would have made my life much simpler today.

Having a better exception would have made life much simpler too

laconic75:
That sounds good. I'll close the request.

shazow:
@laconic75 Would you like to do a PR that fixes the exception message?

PR #657, opened 2015-06-25
Fix usage headers argument for PoolManager

Disassem:
Argument headers for PoolManager doesn't work as it should ("Headers to include with all requests, unless other headers are given explicitly.") for request method.

Sample:

```python
import urllib3
manager = urllib3.PoolManager(headers = urllib3.make_headers(accept_encoding = True, user_agent = 'urllib3'))
request = manager.request('GET', 'http://httpbin.org/headers')
print request.data
```

Result:

```json
{
  "headers": {
    "Accept-Encoding": "identity",
    "Host": "httpbin.org"
  }
}
```

### DIFF
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -130,6 +130,14 @@ class TestPoolManager(HTTPDummyServerTestCase):
     def test_headers(self):
         http = PoolManager(headers={'Foo': 'bar'})
 
+        r = http.request('GET', '%s/headers' % self.base_url)
+        returned_headers = json.loads(r.data.decode())
+        self.assertEqual(returned_headers.get('Foo'), 'bar')
+
+        r = http.request('POST', '%s/headers' % self.base_url)
+        returned_headers = json.loads(r.data.decode())
+        self.assertEqual(returned_headers.get('Foo'), 'bar')
+        
         r = http.request_encode_url('GET', '%s/headers' % self.base_url)
         returned_headers = json.loads(r.data.decode())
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks for reporting this!

Amusingly it works fine if you send anything that uses a request body rather than URL parameters, e.g. a POST request.

The bug is in `urllib3.request.RequestMethods.request_encode_url`, which does not take note of the headers default in PoolManager.

Disassem:
Is it better now?

Lukasa:
Sorry, no, you should follow the example from the method directly below `request_encode_url`:

```python
         if headers is None:
             headers = self.headers
```

Lukasa:
You will also need to add a `headers` keyword argument there, but it should definitely have one anyway.

Disassem:
Like this?

Lukasa:
\o/ Exactly like that! Awesome!


So, we'll need a test for this as well. Let me know if you don't feel like writing one, I can do it instead. =)

Disassem:
Add test.

Lukasa:
Alright, this looks great to me. @shazow want to do a once-over and merge if you're happy?

Disassem:
Yes, of course.
Also may be change version in urllib3.__init__.py? It is still 1.10.2.

shazow:
Looks great, thank you @Disassem! You're welcome to add yourself to the CONTRIBUTORS.txt, btw. :)

And thank you @Lukasa for sheparding this. :)

> Also may be change version in urllib3.__init__.py? It is still 1.10.2.

Ah that's something I screwed up a couple of releases ago, I'll fix that. Thanks for noticing!

PR #660, opened 2015-06-27
close response on connection errors during read

jlatherfold:
I've added a couple of test cases for this in test_socketlevel.py that just compare the poolsize before and after calls to read (that fail). Without the changes to request.py these tests will fail.
I also moved the finally block to above the decoding code, which means we get to release the connection a little bit sooner. If there's an exception during self._fp.read that code will be skipped anyway.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -297,6 +297,64 @@ class TestSocketClosing(SocketDummyServerTestCase):
         self.assertEqual(response.status, 200)
         self.assertEqual(response.data, b'foo')
 
+    def test_connection_cleanup_on_read_timeout(self):
+        timed_out = Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+            buf = b''
+            body = 'Hi'
+            while not buf.endswith(b'\r\n\r\n'):
+                buf = sock.recv(65536)
+            sock.send(('HTTP/1.1 200 OK\r\n'
+                       'Content-Type: text/plain\r\n'
+                       'Content-Length: %d\r\n'
+                       '\r\n' % len(body)).encode('utf-8'))
+
+            timed_out.wait()
+            sock.send(body.encode('utf-8'))
+            sock.close()
+
+        self._start_server(socket_handler)
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            poolsize = pool.pool.qsize()
+            response = pool.urlopen('GET', '/', retries=0, preload_content=False,
+                                    timeout=Timeout(connect=1, read=0.001))
+            try:
+                self.assertRaises(ReadTimeoutError, response.read)
+                self.assertEqual(poolsize, pool.pool.qsize())
+            finally:
+                timed_out.set()
+
+    def test_connection_cleanup_on_protocol_error_during_read(self):
+        body = 'Response'
+        partial_body = body[:2]
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            # Consume request
+            buf = b''
+            while not buf.endswith(b'\r\n\r\n'):
+                buf = sock.recv(65536)
+
+            # Send partial response and close socket.
+            sock.send((
+                'HTTP/1.1 200 OK\r\n'
+                'Content-Type: text/plain\r\n'
+                'Content-Length: %d\r\n'
+                '\r\n'
+                '%s' % (len(body), partial_body)).encode('utf-8')
+            )
+            sock.close()
+
+        self._start_server(socket_handler)
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            poolsize = pool.pool.qsize()
+            response = pool.request('GET', '/', retries=0, preload_content=False)
+
+            self.assertRaises(ProtocolError, response.read)
+            self.assertEqual(poolsize, pool.pool.qsize())
 
 
 class TestProxyManager(SocketDummyServerTestCase):
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -312,7 +312,6 @@ class TestSocketClosing(SocketDummyServerTestCase):
                        '\r\n' % len(body)).encode('utf-8'))
 
             timed_out.wait()
-            sock.send(body.encode('utf-8'))
             sock.close()
 
         self._start_server(socket_handler)
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -234,57 +234,49 @@ class HTTPResponse(io.IOBase):
         data = None
 
         try:
-            if amt is None:
-                # cStringIO doesn't like amt=None
-                data = self._fp.read()
-                flush_decoder = True
-            else:
-                cache_content = False
-                data = self._fp.read(amt)
-                if amt != 0 and not data:  # Platform-specific: Buggy versions of Python.
-                    # Close the connection when no data is returned
-                    #
-                    # This is redundant to what httplib/http.client _should_
-                    # already do.  However, versions of python released before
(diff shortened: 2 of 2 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -297,6 +297,64 @@ def socket_handler(listener):
...
+            body = 'Hi'
+            while not buf.endswith(b'\r\n\r\n'):
+                buf = sock.recv(65536)
+            sock.send(('HTTP/1.1 200 OK\r\n'
+                       'Content-Type: text/plain\r\n'
+                       'Content-Length: %d\r\n'
+                       '\r\n' % len(body)).encode('utf-8'))
+
+            timed_out.wait()
+            sock.send(body.encode('utf-8'))
```
Lukasa: This could throw an exception if the socket got closed. Might be worth doing `try...except` here.

Lukasa:
One small note.

urllib3/response.py
```diff
@@ -231,45 +231,65 @@ def read(self, amt=None, decode_content=None, cache_content=False):
...
-                    raise
-
-                raise ReadTimeoutError(self._pool, None, 'Read timed out.')
-
-            except HTTPException as e:
-                # This includes IncompleteRead.
-                raise ProtocolError('Connection broken: %r' % e, e)
 
+        except SocketTimeout:
+            # the response may not be closed but we're not going to use it anymore
```
shazow: Please write comments in consistent style with the rest of the codebase (sentence-case: starts with a capital, ends with a period).

urllib3/response.py
```diff
@@ -231,45 +231,65 @@ def read(self, amt=None, decode_content=None, cache_content=False):
...
+            # so close it now to ensure that the connection is released back to the pool
+            if self._original_response and not self._original_response.isclosed():
```
shazow: Rather than repeating this block 3 times, maybe we'd be better off factoring this whole try/except block into a helper (e.g. `self._read(...)`)? Then we can wrap it with a single exception handler for closing the original response. I think that would be cleaner and easier to maintain.

test/with_dummyserver/test_socketlevel.py
jlatherfold: It's copied from another test, but I will follow your suggestion

urllib3/response.py
jlatherfold: How about merging all the exceptions in to a single handler, or just catching Exception - after all we may have missed something which would mean we 'lose' the connection on an unhandled exception if the response is still open? It's just a thought - we don't seem to do much different in the handlers apart from rethrow either timeout or protocol error and that could be determined in a single handler by querying the exception object.

jlatherfold: Another thought - is there a way in python to determine in the finally block whether an exception has been thrown? then you could do something like:

if self._original_response and not self._original_response.isclosed() or exception_thrown:

shazow: > How about merging all the exceptions in to a single handler, or just catching Exception - after all we may have missed something which would mean we 'lose' the connection on an unhandled exception if the response is still open? 

You can try, but I don't think that will end up much cleaner. You'll still need to switch on the exception types to re-raise the appropriate internal type.

> Another thought - is there a way in python to determine in the finally block whether an exception has been thrown? then you could do something like:

You could do that by setting a boolean (e.g. `close_response = True` inside every exception block, similar to what we do with `release_conn`).

I suspect refactoring that block into a helper and surrounding the call line with another try/except would be best but I'm open to exploring other variants to see how they end up looking. :) I'd prefer not to leave it as-is, though.

jlatherfold:
First, I didn't change the test - if the socket closes unexpectedly the test will fail and that's probably what we want anyway. Second, instead of refactoring the read code out to a different method I just wrapped it in an outer try (similar to the original code). Although try:try: looks a bit odd we at least get to catch all exceptions this time so nothing gets past us and we get rid of the duplicate code.
On a slightly different note apologies on the branch-name - this is issue 660 not 659.

urllib3/response.py
```diff
@@ -269,7 +270,18 @@ def read(self, amt=None, decode_content=None, cache_content=False):
...
+            # so close it now to ensure that the connection is released back to the pool.
+            if self._original_response and not self._original_response.isclosed():
+                self._original_response.close()
 
+            raise
+        finally:
+            if self._original_response and self._original_response.isclosed():
+                self.release_conn()
+
+        if data:
```
shazow: How can the execution path make it here without `data` being set?

Won't it just escape via `raise` on L279?

urllib3/response.py
```diff
@@ -269,7 +270,18 @@ def read(self, amt=None, decode_content=None, cache_content=False):
             except HTTPException as e:
                 # This includes IncompleteRead.
                 raise ProtocolError('Connection broken: %r' % e, e)
+        except Exception:
```
shazow: I think this can be

```diff
-        except Exception:
+        except:
```

@Lukasa Is that correct/idiomatic these days? (Too much Go in my brain.)

urllib3/response.py
jlatherfold: If no exception is raised during the fp.read opertion, line 279 will never be executed. If we do get an exception it will be caught by one of the inner exception handlers (SocketTimeout, ReadSSLError, HTTPException). All of these re-raise and that will be caught at line 273. It's here that we close the response (but also, this block catches any uncaught exceptions raised in the first block) and then re-raise. The finally block is then executed and we're gone.
If not exception is raised and everything goes according to plan data is set at either line 240 or 244, the finally block is executed and we continue on to decode the data. At least that's what it looks like to me and I'm counting on the test-suite to pick up any blindingly obvious errors with the logic.

jlatherfold: Re-raising at line 279 doesn't change the way the original code behaved. I believe you want to bail out at this point and deal with the exception higher up the callstack. There's not much point running the decoding routine if the read operation has failed.

urllib3/response.py
jlatherfold: I thought about just except: but read in the python docs that it wasn't recommended. I did a search in your code and couldn't find it anywhere so opted for this. Don't all exceptions derive from Exception?

Lukasa:
> First, I didn't change the test - if the socket closes unexpectedly the test will fail and that's probably what we want anyway.

I don't think the socket is getting closed unexpectedly, I think it's getting closed *expectedly*. You're timing the request out, which causes us to explicitly close the socket. This means there's a race between the close of the socket and the send on the other side that could cause an exception to be raised in slow architectures. We want to allow that socket send to fail without blowing the test up.

Unless I've fundamentally misunderstood something here, which I don't rule out.

urllib3/response.py
Lukasa: **Do not use `except:`**.

Not all exceptions derive from `Exception`, and that's why we use `except Exception`. Python implements some very important things as exceptions, like Ctrl-C-ing a program (`KeyboardInterrupt`) and calling `sys.exit()` (`SystemExit`). Those *are* exceptions and so will be caught by `except:`, but they do *not* inherit from `Exception` (they inherit from `BaseException`). You want to let them bubble up, so `except Exception:` is absolutely the idiomatic thing to do.

jlatherfold:
I see - yes you're right. There's no need for the second 'send' (copy paste job) in this case, it can be removed.

urllib3/response.py
shazow: @Lukasa Ah yes, thank you!

shazow:
Looks good to me, thank you @jlatherfold!

shazow:
@jlatherfold Btw, you're welcome to add yourself to the CONTRIBUTORS.txt file. :)

jlatherfold:
Happy to help.
How should I do that - another pull request just on contributors.txt?

shazow:
That works. :)

PR #662, opened 2015-06-29
Fix attribute name in docsting

Disassem:


shazow:
Thank you!! :custard:

PR #664, opened 2015-06-30
App Engine Connection Pool

jonparrott:
This is currently *very* rough and early. I want to make sure I'm on the right track before continuing.

Outstanding questions:

1. What functionality should be covered by the connection pool? (retries, etc.)
2. Should I just inherit from TestConnectionPool and null out the tests that aren't applicable to GAE instead of duplicating code?
3. How should this connection class be wired in? On GAE Sandbox environments w/o sockets, I assume this should be default.
4. What exact edge cases should this class be aware; e.g, where does URLFetch behavior differ from what's expected of httplib/sockets? Do the tests already cover this?

[*shazow edit: Fixes #664]*

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -1,0 +1,222 @@
+import logging
+import os
+import warnings
+
+from ..exceptions import (
+    HTTPError,
+    HTTPWarning,
+    MaxRetryError,
+    ProtocolError,
+    TimeoutError,
+    SSLError
+)
+
+from ..packages.six import BytesIO
+from ..request import RequestMethods
+from ..response import HTTPResponse
+from ..util.timeout import Timeout
+from ..util.retry import Retry
+
+try:
+    from google.appengine.api import urlfetch
+except ImportError:
+    urlfetch = None
+
+
+log = logging.getLogger(__name__)
+
+
+class AppEnginePlatformWarning(HTTPWarning):
+    pass
+
+
+class AppEnginePlatformError(HTTPError):
+    pass
+
+
+class AppEngineManager(RequestMethods):
+    """
+    Connection manager for Google App Engine sandbox applications.
+
+    This manager uses the URLFetch service directly instead of using the
+    emulated httplib, and is subject to URLFetch limitations as described in
+    the App Engine documentation here:
+
+        https://cloud.google.com/appengine/docs/python/urlfetch
+
+    Notably it will raise an AppEnginePlatformError if:
+        * URLFetch is not available.
+        * If you attempt to use this on GAEv2 (Managed VMs), as full socket
+          support is available.
+        * If a request size is more than 10 megabytes.
+        * If a response size is more than 32 megabtyes.
+        * If you use an unsupported request method such as OPTIONS.
+
+    Beyond those cases, it will raise normal urllib3 errors.
+    """
+
+    def __init__(self, headers=None, retries=None, validate_certificate=True):
+        if not urlfetch:
+            raise AppEnginePlatformError(
+                "URLFetch is not available in this environment.")
+
+        if is_prod_appengine_v2():
+            raise AppEnginePlatformError(
+                "Use normal urllib3.PoolManager instead of AppEngineManager"
+                "on Managed VMs, as using URLFetch is not necessary in "
+                "this environment.")
+
+        warnings.warn(
+            "urllib3 is using URLFetch on Google App Engine sandbox instead "
+            "of sockets. To use sockets directly instead of URLFetch see "
+            "https://urllib3.readthedocs.org/en/latest/contrib.html.",
+            AppEnginePlatformWarning)
+
+        RequestMethods.__init__(self, headers)
+        self.validate_certificate = validate_certificate
+
+        self.retries = retries or Retry.DEFAULT
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # Return False to re-raise any potential exceptions
+        return False
+
+    def urlopen(self, method, url, body=None, headers=None,
+                retries=None, redirect=True, timeout=Timeout.DEFAULT_TIMEOUT,
+                **response_kw):
+
+        retries = self._get_retries(retries, redirect)
+
+        try:
+            response = urlfetch.fetch(
+                url,
+                payload=body,
+                method=method,
+                headers=headers or {},
+                allow_truncated=False,
+                follow_redirects=(
+                    redirect and
+                    retries.redirect != 0 and
+                    retries.total),
+                deadline=self._get_absolute_timeout(timeout),
+                validate_certificate=self.validate_certificate,
+            )
+        except urlfetch.DeadlineExceededError as e:
+            raise TimeoutError(self, e)
+
+        except urlfetch.InvalidURLError as e:
+            if 'too large' in e.message:
+                raise AppEnginePlatformError(
+                    "URLFetch request too large, URLFetch only "
+                    "supports requests up to 10mb in size.", e)
+            raise ProtocolError(e)
+
+        except urlfetch.DownloadError as e:
+            if 'Too many redirects' in e.message:
+                raise MaxRetryError(self, url, reason=e)
+            raise ProtocolError(e)
+
+        except urlfetch.ResponseTooLargeError as e:
+            raise AppEnginePlatformError(
+                "URLFetch response too large, URLFetch only supports"
+                "responses up to 32mb in size.", e)
+
+        except urlfetch.SSLCertificateError as e:
+            raise SSLError(e)
+
+        except urlfetch.InvalidMethodError as e:
+            raise AppEnginePlatformError(
+                "URLFetch does not support method: %s" % method, e)
+
+        http_response = self._urlfetch_response_to_http_response(
+            response, **response_kw)
+
+        # Check for redirect response
+        if (http_response.get_redirect_location() and
+                retries.raise_on_redirect and redirect):
+            raise MaxRetryError(self, url, "too many redirects")
+
+        # Check if we should retry the HTTP response.
+        if retries.is_forced_retry(method, status_code=http_response.status):
+            retries = retries.increment(
+                method, url, response=http_response, _pool=self)
+            log.info("Forced retry: %s" % url)
+            retries.sleep()
+            return self.urlopen(
+                method, url,
+                body=body, headers=headers,
+                retries=retries, redirect=redirect,
+                timeout=timeout, **response_kw)
+
+        return http_response
+
+    def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
+
+        if is_prod_appengine_v1():
+            # Production GAE handles deflate encoding automatically, but does
+            # not remove the encoding header.
+            content_encoding = urlfetch_resp.headers.get('content-encoding')
+
+            if content_encoding == 'deflate':
+                del urlfetch_resp.headers['content-encoding']
+
+        return HTTPResponse(
+            # In order for decoding to work, we must present the content as
+            # a file-like object.
+            body=BytesIO(urlfetch_resp.content),
+            headers=urlfetch_resp.headers,
+            status=urlfetch_resp.status_code,
+            **response_kw
+        )
+
+    def _get_absolute_timeout(self, timeout):
+        if timeout is Timeout.DEFAULT_TIMEOUT:
+            return 5  # 5s is the default timeout for URLFetch.
+        if isinstance(timeout, Timeout):
+            if not timeout.read is timeout.connect:
+                warnings.warn(
+                    "URLFetch does not support granular timeout settings, "
+                    "reverting to total timeout.", AppEnginePlatformWarning)
+            return timeout.total
+        return timeout
+
+    def _get_retries(self, retries, redirect):
+        if not isinstance(retries, Retry):
+            retries = Retry.from_int(
+                retries, redirect=redirect, default=self.retries)
+
+        if retries.connect or retries.read or retries.redirect:
+            warnings.warn(
+                "URLFetch only supports total retries and does not "
+                "recognize connect, read, or redirect retry parameters.",
+                AppEnginePlatformWarning)
+
+        return retries
+
+
+def is_appengine():
+    return (is_local_appengine() or
+            is_prod_appengine_v1() or
+            is_prod_appengine_v2())
+
+
+def is_appengine_sandbox():
+    return is_appengine() and not is_prod_appengine_v2()
+
+
+def is_local_appengine():
+    return ('APPENGINE_RUNTIME' in os.environ and
+            'Development/' in os.environ['SERVER_SOFTWARE'])
+
+
+def is_prod_appengine_v1():
+    return ('APPENGINE_RUNTIME' in os.environ and
+            'Google App Engine/' in os.environ['SERVER_SOFTWARE'] and
+            not is_prod_appengine_v2())
+
+
+def is_prod_appengine_v2():
+    return os.environ.get('GAE_VM', False) == 'true'
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -12,3 +12,41 @@ SNI-support for Python 2
 ------------------------
 
 .. automodule:: urllib3.contrib.pyopenssl
+
+
+.. _gae:
+
+Google App Engine 
+-----------------
+
+The :mod:`urllib3.contrib.appengine` module provides a pool manager that
+uses Google App Engine's `URLFetch Service <https://cloud.google.com/appengine/docs/python/urlfetch>`_.
+
+Example usage::
+
+    from urllib3 import PoolManager
+    from urllib3.contrib.appengine import AppEngineManager, is_appengine_sandbox
+
+    # This substitution will be done automagically once appengine code
+    # graduates from the contrib module.
+    if is_appengine_sandbox():
+        # AppEngineManager uses AppEngine's URLFetch API behind the scenes
+        http = AppEngineManager()
+    else:
+        # PoolManager uses a socket-level API behind the scenes
+        http = PoolManager()
+
+    # The client API should be consistent across managers, though some features are not available
+    # in URLFetch and you'll get warnings when you try to use them (like granular timeouts).
+    r = http.request('GET', 'https://google.com/')
+
+
+There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/#Python_Quotas_and_limits>`_ to the URLFetch service and it may not be the best choice for your application. App Engine provides three options for urllib3 users:
+
+1. You can use :class:`AppEngineManager` with URLFetch. URLFetch is cost-effective in many circumstances as long as your usage is within the limitations.
+2. You can use a normal :class:`PoolManager` by enabling sockets. Sockets also have `limitations and restrictions <https://cloud.google.com/appengine/docs/python/sockets/#limitations-and-restrictions>`_ and have a lower free quota than URLFetch. To use sockets, be sure to specify the following in your ``app.yaml``::
+    
+    env_variables:
+        GAE_USE_SOCKETS_HTTPLIB : 'true'
+
+3. If you are using `Managed VMs <https://cloud.google.com/appengine/docs/managed-vms/>`_, you can use the standard :class:`PoolManager` without any configuration or special environment variables.
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -73,6 +73,10 @@ class TestingApp(RequestHandler):
         """ Handle OPTIONS requests """
         self._call_method()
 
+    def head(self):
+        """ Handle HEAD requests """
+        self._call_method()
+
     def _call_method(self):
         """ Call the correct method in this class based on the incoming URI """
         req = self.request
@@ -231,6 +235,13 @@ class TestingApp(RequestHandler):
         chunks.append(compressor.flush())
 
         return Response(chunks, headers=[('Content-Encoding', 'gzip')])
+
+    def nbytes(self, request):
+        length = int(request.params.get('length'))
+        data = b'1' * length
+        return Response(
+            data,
+            headers=[('Content-Type', 'application/octet-stream')])
 
     def shutdown(self, request):
         sys.exit()
--- a/test/contrib/test_gae_manager.py
+++ b/test/contrib/test_gae_manager.py
@@ -1,0 +1,185 @@
+import unittest
+
+from dummyserver.testcase import HTTPSDummyServerTestCase
+from nose.plugins.skip import SkipTest
+
+try:
+    from google.appengine.api import urlfetch
+    (urlfetch)
+except ImportError:
+    raise SkipTest("App Engine SDK not available.")
+
+from urllib3.contrib.appengine import AppEngineManager, AppEnginePlatformError
+from urllib3.exceptions import (
+    TimeoutError,
+    ProtocolError,
+    SSLError)
+from urllib3.util.url import Url
+from urllib3.util.retry import Retry
+
+from test.with_dummyserver.test_connectionpool import (
+    TestConnectionPool, TestRetry)
+
+
+# Prevent nose from running these test.
+TestConnectionPool.__test__ = False
+TestRetry.__test__ = False
+
+
+# This class is used so we can re-use the tests from the connection pool.
+# It proxies all requests to the manager.
+class MockPool(object):
+    def __init__(self, host, port, manager, scheme='http'):
+        self.host = host
+        self.port = port
+        self.manager = manager
+        self.scheme = scheme
+
+    def request(self, method, url, *args, **kwargs):
+        url = self._absolute_url(url)
+        return self.manager.request(method, url, *args, **kwargs)
+
+    def urlopen(self, method, url, *args, **kwargs):
+        url = self._absolute_url(url)
+        return self.manager.urlopen(method, url, *args, **kwargs)
+
+    def _absolute_url(self, path):
+        return Url(
+            scheme=self.scheme,
+            host=self.host,
+            port=self.port,
+            path=path).url
+
+
+# Note that this doesn't run in the sandbox, it only runs with the URLFetch
+# API stub enabled. There's no need to enable the sandbox as we know for a fact
+# that URLFetch is used by the connection manager.
+class TestGAEConnectionManager(TestConnectionPool):
+    __test__ = True
+
+    # Magic class variable that tells NoseGAE to enable the URLFetch stub.
+    nosegae_urlfetch = True
+
+    def setUp(self):
+        self.manager = AppEngineManager()
+        self.pool = MockPool(self.host, self.port, self.manager)
+
+    # Tests specific to AppEngineManager
+
+    def test_exceptions(self):
+        # DeadlineExceededError -> TimeoutError
+        self.assertRaises(
+            TimeoutError,
+            self.pool.request,
+            'GET',
+            '/sleep?seconds=0.005',
+            timeout=0.001)
+
+        # InvalidURLError -> ProtocolError
+        self.assertRaises(
+            ProtocolError,
+            self.manager.request,
+            'GET',
+            'ftp://invalid/url')
+
+        # DownloadError -> ProtocolError
+        self.assertRaises(
+            ProtocolError,
+            self.manager.request,
+            'GET',
+            'http://0.0.0.0')
+
+        # ResponseTooLargeError -> AppEnginePlatformError
+        self.assertRaises(
+            AppEnginePlatformError,
+            self.pool.request,
+            'GET',
+            '/nbytes?length=33554433')  # One byte over 32 megabtyes.
+
+        # URLFetch reports the request too large error as a InvalidURLError,
+        # which maps to a AppEnginePlatformError.
+        body = b'1' * 10485761  # One byte over 10 megabytes.
+        self.assertRaises(
+            AppEnginePlatformError,
+            self.manager.request,
+            'POST',
+            '/',
+            body=body)
+
+    # Re-used tests below this line.
+    # Subsumed tests
+    test_timeout_float = None  # Covered by test_exceptions.
+
+    # Non-applicable tests
+    test_conn_closed = None
+    test_nagle = None
+    test_socket_options = None
+    test_disable_default_socket_options = None
+    test_defaults_are_applied = None
+    test_tunnel = None
+    test_keepalive = None
+    test_keepalive_close = None
+    test_connection_count = None
+    test_connection_count_bigpool = None
+    test_for_double_release = None
+    test_release_conn_parameter = None
+    test_stream_keepalive = None
+    test_cleanup_on_connection_error = None
+
+    # Tests that should likely be modified for appengine specific stuff
+    test_timeout = None
+    test_connect_timeout = None
+    test_connection_error_retries = None
+    test_total_timeout = None
+    test_none_total_applies_connect = None
+    test_timeout_success = None
+    test_source_address_error = None
+    test_bad_connect = None
+    test_partial_response = None
+    test_dns_error = None
+
+
+class TestGAEConnectionManagerWithSSL(HTTPSDummyServerTestCase):
+    nosegae_urlfetch = True
+
+    def setUp(self):
+        self.manager = AppEngineManager()
+        self.pool = MockPool(self.host, self.port, self.manager, 'https')
+
+    def test_exceptions(self):
+        # SSLCertificateError -> SSLError
+        # SSLError is raised with dummyserver because URLFetch doesn't allow
+        # self-signed certs.
+        self.assertRaises(
+            SSLError,
+            self.pool.request,
+            'GET',
+            '/')
+
+
+class TestGAERetry(TestRetry):
+    __test__ = True
+
+    # Magic class variable that tells NoseGAE to enable the URLFetch stub.
+    nosegae_urlfetch = True
+
+    def setUp(self):
+        self.manager = AppEngineManager()
+        self.pool = MockPool(self.host, self.port, self.manager)
+
+    def test_default_method_whitelist_retried(self):
+        """ urllib3 should retry methods in the default method whitelist """
+        retry = Retry(total=1, status_forcelist=[418])
+        # Use HEAD instead of OPTIONS, as URLFetch doesn't support OPTIONS
+        resp = self.pool.request(
+            'HEAD', '/successful_retry',
+            headers={'test-name': 'test_default_whitelist'},
+            retries=retry)
+        self.assertEqual(resp.status, 200)
+
+    #test_max_retry = None
+    #test_disabled_retry = None
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/tox.ini
+++ b/tox.ini
@@ -18,6 +18,12 @@ commands=
         -c {toxinidir}/test/appengine/nose.cfg \
         test/appengine \
         []
+    # For now, this test is ran seperately because the sandbox activation in
+    # the app engine tests will blow up everything.
+    nosetests \
+        -c {toxinidir}/test/appengine/nose.cfg \
+        test/contrib/test_gae_manager.py \
+        []
 setenv =
     PYTHONPATH={env:GAE_PYTHONPATH:}
     {[testenv]setenv}


test/appengine/test_gae_connectionpool.py
```diff
@@ -0,0 +1,193 @@
...
+from nose.plugins.skip import SkipTest
+
+try:
+    from google.appengine.api import urlfetch
+except ImportError:
+    raise SkipTest("App Engine SDK not available.")
+
+try:
+    from urllib.parse import urlencode
+except:
```
t-8ch: `except ImportError`?

test/appengine/test_gae_connectionpool.py
```diff
@@ -0,0 +1,193 @@
+import unittest
+import logging
+import sys
+
+from nose.plugins.skip import SkipTest
+
+try:
+    from google.appengine.api import urlfetch
+except ImportError:
+    raise SkipTest("App Engine SDK not available.")
```
t-8ch: Isn't this file only executed if we explictly want to test GAE?
If so we should blow up.

test/appengine/test_gae_connectionpool.py
```diff
@@ -0,0 +1,193 @@
...
+)
+
+from urllib3.packages.six import b, u
+
+from dummyserver.testcase import HTTPDummyServerTestCase
+
+
+# Note that this doesn't run in the sandbox, it only runs with the URLFetch
+# API stub enabled. There's no need to enable the sandbox as we know for a fact
+# that URLFetch is using used by the connection pool.
```
t-8ch: s/used by//

urllib3/gae_connectionpool.py
```diff
@@ -0,0 +1,96 @@
...
+    SSLError,
+    TimeoutError,
+    InsecureRequestWarning,
+)
+
+from .packages.six import BytesIO
+from .request import RequestMethods
+from .response import HTTPResponse
+from .util.timeout import Timeout
+from .util.url import Url
```
t-8ch: A few of the imports seem to be superfluous

urllib3/gae_connectionpool.py
```diff
@@ -0,0 +1,96 @@
...
+        if not urlfetch:
+            raise EnvironmentError(
+                "URLFetch is not available in this environment.")
+
+        ConnectionPool.__init__(self, host, port)
+        RequestMethods.__init__(self, headers)
+
+        # URLFetch doesn't support fine-grained control over timeouts.
+        if isinstance(timeout, Timeout):
+            timeout = timeout.total
```
t-8ch: Maybe we want to have a `PlatformWarning` for this? @shazow

shazow:
@Lukasa @sigmavirus24 Would love your thoughts on some of the open questions inline:

> What functionality should be covered by the connection pool? (retries, etc.)

Hm, I've been thinking about this. I think it may make more sense to have the AppEngine URLFetch driver be at the Manager-level (see poolmanager.py) rather than ConnectionPool-level because it doesn't make sense for our PoolManager/ProxyManager/etc to manage pools of GAEConnectionPool, right?

This should make things a fair bit more straight-forward. Any other opinions/ideas?

> Should I just inherit from TestConnectionPool and null out the tests that aren't applicable to GAE instead of duplicating code?

Ideally there shouldn't be too many duplicate tests. Any overlapping functionality should be inherited (and thus won't need to be exercised separately) whereas any new functionality should be tested explicitly, right?

> How should this connection class be wired in? On GAE Sandbox environments w/o sockets, I assume this should be default.

This is a good question, and it'll depend if we go with the ConnectionPool-level or Manager-level. The other question is: How automatic/implicit vs explicit should this be? 

> What exact edge cases should this class be aware; e.g, where does URLFetch behavior differ from what's expected of httplib/sockets? Do the tests already cover this?

Honestly I was hoping that's something you could inform us about. We're not too sure how URLFetch differs, as we don't use it ourselves. :/

Another thought: It may be better to start this off as a urllib3/contrib/ submodule. Our standards for those are somewhat lower, so it'll be easier to get this work merged and start to be used by the community while we figure out how we want to integrate it into the core API.

urllib3/gae_connectionpool.py
shazow: Agreed. If we're doing an implicit drop-in replacement, then any features that are explicitly set by users (ie. non-default values) which are not available should yield a `PlatformWarning`.

If we're going to do an explicit activation for URLFetch, then maybe raising an exception would be better.

shazow:
@jonparrott Btw, thank you again for volunteering to tackle this. We very much appreciate your time and effort. :)

Lukasa:
> Hm, I've been thinking about this. I think it may make more sense to have the AppEngine URLFetch driver be at the Manager-level (see poolmanager.py) rather than ConnectionPool-level because it doesn't make sense for our PoolManager/ProxyManager/etc to manage pools of GAEConnectionPool, right?
> 
> This should make things a fair bit more straight-forward. Any other opinions/ideas?

The downside of this is that GAE no longer becomes transparent to users. The ideal case would be that libraries that use urllib3 just magically work on GAE, and I don't think that'd be true if the AppEngine driver was a replacement for PoolManager instead of a replacement for ConnectionPool. Though I'm a little concern that I missed what you were aiming for here.

> This is a good question, and it'll depend if we go with the ConnectionPool-level or Manager-level. The other question is: How automatic/implicit vs explicit should this be?

My vote is for automatic and implicit. The reason to do explicit is to force users to know that something different happens on GAE, but what we're doing there is pushing responsibility for testing and debugging those differences onto our users, when we should be the ones bearing that burden.

test/appengine/test_gae_connectionpool.py
jonparrott: Done.

test/appengine/test_gae_connectionpool.py
jonparrott: No, it's discovered by nose during the regular tests. I tried using nose-exclude to skip the app engine folder completely, but it seems nose-exclude doesn't work correctly w/ python3.

test/appengine/test_gae_connectionpool.py
jonparrott: Done.

urllib3/gae_connectionpool.py
jonparrott: They are copy/pasted from connectionpool.py. I'll do a sweep and remove all unused imports once I make sure everything is covered functionally.

urllib3/gae_connectionpool.py
jonparrott: Agreed w/ not silently ignoring, but URLFetch's default of 5s is really low and a lot of users will want to change this. What's the recommended way to do this?

jonparrott:
> Ideally there shouldn't be too many duplicate tests. Any overlapping functionality should be inherited (and thus won't need to be exercised separately) whereas any new functionality should be tested explicitly, right?

Cool, I'll inherit from the other test and null out the cases that aren't applicable.

> Honestly I was hoping that's something you could inform us about. We're not too sure how URLFetch differs, as we don't use it ourselves. :/

I can help, absolutely, but as I mentioned before I'm not intimately familiar with how URLFetch works under the covers (though I guess I could look). Most importantly, I'd like to know what things you guys are already aware of.

> It may be better to start this off as a urllib3/contrib/ submodule

Fine with me.

> PoolManager vs ConnectionPool

I'm cool with it either way, it seems the implementation is basically the same (urlopen). Let me know what you guys decide.

> @jonparrott Btw, thank you again for volunteering to tackle this. We very much appreciate your time and effort. :)

 No worries, I'm here to help make GCP a great place for Python developers.

> My vote is for automatic and implicit.

We should be aware of what we're doing to users if it's automatic and implicit, and the user should always be able to switch. Here's a comparison of App Engine's environments and the caveats I'm aware of:

_ | Local Dev Server | GAE (Sandbox) | GAE (Managed VMs Compat) | GAE (Managed VMs Custom)
---|------------------------|----------------------|------------------------------------------|-----------------------------------------
URLFetch | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :x: 
Sockets | :heavy_check_mark: | limited<sup>1</sup> | :heavy_check_mark: | :heavy_check_mark: 
PyOpenSSL | maybe<sup>2</sup> | optional<sup>3</sup> | optional<sup>4</sup> | optional<sup>5</sup>
Custom certs | :x: | depends<sup>6</sup>| :heavy_check_mark: | :heavy_check_mark:  

1. Only paid apps, can't access google apis, etc. [limitations](https://cloud.google.com/appengine/docs/python/sockets/#limitations-and-restrictions)
2. It's up to the user to have PyOpenSSL installed on their system.
3. Sandbox users have to [explicitly list this](https://cloud.google.com/appengine/docs/python/sockets/ssl_support), and it's a fixed, google-provided version.
4. Users can either use the method in 3 or modify their dockerfile / requirements to include it.
5. Users should be able to just add this to their requirements, but failing that, can add it to their dockerfile.
6. URLFetch doesn't support custom certs, but sockets + ssl do.

So, we need to figure out what the best implicit behavior is for each environment (I have code that can reliable detect this) and how to give users the ability to change it.

urllib3/gae_connectionpool.py
shazow: I'd be okay with changing the default timeout to something contextually sensitive on L30.

shazow:
> The downside of this is that GAE no longer becomes transparent to users. The ideal case would be that libraries that use urllib3 just magically work on GAE, and I don't think that'd be true if the AppEngine driver was a replacement for PoolManager instead of a replacement for ConnectionPool. Though I'm a little concern that I missed what you were aiming for here.

@Lukasa Why not? We could swap out PoolManager with GAEPoolManager at runtime same way we swap in VerifiedHTTPSConnectionPool. Or maybe we need to finally add another layer above PoolManager which would do this implicitly (we've been talking about adding a Session object? Or another name?).

The silly thing is that ConnectionPool objects are there to manage pools of connections (via our queue collections), which URLFetch does not have (those are managed behind the scenes by URLFetch). In effect, URLFetch _is_ an interface to a ConnectionPool-type object. Then the PoolManager would make an instance of these ConnectionPool managers per host, which would really add nothing in terms of functionality—runtime complexity and stack depth (and possibly GAE-incompatible bugs therein).

Plus, it's not like any of our Manager-features would work with URLFetch without more interception too. For example, ProxyManager wouldn't work.

urllib3/gae_connectionpool.py
jonparrott: Not sure what you mean?

shazow: Ah I misunderstood what you said. All the ways to set a timeout: https://urllib3.readthedocs.org/en/latest/#timeout

So maybe something like...

```python
if isinstance(timeout, Timeout) and not (timeout.read is timeout.connect is Timeout.DEFAULT_TIMEOUT):
    warnings.warn("URLFetch does not support granular timeout settings, reverting to total timeout.", GAEPlatformWarning)
```

jonparrott: Done.

jonparrott:
@shazow I've moved the pool into the contrib directory.

Lukasa:
> Why not? We could swap out PoolManager with GAEPoolManager at runtime same way we swap in VerifiedHTTPSConnectionPool.

I suppose, but...ewww. It feels really weird to do that at runtime if we can possibly avoid it, though I suppose it's an acceptable approach.

Basically, I don't like it, but I can't articulate a good reason. Consider me -0 on that idea. =P

Lukasa:
Actually, no, I think that might be fine. If it were me I'd probably take the stack depth hit, but if that's off the table, the 'patch at runtime' approach is fine.

shazow:
@Lukasa Please do me a favour, sleep on it and try to articulate why you feel negative about it, I'm curious. :) Either way it's going to be a patch at runtime, right? Whether we swap the ConnectionPool or the PoolManager. Only difference is that a ConnectionPool of ConnectionPools is an extra unnecessary layer of abstraction, as far as I can tell. There's no benefit, and some potential surprises (e.g. ProxyManager not working with it).

I'd really like to get somekind of consensus on this as I think this decision will inform how we move forward in our future designs. I still suspect that adding another layer above PoolManager which automagically manages the correct things might be the best and least-surprising option.

jonparrott:
> I still suspect that adding another layer above PoolManager which automagically manages the correct things might be the best and least-surprising option.

It's your library, but adding another layer of complexity just to support GAE's sandbox doesn't sound fun to me. If there's other use cases beyond just supporting GAE for this then I'm cool with it.

shazow:
> It's your library, but adding another layer of complexity just to support GAE's sandbox doesn't sound fun to me. If there's other use cases beyond just supporting GAE for this then I'm cool with it.

Yea it's something we've talked about before. urllib3 in general is largely about being very manual/explicit, but more and more we move towards autodetecting/autoconfiguring which creates a schism. Putting all that magic in one place might help keep things more sane: Things like setting up good TLS defaults, proxy settings from .netrc/env vars, platform-specific injections, maybe someday cookie handling (ugh).

jonparrott:
Gotcha, sounds good to me.

t-8ch:
AFAIK URLfetch does not support chunked encoding, so when we use it implicitly streaming APIs will not work anymore.

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+    def _urlfetch_response_to_http_response(self, urlfetch_resp):
+        return HTTPResponse(
+            # In onder for decoding to work, we must present the content as
+            # a file-like object.
+            body=BytesIO(urlfetch_resp.content),
+            headers=urlfetch_resp.headers,
+            status=urlfetch_resp.status_code
+        )
+
+    def _absolute_url(self, path):
```
t-8ch: Better move the definition from `HTTPConnectionPool` to `ConnectionPool` and use it from there

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class GAEPlatformWarning(HTTPWarning):
+    pass
+
+
+class GAEConnectionPool(ConnectionPool, RequestMethods):
+
+    def __init__(self, host, port=None, scheme='http', timeout=5, headers=None,
+                 retries=None, validate_certificate=False):
```
t-8ch: About `validate_certificate`:

* Do we want to default to `False`?
   For our normal connectionpool we have to as we don't have a cert bundle.
   Recently we began defaulting to `True` as soon as we get a bundle.
* Should we emit a warning when `validate_certficate=True and scheme != 'https'`?

Lukasa:
> AFAIK URLfetch does not support chunked encoding, so when we use it implicitly streaming APIs will not work anymore.

Are you serious?

From [RFC 7230 Section 4.1](https://tools.ietf.org/html/rfc7230#section-4):

> A recipient MUST be able to parse and decode the chunked transfer coding.

The only way URLFetch gets around that requirement is by advertising themselves as a HTTP/1.0 client. Do they do that?

t-8ch:
@Lukasa See #620. AppEngine handles chunking but only gives us the dechunked response.

t-8ch:
@jonparrott Probably a dumb question: What was the advantage of using URLFetch explicitly instead of using implicitly via the httplib-URLFetch shim?

test/appengine/test_gae_connectionpool.py
t-8ch: As the gae connectionpool is now in `contrib` you can move the tests to the contrib tests.
Then they won't be picked up automatically anymore.

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+                pool_timeout=None, release_conn=None, **response_kw):
+
+        url = self._absolute_url(url)
+
+        try:
+            response = urlfetch.fetch(
+                url,
+                payload=body,
+                method=method,
+                headers=headers if headers else {},
```
t-8ch: headers=headers or {}

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+    pass
+
+
+class GAEConnectionPool(ConnectionPool, RequestMethods):
+
+    def __init__(self, host, port=None, scheme='http', timeout=5, headers=None,
+                 retries=None, validate_certificate=False):
+
+        if not urlfetch:
+            raise EnvironmentError(
```
t-8ch: `EnvironmentError` seems to be more for IO stuff. Maybe define our own `PlatformException` matching our `PlatformWarning`

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+                "URLFetch is not available in this environment.")
+
+        ConnectionPool.__init__(self, host, port)
+        RequestMethods.__init__(self, headers)
+
+        if isinstance(timeout, Timeout):
+            if not timeout.read is timeout.connect is Timeout.DEFAULT_TIMEOUT:
```
t-8ch: Use `DEFAULT_TIMEOUT` from the instance in case someone uses a subclass?

test/appengine/test_gae_connectionpool.py
```diff
@@ -0,0 +1,80 @@
...
+from urllib3.exceptions import (
+    ConnectTimeoutError,
+    EmptyPoolError,
+    DecodeError,
+    MaxRetryError,
+    TimeoutError,
+    ProtocolError,
+)
+
+from urllib3.packages.six import b, u
```
t-8ch: Many of the imports are not used.

test/appengine/test_gae_connectionpool.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+from test.with_dummyserver.test_connectionpool import TestConnectionPool
+
+
+# Note that this doesn't run in the sandbox, it only runs with the URLFetch
+# API stub enabled. There's no need to enable the sandbox as we know for a fact
+# that URLFetch is used by the connection pool.
+class TestGAEConnectionPool(TestConnectionPool):
+    nosegae_urlfetch = True
```
t-8ch: Could you note that this is a magic variable used by nosegae directly? (Probably in the class comment earlier)

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+                    "reverting to total timeout.", GAEPlatformWarning)
+            timeout = timeout.total
+
+        self.deadline = timeout
+
+        self.scheme = scheme
+        self.validate_certificate = validate_certificate
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=_Default,
```
t-8ch: `timeout=Timeout.DEFAULT_TIMEOUT`

urllib3/contrib/gaepool.py
```diff
@@ -0,0 +1,101 @@
...
+                pool_timeout=None, release_conn=None, **response_kw):
```
t-8ch: `response_kw` should be used.

jonparrott: What should ``assert_same_host`` do?

urllib3/contrib/gaepool.py
jonparrott: > Do we want to default to False

Probably not, as the default will be True soon in URLFetch itself.

> Should we emit a warning

No, it's not an error for URLFetch to be provided those arguments, and if we default to True then all http requests will fail.

jonparrott:
> @jonparrott Probably a dumb question: What was the advantage of using URLFetch explicitly instead of using implicitly via the httplib-URLFetch shim?

It's more a decision for you guys, but it's one less layer of indirection. GAE's httplib just calls this under the covers. I figured it would be best to work directly with the ``fetch`` method ourselves than try to bend GAE httplib to our will.

Also, there's the case where users can replace httplib so there's no promise that it's using URLFetch anymore.

jonparrott:
@shazow any decision on Pool vs Manager?

test/appengine/test_gae_connectionpool.py
jonparrott: Done, moved the test to contrib.

shazow:
@Lukasa Any new thoughts on ConnectionPool vs Manager? I'm still leaning on Manager.

urllib3/contrib/gaepool.py
jonparrott: Will hold off on this until the decision of pool vs manager is settled.

urllib3/contrib/gaepool.py
jonparrott: I'm actually wrong on that last part, http + validate certificates is fine.

urllib3/contrib/gaepool.py
jonparrott: Done w/ GAEPlatformError.

test/appengine/test_gae_connectionpool.py
jonparrott: Fixed.

test/appengine/test_gae_connectionpool.py
jonparrott: Done.

Lukasa:
I think I'm still with you on Manager, tbh.

Altogether this is just one of those nasty areas where neither solution feels great to me, but avoiding the extra level of indirection is probably the way to go.

jonparrott:
Cool, I'll switch it over to a manager class soon. I just pushed a new updated addressing some of @t-8ch's feedback.

urllib3/contrib/gaepool.py
jonparrott: Done.

urllib3/contrib/gaepool.py
jonparrott: Not sure I follow here.

t-8ch: `timeout.connect is timeout.DEFAULT_TIMEOUT`

shazow: Yea makes sense, lowercase-`timeout.DEFAULT_TIMEOUT` lets people subclass `Timeout` as `MyTimeout` and provide their own `DEFAULT_TIMEOUT` sentinel and it'll still work.

urllib3/contrib/gaepool.py
shazow: When you port this to a Manager, this stuff gets simplified. We don't care about `assert_same_host`, for example. See: https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L153

jonparrott:
Alright, it's been moved over to a connection manager. Still re-using the connection pool tests, as they seem to provide the most relevant tests.

Questions from here:
* What functionality is missing from the connection manager? (retries?)
* What exceptions are expected from urlopen? presently, I just have TimeoutError.

urllib3/contrib/gaemanager.py
```diff
@@ -0,0 +1,83 @@
...
+
+class GAEPlatformWarning(HTTPWarning):
+    pass
+
+
+class GAEPlatformError(HTTPError):
+    pass
+
+
+class GAEConnectionManager(RequestMethods):
```
shazow: `GAEManager` is probably sufficient.

shazow:
> What functionality is missing from the connection manager? (retries?)

In approximate order of priority:

- SSL/TLS configuration
- Optional retries (based on `retries` field, using our `Retry` configuration object)
- Optional redirects (based on `redirect` and `retries` configuration).
- Proxy support (probably configured on init).

> What exceptions are expected from urlopen? presently, I just have TimeoutError.

- SSLError
- MaxRetryError (if retry/redirects are a thing)
- ProxyError (if proxy support is a thing)
- ProtocolError (if the response violated HTTP expectations).
- Regarding TimeoutError, note that there is ReadTimeoutError (request possibly sent but failed did not read the response successfully) and ConnectTimeoutError (failed before sending the request, safe to retry). Would be great to differentiate them if possible.

shazow:
Any thoughts on naming classes `AppEngine*` rather than `GAE*`?

jonparrott:
I'm cool with renaming. 

* SSL/TLS configuration

What configuration is there? For URLFetch, it's either "validate" or "don't validate".

* Optional retries (based on retries field, using our Retry configuration object)

I can do this.

* Optional redirects (based on redirect and retries configuration)

URLFetch can follow redirects, but I'm unsure about the error on depth, and there's no way to configure the depth.

* Proxy support (probably configured on init)

No proxy options for URLFetch.

* Granular timeout (using our Timeout configuration object)

Probably not possible.

shazow:
Sounds good, thanks @jonparrott. Would be good to do a `AppEnginePlatformWarning` for anything that we're explicitly not supporting in urlopen that is otherwise available. Either way, this is already huge progress over what we had. :)

> What configuration is there? For URLFetch, it's either "validate" or "don't validate".

Any way to supply your own cacerts?

jonparrott:
> Any way to supply your own cacerts?

Nope. :(

jonparrott:
Renamed to AppEngineManager. I'm going to wire up the remainder of exceptions then move on to "features". 

Any word on how to get this wired up? I'd like to write a real GAE app that can run through a bunch of cases and make sure everything works in the real environment.

jonparrott:
Okay, all the exceptions that [URLFetch can throw](https://cloud.google.com/appengine/docs/python/urlfetch/exceptions) should be mapped appropriately. I also added a test case to verify that self-signed certs will fail.

urllib3/contrib/appenginemanager.py
```diff
@@ -0,0 +1,117 @@
...
+                allow_truncated=False,  # revisit
+                follow_redirects=redirect,
+                deadline=self._get_absolute_timeout(timeout),
+                validate_certificate=self.validate_certificate,
+            )
+        except urlfetch.DeadlineExceededError as e:
+            raise TimeoutError(self, e)
+
+        except urlfetch.InvalidURLError as e:
+            if 'too large' in e.message:
```
shazow: Should this condition raise `AppEnginePlatformError(e)` rather than ProtocolError(e)?

urllib3/contrib/appenginemanager.py
```diff
@@ -0,0 +1,117 @@
...
+                warnings.warn("URLFetch request too large, URLFetch only "
+                              "supports requests up to 10mb in size.",
+                              AppEnginePlatformWarning)
+            raise ProtocolError(e)
+
+        except urlfetch.DownloadError as e:
+            raise ProtocolError(e)
+
+        except urlfetch.ResponseTooLargeError as e:
+            warnings.warn("URLFetch response too large, URLFetch only supports"
```
shazow: Should this condition raise `AppEnginePlatformError(e)` rather than ProtocolError(e)?

urllib3/contrib/appenginemanager.py
```diff
@@ -0,0 +1,117 @@
...
+                          "responses up to 32mb in size.",
+                          AppEnginePlatformWarning)
+            raise ProtocolError(e)
+
+        except urlfetch.SSLCertificateError as e:
+            raise SSLError(e)
+
+        # TODO: Request too large (there isn't an explicit exception for this)
```
shazow: Is this different from L78?

shazow:
@jonparrott This is great, thank you!

Re filename, I think `urllib3/contrib/appengine.py` is probably sufficient.

To start, we can do:

```python
from urllib3.contrib.appengine import AppEngineManager

http = AppEngineManager()
# Now you can use `http` as any other urllib3 pool/manager

r = http.request('GET', 'https://google.com/')
```

I think where we are is good enough to merge? Next up we'll need to start thinking about how we want to make the instrumentation implicit with the normal urllib3 flow, so you don't need to be as aware of your environment to use urllib3. After that, move it out of contrib and documentation. :)

urllib3/contrib/appenginemanager.py
jonparrott: Up to you. So far I've tried to maintain compatibility with URLFetch's API. If you feel it would be more appropriate to raise the AppEnginePlatformError, that's fine with me.

urllib3/contrib/appenginemanager.py
jonparrott: Left that in, I was able to cover that case. Will remove.

jonparrott:
Got a few simple tests running on production app engine here: http://python-urllib3-gae-int-tests.appspot.com/

Discovered that URLFetch in the live environment handles deflate encoding but doesn't remove the header. Fun times. Pushed code to fix that.

shazow:
Ah fantastic, thank you for your thoroughness. :)

jonparrott:
Alright, I've finished the environment detection code and cleaned up the commit message. I think the outstanding things now are:

1. Implement retries?
2. Should AppEngineManager throw ProtocolErrors or AppEnginePlatformErrors?

Thanks for working with me here, hopefully this can make the "support app engine's weirdness" story much more contained for you guys.

shazow:
> Implement retries?

That would be great, at least some subset of retry functionality.

> Should AppEngineManager throw ProtocolErrors or AppEnginePlatformErrors?

I vote for `AppEnginePlatformError` since the errors are specific to limitations of running on the AppEngine platform, or alternatively a new exception that subclasses it.

jonparrott:
Cool, exceptions updated. Looking into retries now.

jonparrott:
Okay, so I've noticed a inconsistency here:

> Redirects are also considered to be a subset of retries but can be configured or disabled individually.

Presently, ``AppEngineManager`` has a separate ``redirect`` parameter. Should I remove that and use ``retries``'s ``redirect`` value instead? The issue is that URLFetch's follow redirects is boolean and doesn't allow you to configure how many times it'll try.

shazow:
`redirect` controls whether redirects are enabled or disabled (boolean). `retries.redirect` controls how many redirects are followed before giving up and raising a `MaxRetryError`.

So you should primarily look at `redirect` to enable/disable same as URLFetch, then you can do the same platform warning if `retries.redirect` is set (since it's not supported). Also, assuming there is a max-redirects error, would be swell to wrap it in our `MaxRetryError` exception.

jonparrott:
Solid, will do.

jonparrott:
Retry logic implemented:

1. It'll redirect, but treat any truthy values of retries.redirect the same (and warn about it)
2. It'll raise a MaxRetryError if URLFetch indicates too many redirects occurred.
3. It will behave as normal for the forced retry stuff.

(that was way harder than it should've been)

I think that does it @shazow.

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,208 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # Return False to re-raise any potential exceptions
+        return False
+
+    def urlopen(self, method, url, body=None, headers=None,
+                retries=None, redirect=True, timeout=Timeout.DEFAULT_TIMEOUT,
+                **response_kw):
+
+        retries = self._get_retries(retries, redirect)
+        log.info(retries)
```
shazow: Let's try to keep logging messages consistent with other urlopens?

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,208 @@
...
+            raise AppEnginePlatformError(
+                "URLFetch response too large, URLFetch only supports"
+                "responses up to 32mb in size.", e)
+
+        except urlfetch.SSLCertificateError as e:
+            raise SSLError(e)
+
+        except urlfetch.InvalidMethodError as e:
+            raise AppEnginePlatformError(
+                "URLFetch does not support method %s." % method, e)
```
shazow: Style nit: For consistency with our other exceptions, this would be better:

```diff
-                "URLFetch does not support method: %s" % method, e)
+                "URLFetch does not support method %s." % method, e)
```

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,208 @@
...
+    def _get_absolute_timeout(self, timeout):
+        if timeout is Timeout.DEFAULT_TIMEOUT:
+            return 5  # 5s is the default timeout for URLFetch.
+        if isinstance(timeout, Timeout):
+            if not timeout.read is timeout.connect:
+                warnings.warn(
+                    "URLFetch does not support granular timeout settings, "
+                    "reverting to total timeout.", AppEnginePlatformWarning)
+            return timeout.total
+        print "Timeout is a {}".format(type(timeout))
```
shazow: Print statement

shazow:
Looks sensible, just need to cleanup some debugging code and ideally match up logging messages with other pools.

Could use another pair of eyes, though. (@Lukasa @t-8ch @sigmavirus24?)

urllib3/contrib/appengine.py
jonparrott: Oops, will remove.

urllib3/contrib/appengine.py
jonparrott: Done.

urllib3/contrib/appengine.py
jonparrott: debugging statement I accidentally left in, fixed.

jonparrott:
Removed the debugging stuff (that's what I get for trying to get this pushed before I left yesterday).

shazow:
@Lukasa Can you give this a quick look-over? I want to include this in the v1.11 early next week.

@jonparrott Would you be up for adding a bit of docs for how to use this module? It can live here for now: https://github.com/shazow/urllib3/blob/master/docs/contrib.rst -> https://urllib3.readthedocs.org/en/latest/contrib.html 

Something along the lines of...

```python
from urllib3.contrib.appengine import AppEngineManager

http = AppEngineManager()
r = http.request('GET', 'https://google.com/')
```

And maybe a little blurb for detecting which AppEngine environment you're in, a recipe for swapping out the respective manager, and that this will happen automagically as soon as AppEngineManager graduates from contrib-land. :)

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+class AppEnginePlatformError(HTTPError):
+    pass
+
+
+class AppEngineManager(RequestMethods):
+    """
+    Connection manager for Google App Engine sandbox applications.
+
+    This manager uses the URLFetch service directly instead of using the
+    emulated httplib.
```
shazow: Can we add a list of features that AppEngineManager will throw PlatformError/Warnings for in the docstring?

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+    def __init__(self, headers=None, retries=None, validate_certificate=True):
+
+        if not urlfetch:
+            raise AppEnginePlatformError(
+                "URLFetch is not available in this environment.")
+
+        if is_prod_appengine_v2():
+            raise AppEnginePlatformError(
+                "Use normal urllib3.PoolManager instead of AppEngineManager"
+                "on Managed VMs, as using URLFetch is not necessarily in "
```
Lukasa: s/necessarily/necessary/

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+class AppEngineManager(RequestMethods):
+    """
+    Connection manager for Google App Engine sandbox applications.
+
+    This manager uses the URLFetch service directly instead of using the
+    emulated httplib.
+    """
+
+    def __init__(self, headers=None, retries=None, validate_certificate=True):
+
```
Lukasa: Can we remove this whitespace?

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+        if is_prod_appengine_v2():
+            raise AppEnginePlatformError(
+                "Use normal urllib3.PoolManager instead of AppEngineManager"
+                "on Managed VMs, as using URLFetch is not necessarily in "
+                "this environment.")
+
+        warnings.warn(
+            "urllib3 is using URLFetch on Google App Engine sandbox instead "
+            "of sockets. To use sockets directly instead of URLFetch see "
+            "<TODO: documentation link>.", AppEnginePlatformWarning)
```
Lukasa: We should probably clear up TODOs.

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+        if retries.is_forced_retry(method, status_code=http_response.status):
+            retries = retries.increment(
+                method, url, response=http_response, _pool=self)
+            log.info("Forced retry: %s" % url)
+            retries.sleep()
+            return self.urlopen(
+                method, url,
+                body=body, headers=headers,
+                retries=retries, redirect=redirect,
+                timeout=timeout, **response_kw)
```
Lukasa: Ideally this would be done iteratively not recursively (what do you think this is, a Lisp? ;) ) but I don't care enough about it to actually mandate it. It's fine as is. I just wanted to complain. =P

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+
+        return http_response
+
+    def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
+
+        if is_prod_appengine_v1():
```
Lukasa: How can this return `False`? We have a check elsewhere that this isn't `prod_appengine_v2`. What am I missing?

urllib3/contrib/appengine.py
```diff
@@ -0,0 +1,205 @@
...
+            # Production GAE handles deflate encoding automatically, but does
+            # not remove the encoding header.
+            content_encoding = urlfetch_resp.headers.get('content-encoding')
+
+            if content_encoding == 'deflate':
+                del urlfetch_resp.headers['content-encoding']
+
+        return HTTPResponse(
+            # In onder for decoding to work, we must present the content as
```
Lukasa: s/onder/order/

urllib3/contrib/appengine.py
Lukasa: Ah, nevermind, I see there's also an `is_local_appengine()`.

urllib3/contrib/appengine.py
shazow: Other urllib3 code does this recursively too, so it's consistent with the rest of the codebase.

Lukasa:
@shazow I'm certainly happy to have this in contrib land for the next release.

urllib3/contrib/appengine.py
jonparrott: done.

urllib3/contrib/appengine.py
jonparrott: done.

urllib3/contrib/appengine.py
jonparrott: done.

urllib3/contrib/appengine.py
jonparrott: done, now that I know where the docs will live.

urllib3/contrib/appengine.py
jonparrott: done.

jonparrott:
Fixed stuff in @Lukasa and added some docs in contrib (thanks for writing the sample for me, @shazow).

shazow:
@jonparrott This is fantastic and invaluable. Thank you so much for helping us out here, I'm very excited about this and the urllib3/requests community really appreciates your contribution but even moreso: I'm sure all the AppEngine users will be just ecstatic! Make sure to link to this thread in your promo packet, you really deserve some recognition for your work. :) :heart: :custard: :sparkles:

jonparrott:
Happy to help. I'm sure this will have bugs and fun edge cases, so please mention me on any bugs that get filed on this.

webmaven:
@jonparrott Thanks so much for responding when I sent up the Bat-Signal!

PR #667, opened 2015-07-04
show invalid fingerprint in exception

t-8ch:


urllib3/util/ssl_.py
```diff
@@ -123,7 +123,8 @@ def assert_fingerprint(cert, fingerprint):
     digest_length = len(fingerprint)
     hashfunc = HASHFUNC_MAP.get(digest_length)
     if not hashfunc:
-        raise SSLError('Fingerprint is of invalid length.')
+        raise SSLError(
+            'Fingerprint {0} is of invalid length.'.format(fingerprint))
```
shazow: For consistency with other message styles, how about we do: `'Fingerprint is of invalid length: {}'`? (Easier to split/parse if you need to this way.)

shazow: Or maybe grammatically better yet, "Invalid fingerprint length: {}"

t-8ch: I wanted to show the fingerprint, not necessarily the length itself.

shazow:
Perfect, thanks!

PR #668, opened 2015-07-06
Fix IPv6 HTTPS proxy

chripede:
When using an IPv6 proxy for https connections the connection would fail with

`MaxRetryError: HTTPSConnectionPool(host='localhost', port=53619): Max retries exceeded with url: / (Caused by ProxyError('Cannot connect to proxy.', gaierror(-2, 'Name or service not known')))`

That happens because the hard brackets are not removed before calling getaddrinfo.

This fixes that.

### DIFF
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -60,8 +60,8 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     """
 
     host, port = address
-    if host[0] == '[':
-        host = host.replace('[', '').replace(']', '')
+    if host.startswith('['):
+        host = host.strip('[]')
     err = None
     for res in socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM):
         af, socktype, proto, canonname, sa = res
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -60,6 +60,8 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     """
 
     host, port = address
+    if host[0] == '[':
+        host = host.replace('[', '').replace(']', '')
     err = None
     for res in socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM):
         af, socktype, proto, canonname, sa = res
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -163,6 +163,9 @@ In chronological order:
 * James Atherfold <jlatherfold@hotmail.com>
   * Bugfixes relating to cleanup of connections during errors.
 
+* Christian Pedersen <https://github.com/chripede>
+  * IPv6 HTTPS proxy bugfix
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
 
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -14,7 +14,6 @@ from dummyserver.handlers import TestingApp
 from dummyserver.proxy import ProxyHandler
 
 
-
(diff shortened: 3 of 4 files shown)


urllib3/util/connection.py
```diff
@@ -60,6 +60,8 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     """
 
     host, port = address
+    if host[0] == '[':
+        host = host.replace('[', '').replace(']', '')
```
shazow: This might be better:

```python
    if host.startswith('['):
        host = host.strip('[]')
```

shazow:
@Lukasa Thoughts?

t-8ch:
We have the same stuff in `util.url` and this only breaks for https requests.
I think there is a more elegant solution.

shazow:
@t-8ch Yea I was thinking that too but because it's a fairly low-level helper (and `util` helpers shouldn't know anything about urllib3's internals), I thought it was reasonable to duplicate that piece of logic there. What do you suggest?

t-8ch:
Hm we only add this in `util` on formatting. I am fairly sure we should also strip it on parsing.

shazow:
@t-8ch https://github.com/shazow/urllib3/blob/master/urllib3/util/url.py#L172?

t-8ch:
Now I am confused, but will look at it in a little bit.

sigmavirus24:
FWIW, that bit in url.py is correct. Stripping it there means that when we have a request uri that has an IPv6 literal in it, we would do the wrong thing when reconstituting the URL.

Lukasa:
I hate IPv6 literals. The colons were a terrible idea.

Yeah, one way or another we need to handle this, but I think @t-8ch and @sigmavirus24 have got a better idea of how this section of the code should be structured than I do.

sigmavirus24:
The colons aren't even the problem here =P. The problem is that getaddrinfo doesn't like the `[]`s which makes me a sad panda. I know the colons are what necessitate the `[]s` but still, =P

shazow:
Mmm. Given that this is specific to socket.getaddrinfo, I'd like to keep this fix as close to that code as possible but I am open to suggestions for how to structure it better. :)

shazow:
By the way, @chripede, thank you for this excellent contribution with tests right off the bat. We very much appreciate it!

sigmavirus24:
So one way to have this somewhere else is to add a property to the `Url` class, e.g.,

```py
@property
def gaihostname(self):
    return self.hostname.strip('[]')
```

It's simple since

```py
>>> 'foo'.strip('[]')
'foo'
>>> '[::1]'.strip('[]')
'::1'
```

And we probably won't use it in a lot of places, so we can do it dynamically like that. Alternatively, if you're worried about `hostname` being none, make it a method instead. 

I'm not strongly advocating for this, just trying to come up with a compromise.

shazow:
@sigmavirus24 Is this useful for anything outside of the `create_connection` helper? Note that it takes a `(host, port)` tuple, not a Url.

sigmavirus24:
@shazow I'm pretty sure the answer is "No". Like I said, I'm not really advocating for this, just offering another solution that keeps logic pertaining to the URL near the rest of the logic pertaining to it.

That said, either way this PR can be simplified by simply calling `str.strip` without bothering with a conditional. URIs won't have `[]`s in the authority unless you're using IPv6 literals.

shazow:
Good enough for me then. :) @chripede Thanks!

chripede:
@shazow no problem. Thanks for merging so fast. This have been bugging me for a long time in requests.

shazow:
:custard:!!

sigmavirus24:
Good work @chripede! Thanks for fixing it for us! `:D`

PR #671, opened 2015-07-09
Indicate hostname in SubjectAltNameWarning

sigmavirus24:
This adds a new warning subclass for this warning specifically and has
it default to print exactly once (per message and therefore) per
hostname. This will reduce the number of warnings that users see and
provide an overall better user-experience. This allows the user to
determine the offending host and to only see the warning once per time
it connects to that host.

### DIFF
--- a/urllib3/__init__.py
+++ b/urllib3/__init__.py
@@ -58,6 +58,8 @@ del NullHandler
 import warnings
 # SecurityWarning's always go off by default.
 warnings.simplefilter('always', exceptions.SecurityWarning, append=True)
+# SubjectAltNameWarning's should go off once per host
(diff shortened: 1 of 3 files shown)


Lukasa:
\o/ :+1: :shipit:

shazow:
Great idea, thanks!

sigmavirus24:
Fastest merge ever?

shazow:
@sigmavirus24 It's a free Early Preview of Urllib3 Pro: Contributors Edition, $99/mo for fast merges, more emoji in replies, and a mention in the README.

sigmavirus24:
:heart_eyes:

PR #674, opened 2015-07-14
Make stream() satisfy its v1.10 contract

### LINKED ISSUE #673: stream() "contract" changed in 1.10.3, can now raise IncompleteRead
In 1.10.2, `HTTPResponse.stream()` always called `HTTPResponse.read()`. This method catches all `HTTPException` (including `IncompleteRead`), and re-raises them as `ProtocolError`.

In 1.10.3 and 1.10.4, `HTTPResponse.stream()` may call `HTTPResponse.read_chunked()` instead. This method may (via the `_update_chunk_length()` helper method) raise `IncompleteRead`. This exception will not be caught, and will be raised out of the call to `HTTPResponse.stream()`.

A project I work on recently upgraded to requests 2.7.0 (which has vendored urllib3 1.10.4), and has started to see crashes because of `IncompleteRead` being raised from `HTTPResponse.stream()`. We previously would catch `RequestException`, as requests would catch `ProtocolError` (but not `IncompleteRead`) and re-raise a `ChunkedEncodingError` in `Response.iter_content()`.

I'm not familiar with `urllib3.util.retry`, but at a glance it also seems like this change would negatively affect the `_is_read_error()` helper method.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24334962-stream-contract-changed-in-1-10-3-can-now-raise-incompleteread?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jmoldow:
Here's my attempt at a fix for #673. Let me know what you think.

In 1.10 through 1.10.2, `HTTPResponse.stream()` always called `HTTPResponse.read()`.  This method catches all `httplib.HTTPException` (including `httplib.IncompleteRead`), and re-raises them as
`ProtocolError`.

In 1.10.3 and 1.10.4, `HTTPResponse.stream()` may call `HTTPResponse.read_chunked()` instead. This method may (via the `_update_chunk_length()` helper method) raise `httplib.IncompleteRead`.
This exception will not be caught, and will be raised out of the call to `HTTPResponse.stream()`.

Fix the contract of `HTTPResponse.stream()` by making `HTTPResponse.read_chunked()` catch the same low-level exceptions that `HTTPResponse.read()` catches. This is accomplished by refactoring this
try/except logic into a shared context manager.

Fixes #673.

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -7,7 +7,7 @@ try:
 except ImportError:
     import httplib
 from urllib3.response import HTTPResponse
-from urllib3.exceptions import DecodeError, ResponseNotChunked
+from urllib3.exceptions import DecodeError, ResponseNotChunked, ProtocolError
 
 
 from base64 import b64decode
@@ -487,7 +487,7 @@ class TestResponse(unittest.TestCase):
         r.chunked = True
         r.chunk_left = None
         resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
-        self.assertRaises(httplib.IncompleteRead, next, resp.read_chunked())
+        self.assertRaises(ProtocolError, next, resp.read_chunked())
 
     def test_chunked_response_without_crlf_on_end(self):
         stream = [b"foo", b"bar", b"baz"]
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -204,7 +204,7 @@ class HTTPResponse(io.IOBase):
         return data
 
     @contextmanager
-    def _catch_low_level_exceptions_and_release_connection(self):
+    def _error_catcher(self):
         """
         Catch low-level python exceptions, instead re-raising urllib3
         variants, so that low-level exceptions are not leaked in the
@@ -275,7 +275,7 @@ class HTTPResponse(io.IOBase):
         flush_decoder = False
         data = None
 
-        with self._catch_low_level_exceptions_and_release_connection():
+        with self._error_catcher():
             if amt is None:
                 # cStringIO doesn't like amt=None
                 data = self._fp.read()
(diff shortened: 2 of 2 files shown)


Lukasa:
This looks plausible to me. My only question is about the name of the context manager, which seems a bit lengthy, but I have no other problems with this.

sigmavirus24:
Does this catch all of the exceptions that `read_chunked` might raise?

jmoldow:
@Lukasa I'm perfectly happy to change the name, if a different one is proposed. Alternatively, we can handle the `release_conn()` in a different context manager, so that we have ``def _catch_low_level_exceptions`` and ``def _release_conn``.

@sigmavirus24 I don't know. I'm not sure what all the possible exceptions are. I did this under the assumptions that `read_chunked` won't be raising any exceptions that `read` didn't use to raise (inside the try block), since it was refactored out of that. But it's certainly possible that something else was introduced. Though the ``except HTTPException`` will help make sure that most reasonable things are caught. But if anyone more familiar with this stuff wants to look at `read_chunked` and see if we're missing anything, we can catch anything that we've missed, if necessary.

shazow:
I haven't considered using a context manager for this kind of thing, interesting approach. Thank you @jmoldow. We should consider using this pattern in our massive ConnectionPool try/except block too.

Regarding the name, how about `with self._error_catcher():`?

jmoldow:
@shazow I changed the name to that, thanks.

shazow:
Great, thank you! :) You're welcome to add yourself to our CONTRIBUTORS.txt, by the way.

PR #678, opened 2015-07-17
Add a "secure" extras point

### LINKED ISSUE #677: setup.py extra_require for certifi, pyopenssl, other?
Try to setup as many of https://urllib3.readthedocs.org/en/latest/security.html#security's recommended dependencies as possible.

Maybe something like `pip install urllib3[secure]`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24502931-setup-py-extra_require-for-certifi-pyopenssl-other?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
To be user-friendlier, add "secure" as an extra so users can do

    pip install urllib3[secure]

Further, only install the dependencies for SNI on versions of Python
that need them (2.6 and ... sort of 2.7). We install certifi on every
version of Python because it's awesome and the stdlib doesn't provide
any of its functionality on any version of Python.

Closes #677

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -142,6 +142,32 @@ Now you can continue using urllib3 as you normally would.
 
 For more details, check the :mod:`~urllib3.contrib.pyopenssl` module.
 
+Installing urllib3 with SNI support and certificates
+----------------------------------------------------
+
+By default, if you need to use SNI on Python 2.6 or Python 2.7.0-2.7.8, you
+have to install PyOpenSSL, ndghttpsclient, and pyasn1 separately. Further, to
+use certifi you have to install it separately. If you know that you want these
+dependencies when you install urllib3, you can now do::
+
+    pip install urllib3[secure]
+
+This will install the SNI dependencies on Python 2.6 and 2.7 (we cannot yet
+restrict the microversion for 2.7) and certifi on all versions of Python.
+
+.. note::
+
+    If you do this on linux, e.g., Ubuntu 14.04, you will need extra system
+    dependencies for PyOpenSSL. Specifically, PyOpenSSL requires cryptography
+    which will require you to install:
+
+    - build-essential
+    - python-dev
+    - libffi-dev
+    - libssl-dev
+
+    The package names may vary depending on the distribution of linux you are
+    using.
 
 .. _insecurerequestwarning:
 
--- a/setup.py
+++ b/setup.py
@@ -54,4 +54,15 @@ setup(name='urllib3',
           'tornado',
       ],
       test_suite='test',
+      extras_require={
+          'secure;python_version<="2.7"': [
+              'pyOpenSSL',
+              'ndg-httpsclient',
+              'pyasn1',
+              'certifi',
+          ],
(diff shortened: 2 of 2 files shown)


Lukasa:
Can we get sufficiently good granularity to say that we need SNI for Python 2.7.8 and earlier, rather than burying all Python 2.7.9 installs with them?

docs/security.rst
```diff
@@ -142,6 +142,32 @@ Now you can continue using urllib3 as you normally would.
 
 For more details, check the :mod:`~urllib3.contrib.pyopenssl` module.
 
+Installing urllib3 with SNI support and certificates
```
shazow: It's kinda redundant with the section just above, should it be a subsection? Or hmm.. Need somekind of restructuring of this doc.

sigmavirus24: It kind of fits into multiple sections so I made it a separate one

sigmavirus24:
> Can we get sufficiently good granularity to say that we need SNI for Python 2.7.8 and earlier, rather than burying all Python 2.7.9 installs with them?

See the docs that were added to the commit. @rbtcollins can explain why we can't do this with sufficient granularity.

docs/security.rst
shazow: We need a kickstarter for rewriting the urllib3 docs. :P Good enough for now (tm).

shazow:
plz2ping when button is green.

setup.py
```diff
@@ -54,4 +54,11 @@
           'tornado',
       ],
       test_suite='test',
+      extras_require={
+          'secure': [
```
jayvdb: this means it wont have any effect on 2.7.8 and lower, where it is still needed.

docs/security.rst
```diff
@@ -142,6 +142,32 @@ Now you can continue using urllib3 as you normally would.
...
+
+By default, if you need to use SNI on Python 2.6 or Python 2.7.0-2.7.8, you
+have to install PyOpenSSL, ndghttpsclient, and pyasn1 separately. Further, to
+use certifi you have to install it separately. If you know that you want these
+dependencies when you install urllib3, you can now do::
+
+    pip install urllib3[secure]
+
+This will install the SNI dependencies on Python 2.6 and 2.7 (we cannot yet
+restrict the microversion for 2.7) and certifi on all versions of Python.
```
jayvdb: Why not use `python_full_version < '2.7.9'` ?

sigmavirus24:
Transient error on PyPy. Everything else is green.

docs/security.rst
sigmavirus24: @jayvdb according to discussions I've seen elsewhere, that does not work properly everywhere. So there's no point using something that's already broken.

PR #679, opened 2015-07-18
Use HTTPHeaderDict in request

sigmavirus24:
Supersedes #633

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -167,10 +167,6 @@ class HTTPHeaderDict(MutableMapping):
     def __ne__(self, other):
         return not self.__eq__(other)
 
-    values = MutableMapping.values
-    get = MutableMapping.get
-    update = MutableMapping.update
-
     if not PY3: # Python 2
         iterkeys = MutableMapping.iterkeys
         itervalues = MutableMapping.itervalues
@@ -306,8 +302,6 @@ class HTTPHeaderDict(MutableMapping):
 
     def items(self):
         return list(self.iteritems())
-
-    viewitems = items
 
     @classmethod
     def from_httplib(cls, message): # Python 2
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -174,7 +174,7 @@ class HTTPHeaderDict(dict):
     values = MutableMapping.values
     get = MutableMapping.get
     update = MutableMapping.update
-    
+
     if not PY3: # Python 2
         iterkeys = MutableMapping.iterkeys
         itervalues = MutableMapping.itervalues
@@ -236,7 +236,7 @@ class HTTPHeaderDict(dict):
             raise TypeError("extend() takes at most 1 positional "
                             "arguments ({} given)".format(len(args)))
         other = args[0] if len(args) >= 1 else ()
-        
+
         if isinstance(other, HTTPHeaderDict):
             for key, val in other.iteritems():
                 self.add(key, val)
@@ -307,16 +307,16 @@ class HTTPHeaderDict(dict):
     def from_httplib(cls, message): # Python 2
         """Read headers from a Python 2 httplib message object."""
         # python2.7 does not expose a proper API for exporting multiheaders
-        # efficiently. This function re-reads raw lines from the message 
+        # efficiently. This function re-reads raw lines from the message
         # object and extracts the multiheaders properly.
         headers = []
-         
+
         for line in message.headers:
             if line.startswith((' ', '\t')):
                 key, value = headers[-1]
                 headers[-1] = (key, value + '\r\n' + line.rstrip())
                 continue
-    
+
             key, value = line.split(':', 1)
             headers.append((key, value.strip()))
 
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -97,14 +97,7 @@ class RecentlyUsedContainer(MutableMapping):
             return list(iterkeys(self._container))
 
 
-_dict_setitem = dict.__setitem__
-_dict_getitem = dict.__getitem__
-_dict_delitem = dict.__delitem__
-_dict_contains = dict.__contains__
-_dict_setdefault = dict.setdefault
-
-
-class HTTPHeaderDict(dict):
+class HTTPHeaderDict(MutableMapping):
     """
     :param headers:
         An iterable of field-value pairs. Must not contain multiple field names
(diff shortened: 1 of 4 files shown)


sigmavirus24:
This might not be perfect, but it at least passes the `test_collections` tests. And it's late here =P

sigmavirus24:
Tests are failing because there isn't that tests the `get` method directly of the `HTTPHeaderDict`.

urllib3/_collections.py
```diff
@@ -290,33 +294,35 @@ def copy(self):
...
         """Iterate over all headers, merging duplicate ones together."""
         for key in self:
-            val = _dict_getitem(self, key)
+            val = self._container[key.lower()]
             yield val[0], ', '.join(val[1:])
 
     def items(self):
         return list(self.iteritems())
 
+    viewitems = items
```
sigmavirus24: Forgot to delete this. We don't need it.

Lukasa:
Broadly LGTM.

shazow:
Mmmm yea this is a smaller change than I thought we'd need. If you can make the remaining test(s) pass, I'm all for it. :)

shazow:
Btw, :custard:.

sigmavirus24:
So the tests *should* pass now. The thing that I'm uncertain about is [this test](https://github.com/shazow/urllib3/commit/8d7273d97997c34d0f22c06fad7d96312f2db536#diff-1ec38ab7dbd425bce37c385e1e250b95R241). I can definitely see that `dict(HTTPHeaderDict(baz='quuz')) == {'baz': 'quuz'}` but something else is turning that into `{'Baz': 'quuz'}`. I haven't investigated it yet.

shazow:
Hm wonder if it's our dummyserver and/or httplib mangling headers on some versions.

sigmavirus24:
> Hm wonder if it's our dummyserver and/or httplib mangling headers on some versions.

It's more bizarre because this test *was* passing before we fixed the round-trip behaviour. =/

sigmavirus24:
If I debug into the request methods, I get to 

```py
(Pdb) p kw['headers']
{'Accept': '*/*', 'baz': 'quux', 'Host': 'localhost:52261'}
```

On L276 in `urllib3/poolmanager.py`

sigmavirus24:
If I go further, it's still the right case in `connectionpool.py`. I don't have much time to go further though. Sorry. =(

sigmavirus24:
For anyone else looking to follow the same rabbit hole.

1. Add `import pdb; pdb.set_trace()` (or `ipdb` if that's your thing) to `test/with_dummyserver/test_proxy_poolmanager.py` at L238.
2. `tox -e py27 -- -s test.with_dummyserver.test_proxy_poolmanager` (or `py34` or whatever)
3. Start stepping into functions that make the request.

Lukasa:
So this isn't in `httplib`: the header is emitted in the case it was provided to `httplib`. Current money is on the dummyserver doing that.

Lukasa:
Definitely the dummy server. From tcpdump:

```
GET http://localhost:62953/headers HTTP/1.1
Accept-Encoding: identity
Accept: */*
Host: localhost:62953
baz: quux
foo: bar

HTTP/1.1 200 OK
Etag: "feb15a7ce2766d24cf20cf1633197fc5d8a59281"
Date: Sun, 19 Jul 2015 10:22:50 GMT
Content-Length: 146
Server: TornadoServer/4.1
Content-Type: text/html; charset=UTF-8,text/plain

{"Accept": "*/*", "Content-Length": "0", "Foo": "bar", "Connection": "close", "Host": "localhost:62953", "Accept-Encoding": "gzip", "Baz": "quux"}
```

We definitely emitted headers in lowercase, they were definitely reflected to us in upper-case.

Lukasa:
Note also that the dummyserver inserts some headers we didn't send (`Connection: close`) and mangles one we did send (`Accept-Encoding: gzip`).

Lukasa:
Ok, so had a quick chat with @shazow and I think that test should just be rethought. We can't effectively test that case is preserved when we round-trip via the dummy server, so let's just not.

I can see a case for wanting a test that ensures that case is preserved that includes httplib, but if we're going to do that we should do it as a socket-level test where we can confirm that the bytes-on-the-wire are really what we think they are.

sigmavirus24:
So b49490d has a socketlevel test for this.

shazow:
Yay :Custard:!

Are we good to go? Shall I merge and ship v1.11?

sigmavirus24:
That's Cory's call.
-- 

Lukasa:
Let's rock-and-roll. :shipit:

mmedvede:
This causes OpenStack devstack-gate tests to fail on Fedora 21 (but not on Ubuntu). There are interactions between many python packages during the test, so I am not yet certain if it is urlib's fault, but if I use the commit previous to the merge of this pull request, tempest tests pass.

Lukasa:
@mmedvede Can we get more information, e.g. a link to a Launchpad bug?

sigmavirus24:
@Lukasa it appears they're tracking it in https://bugs.launchpad.net/glance/+bug/1476770. The original bug was because the gate is not operating as it is supposed to (and I suspect the same is happening on Fedora). The gate should **always** be installing requests from pip instead of using system packages but they're not so when they install urllib3 1.11.0, it installs over the system urllib3 that was installed. I maintain that this is fundamentally a problem with the gate. Note that instead of fixing the gate, they released a new version of oslo.vmware (the only library to use urllib3 in openstack directly) with a cap to exclude 1.11.0. This is not a problem or a bug in urllib3.

mmedvede:
@sigmavirus24 I was not aware of the bug, thanks. It is exactly what was happening to our third-party Fedora 21 gate (IBM PowerKVM CI) before I disabled the urllib3 1.11 install. I assume that Ubuntu installs newer system python-requests, so when urllib3 1.11 is installed on top, the gate doesn't break. The weird thing is that both Ubuntu and Fedora gates show requests 2.7.0 installed. I have confirmed that doing 'pip install --upgrade --force-reinstall requests' fixes Fedora 21 in my scenario, even though requests would still show as being 2.7.0.

@Lukasa Thanks for quick response. Not urllib3 problem.

sigmavirus24:
@mmedvede check where requests 2.7.0 is installed from. You'll see that it's installed from the system packages and that is *not* what the gate is supposed to do. Anyway, further discussion should happen on LaunchPad, not here.

tfmorris:
This change requires all `requests-cache` caches be deleted.

PR #683, opened 2015-07-22
Clean changelog

Disassem:


shazow:
Erg oops, looks like my release script has a bug in it. Thanks!

PR #685, opened 2015-07-29
Replace select.select with util functions

### LINKED ISSUE #589: ValueError: filedescriptor out of range in select() in urllib3
We are using requests library in our code and after recent urllib3 upgrade we've started seeing following errors in logs on our testing servers:

```
"/usr/lib/python2.7/site-packages/requests-2.6.0-py2.7.egg/requests/packages/urllib3/connectionpool.py", line 762, in _validate_conn
conn.connect()
File "/usr/lib/python2.7/site-packages/requests-2.6.0-py2.7.egg/requests/packages/urllib3/connection.py", line 238, in connect
ssl_version=resolved_ssl_version)
File "/usr/lib/python2.7/site-packages/requests-2.6.0-py2.7.egg/requests/packages/urllib3/contrib/pyopenssl.py", line 302, in ssl_wrap_socket
select.select([sock], [], [])
ValueError: filedescriptor out of range in select()
```

What it basically means is that `sock` that is passed to select[1] is `>= 1024`. That may happen for example if application opens 1024 files on start and then gets `OpenSSL.SSL.WantReadError` during TLS handshake while using urllib3.

Regression was introduced in: a49bec58fa5f7aca76d1d0b2f1975eb094648eab [pyopenssl: wait for data before handshake retry]

Simple fix is to factor out functionality for poll -> select fallback from https://github.com/shazow/urllib3/blob/master/urllib3/util/connection.py#L12 and replace all `select.select()` calls in urllib3 with it.

[1] `select(2)` is bound to `FD_SETSIZE` fds which equals to 1024 on most UNIX systems: http://pubs.opengroup.org/onlinepubs/007908799/xsh/select.html
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10901196-valueerror-filedescriptor-out-of-range-in-select-in-urllib3?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
select.poll is preferable when it is available. Unfortunately it isn't
available on every operating system but select.select is, so we can wrap
the logic to fall back to select.select from select.poll in helper
functions in urllib3.util.

Closes #589 
---

It is late here and this is more of a straw man PR than something I'm certain works. I was careful about the logic though so it should work (but again, I haven't done extensive testing).

### DIFF
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -3,30 +3,63 @@ try:
     from select import poll, POLLIN, POLLOUT, POLLERR
 except ImportError:  # `poll` doesn't exist on OSX and other platforms
     poll = False
-    try:
-        from select import select as select
-    except ImportError:  # `select` doesn't exist on AppEngine.
-        select = False
+
+try:
+    from select import select
+except ImportError:  # `select` doesn't exist on AppEngine.
+    select = False
+
+try:
+    from select import epoll, EPOLLIN, EPOLLOUT, EPOLLERR
+except ImportError:  # `epoll` only exists on Linux
+    epoll = False
+
+try:
+    from select import kqueue, kevent
+except ImportError:  # `kqueue` is only available for BSDs
+    kevent = kqueue = False
+
+
+def _poll(reads, writes, exceptions, timeout):
+    # Assume that if we're being called then poll or epoll is defined
+    pollfunc = poll
+    read_flag = POLLIN
+    write_flag = POLLOUT
+    exc_flag = POLLERR
+
+    if epoll:
+        pollfunc = epoll
+        read_flag = EPOLLIN
+        write_flag = EPOLLOUT
+        exc_flag = EPOLLERR
+
+    p = pollfunc()
+    for fd in reads:
+        p.register(fd, read_flag)
+
+    for fd in writes:
+        p.register(fd, write_flag)
+
+    for fd in exceptions:
+        p.register(fd, exc_flag)
+
+    return p.poll(timeout)
+
+
+def _kqueue(*args):
+    pass
 
 
 def _select(readlist, writelist, exceptionallist, timeout):
     # Some platforms, e.g., OSX do not support poll
-    if not poll:
-        if select:
-            return select(readlist, writelist, exceptionallist, timeout)
+    if epoll or poll:
+        return _poll(readlist, writelist, exceptionallist, timeout)
 
-    # Using poll is better on platforms that support it
-    p = poll()
-    for fd in readlist:
-        p.register(fd, POLLIN)
+    if kqueue and kevent:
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Looks like they're just failing on test coverage.

Lukasa:
If we're going to do this, then we want to look at some other options. For example, while `poll` is preferable, it is generally less efficient than `epoll`/`kqueue`. We should consider whether or not we can use those options instead on platforms that support them.

(Ideally I'd like to use IOCP on Windows, but that's...tricky.)

urllib3/util/connection.py
```diff
@@ -1,14 +1,54 @@
 import socket
 try:
-    from select import poll, POLLIN
+    from select import poll, POLLIN, POLLOUT, POLLERR
 except ImportError:  # `poll` doesn't exist on OSX and other platforms
     poll = False
     try:
-        from select import select
+        from select import select as select
```
shazow: does the `as select` do anything here?

urllib3/util/connection.py
```diff
@@ -1,14 +1,54 @@
...
+
+
+def wait_to_write_data(socket, timeout=0.0):
+    sockets = _select([], [socket], [], timeout)
+
+    # select.select returns ([...], [...], [...])
+    if isinstance(sockets, tuple):
+        return sockets[1]
+
+    return sockets
```
shazow: What would this be if it's not a tuple?

shazow:
:floppy_disk:

urllib3/util/connection.py
sigmavirus24: A list of tuples (from poll). Poll returns a list of things like

```
[(3, POLLIN), (6, POLLOUT), ...]
```

Where `POLLIN` and `POLLOUT` are the actual integers that `select.POLL(IN|OUT)` are assigned to in the module, and `3` and `6` are file descriptors.

urllib3/util/connection.py
sigmavirus24: Oh I had been mucking with naming that `_select` and the function we define `select` but then realized there was a better API for what we needed.

sigmavirus24:
@Lukasa am I mistaken or will `epoll`/`kqueue` be available when `poll` is available, or not really?

Lukasa:
> am I mistaken or will epoll/kqueue be available when poll is available, or not really?

They're orthogonal really. epoll is a Linux kernel special, kqueue is a BSD special. Both those platforms have `poll`, but they may not have both.

Lukasa:
Obviated by #1001.

PR #686, opened 2015-07-29
Use debug instead of info to log new connection

luca3m:


shazow:
Dupe of https://github.com/shazow/urllib3/pull/71

PR #688, opened 2015-07-29
Rely on six to import httplib or http.client

sigmavirus24:
Regarding https://github.com/elastic/elasticsearch-py/issues/253

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -1,7 +1,3 @@
-try:
-    import http.client as httplib
-except ImportError:
-    import httplib
 from contextlib import contextmanager
 import zlib
 import io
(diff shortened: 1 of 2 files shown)


Lukasa:
Does this actually solve the problem in the linked issue? Why is `six` not affected when we are?

sigmavirus24:
`six` handles this better by explicitly checking for Python version: https://github.com/shazow/urllib3/blob/master/urllib3/packages/six.py#L110 so on Python 2, `pies` won't break us. `python-future` shouldn't either but appears to be. Either way, six means we don't need to be conscious of the ordering of our imports for `httplib`.

Lukasa:
Yeah, that's fair enough. Ok, I'm happy to go with this. =)

PR #693, opened 2015-08-10
Moved the Auth part before doing the split on /, ? and #, to allow those characters in passwords

pe2mbs:
This issue i found using PIP. Therefore i made the change.

### DIFF
--- a/urllib3/util/url.py
+++ b/urllib3/util/url.py
@@ -156,6 +156,13 @@ def parse_url(url):
     if '://' in url:
         scheme, url = url.split('://', 1)
 
+    # FIX PE2MBS: 10 Aug 2015 Marc Bertens-Nguyen  <m.bertens@pe2mbs.nl>
+    # Moved the Auth part before doing the split on /, ? and #, to allow those characters in passwords
+    # Auth
+    if '@' in url:
+        # Last '@' denotes end of auth part
+        auth, url = url.rsplit('@', 1)
+
     # Find the earliest Authority Terminator
     # (http://tools.ietf.org/html/rfc3986#section-3.2)
     url, path_, delim = split_first(url, ['/', '?', '#'])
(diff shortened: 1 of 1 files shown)


Lukasa:
This is a tricky problem. Strictly, those characters are not allowed unquoted in the auth part of the URL, so arguably this can lead to incorrect parsing.

sigmavirus24:
@pe2mbs you're absolutely correct that passwords can have whatever characters the user wants. The thing is the user must properly encode those in the auth portion of the URL. URIs (according to RFC 3986) are composed of 5 parts:

1. Scheme
1. Authority
1. Path
1. Query
1. Fragment

Authorities consist of three *sub*parts:

1. user info (username:password)
1. host
1. port

After parsing the first 5, you're supposed to parse the authority information. I strongly suspect that this change would cause the `Url` class here to incorrectly parse a URI like `https://medium.com/@shazow`.

If you're passing credentials via the URI, you should absolutely be encoding the username and password to ensure any reserved characters are properly percent-encoded.

shazow:
Maybe we need a helper for encoding the auth? Or at least a documentation example for how to do it?

@pe2mbs Would you like to do that in a new PR?

pe2mbs:
@shazow I don't know what you are asking of me, can you elaborate on it. I'm fairly new to open source development and even new to python.

sigmavirus24:
@shazow I suspect (since @pe2mbs mentioned they found this in pip) that they were passing this in on the CLI and not using urllib3 to handle these bits

shazow:
(Closing this PR for now but we can continue the discussion)

Mmm, yea not sure that would help then. We could add a helper that you can use to properly-encoded your URL?

Also, should `str(Url(scheme='http', auth='foo#@!', host='localhost'))` convert to something other than `"http://foo#@!bar@localhost"`?

sigmavirus24:
> Also, should str(Url(scheme='http', auth='foo#@!', host='localhost')) convert to something other than "http://foo#@!bar@localhost"?

90% certain that it should. [This](https://github.com/kennethreitz/requests/issues/2703) is a similar issue raised on requests. [My comment](https://github.com/kennethreitz/requests/issues/2703#issuecomment-127729958) pulls out the relevant bits from the RFC.

shazow:
I'd be +1 to that, or add an auth setter which does encoding (might be less surprising than the init params doing magic).

PR #701, opened 2015-08-31
Add support for CA directories.

Lukasa:
This enables support for directories of CAs in urllib3. These are usually used for efficiency purposes, as they allow lazy-lookup of certificate authorities in response to the specific certificates needed for a given connection.

This is a precursor bit of work to kennethreitz/requests#2659.


CHANGES.rst
```diff
@@ -18,10 +20,10 @@ dev (master)
 
 * ``pip install urllib3[secure]`` will install Certifi and
   PyOpenSSL as dependencies. (Issue #678)
-  
+
```
shazow: I love how you re-spaced all the old changes, but you didn't add a space to your new entry. :P

shazow:
I'm always sad to add more params to an already-param-heavy init, but not sure how avoidable this is.

Would it be feasible to have something like **cert_kw and just pass the args all the way down the pipe?

CHANGES.rst
sigmavirus24: trimming trailing whitespace is what he did.

sigmavirus24:
I don't think you can mix `**conn_kw` with `**cert_kw`.

shazow:
Yea, but I mean it can travel along as **conn_kw for a while, then somewhere down the pipe, the leftover args will become **cert_kw. Or is this an empty fantasy?

CHANGES.rst
shazow: Ah right, diffs are hard.

CHANGES.rst
```diff
@@ -6,6 +6,8 @@ dev (master)
 
 * Rely on ``six`` for importing ``httplib`` to work around
   conflicts with other Python 3 shims. (Issue #688)
```
shazow: Newline here

CHANGES.rst
Lukasa: I will never understand why text editors don't have "trim trailing whitespace" as on by default. Is the default assumption really that we have to cater to people writing in Whitespace? Or can we assume that a developer who wants to write Whitespace can find the option and switch it, thereby saving us from novice devs accidentally committing trailing whitespace so that they appear in my diffs later?

I had a point here but I don't remember what it was.

CHANGES.rst
Lukasa: ACK.

Lukasa:
@shazow Regardless of whether this is an empty fantasy or not, I can certainly refactor this change to use `**conn_kw` and pop the argument off if you prefer that approach. It's easy enough to do!

shazow:
@Lukasa If it doesn't look a lot worse, I think that might be better. I'm only +0.5 to this, so feel free to override. :)

tiran:
ceterum censeo you should support a custom SSLContext object or a wrap_socket_object_in_ssl_socket(socket, hostname) callback. You don't want to have options and flags for all SSLContext features ;)

Lukasa:
We have an open pull request for that, @tiran, but unfortunately it's gotten bogged down because this turns out to be a somewhat tricky problem. In the short term, this is an easy enough thing for us to add, and it resolves a known user complaint.

Lukasa:
@shazow Actually I think I lied. Conceptually the SSL arguments are connection arguments, but weirdly they are managed by the connection pool. Arguably we want a refactor there, but...is that refactor in scope for this change?

shazow:
You're right, we may as well include it part of a larger refactor which combines all the cert-related keywords (and other groups too).

matejcik:
is there a good reason to keep this a separate argument, as opposed to, say:
```
if os.path.isdir(ca_certs):
  context.load_verify_locations(None, ca_certs)
else:
  context.load_verify_locations(ca_certs)
```
?

Lukasa:
@matejcik I think at the urllib3 level users should be explicit: it doesn't hurt and catches subtle bugs. I think that logic will definitely happen at the requests level.

matejcik:
ISTM that the file/directory distinction is purely an artifact of openssl's way of handling certificate files? and that pushing it ever-upwards is a bad idea, so stopping it as early as possible (ideally right in python ssl) is desirable?

i mean, it hurts a little bit: forces the user of the library to think about format of their cert store, and also adds a parameter to an already parameter-heavy method.

but i'm not sure about my premise, so that's what i'm asking, basically: are there cases where being explicit about this is a tangible benefit?

Lukasa:
@matejcik You're entirely right, this is an artefact of OpenSSL, and you're entirely right that at some point we should stop propagating the distinction. However, there are costs to doing it lower in the stack, and those costs are discoverability.

Additionally, we are not really abstracting away OpenSSL here. Having a bundle of PEMs in a file is a valid input for OpenSSL, but not necessarily for other SSL libs, and the 'certificate directory' is *very* OpenSSL specific, with a format and layout that is generated directly by OpenSSL itself. We cannot pretend that either of these things has no relationship to OpenSSL, so pasting them together and claiming they have no relationship to OpenSSL is a little disingenuous.

In this case, I think some users should be thinking about the format of their cert store. In either case, I believe this to be an interim solution to this problem, where the SSLContext solution will be the final way to deal with this.

PR #703, opened 2015-09-04
Tests cleanup and deflake-ing, added NewConnectionError exception

shazow:
- [x] Quarantined tests that depend on the /sleep handler, tests are far less flakey when that whole suite is skipped now.
- [x] Need to migrate tests to not use the /sleep handler (and ideally no inline sleep either, but that's of a lesser priority).

- Added `NewConnectionError` exception.
- Removed /sleep handler, ported tests to use event triggers and socket-level handlers.
- A couple test conditions got pruned because they were annoying, but I'm fairly confident they were extraneous anyways.

Tests seem to be waaaay less flakey now. Some weird failures every 30-50 back-to-back runs, but much better in general.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -152,10 +152,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
         conn = pool._get_conn()
         pool._make_request(conn, 'GET', '/')
         tcp_nodelay_setting = conn.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY)
-        assert tcp_nodelay_setting > 0, ("Expected TCP_NODELAY to be set on the "
-                                         "socket (with value greater than 0) "
-                                         "but instead was %s" %
-                                         tcp_nodelay_setting)
+        self.assertGreater(tcp_nodelay_setting, 0)
 
     def test_socket_options(self):
         """Test that connections accept socket options."""
@@ -607,9 +604,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             pool = HTTPConnectionPool(self.host, self.port,
                     source_address=addr, retries=False)
             r = pool.request('GET', '/source_address')
-            assert r.data == b(addr[0]), (
-                "expected the response to contain the source address {addr}, "
-                "but was {data}".format(data=r.data, addr=b(addr[0])))
+            self.assertEqual(r.data, b(addr[0]))
 
     def test_source_address_error(self):
         for addr in INVALID_SOURCE_ADDRESSES:
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -98,6 +98,7 @@ class TestConnectionPoolTimeouts(SocketDummyServerTestCase):
         block_event.set() # Release request
 
         ready_event.wait()
+        block_event.clear()
         self.assertRaises(ReadTimeoutError, pool.request, 'GET', '/')
         block_event.set() # Release request
 
@@ -180,22 +181,21 @@ class TestConnectionPoolTimeouts(SocketDummyServerTestCase):
 
     def test_total_timeout(self):
         block_event = Event()
-        ready_event = self.start_basic_handler(block_read=block_event, num=2)
+        ready_event = self.start_basic_handler(block_send=block_event, num=2)
 
         # This will get the socket to raise an EAGAIN on the read
-        timeout = Timeout(connect=3, read=0)
-        pool = HTTPConnectionPool(self.host, self.port, timeout=timeout)
-        conn = pool._get_conn()
-        self.assertRaises(ReadTimeoutError, pool._make_request, conn, 'GET', '/')
+        timeout = Timeout(connect=3, read=SHORT_TIMEOUT)
+        pool = HTTPConnectionPool(self.host, self.port, timeout=timeout, retries=False)
+        self.assertRaises(ReadTimeoutError, pool.request, 'GET', '/')
 
         block_event.set()
         ready_event.wait()
+        block_event.clear()
 
         # The connect should succeed and this should hit the read timeout
         timeout = Timeout(connect=3, read=5, total=SHORT_TIMEOUT)
-        pool = HTTPConnectionPool(self.host, self.port, timeout=timeout)
-        conn = pool._get_conn()
-        self.assertRaises(ReadTimeoutError, pool._make_request, conn, 'GET', '/')
+        pool = HTTPConnectionPool(self.host, self.port, timeout=timeout, retries=False)
+        self.assertRaises(ReadTimeoutError, pool.request, 'GET', '/')
 
     def test_create_connection_timeout(self):
         timeout = Timeout(connect=SHORT_TIMEOUT, total=LONG_TIMEOUT)
@@ -633,7 +633,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
         for addr in INVALID_SOURCE_ADDRESSES:
             pool = HTTPConnectionPool(self.host, self.port, source_address=addr, retries=False)
             # FIXME: This assert flakes sometimes. Not sure why.
-            self.assertRaises(NewConnectionError, pool.request, 'GET', '/source_address?{}'.format(addr))
+            self.assertRaises(NewConnectionError, pool.request, 'GET', '/source_address?{0}'.format(addr))
 
     def test_stream_keepalive(self):
         x = 2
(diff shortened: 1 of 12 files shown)


shazow:
@Lukasa @sigmavirus24 @dstufft Sooo I kinda got carried away trying to do a 1.11.1 release and ended  up fixing enough things to warrant a 1.12 release. Can I get a review? (Should I split this up into two releases?)

urllib3/connection.py
```diff
@@ -133,11 +134,15 @@ def _new_conn(self):
...
 
-        except SocketTimeout:
+        except SocketTimeout as e:
             raise ConnectTimeoutError(
                 self, "Connection to %s timed out. (connect timeout=%s)" %
                 (self.host, self.timeout))
 
+        except SocketError as e:
+            raise NewConnectionError(
+                self, "Failed to establish a new connection: %s" % e)
```
shazow: Is this sensible?

urllib3/exceptions.py
```diff
@@ -112,6 +112,9 @@ class ConnectTimeoutError(TimeoutError):
     "Raised when a socket timeout occurs while connecting to a server"
     pass
 
+class NewConnectionError(ConnectTimeoutError, PoolError):
```
shazow: Do these base exceptions make sense?

The old behaviour was partly encompassed by ConnectTimeoutError before (not completely), whereas now it's all under NewConnectionError so I think it's sensible to make it backwards-compatible.

Otherwise, not sure what is should go under—PoolError? Something else?

shazow:
Also getting some `Connection reset by peer` on travisci, wonder if it's travis or the tests. Hrm.

urllib3/connection.py
Lukasa: Maybe, but I don't think I like the name (maximum bikeshedding!). `ConnectionSetupError`?

test/with_dummyserver/test_connectionpool.py
```diff
@@ -372,7 +392,7 @@ def test_bad_connect(self):
             pool.request('GET', '/', retries=5)
             self.fail("should raise timeout exception here")
         except MaxRetryError as e:
-            self.assertTrue(isinstance(e.reason, ProtocolError), e.reason)
+            self.assertEqual(type(e.reason), NewConnectionError)
```
Lukasa: Can we use [`assertIsInstance`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsInstance) here? (Ugh I hate unittest).

test/with_dummyserver/test_connectionpool.py
```diff
@@ -669,6 +686,8 @@ def test_cleanup_on_connection_error(self):
             self.assertEqual(http.pool.qsize(), http.pool.maxsize)
 
 
+
+
```
Lukasa: Are these newlines deliberate?

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -287,7 +287,7 @@ def test_https_proxy_timeout(self):
             https.request('GET', self.http_url, timeout=0.001)
             self.fail("Failed to raise retry error.")
         except MaxRetryError as e:
-            assert isinstance(e.reason, ConnectTimeoutError)
+           self.assertEqual(type(e.reason), ConnectTimeoutError)
```
Lukasa: Also `assertIsInstance`.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -298,7 +298,7 @@ def test_https_proxy_pool_timeout(self):
             https.request('GET', self.http_url)
             self.fail("Failed to raise retry error.")
         except MaxRetryError as e:
-            assert isinstance(e.reason, ConnectTimeoutError)
+            self.assertEqual(type(e.reason), ConnectTimeoutError)
```
Lukasa: Also `assertIsInstance`.

urllib3/connectionpool.py
```diff
@@ -592,13 +594,13 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
 
-        except (TimeoutError, HTTPException, SocketError, ConnectionError) as e:
+        except (TimeoutError, HTTPException, SocketError, ProtocolError) as e:
             # Discard the connection for these exceptions. It will be
             # be replaced during the next _get_conn() call.
             conn = conn and conn.close()
             release_conn = True
 
-            if isinstance(e, SocketError) and self.proxy:
+            if isinstance(e, (SocketError, NewConnectionError)) and self.proxy:
```
Lukasa: I don't think we can ever hit this if `e` is `NewConnectionError`, can we? I got lost traipsing through the exception heirarchy, but I don't think any of the parent exceptions of `NewConnectionError` are declared in the `except` block.

urllib3/connection.py
```diff
@@ -133,11 +134,15 @@ def _new_conn(self):
             conn = connection.create_connection(
                 (self.host, self.port), self.timeout, **extra_kw)
 
-        except SocketTimeout:
+        except SocketTimeout as e:
```
Lukasa: Why are we saving the timeout?

urllib3/exceptions.py
Lukasa: I can see the backward compatibility argument for `ConnectTimeoutError`, though you really want to comment that because it's definitely weird and I can see us forgetting that logic.

Otherwise I don't think `PoolError` is right: `ProtocolError` seems better to me.

Lukasa:
Mostly good, feedback inline. <3 thanks @shazow!

test/with_dummyserver/test_connectionpool.py
shazow: Did originally, does not exist in py26.

urllib3/connection.py
shazow: Is that more descriptive? ConnectionSetupError sounds to me like it's a problem with the client's OS, rather than a transient network issue.

urllib3/exceptions.py
shazow: The error happens in the pool though, when `._new_conn()` is called. Also `ProtocolError` is supposed to be for errors that happen mid-request, as opposed to strictly before the request is even sent.

test/with_dummyserver/test_connectionpool.py
Lukasa: Any reason not just to add the [unittest backport module](https://pypi.python.org/pypi/unittest2) for that Python version?

urllib3/exceptions.py
Lukasa: Mm, I suppose, but `PoolError` feels weird to me, but it might just be that I don't have the right feel for the exception hierarchy. =)

urllib3/connection.py
Lukasa: Honestly not sure. =(

urllib3/connectionpool.py
shazow: `TimeoutError` is the parent. But yea, our exception hierarchy needs a big makeover with explicit semantics of the request state during failure. :/

urllib3/exceptions.py
shazow: Yea, maybe I'll cut PoolError. They're mostly supposed to be for pool state violations (ie. if you fiddle with internal mechanics and ruin a counter somewhere).

urllib3/connection.py
shazow: Yea why we're not using it. I think I considered merging the two before but changed my mind. The symmetry is nice, but I think I'll change this back if I remember.

test/with_dummyserver/test_connectionpool.py
shazow: Probably not. :x

test/with_dummyserver/test_connectionpool.py
shazow: More things in our backports, and really we don't get more value in this case.

Ultimately we should "just" port everything to pytest or something. :/

dstufft: +1 pytest.

dstufft: Of course, pytest can run unittest tests just fine, so you can switch to pytest as your runner and just slowly migrate things over to use regular ``assert`` statements if you want. You don't need to do a big bang up "ok we're on pytest now" thing.

shazow: Yes, not a terrible idea, but I strongly suspect it'll devolve into a big several-day ordeal of cleanup that really needs to happen anyways. The structure of the tests is a hot mess right now.

urllib3/exceptions.py
Lukasa: That was my understanding, but then I got confused. =(

shazow:
yoloetc

PR #704, opened 2015-09-09
fix import bug

michael-yin:
fix import bug so the code can be run without modifed


Lukasa:
Yup, that definitely looks wrong! :cake:

shazow:
Thank you!

PR #705, opened 2015-09-15
Add info about disabling warnings for end users

msabramo:
For example, you can use `PYTHONWARNINGS` to disable warnings without modifying code (when using Python >= 2.7).

Relates to:

- https://github.com/pypa/pip/pull/3109
- https://github.com/pypa/pip/issues/2681

Thanks to @dstufft and @sigmavirus24 for the idea!

### DIFF
--- a/docs/security.rst
+++ b/docs/security.rst
@@ -185,21 +185,8 @@ Unverified HTTPS requests will trigger a warning via Python's ``warnings`` modul
 This would be a great time to enable HTTPS verification:
 :ref:`certifi-with-urllib3`.
 
-If you know what you're doing and would like to disable this and other warnings,
-you can use :func:`~urllib3.disable_warnings`::
+For info about disabling warnings, see `Disabling Warnings`_.
 
-    import urllib3
-    urllib3.disable_warnings()
-
-Making unverified HTTPS requests is strongly discouraged. ˙ ͜ʟ˙
-
-Alternatively, if you are using Python's ``logging`` module, you can capture the
-warnings to your own log::
-
-	logging.captureWarnings(True)
-
-Capturing the warnings to your own log is much preferred over simply disabling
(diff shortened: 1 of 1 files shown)


docs/security.rst
```diff
@@ -219,3 +235,18 @@ newer Python version, or that you use pyOpenSSL as described in the
 If you know what you are doing and would like to disable this and other
 warnings, please consult the :ref:`insecurerequestwarning` section for
 instructions on how to handle the warnings.
+
+If you are using a program that uses ``urllib3`` and want to suppress this
```
shazow: I'd prefer not to include these instructions twice.

docs/security.rst
```diff
@@ -201,6 +201,22 @@ warnings to your own log::
 Capturing the warnings to your own log is much preferred over simply disabling
 the warnings.
 
+If you are using a program that uses ``urllib3`` and want to suppress this
```
shazow: How about something simpler, maybe:

```
It is possible to suppress warnings by [setting the `PYTHONWARNINGS` environment variable in Python 2.7+ or using the `-W` flag with the Python interpreter](https://docs.python.org/2/using/cmdline.html#cmdoption-W). For example:

    PYTHONWARNINGS="ignore:Unverified HTTPS request" ./make-insecure-request.py
```

shazow:
Good idea adding a mention about this, suggested some tweaks. :)

msabramo:
Good comments! I updated it with your suggestions.

In order to reduce the duplication, I made a new section called "Disabling Warnings" and the two sections on specific warnings reference that. I think that's a little nicer. Let me know what you think.

shazow:
That's lovely, thank you! :custard:

msabramo:
You're welcome! Thanks for urllib3!

PR #706, opened 2015-09-15
Use proper environment markers

### LINKED ISSUE #684: urllib3 1.11 does not provide the extra 'secure'
I tried with Python 2.7 and 2.6 inside different virtualenv.

```bash
pip install 'urllib3[secure]'
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24839510-urllib3-1-11-does-not-provide-the-extra-secure?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sigmavirus24:
This will intelligently install the right dependencies for a "secure"
urllib3 but only if you have a recent enough version of setuptools and
pip.

Closes #684

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,3 +11,11 @@ max-line-length=99
 
 [wheel]
 universal = 1
+
+[metadata]
+provides-extra = secure
+requires-dist =
+    pyOpenSSL; python_version<="2.7" and extra == 'secure'
+    ndg-httpsclient; python_version<="2.7" and extra == 'secure'
+    pyasn1; python_version<="2.7" and extra == 'secure'
+    certifi; extra == 'secure'
--- a/setup.py
+++ b/setup.py
@@ -55,13 +55,13 @@ setup(name='urllib3',
       ],
       test_suite='test',
       extras_require={
-          'secure;python_version<="2.7"': [
+          'secure:python_version<="2.7"': [
               'pyOpenSSL',
               'ndg-httpsclient',
               'pyasn1',
               'certifi',
           ],
-          'secure;python_version>"2.7"': [
+          'secure:python_version>"2.7"': [
               'certifi',
           ],
       },
--- a/setup.py
+++ b/setup.py
@@ -55,13 +55,10 @@ setup(name='urllib3',
       ],
       test_suite='test',
       extras_require={
-          'secure:python_version<="2.7"': [
+          'secure': [
               'pyOpenSSL',
               'ndg-httpsclient',
               'pyasn1',
-              'certifi',
-          ],
-          'secure:python_version>"2.7"': [
               'certifi',
           ],
       },


Lukasa:
Note that while this works it places some relatively new requirements on setuptools and pip. Given how widely used urllib3 is we may want to be a bit cautious there. Just a thought. Otherwise I'm +1.

shazow:
Will it explode for older versions of setuptools/pip, or just do the same ol' thing?

Lukasa:
For sufficiently old versions, it'll explode.

shazow:
What's a... sufficiently old verison? :P (cc @dstufft?)

shazow:
More specifically, will it explode any moreso than what we already have with incorrect syntax?

sigmavirus24:
> More specifically, will it explode any moreso than what we already have with incorrect syntax?

Yes. I think we might be able to move this to setup.cfg and get the benefits for wheels but not for `tar` files. In those we'll have to do stupid version checking hacks to not install things on versions of python we don't want them on.

dstufft:
pip < 6 and setuptools < uh, 0.7 I think.

shazow:
@dstufft what's your gut feeling about when we should merge this?

And yea I'm happy to merge a setup.cfg workaround if it's safer for now.

dstufft:
I think that pip and setuptools older than that won't understand the syntax and will just ignore it, so it depends on how important you think it is that ``pip install urllib3[secure]`` actually installs ``secure``. The cost of maintaining the ``setup.cfg`` work around is pretty small, but it relies on duplicating your dependencies so that's a possible downside.

shazow:
Given that we have naggy warnings if you don't have [secure], I think it's _reasonably_ okay if pip install urllib3[secure] doesn't succeed.

Wonder if we can print somekind of message on install to indicate that state. <- might be a great idea?

dstufft:
You can't really tell what version of pip you're being installed with.

shazow:
No, I mean based on the state of import availability of dependencies for [secure].

dstufft:
You probably don't want to depend on ``install_requires`` being installed before urllib3 itself is installed.

shazow:
Opened https://github.com/pypa/packaging-problems/issues/77

sigmavirus24:
Looks like wheel/setuptools don't have a way of parsing a setup.cfg for extras without using pbr.

sigmavirus24:
I just updated this to use setup.cfg. It took some work to figure out the syntax for setup.cfg but it now works. This should let older pip/setuptools to do almost the right thing when installing from an `sdist` while always doing the right thing when installing from a wheel. (Also, we should make sure we ship the setup.cfg in sdists)

shazow:
Looks good!

> (Also, we should make sure we ship the setup.cfg in sdists)

How do we do that? My release script is here: https://github.com/shazow/urllib3/blob/master/release.sh#L36

sigmavirus24:
Just checked an sdist and it is. So we're good (that means if someone uses an sdist and then uses a new enough pip, it will build a wheel for them. (The only downside with this is that people using different packages will get different results, but that's better than having a totally broken package.)

PR #708, opened 2015-09-16
We do not require square brackets for httplib.

### LINKED ISSUE #707: Error using https and ipv6: InvalidURL("nonnumeric port: '4f7b'",)
To reproduce:
```Python
import urllib3
p = urllib3.connection_from_url('https://[2001:0:53aa:64c:104c:2c10:2bef:4f7b]')
p.request('GET', '/')
```

produces:
```Python
  File "/usr/lib/python3.4/http/client.py", line 771, in _set_hostport
    port = int(host[i+1:])
ValueError: invalid literal for int() with base 10: '4f7b'

  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 548, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 250, in _get_conn
    return conn or self._new_conn()
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 770, in _new_conn
    strict=self.strict, **self.conn_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connection.py", line 171, in __init__
    timeout=timeout, **kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connection.py", line 119, in __init__
    _HTTPConnection.__init__(self, *args, **kw)
  File "/usr/lib/python3.4/http/client.py", line 750, in __init__
    self._set_hostport(host, port)
  File "/usr/lib/python3.4/http/client.py", line 776, in _set_hostport
    raise InvalidURL("nonnumeric port: '%s'" % host[i+1:])
http.client.InvalidURL: nonnumeric port: '4f7b'

  File "/usr/local/lib/python3.4/dist-packages/urllib3/request.py", line 68, in request
    **urlopen_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/request.py", line 89, in request_encode_url
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 629, in urlopen
    release_conn=release_conn, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 629, in urlopen
    release_conn=release_conn, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 629, in urlopen
    release_conn=release_conn, **response_kw)
  File "/usr/local/lib/python3.4/dist-packages/urllib3/connectionpool.py", line 609, in urlopen
  File "/usr/local/lib/python3.4/dist-packages/urllib3/util/retry.py", line 271, in increment
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='2001:0:53aa:64c:104c:2c10:2bef:4f7b', port=None): Max retries exceeded with url: / (Caused by ProtocolError('Connection aborted.', InvalidURL("nonnumeric port: '4f7b'",)))
```

Using `https://[2001:0:53aa:64c:104c:2c10:2bef:4f7b]:443` instead is a workaround.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26727615-error-using-https-and-ipv6-invalidurl-nonnumeric-port-4f7b?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #707.

I need to come up with a test for this that doesn't suck, but it should be considered illuminating that no tests break with this change.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -68,8 +68,7 @@ class ConnectionPool(object):
         if not host:
             raise LocationValueError("No host specified.")
 
-        # httplib doesn't like it when we include brackets in ipv6 addresses
-        self.host = host.strip('[]')
+        self.host = host
         self.port = port
 
     def __str__(self):


shazow:
@t-8ch You added this in 208ba74c796116bafc95d46dfcf4325449643ab6, any idea why? Is it safe to remove now?

shazow:
The original issue was https://github.com/shazow/urllib3/pull/199

Lukasa:
BTW, I legitimately have no good idea how to test this. I guess I could manually replace the socket to force httplib to do its validation but not actually use the socket, but...that doesn't feel good.

shazow:
According to https://github.com/shazow/urllib3/pull/199#issuecomment-19837893, it sounds like it happens on a specific Py3 version when port is not given? :/

Lukasa:
What's our supported set of Py3 versions? The oldest I have installed right now is Python 3.3.6, which does not reproduce the bug in #199.

sigmavirus24:
Pip supports 3.2 still
-- 

Lukasa:
Ugh, ok, I'll try to reproduce on 3.2.

Lukasa:
Python 3.2.6 also does not reproduce the bug from #199. Looks like this must have been fixed in a version of 3.2 prior to 3.2.6.

@dstufft What's the scope of your supported 3.2 versions? Put another way, how much effort should I be going to to track down a release that actually exhibits this bug?

dstufft:
What versions of X does pip support is always a sort of hard question. In this case I think that the only reason someone would be using Python 3.2 is either if they are using PyPy3 (which I'm not sure if pip even works on....) or if they are getting Python 3.2 from their OS. Typically our support for anything other than the latest version in a particular X.Y series is best effort based on how big of a deal the bug seems to be, and how hard the fix is.

Looking at the bug, it appears this only affects cases where someone is passing a raw IPv6 into urllib3? So for pip that would probably map to someone is using a raw ipv6 address as an ``--index-url``, ``--find-links``, or are using it in a link from a private index. I suspect that very few people are going to be switching to IPv6 *and* still be using Ubuntu Precise. There's a good chance the next major version of pip drops support for 3.2 anyways.

tl;dr I wouldn't spend much effort into it unless someone comes out of the wood work to explicitly complain.

Lukasa:
Thanks @dstufft. From my perspective that makes me want to merge this, but then again I proposed it so I'm naturally in favour. ;)

dstufft:
I don't have an opinion on the change itself, but I wouldn't let a nebulous concern about a <3.2.6 Python 3.2 prevent you.

dstufft:
If you're interested - https://groups.google.com/forum/#!topic/pypa-dev/Ef0PF2ZGAv0

shazow:
I've no feelings about this, @Lukasa totally your call, feel free to hit whichever buttons speaks to you.

sigmavirus24:
Yeah, 3.2 is broken in many ways with httplib. I'd say we just stop trying to support 3.2 in this case. :+1:

Lukasa:
Screw it, let's do it.

PR #711, opened 2015-09-23
Made SSL cert digest check use a constant-time comparison.

### LINKED ISSUE #710: SSL cert digest check uses non-constant-time comparison
This is the problematic line:
https://github.com/shazow/urllib3/blob/master/urllib3/util/ssl_.py#L137

The digests being compared are strings, and `==` is not constant-time, which could potentially allow a timing attack. While it may not immediately be obvious whether an adversary could gain any advantage by successfully timing this line, I am a firm believer in "better safe than sorry."

The best way to fix this is to use a constant-time digest comparison function, such as the built-in `hmac.compare_digest`. However, this function was only introduced in Python 2.7.7, while this library supports Python 2.6+. The next best solution would be to use something like the `constant_time_compare` function from the following package:
https://github.com/mitsuhiko/itsdangerous

Using a constant-time comparison is semantically equivalent to using `==`, and the performance impact is negligible and comes along with improved security. I believe this is a good tradeoff.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26908240-ssl-cert-digest-check-uses-non-constant-time-comparison?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

obi1kenobi:
Resolves #710.

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -17,6 +17,25 @@ HASHFUNC_MAP = {
 
 import errno
 import warnings
+import hmac
+
+
+def _const_compare_digest_backport(a, b):
+    """
+    Compare two digests of equal length in constant time.
+
+    The digests must be of type str/bytes.
+    Returns True if the digests match, and False otherwise.
+    """
+    result = abs(len(a) - len(b))
+    for l, r in zip(bytearray(a), bytearray(b)):
+        result |= l ^ r
+    return result == 0
+
+
+_const_compare_digest = getattr(hmac, 'compare_digest',
+                                _const_compare_digest_backport)
+
 
 try:  # Test for SSL features
     import ssl
@@ -134,7 +153,7 @@ def assert_fingerprint(cert, fingerprint):
 
     cert_digest = hashfunc(cert).digest()
 
-    if cert_digest != fingerprint_bytes:
+    if not _const_compare_digest(cert_digest, fingerprint_bytes):
         raise SSLError('Fingerprints did not match. Expected "{0}", got "{1}".'
                        .format(fingerprint, hexlify(cert_digest)))
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -153,7 +153,7 @@ In chronological order:
 
 * Jon Wayne Parrott <jonwayne@google.com>
   * App Engine environment tests.
-  
+
 * John Krauss <https://github.com/talos>
   * Clues to debugging problems with `cryptography` dependency in docs
 
@@ -168,6 +168,9 @@ In chronological order:
 
 * Jordan Moldow <https://github.com/jmoldow>
   * Fix low-level exceptions leaking from ``HTTPResponse.stream()``.
+
+* Predrag Gruevski <https://github.com/obi1kenobi>
+  * Made cert digest comparison use a constant-time algorithm.
 
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -1,3 +1,4 @@
+import hashlib
 import warnings
 import logging
 import unittest
@@ -18,6 +19,7 @@ from urllib3.util.url import (
 from urllib3.util.ssl_ import (
     resolve_cert_reqs,
     ssl_wrap_socket,
+    _const_compare_digest_backport,
 )
 from urllib3.exceptions import (
     LocationParseError,
@@ -412,3 +414,16 @@ class TestUtil(unittest.TestCase):
         ssl_wrap_socket(ssl_context=mock_context, sock=socket)
         mock_context.wrap_socket.assert_called_once_with(socket)
         ssl_.HAS_SNI = HAS_SNI
+
+    def test_const_compare_digest_fallback(self):
+        target = hashlib.sha256(b'abcdef').digest()
+        self.assertTrue(_const_compare_digest_backport(target, target))
+
+        prefix = target[:-1]
+        self.assertFalse(_const_compare_digest_backport(target, prefix))
+
+        suffix = target + b'0'
+        self.assertFalse(_const_compare_digest_backport(target, suffix))
+
+        incorrect = hashlib.sha256(b'xyz').digest()
+        self.assertFalse(_const_compare_digest_backport(target, incorrect))


shazow:
+1 to fixing this, but I'd rather not introduce a new dependency just for this.

obi1kenobi:
The dependency does not have any dependencies of its own, it's only a single file really.

Would you prefer if I pulled out the code from the dependency and inserted it directly into urllib3?

dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+itsdangerous==0.24
```
shazow: This is only for dependencies used during development (testing, etc). It's not for consumers of the lib.

shazow:
@sigmavirus24 thoughts?

dev-requirements.txt
obi1kenobi: My bad, should the dependency file go in urllib3/contrib or urllib3/packages?

shazow: Let's figure out what we're doing about this first.

tiran:
I don't see the point to make the check constant time. What's your attack vector? In order to abuse a non-constant time comparison a server would have to trick a client into generation hundreds of thousands of millions requests. Each time the server has to generate a new cert. The hash of the new chert has to match the guessed part of the string.  Cert generation and signing is way too costly just to trick a HTTP client, too.

Also you should use the constant timing comparison function from the standard library. In case the function is not available you can roll your own in 5 lines or so.

obi1kenobi:
@tiran respectfully, I disagree with your point of view on the necessity of this. Just because you or I can't find an attack vector doesn't mean one doesn't exist. I'm not willing to bet the safety of all urllib3 users on my computer security skills. Are you?

I can change the PR to use the standard library function, and fall back to an included python implementation of a constant-time check, if that's the approach you'd prefer to take on fixing this.

shazow:
@obi1kenobi So assuming that you can exploit some user app to manufacture hundreds of thousands of consistent requests that let you narrow down your fingerprint assertion hash, what does that give an attacker? Will the attacker need to figure out how to manufacture a cert that collides with the fingerprint?

tiran:
I'm completely and absolutely sure that a constant timing comparison operator doesn't increase security of the SSL digest check. Let's take one step back and think why a constant timing comparison is required. It prevents side-channel attacks on the payload. Without it an attacker gets an advantage so she can guess the payload faster with less operations.

In this case I see three reasons why a constant comparison of the cert digest and a fingerprint don't increase security:

1. A malicious server must trick a client to generator hundred of thousands, maybe millions of request. that's highly unlikely
2.  A malicious server must generate a new cert for every request, get it signed by a CA and then hope that the hash of the signed cert is similiar to the data the attacker has guessed. With with full control over a CA's private key, signing takes long. Pre-image attacks on SHA-256 is not feasible. Even similar prefixes are hard to find (ask some Bitcoin miners)
3. Finally the cert and therefore also the hash of the cert are public knowledge. Everybody can grab the public cert from e.g. github.com and calculate its hash. There is no point to protect the hash with a constant timing comparison op.

obi1kenobi:
Let's reframe the discussion -- assuming I do not include any additional dependencies, and call the Python built-in function whenever possible, what is the downside to using a constant-time comparison?

obi1kenobi:
CI fails with errors that I don't think my code is causing:
```
ERROR: InvocationError: '/home/travis/build/shazow/urllib3/.tox/py32/bin/nosetests'
```

This new commit does not introduce any new dependencies, and would not cause any adverse side-effects. It uses the built-in function whenever it is available, and falls back to a well-commented comparison function otherwise. The fallback function will work both on Python 2.6 and 3.2.

I would really appreciate it if you could take another look.

shazow:
The CI is failing because we require 100% line test coverage and there are a bunch of unneeded branches in the PR which aren't exercised.

I'll accept a minimal change to allow for a constant comparison. I think you're over-generalizing here. There is an invariant in the method that the compared hashes will be of the same length so we don't need izip_longest. We also control the input, so no need to check that. Probably not worth making a crypto module just for this, either.

If you make it into a 1-3 liner private `urllib3.util.ssl_._const_compare_digest(a, b)`, that'd be fine.

obi1kenobi:
Fair enough, I'll update the PR as you suggested.

shazow:
@obi1kenobi Thanks!

@tiran Thanks for your input, too. Much appreciated.

I think this is an acceptable improvement if the amount of work/change involved is proportional, and keeping it simple should satisfy everyone. :)

urllib3/util/ssl_.py
```diff
@@ -17,6 +18,26 @@
...
+if hasattr(hmac, 'compare_digest'):
+    _const_compare_digest = hmac.compare_digest
+else:
+    def _const_compare_digest(a, b):
+        """
+        Compare two digests of equal length in constant time.
+
+        Returns True if the digests match, and False otherwise.
+        """
+        if len(a) != len(b):
```
shazow: We don't need this.

urllib3/util/ssl_.py
```diff
@@ -1,4 +1,5 @@
 from binascii import hexlify, unhexlify
+from urllib3.packages import six
```
shazow: from ..packages import six

shazow: Also, group it in the next import block.

urllib3/util/ssl_.py
```diff
@@ -17,6 +18,26 @@
...
+        """
+        Compare two digests of equal length in constant time.
+
+        Returns True if the digests match, and False otherwise.
+        """
+        if len(a) != len(b):
+            raise ValueError("Mismatched input lengths!")
+
+        result = 0
+        for l, r in six.moves.zip(bytearray(a), bytearray(b)):
```
shazow: How about:

```python
return sum((aa ^ bb) for aa, bb in zip(bytearray(a), bytearray(b))) == 0
```

shazow: This way we don't even need to stress about six.moves.zip, izip, or whatever.

obi1kenobi: The semantics of `zip` are different in Python 2 and 3, as far as I understand. In Python 2, it will pad the sequence with None values, whereas in Python 3 it will stop iterating when an input argument gets exhausted. That could result in 'abc' and 'ab' deemed matching, which is confusing and not the behavior of the built-in function we are trying to replace. I realize this shouldn't happen because different lengths should be impossible, but I don't think not worrying about six is worth this.

obi1kenobi: I also find the 1-liner somewhat more difficult to read, and I don't think the 3 lines saved change that much in the overall project.

shazow: Nah, it'll just truncate to the shortest value. Same thing that `six.moves.zip` will do. In py2, it's just not a generator, which I don't think we care about.

urllib3/util/ssl_.py
```diff
@@ -17,6 +18,23 @@
...
+    _const_compare_digest = hmac.compare_digest
+else:
+    def _const_compare_digest(a, b):
+        """
+        Compare two digests of equal length in constant time.
+
+        Returns True if the digests match, and False otherwise.
+        """
+        result = 0
+        for l, r in six.moves.zip_longest(bytearray(a), bytearray(b)):
```
shazow: This zip_longest exist in our version of six.

obi1kenobi: You mean it doesn't exist?

shazow: Er, yes.

urllib3/util/ssl_.py
```diff
@@ -17,6 +18,23 @@
...
+if hasattr(hmac, 'compare_digest'):
+    _const_compare_digest = hmac.compare_digest
+else:
+    def _const_compare_digest(a, b):
+        """
+        Compare two digests of equal length in constant time.
+
+        Returns True if the digests match, and False otherwise.
+        """
+        result = abs(len(a) - len(b))
```
shazow: Where did this come from?

urllib3/util/ssl_.py
```diff
@@ -17,6 +18,23 @@
...
+        for l, r in six.moves.zip(bytearray(a), bytearray(b)):
```
shazow: You're really not getting anything from six here. :) `zip(...)` will be identical for our purposes.

urllib3/util/ssl_.py
obi1kenobi: It prevents partial matches like 'abc' and 'ab' from being found to be equal, making the function's behavior equivalent to the built-in it replaces. If the lengths are not equal, `result` will be nonzero.

urllib3/util/ssl_.py
obi1kenobi: Python 2's zip will eagerly create a list, which is unnecessary here. I don't think six.moves.zip should be a problem, and it definitely seems to exist in the packaged version of six.py.

obi1kenobi: If you really dislike relying on six, I can replace it with a regular zip call. I just don't see why using the already-included package is bad.

shazow: Reason why having superfluous logic is bad: Something someday will rely on that undocumented/unexpected quirk, and when we refactor and get rid of it, things will break. (e.g. length checks)

Reason why using six unnecessarily is bad: Someday we're going to remove our usage of six, and fewer usages to remove is better.

This entire PR is an exercise in doing things that are not necessary, so allocating an extra tuple is not exactly contradictory with the spirit. :)

shazow: Then again, we are falling back on compare_digest, so that's not gonna save us.

obi1kenobi: The length check in my mind is removing a quirk -- the quirk being the different behavior on Python 2.6 and 2.7.7+. On 2.6, `_const_compare_digest('abcd', 'ab') == True`, whereas on 2.7.7+ it's False.

I'm not entirely convinced that not using six here is better, but that's your call to make -- I'll change it to plain zip.

shazow: Yea, I'm convinced about the minimal length check sadness. Update the zip and I'll merge it. :) Thanks!

obi1kenobi: Done! Thanks!

shazow:
Alright, now we need test coverage. :D

urllib3/util/ssl_.py
```diff
@@ -17,6 +17,23 @@
 
 import errno
 import warnings
+import hmac
+
+if hasattr(hmac, 'compare_digest'):
```
shazow: How about something like:

```python
def _const_compare_digest_backport(a, b):
    ...

_const_compare_digest = getattr(hmac, 'compare_digest', _const_compare_digest_backport)
```

This way we can have a test that exercises `_const_compare_digest_backport`?

obi1kenobi: Where would be a good place to put such a test?

shazow: Off the top of my head, probably `urllib3/test/test_util.py` or something nearby.

shazow: Just exercise that function directly, no need to worry about the request flow.

obi1kenobi:
Finally, all green. Let me know if you like it or if there's anything else you'd like me to change.

shazow:
Thanks!

obi1kenobi:
Thank you!

PR #713, opened 2015-10-04
Add optional support for Brotli content-encoding

Lukasa:
This pull request adds optional support for the [Brotli](http://google-opensource.blogspot.co.uk/2015/09/introducing-brotli-new-compression.html) compression algorithm used as a content-encoding, alongside the current support for gzip and deflate. Chromium have announced their [intention to implement](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/xdVm8c2GOMQ) Brotli, and Firefox 44 included [their support for Brotli](https://bugzilla.mozilla.org/show_bug.cgi?id=366559). Curl also [intend to implement Brotli](http://curl.haxx.se/docs/todo.html#More_compressions), but it'd be nice to beat @bagder to the punch for once!

Google claims that Brotli achieves 20%-26% higher compression ratios than other compression algorithms. They claim the reference implementation decodes just as quickly as deflate in zlib, but with denser compression than LZMA and bzip2.

This implementation currently relies on the hyper project's [brotlipy](https://github.com/python-hyper/brotlipy) sub-project, a CFFI-based wrapper around the Brotli reference implementation. For obvious reasons, as this adds a third-party dependency, the proposed enhancement is a `contrib` module.

Right now this contrib module monkeypatches its way into urllib3, but in the act of writing it I couldn't help but notice that we're really not very far from being able to have a 'formal' content-encoding plugin system. If that sounds like something worth having let me know, and I could write it and then refactor this contrib module to be the first user of that system. (On the other hand, it may simply not be worth it!)

In the short term, I don't think we should merge this just because I'm having trouble actually finding a site that uses Brotli! I think the best way to handle this is for me to offer an enhancement to httpbin that uses brotlipy to provide a `/brotli` endpoint, but that means I need to sit down and actually write that. That's blocked behind python-hyper/brotlipy#2, but I can hopefully get around to that soonish.

### DIFF
--- a/urllib3/contrib/brotlipy.py
+++ b/urllib3/contrib/brotlipy.py
@@ -1,0 +1,55 @@
+"""
+Support for Brotli compression as content-encoding.
+
+This needs the brotlipy package installed from PyPI. You can install it with
+the following command:
+
+    pip install brotlipy
+
+To activate Brotli support, call
+:func:`~urllib3.contrib.brotli.inject_into_urllib3` from your Python code
+before you start making HTTP requests, like this::
+
+    try:
+        import urllib3.contrib.brotlipy
+        urllib3.contrib.brotlipy.inject_into_urllib3()
+    except ImportError:
+        pass
+
+Now you can use :mod:`urllib3` as you normally would, and it will support
+Brotli content-encoding when the required modules are installed.
+"""
+import brotli
+
+from .. import response
+from ..util import request
+
+# Original _get_decoder method.
+original_decoder = response._get_decoder
+
+
+# New _get_decoder method.
+def _new_decoder(mode):
+    if mode == 'brotli':
+        return BrotliDecoder()
+
+    return original_decoder(mode)
+
+
+class BrotliDecoder(object):
+    def __init__(self):
+        self._obj = brotli.Decompressor()
+
+    def __getattr__(self, name):
+        return getattr(self._obj, name)
+
+    def decompress(self, data):
+        if not data:
+            return data
+        return self._obj.decompress(data)
+
+
+def inject_into_urllib3():
+    response.HTTPResponse.CONTENT_DECODERS.append('brotli')
+    request.ACCEPT_ENCODING += ',brotli'
+    response._get_decoder = _new_decoder


shazow:
> Right now this contrib module monkeypatches its way into urllib3, but in the act of writing it I couldn't help but notice that we're really not very far from being able to have a 'formal' content-encoding plugin system. If that sounds like something worth having let me know, and I could write it and then refactor this contrib module to be the first user of that system.

+1 to that, if only to replace the 3 lines of undocumented monkeypatching with a 1 line that we'll commit to not breaking the API of. :P

bagder:
Glad to help (by not doing anything) =)

Lukasa:
Ok, time to define a new interface!

Lukasa:
My gift to you: one shiny plug-in architecture for content-encodings.

Lukasa:
And I just did a quick test against [a custom branch](https://github.com/Lukasa/httpbin/tree/brotli) of `httpbin` with Brotli support on it: this appears to work correctly.

Lukasa:
I've opened a PR against httpbin for their Brotli endpoint given that it seems like it'd be useful, but I'm not enormously confident that it'll get merged. See Runscope/httpbin#254.

shazow:
HMMM I feel this might be a bit much. Some thoughts for debate/feedback:

- This meta-framework for `DECODING_ERRORS` feels out of scope for what we'd like to maintain. Is it unreasonable to have a contract that if you're going to register a decoder, raise exceptions that we already expect? (Rather than having our code mangle them for you)
- Per-request sort/joining feels unnecessary, we should probably be pre-computing the default accept_encoding during registry.
- s/encodings.py/encoding.py/ (the rest of our things are singular); also _compression_ might be a better name, since encoding is fairly overloaded in python-land (ascii, unicode, etc).
- dystopic thought: there's going to be a user who is going to want to toggle two different encoding configurations between every request, lol.

Lukasa:
So that feedback is mostly easy enough, I'll take a swing tomorrow. For `DECODING_ERRORS`: yeah, it might be a bit much. If you think so, my proposal would be to create a new exception for internal use and wrap inside the decoder classes so they raise appropriately.

shazow:
> If you think so, my proposal would be to create a new exception for internal use and wrap inside the decoder classes so they raise appropriately.

Yea that'd be alright. Alternatively, I'd be cool with just accepting some well-documented subset of standard exceptions like IOError or what have you.

shazow:
Thanks again for doing this. :) <3

Lukasa:
> Yea that'd be alright. Alternatively, I'd be cool with just accepting some well-documented subset of standard exceptions like IOError or what have you.

We could define that we catch `IOError`, and that way custom decoders need to either wrap or inherit from `IOError`. (We also catch `zlib.error`, but I don't think anyone is going to wrap/inherit from that)

shazow:
I'm still +1 to having our own DecompressError exception of somekind. Let's say we'll accept that and IOError for convenience, anything else we suggest custom decoders wrap in DecompressError.

Lukasa:
Sounds good to me.

Mattwmaster58:
Personally, I still think this PR would be immensely useful.

immerrr:
@theacodes This PR is important and I believe requires just a little push to get to the finish line. Brotli compression is already ubiquitous and to reap its benefits when using the (also ubiquitous) requests library we need urllib3 to support it.

edmorley:
It's worth noting that Node.js has just added native support for Brotli, which will presumably mean that the Node http-request-related packages will start to add support in the future too. See:
nodejs/node/pull/24938

sethmlarson:
I agree that this would add useful functionality. In order to be implemented we need someone willing and with the time that can follow a patch to completion. :) If you'd like to help out you can contribute and we'd help review your code.

immerrr:
@sethmlarson I have opened #1532. Your feedback is welcome :)

bagder:
@Lukasa better luck next time :grin:, curl got brotli support in 7.57.0, November 2017...

PR #714, opened 2015-10-05
allows user paths for ca_cert verify path

a-tal:
currently you get the following error:

```SSLError("bad ca_certs: '~/example.crt'", Error([('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'BIO_new_file', 'no such file'), ('x509 certificate routines', 'X509_load_cert_crl_file', 'system lib')],))```

### DIFF
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -1,4 +1,5 @@
 import datetime
+import os
 import sys
 import socket
 from socket import error as SocketError, timeout as SocketTimeout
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Lukasa** [comment]
Thanks for this!

Can we apply this change to `ca_cert_dir` as well?

File: urllib3/connection.py [hunk 1]
```diff
@@ -205,11 +206,17 @@ def set_cert(self, key_file=None, cert_file=None,
...
         self.cert_reqs = cert_reqs
-        self.ca_certs = ca_certs
-        self.ca_cert_dir = ca_cert_dir
         self.assert_hostname = assert_hostname
         self.assert_fingerprint = assert_fingerprint
 
+        if ca_certs:
+            ca_certs = os.path.expanduser(ca_certs)
+        if ca_cert_dir:
+            ca_cert_dir = os.path.expanduser(ca_cert_dir)
```
shazow: Hm this line needs test coverage now. :) Could rewrite it as one line if we'd prefer.

```diff
- self.ca_certs = ca_certs
- self.ca_cert_dir = ca_cert_dir
+self.ca_certs = ca_certs and os.path.expanduser(ca_certs)
+self.ca_cert_dir = ca_cert_dir and os.path.expanduser(ca_cert_dir)
```

a-tal: I'm good with that :+1:

shazow:
Thank you! :custard:

shazow:
Now that I think about it, we really don't expand any path settings anywhere else, not sure how to feel about that. :x

PR #716, opened 2015-10-11
contrib.rst: typo fix: automagically --> automatically

staranjeet:


Lukasa:
Thanks for this! However, I think the word "automagically" was chosen deliberately: see [the Wiktionary definition](https://en.wiktionary.org/wiki/automagical). Given that, I'm inclined to let the original commenter keep their wording.

PR #720, opened 2015-10-16
'raise_on_status' option (similar to 'raise_on_redirect')

WGH-:


Lukasa:
Just to clarify @WGH-, the effect of this change is not actually to throw exceptions in more situations but to throw them in fewer situations: that is, if a status code is in the forcelist, you can avoid throwing an exception. Is that correct?

WGH-:
Yeah, exactly. As of now, once retry limit is hit by response with return code in `status_forcelist`, exception `MaxRetryError` is always raised.
My proposed change adds an option to suppress this exception, without changing the defaults, just like `raise_on_redirect` does.

Lukasa:
I see. Well, from that perspective I suppose it's reasonable enough, though I know that @shazow is concerned about the number of parameters that now go on urlopen. I'm going to let him make the final call here.

urllib3/connectionpool.py
```diff
@@ -652,7 +652,15 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
 
         # Check if we should retry the HTTP response.
         if retries.is_forced_retry(method, status_code=response.status):
-            retries = retries.increment(method, url, response=response, _pool=self)
+            try:
+                retries = retries.increment(method, url, response=response, _pool=self)
```
shazow: Hmm looks like we're calling retries.increment(...) an extra time with this change than before. Was this a bug before where we weren't incrementing during retries, or is it a bug now that we're double-incrementing?

shazow:
@kevinburke You helped design the retry API, any feelings on this?

urllib3/connectionpool.py
WGH-: I don't see where that extra time is. I merely wrapped increment with try...except

shazow: Ah oops you're right, misread the diff.

WGH-:
Bump.

shazow:
If there are no objections, I'm merging this next week. (cc @kevinburke @sigmavirus24)

sigmavirus24:
I'm :+1: on this.

kevinburke:
Hey! Whatever you all decide is fine by me!

On Friday, January 15, 2016, Ian Cordasco <notifications@github.com> wrote:

> I'm [image: :+1:] on this.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shazow/urllib3/pull/720#issuecomment-172148610>.
>


-- 

--
Kevin Burke
925.271.7005 | kev.inburke.com

PR #722, opened 2015-10-19
There's no reason not to preload chunked bodies.

### LINKED ISSUE #718: After read timeout, connection broken for a second timeout period
Using urllib3 1.12 to access etcd with a HTTPConnectionPool.  I do a long-poll read in a retry loop:
```
http = HTTPConnectionPool("localhost", 4001, maxsize=2)
while True:
    try:
        resp = http.request("GET", "http://localhost:4001/v2/keys/calico/v1",
                            fields={"recursive": "true", "wait": "true",
                                    "waitIndex": next_index},
                            timeout=5)
        resp_body = loads(resp.data)
    except ReadTimeoutError:
        _log.info("Watch read timed out, restarting watch at index %s",
                  next_index)
        continue
    else:
        ...
```
I expect some reads to time out because there was no data available.  In that case, I expect to hit the ReadTimeoutError, loop again and retry.  If there's then data waiting, I should get it immediately.

I do see the ReadTimeoutError after the first failure.  However, when my code retries, I see a 5s delay and then this log:
```
2015-10-15 09:38:16,311[DEBUG]273: Incremented Retry for (url='http://localhost:4001/v2/keys/calico/v1?waitIndex=7611342&recursive=true&wait=true'): Retry(total=2, connect=None, read=None, redirect=None)
2015-10-15 09:38:16,311[WARNING]625: Retrying (Retry(total=2, connect=None, read=None, redirect=None)) after connection broken by 'ReadTimeoutError("HTTPConnectionPool(host='localhost', port=4001): Read timed out.",)': http://localhost:4001/v2/keys/calico/v1?waitIndex=7611342&recursive=true&wait=true
```
Only then do I receive a response to my retry.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27428129-after-read-timeout-connection-broken-for-a-second-timeout-period?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #718.

This was introduced during our work on chunked transfer encoding. It's actually a fairly dramatic functional change, and potentially moves the point of exceptions being thrown quite dramatically, as well as changes how long each function call will take. However, I think it's more in line with our internal implementation and our documented behaviour.

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -130,8 +130,8 @@ class HTTPResponse(io.IOBase):
         if "chunked" in encodings:
             self.chunked = True
 
-        # We certainly don't want to preload content when the response is chunked.
-        if not self.chunked and preload_content and not self._body:
+        # If requested, preload the body.
+        if preload_content and not self._body:
             self._body = self.read(decode_content=decode_content)
 
     def get_redirect_location(self):


shazow:
Thanks! Wonder if we can get rid of the notion of preloading altogether in urllib3 v2.x or something and just force people to use context managers to mark the scope of a connection.

Lukasa:
That seems like a solid design pattern to me, it'd be an idea worth pursuing.

PR #724, opened 2015-10-20
Sync fetch_gae_sdk.py with oauth2client

jayvdb:
Fixes Python 3 syntax errors, and Python 2.6 str.format incompatibilities.


shazow:
Thanks!

PR #725, opened 2015-10-21
Fix flake8 violations

jayvdb:
Add tox rule to prevent regressions.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,7 @@ cover-min-percentage=100
 cover-erase=true
 
 [flake8]
+exclude=./docs/conf.py,./test/*,./urllib3/packages/*
 max-line-length=99
 
 [wheel]
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -167,7 +167,7 @@ class HTTPHeaderDict(MutableMapping):
     def __ne__(self, other):
         return not self.__eq__(other)
 
-    if not PY3: # Python 2
+    if not PY3:  # Python 2
         iterkeys = MutableMapping.iterkeys
         itervalues = MutableMapping.itervalues
 
@@ -304,7 +304,7 @@ class HTTPHeaderDict(MutableMapping):
         return list(self.iteritems())
 
     @classmethod
-    def from_httplib(cls, message): # Python 2
+    def from_httplib(cls, message):  # Python 2
         """Read headers from a Python 2 httplib message object."""
         # python2.7 does not expose a proper API for exporting multiheaders
         # efficiently. This function re-reads raw lines from the message
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -7,18 +7,13 @@ import warnings
 from .packages import six
 
 try:  # Python 3
-    from http.client import HTTPConnection as _HTTPConnection, HTTPException
+    from http.client import HTTPConnection as _HTTPConnection
+    from http.client import HTTPException  # noqa: unused in this module
 except ImportError:
-    from httplib import HTTPConnection as _HTTPConnection, HTTPException
-
-
-class DummyConnection(object):
-    "Used to detect a failed ConnectionCls import."
-    pass
-
+    from httplib import HTTPConnection as _HTTPConnection
+    from httplib import HTTPException  # noqa: unused in this module
 
 try:  # Compiled with SSL?
-    HTTPSConnection = DummyConnection
     import ssl
     BaseSSLError = ssl.SSLError
 except (ImportError, AttributeError):  # Platform-specific: No SSL.
@@ -60,6 +55,11 @@ port_by_scheme = {
 }
 
 RECENT_DATE = datetime.date(2014, 1, 1)
+
+
+class DummyConnection(object):
+    """Used to detect a failed ConnectionCls import."""
+    pass
 
 
 class HTTPConnection(_HTTPConnection, object):
@@ -266,8 +266,8 @@ class VerifiedHTTPSConnection(HTTPSConnection):
                 )
             match_hostname(cert, self.assert_hostname or hostname)
 
-        self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED
-                            or self.assert_fingerprint is not None)
+        self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED or
+                            self.assert_fingerprint is not None)
 
 
 if ssl:
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -10,7 +10,8 @@ try:  # Python 3
     from queue import LifoQueue, Empty, Full
 except ImportError:
     from Queue import LifoQueue, Empty, Full
-    import Queue as _  # Platform-specific: Windows
+    # Queue is imported for side effects on MS Windows
+    import Queue as _unused_module_Queue  # noqa: unused
 
 
 from .exceptions import (
@@ -22,7 +23,6 @@ from .exceptions import (
     LocationValueError,
     MaxRetryError,
     ProxyError,
-    ConnectTimeoutError,
     ReadTimeoutError,
     SSLError,
     TimeoutError,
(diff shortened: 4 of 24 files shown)


urllib3/connectionpool.py
```diff
@@ -10,7 +10,8 @@
     from queue import LifoQueue, Empty, Full
 except ImportError:
     from Queue import LifoQueue, Empty, Full
-    import Queue as _  # Platform-specific: Windows
+    # Queue is imported for side effects on MS Windows
+    import Queue as _unused_module_Queue  # noqa: unused
```
jayvdb: note that just `_` meant that on line 186 the variable `_` in `for _ in xrange(maxsize):` was shadowing this symbol.

shazow:
Thanks, I appreciate the thought here. :)

Some requests for changes:
- vendored packages/ like six.py (and maybe contrib/? we typically have lower standards for these) should be omitted.
- Some code is bundled in logical sections (like in the `__init__.py`) and unbundling it into standard module format would hurt more than help. I could be convinced otherwise, though.

The rest of the changes are actually quite good, especially consistency in spacing. Not a big fan of the `noqa` markers but there are few enough that I can live with it. :)

test/benchmark.py
```diff
@@ -39,7 +39,7 @@ def urllib_get(url_list):
     assert url_list
     for url in url_list:
         now = time.time()
-        r = urllib.urlopen(url)
```
jayvdb: these test/ changes are unnecessary as they are excluded.  looking at them again, these changes are probably not worth doing as part of this changeset.

setup.cfg
```diff
@@ -7,8 +7,13 @@ cover-min-percentage=100
 cover-erase=true
 
 [flake8]
+exclude=./docs/conf.py,./test/*
```
jayvdb: happy to also exclude dummyserver/, as it is mostly part of tests.  however it is quite clean, and could be re-used by other projects, so it is nice to provide clean code for other projects to copy and re-use.

urllib3/_collections.py
```diff
@@ -167,7 +167,7 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
-    if not PY3: # Python 2
+    if not PY3:  # Python 2
```
jayvdb: do we want to keep these comments on the same line as `PY3`?  They dont add much.

shazow: They're used for omitting platform-specific coverage (see .coveragerc).

urllib3/connectionpool.py
```diff
@@ -22,7 +23,6 @@
     LocationValueError,
     MaxRetryError,
     ProxyError,
-    ConnectTimeoutError,
```
jayvdb: this, and line 38, change the external API of this module.  Not sure whether that is OK.
The alternative would be to use `noqa`, or add `__all__`.

shazow: I think it's fine.

jayvdb:
I was almost going to propose putting `six` on the exclude list, so no drama from me.  I wouldnt be bothered by excluding all of ` urllib3/packages/` ; very little changes occurring there, and no doubt these issues would be resolved over time when they are sync'd with upstream.

I would advise against lowering the standards in `contrib/`, as urllib has a very critical dependency on parts of it, and undergo a lot of change.

I also was tossing up the idea of excluding `./__init__.py` or the import order error code, and I was guessing they would be the most controversial.  But we may as well have a discussion about those changes.
IMO moving `import warnings` up to the top is quite crucial for pep8 -- easy access to dependencies, _and_ imports in code are always suspect to readers who wonder whether earlier code or the import has side effects.
Without `__all__`, more #noqa would be needed.
I prefer module metadata is placed above imports, so no objections from me with reverting that part.

shazow:
On second look, the changes to `__init__.py` are probably fine.

@Lukasa Do you remember if contrib/pyopenssl.py was copypasta at all? Are we gonna regret fiddling with the formatting here?

jayvdb:
afaics, OpenSSL support was added by @kirkeby in 953b5a41e .

Lukasa:
AFAIK, contrib/pyopenssl.py is custom for this module. Even if it wasn't when it was originally introduced, it sure as hell is now.

Aside from removing changes to the vendored modules, though, I think this is entirely reasonable.

setup.cfg
```diff
@@ -7,8 +7,13 @@ cover-min-percentage=100
 cover-erase=true
 
 [flake8]
+exclude=./docs/conf.py,./test/*,./urllib3/packages/*
 max-line-length=99
 
+[pep8]
```
sigmavirus24: This shouldn't need to be here.

shazow:
Coolio.

urllib3/connection.py
```diff
@@ -62,6 +57,14 @@ class ConnectionError(Exception):
 RECENT_DATE = datetime.date(2014, 1, 1)
 
 
+class DummyConnection(object):
+    "Used to detect a failed ConnectionCls import."
+    pass
+
+
+HTTPSConnection = DummyConnection
```
shazow: I think we can get rid of this assignment, we override it twice more below anyways.

shazow:
+1, thank you for doing this! Ping me when the feedback is addressed and I'll merge.

Btw, I welcome you to add yourself to our CONTRIBUTORS.txt file. :)

setup.cfg
jayvdb: Not needed; I usually do it so people can run pep8 directly and see reasonable results.

urllib3/connection.py
jayvdb: Agreed.

I must admit, I am not thrilled with the `HTTPSConnection` class being hidden at the end of the module.  Any objection to the `HTTPSConnection` class being renamed to `_HTTPSConnection` or `UnverifiedHTTPSConnection`?  Or simply merged it into `VerifiedHTTPSConnection`.

I could do that as a separate earlier commit as part of this pull request.

urllib3/_collections.py
jayvdb: That can be avoided by adding appveyor builds covering the win32 specific code paths, and merging all job coverage on codecov.io which can notify the total coverage across all jobs.

shazow: Will it work in local dev environments?

urllib3/connection.py
shazow: HTTPSConnection is a publicly exposed API component.

setup.cfg
sigmavirus24: So now any time we update one of these stanzas we have to update the other. How convenient

urllib3/connection.py
jayvdb: I was referring to the class `HTTPSConnection` defined on line ~168, which is the parent class of `VerifiedHTTPSConnection`.

on lines ~279 the name `HTTPSConnection` is assigned the value of `VerifiedHTTPSConnection` or `DummyConnection`, and therefore the class defined on ~168 is hidden except in the bases of `VerifiedHTTPSConnection`.  That is a little confusing.(and causes conversations like this one.)
only the value assigned on lines ~279 is part of the public API.

setup.cfg
jayvdb: Through your narky comment, I get this very strong sense that you consider that duplication to be more important than allowing pep8 to be used directly.  No worries.  As I said, this is not needed, and I'll remove it in the next update.

urllib3/_collections.py
jayvdb: Depends on your definition of 'work'.

If you mean that local dev environments showing 100% reports obtained after running only a single test run on a single environment, then ... no. ;-)

However we might be able to come closer to achieving that broad objective, if a developer uses tox instead of nosetests directly.  https://github.com/shazow/urllib3/pull/726 provides a proof of concept.

jayvdb:
@shazow ping ;-)

shazow:
Thanks for the ping. Merged! :)

PR #726, opened 2015-10-22
[POC] Tox combined coverage report

jayvdb:
Store the coverage data from each tox environment,
providing coverage data for each Python version,
and combine them into a final report which requires
100% coverage.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,14 +2,19 @@
 omit =
     urllib3/packages/*
     urllib3/contrib/*
+    test/*
+    dummyserver/*
+    */site-packages/*
+    /usr/bin/*
 
 [report]
+show_missing = true
 exclude_lines =
     except ImportError:
     pass
     import
     raise
     .* # Platform-specific.*
-    .*:.* # Python \d.*
+    .*:.* # Python 2.6.4
     .* # Abstract
     .* # Defensive:
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,6 +3,8 @@ nose-exclude==0.4.1
 mock==1.3.0
 coverage==3.7.1
 tox==2.1.1
+pathlib
+reFILE
 twine==1.5.0
 wheel==0.24.0
 tornado==4.2.1
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,15 @@
 [tox]
-envlist = py26, py27, py32, py33, py34, pypy
+envlist = py26, py27, py32, py33, py34, pypy, coveragereport
 
 [testenv]
 deps= -r{toxinidir}/dev-requirements.txt
+whitelist_externals=
+    mkdir
+    refile
 commands=
     nosetests []
+    mkdir -p {distshare}/coverage/urllib3
+    refile mv -m {distshare}/coverage/urllib3 '\.coverage$' "{envname}.dat" .
 setenv =
     PYTHONWARNINGS=always::DeprecationWarning
 
@@ -13,6 +18,9 @@ basepython = python2.7
 deps= 
     {[testenv]deps}
     -r{toxinidir}/test/appengine/requirements.txt
+whitelist_externals=
+    mkdir
+    refile
 commands=
     nosetests \
         -c {toxinidir}/test/appengine/nose.cfg \
@@ -24,6 +32,18 @@ commands=
         -c {toxinidir}/test/appengine/nose.cfg \
         test/contrib/test_gae_manager.py \
         []
+    mkdir -p {distshare}/coverage/urllib3
+    refile mv -m {distshare}/coverage/urllib3 '\.coverage$' "{envname}.dat" .
 setenv =
     PYTHONPATH={env:GAE_PYTHONPATH:}
     {[testenv]setenv}
+
+[testenv:coveragereport]
+deps= -r{toxinidir}/dev-requirements.txt
+whitelist_externals=
+    mkdir
+    refile
+commands=
+    refile mv -m . '^(.*)$' '.coverage.\1' {distshare}/coverage/urllib3
+    coverage combine
+    coverage report  --fail-under=100 --rcfile={toxinidir}/.coveragerc
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,9 @@
 *.egg
 *.egg-info
 *.log
+.coverage
+.coverage.*
+coverage.xml
 /dist
 /build
 /docs/_build
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,9 +2,9 @@
 logging-clear-handlers=true
 with-coverage=true
 cover-package=urllib3
-# no rounding problems here
-cover-min-percentage=100
-cover-erase=true
+# tox.ini requires 100% coverage
+# Individual test runs can not be used to determine
+# the complete coverage.
 
 [flake8]
 max-line-length=99


.coveragerc
```diff
@@ -2,14 +2,19 @@
...
 [report]
+show_missing = true
 exclude_lines =
     except ImportError:
     pass
     import
     raise
     .* # Platform-specific.*
-    .*:.* # Python \d.*
+    .*:.* # Python 2.6.4
```
jayvdb: This is needed until a 2.6.4 build is added to the build matrix.

.coveragerc
```diff
@@ -2,14 +2,19 @@
...
+    /usr/bin/*
 
 [report]
+show_missing = true
 exclude_lines =
     except ImportError:
     pass
     import
     raise
     .* # Platform-specific.*
```
jayvdb: This can also be replaced with more specific exclusions, and/or remove this comment from areas that are covered by the tox envlist matrix.  especially the 'AppEngine' code removed from coverage.

dev-requirements.txt
```diff
@@ -3,6 +3,8 @@ nose-exclude==0.4.1
 mock==1.3.0
 coverage==3.7.1
 tox==2.1.1
+pathlib
```
jayvdb: this is a bug in reFILE; in the packaging I suspect (https://github.com/Sean1708/reFILE/issues/1)

dev-requirements.txt
```diff
@@ -3,6 +3,8 @@ nose-exclude==0.4.1
...
+reFILE
```
jayvdb: happy to switch this with an alternative tool.

tox.ini
```diff
@@ -24,6 +32,18 @@ commands=
...
+    mkdir -p {distshare}/coverage/urllib3
+    refile mv -m {distshare}/coverage/urllib3 '\.coverage$' "{envname}.dat" .
 setenv =
     PYTHONPATH={env:GAE_PYTHONPATH:}
     {[testenv]setenv}
+
+[testenv:coveragereport]
+deps= -r{toxinidir}/dev-requirements.txt
+whitelist_externals=
+    mkdir
```
jayvdb: whoops; not needed here.

.coveragerc
```diff
@@ -2,14 +2,19 @@
 omit =
     urllib3/packages/*
     urllib3/contrib/*
+    test/*
```
jayvdb: I found these useful when combining and reporting coverage data, otherwise the final report would include many undesirable packages.

Lukasa:
How does this affect Travis?

jayvdb:
This follows from a discussion started on https://github.com/shazow/urllib3/pull/725 .  It has been on my todo list for a different project (https://phabricator.wikimedia.org/T74863), but it seemed worth exploring it here.

Immediate affect on Travis is none (other than the dependency on reFILE, which needs to be is fixed one way or another).

Ideally coverage data from Travis and Appveyor is sent to a service that can combine the coverage data, and report back to pull requests if the _combined_ coverage reaches 100%.  So far, https://codecov.io/ looks like the best option.

e.g. https://github.com/codecov/codecov-python/pull/40 shows the coverage report in github, and on the Unified coverage report of that commit (https://codecov.io/github/codecov/codecov-python/codecov/__init__.py?ref=963697ffe8ea863986ea21ffce896751d08d85d9) you will see all platform specific branches of the code are show as green , but they are not covered on the respective individual job coverage reports (e.g. https://codecov.io/github/codecov/codecov-python/codecov/__init__.py?ref=963697ffe8ea863986ea21ffce896751d08d85d9&build=369.1 )

The outstanding issue with this model, IMO, is that travis and codecov.io do not offer the coverage data artifacts for download freely, so there is an element of lock-in which I dont like.  But there are workarounds that I am exploring.

shazow:
I'm a strong -1 for anything that doesn't work in a purely-offline context. I do a lot of my dev on urllib3 while offline and I'd hate to depend on an online service to get the job done. :)

sigmavirus24:
== @shazow on this, being able to get the feedback locally is as important as it is to be able to get it from a CI service.

jayvdb:
@shazow & @sigmavirus24 ...
**This** patch provides the desired feedback locally.  This patch has nothing to do with a CI service.

It would remove '100% coverage feedback' from an individual `nosetests` run locally.  This has been inaccurate feedback.  No doubt you are well aware of those inaccuracies, and work with those inaccuracies in your mind when reading the `nosetests` coverage reports, but it still allows slippages in the real coverage, as the coverage report is explicitly omitting many branches.

With this changeset, to achieve such feedback, instead of relying on `nosetests` output the developer would need to use `tox`, or slightly quicker : `tox -epy26,py27,py33,coveragereport` , as that is sufficient to provide the desired combined coverage.
It is set up so that coverage data is stashed locally, so that after the initial coverage data has been stashed for all platforms, running `tox -epy33,coveragereport` will generate combined updated coverage report.

tox.ini
```diff
@@ -24,6 +32,18 @@ commands=
...
+    refile
+commands=
+    refile mv -m . '^(.*)$' '.coverage.\1' {distshare}/coverage/urllib3
```
jayvdb: note this does a `mv` because `cp` doesnt exist yet https://github.com/Sean1708/reFILE/issues/4 .
This is not ideal, as it means that re-running `tox -epy34,coveragereport` will only show Python 3.4 coverage.
`refile` was the best tool that I could quickly find, but I'd be more than happy to switch to an alternative, or I'll add `cp` support to `refile` before this changeset moves out of the `POC` stage.

shazow:
It's not _inaccurate_ so much that we've whitelisted platform-specific branches. Code that we want exercised is still 100% exercised. I really don't see a way around this. To get full branch coverage, it would need to run on a permutated matrix of {multiple versions of windows, linux, multiple versions of osx, every version of python, with and without +ssl compiled, with and without pyopenssl} and probably some other cases I'm not remembering.

At best, we'll get slightly improved yet still incomplete coverage with a lot of changes/work in the process. :/

jayvdb:
This changeset is only a proof of concept of one possible way to achieve cross-version branch coverage, by using `tox` instead of `nosetests` to provide the 100% coverage assurance locally.

I agree that some branches will be very difficult to provide accurate coverage for, as they would be prohibitively difficult for a developer to set up suitable environments in order to collect coverage for a few extra lines.  This changeset doesn't attempt to do that, and is not advocating for that.  Perfect is the enemy of good.  Using coverage omit or pragma's is more sensible for those hard to reach branches.

This changeset shows that, with minimal effort, using `tox` could eliminate a large cohort of branches currently being omitted (i.e. versions), and still achieve the desired 100% coverage assurance **locally** (and creating a path forward for CI-based verification of that assurance).  IMO that is striking the right balance, as tox makes it trivial and time-efficient for a developer to run tests in these version-based environments.

I can appreciate if you feel that cost/benefit is not the right balance for this project, based on its core teams needs.  This is just a proof of concept, and will be abandoned if the concept itself doesn't interest the core team.

shazow:
Right now we have a simple heuristic: If it's a platform-specific branch, annotate it and move on. With a system like this, we'll have a much more complex heuristic that will require much more investigation.

I'm not opposed having combined coverage analytics _in addition_ to what we already do, but I'm not convinced that this is easier/better in replacement to what we currently do.

May I propose a standalone script (or tox config) which gathers coverage while ignoring coveragerc across multiple tox points then collates it?

jayvdb:
Certainly, having separate configurations is a good approach.
It would require a little duplication, as I assume a combined coverage would still want many of the existing coverage `omit` and `exclude_line` entries to produce a meaningful report.

dev-requirements.txt
shazow: What do we need reFILE for?

shazow:
Cool. If you can package this up into a reasonably isolated PR, we can keep it around. :)

dev-requirements.txt
jayvdb: It is currently being used in `tox.ini` to stash the coverage data of each `nosetests` environment (simple `cp` could suffice instead), and then also in the `coveragereport` task to pull all of the stashed data back into the working directory and rename it so that coverage combined picks them all up (I have only done a quick search for a better approach to achieve this.).  Very open to suggestions on alternatives.

shazow: What's wrong with just using mv? e.g. `for name in *; do mv "$name" ".coverage.$name"; done`?

jayvdb: That could be done in an external script, but not inline in tox, and would not work on Windows.

jayvdb:
Im not going to get back to this any time soon, and it needs to be restarted from the ground up to make it acceptable.

shazow:
Thanks for giving it a try back then, sorry it never made it in. I'm sure this change will serve as an important reference in the future!

PR #729, opened 2015-10-27
Catch exceptions on connection close

### LINKED ISSUE #728: conn.close() can raise OpenSSL exception after idle timeout
When returning a connection to the pool, urllib3 calls `conn.close()`:

~~~
        if conn and is_connection_dropped(conn):
            log.info("Resetting dropped connection: %s" % self.host)
            conn.close()
~~~

This can raise an exception under the following conditions:

* urllib3 is using pyOpenSSL and https
* Remote webserver is running Windows IIS (7.5 and 8 tested)
* KeepAlives are active (default for IIS)
* Connection drops after idle timeout

tcpdump shows that the remote IIS host sends a RST on idle timeout. This appears to be IIS-specific behaviour (Apache2 and haproxy both close the connection normally)

At that point nothing bad happens immediately. However the connection remains in the pool. The next time you send a request to this host, it is reused; some data is sent down it; this then raises an `OpenSSL.SSL.SysCallError` exception. I am guessing that it it is trying to send an SSL shutdown message down the already-closed socket.

More detail at <https://github.com/zatosource/zato/issues/465>

Proposed patch:

~~~
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -239,11 +239,15 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # If this is a persistent connection, check if it got disconnected
         if conn and is_connection_dropped(conn):
             log.info("Resetting dropped connection: %s" % self.host)
-            conn.close()
-            if getattr(conn, 'auto_open', 1) == 0:
-                # This is a proxied connection that has been mutated by
-                # httplib._tunnel() and cannot be reused (since it would
-                # attempt to bypass the proxy)
+            try:
+                conn.close()
+                if getattr(conn, 'auto_open', 1) == 0:
+                    # This is a proxied connection that has been mutated by
+                    # httplib._tunnel() and cannot be reused (since it would
+                    # attempt to bypass the proxy)
+                    conn = None
+            except Exception as e:
+                log.info("Closing connection failed: %s, force new connection" % str(e))
                 conn = None

         return conn or self._new_conn()
~~~

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/27746311-conn-close-can-raise-openssl-exception-after-idle-timeout?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

candlerb:
Typically caused by remote IIS server dropping HTTPS keepalive
connections without shutting down properly

Fixes #728

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -217,7 +217,10 @@ class WrappedSocket(object):
 
     def close(self):
         if self._makefile_refs < 1:
-            return self.connection.close()
+            try:
+                return self.connection.close()
+            except OpenSSL.SSL.Error:
+                return
         else:
             self._makefile_refs -= 1
 


urllib3/connectionpool.py
```diff
@@ -239,11 +239,15 @@ def _get_conn(self, timeout=None):
...
-                # httplib._tunnel() and cannot be reused (since it would
-                # attempt to bypass the proxy)
+            try:
+                conn.close()
+                if getattr(conn, 'auto_open', 1) == 0:
+                    # This is a proxied connection that has been mutated by
+                    # httplib._tunnel() and cannot be reused (since it would
+                    # attempt to bypass the proxy)
+                    conn = None
+            except Exception as e:
```
shazow: We don't want to eat all exceptions, this makes finding bugs in the future really hard. What specific exception are we expecting?

urllib3/connectionpool.py
```diff
@@ -239,11 +239,15 @@ def _get_conn(self, timeout=None):
...
         if conn and is_connection_dropped(conn):
             log.info("Resetting dropped connection: %s" % self.host)
-            conn.close()
-            if getattr(conn, 'auto_open', 1) == 0:
-                # This is a proxied connection that has been mutated by
-                # httplib._tunnel() and cannot be reused (since it would
-                # attempt to bypass the proxy)
+            try:
+                conn.close()
+                if getattr(conn, 'auto_open', 1) == 0:
```
shazow: I don't think we need to wrap this if branch in the try/except too?

shazow: Nevermind.

urllib3/connectionpool.py
candlerb: `OpenSSL.SSL.SysCallError` is what you get with pyOpenSSL 0.15+: this was enabled in
<https://github.com/pyca/pyopenssl/commit/bff1d1ae72e743e41254bb31a5b50b5b63a37b85> [(changelog)](https://github.com/pyca/pyopenssl/blob/0.15/ChangeLog#L25)

With pyOpenSSL 0.14 I got some generic `Error` whose exact class I'm not sure of.

In general I'd agree to minimise catch/discard of errors. However in this particular case, it seems reasonable to catch Exception.  We are specifically only closing a connection where the underlying socket is known to have closed (`is_connection_dropped(conn)`), and we know we are going to discard it and create a new one.

candlerb:
I don't know why Travis failed. `make test` works for me in a venv, and the [specific error](https://travis-ci.org/shazow/urllib3/jobs/87755077) doesn't seem to be anything to do with this change:

~~~
No handlers could be found for logger "nose.plugins.cover"
~~~

urllib3/connectionpool.py
shazow: I disagree that it's a good idea to catch all errors and silently discard even in this case. We could have a syntax error in conn.close somehow, or a new scenario that we'd want to be aware of, or any number of conditions where this behaviour will hide what's going on underneath and make it a lot harder to debug.

We'll need to track down the specific exception and maybe add a wrapper that we can still catch even if pyopenssl is not available.

shazow:
Travis is failing due to missing test coverage.

urllib3/connectionpool.py
candlerb:
~~~
>>> import inspect
>>> inspect.getmro(OpenSSL.SSL.SysCallError)
(<class 'OpenSSL.SSL.SysCallError'>,
 <class 'OpenSSL.SSL.Error'>,
 <type 'exceptions.Exception'>,
 <type 'exceptions.BaseException'>,
 <type 'object'>)
~~~

Unfortunately not a child of `EnvironmentError`, but that might be a reasonable thing to catch as well, whether or not pyopenssl is available.

shazow: Hm, would it be feasible to add this wrapping into our pyopenssl wrapper? (contrib/pyopenssl.py)

Lukasa: Wrapping this is almost certainly what we want to do.

candlerb:
A fix in `contrib.pyopenssl` should work. Here is a backtrace:

~~~
...
  File "/opt/zato/2.0.5/eggs/urllib3-1.10.4-py2.7.egg/urllib3/contrib/pyopenssl.py", line 213, in close
    return self.connection.shutdown()
  File "/opt/zato/2.0.5/eggs/pyOpenSSL-0.15.1-py2.7.egg/OpenSSL/SSL.py", line 1522, in shutdown
    self._raise_ssl_error(self._ssl, result)
  File "/opt/zato/2.0.5/eggs/pyOpenSSL-0.15.1-py2.7.egg/OpenSSL/SSL.py", line 1178, in _raise_ssl_error
    raise SysCallError(errno, errorcode.get(errno))
OpenSSL.SSL.SysCallError: (104, 'ECONNRESET')
~~~

So I made this patch:

~~~
--- /opt/zato/2.0.5/eggs/urllib3-1.10.4-py2.7.egg/urllib3/contrib/pyopenssl.py.orig	2015-06-16 07:50:30.000000000 +0000
+++ /opt/zato/2.0.5/eggs/urllib3-1.10.4-py2.7.egg/urllib3/contrib/pyopenssl.py	2015-10-28 11:55:33.616085405 +0000
@@ -210,7 +210,10 @@

     def close(self):
         if self._makefile_refs < 1:
-            return self.connection.shutdown()
+            try:
+                return self.connection.shutdown()
+            except OpenSSL.SSL.SysCallError:
+                return
         else:
             self._makefile_refs -= 1
~~~

This is sufficient to stop the problem using the reproducing code posted to #728 

Alternatively you might want to catch the slightly wider `OpenSSL.SSL.Error`, which I believe is what pyOpenSSL 0.14 returned in this scenario.

shazow:
This looks like a much safer fix, thank you. @Lukasa any suggestions before I merge?

urllib3/contrib/pyopenssl.py
```diff
@@ -217,7 +217,10 @@ def shutdown(self):
 
     def close(self):
         if self._makefile_refs < 1:
-            return self.connection.close()
+            try:
+                return self.connection.close()
+            except OpenSSL.SSL.Error:
+                return
```
sigmavirus24: Do we want to return `None` here?

The rest of this function implicitly returns `None` (e.g., when `self._makefile_refs >= 1`) so why not just do:

```py
try:
    return self.connection.close()
except OpenSSL.SSL.Error:
    # Nothing to do here
    pass
```

candlerb: I used "return" rather than "pass" so that you don't need to look any further in the function to see that falling through does nothing, but it's just a matter of style.

As for return value, there were already circumstances in which `None` was returned implicitly. The only potentially useful return value is that from `self.connection.close()`, but I don't know if that ever returns anything, nor if anyone ever checks it.

Lukasa: I think with a body as short as this one `pass` or `return` are about equally clear. I don't object to the explicit indication that we return `None` in some cases. I think this is 50/50, so I'm inclined to just leave this as-is.

Lukasa:
@shazow This LGTM. I don't think close should ever really fail, so I'm happy with this.

urllib3/contrib/pyopenssl.py
shazow: Yea not sure which is better here, good enough!

shazow:
@candlerb Thanks for digging through this!

Issue #730, opened 2015-10-29
contrib/pyopenssl should be wrapping SysCallError


steveoh:
https://github.com/kennethreitz/requests/issues/2852

```
Traceback (most recent call last):
  File "..\requests-2.8.1-py2.7.egg\requests\api.py", line 122, in put
    return request('put', url, data=data, **kwargs)
  File "..\requests-2.8.1-py2.7.egg\requests\api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "..\requests-2.8.1-py2.7.egg\requests\sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "..\requests-2.8.1-py2.7.egg\requests\sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "..\requests-2.8.1-py2.7.egg\requests\adapters.py", line 370, in send
    timeout=timeout
  File "..\requests-2.8.1-py2.7.egg\requests\packages\urllib3\connectionpool.py", line 559, in urlopen
    body=body, headers=headers)
  File "..\requests-2.8.1-py2.7.egg\requests\packages\urllib3\connectionpool.py", line 376, in _make_request
    httplib_response = conn.getresponse(buffering=True)
  File "C:\Python27\ArcGIS10.3\lib\httplib.py", line 1067, in getresponse
    response.begin()
  File "C:\Python27\ArcGIS10.3\lib\httplib.py", line 409, in begin
    version, status, reason = self._read_status()
  File "C:\Python27\ArcGIS10.3\lib\httplib.py", line 365, in _read_status
    line = self.fp.readline(_MAXLINE + 1)
  File "C:\Python27\ArcGIS10.3\lib\socket.py", line 476, in readline
    data = self._sock.recv(self._rbufsize)
  File "..\requests-2.8.1-py2.7.egg\requests\packages\urllib3\contrib\pyopenssl.py", line 179, in recv
    data = self.connection.recv(*args, **kwargs)
  File "..\OpenSSL\SSL.py", line 1320, in recv
    self._raise_ssl_error(self._ssl, result)
  File "..\OpenSSL\SSL.py", line 1178, in _raise_ssl_error
    raise SysCallError(errno, errorcode.get(errno))
SysCallError: (10054, 'WSAECONNRESET')
```

Lukasa:
So, specifically the problem is `SysCallError` in our `recv` wrapper.

shazow:
[...]

PR #731, opened 2015-10-29
wrap error in named exception

steveoh:
refs #730


shazow:
LGTM. @sigmavirus24 @Lukasa will that satisfy the original requests bug?

Lukasa:
I believe so, yes.

shazow:
Thanks @steveoh !

PR #733, opened 2015-11-03
Add proxy authentication example using ProxyManager

bananos:
I've struggled a bit while trying to google an example of using basic proxy authentication using urllib3,  therefore decided to add a bit of code sample to ProxyManager docs.
Not sure if I've placed it to the right place, doctests in source code seemed to be not appropriate.

### DIFF
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -70,6 +70,17 @@ connections and individual per-server:port
 :class:`~urllib3.connectionpool.HTTPSConnectionPool` instances for tunnelled
 HTTPS connections.
 
+Example using proxy authentication:
+
+::
+
+	>>> headers = urllib3.make_headers(proxy_basic_auth='myusername:mypassword')
+	>>> proxy = urllib3.ProxyManager('http://localhost:3128', proxy_headers=headers)
+	>>> r = proxy.request('GET', 'http://example.com/')
+	>>> r.status
+	200
+
+
 API
 ---
     .. autoclass:: ProxyManager


docs/managers.rst
```diff
@@ -70,6 +70,21 @@ connections and individual per-server:port
...
 
+Example using proxy authentication:
+
+::
+
+	>>> import base64
+	>>> user = "user"
+	>>> passwd = "123123"
+	>>> headers = {'Proxy-Authorization': 'Basic '+base64.b64encode(user+':'+passwd) }
+	>>> proxy_url = 'http://'+user+':'+passwd+'@localhost:3128'
```
shazow: I don't think you need the user/passwd in the URL if you're using the `Proxy-Authorization` header?

bananos: My bad, you're correct.  Let me fix it.  
p.s. Why the build is failing?

Lukasa: Build failure is probably incidental, I'll take a look.

Lukasa: Yup, there we go. All green.

docs/managers.rst
```diff
@@ -70,6 +70,20 @@ connections and individual per-server:port
 :class:`~urllib3.connectionpool.HTTPSConnectionPool` instances for tunnelled
 HTTPS connections.
 
+Example using proxy authentication:
+
+::
+
+	>>> import base64
```
shazow: This would be simpler:

```python
>>> headers = urllib3.make_headers(proxy_basic_auth='myusername:mypassword')
>>> proxy = urllib3.ProxyManager('http://localhost:3128', proxy_headers=headers)
>>> r = proxy.request('GET', 'http://example.com/')
>>> r.status
200
```

bananos: Cool, didn't know about that shortcut.
Build is failing once again...

shazow: Don't worry about the "connection reset by peer" errors, that's just travisci machines being overloaded. :(

shazow:
Thank you!

PR #734, opened 2015-11-05
Defensively close connections

Lukasa:
Generally speaking, if we hit an exception when working with a connection we should assume that the connection is dead. We've not been defensive enough with this in the past, and that attitude pretty much just leads us to a constant cycle of bug-squashing each time we discover a new connection-invalidating exception.

In this case, the problem (and the fix) were inspired by [this HN post](https://news.ycombinator.com/item?id=10512343). For that reason, I'd like to credit @rkday (in his guise as @rkd-msw) for the actual fix. Rob, if you'd like to, I'm happy for you to open a new pull request with these commits in place. If you don't care about the commit log, I've made sure you're added to CONTRIBUTORS.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -181,5 +181,8 @@ In chronological order:
 * John Vandenberg <jayvdb@gmail.com>
   * Python 2.6 fixes; pyflakes and pep8 compliance
 
+* Andy Caldwell <andy.m.caldwell@googlemail.com>
+  * Bugfix related to reusing connections in indeterminate states.
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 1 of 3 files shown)


rkd-msw:
Credit where credit's due - this is @bossmc's fix, I merely blogged about it.

Lukasa:
Heh, fair enough. In that case, same offers to @bossmc: if you want to propose this patch against the repo yourself, Andy, let me know. I'd rather merge your work than mine (I have enough commits against this repo as it is).

bossmc:
I'm not too bothered about getting my name on the git commits (which would just be carbon-copies of your changes anyway).  If you really want me to own the PR, I'm happy to do so, otherwise propose/merge away!

Lukasa:
That's fine then, let's leave this as is. Your name is in the CONTRIBUTORS list nonetheless. =D

bossmc:
Woo!

shazow:
Looks safe, thank you everyone! <3

PR #735, opened 2015-11-05
Add absolute_import feature from the future

sigmavirus24:
This should prevent local files or modules from shadowing standard
library modules and will provide an all around better experience. This
does not, however, help us in the case where someone uses a library
like pies (which installs broken standard library shims for Python 2/
Python 3 compatibility).

Related to: https://github.com/kennethreitz/requests/issues/2771

### DIFF
--- a/urllib3/__init__.py
+++ b/urllib3/__init__.py
@@ -1,6 +1,7 @@
 """
 urllib3 - Thread-safe connection pooling and re-using.
 """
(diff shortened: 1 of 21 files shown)


Lukasa:
\o/

sigmavirus24:
:custard:

shazow:
:sparkles: :custard: :sparkles:

PR #737, opened 2015-11-10
fix race condition around ready_event (hopefully)

matejcik:
`test_timeout` and friends rely on `ready_event` event variable to indicate that a socket is (about to be) listening. But the set/clear sequence falls on its head if the test thread is not already waiting on `ready_event`.

Which, for some reason, happens really frequently in our build environment.

This patch seems to ensure that the test thread never gets locked out of the `ready_event`.

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -59,11 +59,6 @@ class SocketDummyServerTestCase(unittest.TestCase):
         return ready_event
 
     @classmethod
-    def wait_for_socket(cls, ready_event):
-        ready_event.wait()
-        ready_event.clear()
-
-    @classmethod
     def start_basic_handler(cls, **kw):
         return cls.start_response_handler(
             b'HTTP/1.1 200 OK\r\n'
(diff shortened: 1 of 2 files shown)


Lukasa:
This change *looks* reasonable to me, but @shazow should definitely think on it.

shazow:
Yea this part is a bit of a mess, I was aggressively trying to fix some flakiness issues so I shoved a bunch of events in there and it's definitely not optimal. I have some ideas for rewriting this to use a ContextManager that would make things a lot cleaner...

For now, I'd say let's get rid of the wait_for_socket classmethod helper in the base that is only being used in one place. Could make it a module helper in the respective test suite if you'd like, or make the wait/clear inline like the rest of the code is anyways. Otherwise, happy to merge.

matejcik:
my thinking with the classmethod helper was that it now forms a usage contract: before you want to use a socket with ready_event, you call wait_for_socket.
but i have no problem with moving that to a module helper, esp. if this is slated for rewrite anyway

shazow:
Hm I think this breaks the rest of our tests: https://travis-ci.org/shazow/urllib3/jobs/90915501

shazow:
Hmmm looks like travis is just freaking out in general.

matejcik:
hmm is this something i can fix? it doesn't look like something i can fix

sigmavirus24:
Restarting this build as well

sigmavirus24:
@shazow this is good to merge

shazow:
:boat:

PR #740, opened 2015-11-14
dev-requirements piprot bump

shazow:
Travis is failing a lot lately, let's try newer versions of things.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,8 +1,8 @@
 nose==1.3.7
 nose-exclude==0.4.1
 mock==1.3.0
-coverage==3.7.1
-tox==2.1.1
-twine==1.5.0
-wheel==0.24.0
+coverage==4.0.2
+tox==2.2.1
+twine==1.6.4
+wheel==0.26.0
 tornado==4.2.1


sigmavirus24:
So I was able to reproduce this (by upgrading to 2.2.1). If I remove the following line this problem disappears:

```
PYTHONPATH={env:GAE_PYTHONPATH:}
```

I'm going to find a version of tox that works and file a bug against tox about what appears to be a regression.

shazow:
@sigmavirus24 <3

sigmavirus24:
So this is a regression between 2.1.x and 2.2.x. The problem happens on 2.2.0 and 2.2.1 but not 2.1.1 or prior.

sigmavirus24:
Ah it's already been reported: https://bitbucket.org/hpk42/tox/issues/285/tox-220-breaks-some-toxini-config-files

sigmavirus24:
So 2 things:

1. I'm confused as to how we're intending dev-requirements.txt to be used. It looks like we're using it as test-requirements.txt instead of dev-requirements.txt

2. We're installing tox inside a virtualenv managed by tox in this set-up. Should we separate test and dev requirements so dev-requirements installs tox and whatever else we think developers need and test-requirements is used inside of tox?

shazow:
We don't have a test-requirements.txt; basically dev-requirements.txt has extra deps that we're expecting developers working on urllib3 to use/install in order to have a working environment that can run all the tests, generate coverage, etc.

shazow:
Will push this off for another day (prolly when tox is fixed).

PR #744, opened 2015-11-19
Specify minimum pyOpenSSL version

lorenzph:
urllib3 requires "set_tlsext_host_name" which was only added in
pyOpenSSL 0.13. As some distributions (e.g. Ubuntu 12.04) still ship an
older version enforce the correct minimum version during installation.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,7 +16,7 @@ universal = 1
 [metadata]
 provides-extra = secure
 requires-dist =
-    pyOpenSSL; python_version<="2.7" and extra == 'secure'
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks again for this!

Lukasa:
Ah, I can't merge this from my phone. Someone else should feel free. =)

sigmavirus24:
I would hold off on merging this until we can figure out why Travis is perpetually broken

sigmavirus24:
Restarted the build after merging the tox pin

sigmavirus24:
@shazow :+1: here too

PR #747, opened 2015-11-22
Fix chunked decompression logic

### LINKED ISSUE #743: Could calling  self._decoder.decompress after flush() has been called on the decompressobj instance cause some problems ?
Disclaimer : I don’t know much about python. I am calling, from Java (through jython), a python script that uses a 3rd party python API that makes use of urllib3 to call a REST api. This whole setup obviously makes it challenging to troubleshoot issues.
The 3rd party python API we’re using uses urllib3 version 1.12
While I know it would be very useful, I do not have a small sample to replicate the issue (I’d have to peel a lot of layers that I do not have a good understanding of at this point). I might need to work on it if nobody can help me without it.

The issue I am having is thrown by the zlib java implementation that is called by jython:

```
java.lang.NullPointerException: java.lang.NullPointerException: Inflater has been closed

File "C:\Users\laurentd\.gradle\caches\modules-2\files-2.1\org.python\jython-standalone\2.7.0\cdfb38bc6f8343bcf1d6accc2e1147e8e7b63b75\jython-standalone-2.7.0.jar\Lib\zlib.py", line 202, in _get_inflate_data
	at java.util.zip.Inflater.ensureOpen(Inflater.java:389)
	at java.util.zip.Inflater.inflate(Inflater.java:257)
	at java.util.zip.Inflater.inflate(Inflater.java:280)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
```

The issue seems to be linked to the fact that the flush() method is called on a zlip.decompressobj object but that somehow, somewhere, we’re still trying to call the decompress method on that zlibdecompressobj object.
This is something that the zlib documentation mentions should not be done:
“After calling flush(), the decompress() method cannot be called again; the only realistic action is to delete the object.” (from https://docs.python.org/2/library/zlib.html)

Further debugging took me to the response.py file, in the read_chunked method, especially in the last line of this part:
```
            while True:
                self._update_chunk_length()
                if self.chunk_left == 0:
                    break
                chunk = self._handle_chunk(amt)
                yield self._decode(chunk, decode_content=decode_content,
                                   flush_decoder=True)
```
It looks like when self._decode is called, we always set flush_decoder to True.

Because of this, if I understand correctly, the self._decoder object (which is, in my case, a GzipDecoder, and a decompressobj?) is always flushed at the end of the call to self._decode (line 200).
However, the self._decoder seems to be initialized only once and seems to be reused (at least in my case) throughout the multiple calls to self._decode.

Though I am probably missing a few things, it could explain why decompressobj.decompress is called on a decompressobj object that has been flushed, and this could (?) have unpredictable results, including that NPE in the jython implementation of zlib (and ultimately in java).

Here is a deeper (and sanitized) portion of the call stack when the exception happens.
```
....\scripts\requests\sessions.py", line 608, in send
    r.content
....\scripts\requests\models.py", line 734, in content
    self._content = bytes().join(self.iter_content(CONTENT_CHUNK_SIZE)) or bytes()
....\scripts\requests\models.py", line 734, in content
    self._content = bytes().join(self.iter_content(CONTENT_CHUNK_SIZE)) or bytes()
....\scripts\requests\models.py", line 657, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
....\scripts\requests\models.py", line 657, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
....\scripts\requests\packages\urllib3\response.py", line 322, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
....\scripts\requests\packages\urllib3\response.py", line 471, in read_chunked
    yield self._decode(chunk, decode_content=decode_content,
....\scripts\requests\packages\urllib3\response.py", line 190, in _decode
....\scripts\requests\packages\urllib3\response.py", line 190, in _decode
....\scripts\requests\packages\urllib3\response.py", line 56, in decompress
    return self._obj.decompress(data)
  File "C:\Users\laurentd\.gradle\caches\modules-2\files-2.1\org.python\jython-standalone\2.7.0\cdfb38bc6f8343bcf1d6accc2e1147e8e7b63b75\jython-standalone-2.7.0.jar\Lib\zlib.py", line 155, in decompress
  File "C:\Users\laurentd\.gradle\caches\modules-2\files-2.1\org.python\jython-standalone\2.7.0\cdfb38bc6f8343bcf1d6accc2e1147e8e7b63b75\jython-standalone-2.7.0.jar\Lib\zlib.py", line 202, in _get_inflate_data
  File "C:\Users\laurentd\.gradle\caches\modules-2\files-2.1\org.python\jython-standalone\2.7.0\cdfb38bc6f8343bcf1d6accc2e1147e8e7b63b75\jython-standalone-2.7.0.jar\Lib\zlib.py", line 202, in _get_inflate_data
```
Is it possible that the _decode method is causing the problem by calling decompress on a decompressobj that has been flushed or am I going down the wrong track?

Note that the problem is gone if I reinitialize the decompressobj at the beginning of _decode by mimicking what _init_decoder does, which I can’t believe is a good fix (although self._decoder seems to be really used only in _decode):
```
        content_encoding = self.headers.get('content-encoding', '').lower()
        self._decoder = _get_decoder(content_encoding)
```

Thanks in advance.

Laurent.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28367602-could-calling-self-_decoder-decompress-after-flush-has-been-called-on-the-decompressobj-instance-cause-some-problems?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #743.

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -475,8 +475,14 @@ class HTTPResponse(io.IOBase):
                 if self.chunk_left == 0:
                     break
                 chunk = self._handle_chunk(amt)
-                yield self._decode(chunk, decode_content=decode_content,
-                                   flush_decoder=True)
+                decoded = self._decode(chunk, decode_content=decode_content,
+                                       flush_decoder=False)
+                if decoded:
+                    yield decoded
+
+            decoded = self._decode(b'', decode_content, flush_decoder=True)
+            if decoded:
+                yield decoded
 
             # Chunk content ends with \r\n: discard it.
             while True:
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -195,11 +195,21 @@ class HTTPResponse(io.IOBase):
                 "Received response with content-encoding: %s, but "
                 "failed to decode it." % content_encoding, e)
 
-        if flush_decoder and decode_content and self._decoder:
-            buf = self._decoder.decompress(binary_type())
-            data += buf + self._decoder.flush()
+        if flush_decoder and decode_content:
+            data += self._flush_decoder()
 
         return data
+
+    def _flush_decoder(self):
+        """
+        Flushes the decoder. Should only be called if the decoder is actually
+        being used.
+        """
+        if self._decoder:
+            buf = self._decoder.decompress(b'')
+            return buf + self._decoder.flush()
+
+        return b''
 
     @contextmanager
     def _error_catcher(self):
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -490,9 +490,13 @@ class HTTPResponse(io.IOBase):
                 if decoded:
                     yield decoded
 
-            decoded = self._decode(b'', decode_content, flush_decoder=True)
-            if decoded:
-                yield decoded
+            if decode_content:
+                # On CPython and PyPy, we should never need to flush the
+                # decoder. However, on Jython we *might* need to, so
+                # lets defensively do it anyway.
+                decoded = self._flush_decoder()
+                if decoded:  # Platform-specific: Jython.
+                    yield decoded
 
             # Chunk content ends with \r\n: discard it.
             while True:
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ dev (master)
 
 * Close connections more defensively on exception. (Issue #734)
 
+* Adjusted ``read_chunked`` to handle gzipped, chunk-encoded bodies without
+  repeatedly flushing the decoder, to function better on Jython. (Issue #743)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 


shazow:
Missing coverage. Also do we need that extra _decode? Should we re-order the loop instead?

Lukasa:
Yeah, I meant to write a thing about that but then...I didn't. I don't know why.

What I'm not sure about is whether that last flush is needed *at all*. I've yet to be able to construct a scenario where we need it. You're right though, we might be able to resolve this problem by re-ordering the loop, let me have a think.

Lukasa:
Hmm. The way the loop is written makes it really hard to have a single decode point. In particular, there's a lot of state being hung off the object. This means, for example, we cannot safely call `handle_chunk` on the final zero-length chunk. That's somewhat frustrating.

However, we *can* refactor to pull the flush logic out of `_decode` and then call that directly, which will at least make the whole thing conceptually somewhat cleaner. Separation of concerns and all that.

Lukasa:
So I can do that refactor, but we still don't hit the line in my testing. We're at a stage now where I have to investigate exactly how flush behaves in all the cases we use it.

Lukasa:
AHA! I got it!

We don't need to call `flush` on CPython, at the very least. I give you the weirdly hard to find [CPython bug 23200](https://bugs.python.org/issue23200), which says:

> The decompress() method changed from Z_NO_FLUSH to Z_SYNC_FLUSH in Feb 2001; see revision [01c2470eeb2e](http://hg.python.org/lookup/01c2470eeb2e). I guess previously flush() was necessary to get all your data.

So, on CPython (and probably PyPy, I'll have to check), we definitely don't need it *at all*, in any circumstance. That leaves me with a question: what does Jython need?

Lukasa:
Just did a source code check: PyPy also always calls the decompressor with Z_SYNC_FLUSH from the decompressobj, so we don't need it there either (good work PyPy!). That potentially leaves Jython as the odd-one-out. If we can prove that Jython doesn't automatically use Z_SYNC_FLUSH, then *this* is a bug worth reporting to them @sigmavirus24.

Lukasa:
And I just walked straight into a Jython zlib bug unrelated to this one.

Lukasa:
For those who care, the bug is [here](http://bugs.jython.org/issue2433).

However, on the flush issue, I haven't been able to demonstrate that Jython needs it. However, that doesn't prove much: it *may* need it and I just haven't spotted it yet. I need to dive deeper, sadly.

Lukasa:
Ok, so sadly Jython does *not* pass Z_SYNC_FLUSH, it passes Z_NO_FLUSH. I don't *know* if this is a problem or not, because I don't know what zlib does with this information: I'll see if I can find out. Regardless, I've mentioned this inconsistency as [Jython bug 2434](http://bugs.jython.org/issue2434).

shazow:
:nut:

jimbaker:
So here's the input from the Jython side - and really from the underlying Java support - since Java 7, [Z_SYNC_FLUSH is supported](https://docs.oracle.com/javase/7/docs/api/java/util/zip/Deflater.html#SYNC_FLUSH)

More in this Java dev blog [post](https://blogs.oracle.com/xuemingshen/entry/the_flushable_deflateroutputstream)

So we can readily support in Jython. Thanks for filing that [bug](http://bugs.jython.org/issue2434) with us!

Lukasa:
Right, so in the meantime we need to decide what we're doing while Jython is suspected to require flushing. @shazow, got a preference?

shazow:
What are the options? Always flush vs what? What's the down-side of always-flush in this case?

Lukasa:
The only downside of always flush is testing-based: specifically, I'm currently unable to construct a test-case that hits the flush statement.

shazow:
I'm alright with doing the `# Platform-specific: Jython` coverage omission if it's documented why (link to here I guess).

Lukasa:
Ok, cool, I'll fix that up.

Lukasa:
Hmm, why won't Travis build this?

Lukasa:
Ok, *now* you build. Stupid Travis.

sigmavirus24:
:+1:

urllib3/response.py
```diff
@@ -475,8 +485,18 @@ def read_chunked(self, amt=None, decode_content=None):
...
+                                       flush_decoder=False)
+                if decoded:
+                    yield decoded
+
+            if decode_content:
+                # On CPython and PyPy, we should never need to flush the
+                # decoder. However, on Jython we *might* need to, so
+                # lets defensively do it anyway.
+                decoded = self._flush_decoder()
+                if decoded:  # Platform-specific: Jython.
```
shazow: Hm why is the yield Jython-specific? (Why does Jython yield an extra thing?)

Lukasa: Because the flush works on all platforms, but it doesn't return anything on any other platform (the decoder was flushed at the last decompress call). The `if` block is not platform specific because `decode_content` may be `True` on any platform.

shazow: ¯\\\_(ツ)\_/¯

jimbaker:
Just closed out the Jython related bug: http://bugs.jython.org/issue2434 - we should have compliant support, specifically all flush behaviors, thanks to using the underlying support available as of Java 7 (which Jython 2.7 requires). I don't think you need to do anything specific for Jython going forward, although this will of course require >= Jython 2.7.1

shazow:
Yay this is excellent news, we should add Jython on our list of supported platforms in the README.

shazow:
Wonder how hard it would be to setup a jython 2.7.1 endpoint for travisci...

sigmavirus24:
@jimbaker :cake:

@shazow I think we'd have to use pyenv to get jython 2.7.1

PR #376, opened 2014-04-18
Add source_address support to PoolManager. Issue warnings if source_address is used with Python 2.6. Fix Python 2.6 testing bug.

gruns:

PR #751, opened 2015-11-26
Start keying connection pools off SSL arguments as well.

Lukasa:
This is a proposed patch that begins to implement the requirements for kennethreitz/requests#2863. In particular, it adds hooks to provide additional keyword arguments as part of the key for the connection pool.

There are some outstanding questions. For example:

- should we actually just use all the connection pool keyword arguments as part of the key? Why are `headers` et al exempt?
- Are we happy with this API, where `connection_from_host` picks up the new argument but `connection_from_url` does not?

Code review encouraged, please. :custard:

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,6 +1,6 @@
 import unittest
 
-from urllib3.poolmanager import PoolManager
+from urllib3.poolmanager import PoolManager, SSL_KEYWORDS
 from urllib3 import connection_from_url
 from urllib3.exceptions import (
     ClosedPoolError,
@@ -86,6 +86,64 @@ class TestPoolManager(unittest.TestCase):
         self.assertRaises(ClosedPoolError, conn_pool._get_conn)
 
         self.assertEqual(len(p.pools), 0)
+
+    def test_pools_keyed_on_ssl_arguments_from_host(self):
+        ssl_kw = {
+            'key_file': '/root/totally_legit.key',
+            'cert_file': '/root/totally_legit.crt',
+            'cert_reqs': 'CERT_REQUIRED',
+            'ca_certs': '/root/path_to_pem',
+            'ssl_version': 'SSLv23_METHOD',
+        }
+        p = PoolManager(5)
+        conns = []
+        conns.append(
+            p.connection_from_host(
+                'example.com', 443, scheme='https', ssl_kwargs=ssl_kw
+            )
+        )
+
+        for k in ssl_kw:
+            ssl_kw[k] = 'newval'
+            conns.append(
+                p.connection_from_host(
+                    'example.com', 443, scheme='https', ssl_kwargs=ssl_kw
+                )
+            )
+
+        self.assertTrue(
+            all(
+                x is not y
+                for i, x in enumerate(conns)
+                for j, y in enumerate(conns)
+                if i != j
+            )
+        )
+
+    def test_pools_keyed_on_ssl_arguments_from_url(self):
+        ssl_kw = {
+            'key_file': '/root/totally_legit.key',
+            'cert_file': '/root/totally_legit.crt',
+            'cert_reqs': 'CERT_REQUIRED',
+            'ca_certs': '/root/path_to_pem',
+            'ssl_version': 'SSLv23_METHOD',
+        }
+        p = PoolManager(5, **ssl_kw)
+        conns = []
+        conns.append(p.connection_from_url('https://example.com/'))
+
+        for k in ssl_kw:
+            p.connection_pool_kw[k] = 'newval'
+            conns.append(p.connection_from_url('https://example.com/'))
+
+        self.assertTrue(
+            all(
+                x is not y
+                for i, x in enumerate(conns)
+                for j, y in enumerate(conns)
+                if i != j
+            )
+        )
 
 
 if __name__ == '__main__':
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -95,20 +95,16 @@ class TestPoolManager(unittest.TestCase):
             'ca_certs': '/root/path_to_pem',
             'ssl_version': 'SSLv23_METHOD',
         }
-        p = PoolManager(5)
+        p = PoolManager(5, **ssl_kw)
         conns = []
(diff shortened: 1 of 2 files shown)


alex:
`headers` are handled on a per request, not per connection basis, so there's no reason they need to be included.

Lukasa:
@alex Well, that's true-ish, but we have parameters that are added by default to all requests through a given ConnectionPool.

shazow:
Any feelings about generalizing this beyond SSL?

I'm imagining people might want to key based off sessions (some internal header-related value), resolved ip (as opposed to hostname), source address (a few people brought this up #376), others?

Not sure how many of these usecases we can cover at once, but worth considering while we're touching this part of the plumbing.

shazow:
> Are we happy with this API, where `connection_from_host` picks up the new argument but `connection_from_url` does not?

When I wrote these, they were shortcut helpers for fairly specific flows. I'm open to expanding them, or adding more, or getting rid of them.

Lukasa:
> Any feelings about generalizing this beyond SSL?

It should be trivial to generalise to anything that is in `connection_pool_kw`, and probably to `headers` as well. My bigger concern is: will that be too many things?

shazow:
I'm imagining something like:

```python
def connection_from_key(self, *args):
    ...

def connection_from_request(self, <a bunch of request context>):
    ... logic you can override to convert to a connection_from_key call of your desire
```

Possibly as a callable you pass instead, whatever makes more sense.

Lukasa:
Yeah, that can totally work. Would those be in addition to `connction_from_url`/`connection_from_host`?

shazow:
Probably? ¯\\\_(ツ)\_/¯

Lukasa:
So the problem with `connection_from_key` is that, if it needs to create a new ConnectionPool, it needs to know the scheme, host, and port. Given that we already have a method that basically has that signature (`connection_from_host`), I think we should either consider changing the name of that method or just accept that that is the base method for keying.

So, in the short term I'm just going to do a small refactor that means we pass all of the relevant `connection_pool_kw`. I'll also not going to key off `headers` because they don't actually get set on the relevant `ConnectionPool`, so it doesn't really matter. This means we'll be keyed off:

- scheme
- host
- port
- strict
- default timeout
- default retries
- maxsize
- block
- arguments that are applied directly to connections (including the SSL ones)

I think that's probably sufficient for now. =)

Lukasa:
Ok, I've uploaded a change that does this. It might be a bit heavyweight, so I haven't focused on exhaustively testing it yet, but I'd love feedback. =)

shazow:
Hm I think this might be going in the wrong direction, we need to go more towards treating a pool as a cache. This means we'll probably need to provide a key (which should strictly be a flat tuple IMO) and a callable that returns a new value (a fn() -> ConnectionPool).

I'm not a fan of the automagically-keying-off-a-bunch-of-things-based-on-the-scheme approach right now. It's starting to look like a half-way adapter registry (not that it's bad, but we need to commit to some direction I think).

For now, I'd be happier with PoolManager having a scheme-based if-statement which replaces the default HTTP tuple key with one that includes more HTTPS-specific stuff.

Does that make sense? Sorry for the fairly vague feedback. :P

Lukasa:
@shazow So, can you look back at the commit prior to the HEAD of this branch and confirm whether that looks closer to what you wanted?

shazow:
Bits that I'd like to avoid:

```python
ssl_kwargs = frozenset(ssl_kwargs.items())
```

I'd like to avoid the key being a recursive structure. I think it should be a flat well-defined set of values that we're keying on in specific scenarios. Maybe we should even have a set of helpers for each scenario, where given request context it will generate the appropriate key.

One big problem here is that urllib3 still doesn't have a notion of request contexts (or pre-built requests) so serializing that context is tricky. It's also something we need for retry history.

```python
        ssl_kwargs = {
            k: v for k, v in self.connection_pool_kw.items()
            if k in SSL_KEYWORDS
        }
```

Maybe, at the least, we need some idea of a hashable SSL configuration context? Wiring all of these random SSL-related keywords to be further propagated is likely to yield more pain in the future, especially if some of these values become transient or don't really affect the request itself (but rather just the validation configuration).

sigmavirus24:
So I agree about flat being better than nested. Absolutely. The problem is creating a key that's immutable that the dictionary will know is immutable and it won't barf on. Using a list or dictionary inside a tuple does not an immutable key make.

Also, perhaps since the key object is growing so potentially long, we should look at using a namedtuple simply to make things easier to reason about when handling the keys.

Lukasa:
Yeah, if we're moving to a flat structure we're definitely moving to namedtuple. This is frankly more for debugging reasons than anything else: it'll be a bloody nightmare to work out what's going on in a large tuple of random-ish values and easy to subtly get wrong.

shazow:
Yea, urllib3's internal keying should be well structured and defined, but I'm expecting we'll want to allow third parties to change the keying to their own needs. Or maybe this is over-generalizing.

sigmavirus24:
> we'll want to allow third parties to change the keying to their own needs

Does anyone do this now? I'm not sure I see a need for it.

shazow:
See https://github.com/shazow/urllib3/pull/751#issuecomment-160181929

jeremycline:
Would it be possible to have a ruling on a direction to take? I've been hitting this issue while using requests and I'd really like to fix it, but until I know what's going to happen with this patch, there's not much I can do :smile:

shazow:
@jeremycline Would you be interested in putting together a draft PR for review with the feedback we talked about? Unfortunately we're all pretty busy right now so I'm not sure this issue will move forward on its own any time soon.

Alternatively, could always add a bounty and see what happens. :) If you want to do that, I'd recommend opening a separate umbrella issue for the bounty specifically.

jeremycline:
No worries, I completely understand being too busy. I'm happy to give a PR a shot. I should have something that's worth an initial review in the next few days.

shazow:
Hey everyone in this thread, please take a look at #830 which we're quite happy with as a solution but if you have feedback then now is the time. :)

Lukasa:
We can close this in favour of #830 at this point.

Issue #748, opened 2015-11-23
Experiment with flaky project to re-run designated flakey tests


sigmavirus24:
https://github.com/box/flaky looks like it might help our CI problem over come it's flaky-ness (for a temporary solution until we can make the test suite itself less flaky by default).

shazow:
Ideally we'd want to re-run tests that fail with a specific error. All of the flaky travis failures look like this:

```
  ...
  File "/opt/python/3.2.5/lib/python3.2/socket.py", line 287, in readinto
    return self._sock.recv_into(b)
nose.proxy.error: [Errno 104] Connection reset by peer
```

As far as I can tell, the machine is overloaded and is randomly killing sockets.

Since like 80% of our tests depend on sockets, it won't be super productive to annotate them. I (mostly) deflaked all the failures that were related to using time.sleep. I think there might be one or two left, but they're almost never the cause of failures anymore.

Lukasa:
I'd like to fix this properly, if we can. The problem is something in the Travis environment, and it should be possible for us to track it down and fix it.

I seem to recall seeing two failures. The first, as pointed out by @shazow, is:

```
======================================================================
ERROR: test_connect_timeout (test.with_dummyserver.test_connectionpool.TestConnectionPoolTimeouts)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/shazow/urllib3/test/with_dummyserver/test_connectionpool.py", line 158, in test_connect_timeout
    self.assertRaises(ConnectTimeoutError, pool._make_request, conn, 'GET', url)
  File "/opt/python/pypy-2.5.0/lib-python/2.7/unittest/case.py", line 473, in assertRaises
    callableObj(*args, **kwargs)
  File "/home/travis/build/shazow/urllib3/urllib3/connectionpool.py", line 376, in _make_request
    httplib_response = conn.getresponse(buffering=True)
  File "/opt/python/pypy-2.5.0/lib-python/2.7/httplib.py", line 1068, in getresponse
    response.begin()
  File "/opt/python/pypy-2.5.0/lib-python/2.7/httplib.py", line 409, in begin
    version, status, reason = self._read_status()
  File "/opt/python/pypy-2.5.0/lib-python/2.7/httplib.py", line 365, in _read_status
    line = self.fp.readline(_MAXLINE + 1)
  File "/opt/python/pypy-2.5.0/lib-python/2.7/socket.py", line 534, in readline
[...]

PR #752, opened 2015-11-26
Repair flaky connection timeout tests.

Lukasa:
This PR changes the tests that previously used the `noop_handler` to instead use the `TARPIT_HOST`, which should fix their flakiness. See also #748.


Lukasa:
So, no more "connection reset by peer errors", but we did hit another error that I'm adding to the flaky test list in #748.

Lukasa:
Ok, this has expanded a bit to start including a few fixes from #748.

Lukasa:
LOOK, A GREEN TEST RUN FIRST TIME! \o/ :v:

shazow:
:four_leaf_clover:

PR #754, opened 2015-12-02
Define an SNIMissingWarning.

### LINKED ISSUE #753: Warn if we do not have access to SNI
Suggested in kennethreitz/requests#2910. This would make it easier for people to correlate unexpected cert verification failures with the specific action they need to take. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28763221-warn-if-we-do-not-have-access-to-sni?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
This fires whenever we make a HTTPS request and don't have SNI present. Should only fire once per run.

Resolves #753.

While we're here, @shazow @sigmavirus24: I'm thinking about having the docs URL be a property on the warning itself, so that requests can override them to point at our own docs when requests is being used. Thoughts?

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -305,7 +305,7 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         return context.wrap_socket(sock, server_hostname=server_hostname)
 
     warnings.warn(
-        'A HTTPS request has been made, but the SNI (Subject Name '
+        'An HTTPS request has been made, but the SNI (Subject Name '
         'Indication) extension to TLS is not available on this platform. '
         'This may cause the server to present an incorrect TLS '
         'certificate, which can cause validation failures. For more '
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -6,7 +6,7 @@ import hmac
 from binascii import hexlify, unhexlify
 from hashlib import md5, sha1, sha256
 
-from ..exceptions import SSLError, InsecurePlatformWarning
+from ..exceptions import SSLError, InsecurePlatformWarning, SNIMissingWarning
 
 
 SSLContext = None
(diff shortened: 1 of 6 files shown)


sigmavirus24:
> I'm thinking about having the docs URL be a property on the warning itself, so that requests can override them to point at our own docs when requests is being used.

I'm curious how you would propose doing that. Otherwise I'm +1

Lukasa:
Like this:

```python
class SNIMissingWarning(HTTPWarning):
    DocsURL = 'http://urllib3.readthedocs.org/security.html#snimissingwarning'
```

Then, change the `warn` calls to use the URLs from the specific warning they raise.

At which point, on import time, requests can write to those properties and set them to what we want them to be.

sigmavirus24:
Ah so then it works (in requests) like:

```py
from .packages.urllib3 import exceptions

exceptions.SNIMissingWarning.DocsURL = '...'
```

This will cause problems for distros that unvendor urllib3 and then we're overwriting what urllib3 would be using (e.g., in a case where the same process uses a library that imports urllib3 independently and another thing that uses requests such that urllib3's warnings would point to requests when not using requests).

Are we sure we want to do that?

Lukasa:
@sigmavirus24 Yeah, that problem would exist. Frankly, I'm pretty unsure of how to fix it otherwise, but the UX for the warnings from requests is not great at the moment.

sigmavirus24:
I agree with that @Lukasa. I just wanted to point that out as something that people will almost certainly run into.

urllib3/util/ssl_.py
```diff
@@ -303,4 +303,15 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         context.load_cert_chain(certfile, keyfile)
     if HAS_SNI:  # Platform-specific: OpenSSL with enabled SNI
         return context.wrap_socket(sock, server_hostname=server_hostname)
+
+    warnings.warn(
+        'A HTTPS request has been made, but the SNI (Subject Name '
```
shazow: s/A HTTPS/An HTTPS/?

shazow:
Regarding the configurable message, would it be easier to have requests override which warning class is used and provide its own? (Or monkeypatch.)

Lukasa:
@shazow It's certainly possible, but the message is hardcoded at the warn site at the moment.

sigmavirus24:
Monkey-patching will have the same (confusing) effect though. :/

Lukasa:
Ok, I think we should be merging this regardless of how we decide to deal with warning messages. I'll open a separate issue to track that.

sigmavirus24:
:+1: from me.

shazow:
Looks good for now, though I think we should clarify the message to emphasize that the scenario yields false negatives (rather than false positives).

PR #757, opened 2015-12-03
Catch socket errors in error_catcher

### LINKED ISSUE #756: Catch socket.error in error_catcher
Originally spotted in kennethreitz/requests#1236. It turns out that during `stream` a `socket.error` can get raised that isn't wrapped by httplib. We should catch and wrap those ourselves.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28810607-catch-socket-error-in-error_catcher?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #756.

This is fundamentally a defensive change because it's pretty hard to hit this, but it clearly does happen (usually involving weird SSL issues and long-running connections).

### DIFF
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -3,6 +3,7 @@ from contextlib import contextmanager
 import zlib
 import io
 from socket import timeout as SocketTimeout
+from socket import error as SocketError
 
 from ._collections import HTTPHeaderDict
 from .exceptions import (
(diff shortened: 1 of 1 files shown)


sigmavirus24:
:+1: from me

PR #761, opened 2015-12-15
Strip brackets from hostnames before validating.

### LINKED ISSUE #760: Must remove square brackets from IPv6 addresses when using match_hostname
Spotted in [OpenStack](https://bugs.launchpad.net/nova/+bug/1526413). Introduced by my fix #708.

Apparently, while `httplib` itself doesn't care about the square brackets, the `match_hostname` function needs them gone. I suggest we move the stripping down into our `ssl.py` module.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29146492-must-remove-square-brackets-from-ipv6-addresses-when-using-match_hostname?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
This resolves #760.


urllib3/connection.py
```diff
@@ -265,7 +265,9 @@ def connect(self):
                     'for details.)'.format(hostname)),
                     SubjectAltNameWarning
                 )
-            match_hostname(cert, self.assert_hostname or hostname)
+            asserted_hostname = self.assert_hostname or hostname
+            asserted_hostname = asserted_hostname.strip('[]')
```
shazow: Could you add a comment explanation here?

Lukasa: Ah yeah, sorry.

Lukasa: Done. =)

urllib3/connection.py
```diff
@@ -265,7 +265,14 @@ def connect(self):
                     'for details.)'.format(hostname)),
                     SubjectAltNameWarning
                 )
-            match_hostname(cert, self.assert_hostname or hostname)
+
+            # In case the hostname is an IPv6 address, strip the square
+            # brackets from it before using it to validate. Non IPv6 addresses
+            # should not start/end with square brackets, so this should be
+            # safe.
```
shazow: I don't mean explaining what it's doing but moreso why. Worth saying that this is something that match_hostname specifically demands?

Lukasa: Done again. =)

sigmavirus24: It's not that it demands that, it's that we strip `[]` from IPv6 hostnames (because of httplib on some Python 3 version (I think 3.2)) and when we then pass them to `ssl_match_hostname` we have a mismatch between `::1` and `[::1]`.

sigmavirus24:
This looks good to me. Thanks for tackling this so quickly @Lukasa

PR #692, opened 2015-08-09
[wip] SOCKS support, take 3

Anorov:
**[WIP]**

Here is an attempt at SOCKS support with the latest version of PySocks. At the moment there are no tests, though I plan to add them, and definitely encourage others to add them and help with them. I've been testing it with a local script so far. The functionality should all work at the moment, but I can't say that with serious confidence until full tests have been added.

I had to choose between peppering ProxyManager with a few if-branches, or subclassing it and making a SOCKSProxyManager. I decided to go with the former, to preserve full backwards compatibility and allow end users to use the familiar `ProxyManager("socks5://localhost:9050")` syntax. A way to make it more OO while preserving compatibility would be to add override __new__ for ProxyManager and defer to the SOCKSProxyManager class when a SOCKS URL scheme is seen, but it's debatable whether that is more or less Pythonic.

Unfortunately, a pure-Python SOCKS4 and SOCKS5 server may not be feasible for testing, unless someone has one already written. The ones I found on the Internet have various issues. This is why I've held off on comitting anything since then, and frankly didn't want to spend the time hunting around to get a full test suite working. So, test assistance is very much needed.

I am not interested in the bounty, and would like it split between the other contributors.

PR #762, opened 2015-12-15
Add SOCKS support via contrib module (AKA SOCKS support, take 4)

Lukasa:
@Anorov did fantastic work in #692 but did not have time to go back through and respond to code review, so I've adopted the work done there and refactored it. Credit should be joint between me and @Anorov should this satisfy the bounty: however, @Anorov has previously claimed they don't want the bounty. For that reason, if @Anorov doesn't change their mind and this submission does qualify for it, I'll donate the bounty, in full, to PyLadies. (I don't need it because I'm paid full-time to work on libraries like this.)

This change essentially pulls most of the work done in #692 out into a contrib module. This allows us to avoid having to change too many things in the mainline of urllib3, and also ensures that it is not affected by the requirements to have tests and coverage. This helps us get the work initially off the ground and out into the world.

If we decide to take this change as-is, I'll also volunteer to be "on the hook" for supporting it going forward until such time as I can add sufficient test coverage for it.

Feedback is encouraged!

### DIFF
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -55,6 +55,8 @@ There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/
 SOCKS Proxies
 -------------
 
+.. versionadded:: 1.14.0
+
 The :mod:`urllib3.contrib.socks` module enables urllib3 to work with proxies
 that use either the SOCKS4 or SOCKS5 protocols. These proxies are common in
 environments that want to allow generic TCP/UDP traffic through their borders,
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -69,6 +69,12 @@ extra, like so:
 
     $ pip install urllib3[socks]
 
+If you have already got urllib3 1.14.0 or later installed, run:
+
+.. code-block:: bash
+
+    $ pip install -U urllib3[socks]
+
 The SOCKS module provides a
 :class:`SOCKSProxyManager <urllib3.contrib.socks.SOCKSProxyManager>` that can
 be used when SOCKS support is required. This class behaves very much like a
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -80,8 +80,8 @@ machine, listening on port 8889:
 
     from urllib3.contrib.socks import SOCKSProxyManager
 
-    pm = SOCKSProxyManager('socks5://localhost:8889/')
-    r = pm.request('GET', 'https://www.google.com/')
+    http = SOCKSProxyManager('socks5://localhost:8889/')
+    r = http.request('GET', 'https://www.google.com/')
 
 The SOCKS implementation supports the full range of urllib3 features. It also
 supports the following SOCKS features:
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -4,7 +4,7 @@ Contrib Modules
 ===============
 
 These modules implement various extra features, that may not be ready for
-prime time.
+prime time or that require optional third-party dependencies.
 
 .. _contrib-pyopenssl:
 
--- a/docs/contrib.rst
+++ b/docs/contrib.rst
@@ -16,7 +16,7 @@ SNI-support for Python 2
 
 .. _gae:
 
-Google App Engine 
+Google App Engine
 -----------------
 
 The :mod:`urllib3.contrib.appengine` module provides a pool manager that
@@ -45,8 +45,54 @@ There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/
 
 1. You can use :class:`AppEngineManager` with URLFetch. URLFetch is cost-effective in many circumstances as long as your usage is within the limitations.
 2. You can use a normal :class:`PoolManager` by enabling sockets. Sockets also have `limitations and restrictions <https://cloud.google.com/appengine/docs/python/sockets/#limitations-and-restrictions>`_ and have a lower free quota than URLFetch. To use sockets, be sure to specify the following in your ``app.yaml``::
-    
+
     env_variables:
         GAE_USE_SOCKETS_HTTPLIB : 'true'
 
 3. If you are using `Managed VMs <https://cloud.google.com/appengine/docs/managed-vms/>`_, you can use the standard :class:`PoolManager` without any configuration or special environment variables.
+
+
+SOCKS Proxies
+-------------
+
+The :mod:`urllib3.contrib.socks` module enables urllib3 to work with proxies
+that use either the SOCKS4 or SOCKS5 protocols. These proxies are common in
+environments that want to allow generic TCP/UDP traffic through their borders,
+but don't want unrestricted traffic flows.
+
+To use it, either install ``PySocks`` or install urllib3 with the ``socks``
+extra, like so:
+
+.. code-block:: bash
+
+    $ pip install urllib3[socks]
+
+The SOCKS module provides a
+:class:`SOCKSProxyManager <urllib3.contrib.socks.SOCKSProxyManager>` that can
+be used when SOCKS support is required. This class behaves very much like a
+standard :class:`ProxyManager <urllib3.poolmanager.ProxyManager>`, but allows
+the use of a SOCKS proxy instead.
+
+Using it is simple. For example, with a SOCKS5 proxy running on the local
+machine, listening on port 8889:
+
+.. code-block:: python
+
+    from urllib3.contrib.socks import SOCKSProxyManager
+
+    pm = SOCKSProxyManager('socks5://localhost:8889/')
+    r = pm.request('GET', 'https://www.google.com/')
+
+The SOCKS implementation supports the full range of urllib3 features. It also
+supports the following SOCKS features:
+
+- SOCKS4
+- SOCKS4a
+- SOCKS5
+- Usernames and passwords for the SOCKS proxy
+
+The SOCKS module does have the following limitations:
+
+- No support for contacting a SOCKS proxy via IPv6.
+- No support for reaching websites via a literal IPv6 address: domain names
+  must be used.
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -14,6 +14,18 @@ SOCKS_VERSION_SOCKS4 = b'\x04'
 SOCKS_VERSION_SOCKS5 = b'\x05'
 
 
+def _get_free_port(host):
+    """
+    Gets a free port by opening a socket, binding it, checking the assigned
+    port, and then closing it.
+    """
+    s = socket.socket()
+    s.bind((host, 0))
+    port = s.getsockname()[1]
+    s.close()
+    return port
+
+
 def _read_exactly(sock, amt):
     """
     Read *exactly* ``amt`` bytes from the socket ``sock``.
@@ -347,6 +359,40 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
             self.assertTrue("SOCKS5 authentication failed" in str(e))
         else:
             self.fail("Did not raise")
+
+    def test_source_address_works(self):
+        expected_port = _get_free_port(self.host)
+
+        def request_handler(listener):
+            sock = listener.accept()[0]
+            self.assertEqual(sock.getpeername()[0], '127.0.0.1')
+            self.assertEqual(sock.getpeername()[1], expected_port)
+
+            handler = handle_socks5_negotiation(sock, negotiate=False)
+            addr, port = next(handler)
+
+            self.assertEqual(addr, '16.17.18.19')
+            self.assertTrue(port, 80)
+            handler.send(True)
+
+            while True:
+                buf = sock.recv(65535)
+                if buf.endswith(b'\r\n\r\n'):
+                    break
+
+            sock.sendall(b'HTTP/1.1 200 OK\r\n'
+                         b'Server: SocksTestServer\r\n'
+                         b'Content-Length: 0\r\n'
+                         b'\r\n')
+            sock.close()
+
+        self._start_server(request_handler)
+        proxy_url = "socks5://%s:%s" % (self.host, self.port)
+        pm = socks_proxy.SOCKSProxyManager(
+            proxy_url, source_address=('127.0.0.1', expected_port)
+        )
+        response = pm.request('GET', 'http://16.17.18.19')
+        self.assertEqual(response.status, 200)
 
 
 class TestSOCKS4Proxy(IPV4SocketDummyServerTestCase):
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -1,7 +1,7 @@
 import threading
 import socket
 
-from urllib3.contrib import socks_proxy
+from urllib3.contrib import socks
 from urllib3.exceptions import ConnectTimeoutError, NewConnectionError
 
 from dummyserver.server import DEFAULT_CERTS
@@ -222,7 +222,7 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
 
         self._start_server(request_handler)
         proxy_url = "socks5://%s:%s" % (self.host, self.port)
-        pm = socks_proxy.SOCKSProxyManager(proxy_url)
+        pm = socks.SOCKSProxyManager(proxy_url)
         response = pm.request('GET', 'http://16.17.18.19')
 
         self.assertEqual(response.status, 200)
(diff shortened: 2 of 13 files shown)


Lukasa:
Ok, while writing tests for this I've encountered the first problem: PySocks' `create_connection` method can only create IPv4 connections. I'm going to try to work around that in the contrib module itself, but we may also want to propose this fix upstream.

Lukasa:
And the second problem: PySocks doesn't support using IPv6 on the other side of the connection, even though you can coerce it to do IPv6 locally. I think that at this point it no longer makes any sense to retrofit IPv6 support onto PySocks from the outside, and instead we can look into doing upstream enhancements to PySocks to enable support for it. For now, as a contrib module, I think it's an acceptable limitation to have IPv4-only support for SOCKS.

shazow:
Sounds good, let's document the limitations in the module's header docstring.

Lukasa:
Ok @shazow, I've now added some tests. How does this feel to you? In particular, please check what I had to do to tox to make the socks tests run by default, and tell me if you'd rather we did it a different way.

test/contrib/test_socks.py
```diff
@@ -0,0 +1,492 @@
...
+            sock.sendall(b'HTTP/1.1 200 OK\r\n'
+                         b'Server: SocksTestServer\r\n'
+                         b'Content-Length: 0\r\n'
+                         b'\r\n')
+            sock.close()
+
+        self._start_server(request_handler)
+        proxy_url = "socks5://%s:%s" % (self.host, self.port)
+        pm = socks_proxy.SOCKSProxyManager(proxy_url)
+        response = pm.request('GET', 'http://http2bin.org')
```
shazow: Can we avoid external requests? (Could setup a second socket listener/handler)

tox.ini
```diff
@@ -4,6 +4,7 @@ envlist = flake8-py3, py26, py27, py32, py33, py34, pypy
 [testenv]
 deps= -r{toxinidir}/dev-requirements.txt
 commands=
+    pip install .[socks]
```
shazow: wonder if we need an [all] tag?

urllib3/contrib/socks_proxy.py
```diff
@@ -0,0 +1,156 @@
+# -*- coding: utf-8 -*-
+"""
+SOCKS support for urllib3
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This contrib module contains provisional support for SOCKS proxies from within
```
shazow: Supports both SOCKS4 and SOCKS5, correct? (worth noting I think)

shazow:
Looks fantastic, excellent work @Lukasa! Couldn't have done it better myself. :)

Let's get one more pair of eyes on this, then I'm happy to merge.

shazow:
Btw, what's the coverage like right now? I'd be willing to skip contrib if it's within reach of 100%. Very happy with the design here.

test/contrib/test_socks.py
Lukasa: This doesn't make an external request: it goes via the SOCKS proxy.

tox.ini
Lukasa: Do we want it? It'll cause PyOpenSSL et al to get installed here, which may not be all that sensible a thing to do.

urllib3/contrib/socks_proxy.py
Lukasa: Done.

test/contrib/test_socks.py
shazow: Ah fair enough, in that case let's make it a non-URL (example.com or something).

Lukasa:
Coverage-wise we're three lines short: 52, 96, 119. These *should* be hittable, I think, with a bit more work.

If you wanted this out of contrib (which I'd be fine with), how would you want to handle the optional dependency on PySocks?

test/contrib/test_socks.py
Lukasa: Changed to example.com.

shazow:
On Dec 16, 2015 10:38 AM, "Cory Benfield" <notifications@github.com> wrote:
>
> Coverage-wise we're three lines short: 52, 96, 119. These should be
hittable, I think, with a bit more work.
>
> If you wanted this out of contrib (which I'd be fine with), how would you
want to handle the optional dependency on PySocks?

Good point, we've been keeping things with external deps in contrib.

What would you prefer to do? Could omit it from the coverage exclusion and
keep it in contrib.

Lukasa:
Ok @shazow, this now has `socks_proxy.py` included in the coverage and full test coverage achieved.

One thing this doesn't have, and I'd like to have, is a test that does HTTPS over the proxy. This is a bit tricky to do, but I think it's achievable, so let's not merge until I can get that done. Does that sound good?

sigmavirus24:
I'm happy to review when this is ready for review. Just let me know @Lukasa :dancer:

Lukasa:
I believe this is ready for preliminary review: I don't think I'll get to the HTTPS test until tomorrow.

urllib3/contrib/socks_proxy.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
```
shazow: IMO `urllib3/contrib/socks.py` is good enough.

shazow:
@Lukasa Sounds good!

urllib3/contrib/socks_proxy.py
sigmavirus24: Why not `urllib3/contrib/tubesocks.py`? :-P

shazow: festivesocks.py

Lukasa: Well I might have to use one of those, because `socks.py` busts the import logic (I import `socks` in this module, so if I have a `socks.py` in this directory everything gets screwy.)

Lukasa:
Ok, I've just pushed up a TLS test that verifies that everything works correctly. To get this to work on all platforms I did need to extend our fallback `SSLContext.wrap_socket` function to take the `server_side` argument, but I don't think that's too severe a burden frankly. Let me know if you disagree.

urllib3/contrib/socks_proxy.py
shazow: Hm normally I'd suggest `socks_.py` for internal stuff but that would be annoying for people importing.

How about some `absolute_import` action?

urllib3/util/ssl_.py
```diff
@@ -110,7 +110,7 @@ def set_ciphers(self, cipher_suite):
                 )
             self.ciphers = cipher_suite
 
-        def wrap_socket(self, socket, server_hostname=None):
+        def wrap_socket(self, socket, server_hostname=None, server_side=False):
```
shazow: Casual note: I'd have called it `is_server_side=False`, but this is fine too.

urllib3/contrib/socks_proxy.py
Lukasa: Done. =)

urllib3/util/ssl_.py
Lukasa: Yeah, I stole this name from the actual `wrap_socket` function in the `ssl` module. So you can take it up with python-dev. ;)

Lukasa:
Ok, so this is now up for review again. =)

urllib3/util/ssl_.py
shazow: Ah, precedent always wins, gj. :)

shazow:
Looks great to me! :custard:!

Only thing left is docs, an [example with details to live here](https://urllib3.readthedocs.org/en/latest/contrib.html) with a link from the landing page. Could do in this PR or a separate one, up to you. :)

Lukasa:
Let's do it all here.

Lukasa:
Initial SOCKS documentation pushed. =)

urllib3/contrib/socks.py
```diff
@@ -0,0 +1,157 @@
...
+Known Limitations:
+
+- Currently PySocks does not support contacting remote websites via literal
+  IPv6 addresses. Any such connection attempt will fail.
+- Currently PySocks does not support IPv6 connections to the SOCKS proxy. Any
+  such connection attempt will fail.
+"""
+from __future__ import absolute_import
+
+import socks
```
shazow: Should we add an extra helpful message on import failure? (That we require pysocks, and maybe link to the docs on the module).

shazow:
Looks good.

Eventually we should add a listing to the available contrib modules on the docs landing page. Also need to update the contrib docs intro that it's more about things that require third-party dependencies rather than "not ready for prime time".

urllib3/contrib/socks.py
Lukasa: Sure. Printed, logged, or on the exception itself?

shazow: Hm not sure, maybe a DependencyWarning or something?

Lukasa: Done.

Lukasa:
Ok, I've added a warning, changed the variable names in the docs, and added "or require third-party dependencies" to the contrib blurb.

shazow:
LGTM. :custard:-quality stuff. Thank you @Lukasa!

@sigmavirus24 Got a chance to give it a final quick look?

docs/contrib.rst
```diff
@@ -45,8 +45,54 @@ There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/
...
+
     env_variables:
         GAE_USE_SOCKETS_HTTPLIB : 'true'
 
 3. If you are using `Managed VMs <https://cloud.google.com/appengine/docs/managed-vms/>`_, you can use the standard :class:`PoolManager` without any configuration or special environment variables.
+
+
+SOCKS Proxies
+-------------
+
```
sigmavirus24: It would be good to add

```rst
.. versionadded:: 1.14.0
```

So people know when this was added

docs/contrib.rst
```diff
@@ -45,8 +45,54 @@ There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/
...
+that use either the SOCKS4 or SOCKS5 protocols. These proxies are common in
+environments that want to allow generic TCP/UDP traffic through their borders,
+but don't want unrestricted traffic flows.
+
+To use it, either install ``PySocks`` or install urllib3 with the ``socks``
+extra, like so:
+
+.. code-block:: bash
+
+    $ pip install urllib3[socks]
```
sigmavirus24: We might want to suggest that if someone already has `urllib3` installed at 1.14.0 then they use `pip install -U urllib3[socks]` to get the extra dependencies.

docs/contrib.rst
```diff
@@ -45,8 +45,54 @@ There are `limitations <https://cloud.google.com/appengine/docs/python/urlfetch/
...
+machine, listening on port 8889:
+
+.. code-block:: python
+
+    from urllib3.contrib.socks import SOCKSProxyManager
+
+    http = SOCKSProxyManager('socks5://localhost:8889/')
+    r = http.request('GET', 'https://www.google.com/')
+
+The SOCKS implementation supports the full range of urllib3 features. It also
```
sigmavirus24: Are the features listed somewhere that we could provide something like

```rst
:ref:`features <link-attribute>`
```

tox.ini
sigmavirus24: @shazow worry not, someone can do 

```
pip install .[socks,secure,...]
```

urllib3/contrib/socks.py
```diff
@@ -0,0 +1,170 @@
...
+except ImportError:
+    ssl = None
+
+
+class SOCKSConnection(HTTPConnection):
+    """
+    A plain-text HTTP connection that connects via a SOCKS proxy.
+    """
+    def __init__(self, *args, **kwargs):
+        self._socks_options = kwargs.pop('_socks_options')
```
sigmavirus24: Potentially silly question, why is the kwarg named `_socks_options` (why the preceding `_`)?

urllib3/exceptions.py
```diff
@@ -180,6 +180,14 @@ class SNIMissingWarning(HTTPWarning):
     pass
 
 
+class DependencyWarning(HTTPWarning):
```
sigmavirus24: I know `HTTPWarning` is our base warning, but it is irksome here. Not sure we can fix this pre-2.0 though.

sigmavirus24: Unless we do:

```py
class Urllib3Warning(Warning):
    ...

HTTPWarning = Urllib3Warning
```

It would be backwards compatible but I don't like that very much anyway because HTTPWarnings still catch dependency warnings. 

Bleh

urllib3/poolmanager.py
```diff
@@ -18,11 +18,6 @@
 __all__ = ['PoolManager', 'ProxyManager', 'proxy_from_url']
 
 
-pool_classes_by_scheme = {
```
sigmavirus24: This is going to be subtly backwards incompatible.

I know there are projects that rely on being able to mutate this to change urllib3's (and as a consequence, requests') behaviour.

You're also removing something that was a module level attribute which is even more backwards incompatible. I have a suggestion:

Let's leave this, and in the `PoolManager` class, let's do something like

```py
if not hasattr(self, 'pool_classes_by_scheme'):
    self.pool_classes_by_scheme = pool_classes_by_scheme
```

This gets you the behaviour you want without breaking other people.

We should also add a note to break this in 2.0

sigmavirus24:
With the exception of my `pool_classes_by_scheme` comment, everything else here is a "Nice to Have" comment. The exception, however, is how backwards incompatible that one change is. If y'all are okay with it, I can try to find the project I remember mucking with that stuff and try and help them fix that, but I doubt it'll happen.

docs/contrib.rst
Lukasa: I don't think so. =(

urllib3/contrib/socks.py
Lukasa: The main reason is that I don't want this to be viewed as user-editable. There's also consistency with the other proxy options, which are passed as `_proxy_options`.

urllib3/exceptions.py
Lukasa: I'm inclined to leave this as-is.

urllib3/contrib/socks.py
sigmavirus24: I didn't realize the other proxy options were passed that way. Ignore me :)

urllib3/exceptions.py
sigmavirus24: Yeah, my "Bleh" was a remark of discontent because I don't see a good way of "fixing" this. It's not broken, just semantically irksome to me.

urllib3/poolmanager.py
Lukasa: Yup, let's fix this up.

Lukasa:
Ok, I believe I've done that. Let's see how the tests go.

sigmavirus24:
:+1: from me.

urllib3/exceptions.py
shazow: Hm doesn't really bother me,I feel better about something like BaseWarning (or HTTPWarning) than urllib3.Urllib3Warning.

I'm +0 to making it its own base warning though.

Lukasa: I'm inclined to leave it: if you want to filter out all urllib3 warnings, I feel like that should be do-able just as before.

urllib3/poolmanager.py
```diff
@@ -65,6 +65,10 @@ def __init__(self, num_pools=10, headers=None, **connection_pool_kw):
         self.pools = RecentlyUsedContainer(num_pools,
                                            dispose_func=lambda p: p.close())
 
+        # Locally set the pool classes so other PoolManagers can override them.
+        if not hasattr(self, 'pool_classes_by_scheme'):
```
shazow: Hm is this check necessary?

Lukasa: *Probably* not? @sigmavirus24?

Lukasa: Oh, actually, maybe. It depends how careful you want to be about the order in which people call `super()` and then do their own initialization.

shazow: Isn't it just a matter of whether they override before they call super().__init__()?

If it is necessary, I can think of a dozen other places we might need to add it. :P

Lukasa: Yeah, it is. You're right, we can leave it out: just saying why it was added in the first place. ;)

Lukasa: Gone.

Anorov:
Apologies for the lack of activity in the past few months; I've been very busy with other things.

This looks fantastic, @Lukasa. I actually already factored my original changes out into a contrib module some time ago, but didn't get a chance to push them. However, your version looks a bit cleaner than what I have, so I'm glad this was used instead. And yes, I'm not interested in any part of the bounty. Please feel free to donate it.

I noticed you forked PySocks and made a few IPv6-related changes. I'd be glad to merge those in once they're complete.

Lukasa:
@Anorov Don't worry about the absence: when you're busy, you're busy, it's not your fault!

Regarding the IPv6 stuff, I have two pull requests open. Please feel free to give code review feedback and I'll do my best to apply that in a timely manner!

Anorov:
@Lukasa Sorry about that, somehow missed that they were opened already. Merged.

PR #763, opened 2015-12-17
Fixes the AppEngineManager to work within the requests framework

mikelambert:
- Properly matches HTTPConnectionPool spec for __init__, clear, connection_from_url, and urlopen.
- Supports connection_from_url(url).urlopen(method, host_relative_path) calls
- If we process chunked data from a server through URLFetch, we properly unchunk it for the caller.
- Fix broken timeout detection code.

If I then use this requests_toolbelt code:
https://github.com/sigmavirus24/requests-toolbelt/blob/feature/gae/requests_toolbelt/adapters/appengine.py

I can get requests working on my dev appengine system with URLFetch. This should help fix the issues mentioned in:
https://github.com/kennethreitz/requests/issues/1905
https://github.com/sigmavirus24/requests-toolbelt/pull/114
https://github.com/shazow/urllib3/issues/618

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -34,6 +34,22 @@ class AppEnginePlatformWarning(HTTPWarning):
 class AppEnginePlatformError(HTTPError):
     pass
 
+class _AppEngineConnection(object):
+    """A dummy Connection object that supports the urlopen() interface.
+
+    This Connection's urlopen() is called with a host-relative path,
+    so in order to properly support opening the URL, we need to store the full URL
+    when we are constructed from the Manager.
+    """
+
+    def __init__(self, manager, url):
+        self.manager = manager
+        self.url = url
+
+    def urlopen(self, **kwargs):
+        # We need to use the full URL here, not the host-relative path that is passed in
+        kwargs['url'] = self.url
+        return self.manager.urlopen(**kwargs)
 
 class AppEngineManager(RequestMethods):
     """
@@ -85,9 +101,25 @@ class AppEngineManager(RequestMethods):
         # Return False to re-raise any potential exceptions
         return False
 
-    def urlopen(self, method, url, body=None, headers=None,
-                retries=None, redirect=True, timeout=Timeout.DEFAULT_TIMEOUT,
-                **response_kw):
+    def clear(self):
+        pass
+
+    def connection_from_url(self, url):
+        return _AppEngineConnection(self, url)
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        """Perform an HTTP request using AppEngine's URLFetch.
+
+        To keep compatibility with HTTPConnectionPool.urlopen(), we accept dummy arguments:
+        - assert_same_host
+        - pool_timeout
+        - release_conn
+
+        The other parameters should mostly work as described in HTTPConnectionPool.urlopen,
+        except where necessary to support the limitations of AppEngine's URLFetch.
+        """
 
         retries = self._get_retries(retries, redirect)
 
@@ -164,6 +196,14 @@ class AppEngineManager(RequestMethods):
             if content_encoding == 'deflate':
                 del urlfetch_resp.headers['content-encoding']
 
+        transfer_encoding = urlfetch_resp.headers.get('transfer-encoding')
+        # We have a full response's content,
+        # so let's make sure we don't report ourselves as chunked data.
+        if transfer_encoding == 'chunked':
+            encodings = transfer_encoding.split(",")
+            encodings.remove('chunked')
+            urlfetch_resp.headers['transfer-encoding'] = ','.join(encodings)
+
         return HTTPResponse(
             # In order for decoding to work, we must present the content as
             # a file-like object.
@@ -177,7 +217,7 @@ class AppEngineManager(RequestMethods):
         if timeout is Timeout.DEFAULT_TIMEOUT:
             return 5  # 5s is the default timeout for URLFetch.
         if isinstance(timeout, Timeout):
-            if timeout.read is not timeout.connect:
+            if timeout._read is not timeout._connect:
                 warnings.warn(
                     "URLFetch does not support granular timeout settings, "
                     "reverting to total timeout.", AppEnginePlatformWarning)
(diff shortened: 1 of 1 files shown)


urllib3/contrib/appengine.py
```diff
@@ -177,7 +216,7 @@ def _get_absolute_timeout(self, timeout):
         if timeout is Timeout.DEFAULT_TIMEOUT:
             return 5  # 5s is the default timeout for URLFetch.
         if isinstance(timeout, Timeout):
-            if timeout.read is not timeout.connect:
+            if timeout._read is not timeout._connect:
```
Lukasa: What's this change for?

Lukasa:
This looks reasonable to me at a first glance. You'll need to fix up the linting problems though.

urllib3/contrib/appengine.py
```diff
@@ -85,9 +101,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def clear(self):
+        pass
+
+    def connection_from_url(self, url):
+        return _AppEngineConnection(self, url)
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        """This should be an AppEngine version of HTTPConnectionPool.urlopen().
```
jayvdb: "This should be an " prefix doesnt appear to add much value.

"AppEngine version of HTTPConnectionPool.urlopen()." says the same?

Or pep257 compliant would be something like

"Perform an HTTP request using AppEngine's URLFetch."

shazow: Either of those sound like an improvement, +1.

shazow:
I'm still not convinced it makes sense to add all the extraneous params that are not supported by this backend. See https://github.com/sigmavirus24/requests-toolbelt/pull/114#issuecomment-159454464:

> I think it makes sense for the constructors's options to be relevant to the backend they're implementing, as long as the instance's interface is sufficiently common. I'd be a strong -1 against accepting arguments that are not relevant/not implemented and simply ignoring them (sounds like a recipe for bugs/incorrect user expectations).
>
> If anything, could add a "backport"-ish constructor helper, but honestly I think that's the job of the adapter.

sigmavirus24:
Yeah after the discussion on https://github.com/sigmavirus24/requests-toolbelt/pull/114#issuecomment-159454464 I began to agree with @shazow. I just don't have the motivation to put very much effort into the requests-side of this. I also have higher priority tasks IRL, at work, and on other projects. @mikelambert if you want to take over 114 on the toolbelt side, that's fine.

urllib3/contrib/appengine.py
```diff
@@ -164,6 +196,13 @@ def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
             if content_encoding == 'deflate':
                 del urlfetch_resp.headers['content-encoding']
 
+        transfer_encoding = urlfetch_resp.headers.get('transfer-encoding')
```
jonparrott: +1

urllib3/contrib/appengine.py
mikelambert: The change is to...make things work. :) There is no "read" or "connect" at all in urllib3.util.timeout. Though there is _read and _connect (object properties), and read_timeout and connect_timeout (@properties). https://github.com/kennethreitz/requests/blob/master/requests/packages/urllib3/util/timeout.py

mikelambert:
I've fixed up the docstrings (which was the majority of the linting errors added in my change), and removed the extra **kwargs to __init__, which weren't necessary. (woops!)

Though I've kept the extra parameters listed in urlopen(), because they are necessary for compatibility at some level, due to requests/adapters.py HTTPAdapter.send() calling:
conn.urlopen(...
                    assert_same_host=False,
...)

Since we need to keep "assert_same_host" out of **response_kw (which gets passed to HTTPResponse()), the options as I see them are:
1) Current situation, where AppEngineManager.urlopen supports the same calling spec as HTTPConnectionPool.urlopen, and takes the parameters to keep them out of **response_kw.
2) Make my _AppEngineConnection shim's urlopen() accept the same set of parameters as HTTPConnectionPool.urlopen, but then only pass the relevant subset to AppEngineManager.urlopen (ie, not assert_same_host, pool_timeout, or release_conn).
3) Rewrite request_toolbelt's AppEngineAdapter.send() to copy over a lot of HTTPAdapter.send(), but then call urlopen() with a smaller set of parameters.

I dislike #3 because it feels like AppEngineManager.urlopen is diverging from HTTPConnectionPool.urlopen, when clients (such as requests, and possibly other urllib3 clients?) might want them to remain API-compatible to make it easier to run on AppEngine.

But I'd rather see something working committed, so if the gatekeepers prefer #3, I'm happy to rework it that way.

mikelambert:
@jonparrott since I think he's arguing for #3 in https://github.com/sigmavirus24/requests-toolbelt/pull/114

mikelambert:
Any preferences on which direction to go with this? @jonparrott , @sigmavirus24, @shazow Happy to rewrite whichever way you prefer, to get this committed.

sigmavirus24:
I agree with @jonparrott and @shazow that the adapter for requests that will be added to the toolbelt should be what handles this complexity.

mikelambert:
Alright folks, thanks for voting. I've reworked/rebased my changes to avoid any additional complexity in urllib3. Though honestly, I think AppEngineManager's API is very confusing as-is, and I'm now even more confused what you want urllib3's appengine code to be...

The existing urllib3 AppEngineManager looks like a strange mix of an urllib3 PoolManager, ConnectionPool, and a fake RequestMethods all-rolled-into-one:
- request-toolbelt's AppEngineAdapter sets self.poolmanager = gaecontrib.AppEngineManager(...), expecting it to act like a PoolManager.
- urllib3's AppEngineManager's urlopen() more closely matches ConnectionPool's API than PoolManager's API.
- AppEngineManager subclasses RequestMethods, but then implements urlopen with a different API and doesn't use any of the convenient request* methods, so I'm not sure what the point is.

Instead, as requested, I've moved the urllib3 wrapper code for PoolManager and ConnectionPool to requests-toolbelt. And I try to work-around the strange AppEngineAdapter API (whatever it is, though I am treating it more like a ConnectionPool), and just call it from requests-toolbelt.

shazow:
It's based off of ProxyManager. :)

shazow:
Thanks for the fixes!

PR #767, opened 2015-12-26
Let logger format messages on demand

scop:


Lukasa:
Seems reasonable to me. =)

shazow:
Ah is this equivalent? Any downside? (TIL)

scop:
I don't think there are any downsides. The most prominent upside is avoiding formatting of messages that end up being thrown away due to logging level, and the secondary is slightly cleaner code (IMO).

shazow:
That's great, thank you. :)

PR #775, opened 2016-01-03
WIP: Chunked transfer encoding on HTTP requests

ml31415:
I recently stumbled upon this quite hacky lines in requests: https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L380 What happens there is, that requests has to do dirty low level work, to get chunked transfer encoding going on in HTTP requests. This low level hack is fairly untested there (if I didn't miss anything obvious), and doesn't really integrate well with all the proxy stuff etc. going on on higher levels, especially error handling, as many higher levels of urllib3 are just bypassed. 

But it looks like currently there was no better way to do this, as urllib3 does not support any chunked request by itself. The original `httplib.HTTPConnection` is also lacking this feature. So this pull request is basically adding some separate `request_chunked` method on HTTPConnection and exposes this functionality to upstream libraries, i.e. requests, so that it can make use of it in some later version and get rid of the misplaced lowlevel code.

In terms of implementation, it would have probably been more elgant to just add a `chunked` flag to the `HTTPConnection.request` method itself, but I thought it may be better for backwards compatibility, to have a separate function.


shazow:
Had a quick superficial look, first thoughts:
- Overall looks like a good thing to add, thank you for taking the time of putting together this draft! :)
- As usual, I'd prefer to avoid adding more keywords to urlopen if possible, but I can't think of a better way here. Open to suggestions?
- Some minor test failures in py3 and linting that will need to be fixed. (See https://travis-ci.org/shazow/urllib3/builds/99955520)

Would like to wait on @Lukasa @sigmavirus24 to add their thoughts. :)

Lukasa:
> As usual, I'd prefer to avoid adding more keywords to urlopen if possible, but I can't think of a better way here.

The only way to avoid doing that is to take a Requests-like approach where the behaviour is implicitly controlled by the type of the body. That's probably not a great approach to take for urllib3 (and potentially thorny as hell if you choose a different pattern to the one used by Requests!). I think we might be stuck with a flag here, sadly.

I'll dive in and do some code review.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +842,34 @@ def test_empty_head_response_does_not_hang(self):
...
+    def test_chunked(self):
+        self.buffer = b''
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            while not self.buffer.endswith('\r\n0\r\n\r\n'):
+                self.buffer += sock.recv(65536)
+
+            # Send incomplete message (note Content-Length)
```
Lukasa: This comment is not really valid anymore, I think it's worth removing it.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +842,34 @@ def test_empty_head_response_does_not_hang(self):
...
+        chunks = ['foo', 'bar', 'bazzzzzz']
+        pool = HTTPConnectionPool(self.host, self.port, retries=False)
+        r = pool.urlopen('GET', '/', chunks, headers=dict(DNT='1'), chunked=True)
+
+        self.assertTrue('Transfer-Encoding' in self.buffer)
+        body = self.buffer.split('\r\n\r\n')[1]
+        lines = body.split('\r\n')
+        for i, chunk in enumerate(chunks):
+            self.assertEqual(lines[i * 2], str(len(chunk)))
+            self.assertEqual(lines[i * 2 + 1], chunk)
```
Lukasa: I think we need substantially more testing around this. For example, let's test sending empty chunked bodies, or test the behaviour of setting the `chunked` flag but providing no body at all.

urllib3/connection.py
```diff
@@ -162,6 +162,27 @@ def connect(self):
...
+        """ 
+        Alternative to the common request method, which sends the 
+        body with chunked encoding and not as one block 
+        """
+        header_names = set(k.lower() for k in headers)
+        self.putrequest(method, url, skip_accept_encoding=True)
+        for header, value in headers.items():
+            self.putheader(header, value)
+        if 'transfer-encoding' not in headers:
+            self.putheader('Transfer-Encoding', 'chunked')
```
Lukasa: Hmm. I'm a little nervous about this. There are a few other transfer-encodings defined for HTTP (compress, deflate, and gzip are all defined in RFC 7230. It's *possible*, albeit unlikely, that users may be attempting to use one of those transfer encodings *and* want us to chunk-encode it. If they did, we'd get this wrong.

I think the correct thing to look for is whether `chunked` is in any of the values of the `transfer-encoding` header, and if not to add it by appending to it. That's a little more complex and we may not need to do it if we're happy with the edge-case.

urllib3/connection.py
```diff
@@ -162,6 +162,27 @@ def connect(self):
         conn = self._new_conn()
         self._prepare_conn(conn)
 
+    def request_chunked(self, method, url, body=None, headers={}):
+        """ 
+        Alternative to the common request method, which sends the 
+        body with chunked encoding and not as one block 
+        """
+        header_names = set(k.lower() for k in headers)
```
Lukasa: What's this for? This isn't actually used. I presume you want it a few lines down.

urllib3/connection.py
```diff
@@ -162,6 +162,27 @@ def connect(self):
...
+            self.putheader('Transfer-Encoding', 'chunked')
+        self.endheaders()
+
+        if body is not None:
+            for chunk in body:
+                self.send(hex(len(chunk))[2:].encode('utf-8'))
+                self.send(b'\r\n')
+                self.send(chunk)
+                self.send(b'\r\n')
+            self.send(b'0\r\n\r\n')
```
Lukasa: Right now this block doesn't handle absent bodies correctly. We'll potentially add Transfer-Encoding: chunked as a header, but then send no data at all, leaving the remote server waiting forever for us to terminate the body. I think if we pull this `send` call out of the `if` block that'll be fixed. We should also add the case I just identified as a test case in the test suite. =)

urllib3/connectionpool.py
```diff
@@ -318,7 +318,9 @@ def _raise_timeout(self, err, url, timeout_value):
         if 'timed out' in str(err) or 'did not complete (read)' in str(err):  # Python 2.6
             raise ReadTimeoutError(self, url, "Read timed out. (read timeout=%s)" % timeout_value)
 
-    def _make_request(self, conn, method, url, timeout=_Default,
+
+
+    def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
```
Lukasa: Do we need the extra whitespace here?

ml31415: Nope ) My bad.

test/with_dummyserver/test_socketlevel.py
ml31415: Yes, copy&paste stuff ... Will remove it.

test/with_dummyserver/test_socketlevel.py
ml31415: Agreed, so far it only checks the graceful handling, no errors or oddities. I'll implement your suggestion as tests, open for further suggestions.

urllib3/connection.py
ml31415: Just saw another bug here. It should check for in header_names, not headers.

Is it allowed to define several encodings in the headers? Like chunked gzip? I don't think so. Maybe it should throw an error, if request_chunked is run with another encoding set?

Side note: The original request also sets a transfer-encoding automatically (identity), if none is set.

urllib3/connection.py
ml31415: Yes, had added that a bit later, and then forgot to actually use it... doh.

urllib3/connection.py
ml31415: Ok, good catch.

ml31415:
Please have another look. Hope I didn't forget anything. Any further ideas for test cases?

urllib3/connection.py
Lukasa: It's entirely allowed to have several different transfer encodings, and they're order-specific too: for example, `Transfer-Encoding: gzip, deflate, chunked` says that the originally representation was on-the-fly gzipped, which was then on-the-fly deflated, which was then chunk-encoded and sent, and the remote party needs to dechunk it, then inflate it, then gunzip it to get the original.

Like I said, it's *very* uncommon, but it is technically allowed.

Automatically setting identity is a bit different, though: which library does that? We should really stop, RFC 7230 deprecated it.

ml31415: I was referring to that code: https://github.com/python/cpython/blob/2.7/Lib/httplib.py#L1005 but I had remembered it wrong, it's the accept-encoding that get's set. Looking at the comments there, looks like we're not the only ones who are bit undecided about what to set automatically.

ml31415:
I just wanted to ping this PR. Anything left to do before merging this?

urllib3/connection.py
```diff
@@ -162,6 +162,31 @@ def connect(self):
...
+            self.putheader(header, value)
+        if 'transfer-encoding' not in header_names:
+            self.putheader('Transfer-Encoding', 'chunked')
+        self.endheaders()
+
+        if body is not None:
+            for chunk in body:
+                if not chunk:
+                    continue
+                self.send(hex(len(chunk))[2:].encode('utf-8'))
```
shazow: nit: This line is hard to read, would be nice to break it up or clarify.

shazow:
Still looks fine on my end, but I haven't been as intimate with the protocol lately as @Lukasa so I'll leave it to him. :)

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +842,56 @@ def test_empty_head_response_does_not_hang(self):
...
+
+        self._start_server(socket_handler)
+
+    def test_chunks(self):
+        self._start_chunked_handler()
+        chunks = ['foo', 'bar', '', 'bazzzzzz']
+        pool = HTTPConnectionPool(self.host, self.port, retries=False)
+        r = pool.urlopen('GET', '/', chunks, headers=dict(DNT='1'), chunked=True)
+
+        self.assertTrue(b'Transfer-Encoding' in self.buffer)
```
Lukasa: We can be a bit more aggressive here with this check: we should look for the exact header line and confirm it's well formed. `b'\r\nTransfer-Encoding: chunked\r\n`.

urllib3/connection.py
```diff
@@ -162,6 +162,31 @@ def connect(self):
...
+        header_names = set(k.lower() for k in headers)
+        self.putrequest(method, url, skip_accept_encoding=True)
+        for header, value in headers.items():
+            self.putheader(header, value)
+        if 'transfer-encoding' not in header_names:
+            self.putheader('Transfer-Encoding', 'chunked')
+        self.endheaders()
+
+        if body is not None:
+            for chunk in body:
```
Lukasa: I think this behaves surprisingly for string bodies. It might be good to have a conditional that looks for string bodies and puts them in a one-item list.

urllib3/connection.py
Lukasa: As a follow-up, it is also subtly wrong.

There is an unconditional encode of the chunk to UTF-8 two lines below. If a Unicode string is passed, the `len()` call here may return a smaller number of bytes than the actual number used in the chunk. Given that the unconditional call to `.encode` doesn't work on Python 3 for byte string chunks, I recommend that this method simply not accept Unicode strings at all: remove the call to `.encode` from the chunk send and it essentially becomes mandatory to pass this code bytes.

Lukasa:
The general direction of this is really good, thanks! There are a few small details that need to be addressed though.

ml31415:
Sorry for misusing travisci for "debugging", I got lots of false positives on my setup, so I basically just do the py2 checks locally. Besides that, please have another look, all points should be all addressed now. Good catch with the `type(body) == str` issue.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +842,67 @@ def test_empty_head_response_does_not_hang(self):
...
+    def test_string_body(self):
+        self._start_chunked_handler()
+        pool = HTTPConnectionPool(self.host, self.port, retries=False)
+        body_str = 'thisshouldbeonechunk'
+        r = pool.urlopen('GET', '/', body_str, chunked=True)
+        header, body = self.buffer.split(b'\r\n\r\n', 1)
+
+        self.assertTrue(b'Transfer-Encoding: chunked' in header.split(b'\r\n'))
+        self.assertTrue(b'\r\n' + body_str.encode('utf-8') + b'\r\n' in body)
+        self.assertTrue(body.endswith(b'\r\n0\r\n\r\n'))
```
Lukasa: I think here we can actually just assert exactly what the body should look like. There's really only one valid encoding here.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +842,67 @@ def test_empty_head_response_does_not_hang(self):
...
+        r = pool.urlopen('GET', '/', chunks, headers=dict(DNT='1'), chunked=True)
+
+        self.assertTrue(b'Transfer-Encoding' in self.buffer)
+        body = self.buffer.split(b'\r\n\r\n', 1)[1]
+        lines = body.split(b'\r\n')
+        # Empty chunks should have been skipped, as this could not be distinguished
+        # from terminating the transmission
+        for i, chunk in enumerate([c for c in chunks if c]):
+            self.assertEqual(lines[i * 2], str(len(chunk)).encode('utf-8'))
+            self.assertEqual(lines[i * 2 + 1], chunk.encode('utf-8'))
```
Lukasa: This test should also assert that `len(lines[i * 2 + 1]) == int(lines[i * 2])`

urllib3/connection.py
```diff
@@ -162,6 +162,34 @@ def connect(self):
...
+            if isinstance(body, six.string_types):
+                body = [body]
+            for chunk in body:
+                if not chunk:
+                    continue
+                len_str = hex(len(chunk))[2:]
+                self.send(len_str.encode('utf-8'))
+                self.send(b'\r\n')
+                self.send(chunk.encode('utf-8'))
+                self.send(b'\r\n')
```
Lukasa: So this code is still hurting in one of two cases:

- `chunk` is a Unicode string containing characters outside the ASCII charset (e.g. `u"終あ権7載チ"`). We'll want a test for this because it's easy to get wrong, but your tests actually make the same mistake! I've put a note in that should help the tests catch the error that the code is making here: if you make that change and then add my sample string to the test case you should discover the problem on the Python 3 runs.
- `chunk` is a bytestring on Python 3. On Python 3, calling `encode` on a bytestring is an error.
- `chunk` is a bytestring on Python 2 containing characters encoded using an encoding that is *not* the system default. Calling `.encode()` on a bytestring in Python 2 implicitly calls `decode()` using the system default encoding. This could easily explode.

Again, the simplest fix is to simply not allow unicode strings as chunks *at all*. If you're determined to allow them, then you should check for them explicitly at the top of the loop and encode them there, before calling `len` on them.

Lukasa:
Thanks, this is good progress! I've got a note for you. =)

test/with_dummyserver/test_socketlevel.py
ml31415: I basically tried to keep the body_str variable. If I check for a fixed output, I also have to reimplement the encoding e.g. of the len of the chunk etc.

In general I'd be a big fan of parametrizing tests, so this may be run e.g. also with a randomly generated 1k string, or 100k, just to make sure it doesn't break. Though, unittest sucks at parametrizing, that's why I skipped it.

Lukasa: Yeah, I think right now we're not going to parametrize the tests, so there's really no reason to not be pretty specific here.

urllib3/connection.py
ml31415: In terms of unicode, please have a look at https://github.com/python-mirror/python/blob/2.7/Lib/httplib.py#L814 this is the part where httplib is sending it's body. No check on anything. Feed in a unicode string, and it tries to send it as a file etc.

Now of course there is no point in repeating the mistakes from httplib. Still, httplib relies on the higher level API to present sane input to `request`. The same should be expected for `request_chunked`, as it's supposed to serve as a drop in.

Lukasa: Python 2.7's httplib is not the bar of acceptable that we're trying to meet, Python 3's is. The odds of a Python 2 user passing a unicode chunk here is low, but if they do it we should either error helpfully (rather than letting the implicit encode/decode catch it) or we should try to help them (I'd prefer to do the first of those things). However, the odds of a Python 3 user passing bytes here is very high, and we absolutely **must** do the right thing here in the face of that situation.

To be clear, here is the space of things a user might do and whether this function works correctly:

|                            | Python 2                           | Python 3                           |
|----------------------------|------------------------------------------|----------------------------------|
| Bytestring: only ASCII     | Works (only on some platforms, may fail) | **Fails**: `AttributeError`        |
| Bytestring: not ASCII      | **Fails**: `UnicodeDecodeError`    | **Fails**: `AttributeError`          |
| Unicode string: only ASCII | Works                              | Works                              |
| Unicode string: not ASCII  | **Invalid Output**: Incorrect chunk header. | **InvalidOutput**: Incorrect chunk header. |

It is not acceptable to fail in the face of a bytestring, so all of those failures need to be dealt with. It's also unacceptable to generate invalid output, because this amounts to silent data corruption, which we must not do.

I strongly recommend adding tests for each of the four data types I've listed there and confirm that they all work.

ml31415: Ok, you got a point here. In that case, the test cases should also cover all of the mentioned cases. I'll add that.

ml31415:
Any idea why travis is choking on my umlauts? Looks like valid code for me :(

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,70 @@
...
+            self.assertTrue(body.endswith(b'\r\n0\r\n\r\n'))
+        else:
+            self.assertEqual(body, b'0\r\n\r\n')
+        return header, body
+
+    def test_bytestring_body(self):
+        self._test_body(b'thisshouldbeonechunk')
+
+    def test_unicode_body(self):
+        self._test_body(u'letshavesomefunchars äöüß')
```
sigmavirus24: Python 3.2 is choking on this because `u'foobar'` is invalid syntax. Python 3.0, 3.1, and 3.2 did not support the `u` for a string prefix.

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertTrue(b'Transfer-Encoding: chunked' in header.split(b'\r\n'))
+        if data:
+            self.assertTrue(b'\r\n' + data.encode('utf-8') + b'\r\n' in body)
+            self.assertTrue(body.endswith(b'\r\n0\r\n\r\n'))
+        else:
+            self.assertEqual(body, b'0\r\n\r\n')
+        return header, body
+
+    def test_bytestring_body(self):
+        self._test_body(b'thisshouldbeonechunk')
```
sigmavirus24: This is choking up the rest of your Python 3 tests because when you iterate over a bytes object in Python 3 you get an integer, e.g.,

```py
for byte in b'thisshouldbeonechunk':
    print('type of byte is %s' % type(byte))
```

If you want to iterate over a bytes object the same way in both Python 2 and 3 you have to do

```py
bytes_obj = b'thisshouldbeonechunk'
for i in range(0, len(bytes_obj)):
    byte = bytes_obj[i:i + 1]
    print('type of byte is %s'  % type(byte))
```

test/with_dummyserver/test_socketlevel.py
```diff
@@ -70,7 +37,7 @@ def test_hostname_in_first_request_packet(self):
         def socket_handler(listener):
             sock = listener.accept()[0]
 
-            self.buf = sock.recv(65536) # We only accept one packet
+            self.buf = sock.recv(65536)  # We only accept one packet
```
sigmavirus24: @shazow how do you feel about random fixing of pep8/flake8 violations? Do you like two spaces before an in-line comment or not?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -842,3 +715,4 @@ def test_empty_head_response_does_not_hang(self):
 
         # stream will use the read method here.
         self.assertEqual([], list(r.stream()))
+
```
sigmavirus24: Extra new-line?

ml31415:
Please have another look.

test/with_dummyserver/test_chunked_transfer.py
ml31415: Brr, the iteration wasn't intended at all. Should be fixed now.

ml31415:
Taking care for every single python versions peculiarities is really more pain in the ass, than I thought. Hope this is all sorted out now.

Lukasa:
@ml31415 Welcome to the fun fun world of writing cross-version Python code. ;)

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @onlyPy2
+    def test_unicode_body(self):
+        # Define u'thisshouldbeonechunk äöüß' in a way, so that python3.1
+        # does not suffer a syntax error
+        body = b'thisshouldbeonechunk \xc3\xa4\xc3\xb6\xc3\xbc\xc3\x9f'.decode('utf-8')
+        self._test_body(body)
+
+    @onlyPy3
+    def test_unidoce_body_py3(self):
```
Lukasa: typo: `unicode`. =)

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,80 @@
...
+        self._test_body('thisshouldbeonechunk äöüß')
+
+    def test_empty_body(self):
+        header, body = self._test_body(None)
+
+    def test_empty_string_body(self):
+        header, body = self._test_body('')
+
+    def test_empty_iterable_body(self):
+        header, body = self._test_body([])
```
Lukasa: Presumably all these assignments are unnecessary: let's just leave it as `self._test_body(<thing>)`.

test/with_dummyserver/test_headers.py
```diff
@@ -0,0 +1,131 @@
...
+                      b'Set-Cookie: foo=1\r\n'
+                      b'Set-Cookie: bar=1\r\n'
+                      b'\r\n')
+            sock.close()
+
+        self._start_server(multicookie_response_handler)
+        pool = HTTPConnectionPool(self.host, self.port)
+        r = pool.request('GET', '/', retries=0)
+        self.assertEqual(r.headers, {'set-cookie': 'foo=1, bar=1'})
+        self.assertEqual(r.headers.getlist('set-cookie'), ['foo=1', 'bar=1'])
```
Lukasa: Uh...why did all this get pulled out of its previous location?

urllib3/connection.py
```diff
@@ -162,6 +162,37 @@ def connect(self):
         conn = self._new_conn()
         self._prepare_conn(conn)
 
+    def request_chunked(self, method, url, body=None, headers={}):
+        """
+        Alternative to the common request method, which sends the
+        body with chunked encoding and not as one block
+        """
+        header_names = set(k.lower() for k in headers)
+        self.putrequest(method, url, skip_accept_encoding=True)
```
Lukasa: Why do we need to remove the `accept-encoding` header?

urllib3/connection.py
```diff
@@ -162,6 +162,37 @@ def connect(self):
...
+            for chunk in body:
+                if not chunk:
+                    continue
+                len_str = hex(len(chunk))[2:]
+                self.send(len_str.encode('utf-8'))
+                self.send(b'\r\n')
+                if isinstance(chunk, six.binary_type):
+                    self.send(chunk)
+                else:
+                    self.send(chunk.encode('utf-8'))
```
Lukasa: Let's save ourselves a line here:

```python
if not isinstance(chunk, six.binary_type):
    chunk = chunk.encode('utf-8')
self.send(chunk)
```

This is a style @shazow slightly prefers. =)

Lukasa:
Tests mostly look good, just some code review stuff now. Thanks for bearing with us through this @ml31415, I appreciate that its been a lot of work. You've done great stuff, and now it's just polishing everything up. =) :custard:

test/with_dummyserver/test_chunked_transfer.py
ml31415: You mean, to group it together as one single test_empty_body, for all different versions of "emptiness"? Could be done, though it sucks a bit to do that, as it e.g. won't show when several versions of emptiness fail. pytest is so much better for that stuff ...

test/with_dummyserver/test_headers.py
ml31415: Well, I just took the #TODO on top of test_socketlevel.py serious. I guess the grouping should be ok.

urllib3/connection.py
ml31415: It's not skipping the header, it's skipping the auto-insertion of an 'identity' accept_encoding header. This is the line from httplib:

```
            if not skip_accept_encoding:
                self.putheader('Accept-Encoding', 'identity')
```

test/with_dummyserver/test_chunked_transfer.py
Lukasa: No, I don't, I just mean for each test remove the assignment to `header, body`. Those variables are unused, so we simply don't need them. =)

test/with_dummyserver/test_headers.py
Lukasa: To be clear, I don't object to the change, I just don't want it in *this* pull request: it makes the diff hard to evaluate. If you'd like to make a separate pull request with this change that'd be fine.

urllib3/connection.py
Lukasa: Sure, but *why* are we doing it here and nowhere else?

test/with_dummyserver/test_chunked_transfer.py
ml31415: Yeah, should be more careful with my c&p.

urllib3/connection.py
ml31415: httplib.request also does it, so it basically happens everywhere else, too.

Lukasa: Huh, perplexing. Fair enough then!

test/with_dummyserver/test_headers.py
ml31415: Well, do you really think copy would be worth a separate PR?

urllib3/connection.py
ml31415: Though, I should mimic the original behavoiur better and check for an already set accept-encoding header bevor skipping...

test/with_dummyserver/test_headers.py
Lukasa: Absolutely, yes. =)

A general guideline is that large diffs are hard to review effectively: it's difficult to isolate exactly why a change is being made and whether the change is sensible. For that reason, refactors should be as much as possible isolated from feature improvements, because reviewing the two types of change involves very different processes.

ml31415:
The test fail seems unrelated.

Lukasa:
Yup, test back to passing.

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,80 @@
...
+        return header, body
+
+    def test_bytestring_body(self):
+        self._test_body(b'thisshouldbeonechunk')
+
+    @onlyPy2
+    def test_unicode_body(self):
+        # Define u'thisshouldbeonechunk äöüß' in a way, so that python3.1
+        # does not suffer a syntax error
+        body = b'thisshouldbeonechunk \xc3\xa4\xc3\xb6\xc3\xbc\xc3\x9f'.decode('utf-8')
```
Lukasa: This test should be sufficient for both Python 2 and Python 3, I don't believe either will have a problem with it. There's probably no reason to duplicate the test with a skip. =) Mind removing the decorator here and seeing if it works?

ml31415:
The fail seems unrelated again.

ml31415:
I fixed another bug with the length encoding of unicode strings and improved the tests to catch that.

test/__init__.py
```diff
@@ -67,6 +67,17 @@ def wrapper(*args, **kwargs):
...
+    """Skips this test unless you are on Python3.x"""
+
+    @functools.wraps(test)
+    def wrapper(*args, **kwargs):
+        msg = "{name} requires Python2.x to run".format(name=test.__name__)
+        if six.PY3:
+            raise SkipTest(msg)
+        return test(*args, **kwargs)
+    return wrapper
+
```
Lukasa: I think we no longer need this.

test/with_dummyserver/test_headers.py
Lukasa: This big copy is still here. =)

test/__init__.py
ml31415: Gone.

shazow:
You're all amazing, :custard: for all. Please ping me for final review before merge. :)

test/with_dummyserver/test_headers.py
ml31415: Well, let's say I learned my lessen with the copy and will file separate PRs for that in the future? Reverting this and filing another PR would result in the same outcome right now, and the extra time for having a look at this was already spent. So spending more time on that simple refactor doesn't make it better either.

sigmavirus24: @ml31415 that's a pretty poor argument. Having pull requests that are focused on one thing is good and I'm strongly :-1: on merging this until you split this out.

ml31415:
I guess you can have the final review. Nothing more to add from my side.

urllib3/connection.py
```diff
@@ -162,6 +162,37 @@ def connect(self):
         conn = self._new_conn()
         self._prepare_conn(conn)
 
+    def request_chunked(self, method, url, body=None, headers={}):
+        """
+        Alternative to the common request method, which sends the
+        body with chunked encoding and not as one block
+        """
+        header_names = set(k.lower() for k in headers)
```
sigmavirus24: @shazow would it be better to just use an `HTTPHeaderDict` here? It would avoid having to `str.lower` everything in this method.

urllib3/connection.py
```diff
@@ -162,6 +162,37 @@ def connect(self):
...
+        skip_accept_encoding = 'accept-encoding' in header_names
+        self.putrequest(method, url, skip_accept_encoding=skip_accept_encoding)
+        for header, value in headers.items():
```
sigmavirus24: The header dict also preserves case so we could just have

```py
headers = HTTPHeaderDict(headers)
```

urllib3/connectionpool.py
```diff
@@ -741,7 +745,7 @@ def _prepare_proxy(self, conn):
         except AttributeError:  # Platform-specific: Python 2.6
             set_tunnel = conn._set_tunnel
 
-        if sys.version_info <= (2, 6, 4) and not self.proxy_headers:   # Python 2.6.4 and older
+        if sys.version_info <= (2, 6, 4) and not self.proxy_headers:  # Python 2.6.4 and older
```
sigmavirus24: Another Flake8 "fix"?

urllib3/connection.py
ml31415: HTTPHeaderDict also has to lower everything internally, plus adds a lot of other overhead. I'd be a big fan of reusing HTTPHeaderDict throughout requests+urllib3, but as long as this doesn't happen consistently, there is nothing gained from having it here.

ml31415:
Not sure how you see this, but I'd rather prefer a workflow, where you guys make your comments once, I can fix them, and any further discussion focusses on the changes made. It gets rather annoying, to iterate over and over, and always coming up with something new. No need to make some occupational therapy out of this tiny fix.

Lukasa:
@ml31415 Part of the reason we have a multiple loop approach to this kind of change is that it leads to better code. Any code that lands into urllib3 is code that we're going to have to support for the foreseeable future, so it's really important that there are as few "broken windows" introduced into the codebase as possible.

This is doubly true in a case like this which will have a knock-on effect on requests, as there is an *implicit* partner to this change in the form of code deletion in requests. Given how frequently changes introduced in urllib3 have led to subtle breakages in requests, and given how widely these two projects are used, it's really important that we get this as right as possible.

We're aware that this is a burden, and we're sorry about that, but this is the process behind building good software. To artificially limit ourselves to only commenting on the original diff the first time, and then *not* addressing issues we see in that diff subsequently, is an arbitrary restriction that can only hurt code quality. If at any point you feel like you have maxed out on the amount of time you want to spend on this change, please let us know and someone will take the change off your hands and move it forward.

urllib3/connection.py
sigmavirus24: "It's not used everywhere so there's no point using it here" is a terrible argument for not making a change. If you don't want to continue addressing feedback, that's fine. @Lukasa or I will finish this up for you.

I don't think the header dict adds more overhead in this case. On the contrary, it means we can deal with the headers in the method in an intuitive manner (relying on case-insensitivity when we need it and preserving case to send to the server when we need that.

ml31415:
Of course restricting the discussion exclusively to the changes doesn't make sense, that was not the idea. I just wonder why some things come up now, and not much earlier, as they were there from the first commit on.

Anyways, as @sigmavirus24 suggested, I guess I really leave it up to you, to finish this to your likes, and save some hazzle for all sides involved.

sigmavirus24:
> I just wonder why some things come up now, and not much earlier, as they were there from the first commit on.

Because not all of us look at things from the very beginning. And each time we look at it we may see something different. We're not all super human beings who see every possibility in code the first time we look at something, especially when the size of the changes grows to a very large size.

ml31415:
Sorry guys. I know patience is not one of my strong sides, and I didn't want to offend anyone. As I said, I'll leave this PR as is. When you adjust it to your likes on your own, we save ourselfs lots of time and forth and back, and everyone is happy.

Lukasa:
Don't worry, you didn't cause offence. =) This kind of difficulty happens whenever people attach different amounts of importance to a patch: generally, maintainers come across as absurdly demanding even for small patches. This is because for most contributors, they make the patch and they go: the maintainers have promised to deal with the consequences of that patch for as long as the project lasts. For us, there are no small patches. =)

Generally speaking we'd rather work with people to improve the patch, so that the original author feels a degree of ownership over the work: we're a community project, and that's important to us. However, not everyone has the time to devote to that task, as is the case here. In that case, it's totally fine for you to say that you've done all you can.

I'll grab your branch and pull it out to a local one and update it with the feedback from here.

ml31415:
Sounds great! Thanks for continuing this!

Lukasa:
Ok, this is superseded by #790. Thanks for all your work folks!

shazow:
@ml31415 Thanks for getting the ball rolling, we appreciate your work and hope you stick around. Any feedback/code reviews/whatever is always welcome. :)

Also you're invited to add yourself to our CONTRIBUTORS file.

PR #782, opened 2016-01-14
Support non-tunnelled https proxying

abortz:
CONNECT is nice and all, but terrible for caching -- the world has happily become much more secure! Turns out Squid supports proxying "METHOD https://... HTTP/1.1" requests out-of-the-box, if only a user agent can send them. Yes, you obviously need to do all the SSL verification at the proxy layer, but for a trusted proxy this isn't hard or unreasonable. The alternative this is pseudo-MITM by the proxy which *can* be done (see squid ssl_bump) but is quite terrible.

PR applies cleanly to 1.7.1 as that's Ubuntu LTS's pinned version -- I don't know if it's still compatible with master.

### DIFF
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -199,7 +199,7 @@ class ProxyManager(PoolManager):
     """
 
     def __init__(self, proxy_url, num_pools=10, headers=None,
-                 proxy_headers=None, **connection_pool_kw):
+                 proxy_headers=None, allow_connect=True, **connection_pool_kw):
 
         if isinstance(proxy_url, HTTPConnectionPool):
             proxy_url = '%s://%s:%i' % (proxy_url.scheme, proxy_url.host,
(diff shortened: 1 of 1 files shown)


Lukasa:
This *looks* plausible. However, it definitely doesn't merge cleanly: can you rebase onto the current tip?

I have some mixed feelings about this, but for urllib3 I think it's ok. In requests enabling this would be absurdly footgunny, so I'll have to think really hard about whether we use it. urllib3 is sufficiently explicit about it that I think it's fine.

We'll also need to write up some docs for this feature.

schlamar:
Well, the problem is that the proxy connection is HTTP so you cannot trust the proxy! A rogue proxy could intercept all your data and you won't notice.

If you want caching at proxy level you need to configure your proxy as MITM with a custom cert. In this case you are able to detect a rogue proxy because your custom cert doesn't validate.

PR #783, opened 2016-01-14
Close the underlying connection when closing a response.

### LINKED ISSUE #779: HTTPResponse.close may not close underlying connection.
Found while investigating kennethreitz/requests#2963

The `HTTPResponse` class has a `close` method that rather suggests it will try to close the backing TCP connection behind the given HTTP response. Right now, that's not what happens if the connection is kept alive for any reason (that is, if the server did not send `Connection: close`): instead, the TCP connection will be kept alive and handled as normal.

This seems moderately surprising to me. What it means, in practice, is that calling `HTTPResponse.close()` in both urllib3 and httplib/http.client does not guarantee the closure of the backing TCP connection: instead, in both cases it says "I'm done with the TCP connection, but the underlying connection is free to re-use it". The problems this causes can be see in the `_error_catcher` context manager on the HTTPResponse which does not actually call the class `close` method, presumably because it's too deficient to do the job.

This behaviour affects the chunked transfer encoding decoding logic which calls `self.close()` and therefore may incorrectly keep the connection alive, though it does not itself return the connection to the pool.

I believe it *should* be safe to have `close` close the underlying connection if it is present. As something of an optimisation, we can then safely assume that `close` can call `release_conn`, which will allow us to keep hold of the `HTTPConnection` object in a situation where otherwise we might lose it.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29731274-httpresponse-close-may-not-close-underlying-connection?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
If the connection is still on the response object, closing the response should assume the connection is busted and close that too.

Resolves #779.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -436,6 +436,13 @@ class TestSocketClosing(SocketDummyServerTestCase):
     def test_closing_response_actually_closes_connection(self):
         done_closing = Event()
         complete = Event()
+        # The insane use of this variable here is to get around the fact that
+        # Python 2.6 does not support returning a value from Event.wait(). This
+        # means we can't tell if an event timed out, so we can't use the timing
+        # out of the 'complete' event to determine the success or failure of
+        # the test. Python 2 also doesn't have the nonlocal statement, so we
+        # can't write directly to this variable, only mutate it. Hence: list.
+        successful = []
 
         def socket_handler(listener):
             sock = listener.accept()[0]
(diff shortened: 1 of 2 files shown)


Lukasa:
Lovers of crappy old Python versions will enjoy the hack I had to use to get the test to run properly on Python 2.6.

shazow:
:custard:

PR #785, opened 2016-01-20
Always use setuptools

dstufft:
Using distutils instead of setuptools is discouraged, it doesn't properly add all of the metadata that Python packaging requires. In addition, in 2016 you are almost certainly guaranteed to have setuptools exist on a user's machine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,10 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+from setuptools import setup
 
 import os
 import re
 
-try:
-    import setuptools  # noqa: unused
-except ImportError:
-    pass  # No 'develop' command, oh well.
 
 base_path = os.path.dirname(__file__)
 


shazow:
Any idea if this would affect any distro maintained bundling?

dstufft:
No, it'll just add a build-depends on python-setuptools.

shazow:
Is there no Py32 support? https://travis-ci.org/shazow/urllib3/jobs/103506521

dstufft:
That's unrelated, virtualenv and pip no longer support Python 3.2 and you're upgrading virtualenv to the new version in your travis job.

shazow:
YAY more stuff to fix.

shazow:
@sigmavirus24 @Lukasa any other concerns I'm not thinking of? (Good to merge otherwise)

sigmavirus24:
No concerns, I'm very :+1: on this.

Lukasa:
I'm +1 as well, make it so.

Issue #651, opened 2015-06-13
Likely bug in retry/release-conn code, need tests


shazow:
In this change: https://github.com/shazow/urllib3/pull/647/files#diff-211abff6a07837e4a89a8663a89d2c84R582

@jlatherfold writes:

> For the (TimeoutError, HTTPException, SocketError, ConnectionError) exception handler you don't raise (presumably because there may be retries) and you will go on to release the conn and make another call to urlopen (if retries). But you call it with release_conn=True (since you've just set it to true) - what happens if that call succeeds? You'll release the conn back to the pool before the response is read. At least that's what it looks like to me.

jlatherfold:
I will try and confirm this next week sometime, but writing a test to force it might not be so easy.

shazow:
Take a look at the socketlevel tests in #647, should be some good examples.

jlatherfold:
Hi,

I ran a few trial runs with your changes and set read timeout low enough to force the issue. I set retries=1 and got a lot of retries that succeeded so the problem that I thought would occur didn't - I couldn't confirm that this is an issue. You could probably be certain if you did something like this (ie. use a local varaible instead of release_conn):

put_conn_on_error = False
try:
    # make call...
except:
    conn = conn and conn.close()
    put_conn_on_error = True
finally:
    if release_conn or put_conn_on_error:
        self._put_conn(conn)
    
if retries:
    # try again...

Anyway, it seems ok as it is and I won't be doing any further tests on this for now.

shazow:
Maybe, though not as certain as actually having a test in the suite which exercises this scenario. :) We'll see if we can add that later, thanks for trying!

PR #788, opened 2016-01-24
fix 'bool' object has no attribute 'close'

boyxuper:
HTTPConnectionPool.urlopen may pass connection=False to HTTPResponse when release_conn set to True
which can cause AttributeError: 'bool' object has no attribute 'close'

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -239,6 +239,33 @@ class TestSocketClosing(SocketDummyServerTestCase):
         finally:
             timed_out.set()
 
+    def test_delayed_body_read_timeout_with_preload(self):
+        timed_out = Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+            buf = b''
+            body = 'Hi'
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+            sock.send(('HTTP/1.1 200 OK\r\n'
+                       'Content-Type: text/plain\r\n'
+                       'Content-Length: %d\r\n'
+                       '\r\n' % len(body)).encode('utf-8'))
+
+            timed_out.wait(5)
+            sock.close()
+
+        self._start_server(socket_handler)
+        pool = HTTPConnectionPool(self.host, self.port)
+
+        try:
+            self.assertRaises(ReadTimeoutError, pool.urlopen,
+                              'GET', '/', retries=False,
+                              timeout=Timeout(connect=1, read=0.001))
+        finally:
+            timed_out.set()
+
     def test_incomplete_response(self):
         body = 'Response'
         partial_body = body[:2]
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -252,7 +252,7 @@ class HTTPResponse(io.IOBase):
             # Closing the response may not actually be sufficient to close
             # everything, so if we have a hold of the connection close that
             # too.
-            if self._connection is not None:
+            if self._connection:
                 self._connection.close()
 
             raise
@@ -387,7 +387,7 @@ class HTTPResponse(io.IOBase):
         if not self.closed:
             self._fp.close()
 
-        if self._connection is not None:
+        if self._connection:
             self._connection.close()
 
     @property
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -561,7 +561,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # the request doesn't need to know about the connection. Otherwise
             # it will also try to release it and we'll have a double-release
             # mess.
-            response_conn = not release_conn and conn
+            response_conn = conn if not release_conn else None
 
             # Import httplib's response into our own wrapper object
             response = HTTPResponse.from_httplib(httplib_response,


boyxuper:
and I also found that before HTTPResponse.urlopen retires, response_conn can be changed, which makes unexpected behavior.
as #651 described

Lukasa:
Thanks for this!

Firstly, it'd be good to have a test that reproduces this behaviour so that we don't regress it. Secondly, would it be better to change the logic so that instead of passing `connection=False` we pass `connection=None` instead?

boyxuper:
hi, @Lukasa 
1. The original exception has been overridden by this exception, the traceback:
[E 160124 16:43:23 client:840] Unexpected request failure, re-raising.
      File "/opt/www/b/lib/python2.7/site-packages/etcd/client.py", line 802, in wrapper
        _ = response.data
      File "/opt/www/b/lib/python2.7/site-packages/urllib3/response.py", line 166, in data
        return self.read(cache_content=True)
      File "/opt/www/b/lib/python2.7/site-packages/urllib3/response.py", line 311, in read
        flush_decoder = True
      File "/usr/local/lib/python2.7/contextlib.py", line 35, in __exit__
      File "/opt/www/b/lib/python2.7/site-packages/urllib3/response.py", line 256, in _error_catcher
        self._connection.close()

2. I think it's good idea, but I'm not very familiar with urllib3 source code. I think I'll also change all `if self._connection` check to if self._connection is not None, will this be okay?

Lukasa:
Hmm, that's a moderately tricky one. Do you have a good repro scenario?

shazow:
Hm I think it would be more defensive to check for a truth-y/false-y object where the `has no attribute` line occurs (where is that?).

I don't think we get any benefit out of checking whether it's `None` specifically (we'd probably just introduce this same bug into a differently flow).

shazow:
Ah your hypothesis is it's happening here? https://github.com/shazow/urllib3/commit/3cac4497f8a006a08946aa1ed991a9d99d1ed80a#commitcomment-15635601

Right this would happen if `response_conn is True`, which should not happen.

boyxuper:
update, I added a line to log exception, Traceback is here
      File "/opt/www/a/lib/python2.7/site-packages/urllib3/response.py", line 235, in _error_catcher
        raise ReadTimeoutError(self._pool, None, 'Read timed out.')
    ReadTimeoutError: HTTPConnectionPool(host='10.13.85.97', port=2379): Read timed out.

boyxuper:
My solution is quick fix, yes, "we'd probably just introduce this same bug into a differently flow“. I think the self._connection can ONLY be two values: None or valid urllib3.HTTPConnection instance, any other type is invalid. And current fix uniforms the "should operate self._connection" decision making logic.
I'm not sure we should refactor those logic this time.

Lukasa:
@boyxuper Out of interest, what version of urllib3 are you using to produce that traceback?

boyxuper:
urllib3==1.14

boyxuper:
here's reproduce case: https://gist.github.com/boyxuper/bcdd811a540405feab0d

Lukasa:
Ok, so it should be fairly easy to add a test for this case. @boyxuper, do you feel like you're prepared to write that test, or would you rather I did it and provided it to you as a pull request to your branch?

boyxuper:
@Lukasa 
I'll try to write this test, I'm reading test/contrib/test_socks.py.

Lukasa:
test_socketlevel.py might be more instructive.

boyxuper:
thx, @Lukasa 
test case added.

urllib3/response.py
```diff
@@ -150,7 +150,7 @@ def get_redirect_location(self):
         return False
 
     def release_conn(self):
-        if not self._pool or not self._connection:
+        if not self._pool or self._connection is None:
```
shazow: Is there any benefit to changing `not self._connection` to `self._connection is None`? I feel like the former is safer, no?

boyxuper: I think the self._connection can only be two values: None or urllib3.HTTPConnection, any other type should considered invalid. Strict type checking maybe better?

shazow: My arguments would be:

1. We should be consistent, and we rely on falsey values everywhere else where None does not have a special meaning separate from False.
2. This bug happened precisely because we're doing a check for `is None` rather than a falsey check: https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L255

I'd prefer we fix https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L255 rather than change this to a None check.

boyxuper: I agree with you if the code is already relying on falsey values.
Also take a look at https://github.com/boyxuper/urllib3/blob/master/urllib3/response.py#L390 .
I'll update the pull request, please correct me if something wrong.

boyxuper:
@shazow  @Lukasa 
The pull request updated, please correct me if something wrong.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -239,6 +239,34 @@ def socket_handler(listener):
...
+            body = 'Hi'
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+            sock.send(('HTTP/1.1 200 OK\r\n'
+                       'Content-Type: text/plain\r\n'
+                       'Content-Length: %d\r\n'
+                       '\r\n' % len(body)).encode('utf-8'))
+
+            timed_out.wait()
+            sock.send(body.encode('utf-8'))
```
Lukasa: We should avoid this `send` call because there's a risk it'll cause an exception to be thrown and prevent the cleanup of the thread. Just go straight to the close call. =)

test/with_dummyserver/test_socketlevel.py
```diff
@@ -239,6 +239,34 @@ def socket_handler(listener):
...
+            buf = b''
+            body = 'Hi'
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+            sock.send(('HTTP/1.1 200 OK\r\n'
+                       'Content-Type: text/plain\r\n'
+                       'Content-Length: %d\r\n'
+                       '\r\n' % len(body)).encode('utf-8'))
+
+            timed_out.wait()
```
Lukasa: You should set a timeout here, to make sure we don't get stuck here forever and leak the thread. Change this to `timed_out.wait(5)`.

Lukasa:
This is great @boyxuper! I have two really small notes about the test, but once those are fixed I think it should be good to merge.

test/with_dummyserver/test_socketlevel.py
boyxuper: There's a finally block which will set the event, is that doesn't work? Or I've missed something?

Lukasa: The finally block *should* work, but in general it's a bit better to avoid the risk of deadlocking. It also prevents later changes to this code from missing that subtlety and accidentally removing the finally block.

boyxuper: @Lukasa 
thx for explanation, updating this pr.

shazow:
Thank you, looks excellent! :custard:

PR #789, opened 2016-01-25
Support a concatenated gzip data

smtakeda:
Currently `requests`/`urllib3` doesn't support a concatenated gzip data. Only the first gz content is extracted. Here is an reproducible test case.

1.Create a text file `abc.txt` including three lines of data:

```
1
2
3
```

2.Update `GzipSimpleHttpServer` 
https://github.com/smtakeda/GzipSimpleHTTPServer/commit/68506651620840ba15cc6f60b4d15a3e7719b8c9

3.Launch `GzipSimpleHttpServer`

```
python GzipSimpleHTTPServer.py 8001
```

4.Run this:

```
import requests
ret = requests.get('http://localhost:8001/abc.txt')
print(ret.text)
```

You'll get `1` only, since `zlib.decompress` stops decoding at the end of the first gz data.

The proposed fix is feed https://docs.python.org/3/library/zlib.html#zlib.Decompress.unused_data into a new `Decompress` object until no `unused_data` is observed.

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -292,6 +292,29 @@ class TestResponse(unittest.TestCase):
 
         self.assertRaises(StopIteration, next, stream)
 
+    def test_concatenated_gzip_streaming(self):
+        import zlib
+        data = []
+        uncompressed_data = []
+        for i in range(3):
+            compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
+            ud = b'foo{0}'.format(i)
+            d = compress.compress(ud)
+            d += compress.flush()
+            data.append(d)
+            uncompressed_data.append(ud)
+        data = ''.join(data)
+        uncompressed_data = ''.join(uncompressed_data)
+        fp = BytesIO(data)
+        resp = HTTPResponse(fp, headers={'content-encoding': 'gzip'},
+                         preload_content=False)
+        stream = resp.stream()
+
+        # Read everything
+        payload = next(stream)
+        self.assertEqual(payload, uncompressed_data)
+        
+
     def test_deflate_streaming_tell_intermediate_point(self):
         # Ensure that ``tell()`` returns the correct number of bytes when
         # part-way through streaming compressed content.
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -55,7 +55,12 @@ class GzipDecoder(object):
     def decompress(self, data):
         if not data:
             return data
-        return self._obj.decompress(data)
+        ret = self._obj.decompress(data)
+        while len(self._obj.unused_data) > 0:
+            unused_data = self._obj.unused_data
+            self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
+            ret += self._obj.decompress(unused_data)
+        return ret
 
 
 def _get_decoder(mode):


smtakeda:
ah ok, the test case should be included... will work on it.

Lukasa:
This is a tricky one.

The concatenated gzip format is...unusual. Primarily, it's unusual because it is used as a hack for concatenating multiple bits of unrelated output: for example, when used from the command line it allows you to essentially build an archive that preserves the filenames of each compressed file.

Certainly, I can't find evidence of support for this format in any other user-agent: curl doesn't support it (it prints `1` and then exits), Chrome doesn't support it (it gets the 1 but then basically writes out the remaining unused data), Safari doesn't support it (it prints 1).

This makes me disinclined to want to add support for the format into urllib3. In particular, it makes us prone to attacks on the zlib compression logic by including unused gunk at the back of the file that we will then attempt to decompress. That seems like an unnecessary vector to me.

Is there any reason that incremental gzip compression wouldn't be more useful in this case? (i.e. using zlib.compressobj)

smtakeda:
Hi, thanks for a quick response. Yes, our company's DB engine compresses data in gzip in parallel, concatenates them and stores it in S3 in a special mode. I've implemented Python connector that downloads data using `requests`, and we discovered the second and further gzip data were ignored.

As you pointed out, Chrome, Safari don't support it, yet IE does according to my colleague who implemented the server code. Java has some library that can have the fix. So far I've done monkey patch to make this work.

In my understanding, if the data are compressed in parallel, it would be faster DB engine to send back to the client, i.e., compress each and concat multiple gz files should be faster than compressing a large file.

Lukasa:
@smtakeda Given the use-case you have, which is presumably encrypting quite a lot of files, it seems like it may be better to simply not use the transparent gzip decryption that urllib3 provides, and instead to handle this with a custom decoder. You can disable the transparent decompression that urllib3 provides when using requests by using `response.raw` directly.

Another alternative would be to use something like [`gzjoin`](http://svn.ghostscript.com/ghostscript/tags/zlib-1.2.3/examples/gzjoin.c) at the end of your compression pipeline to correctly concatenate those files together: this should resolve the problem of only seeing one chunk, and works across all user-agents rather than relying on each one to independently make a change that works for this use-case. This does have the downside of needing to *decompress* the data in order to correctly build the CRC32, but should otherwise be ok.

I remain reluctant to add support for this in urllib3: this is such an uncommon use-case that I don't know it justifies the maintenance burden of the code.

smtakeda:
ah ok, `response.raw` worked. I thought it would not, because `iter_content` didn't:
http://docs.python-requests.org/en/latest/user/quickstart/#raw-response-content

This is a test code, and I can live with this in my application.
```
import requests
import zlib

ret = requests.get(
    'http://localhost:8001/abc.txt', stream=True)

chunk_size = 8192

data = []
d = ret.raw.read(chunk_size)
while d:
    data.append(d)
    d = ret.raw.read(chunk_size)

all_data = ''.join(data)

idx = 0
max_len = len(all_data)

obj = zlib.decompressobj(16+zlib.MAX_WBITS)
output = ''
while idx < max_len:
    if max_len - idx < chunk_size:
        end_idx = max_len
    else:
        end_idx = idx + chunk_size
    print("> start {0}, end {1}".format(idx, end_idx))
    file_content = all_data[idx:end_idx]
    output += obj.decompress(file_content)
    while len(obj.unused_data) > 0:
        print('>> unused_data: {0}'.format(len(obj.unused_data)))
        unused_data = obj.unused_data
        obj = zlib.decompressobj(16+zlib.MAX_WBITS)
        output += obj.decompress(unused_data)
    idx += chunk_size
    print('>>> {0}'.format(output.split('\n')))
output += obj.flush()
print(output)
```

Great suggestion about the compression end. I'll let the colleague know and see if we can incorporate it.

Thank you very much!

shazow:
This is a very interesting use case, @smtakeda would you be up for writing a little recipe tutorial to be included in our docs?

smtakeda:
sure, @shazow. Would the doc be in requests or urllib3? Btw, I came up with a simpler code where the parameter is `response.raw`. This assumes the data encoding is text UTF-8 and returns a Unicode data.

```
CHUNK_SIZE = 8192
def decompress_raw_data(raw_data_fd):
    obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
    output = []
    d = raw_data_fd.read(CHUNK_SIZE)
    while d:
        output.append(obj.decompress(d))
        while obj.unused_data != b'':
            unused_data = obj.unused_data
            obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
            output.append(obj.decompress(unused_data))
        d = raw_data_fd.read(CHUNK_SIZE)
    output.append(obj.flush())
    return b''.join(output).decode(u'utf-8')
```

Lukasa:
@smtakeda It'd be much better to have the doc in urllib3's docs, I think. =)

shazow:
Yea, feel free to add a new Recipes section in our docs/ and toss it in there (and add yourself to the contributors file!).

PR #790, opened 2016-01-25
Chunked transfer encoding on HTTP requests

Lukasa:
Supersedes #775.


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+
+from urllib3 import HTTPConnectionPool
+from urllib3.packages import six
+from dummyserver.testcase import SocketDummyServerTestCase
+
+
+class TestChunkedTransfer(SocketDummyServerTestCase):
+    def _start_chunked_handler(self):
```
sigmavirus24: Does this need to be private? I know start_basic_handler isn't.

Lukasa: Uh, I don't think so, no. Let me fix that up.

Lukasa: Done.

sigmavirus24:
:+1: from me.

urllib3/connectionpool.py
```diff
@@ -434,7 +437,8 @@ def is_same_host(self, url):
 
     def urlopen(self, method, url, body=None, headers=None, retries=None,
                 redirect=True, assert_same_host=True, timeout=_Default,
-                pool_timeout=None, release_conn=None, **response_kw):
+                pool_timeout=None, release_conn=None, chunked=False,
```
shazow: docstring for the new keyword please

urllib3/connection.py
```diff
@@ -162,6 +164,38 @@ def connect(self):
         conn = self._new_conn()
         self._prepare_conn(conn)
 
+    def request_chunked(self, method, url, body=None, headers={}):
```
shazow: `headers={}` as a kw value is a mutability bug waiting to happen, let's do what we do everywhere else (`headers=None` and set the default right after).

shazow:
Minor style comments, rest looks good. Ping when ready. :) :custard:

Lukasa:
Ping!

urllib3/connection.py
sigmavirus24: We're not using that value at all. We make a copy with `HTTPHeaderDict`. I'm not sure using `None` as a default really buys us much more safety.

PR #795, opened 2016-01-26
Enable PyOpenSSL testing.

### LINKED ISSUE #791: Add PyOpenSSL tests to the regular test runs.
I've just noticed that, while we have tests that can run under PyOpenSSL, they don't run either under tox or Travis. I suggest that we promote those tests up to be full parts of our test suite, especially as PyOpenSSL can be critically important for some features.

While we're there, we should add `TestSSL` to the import from `test_socketlevel`, as the `SSL` tests are probably a good thing to have run under PyOpenSSL as well!

This will give us a hook for writing tests that only run under PyOpenSSL and the expectation that they will actually run on multiple platforms.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30191816-add-pyopenssl-tests-to-the-regular-test-runs?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
This change brings testing of the PyOpenSSL logic into urllib3's CI and testing frameworks. It also enables testing of the PyOpenSSL code in Python 3, and fixes a few bugs we had that prevented that from functioning properly.

Unfortunately, I had to rewrite the server certificate, because this change revealed that our server certificate would raise warnings under PyOpenSSL because it had no dnsname subject alternative name fields. This didn't affect the standard library because the standard library would also return email fields. This change unfortunately extends the footprint of the total patch quite a bit.

I so far have not enabled code coverage for PyOpenSSL. This is in part because getting accurate numbers for it will be tricky: there's large chunks of code that are specific to error conditions or to individual Python versions that we have not fully investigated. We should investigate how best to obtain code coverage here, which may involve changing our reporting logic to look a bit more like what @hynek suggests [here](https://hynek.me/articles/testing-packaging/).

For the sake of @sigmavirus24 I should also note that this change has a potentially destabilising effect on requests. Previously requests would *not* automatically use the PyOpenSSL code because the import failed (due to the import of `socket._fileobject`). This change makes that code path functional, which means that Python 3 installations of requests may suddenly take a new and potentially under-tested code path. We'll want to come up with a battle plan here.

Resolves #791.


Lukasa:
Ok, so the remaining failure is PyPy, because of restrictions about the version of cryptography that can be used with Travis' ancient version of PyPy. Any suggestions for what we should do here @reaperhulk?

reaperhulk:
The only thing that can be done is a manual install of pypy 4.0.1 :disappointed: . It's not hard but whether you want to accept that complexity into your travis config is another question. Unfortunately, Travis env updates are likely still several months away.

urllib3/contrib/pyopenssl.py
```diff
@@ -167,13 +178,60 @@ def __init__(self, connection, socket, suppress_ragged_eofs=True):
...
         return self.socket.fileno()
 
-    def makefile(self, mode, bufsize=-1):
-        self._makefile_refs += 1
-        return _fileobject(self, mode, bufsize, close=True)
+    if _fileobject:  # Platform-specific: Python 2
+        def makefile(self, mode, bufsize=-1):
+            self._makefile_refs += 1
+            return _fileobject(self, mode, bufsize, close=True)
+    else:  # Platform-specific: Python 3
```
shazow: I'd vote on factoring this out, maybe into a backport submodule?

urllib3/util/__init__.py
```diff
@@ -6,6 +6,7 @@
 from .ssl_ import (
     SSLContext,
     HAS_SNI,
+    IS_PYOPENSSL,
```
shazow: Maybe `HAS_PYOPENSSL`? (~0 on it either way, just a thought.)

urllib3/contrib/pyopenssl.py
Lukasa: How do you envision this looking? `backports.makefile` exposing a pair of functions that I just assign, eg:

```python
try:  # Platform-specific: Python 2
    from socket import _fileobject
except ImportError:  # Platform-specific: Python 3
    _fileobject = None
    from .packages.backport_makefile import makefile, _decref_socketios as backport_makefile, backport_decref_socketios

class WrappedSocket(object):
    if _fileobject:
        def makefile(self, mode, bufsize=-1):
            self._makefile_refs += 1
            return _fileobject(self, mode, bufsize, close=True)
    else:
        makefile=backport_makefile
        _decref_socketios=backport_decref_socketios
```

That makes me a bit nervous because it requires that implementation details of the `WrappedSocket` class leak into the backport module, but it might be fine.

urllib3/util/__init__.py
Lukasa: How about `USING_PYOPENSSL`?

shazow: IS_ is fine, I was just wondering if we wanted to normalize towards consistency with `HAS_SNI`

urllib3/contrib/pyopenssl.py
shazow: Could handle the if-exists logic in the backport module itself, so it would just be a plain import. I'm thinking of something like what we're doing with `ssl_match_hostname` (in fact, it should probably be in `packages` too).

urllib3/util/__init__.py
Lukasa: Yeah, I thought about that, but they're not really the same kind of thing. `HAS_SNI` is feature detection: `IS_PYOPENSSL` is mostly just a warning flag (and in this case only really exists for testing at this time, though we may want to use it later).

urllib3/contrib/pyopenssl.py
Lukasa: So instead we'd have:

```python
from .packages.backport_makefile import makefile, _decref_socketios as backport_makefile, backport_decref_socketios

class WrappedSocket(object):
    makefile = backport_makefile
    _decref_socketios = backport_decref_socketios
```

Is that what you're thinking?

shazow: Sure, that's fine (though I think the `as` syntax is incorrect in your example, should be `import foo as bar, baz as quux` I think?).

Also, is it worth including _decref_socketios? that seems like a small enough standalone thing to keep inline.

urllib3/util/__init__.py
shazow: Ah, they both seem like feature detection to me. Meh, I don't have strong feelings about this. :)

urllib3/contrib/pyopenssl.py
Lukasa: Ok, I've done that backport.

Lukasa:
Tests will still be failing because of PyPy, which is what I'll look at next.

Lukasa:
Green test run! :sparkles:

shazow:
+1, looks good. Anything left to do? Feel free to merge. :)

Lukasa:
I think we're golden, but I'd like to let @sigmavirus24 pull the trigger (just to get all the people involved!)

shazow:
```python
for chant in xrange(50):
  print("sigma!")
```

caizixian:
Hooray!

PR #799, opened 2016-02-04
Changes to support sending a complete response object when 407 errors…

angelcaido19:
… is found while doing the CONNECT to proxy. Added a way to stablish a connection using connect over an already connected socket

Trying to connect to a "https" web page using a proxy uses a "HTTP CONNECT" verb to tell the proxy that we want to connect to a https destination.
With "basic authentication" the "CONNECT" is sent with the appropriate headers so the proxy can authenticate the request and do the request to the https web server. (That headers are setup by the requests library)
With "NTLM" we use a plugin (requests-ntlm) that is set as a hook and it is called on every request response. For example: with HTTP, the requests library sends a "GET" to a web page, as a response we get a HTTP-407 that is handled by the plugin hook to do the NTLM negotiation and authentication
In a HTTPS scenario, the first connection to the proxy is being done with the "CONNECT". At that time, requests does not set any authentication header (because it is handled by a plugin) and we get an 407 authentication Error.

This Patch tries to avoid raising an exception on "Proxy Authentication" errors and it sends back a complete http response so it can be handled by the requests-ntlm plugin and reuse an already connected and authenticated socket to do the connection

### DIFF
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -10,9 +10,11 @@ from .packages import six
 try:  # Python 3
     from http.client import HTTPConnection as _HTTPConnection
     from http.client import HTTPException  # noqa: unused in this module
+    from http.client import PROXY_AUTHENTICATION_REQUIRED
 except ImportError:
     from httplib import HTTPConnection as _HTTPConnection
     from httplib import HTTPException  # noqa: unused in this module
+    from httplib import PROXY_AUTHENTICATION_REQUIRED
 
 try:  # Compiled with SSL?
     import ssl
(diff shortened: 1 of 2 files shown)


Lukasa:
Firstly, thanks for this pull request!

So this is a great idea, but I have some concerns with the specifics of the implementation. In particular, I think I very much don't like changing `connect` to have a return value. `connect` in this context really has nothing to do with the CONNECT verb *per se* and everything to do with the underlying socket connection. This method is frequently called transparently by httplib, and httplib absolutely does not expect that we'll return a value out of the connect function.

This means that I'm extremely unsure about how best to handle this problem. It's not even entirely clear to me that it's a good idea to let the 407 bubble up in the standard way. Is it potentially more useful to throw a specific exception and let the requests_ntlm library (and anyone else who cares) catch it?

angelcaido19:
My approach was to return a full response object from urrlib3 so we can use it directly at requests and then interact with request_ntlm via hooks.
Catching the exception would involve to create a "mocked" response to be used by requests. Do you thing doing that is better?
Right now I dont have much time to do that but if we can discuss the better way to solve it maybe I can do it later!

sigmavirus24:
Hey @angelcaido19 thanks for pointing out this deficiency to us. None of us deal with proxies on a regular basis, let alone proxies that require NTLM authentication. Your usage and experience is invaluable here.

I tend to agree with @Lukasa. The `connect` method now *might* return something and when it does, it's a response object. That seems like it will produce some very difficult to debug problems. I wonder if we could potentially handle this case via some exception handling. I think the standard library raises an exception (IOError? OSError? I forget which of these) when the status code is not 200 (from `_tunnel`). I wonder if it's kind enough to attach the status code to the exception as well.

PR #801, opened 2016-02-10
Better handle the IPv6 debacle.

Lukasa:
This is intended to resolve kennethreitz/requests#3002.

It reverts #708 and #760.

The basic problem originally found in #707 was *actually* because `connection_from_url` would sometimes pass `None` to the connection, which screws up the logic. For that reason, I thought I'd bring in the logic used in `PoolManager.connection_from_url`.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -69,7 +69,8 @@ class ConnectionPool(object):
         if not host:
             raise LocationValueError("No host specified.")
 
-        self.host = host
+        # httplib doesn't like it when we include brackets in ipv6 addresses
+        self.host = host.strip('[]')
         self.port = port
 
     def __str__(self):
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -70,6 +70,11 @@ class ConnectionPool(object):
             raise LocationValueError("No host specified.")
 
         # httplib doesn't like it when we include brackets in ipv6 addresses
+        # Specifically, if we include brackets but also pass the port then
+        # httplib crazily doubles up the square brackets on the Host header.
+        # Instead, we need to make sure we never pass ``None`` as the port.
+        # However, for backward compatibility reasons we can't actually
+        # *assert* that.
         self.host = host.strip('[]')
         self.port = port
 
@@ -830,6 +835,7 @@ def connection_from_url(url, **kw):
         >>> r = conn.request('GET', '/')
     """
     scheme, host, port = get_host(url)
+    port = port or port_by_scheme.get(scheme, 80)
     if scheme == 'https':
         return HTTPSConnectionPool(host, port=port, **kw)
     else:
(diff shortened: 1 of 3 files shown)


dummyserver/certs/server.ipv6addr.crt
```diff
@@ -1,16 +0,0 @@
------BEGIN CERTIFICATE-----
```
sigmavirus24: Why are we removing these files?

Lukasa: Because they were introduced in #761, which is being reverted here. We could leave the tests in place if you think that's sensible.

urllib3/connectionpool.py
```diff
@@ -829,6 +835,7 @@ def connection_from_url(url, **kw):
         >>> r = conn.request('GET', '/')
     """
     scheme, host, port = get_host(url)
+    port = port or port_by_scheme.get(scheme, 80)
```
shazow: Is this actually necessary in combination of the [] stripping?

dummyserver/certs/server.ipv6addr.crt
shazow: I think the tests would be nice to have, good reference for doing similar stuff in the future at least.

urllib3/connectionpool.py
Lukasa: Yes.

If you pass an IPv6 literal without square brackets to httplib while passing `port=None` then it will attempt to parse a port out of the IP address, which will cause it to throw an exception. We *need* to pass a non-`None` port.

Lukasa:
Looks like the sum total of this change is that we now have a branch of a comparison that never gets used. I'll see if I can hit it in testing.

Lukasa:
So the reason this comparison is never hit is because `self.port` is never `None` anymore on connection pool objects. I think that might just be fine.

Lukasa:
Oh, crap, I didn't give you a good action there @shazow. My proposal is to either exclude that branch from coverage or assert it to be unreachable. That, or we have to manually initialize a `ConnectionPool` for this test. Preferences?

shazow:
Either one is fine, I'm happy with a manually-initialized object just for coverage.

Half the reason why I like coverage is for the same benefits we get from static languages—makes sure the code doesn't crash on basic usage and maintaining that things still work once we change it. :P

Lukasa:
Ok, there's our test coverage. =)

sigmavirus24:
:+1: from me

shazow:
:custard:

ralphbean:
FYI, we're going to look at applying this patch to the 1.13.1 release downstream in Fedora (unless there are objections).

shazow:
I was thinking of pushing out v1.15 next week, fwiw.

sigmavirus24:
@shazow I think @ralphbean is keeping urllib3 at 1.13.1 for requests.

PR #802, opened 2016-02-10
[WIP] Experimental code for cert validation in OS X and Windows

Lukasa:
This is related to kennethreitz/requests#2966.

This change makes it possible to trust the system certificate store explicitly on systems that would like to. The goal is to have this be a cross-platform way to solve the problem, so that Windows and OS X aren't left out in the cold in this kind of world. The goal here is to divide the world into two sets of users:
- Non-Windows and Non-OS X users. These users are assumed to be using OpenSSL, configured appropriately, such that `set_default_verify_paths` will work correctly for them. In that instance, if `trust_system` is `True`, they will automatically use OpenSSL with the system cert stores.
- Windows and OS X users. These users cannot use `set_default_verify_paths` because those platforms have their own, complex, certificate stores that do not and cannot interface properly with OpenSSL. These platforms will therefore need to use certitude to validate their trust stores: essentially, they'll pass the X509 cert chain to certitude which will call into the relevant system functions to do the validation.

This is very much WIP right now, and there are the following limitations:
1. We need to bikeshed the API a bit.
2. This has made `ssl_wrap_socket` a bit crazy: we may want to factor it out a bit to keep the code cleaner.
3. certitude doesn't work on Windows yet.

However, feedback would be appreciated.

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -51,12 +51,20 @@ try:
 except SyntaxError as e:
     raise ImportError(e)
 
+try:
+    import certitude
+except ImportError:
+    pass
+
 import OpenSSL.SSL
+import OpenSSL.crypto
 from pyasn1.codec.der import decoder as der_decoder
 from pyasn1.type import univ, constraint
 from socket import _fileobject, timeout, error as SocketError
 import ssl
 import select
+import platform
+
 
 from .. import connection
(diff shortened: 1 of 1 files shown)


shazow:
+1, looks like a sensible start. More docs would be good though. @sigmavirus24 any concerns?

sigmavirus24:
This looks good to me so far

Issue #805, opened 2016-02-24
Connection pool queue can still be exhausted by gevent timeouts


ml31415:
A similar error was discussed here https://github.com/shazow/urllib3/issues/644 already and some fix applied, but the problem still partly exists. Consider this code:

```python
from gevent import monkey, spawn, sleep, Timeout, joinall
monkey.patch_all()

import sys
from itertools import product
from urllib3.connectionpool import HTTPConnectionPool

def timed_read(cp, i, timeout, release=None, preload_content=True):
    try:
        with Timeout(timeout):
            resp = cp.urlopen('GET', 'http://httpbin.org/redirect/:1',
                              release_conn=release,
                              preload_content=preload)
            print "%s opened" % i
            sys.stdout.flush()
            data = resp.data
    except Timeout:
        print "%s timed out" % i
    else:
        content_len = int(resp.headers['content-length'])
        if len(data) != content_len:
            print "%s incomplete read: %s/%s" % (i, len(data), content_len)
        else:
            print "%s finished" % i
    sys.stdout.flush()


def main(timeout=10, release=None, preload=True, poolsize=2):
    print '-' * 10 + "  timeout=%s preload=%s release=%s " % (timeout, preload, release) + '-' * 10
    cp = HTTPConnectionPool('httpbin.org', block=True, maxsize=poolsize, timeout=None)
    group = []
    for i in range(3):
        group.append(spawn(timed_read, cp, i, timeout, release=release, preload_content=preload))
    joinall(group)

    current_size = cp.pool.qsize()
    if current_size != poolsize:
        print "poolsize decreased: %s" % current_size


for timeout, preload in product([10, 4, 0.1], [True, False]):
    main(timeout=timeout, preload=preload)
```

Setting a timeout that way may seem stupid, as it could be set on the connection itself, but it's just an example for having a timeout running on a longer running task including several calls into the connection pool. What happens is, that the connections are not properly returned back into the queue, and the queue finally runs empty, and any further call to it blocks forever:

<pre>
----------  timeout=10 preload=True release=None ----------
1 opened
1 finished
0 opened
0 finished
2 opened
2 finished
----------  timeout=10 preload=False release=None ----------
0 opened
0 finished
1 opened
1 finished
2 opened
2 finished
----------  timeout=4 preload=True release=None ----------
1 opened
1 finished
0 timed out
2 timed out
----------  timeout=4 preload=False release=None ----------
1 opened
1 finished
0 opened
0 finished
2 timed out
[...]

PR #806, opened 2016-02-25
Pull cleanup code into finally.

Lukasa:
As discussed in #805.


sigmavirus24:
:+1: looks good to me.

shazow:
:custard:

Does this fix #805?

Lukasa:
It should do, yes.

PR #807, opened 2016-02-29
Yet another pass at connection cleanup.

### LINKED ISSUE #805: Connection pool queue can still be exhausted by gevent timeouts

Lukasa:
Resolves #805 (again).

This patch ensures that we correctly cleanup connections that fail inside `HTTPConnectionPool.urlopen`, not just ones that fail after opening. This should ensure that weird gevent timeouts during connection setup also correctly eliminate the connection, rather than leaking it.

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -283,6 +283,28 @@ class TestConnectionPool(unittest.TestCase):
             conn = pool._get_conn()
             self.assertEqual(conn.cert_reqs, 'CERT_REQUIRED')
 
+    def test_cleanup_on_extreme_connection_error(self):
+        class RealBad(BaseException):
+            pass
+
+        def kaboom(*args, **kwargs):
+            raise RealBad()
+
+        c = connection_from_url('http://localhost:80')
+        c._make_request = kaboom
+
+        initial_pool_size = c.pool.qsize()
+
+        try:
+            # We need to release_conn this way or we'd put it away regardless.
+            c.urlopen('GET', '/', release_conn=False)
+        except RealBad:
+            pass
+
+        new_pool_size = c.pool.qsize()
+        self.assertEqual(initial_pool_size, new_pool_size)
+
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -284,6 +284,11 @@ class TestConnectionPool(unittest.TestCase):
             self.assertEqual(conn.cert_reqs, 'CERT_REQUIRED')
 
     def test_cleanup_on_extreme_connection_error(self):
+        """
+        This test validates that we clean up properly even on exceptions that
+        we'd not otherwise catch, i.e. those that inherit from BaseException
+        like KeyboardInterrupt or gevent.Timeout. See #805 for more details.
+        """
         class RealBad(BaseException):
             pass
 
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -630,6 +630,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             if not clean_exit:
                 # We hit some kind of exception, handled or otherwise. We need
                 # to throw the connection away unless explicitly told not to.
+                # Close the connection, set the variable to None, and make sure
+                # we put the None back in the pool to avoid leaking it.
                 conn = conn and conn.close()
                 release_conn = True
 
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -556,6 +556,10 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # complains about UnboundLocalError.
         err = None
 
+        # Keep track of whether we cleanly exited the except block. This
+        # ensures we do proper cleanup in finally.
+        clean_exit = False
+
         try:
             # Request a connection from the queue.
             timeout_obj = self._get_timeout(timeout)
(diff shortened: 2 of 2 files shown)


test/test_connectionpool.py
```diff
@@ -283,6 +283,28 @@ def test_ca_certs_default_cert_required(self):
             conn = pool._get_conn()
             self.assertEqual(conn.cert_reqs, 'CERT_REQUIRED')
 
+    def test_cleanup_on_extreme_connection_error(self):
+        class RealBad(BaseException):
```
sigmavirus24: So, perhaps it's just me, but I'd love to have a comment or docstring for this that describes why we have an exception class inheriting from BaseException and why we're testing that here.

(In reality, I'd love for all tests to have docstrings, but this one especially.)

urllib3/connectionpool.py
```diff
@@ -628,6 +627,12 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
             err = e
 
         finally:
+            if not clean_exit:
+                # We hit some kind of exception, handled or otherwise. We need
+                # to throw the connection away unless explicitly told not to.
+                conn = conn and conn.close()
```
sigmavirus24: I take it that `conn.close()` will return `None`, right? In other words, the result of this will be a state where the locals has `conn = None` and then `self._put_conn` will add `None` back to the pool?

Then we'll go onto the retry logic around L646?

sigmavirus24:
tl;dr this looks fine except I'd love to get the @ml31415's feedback on this + better commenting around this code so we know what it is and why it is later on when we've all forgotten the joyride that #805 has become. ;)

urllib3/connectionpool.py
Lukasa: Exactly right. This pattern was present in the other blocks of code that I consolidated here.

Lukasa:
@sigmavirus24 I think I've added some clarifying material, let me know how you feel about it.

shazow:
This is one of those refactors where I'm really glad we have great coverage for that code. :) 

I suspect this will be an intermediary step towards a further refactor in the future but +1 from me!

sigmavirus24:
:+1: from me as well. Thanks for the comments @Lukasa. I think future us (or the poor souls who replace us) will thank us.

Lukasa:
Yeah, code coverage here was huge, I felt much more comfortable with this change than I would have without it.

I'm going to let someone else merge. =)

shazow:
:custard:

PR #808, opened 2016-02-29
README: Add maintainers list

shazow:
@Lukasa @sigmavirus24 any objections?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,7 +128,7 @@ Contributing
 ============
 
 Thank you for giving back to urllib3. Please meet our jolly team
-of jolly code-sherpas:
+of code-sherpas:
 
 Maintainers
 -----------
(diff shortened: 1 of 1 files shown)


Lukasa:
The irony of the build failing here is strong. ;)

The only objection I have is ordering: I know you ordered alphabetically, but FWIW it feels a bit weird to not have you at the top of that list. Not a strong objection.

sigmavirus24:
tl;dr; what @Lukasa said

> The irony of the build failing here is strong. ;)

Seconded.

> The only objection I have is ordering: I know you ordered alphabetically, but FWIW it feels a bit weird to not have you at the top of that list. Not a strong objection.

Also seconded.

shazow:
lol actually didn't notice it was alphabetic, I'd prefer you both get more recognition than you do. I'm also considering moving the notice further up in the README if you can find a place that makes sense for it.

sigmavirus24:
I don't really see somewhere else that makes sense tbh.

shazow:
We can change the ordering if it keeps bugging you, let me know. :)

PR #813, opened 2016-03-14
New PyOpenSSL requires bytes for cipher list.

Lukasa:
PyOpenSSL is planning to add urllib3 as a tested project to make sure that changes there don't break here (that's really convenient and very kind of @hynek). I should note that we only managed to get here because we added PyOpenSSL to our own testing matrix in #795, which is a change I'm feeling increasingly good about!

The current master (and therefore next release) of PyOpenSSL currently works fine with urllib3 *except* that it emits a deprecation warning when we pass a `str` as the default cipher list. This change fixes that warning, brings our PyOpenSSL tests back to green against the current PyOpenSSL master, and allows the merging of pyca/pyopenssl#446.


shazow:
Seems harmless, thankies! :)

PR #815, opened 2016-03-15
Reordering of parse_url to handle separator characters inside auth information

### LINKED ISSUE #814: parse_url fails when given credentials in the URL with '/', '#', or '?'
Technically, this behavior is compliant with RFC-3986 in that the "authority" section should end at the next instance of one of those characters. However, RFC-2617, which specifies the HTTP Basic Auth scheme, does not specifically exclude those characters as valid, which means that there are valid RFC-2617 usernames and passwords which cannot currently be transmitted inside the URL for handling.

This behavior has received an issue filing on pip in the context of an HTTP proxy.

The attached pull request changes behavior such that any authorization information (separated from the rest of the URL by the reserved character '@') is pulled from the URL prior to scanning for the end of the authority section (separated from the rest of the URL by the aforementioned reserved characters). The PR also includes regression tests to ensure that this behavior change stays in place.

Sources:
https://github.com/pypa/pip/issues/3267
http://tools.ietf.org/html/rfc3986#section-3.2
https://tools.ietf.org/html/rfc2617#section-2

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/31911715-parse_url-fails-when-given-credentials-in-the-url-with-or?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

haikuginger:
Fixes #814.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -123,7 +123,9 @@ class TestUtil(unittest.TestCase):
         'http://foo:bar@localhost/': Url('http', auth='foo:bar', host='localhost', path='/'),
         'http://foo@localhost/': Url('http', auth='foo', host='localhost', path='/'),
         'http://foo:bar@baz@localhost/': Url('http', auth='foo:bar@baz', host='localhost', path='/'),
-        'http://@': Url('http', host=None, auth='')
+        'http://@': Url('http', host=None, auth=''),
+        'http://foo:bar#@localhost': Url('http', auth='foo:bar#', host='localhost'),
+        'http://foo:bar/baz@localhost': Url('http', auth='foo:bar/baz', host='localhost')
     }
 
     non_round_tripping_parse_url_host_map = {
--- a/urllib3/util/url.py
+++ b/urllib3/util/url.py
@@ -158,6 +158,11 @@ def parse_url(url):
     if '://' in url:
         scheme, url = url.split('://', 1)
 
+    # Auth
+    if '@' in url:
+        # Last '@' denotes end of auth part
+        auth, url = url.rsplit('@', 1)
+
     # Find the earliest Authority Terminator
     # (http://tools.ietf.org/html/rfc3986#section-3.2)
     url, path_, delim = split_first(url, ['/', '?', '#'])
@@ -165,11 +170,6 @@ def parse_url(url):
     if delim:
         # Reassemble the path
         path = delim + path_
-
-    # Auth
-    if '@' in url:
-        # Last '@' denotes end of auth part
-        auth, url = url.rsplit('@', 1)
 
     # IPv6
     if url and url[0] == '[':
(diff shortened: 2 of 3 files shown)


Lukasa:
I'm happy with this change (and the associated reasoning in #814), so I'm :+1:.

sigmavirus24:
I'm :-1: and will discuss it on #814.

urllib3/util/url.py
```diff
@@ -158,6 +158,11 @@ def parse_url(url):
     if '://' in url:
         scheme, url = url.split('://', 1)
 
+    # Auth
+    if '@' in url:
+        # Last '@' denotes end of auth part
```
sigmavirus24: The "auth" part is not one of the actual components of a URI, however. The components are:

- scheme
- authority (not auth by way urllib3 refers to auth)
- path
- query
- fragment

An authority can be parsed into credentials, hostname, and port, but the parsing of those three sub-components should be done after parsing out the authority.

test/test_util.py
```diff
@@ -123,7 +123,9 @@ def test_invalid_host(self):
         'http://foo:bar@localhost/': Url('http', auth='foo:bar', host='localhost', path='/'),
         'http://foo@localhost/': Url('http', auth='foo', host='localhost', path='/'),
         'http://foo:bar@baz@localhost/': Url('http', auth='foo:bar@baz', host='localhost', path='/'),
-        'http://@': Url('http', host=None, auth='')
+        'http://@': Url('http', host=None, auth=''),
+        'http://foo:bar#@localhost': Url('http', auth='foo:bar#', host='localhost'),
+        'http://foo:bar/baz@localhost': Url('http', auth='foo:bar/baz', host='localhost')
```
sigmavirus24: Please add a test for the following:

```
http://localhost/@username
http://localhost:5000/?foo=@bar
http://localhost/#biz@baz
```

None of those should have auth parameters (in case that wasn't clear).

haikuginger:
I think I see your point. Essentially, if we make fixes for and provide test cases for one non-RFC-compliant edge case, where does it stop?

I do think there's an argument to be made: the separator characters are at least RFC-valid credentials, whereas there's no RFC-valid reason to have a non-urlencoded "@" elsewhere in the URL, but I'll defer to you guys on which non-standard URLs you want to fail.

Lukasa:
> whereas there's no RFC-valid reason to have a non-urlencoded "@" elsewhere in the URL

That's not entirely true. RFC 3986 allows a non-urlencoded "@" in a path segment, for example, and RFC 7230 allows it as well because it imports the same definition.

haikuginger:
Ah, of course you're right. I saw it in the "reserved" table and didn't see it was specifically allowed lower down.

PR #818, opened 2016-03-16
Handling errors around Unicode encoding differently

### LINKED ISSUE #793: UnicodeDecodeError in format_header_params
This issue was discussed here: https://github.com/kennethreitz/requests/issues/2639 and it seemed like the consensus was that this should be fixed in urllib3. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30203414-unicodedecodeerror-in-format_header_params?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

haikuginger:
Fixes #793; allows unicode characters to come in in the form of a Python 2 bytes/string object without uncaught exceptions.

### DIFF
--- a/test/test_fields.py
+++ b/test/test_fields.py
@@ -2,6 +2,7 @@ import unittest
 
 from urllib3.fields import guess_content_type, RequestField
 from urllib3.packages.six import u, PY3
+from .. import onlyPy2
 
 
 class TestRequestField(unittest.TestCase):
@@ -48,8 +49,8 @@ class TestRequestField(unittest.TestCase):
         param = field._render_part('filename', u('n\u00e4me'))
         self.assertEqual(param, "filename*=utf-8''n%C3%A4me")
 
+    @onlyPy2
     def test_render_unicode_bytes_py2(self):
-        if not PY3:
-            field = RequestField('somename', 'data')
-            param = field._render_part('filename', 'n\xc3\xa4me')
-            self.assertEqual(param, "filename*=utf-8''n%C3%A4me")
+        field = RequestField('somename', 'data')
+        param = field._render_part('filename', 'n\xc3\xa4me')
+        self.assertEqual(param, "filename*=utf-8''n%C3%A4me")
--- a/test/test_fields.py
+++ b/test/test_fields.py
@@ -2,7 +2,7 @@ import unittest
 
 from urllib3.fields import guess_content_type, RequestField
 from urllib3.packages.six import u, PY3
-from .. import onlyPy2
+from . import onlyPy2
 
 
 class TestRequestField(unittest.TestCase):
--- a/test/test_fields.py
+++ b/test/test_fields.py
@@ -1,7 +1,7 @@
 import unittest
 
 from urllib3.fields import guess_content_type, RequestField
-from urllib3.packages.six import u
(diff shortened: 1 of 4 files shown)


test/test_fields.py
```diff
@@ -47,3 +47,9 @@ def test_render_part(self):
         field = RequestField('somename', 'data')
         param = field._render_part('filename', u('n\u00e4me'))
         self.assertEqual(param, "filename*=utf-8''n%C3%A4me")
+
+    def test_render_unicode_bytes_py2(self):
+        if not PY3:
```
shazow: We have [test decorators](https://github.com/shazow/urllib3/blob/master/test/__init__.py#L37-L79) for this.

shazow:
Seems safe to me, @lukasa?

urllib3/fields.py
```diff
@@ -36,11 +36,11 @@ def format_header_param(name, value):
         result = '%s="%s"' % (name, value)
         try:
             result.encode('ascii')
-        except UnicodeEncodeError:
+        except (UnicodeEncodeError, UnicodeDecodeError):
             pass
         else:
             return result
-    if not six.PY3:  # Python 2:
+    if not six.PY3 and isinstance(value, unicode):  # Python 2:
```
shazow: Maybe we should use `six.text_type` instead? (Also flake8 fails on unicode for some reason)

haikuginger: I took a look at six; the issue there is that six.text_type evaluates as true for Python 3 strings as well as Python 2 unicode, and encode_rfc2231 doesn't take Python 3 bytes.

shazow: Yea, we'd still need the `not six.PY3` pre-check. Wonder why flake8 is failing though.

Lukasa: Python 3 `str` type is the same as Python 2 `unicode` type. That's also why Flake8 is complaining BTW: there is no `unicode` type on Python 3.

shazow: Ah that makes sense. What's the safest thing to do then?

Lukasa:
After fixing the nit around `unicode` I'm fine with this, though I'd like a bit more testing around it to validate that it works properly.

urllib3/fields.py
haikuginger: I took a look at the logs; looks like it's failing because unicode isn't a valid type in Py3. We could definitely get around THAT by switching to `six.text_type`, but yeah, we'd still need `not six.PY3`.

Lukasa: Yup, switching to `six.text_type` is all we need here.

haikuginger:
@Lukasa: what sort of additional testing would you like? There's existing testing for both Python 2 `unicode` types and Python 3 `str` types; this adds support and testing for Unicode bytes in a Python 2 `str`.

Lukasa:
Oh I'm sorry, I didn't properly process that there is a test added in the diff. In that case I'm happy when @shazow is happy!

shazow:
:custard:!

PR #820, opened 2016-03-17
Show certificate when there is a hostname mismatch

msabramo:
I am getting a hostname mismatch error for one of three servers in a pool so I am suspecting that one server has an invalid certificate.

    requests.exceptions.SSLError: hostname '10.10.194.37' doesn't match u'vault.service.trouper'

by itself is not much info to go on. It would be nice to see the server certificate that was presented.

I found that it was surprisingly difficult to see the server certificate, so I decided to see if urllib3 could show the certificate.

With this, urllib3 logs an error with the certificate just before raising the exception:

```
ERROR:requests.packages.urllib3.connection:Certificate did not match expected hostname: 10.10.194.38.
Certificate: {'subjectAltName': (('IP Address', '127.0.0.1'), ('IP Address', '10.10.194.36'), ('IP Address', '10.10.194.37'), ('IP Address', '10.10.194.38')),
'notBefore': u'Mar 17 01:18:27 2016 GMT', 'serialNumber': u'9BFDB58B08E7DCCA', 'notAfter': 'Mar 15 01:18:27 2026 GMT', 'version': 3L,
'subject': ((('countryName', u'US'),), (('stateOrProvinceName', u'California'),), (('localityName', u'Palo Alto'),), (('organizationName', u'DevOps'),), (('commonName', u'vault
```

which allows me to check whether the server is presenting the correct certificate.

I also add the certificate to the exception that is raised so that client code can inspect it.

### DIFF
--- a/test/test_connection.py
+++ b/test/test_connection.py
@@ -1,0 +1,48 @@
+import unittest
+
+import mock
+
+from urllib3.connection import (
+    CertificateError,
+    VerifiedHTTPSConnection,
+    _match_hostname,
+)
+
+
+class TestConnection(unittest.TestCase):
+    """
+    Tests in this suite should not make any network requests or connections.
+    """
+    def test_match_hostname_no_cert(self):
+        cert = None
+        asserted_hostname = 'foo'
+        self.assertRaises(ValueError, _match_hostname, cert, asserted_hostname)
+
+    def test_match_hostname_empty_cert(self):
+        cert = {}
+        asserted_hostname = 'foo'
+        self.assertRaises(ValueError, _match_hostname, cert, asserted_hostname)
+
+    def test_match_hostname_match(self):
+        cert = {'subjectAltName': [('DNS', 'foo')]}
+        asserted_hostname = 'foo'
+        _match_hostname(cert, asserted_hostname)
+
+    def test_match_hostname_mismatch(self):
+        cert = {'subjectAltName': [('DNS', 'foo')]}
+        asserted_hostname = 'bar'
+        try:
+            with mock.patch('urllib3.connection.log.error') as mock_log:
+                _match_hostname(cert, asserted_hostname)
+        except CertificateError as e:
+            self.assertEqual(str(e), "hostname 'bar' doesn't match 'foo'")
+            mock_log.assert_called_once_with(
+                'Certificate did not match expected hostname: %s. '
+                'Certificate: %s',
+                'bar', {'subjectAltName': [('DNS', 'foo')]}
+            )
+            self.assertEqual(e._peer_cert, cert)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 import datetime
+import logging
 import os
 import sys
 import socket
@@ -38,7 +39,7 @@ from .exceptions import (
     SubjectAltNameWarning,
     SystemTimeWarning,
 )
-from .packages.ssl_match_hostname import match_hostname
+from .packages.ssl_match_hostname import match_hostname, CertificateError
 
 from .util.ssl_ import (
     resolve_cert_reqs,
@@ -51,6 +52,8 @@ from .util.ssl_ import (
 from .util import connection
 
 from ._collections import HTTPHeaderDict
+
+log = logging.getLogger(__name__)
 
 port_by_scheme = {
     'http': 80,
@@ -299,7 +302,18 @@ class VerifiedHTTPSConnection(HTTPSConnection):
                     'for details.)'.format(hostname)),
                     SubjectAltNameWarning
                 )
-            match_hostname(cert, self.assert_hostname or hostname)
+            asserted_hostname = self.assert_hostname or hostname
+            try:
+                match_hostname(cert, asserted_hostname)
+            except CertificateError as e:
+                log.error(
+                    'Certificate did not match expected hostname: %s. '
+                    'Certificate: %s', asserted_hostname, cert
+                )
+                # Add cert to exception and reraise so client code can inspect
+                # the cert when catching the exception, if they want to
+                e.cert = cert
+                raise
 
         self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED or
                             self.assert_fingerprint is not None)
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -307,6 +307,7 @@ class VerifiedHTTPSConnection(HTTPSConnection):
         self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED or
                             self.assert_fingerprint is not None)
 
+
 def _match_hostname(cert, asserted_hostname):
     try:
         match_hostname(cert, asserted_hostname)
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -302,21 +302,23 @@ class VerifiedHTTPSConnection(HTTPSConnection):
                     'for details.)'.format(hostname)),
                     SubjectAltNameWarning
                 )
-            asserted_hostname = self.assert_hostname or hostname
-            try:
-                match_hostname(cert, asserted_hostname)
-            except CertificateError as e:
-                log.error(
-                    'Certificate did not match expected hostname: %s. '
-                    'Certificate: %s', asserted_hostname, cert
-                )
-                # Add cert to exception and reraise so client code can inspect
-                # the cert when catching the exception, if they want to
-                e.cert = cert
-                raise
+            _match_hostname(cert, self.assert_hostname or hostname)
 
         self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED or
                             self.assert_fingerprint is not None)
+
+def _match_hostname(cert, asserted_hostname):
+    try:
+        match_hostname(cert, asserted_hostname)
+    except CertificateError as e:
+        log.error(
+            'Certificate did not match expected hostname: %s. '
+            'Certificate: %s', asserted_hostname, cert
+        )
+        # Add cert to exception and reraise so client code can inspect
+        # the cert when catching the exception, if they want to
+        e._peer_cert = cert
+        raise
 
 
 if ssl:


Lukasa:
Thanks for this!

I have to admit, I'm not sure that I think firing a warning here is a good idea. Firing a warning followed immediately by an exception seems entirely superfluous to me. I think I'd be happier with logging this (at, say, ERROR level) before throwing the exception.

msabramo:
Yeah, I wasn't sure about the warning either, but I decided to just go with something to start a discussion about what would be the best mechanism and that's what I got. Thanks! Updated.

urllib3/connection.py
```diff
@@ -299,7 +302,15 @@ def connect(self):
...
-            match_hostname(cert, self.assert_hostname or hostname)
+            try:
+                match_hostname(cert, self.assert_hostname or hostname)
+            except CertificateError as e:
+                log.error(
+                    'Certificate did not match expected hostname: %s. '
+                    'Certificate: %s', asserted_hostname, cert
+                )
+                e.cert = cert
+                raise e
```
Lukasa: I'm a *little* nervous about the fact that this exception will have a misleading traceback. We might need to do something a bit nasty here to preserve the original traceback, rather than having it be emitted from here. I think [this code](https://stackoverflow.com/questions/8760267/re-raise-python-exception-and-preserve-stack-trace) will do the job, but I admit to not having tried.

Lukasa:
M'kay, I've made a note about a problem with the re-thrown exception. You've also got some test failures that need fixing. =)

Finally, it'd be nice to have a test that demonstrates that the certificate really is hung off the exception, just so we don't accidentally lose this behaviour.

urllib3/connection.py
msabramo: Other alternative is to just not try to put the `cert` in there and just do `raise` instead of `raise e`. Maybe we don't need the cert in the exception?

Lukasa: That's definitely another alternative. I'm a bit sad about that though: it'd be convenient to have it.

You could investigate whether leaving the patch in and changing to a naked `raise` statement still successfully patches the cert onto the exception. That would be the best of both worlds I think.

msabramo:
@Lukasa: I have a question for you if you don't mind, since I know you are expert on all things SSL/TLS and this is the issue that inspired this PR, so it's kind of relevant.

How is one supposed to deal with verifying certificates when doing HTTPS requests to URLs with IP addresses in them?

My certificate lists the IP address in the `subjectAltName` and I thought that would be enough, but it looks like the Python `ssl` library is purposely ignoring IP address entries in `subjectAltName`?

```python
def match_hostname(cert, hostname):
    """Verify that *cert* (in decoded format as returned by
    SSLSocket.getpeercert()) matches the *hostname*.  RFC 2818 and RFC 6125
    rules are followed, but IP addresses are not accepted for *hostname*.

    CertificateError is raised on failure. On success, the function
    returns nothing.
    """
    if not cert:
        raise ValueError("empty or no certificate, match_hostname needs a "
                         "SSL socket or SSL context with either "
                         "CERT_OPTIONAL or CERT_REQUIRED")
    dnsnames = []
    san = cert.get('subjectAltName', ())
    for key, value in san:
        if key == 'DNS':
```

Lukasa:
The [latest copy of the stdlib](https://github.com/python/cpython/blob/master/Lib/ssl.py) actually *does* handle IP addresses. However, it uses the stdlib's `ipaddress` module, which is actually a relatively new addition, so to bring that support back to urllib3 we'd need to backport the ipaddress module too. That's a bit sad, but it's the world we live in. =(

msabramo:
That makes me a sad :panda_face:

Lukasa:
You and me both. I plan to tackle this at some point soon, but I'm currently spread pretty thin and trying to tackle one thing at a time!

msabramo:
Ah, just found https://bugs.python.org/issue23239 - so it looks like the stdlib change you mentioned was in Python 3.5, for future reference for anyone who finds this while searching. And even that change has some issues that no one has had time to work on.

urllib3/connection.py
msabramo: > You could investigate whether leaving the patch in and changing to a naked
> raise statement still successfully patches the cert onto the exception.
> That would be the best of both worlds I think.

Just tested and it works so I updated it. Thanks!

Note that because exceptions get wrapped by other exceptions, it's somewhat of a pain to get at that `cert` attribute, though possible is better than not possible.

```
(Pdb++) e.cert
*** AttributeError: 'SSLError' object has no attribute 'cert'
(Pdb++) e.args[0].cert
*** AttributeError: 'SSLError' object has no attribute 'cert'
(Pdb++) e.args[0].args[0].cert
{'subjectAltName': (('IP Address', '127.0.0.1'), ('IP Address', '10.10.194.36'), ('IP Address', '10.10.194.37'), ('IP Address', '10.10.194.38')), 'notBefore': u'Mar 17 01:18:27 2016 GMT', 'serialNumber': u'9BFDB58B08E7DCCA', 'notAfter': 'Mar 15 01:18:27 2026 GMT', 'version': 3L, 'subject': ((('countryName', u'US'),), (('stateOrProvinceName', u'California'),), (('localityName', u'Palo Alto'),), (('organizationName', u'DevOps'),), (('commonName', u'vault.service.trouper'),)), 'issuer': ((('countryName', u'US'),), (('stateOrProvinceName', u'California'),), (('localityName', u'Palo Alto'),), (('organizationName', u'DevOps'),), (('commonName', u'vault.service.trouper'),))}
```

Lukasa: Yeah, this is an ongoing downside of the wrapped exceptions in requests: trying to *actually use* something that is in one of the lower exceptions is hard. However, my suspicion is that this mostly actually happens in some kind of semi-interactive setting, where people can drill down.

For the automated case, the log is probably sufficient. Having both covers all the bases. =)

urllib3/connection.py
```diff
@@ -299,7 +302,15 @@ def connect(self):
...
-            match_hostname(cert, self.assert_hostname or hostname)
+            try:
+                match_hostname(cert, self.assert_hostname or hostname)
+            except CertificateError as e:
+                log.error(
+                    'Certificate did not match expected hostname: %s. '
+                    'Certificate: %s', asserted_hostname, cert
+                )
+                e.cert = cert
+                raise
```
Lukasa: Heh, so this is great, but now I think I want a little comment here that says "trust me, the re-raised exception will now have the cert on it". Oh, and a test to prove it. =D

msabramo: Yeah, agreed. Once I get the failing tests to pass. :smile:

msabramo: Yay! Tests passing now!

urllib3/connection.py
```diff
@@ -299,7 +302,16 @@ def connect(self):
                     'for details.)'.format(hostname)),
                     SubjectAltNameWarning
                 )
-            match_hostname(cert, self.assert_hostname or hostname)
+            try:
+                asserted_hostname = self.assert_hostname or hostname
```
Lukasa: A minor style nit, but I'd rather see this outside the `try` block. While this *shouldn't* be able to throw an exception that causes a problem, there is an attribute access in here that could do all kinds of wacky stuff. Best to keep `try` block scope small. =)

msabramo: :+1:. Updated that and also added the "trust me" comment that you suggested.

Lukasa:
Cool, I pronounce myself happy. Pinging @shazow for final review and (almost certainly) merge.  Thanks @msabramo! :custard:

msabramo:
I didn't add the test yet though.

> now I think I want a little comment here that says "trust me, the re-raised exception will now have the 
> cert on it". Oh, and a test to prove it.

Lukasa:
Oh yeah, in that case I rescind all my praise and replace it with criticism! ;)

(But seriously, the test would be good thanks.)

msabramo:
Dang it. Why did I say anything? I just created more work for myself. :smile:

sigmavirus24:
I'm also waiting for the test but then I'd be happy with this.

urllib3/connection.py
```diff
@@ -299,7 +302,18 @@ def connect(self):
...
+            try:
+                match_hostname(cert, asserted_hostname)
+            except CertificateError as e:
+                log.error(
+                    'Certificate did not match expected hostname: %s. '
+                    'Certificate: %s', asserted_hostname, cert
+                )
+                # Add cert to exception and reraise so client code can inspect
+                # the cert when catching the exception, if they want to
+                e.cert = cert
```
shazow: Should we be wrapping this in a urllib3-local exception that always has this field to avoid monkeypatching?

Lukasa: What's your monkeypatching concern?

shazow: I always feel uncomfortable relying on properties on classes which aren't there after init. Someday it'll get raised elsewhere and it'll be missing `cert` and a bunch of consumer code will get `AttributeError`s.

sigmavirus24: Hm, maybe we should be centralizing the SSL hostname matching logic elsewhere so we can point people towards the tools to use instead of implementing this kind of logic from a vendored package themselves?

That said, if we are raising this exception to the user (which I'm fairly certain we aren't) we should stop that and replace it with a different exception.

shazow: It gets wrapped in `SSLError` and then raised to the user.

shazow: If we want to keep it monkeypatched, I'd vote to at least underscore-prefix it to discourage people from depending on it. Maybe `e._peer_cert = cert`.

Lukasa: I can get behind that.

msabramo: OK, I can do that.

msabramo:
OK, I just added a few tests in `test/test_connection.py`. The most notable test for this PR is `test_match_hostname_mismatch`.

test/test_connection.py
```diff
@@ -0,0 +1,44 @@
...
+            self.assertEqual(str(e), "hostname 'bar' doesn't match 'foo'")
+            mock_log.assert_called_once_with(
+                'Certificate did not match expected hostname: %s. '
+                'Certificate: %s',
+                'bar', {'subjectAltName': [('DNS', 'foo')]}
+            )
+            self.assertEqual(e._peer_cert, cert)
+
+
+if __name__ == '__main__':
```
sigmavirus24: Is this necessary?

msabramo: What specifically? This test or just the `if __name__ == '__main__':`? If it's the latter, it's because I copied from another file (`test/test_connectionpool.py`) so that I could match existing style.

test/test_connection.py
```diff
@@ -0,0 +1,49 @@
...
+        _match_hostname(cert, asserted_hostname)
+
+    def test_match_hostname_mismatch(self):
+        cert = {'subjectAltName': [('DNS', 'foo')]}
+        asserted_hostname = 'bar'
+        try:
+            with mock.patch('urllib3.connection.log.error') as mock_log:
+                _match_hostname(cert, asserted_hostname)
+        except CertificateError as e:
+            # import pdb; pdb.set_trace()
```
Lukasa: Probably this comment can be removed.

msabramo: Probably? :smile:

msabramo: Done!

urllib3/connection.py
```diff
@@ -299,12 +302,26 @@ def connect(self):
...
                     SubjectAltNameWarning
                 )
-            match_hostname(cert, self.assert_hostname or hostname)
+            _match_hostname(cert, self.assert_hostname or hostname)
 
         self.is_verified = (resolved_cert_reqs == ssl.CERT_REQUIRED or
                             self.assert_fingerprint is not None)
 
 
+def _match_hostname(cert, asserted_hostname):
```
shazow: @sigmavirus24 Is this what you had in mind?

sigmavirus24: I mean this doesn't hurt, but I didn't mean that this *had* to happen. I think we were talking about monkey-patching as if someone was going to reuse this code (which I don't think anyone will).  If we are expecting people to reuse this, this should be a function that is not faux-private living in `urllib3.util._ssl`.

This is fine now but just seems like unnecessary refactoring.

msabramo: My motivation for extracting out the function was to make it more easily testable.

shazow: Fair point!

msabramo: And I made it private because if there's not some deliberate plan to make it public for some good reason, then better not to expose too much and create something that people rely on and that can't easily be changed in good conscience.

shazow:
Looks good to me. Any other objections before I merge?

Lukasa:
Nope. =)

urllib3/connection.py
msabramo: Also this function is very simple right now, but I was anticipating that it could become a bit more complex, if @Lukasa or someone else added the ability to verify certificates for IP addresses by backporting all of the stuff from Python 3.5, as @Lukasa mentioned in https://github.com/shazow/urllib3/pull/820#issuecomment-197956174

shazow:
@msabramo Thank you again for your hard work, we very much appreciate it! :) :custard:

PR #825, opened 2016-03-23
Preserve order of response headers

davidfstr:
Patch for issue: https://github.com/shazow/urllib3/issues/821

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -352,7 +352,9 @@ class TestConnectionPool(unittest.TestCase):
         are in the same order as received from the underlying httplib.
         """
         
-        expected_response_headers = OrderedDict([('X-Header-%d' % i, str(i)) for i in range(16)])
+        # NOTE: Using lowercase response header names since Python 2.x doesn't
+        #       preserve their case, normalizing them all to lowercase.
+        expected_response_headers = OrderedDict([('x-header-%d' % i, str(i)) for i in range(16)])
         
         c = connection_from_url('http://localhost:80')
         
@@ -387,9 +389,12 @@ class TestConnectionPool(unittest.TestCase):
             for (k, v) in headers.items():
                 httplib_message.add_header(k, v)
         else:
-            httplib_message = HTTPMessage(BytesIO(b''))
             for (k, v) in headers.items():
-                httplib_message.addheader(k, v)
+                assert k == k.lower(), \
+                    'Unwise to use anything but lowercase header names ' + \
+                    'since Python 2.x normalizes them to lowercase internally.'
+            header_text = ''.join(['%s: %s\r\n' % (k, v) for (k, v) in headers.items()])
+            httplib_message = HTTPMessage(BytesIO(header_text.encode('utf8')))
         return httplib_message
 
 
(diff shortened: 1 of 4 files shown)


davidfstr:
Apparently still needs some work to function on Python 2.x. Will take another look in a day or two.

test/test_connectionpool.py
```diff
@@ -308,7 +311,80 @@ def kaboom(*args, **kwargs):
...
+            conn = original_new_conn(*args, **kwargs)
+            conn.request = lambda *args, **kwargs: None
+            conn.getresponse = lambda *args, **kwargs: httplib_response
+            return conn
+        c._new_conn = patched_new_conn
+        
+        actual_response = c.urlopen('GET', '/', release_conn=False)
+        self.assertEqual(
+            list(actual_response.headers.items()),
+            list(expected_response_headers.items()))
```
Lukasa: These tests almost certainly require more work than simply writing a socketlevel test like we have elsewhere. You may want to give that a shot.

davidfstr:
The last failing test is the same one that is failing on the `master` branch:

* ERROR: test_delayed_body_read_timeout (test.with_dummyserver.test_socketlevel.TestSocketClosing)

I'll wait until `master` is passing, then I'll rebase & squish this branch to be more readable.

test/test_connectionpool.py
davidfstr: I didn't understand your suggestion. Could you rephrase?

test/test_connectionpool.py
```diff
@@ -308,7 +311,91 @@ def kaboom(*args, **kwargs):
...
+    
+    def test_preserves_request_header_order(self):
+        """
+        Ensure that headers provided to ConnectionPool.urlopen()
+        are sent to the underlying httplib in the same order.
+        """
+        class AbortRequest(BaseException):
+            pass
+        
+        expected_request_headers = OrderedDict([('X-Header-%d' % i, str(i)) for i in range(16)])
```
sigmavirus24: Why are these tests making 16 headers? Why not 5? Why not 20? What is special about 16?

test/test_connectionpool.py
sigmavirus24: @davidfstr you're duplicating a lot of code between these tests when you could instead write a different type of test (a socket-level test) with the dummyserver that would require a lot less work in each test, be easier to read, understand, and maintain.

davidfstr: Sounds good to me. I'll take a look.

test/test_connectionpool.py
davidfstr: Nothing too special. Just a large enough number that it's a low probability that (shuffle(X) == X), where X is a list of headers. Maybe I'll change it to something like 8 just so that it's not a power of 2 (which might imply the number was special). :-)

Lukasa: Eh, I think I'd go the other way: make it vanishingly unlikely that the test passes. 500 entries, say.

shazow: Right now it gets inserted in alphabetical order, no? How about we just random.shuffle(...) it and check that we get it back in the same shuffled order? That should be fine with a small number of entries.

davidfstr:
Rescinding this PR in favor of: https://github.com/shazow/urllib3/pull/827

PR #828, opened 2016-03-28
Make each PoolManager get a copy of pool_classes_by_scheme

jeremycline:
Each PoolManager should have its own copy of ``pool_classes_by_scheme`` to modify as it sees fit without changing other PoolManager dictionaries.

I ran across this while working on issue #751 and I thought it would be a nice way to get acquainted with the pull request review process for this project.


shazow:
Hm never ran into this in practice but don't see any harm in it.

Lukasa:
This seems reasonable to me: *however*, it's entirely possible that people have been relying on this undocumented behaviour. While we don't owe them anything necessarily (we never expressed that this could be done or encouraged anyone to do it), we should almost certainly make sure that the CHANGES.rst entry for this warns people that the behaviour has changed.

sigmavirus24:
Not to dissuade you from accepting this but at least one or two openstack projects rely on this not being a per-instance dictionary

Lukasa:
@sigmavirus24 Is that entirely accurate? That is, those projects rely on being able to update the dictionary at module scope and have that affect already-existing pool managers?

jeremycline:
I've updated the CHANGES.rst with a note about this change. 

I can definitely see how projects could be relying on this particular behaviour, but it struck me as very unexpected. If this doesn't get accepted, I'd like to change the inline documentation as it leads one to believe each PoolManager has its own copy.

sigmavirus24:
I'm rather certain that either glance or glance client update that to append custom schemes and possibly pool classes.
-- 

Lukasa:
@sigmavirus24 I get the idea that they might do that as some form of start-of-day processing, but it seems totally bonkers to dynamically add new schemes, right?

sigmavirus24:
On March 28, 2016 8:02:34 AM CDT, Cory Benfield <notifications@github.com> wrote:
>@sigmavirus24 I get the idea that they might do that as some form of
>start-of-day processing, but it seems totally bonkers to dynamically
>add new schemes, right?
>
>---
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/shazow/urllib3/pull/828#issuecomment-202382523

It's based off the available schemes for a deployed instance of glance iirc. It happens once in a process but is somewhat dynamic
-- 

shazow:
FWIW, my intended use of it when I wrote the cost was more along the lines of:

```python
my_scheme_mapping = {...}
pool = PoolManager()
pool.pool_classes_by_scheme = my_scheme_mapping
```

But I can totally understand why people would be tempted to mutate the default. Maybe we should have used somekind of frozendicts for these kinds of things.

shazow:
Soooooo what's the consensus on this?

haikuginger:
I think this needs more discussion, so I'm -1 for the moment, given the misgivings that @sigmavirus24 expressed.

What if we made `pool_classes_by_scheme` a class attribute? That way, it can either be overridden on a global basis (by doing `PoolManager.pool_classes_by_scheme = mapping`), or on a per-instance basis (by doing `pool.pool_classes_by_scheme = mapping`). glance would still need to change its code, but overall behavior would be preserved.

jeremycline:
On the one hand, it's currently not consistent with how the pool keys are stored (and that's my own fault), but on the other hand it could break existing projects. Given the very minimal potential gain for new library users versus the real chance of breaking things, I suppose I am also -1 for the moment.

jeremycline:
I've had some coffee and thought about this a bit more. If this PR doesn't make it in, would it be reasonable to change the ``key_fn_by_scheme`` behaviour[0] to match ``pool_classes_by_scheme``? 1.16 hasn't been released yet so we still have time to make them behave the same way.

[0] https://github.com/shazow/urllib3/blob/9a3d6f6c99372f21bc80b0f05e7979afbe3358b5/urllib3/poolmanager.py#L125-L126

shazow:
That's a fair request, it does make sense for them to be consistent.

Also maybe we should just go all-out with callables and avoid exposing raw dict registries. Maybe a v2.0 wishlist thing?

sigmavirus24:
> Also maybe we should just go all-out with callables and avoid exposing raw dict registries. Maybe a v2.0 wishlist thing?

:+1:

haikuginger:
> Also maybe we should just go all-out with callables and avoid exposing raw dict registries. Maybe a v2.0 wishlist thing?

Yes.

jeremycline:
I've opened https://github.com/shazow/urllib3/pull/876 to make everything consistent

jeremycline:
I've rebased and resolved the merge conflicts for this branch.

PR #830, opened 2016-03-28
Create connection pool keys by scheme

jeremycline:
This commit modifies how the PoolManager stores and retrieves ``ConnectionPool`` objects for later use. It adds a new instance variable to the class, ``pool_keys_by_scheme``, which allows a user to define a list of fields from ``connection_pool_kw`` to consider (by scheme) when storing and retrieving pools.

This is my first stab at addressing the feedback from PR #751. It may or (much more likely) may not be what everyone had in mind, so I thought I would make a pull request before spending too much time polishing. Thanks for taking a look!

### DIFF
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -53,12 +53,13 @@ the ``key_fn_by_scheme`` instance variable on :class:`.PoolManager`.
 
 .. doctest ::
 
+    >>> import functools
     >>> from collections import namedtuple
     >>> from urllib3.poolmanager import PoolManager, HTTPPoolKey
     >>> from urllib3.poolmanager import default_key_normalizer as normalizer
     >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('source_address',))
     >>> manager = PoolManager(10, source_address='127.0.0.1')
-    >>> manager.key_fn_by_scheme['http'] = lambda context: normalizer(context, CustomKey)
+    >>> manager.key_fn_by_scheme['http'] = functools.partial(normalizer, CustomKey)
     >>> manager.connection_from_url('http://example.com')
     >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
     >>> manager.connection_from_url('http://example.com')
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -28,6 +28,41 @@ so you don't have to.
     >>> conn.num_requests
     3
 
+A :class:`.PoolManager` will create a new :doc:`ConnectionPool <pools>`
+when no :doc:`ConnectionPools <pools>` exist with a matching pool key.
+The pool key is derived using the requested URL and the current values
+of the ``connection_pool_kw`` instance variable on :class:`.PoolManager`.
+
+The keys in ``connection_pool_kw`` used when deriving the key are
+configurable. For example, by default the ``source_address`` key is not
+considered.
+
+.. doctest ::
+
+    >>> from urllib3.poolmanager import PoolManager
+    >>> manager = PoolManager(10, source_address='127.0.0.1')
+    >>> manager.connection_from_url('http://example.com')
+    >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
+    >>> manager.connection_from_url('http://example.com')
+    >>> len(manager.pools)
+    1
+
+To make the pool manager create new pools when the value of
+``source_address`` changes, you can define a custom pool key.
+
+.. doctest ::
+
+    >>> from collections import namedtuple
+    >>> from urllib3.poolmanager import PoolManager, HTTPPoolKey
+    >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('source_address',))
+    >>> manager = PoolManager(10, source_address='127.0.0.1')
+    >>> manager.pool_keys_by_scheme['http'] = CustomKey
+    >>> manager.connection_from_url('http://example.com')
+    >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
+    >>> manager.connection_from_url('http://example.com')
+    >>> len(manager.pools)
+    2
+
 The API of a :class:`.PoolManager` object is similar to that of a
 :doc:`ConnectionPool <pools>`, so they can be passed around interchangeably.
 
@@ -58,6 +93,12 @@ API
 ---
 
     .. autoclass:: PoolManager
+       :inherited-members:
+    .. autoclass:: BasePoolKey
+       :inherited-members:
+    .. autoclass:: HTTPPoolKey
+       :inherited-members:
+    .. autoclass:: HTTPSPoolKey
        :inherited-members:
 
 ProxyManager
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -48,15 +48,17 @@ considered.
     1
 
 To make the pool manager create new pools when the value of
-``source_address`` changes, you can define a custom pool key.
+``source_address`` changes, you can define a custom pool key and alter
+the ``pool_key_funcs_by_scheme`` instance variable on :class:`.PoolManager`.
 
 .. doctest ::
 
     >>> from collections import namedtuple
     >>> from urllib3.poolmanager import PoolManager, HTTPPoolKey
+    >>> from urllib3.poolmanager import default_key_normalizer as normalizer
     >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('source_address',))
     >>> manager = PoolManager(10, source_address='127.0.0.1')
-    >>> manager.pool_keys_by_scheme['http'] = CustomKey
+    >>> manager.pool_key_funcs_by_scheme['http'] = lambda context: normalizer(context, CustomKey)
     >>> manager.connection_from_url('http://example.com')
     >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
     >>> manager.connection_from_url('http://example.com')
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -49,7 +49,7 @@ considered.
 
 To make the pool manager create new pools when the value of
 ``source_address`` changes, you can define a custom pool key and alter
-the ``pool_key_funcs_by_scheme`` instance variable on :class:`.PoolManager`.
+the ``key_fn_by_scheme`` instance variable on :class:`.PoolManager`.
 
 .. doctest ::
 
@@ -58,7 +58,7 @@ the ``pool_key_funcs_by_scheme`` instance variable on :class:`.PoolManager`.
     >>> from urllib3.poolmanager import default_key_normalizer as normalizer
     >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('source_address',))
     >>> manager = PoolManager(10, source_address='127.0.0.1')
-    >>> manager.pool_key_funcs_by_scheme['http'] = lambda context: normalizer(context, CustomKey)
+    >>> manager.key_fn_by_scheme['http'] = lambda context: normalizer(context, CustomKey)
     >>> manager.connection_from_url('http://example.com')
     >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
     >>> manager.connection_from_url('http://example.com')
--- a/docs/managers.rst
+++ b/docs/managers.rst
@@ -34,21 +34,21 @@ The pool key is derived using the requested URL and the current values
 of the ``connection_pool_kw`` instance variable on :class:`.PoolManager`.
 
 The keys in ``connection_pool_kw`` used when deriving the key are
-configurable. For example, by default the ``source_address`` key is not
+configurable. For example, by default the ``my_field`` key is not
 considered.
 
 .. doctest ::
 
     >>> from urllib3.poolmanager import PoolManager
-    >>> manager = PoolManager(10, source_address='127.0.0.1')
+    >>> manager = PoolManager(10, my_field='wheat')
     >>> manager.connection_from_url('http://example.com')
-    >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
+    >>> manager.connection_pool_kw['my_field'] = 'barley'
     >>> manager.connection_from_url('http://example.com')
     >>> len(manager.pools)
     1
 
 To make the pool manager create new pools when the value of
-``source_address`` changes, you can define a custom pool key and alter
+``my_field`` changes, you can define a custom pool key and alter
 the ``key_fn_by_scheme`` instance variable on :class:`.PoolManager`.
 
 .. doctest ::
@@ -57,11 +57,11 @@ the ``key_fn_by_scheme`` instance variable on :class:`.PoolManager`.
     >>> from collections import namedtuple
     >>> from urllib3.poolmanager import PoolManager, HTTPPoolKey
     >>> from urllib3.poolmanager import default_key_normalizer as normalizer
-    >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('source_address',))
-    >>> manager = PoolManager(10, source_address='127.0.0.1')
+    >>> CustomKey = namedtuple('CustomKey', HTTPPoolKey._fields + ('my_field',))
+    >>> manager = PoolManager(10, my_field='wheat')
     >>> manager.key_fn_by_scheme['http'] = functools.partial(normalizer, CustomKey)
     >>> manager.connection_from_url('http://example.com')
-    >>> manager.connection_pool_kw['source_address'] = '127.0.0.2'
+    >>> manager.connection_pool_kw['my_field'] = 'barley'
     >>> manager.connection_from_url('http://example.com')
     >>> len(manager.pools)
     2
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,6 +1,6 @@
 import unittest
 
-from urllib3.poolmanager import PoolManager
+from urllib3.poolmanager import PoolManager, SSL_KEYWORDS
 from urllib3 import connection_from_url
 from urllib3.exceptions import (
     ClosedPoolError,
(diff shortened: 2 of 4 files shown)


urllib3/poolmanager.py
```diff
@@ -152,7 +202,11 @@ def urlopen(self, method, url, redirect=True, **kw):
         :class:`urllib3.connectionpool.ConnectionPool` can be chosen for it.
         """
         u = parse_url(url)
-        conn = self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
+        request_context = self.connection_pool_kw.copy()
+        request_context['scheme'] = u.scheme or 'http'
+        request_context['port'] = u.port or port_by_scheme.get(u.scheme, 80)
+        request_context['host'] = u.host
```
Lukasa: Why have you changed this to not delegate to `connection_from_host`? It seems like you've duplicated a function that was created to avoid this duplication.

test/test_poolmanager.py
```diff
@@ -87,6 +87,60 @@ def test_contextmanager(self):
...
+
+        self.assertTrue(
+            all(
+                x is not y
+                for i, x in enumerate(conns)
+                for j, y in enumerate(conns)
+                if i != j
+            )
+        )
+
```
Lukasa: So these tests are fine (I should think so, I wrote them!), but there are a few other values in the pool key now which means these tests need updating.

urllib3/poolmanager.py
jeremycline: I am not crazy about how I changed ``connection_from_host`` to not really do what it says because it now takes all the ``connection_pool_kw`` into account. I meant to either change this back to ``connection_from_host`` and accept it being a bit odd or mark ``connection_from_host`` as deprecated (or something) and make something else, but clearly I forgot. Maybe this was a _bit_ too rough to be a PR, but I was worried that I was heading in the wrong direction entirely with the per-PoolManager keys.

In any case, this should change, but I'm not certain about how I want to do it yet.

test/test_poolmanager.py
jeremycline: I definitely plan on adding more tests once I'm convinced I haven't gotten the wrong end of the stick with this keying approach.

urllib3/poolmanager.py
Lukasa: I don't think there's any problem with the change in `connection_from_host`, tbh.

test/test_poolmanager.py
```diff
@@ -87,6 +93,136 @@ def test_contextmanager(self):
...
+    def test_http_pool_key_extra_kwargs(self):
+        """Assert non-HTTPPoolKey fields are ignored when selecting a pool."""
+        p = PoolManager()
+        conn_pool = p.connection_from_url('http://example.com/')
+        p.connection_pool_kw['some_kwarg'] = 'that should be ignored'
+        other_conn_pool = p.connection_from_url('http://example.com/')
+
+        self.assertTrue(conn_pool is other_conn_pool)
```
Lukasa: A good extra test to add, like this one, is to test that the HTTPSPoolKey fields are ignored when selecting HTTP pools.

Lukasa:
First things first: this is a model patch, thank you so much! We've got great tests, good test coverage, and really lucid and clean diff: I'm absolutely delighted with it, and happy to say that it's an order of magnitude better than the original patch I proposed. :custard:

I've left one note in the diff, which is a suggestion for an extra test. I think it'd be a nice thing to have, just to ensure that we've got a test enforcing an invariant that is implicit in the code but not formally tested. It seems like the kind of thing that can be accidentally broken years after being introduced, so I'd like to see that test.

However, with the addition of just that test, I'd be totally happy to call this ready to merge. @shazow should probably review this as well because it refactors some important code, but for my part I'm totally stoked with this patch: it's just wonderful.

Thanks so much @jeremycline!

Lukasa:
Alright, beautiful. Tapping @shazow for the final review-and-merge dance!

shazow:
+1 to everything @lukasa said, this is fantastic. @jeremycline thank you so much for contributing. :)

I'd love an addition to the docs that mentions this, perhaps a short example of how to change the pool to key on source_address? Namely to satisfy https://github.com/shazow/urllib3/pull/376.

Otherwise, this is great and will be happy to merge. Please ping again when you're ready for final review.

jeremycline:
@shazow I added some documentation, let me know if it's not quite what you had in mind!

shazow:
Excellent! @Lukasa Should we include this in v1.15 or hold it over for the next release? (Was gonna push it out today/tomorrow)

Lukasa:
@shazow I think I don't strongly mind, but I'm pretty comfortable with the change, so I guess I'd err very slightly on shipping it in 1.15.

shazow:
Hmmmm for extending the keys, should we require people to extend some base class? For when want to add additional functionality to the key coercer. See: https://github.com/shazow/urllib3/issues/833

Or maybe we should just require some duck interface (or even a callable) and provide a helper for creating that callable from any tuple-like?

jeremycline:
Hmm. Personally I'd prefer to do that in a follow-up PR that fixes the case-sensitivity across the board. The key classes could automatically fix the case by doing something like
```
from collections import namedtuple


class HTTPPoolKey(namedtuple('BasePoolKey', ('scheme', 'host', 'port'))):

    def __new__(cls, scheme, host, port):
        self = super(HTTPPoolKey, cls).__new__(cls, scheme.lower(),
                                               host.lower(), int(port))
        return self
```
but I find that pattern to be somewhat odd when trying to extend a key.

Lukasa:
It's also not clear to me that this would work well as part of the key logic. I'm inclined to allow keys to be case sensitive, but for urllib3 to consistently lowercase schemes and hosts when it builds keys.

shazow:
I mean, keys would generally be used however they're used. It's a matter of where we want to encapsulate the normalization logic. The point being: Somebody out there is gonna be like "well actually, i want _my_ host to be case-sensitive, how do I override?" and if we can be like "oh that's easy, just provide your own key class and enjoy" then that would be nice. :)

Lukasa:
Sure, but I don't think this is a matter of normalising in just one place. Put another way: the problems with case sensitivity are not in only the connection keys, which means putting the normalisation logic in the keys doesn't complete the fix.

shazow:
> Put another way: the problems with case sensitivity are not in only the connection keys, which means putting the normalisation logic in the keys doesn't complete the fix.

Can you expand on that? Where else do we get issues with case sensitivity?

Lukasa:
Here are two examples I found in a quick scan of the poolmanager code:

- we call `port_by_scheme.get(scheme, 80)` with the scheme passed into `connection_for_host`. That call doesn't work properly for improperly cased schemes. This means we need to lowercase the scheme *before* building the key or we may incorrectly choose port 80.
- in urlopen we parse the URL and check its scheme against the lowercase string `"http"`, rather than using the scheme stored on the connection pool.

Both of these can be fixed by re-ordering some code, but I'm inclined to want to call that a separate patch that builds on top of this one rather than extending this.

jeremycline:
Maybe making a change to ``urllib3.util.url.Url`` to lower-case things like scheme and/or making the ``<thing>_by_scheme`` dictionaries case-insensitive? Either way, I think it would probably be better as a separate PR (I could even be talked into making that PR :smile:).

shazow:
If we're firm that this logic shouldn't live in the key-extraction mechanism, then I'm cool with it being a separate PR. On the other hand, if it is going to live there, we're going to do an API-breaking release which is always sadpanda. :(

-1 on entangling the url parser into this though, I'd like to keep those concerns separated.

jeremycline:
This PR is guilty of creating a dictionary by scheme and not accessing it with ``scheme.lower()``, so I may as well fix that here and now.

It sounds like what needs to be decided is whether the default pool keys are case-sensitive or not. It won't fix #833, but I don't think extra enforcement of case-insensitivity would hurt especially since I think I would be a little surprised to have one pool for ``http://somehost:80`` and a second one for ``HTTP://somehost:80``.

Lukasa:
I'd certainly be open to having pool keys be case insensitive. @shazow?

shazow:
I think the question was where does that logic live and do we allow people to override that functionality. Is there any downside to having the normalization logic encapsulated in the pool_key object we accept?

jeremycline:
@shazow Perhaps I am misunderstanding what you're getting at, but looking at PoolManager and ProxyPoolManager, I think it would be very difficult to encapsulate the normalization logic in the pool key. There are ``<thing>_by_scheme`` dictionaries throughout the code and they are all keyed with lower-case schemes. When they're accessed, it seems to always come from the scheme pulled from the parsed URL and often happens long before a pool key gets create. Indeed, the pool key class used is determined by the scheme.

While having the pool keys normalize the scheme and host would be okay, it wouldn't fix a lot of the problems I see and would be a lot more code than just normalizing the values before creating the key. I keep changing my mind, but I think this time I've really convinced myself that the keys can be case-sensitive and the values should just be normalized before the key is created. Users won't be able to configure ``PoolManager`` to being case-sensitive, but as @Lukasa noted the RFC states the scheme[0] and host[1] are case-insensitive. I'm not sure I see the value in adding complexity to allow people to override functionality that they probably shouldn't ever override.

[0] https://tools.ietf.org/html/rfc3986#section-3.1
[1] https://tools.ietf.org/html/rfc3986#section-3.2.2

shazow:
That reminds me, should we apply these changes to the ProxyManager as well?

Sorry I didn't have a chance to really dig in and expand on what I was trying to say. What I meant is that we're already changing the code to select the scheme and whatnot based off the key object rather than the original url, right?

Things like this in this PR:

```diff
-            pool = self._new_pool(scheme, host, port)
+            pool = self._new_pool(pool_key.scheme, pool_key.host, pool_key.port)
```

A super rough sketch of what I would propose is something like this (roughly diff'd from this PR):

```diff
    def connection_from_context(self, request_context):
         """Get a :class:`ConnectionPool` based on the request context."""
-        pool_key_cls = self.pool_keys_by_scheme[request_context['scheme']]
+        pool_key_fn = self.pool_keys_by_scheme[request_context['scheme']]
-        key_kwargs = {}
-        for key_field in pool_key_cls._fields:
-            key_kwargs[key_field] = request_context.get(key_field)
-        pool_key = pool_key_cls(**key_kwargs)
+        pool_key = pool_key_fn(request_context)
  
         return self.connection_from_pool_key(pool_key)
```

Specifically, `pool_key_fn` would be a callable that returns a namedtuple like the ones we defined (those are totally fine and I think we should keep them), rather than it necessarily being a namedtuple itself. If we wanted to backport this PR's existing functionality, the base-case would look like this:

```python
def namedtuple_to_key_fn(tup):
    "Helper for converting namedtuples into key functions, if we need it."
    def fn(request_context):
         key_kwargs = {}
         for key_field in tup._fields:
             key_kwargs[key_field] = request_context.get(key_field)
         return tup(**key_kwargs)
    return fn
```

And that would put us back at where we are now. Buuuut I don't think we'll even need to do that. With this approach, we can inject all kinds of arbitrary normalization logic into the key fn. For example, we could have...

```python
def default_http_key_fn(request_context):
    return HTTPPoolKey(
        scheme=request_context['scheme'].lower(),
        host=request_context['host'].lower(),
        port=request_context['port'],
        ...,
    )
```

I don't want to point too much attention to `namedtuple_to_key_fn`, that's just a shim to show how we'd implement the current state of the PR with the same system. Instead I imagine us pre-defining a handful of `*_key_fn` things with baked-in logic that converts a request_context. Then we'd make sure our code uses the request key for base values.

Again, feel free to say that this is isn't as good and point out increased complexity from this approach (I very much prefer things that reduce complexity in affected areas). I hope this clarifies more what I had in mind, but I'm totally not married to this approach. :)

jeremycline:
@shazow That's a much nicer way to modify the key construction process than mucking about in the namedtuple class definitions (which is what I had in mind). I like it. I'll add a commit later today with that and some tests unless anyone objects.

There are a few other places[0][1][2][3] I've found that break when using URLs like ``HTTP://EXAMPLE.COM`` that won't get fixed, but how about I fix those up in a separate PR? As you can see, they all happen after the URL is parsed into a ``Url`` object which is why I originally suggested involving that class, but if you're not comfortable with that, all these instances can have ``lower()`` called on them.

[0] https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L162
[1] https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L234
[2] https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L246-L252
[3] https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L271-L273

shazow:
Separate PR is totally cool. Thank you! :)

jeremycline:
@shazow Let me know what you think of that. I looked at ``ProxyManager`` again and I think nothing needs to change there for this PR.

urllib3/poolmanager.py
```diff
@@ -42,12 +42,39 @@
...
 )
 
-# The default pool key for each scheme; each PoolManager makes a copy of this
-# dictionary so they can be configured globally here, or individually on the
-# instance.
-pool_keys_by_scheme = {
-    'http': HTTPPoolKey,
-    'https': HTTPSPoolKey,
+
+def default_key_normalizer(request_context, key_class):
```
shazow: Hmmm I imagine the name/behaviour of this function changing, let's make it `_`-prefixed just in case.

urllib3/poolmanager.py
```diff
@@ -42,12 +42,39 @@
...
+    context['scheme'] = context['scheme'].lower()
+    context['host'] = context['host'].lower()
+    return key_class(**context)
+
+
+# A dictionary that maps a scheme to a callable that creates a pool key.
+# This can be used to alter the way pool keys are constructed, if desired.
+# Each PoolManager makes a copy of this dictionary so they can be configured
+# globally here, or individually on the instance.
+pool_key_funcs_by_scheme = {
```
shazow: IMO `key_func_by_scheme` (or `key_fn_by_scheme`?)

Honestly could really use a fresher idea for namespacing these registries, the current naming scheme is getting out of hand.

shazow:
Looks good, but could really use some new naming ideas. Anyone else? (Fine with merging if we can't think of anything better)

jeremycline:
Nothing better came to mind, so I just switched it to ``key_fn_by_scheme``.

shazow:
Looks good to me. Let's get a final approval/feedback from @sigmavirus24 or @Lukasa and merge away. :)

Lukasa:
I'm not going to be in a position to review until Tuesday: either @sigmavirus24 does or we revisit then.

docs/managers.rst
```diff
@@ -28,6 +28,43 @@ so you don't have to.
     >>> conn.num_requests
     3
 
+A :class:`.PoolManager` will create a new :doc:`ConnectionPool <pools>`
```
sigmavirus24: Why the `.` before PoolManager?

sigmavirus24: Hm, I see it outside this diff too. Must be sphinx magic that I haven't seen before.

test/test_poolmanager.py
```diff
@@ -87,6 +96,224 @@ def test_contextmanager(self):
 
         self.assertEqual(len(p.pools), 0)
 
+    def test_http_pool_key_fields(self):
+        """Assert the HTTPPoolKey fields are honored when selecting a pool."""
+        connection_pool_kw = {
+            'timeout': timeout.Timeout(3.14),
```
sigmavirus24: Want some desert to go with that timeout?

urllib3/poolmanager.py
```diff
@@ -23,6 +24,59 @@
...
+    context['host'] = context['host'].lower()
+    return key_class(**context)
+
+
+# A dictionary that maps a scheme to a callable that creates a pool key.
+# This can be used to alter the way pool keys are constructed, if desired.
+# Each PoolManager makes a copy of this dictionary so they can be configured
+# globally here, or individually on the instance.
+key_fn_by_scheme = {
+    'http': lambda context: _default_key_normalizer(context, HTTPPoolKey),
```
sigmavirus24: Lambdas are intentionally slow (because Guido doesn't like them). I wonder if it will affect urllib3/urllib3 users and if functools.partial would be better here.

jeremycline: I'm not attached to the lambdas. I'll just replace these rather than trying to figure out whether or not the slowdown is negligible.

sigmavirus24:
@Lukasa want to give this one more pass?

docs/managers.rst
```diff
@@ -28,6 +28,44 @@ so you don't have to.
...
+when no :doc:`ConnectionPools <pools>` exist with a matching pool key.
+The pool key is derived using the requested URL and the current values
+of the ``connection_pool_kw`` instance variable on :class:`.PoolManager`.
+
+The keys in ``connection_pool_kw`` used when deriving the key are
+configurable. For example, by default the ``source_address`` key is not
+considered.
```
Lukasa: Urh...so, I appreciate this documentation because it raises a really good question: *why not*? `source_address` almost certainly *should* be part of the pool key.

Lukasa:
One question inline.

docs/managers.rst
jeremycline: I suppose I didn't do a lot of creative thinking and took https://github.com/shazow/urllib3/pull/830#issuecomment-205453250 a little too literally. It makes sense to me to be a part of the key, so I'm fine adding it as well.

jeremycline:
@Lukasa What do you think of 4a3078e184a14170845378523e291d0e604d6b52?

Lukasa:
LGTM.

jeremycline:
So is everyone satisfied with the state of things?

shazow:
YAY!!! :heart: @jeremycline

jeremycline:
Great! Thank you all so much for working with me on this PR and for all the work you do maintaining this project. I've really enjoyed contributing and hope to do so again in the not-to-distant future!

shazow:
Us as well, you did an excellent job and hope to see you around much more! :) You're very welcome to hang around and help us review PRs and all that kinda stuff.

Lukasa:
Thanks so much @jeremycline! :sparkles: :custard:

PR #835, opened 2016-04-07
[Second Attempt] Add ssl_context parameter to Connection classes

Lukasa:
This builds upon the work done by @sigmavirus24 in #507.

The differences between this PR and the original are:

- The tests pass! The original PR had failing tests, and was written at a time when we didn't test PyOpenSSL at all so those tests failed *miserably*. This patch brings everything up to speed.
- Adds PyOpenSSL support, including a new wrapper context object for PyOpenSSL that has a similar interface to the builtin context. The goal here is to keep the experience the same for users who are using PyOpenSSL contexts. As a side benefit, this now allows users to use both PyOpenSSL and the standard library SSL module at the same time if they so choose (by passing appropriate context objects).
- Refactors the code to continue to use as much of the original code as possible. @sigmavirus24 had hoisted a lot of code out of `urllib3.util.ssl_.ssl_wrap_socket`, but that method now takes a `context` argument, which means that we really should just re-use it. This patch has been adjusted to do that, ensuring that code coverage remains high and that logic exists in as few places as possible.

If @sigmavirus24 is happy to agree to it, I'd say that this patch supersedes #507.


Lukasa:
Note, please, that this *does* introduce a very subtle behavioural change when using PyOpenSSL (one I wasn't even aware we had). Weirdly, the original logic in PyOpenSSL would allow you to set `cert_reqs='CERT_REQUIRED'` and provide no cert bundle, and would then transparently call `load_default_verify_locations`. That behaviour is out of line with the way we used the stdlib ssl module, so I removed it.

However, we should consider whether we want to do the opposite: have that fallback work in the general case.

haikuginger:
To make sure I'm understanding correctly, it sounds like changing that behavior across the board would automatically get system roots when possible, but require use of the stdlib-accessible certificate stores provided by PEP-476 in non-PyOpenSSL versions, and likely failover to empty stores on lower versions of Python.

If so, I'm in favor of that behavior change, provided there's also a warning thrown for those on configurations that don't support that feature - I believe the cutoffs for the stdlib are 2.7.9 and 3.4.3. For platforms newer than that, we'd call `context.load_default_certs()`.

It might also be worth talking about whether that's worth moving to across the board in the future.

Lukasa:
Bump. =)

test/with_dummyserver/test_https.py
```diff
@@ -166,8 +166,13 @@ def test_unverified_ssl(self):
             self.assertEqual(r.status, 200)
             self.assertTrue(warn.called)
 
-            call, = warn.call_args_list
-            category = call[0][1]
+            calls = warn.call_args_list
+            if sys.version_info >= (2, 7, 9) or util.IS_PYOPENSSL:
```
sigmavirus24: Please, can we have a comment about this?

urllib3/connection.py
```diff
@@ -281,17 +292,21 @@ def connect(self):
...
+            ca_certs=self.ca_certs,
+            ca_cert_dir=self.ca_cert_dir,
+            server_hostname=hostname,
+            ssl_context=context)
 
         if self.assert_fingerprint:
             assert_fingerprint(self.sock.getpeercert(binary_form=True),
                                self.assert_fingerprint)
-        elif resolved_cert_reqs != ssl.CERT_NONE \
+        elif self.ssl_context.verify_mode != ssl.CERT_NONE \
```
sigmavirus24: Can't we just use `context` here too instead of `self.ssl_context`?

sigmavirus24:
A couple comments but otherwise I'm :+1:

urllib3/connection.py
```diff
@@ -204,13 +205,21 @@ class HTTPSConnection(HTTPConnection):
     default_port = port_by_scheme['https']
 
     def __init__(self, host, port=None, key_file=None, cert_file=None,
-                 strict=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT, **kw):
+                 strict=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
+                 ssl_context=None, ssl_version=None, **kw):
```
shazow: Do we need both `ssl_context` and `ssl_version` here?

shazow:
Overall I like it, but as always I'm trying to think of ways to reduce the addition of new keywords. I'd love to get rid of `ssl_version` in the main constructor (would it be sufficient to provide a helper for pre-constructing an `ssl_context`?) but it's not a hard blocker.

urllib3/connection.py
sigmavirus24: I think I did this on the first attempt at this PR. The logic is preserving backwards compatibility because it already accepts `ssl_version`. Most people using `ssl_version` probably will not want to switch to creating a context object.

Would it be better for them to be mutually exclusive?

shazow: Hrm, is it? I see it as a property on `VerifiedHTTPSConnectionPool` but it's not a param on the constructor is it? Plus it's not on any of the other pools, so I don't think it's an issue?

sigmavirus24: It may have been passed via `**kw`. @Lukasa do you mind updating this?

Lukasa: ACK.

urllib3/connection.py
Lukasa: ACK.

Lukasa:
@shazow There's no need to put `ssl_version` in the constructor, actually. It's only there because we used to use defaults in `ssl_wrap_socket` and now we can't, but that's ok: we can just manually default them ourselves. Easy peasy! See the new diff.

test/with_dummyserver/test_https.py
sigmavirus24: Ping @Lukasa

Lukasa: Arg, missed this. Yeah, sure.

Lukasa: ACK.

sigmavirus24:
I'm very much in the :shipit: camp.

shazow:
I like.

Any desire to include a recipe for providing your own ssl_context in our docs? :) (Separate PR is fine too.)

Otherwise good to merge.

sigmavirus24:
That's a good point @shazow

Lukasa:
So I updated the docs, but this doesn't quite work right yet. In particular, if you pass an `SSLContext` you still need to pass `cert_reqs`, which doesn't seem right.

shazow:
docs-driven-development!

Lukasa:
So, million dollar question: do we believe `cert_reqs` will change on a `VerifiedHTTPSConnection` after the first request goes through? I feel like the answer *has* to be no: if it is, then we can skip the problematic LoC.

shazow:
Anything that mutates the state of a socket in a pool needs to remain unchanged after pool initialization. Not sure if we want to enforce that in code, generally I've been assuming this is part of the contract though.

Lukasa:
Agreed. The only reason it's confusing is that there's a `set_cert` method that doesn't clearly only get called once: at least, it doesn't assert that that's part of the contract. But fine, let's do that then. =)

shazow:
Hrm agreed, want to add a little docstring to mention that it should only be called once before the pool is used?

haikuginger:
I'm of the :shipit: mindset on this; could we include it in 1.17 with `SessionManager`? That gives us two marquee features, with 1.16 to fall back on if we find that either one needs some retooling.

Lukasa:
I'm with @haikuginger, I think that's a very good idea.

shazow:
Maybe it's time to figure out how to do an opt-in beta release on pypi.

haikuginger:
I think PyPI discourages that in general- and we probably wouldn't get enough usage information back for it to be meaningful anyway.

Let's go ahead with the plan, but be prepared to bang out 1.17.1, 1.17.2, and/or 1.17.3 in quick succession.

jayvdb:
I suspect there would be a decent amount of usage if requests also did a pre-release on pypi that included the pre-released urllib3.  I would add a requests-pre-release job to my build matrix, and we could work with some significant downstream users to ask them to do the same.

Lukasa:
The requests team can probably get our act together enough to do something like that around PyCon time.

sigmavirus24:
> I think PyPI discourages that in general- and we probably wouldn't get enough usage information back for it to be meaningful anyway.

Pre-releases need only be properly versioned and then users need to opt-in with `pip install --pre`. We can use download count to determine that but some amount of that will be bandersnatch mirrors fetching the package once uploaded.

haikuginger:
@sigmavirus24: that's good to know. I think I must have been looking at outdated information, then.

docs/security.rst
```diff
@@ -111,6 +111,41 @@ Once you find your root certificate file::
         ...
 
 
+Custom SSL Contexts
+-------------------
+
+.. versionadded:: 1.16.0
```
shazow: If we're saving this for 1.17, should update this.

Lukasa: Done.

shazow:
Hm I think this is safe to merge even without the sessionmanager stuff, right?

haikuginger:
Safe to my knowledge; we had initially wanted to keep them together to keep all the API breakage in a single release, but that's the only reason I'm aware of that it didn't go into 1.16 in the first place.

sigmavirus24:
I would think we should merge this. This is a feature we've been promising for a while. Plus, I'm not sure how it breaks anyone's usage of the API

shazow:
Let's just merge it now once Travis confirms it's good, since this is stable.

sigmavirus24:
👍 Agreed. (Took a quick look at the code again, looks great to me. Still don't understand why this is an API breaking change as it's completely optional for consumers.)

Lukasa:
Cool, let's do it! This also enables something I've been wanting to look at, which is attempting to reduce the number of optional dependencies required for PyOpenSSL injection.

PR #840, opened 2016-04-11
Disables IPv6 DNS when IPv6 connections are not possible

### LINKED ISSUE #838: urllib3 attempts to use IPv6 even when IPv6 is disabled
This is an issue when running on a server without IPv6 (must be disabled because the network does not support it). Example when connecting to https://graph.facebook.com using requests and IPv4 happens to fail:

```
HTTPSConnectionPool(host='graph.facebook.com', port=443): Max retries exceeded with url: /v2.5/me/feed (Caused by NewConnectionError('<requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x7f4dbd158518>: Failed to establish a new connection: [Errno 97] Address family not supported by protocol',))
  File "/home/lib/python3.4/site-packages/requests/packages/urllib3/connection.py", line 137, in _new_conn
  File "/home/lib/python3.4/site-packages/requests/packages/urllib3/util/connection.py", line 91, in create_connection
    raise err
  File "/home/lib/python3.4/site-packages/requests/packages/urllib3/util/connection.py", line 71, in create_connection
    sock = socket.socket(af, socktype, proto)
  File "/usr/lib/python3.4/socket.py", line 126, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
OSError: [Errno 97] Address family not supported by protocol
```

urllib3 should throw an exception after exhausting all IPv4 options instead of trying (and invariably failing) IPv6.

See closed issue https://github.com/kennethreitz/requests/issues/3084.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32805768-urllib3-attempts-to-use-ipv6-even-when-ipv6-is-disabled?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

haikuginger:
Fixes #838 

Note that this will only work in cases where either Python has been compiled without IPv6 support or the host has IPv6 networking disabled (an exception is raised when opening a socket to `::1`); beyond that, we have no real way to check IPv6 availability.

Credit where it's due, this is existing code from the test module with unit tests added around it.

### DIFF
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -46,6 +46,8 @@ def is_connection_dropped(conn):  # Platform-specific
 
 # This function is copied from socket.py in the Python 2.7 standard
 # library test suite. Added to its signature is only `socket_options`.
+# One additional modification is that we avoid binding to IPv6 servers
+# discovered in DNS if the system doesn't have IPv6 functionality.
 def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
                       source_address=None, socket_options=None):
     """Connect to *address* and return the socket object.
@@ -64,7 +66,8 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     if host.startswith('['):
         host = host.strip('[]')
     err = None
-    for res in socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM):
+    family = supported_ip_family()
+    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
         af, socktype, proto, canonname, sa = res
         sock = None
         try:
@@ -99,3 +102,35 @@ def _set_socket_options(sock, options):
 
     for opt in options:
         sock.setsockopt(*opt)
+
+
+def supported_ip_family():
+    family = socket.AF_INET
+    if HAS_IPV6:
+        family = 0
+    return family
+
+
+def _has_ipv6(host):
+    """ Returns True if the system can bind an IPv6 address. """
+    sock = None
+    has_ipv6 = False
+
+    if socket.has_ipv6:
+        # has_ipv6 returns true if cPython was compiled with IPv6 support.
+        # It does not tell us if the system has IPv6 support enabled. To
+        # determine that we must bind to an IPv6 address.
+        # https://github.com/shazow/urllib3/pull/611
+        # https://bugs.python.org/issue658327
+        try:
+            sock = socket.socket(socket.AF_INET6)
+            sock.bind((host, 0))
+            has_ipv6 = True
+        except:
+            pass
+
+    if sock:
+        sock.close()
+    return has_ipv6
+
+HAS_IPV6 = _has_ipv6('::1')
--- a/urllib3/util/connection.py
+++ b/urllib3/util/connection.py
@@ -67,7 +67,7 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
         host = host.strip('[]')
     err = None
 
-    # Using the value from family_filter() in the context of getaddrinfo lets 
+    # Using the value from family_filter() in the context of getaddrinfo lets
     # us select whether to work with IPv4 DNS records, IPv6 records, or both.
     # The original create_connection function always returns all records.
     family = family_filter()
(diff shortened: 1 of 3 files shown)


haikuginger:
Looks like the failing checks need to be re-run; passes locally and on Travis for my branch.

kz26:
Looks good, many thanks!

urllib3/util/connection.py
```diff
@@ -99,3 +102,35 @@ def _set_socket_options(sock, options):
...
+        # has_ipv6 returns true if cPython was compiled with IPv6 support.
+        # It does not tell us if the system has IPv6 support enabled. To
+        # determine that we must bind to an IPv6 address.
+        # https://github.com/shazow/urllib3/pull/611
+        # https://bugs.python.org/issue658327
+        try:
+            sock = socket.socket(socket.AF_INET6)
+            sock.bind((host, 0))
+            has_ipv6 = True
+        except:
```
Lukasa: Can this be `except Exception` please?

urllib3/util/connection.py
```diff
@@ -64,7 +66,8 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     if host.startswith('['):
         host = host.strip('[]')
     err = None
-    for res in socket.getaddrinfo(host, port, 0, socket.SOCK_STREAM):
+    family = supported_ip_family()
+    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
```
Lukasa: Ok, so this is fine but the code is moderately unclear. I don't think `supported_ip_family` is a good name for this function: both here and lower down I was a bit confused about what it was actually achieving.

I wonder if the better approach here is to call it something like `family_filter`, and then add a comment here that explains how it works in the context of `getaddrinfo`, *and* a docstring in the actual function itself explaining why it sometimes returns zero.

Lukasa:
Ok, this looks like a reasonable approach to me! I've left some notes inline for cleaning up the code, but I think this fix would be absolutely fine when those get tidied up.

haikuginger:
Looks like Travis needs another nudge.

Lukasa:
Ok, I'm happy. @sigmavirus24, want to review?

Please don't merge this yet though: I'd like to get a .1 release out the door today and that works best if we don't accidentally slip other bugfixes into it.

urllib3/util/connection.py
```diff
@@ -99,3 +106,39 @@ def _set_socket_options(sock, options):
 
     for opt in options:
         sock.setsockopt(*opt)
+
+
+def family_filter():
```
shazow: Not sure "filter" is the right word here. Maybe `get_address_family()`?

Filter makes me think that it takes some input and returns a subset output.

shazow:
One note. Seems reasonable. :)

urllib3/util/connection.py
haikuginger: I originally had `supported_ip_family()`, but I think `filter` works as well. Seems like it might be worth more discussion; I'll wait until there's a consensus on method name to re-commit.

Lukasa: This name is really tricky, because it's contextual.

`get_address_family()` feels wrong because this function will *never* return `AF_INET6`. It has exactly two return values, `AF_INET` and `0`, where `0` is only meaningful when you realise that it gets passed into `getaddrinfo`. If the code returned `AF_UNSPEC` instead of `0` that would get us closer to something sensible.

Maybe return `AF_UNSPEC` and change the name to indicate that this is the family flag for `getaddrinfo` in some way: `gai_family`, for example? I dunno, just some thoughts.

shazow: > Maybe return `AF_UNSPEC` and change the name to indicate that this is the family flag for `getaddrinfo` in some way: `gai_family`, for example? 

+1, I'd be comfortable with that.

haikuginger: `allowed_gai_family()`? Conveys the context (`gai`) and makes clear that it's restricting some options.

:+1: on the change to AF_UNSPEC; didn't come across that constant when initially setting this up.

shazow: Both sound good to me. :)

Lukasa: Both are equally fine for me. =)

sigmavirus24:
Looks good to me :)

shazow:
Thank you!

Issue #843, opened 2016-04-13
Support cookies when the PoolManager or ConnectionPool is passed a CookieJar object


haikuginger:
I think I can support cookie handling with the stdlib `CookieJar` object without too much trouble, and wanted to get some thoughts on implementation first.

The biggest change is that both `CookieJar.extract_cookies` and `CookieJar.add_cookie_header` both expect to work with a `urllib.request.Request`-like object. We can work around this by temporarily loading basic request information (essentially URL) into the appropriate object and then pull the modifications back out, but if we're doing that, it may also be worthwhile to consider a `urllib3.request.Request` object that can be passed around like `urllib.request.Request`.

The second, more minor, change is that `CookieJar.extract_cookies` expects an `info()` method on the response which returns response headers. Given that there's an existing urllib3 header API, for this case, I'd rather just build a shim to give `CookieJar` what it needs.

Regardless, thoughts are appreciated on two questions: is this an appropriate feature for urllib3 to support, and if so, is using the stdlib to this extent an appropriate way to implement it?

shazow:
I'm definitely open to proposals for this. In my mind, I've been planning on including a higher-level SessionManager (which is a composition of a PoolManager or ConnectionPool) which would be cookie-aware among other things. This could be a first iteration of this.

Lukasa:
In terms of looking for inspiration, Requests does some shimming to talk to the stdlib. Take a look at requests' `cookies.py` file to see what we do. The approach is a sensible one though!

shazow:
Also some related discussion here: https://github.com/shazow/urllib3/issues/776

haikuginger:
Preference question: if the request both explicitly declares a `Cookie` header AND a `CookieJar` has been provided, should the behavior be to rely on the CookieJar's Cookie header, the explicitly stated Cookie header, or should we try to merge them with `'; '.join()`?

Lukasa:
@haikuginger I'd be inclined to send multiple Cookie headers, in fact, though right now I don't know that urllib3 will let us do that.

haikuginger:
Per [RFC-6265](https://tools.ietf.org/html/rfc6265#section-5.4), we should only be going with a single Cookie header, but we can join them with semicolons to send both cookies.

> When the user agent generates an HTTP request, the user agent MUST NOT attach more than one Cookie header field.

Given that, I think merging them into a single Cookie header might be the way to go.

One other item while we're in the RFC:

[...]

PR #844, opened 2016-04-15
Add cookie support

haikuginger:
Initial implementation of #843; updates the bundled copy of `six` to v1.10.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -38,7 +38,10 @@ from .connection import (
     HTTPConnection, HTTPSConnection, VerifiedHTTPSConnection,
     HTTPException, BaseSSLError,
 )
-from .request import RequestMethods
+from .request import (
+    RequestMethods,
+    Request
+)
 from .response import HTTPResponse
 
 from .util.connection import is_connection_dropped
@@ -46,9 +49,11 @@ from .util.response import assert_header_parsing
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import get_host, Url
+from .util.request import make_cookie_header
 
 
 xrange = six.moves.xrange
+CookieJar = six.moves.http_cookiejar.CookieJar
 
 log = logging.getLogger(__name__)
 
@@ -167,7 +172,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
     def __init__(self, host, port=None, strict=False,
                  timeout=Timeout.DEFAULT_TIMEOUT, maxsize=1, block=False,
                  headers=None, retries=None,
-                 _proxy=None, _proxy_headers=None,
+                 _proxy=None, _proxy_headers=None, cj=None,
                  **conn_kw):
         ConnectionPool.__init__(self, host, port)
         RequestMethods.__init__(self, headers)
@@ -188,6 +193,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         self.proxy = _proxy
         self.proxy_headers = _proxy_headers or {}
+
+        self.cookie_jar = cj
 
         # Fill the queue up so that doing get() on it will block properly
         for _ in xrange(maxsize):
@@ -443,7 +450,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
     def urlopen(self, method, url, body=None, headers=None, retries=None,
                 redirect=True, assert_same_host=True, timeout=_Default,
-                pool_timeout=None, release_conn=None, chunked=False,
+                pool_timeout=None, release_conn=None, chunked=False, cj=None,
                 **response_kw):
         """
         Get a connection from the pool and perform an HTTP request. This is the
@@ -461,6 +468,12 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
            `preload_content=False` because we want to make
            `preload_content=False` the default behaviour someday soon without
            breaking backwards compatibility.
+
+        .. note::
+
+           `url` typically receives a subpath, rather than a full URI. Other
+           aspects of the URI are specified during the instantiation of the
+           ConnectionPool.
 
         :param method:
             HTTP request method (such as GET, POST, PUT, etc.)
@@ -530,6 +543,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             Additional parameters are passed to
             :meth:`urllib3.response.HTTPResponse.from_httplib`
         """
+        original_headers = headers
+
         if headers is None:
             headers = self.headers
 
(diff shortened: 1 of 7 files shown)


urllib3/request.py
```diff
@@ -6,8 +6,31 @@
...
+
+
+class Request(_Request):
+    """
+    Currently used as a shim to allow us to work with the stdlib cookie
+    handling, which expects a `urllib.request.Request`-like object.
+    """
+    def __init__(self, *args, **kwargs):
+        self._method = kwargs.pop('method', None)
+        _Request.__init__(self, *args, **kwargs)
```
Lukasa: Any reason not to use `super()` here?

urllib3/connectionpool.py
```diff
@@ -167,7 +172,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
     def __init__(self, host, port=None, strict=False,
                  timeout=Timeout.DEFAULT_TIMEOUT, maxsize=1, block=False,
                  headers=None, retries=None,
-                 _proxy=None, _proxy_headers=None,
+                 _proxy=None, _proxy_headers=None, cj=None,
```
Lukasa: I don't know if I like calling the cookie jar argument `cj`: seems like needless brevity.

urllib3/request.py
haikuginger: No; will change.

urllib3/connectionpool.py
haikuginger: Agreed; `cookie_jar`?

haikuginger:
One behavior in this revision that may be worth more discussion is handling the case where the ConnectionPool has a `CookieJar` already, and one is passed in the `request` method.

Currently, in that scenario, the method-passed `CookieJar` alone is used for both cookie retrieval and storage to avoid surprises; I think it'd be a bit odd if we pulled from a jar and didn't store in it, and even odder if we stored the same cookie in two places when a specific place to store it has been passed to us.

Of course, open to changes, though.

sigmavirus24:
ce9394a2564608823019ad6b59276bfa82bad642 should be a separate pull request.

urllib3/connectionpool.py
sigmavirus24: Why is the connection pool handling this? This feels a bit ... out of scope for this class.

haikuginger: So, in re: the connection pool having the cookie jar as an instance, the decision there will be related to the decision about instance vs. argument `CookieJar`s (see below). If we're just going to use one or the other, then we can move the jar selection logic into `PoolManager.urlopen()` and just have the `CookieJar` as an instance attribute of that object, passing a single jar into `ConnectionPool.urlopen()`.

If, on the other hand, we want to use both instance and argument jars, then we need to be able to access them separately from within `ConnectionPool.urlopen()`, since cookie handling logic requires us to have both the `Request`-like and `Response`-like objects at the same time. In that case, we'd initialize each `ConnectionPool` with the same `CookieJar` using `PoolManager.connection_pool_kw`, and leave the option open to receive another `CookieJar` in the `.request` kwargs. This implementation also allows setting up single ConnectionPools that can handle cookies independently.

Regardless, the most efficient way to do all the reset of the cookie-handling logic will be inside `ConnectionPool.urlopen()`.

shazow:
I would very very much prefer to keep the current pools unaware of cookies. Ideally we should have composition by using a higher-level class (such as `SessionManager` mentioned in the other thread) that contains/embeds a pool-like object and proxies requests/responses to these other pools, fiddling with headers along the way.

haikuginger:
@shazow, the cost to going that way is that all the redirect logic happens inside `ConnectionPool.urlopen()`; we'd lose the chance to get cookies on 303s (not sure if that's a common use case). There's also the chance, if we're setting any headers in `PoolManager` or `ConnectionPool`, that we'll lose the `Cookie` header we set on the higher level.

Rather than having cookie logic inside `ConnectionPool`, what would the thoughts be on an optional `Context` object passed into `ConnectionPool.urlopen()` which, if present, gets sent a copy of the final request and response to manipulate? `SessionManager` could pass it into every request it makes, and `ConnectionPool.urlopen()` could use it in a relatively naïve manner - `Context.retrieve(request)` and `Context.update(request, response)`, for example. Right now, that'd be a pretty thin wrapper for the `SessionManager`'s instance `CookieJar`, but that could change in the future, and could have an API to expand its use.

Let me do some fiddling around a `SessionManager` and see where I get. This whole PR is a first draft, so I'm not too attached to it.

Lukasa:
Cookies on redirects are remarkably common, and need to be handled.

sigmavirus24:
You'll also want to be secure by default by using the right cookie jar policy so that you don't send unscoped cookies to every domain under the sun. (You also need to make sure you parse the cookie headers appropriately that way domains that don't specify a host are associated with one anyway.)

shazow:
>  the cost to going that way is that all the redirect logic happens inside ConnectionPool.urlopen(); we'd lose the chance to get cookies on 303s (not sure if that's a common use case).

The redirect logic largely lives in the Retries abstraction. The piece that live inside of ConnectionPool is only a few lines long which I'd be happy to replicate or abstract into a shared helper. I don't feel that's a strong reason to keep it there, at least. Is there anything else?

> There's also the chance, if we're setting any headers in `PoolManager` or `ConnectionPool`, that we'll lose the `Cookie` header we set on the higher level.

I think it's reasonable to make the assumption that if you're using a higher-level SessionManager, then mutating the state of the encapsulated Pool is out of scope of our guarantees.

The Context object proposal is not terrible but I'd prefer to avoid it if we can. I feel that's what the Retry object is basically right now, and if we go that route then we should merge that functionality somehow, and further creep the scope into request history also.

haikuginger:
Okay! Lots of stuff here.

@sigmavirus24: the security concern should be getting handled by the default policy; I'll add some unit testing to ensure that's the case, and include an API in the `SessionManager` to let users set their own policy.

It sounds like encapsulation of the PoolManager/ConnectionPools is the name of the game, so we can definitely go that way. Essentially, I'll need to build the Response object out a bit so we can work with it at a higher level, and play with the redirect logic.

@shazow: Would you still prefer to avoid a Context object if it incorporated the items you mentioned (retry and request history)? That's work I'm happy to do if it shifts the balance.

Any other thoughts before I get to work on a second draft?

sigmavirus24:
> the security concern should be getting handled by the default policy

Yeah, you would *hope* that, but it is *not* the case. See https://github.com/kennethreitz/requests/issues/2576 for more details.

haikuginger:
Very good to know. Looks like we'd want `DomainStrict`, then, unless I'm mistaken (as always, quite possible).

shazow:
@haikuginger I would suggest avoiding the `Context` object if we can. That's going to be a bigger change than it seems because we're going to insist on merging a whole bunch of existing functionality in there. :) 

In general, when possible, try to make the minimal amount of changes in existing components. This helps keep reduce the API surface that we're making compatibility commitments on. The other reason is that we want to be able to swap out functionality if needed (e.g. someone should be able to modify or bring their own cookie management logic).

haikuginger:
I'm closing this PR as it's superseded by #849.

PR #846, opened 2016-04-16
pyopenssl: add support for IP Address SAN

FiloSottile:
See docker/docker-py#1037


Lukasa:
Thanks for this!

Just a quick question, but what version of Python are you using this with? Because this is only really half the code needed to get this to work properly: we also need to update the match_hostname function for older Pythons.

Issue #578, opened 2015-03-25
Attach last retry object to response before returning


shazow:
Would be swell if our `response.HTTPResponse` object had a `retries` attribute (or maybe a better name?) which contained the latest (immutable) Retry object.

At least that way users can get some idea of how many retries occurred (derived from their original retries configuration). Bonus points if we store more useful data in the retry object for the user to inspect without having to do math.

This attachment could happen right before [returning the response here](https://github.com/shazow/urllib3/blob/431e80619efb73201f483467a68ee0c8a215e5d3/urllib3/connectionpool.py#L650).

(Mildly related to #576.)

shazow:
Ping @wolever, he expressed some interest in doing this. :)

nateprewitt:
This should be resolved in #848.

Issue #576, opened 2015-03-23
Retry history (or maybe request history via retries?)


shazow:
Would be great if we have somekind of retry request history in the retry object. There's a TODO for this in the code.

Ideally we'd do this by replacing our observed errors counter with a list of objects describing the respective requests.

If we end up doing a request history (and requests have a retry attached per #578), then this issue would be obsolete.

Related: #185 which adds redirect_history, but we should be able to do better now that we've refactored some of the code.

sigmavirus24:
I might take a look at this sometime in the future. Of course if someone wants to get involved in urllib3 we can work together on it if you'd like.

nateprewitt:
This should be resolved in #848.

PR #848, opened 2016-04-18
Attach last Retry object to response and add Retry history

palaviv:
This PR is trying to solve both Issue #576 and #578.
The last Retry object of the request is added to the response object. Thus allowing the user to check if any errors occurred during the request. In addition to give more information to the user the Retry object now has a history attribute. This attribute contains a list of the errors that were thrown at each connection retry.

example code:

    retry = Retry(total=2)
    resp = pool.request('GET', '/', retries=retry)
    if resp.retries.total != 2:
        print(resp.retries.history)  # will show list of errors


shazow:
Had a quick look, some initial feels: Looks like a good step in the right direction, but I think we're going to need for history to be a list of an object that encapsulates some request (method and resource at least) and response information (status code, exception, other things?), rather than just a list of exceptions encountered. The use case is we should be able to look up redirect history based on this list.

palaviv:
How about a new object ("RetryHistory" working title) that will contain all the data passed to the Retry.increment: method, url, response, error?

shazow:
> How about a new object ("RetryHistory" working title) that will contain all the data passed to the Retry.increment: method, url, response, error?

RetryHistory makes sense, or RequestHistory (or ResponseHistory, or both?).

I want to avoid memory leaks, so let's be careful storing too much historical request/response data, specifically we want to avoid hanging on to the bodies. That is to say, avoid storing the full response and just store the minimal metadata for it.

palaviv:
I changed the history to a list of ``RequestHistory`` objects. Each object contains the following fields:
method, url, error, status, redirect_location.

palaviv:
I added the test for multiple redirects from PR #185. I hope this is ok @cezarsa.

urllib3/util/retry.py
```diff
@@ -15,6 +16,16 @@
...
 
+class RequestHistory(namedtuple('RetryHistory', ["method", "url", "error",
+                                                 "status", "redirect_location"])):
+    """
+    Datastructure for representing the metadata of requests that result in a retry.
+    """
+    def __new__(cls, method, url, error, status, redirect_location):
+        return super(RequestHistory, cls).__new__(cls, method, url, error,
+                                                  status, redirect_location)
+
```
Lukasa: Why has this been turned into a subclass of `namedtuple`?

urllib3/util/retry.py
```diff
@@ -169,10 +184,10 @@ def get_backoff_time(self):
 
         :rtype: float
         """
-        if self._observed_errors <= 1:
+        if len(self.history) <= 1:
             return 0
 
-        backoff_value = self.backoff_factor * (2 ** (self._observed_errors - 1))
+        backoff_value = self.backoff_factor * (2 ** (len(self.history) - 1))
```
Lukasa: This change subtly affects the way the backoff works. In particular, redirects will now count towards the backoff count. This means that redirects and retries can interact in a way to dramatically slow things down. Presumably we should count redirects and retries separately from the perspective of backing off.

We may even want to go further here: perhaps each host should handle its backoff separately.

shazow: Good point, only errors should trigger backoff.

I see the value of doing it per-host but at this point it's probably more effort than it's worth. :/

shazow: Also, do we want to reset the backoff every time we get a non-error response? So only look at the last sequence of contiguous errors?

palaviv:
Fixed the CR comment's

urllib3/util/retry.py
```diff
@@ -169,10 +179,13 @@ def get_backoff_time(self):
 
         :rtype: float
         """
-        if self._observed_errors <= 1:
+        # We want to consider only the last consecutive errors sequence (Ignore redirects).
+        consecutive_errors_len = len(list(takewhile(lambda x: x.redirect_location is None,
```
shazow: Not necessarily clearer but could also write this as:

```python
consecutive_errors_len = next(i for i, h in enumerate(reversed(self.history)) if not h.redirect_location)
```

At least no lambdas and itertools, but either way is fine. :) Just a nit.

Also any thoughts if it would it make more sense to look at `.error` or `.status` instead?

palaviv: >```python
consecutive_errors_len = next(i for i, h in enumerate(reversed(self.history)) if not h.redirect_location)
```

This will raise ``StopIteration`` if there aren't any redirects in the history list.

The only retry event we don't consider as error is redirect. Currently in ``get_redirect_location`` we check that ``status`` is a redirect status. The only way that ``.redirect_location`` will be different then None is if the call to ``get_redirect_location`` does not return False. IMO I think this is a simpler check then ``x.error or (x.status and x.status not in REDIRECT_STATUSES)``.

Comment (shazow) [inline urllib3/util/retry.py]: Er sorry I meant `consecutive_errors_len = next((i for i, h in enumerate(reversed(self.history)) if not h.redirect_location), 0)`

Fair enough.

**Comment by haikuginger** [comment]
Looks like an ephemeral bug. I've restarted the specific test that failed.

**Comment by palaviv** [comment]
Thank you @haikuginger.

File: dummyserver/handlers.py [hunk 4]
```diff
@@ -159,6 +161,18 @@ def redirect(self, request):
         headers = [('Location', target)]
         return Response(status='303 See Other', headers=headers)
 
+    def multi_redirect(self, request):
+        "Performs a redirect chain based on ``redirect_codes``"
+        codes = request.params.get('redirect_codes', '200').decode('utf-8')
+        codes = codes.split(',')
+        head, tail = codes[0], codes[1:]
```
Lukasa: You can simplify this somewhat by changing to:

```python
head, tail = codes.split(',', 1)
```

Lukasa: That also removes the need for a `join` later on, on line 173: the string will just remain intact.

test/contrib/test_gae_manager.py
```diff
@@ -177,8 +177,24 @@ def test_default_method_whitelist_retried(self):
...
+        self.assertEqual(resp.retries.total, 1)
+        # URLFetch use absolute urls.
+        self.assertEqual(resp.retries.history,
+                         [RequestHistory('GET',
+                                         self.pool._absolute_url('/successful_retry'),
+                                         None, 418, None)])
+
     #test_max_retry = None
     #test_disabled_retry = None
+    # URLFetch resolve redirects.
```
Lukasa: Can we reword this to be slightly longer and clearer? Maybe

```python
# We don't need these tests because URLFetch resolves its own redirects.
```

test/test_response.py
```diff
@@ -525,6 +525,13 @@ def test_get_case_insensitive_headers(self):
         self.assertEqual(r.headers.get('host'), 'example.com')
         self.assertEqual(r.headers.get('Host'), 'example.com')
 
+    def test_retries(self):
+        fp = BytesIO(b'')
+        resp = HTTPResponse(fp)
+        self.assertEqual(resp.retries, None)
+        resp = HTTPResponse(fp, retries="test")
+        self.assertEqual(resp.retries, "test")
```
Lukasa: What exactly is this testing?

Lukasa: Put another way: it'd be better if this set `retries` to a value that is more like what it would *actually* be set to.

test/test_retry.py
```diff
@@ -131,6 +131,18 @@ def test_zero_backoff(self):
...
+        retry = retry.increment()
+        retry = retry.increment()
+        self.assertEqual(retry.get_backoff_time(), 0.4)
+        redirect_response = HTTPResponse(status=302, headers={'location': 'test'})
+        retry = retry.increment(response=redirect_response)
+        self.assertEqual(retry.get_backoff_time(), 0)
+        retry = retry.increment()
+        retry = retry.increment()
+        self.assertEqual(retry.get_backoff_time(), 0.4)
+
```
Lukasa: This test doesn't distinguish between linear and exponential backoff. Can we add one more increment?

Lukasa:
Sorry to come driving in with more review @palaviv, but @shazow is taking a break so I'm taking over reviewing this patch. However, you'll be happy to know that I basically only have the notes I've made here to give: so long as no more are introduced when you fix those up, I'll be happy to merge this soon!

dummyserver/handlers.py
palaviv: done

test/contrib/test_gae_manager.py
palaviv: done

test/test_response.py
palaviv: done

test/test_retry.py
palaviv: This test is not checking that we increment exponentially. The test check that we reset after a redirect. I actually think that I should remove one of the  ``retry.increment()`` at each stage as it is not contributing to this test purpose.

Lukasa: That's fine: feel free to do that. I'd *also* like a check for the backoff function, just because it's one of those implicit things that can easily be overlooked.

palaviv: I think that ``RetryTest.test_backoff`` already covers what you ask for.

Lukasa: Oh apologies, so it does.

Lukasa:
Ok, I think I'm happy. Any other contributors got thoughts? If not, I'm happy for us to go ahead and merge this.

sigmavirus24:
I'm only concerned about history being a list and the expensive slice copy we make of it. Is it really important that history be a list and not a tuple?

palaviv:
I changed history from list to tuple @sigmavirus24.

sigmavirus24:
This looks good to me.

haikuginger:
Looks good to me. @palaviv, do you want to add a line item to `CHANGES.rst`?

palaviv:
Added the `CHANGES.rst` entry @haikuginger.

Lukasa:
Hooray, much happiness was had by all. Thanks for reviews @sigmavirus24 and @haikuginger, and thanks so much for doing the work @palaviv! :sparkles: :custard: :cookie:

PR #849, opened 2016-04-20
SessionManager (cookies and autoconfiguration)

haikuginger:
Implements #843; supersedes #844.


urllib3/poolmanager.py
```diff
@@ -142,7 +137,7 @@ def connection_from_url(self, url):
         u = parse_url(url)
         return self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
 
-    def urlopen(self, method, url, redirect=True, **kw):
+    def urlopen(self, method, url, redirect=True, retries=None, **kw):
```
Lukasa: I can't see why `retries` needed to become an explicit parameter here. Care to explain the reasoning?

urllib3/poolmanager.py
```diff
@@ -153,44 +148,23 @@ def urlopen(self, method, url, redirect=True, **kw):
...
+        if self.proxy is not None and u.scheme == "http":
+            response = conn.urlopen(method, url, retries=retries, **kw)
+        else:
+            response = conn.urlopen(method, u.request_uri, retries=retries, **kw)
 
-        log.info("Redirecting %s -> %s", url, redirect_location)
-        return self.urlopen(method, redirect_location, **kw)
+        if redirect and response.get_redirect_location():
+            kw['redirect'] = redirect
+            return handle_redirect(response, self, retries, method, url, pool=conn, **kw)
```
Lukasa: I'd really love it if you used keyword arguments here rather than positional: in particular, the `self` in here is extremely mysterious!

urllib3/request.py
```diff
@@ -1,13 +1,47 @@
...
-except ImportError:
-    from urllib import urlencode
 
 from .filepost import encode_multipart_formdata
 
+from .packages import six
 
-__all__ = ['RequestMethods']
+_Request = six.moves.urllib.request.Request
+urlencode = six.moves.urllib.parse.urlencode
```
Lukasa: Can we do this as `from six.moves.urllib.request import Request` and `from six.moves.urllib.parse import urlencode`?

urllib3/request.py
```diff
@@ -1,13 +1,47 @@
...
+__all__ = ['RequestMethods', 'Request']
+
+
+class Request(_Request):
+    """
+    Currently used as a shim to allow us to work with the stdlib cookie
+    handling, which expects a :class:`urllib.request.Request`-like object.
+    """
+    def __init__(self, *args, **kwargs):
+        self._method = kwargs.pop('method', None)
```
Lukasa: Why are we saving `_method`?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,78 @@
+from .util.retry import Retry, handle_redirect
+from .request import RequestMethods, Request
+from .poolmanager import PoolManager, ProxyManager
+from .util.context import Context
+
+
+class SessionManager(RequestMethods):
+    """
+    Allows arbitrary requests while maintaining browser-session-like
+    context across those requests.
```
Lukasa: I'd love this to be much more clear about what, exactly, constitutes browser-session-like context. What is being maintained?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,78 @@
...
+    :param \**pm_kwargs:
+        Additional paramaters to be passed to the encapsulated
+        :class:`.PoolManager` or :class:`.ProxyManager` instance.
+    """
+    def __init__(self, headers=None, cookie_jar=None, **pm_kwargs):
+
+        super(SessionManager, self).__init__(headers=headers)
+
+        # Detect if we should create a PoolManager or a ProxyManager
+        if pm_kwargs.get('proxy_url', None) is None:
```
Lukasa: Might be better to check truthiness here, rather than `None` identity, to allow for the admittedly minor possibility that a falsy string ends up here.

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Set the encapsulated context's cookie policy
+        """
+        self.context.set_cookie_policy(policy)
+
+    def apply_context(self, request):
+        self.context.apply_to(request)
+
+    def extract_context(self, response, request):
+        self.context.extract_from(response, request)
```
Lukasa: These methods need docstrings explaining their purpose.

urllib3/util/context.py
```diff
@@ -0,0 +1,84 @@
+import time
+
+from ..packages import six
+
+cookielib = six.moves.http_cookiejar
+PythonCookiePolicy = cookielib.DefaultCookiePolicy
+PythonCookieJar = cookielib.CookieJar
```
Lukasa: Same thing here, can we do these using `from x import y as z`?

urllib3/util/retry.py
```diff
@@ -11,7 +11,7 @@
 )
 from ..packages import six
 
-
+urljoin = six.moves.urllib.parse.urljoin
```
Lukasa: Same here.

urllib3/util/retry.py
```diff
@@ -290,5 +290,31 @@ def __repr__(self):
...
+    """
+    Abstraction of the redirect process so that we can access it from
+    multidomain levels in addition to from the :class:`.ConnectionPool`.
+    """
+    # urljoin will give us either a full URL or just a path, based on what
+    # the original URL is, making it appropriate for the ConnectionPool
+    # level as well as the PoolManager or SessionManager
+    redirect_location = urljoin(url, response.get_redirect_location())
+    if response.status == 303:
+        method = 'GET'
```
Lukasa: This is somewhat incomplete. HEAD requests should also stay HEAD requests (a general rule in redirects). There are also questions about whether you should go further in emulating browsers, though that's up to you.

urllib3/util/retry.py
```diff
@@ -290,5 +290,31 @@ def __repr__(self):
...
 
+def handle_redirect(response, requester, retries, method, url, *args, **kwargs):
+    """
+    Abstraction of the redirect process so that we can access it from
+    multidomain levels in addition to from the :class:`.ConnectionPool`.
+    """
+    # urljoin will give us either a full URL or just a path, based on what
+    # the original URL is, making it appropriate for the ConnectionPool
+    # level as well as the PoolManager or SessionManager
+    redirect_location = urljoin(url, response.get_redirect_location())
```
Lukasa: Hrrrm. Does this actually work? Location headers can come in a wide variety of forms and styles, and in particular can come without a scheme.

Additionally, my reading of this PR is that it *changes* the redirect logic currently possessed by urllib3, which previously didn't do this urljoin. Why the change?

Lukasa:
This is a biggish change and will need a *lot* of code review, but it's a great start! I've left some notes inline. =)

urllib3/util/retry.py
haikuginger: At a minimum, it handles all the cases that are currently tested for on `ConnectionPool` and `PoolManager`.

I'm not clear why it wasn't used across the board in the past; it was present in `PoolManager`'s block of redirect logic, but not in `ConnectionPool`'s. I suspect that it may have been discovered when setting up the cross-domain redirect logic for `PoolManager`, but wasn't applied to `ConnectionPool` because at the time they were independent pieces of logic.

In any manual tests I do, it seems to behave as I'd hope, including with locations not containing a scheme. Essentially:

    >>> urljoin('http://host/path/resource', 'http://other_host/path/resource')
    'http://otherhost/path/resource'
    >>> urljoin('http://host/path/resource', '/otherpath/otherresource')
    'http://host/otherpath/otherresource'
    >>> urljoin('http://host/path/resource', 'otherresource')
    'http://host/path/otherresource'
    >>> urljoin('/path/resource', '/otherresource')
    '/otherresource'
    >>> urljoin('/path/resource', 'nested/otherresource')
    '/path/nested/otherresource'

urllib3/util/retry.py
haikuginger: All right; should I pull from `RequestMethods._encode_url_methods` to determine which methods can be left alone? Additionally, should I be using `response.REDIRECT_STATUSES` instead of just 303? I didn't want to change the external behavior on that unilaterally.

urllib3/sessionmanager.py
haikuginger: Will do!

urllib3/sessionmanager.py
haikuginger: I'm of two minds on this; on the one, a falsy string will pretty much never be a valid proxy. On the other, should urllib3 be expected to do anything to check validity on a high level when it receives a proxy URL, or just raise the appropriate exception when the connection fails? I'm fine either way, though, so I can make that change if you want.

urllib3/sessionmanager.py
haikuginger: Gotcha, will clarify.

urllib3/request.py
haikuginger: The `urllib2.Request` object in Python 2 actually can't handle a `method` parameter natively, unfortunately, so we have to pull it before initializing. If we ever end up using it, we can add a `get_method()` method that'll check for that variable and fall through to the superclass's `get_method()` method.

urllib3/request.py
haikuginger: Odd, I could have sworn that failed earlier when I tried it. Will go through and make those changes.

urllib3/poolmanager.py
haikuginger: Yeah, I'm not thrilled with `handle_redirect`'s signature at the moment. One thing I thought about was making it an instance method on `RequestMethod`, but I'm not sure if that'd be appropriate.

It's also tricky because a few of those are legitimately required arguments - for example, without `retries`, we'd redirect indefinitely.

urllib3/poolmanager.py
haikuginger: This is done to pop it out of `kw` early in support of the awful method signature for `handle_redirect()`; see below.

haikuginger:
Explanation on the last commit: essentially, doing `from six.moves.http_cookies import {name}` brings along `http.cookiejar`, which coverage notices and now wants to test. In addition to the specific excludes we already had in place (`.packages`, etc), I modified .coveragerc to only require coverage on files within the `urllib3` directory, of which category `http.cookiejar` is not an example.

urllib3/util/retry.py
Lukasa: Hrm. Well, we'll leave it then. =)

urllib3/util/retry.py
Lukasa: I wouldn't use `_encode_url_methods`, that just defines methods with bodies. The definition for redirects is exclusive (i.e. everything but HEAD gets changed).

And no, you shouldn't be using `REDIRECT_STATUSES`, that would also be wrong. This is somewhat more complex than you'd expect. See [`requests.sessions.SessionRedirectMixin.rebuild_method`](https://github.com/kennethreitz/requests/blob/aa1c3adf9a2e5dcccd9c4d53b8d10c9dc4348f2f/requests/sessions.py#L249-L269) for the logic requests uses.

urllib3/sessionmanager.py
Lukasa: @shazow can disambiguate here.

urllib3/request.py
Lukasa: My question isn't why we're yanking it out, it's why are we saving it?

haikuginger: Force of habit, I guess. Given that this part of urllib3 isn't a public API, we can certainly not save it, and add it back later if we have a need.

urllib3/sessionmanager.py
shazow: There are a few places where we distinguish between `None` and `False` as specific behaviour. If we're just checking presence, a `if 'proxy_url` in pm_kwargs:` should be sufficient.

That said, I would prefer to avoid this flow altogether and instead pass in a pre-made Manager to the constructor. We can add classmethod helpers as shortcuts for creating SessionManager objects with uninitialized Manager components later.

shazow:
Haven't had a full look at this, thank you @Lukasa for providing the first review! I'll take a look in a day or two. Very excited about this, @haikuginger. :)

urllib3/util/retry.py
haikuginger: All right. I'm going to spend some quality time with RFC-7231 and a few different browsers and come up with something.

urllib3/sessionmanager.py
haikuginger: > That said, I would prefer to avoid this flow altogether and instead pass in a pre-made Manager to the constructor. 

Do you mind walking me through the thinking here? Writing this, I sort of was seeing `SessionManager` as abstracting the `PoolManager` and `ProxyManager` in the same way those classes abstract the `ConnectionPool`, so I want to make sure that my thinking lines up with yours when revisiting it.

shazow: Primarily because the coupling in PoolManager was unavoidable since it needs to create new instances of ConnectionPool per host. Whereas SessionManager is 1:1 so we can avoid it and it should make the code cleaner by avoiding switches based on input conditions. For example, there may come a point when PoolManager and ProxyManager or OtherFutureManager have the same kw param with conflicting meanings, and then it will be impossible to create a SessionManager with the appropriate type of internal manager.

I'm all for shortcuts (we have `connection_from_*` and similar all over the place for this), but I feel like constructors should be as decoupled as possible to allow better substitution.

Also, there's no reason a user shouldn't be allowed to provide their own `CustomPoolManager` as the driver for the `SessionManager`.

haikuginger: Makes sense! Any thoughts on the classmethod verbiage? I was thinking something along the lines of `SessionManager.with_pool_manager(*args, **kwargs)`.

Alternatively, would you be open to optionally taking an existing *Manager, and if none was received, initialize a standard `PoolManager` with the received args? That way, we'd get one-step init for the common use case, and would also have a canonical way to initialize with an alternative of any appropriate type. It could look something like this:

    class SessionManager(object):

        def __init__(self, manager=None, headers=None, cookie_jar=None, manager_class=PoolManager, **manager_kw):
            if manager:
                self.manager = manager
            else:
                self.manager = manager_class(**manager_kw)
            # Do other stuff

Something like that could be subclassed relatively simply to use any variety of manager:

    class ProxySessionManager(SessionManager):

        def __init__(self, **kw):
            kw['manager_class'] = ProxyManager
            super(ProxySessionManager, self).__init__(**kw)

shazow: I was thinking:

```python
class SessionManager(object):
    def __init__(self, manager, cookie_jar):
        ...
```

Nice and simple. :)

haikuginger: All right! I'll get on that and the class method(s) to go with it.

shazow: Don't worry about the helper class methods to start with, those are things we can always add later. Let's focus on the core functionality first.

haikuginger:
Okay, I've made most of the above changes. In addition, as there weren't any out-and-out objections to the idea, I've moved redirection handling to the `RequestMethods` class; this lets us do `self.redirect` easily to keep the redirection in the same scope. This also establishes the contract that the scope calling `.redirect` will be doing so from a context that has a `urlopen` method.

urllib3/poolmanager.py
haikuginger: I decided to leave `retries` in the signature for `PoolManager.urlopen`; if we let it filter into `**kw`, we'd have to pop it anyway, as .redirect requires a true `Retries` object to be passed to it, and the one in `**kw` might be an int. In the previous version, this didn't matter, as the redirect logic wasn't a separate function we were passing `**kw` to.

haikuginger:
Bump! 🍪

shazow:
Thanks for the bump. :) I'm gone for a few days but can take a look on Monday. @sigmavirus24 Any thoughts?

urllib3/connectionpool.py
```diff
@@ -651,28 +651,13 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
-
-            log.info("Redirecting %s -> %s", url, redirect_location)
-            return self.urlopen(
-                method, redirect_location, body, headers,
-                retries=retries, redirect=redirect,
-                assert_same_host=assert_same_host,
-                timeout=timeout, pool_timeout=pool_timeout,
-                release_conn=release_conn, **response_kw)
+        if redirect and response.get_redirect_location():
+            return self.redirect(
```
shazow: There's no `self.redirect` on this class though, right?

shazow:
As it is now, using an `HTTPConnectionPool` directly with `urlopen` will explode if redirects are enabled, right? That seems like a problem to me. Maybe urlopen needs to forget about redirects entirely (so that it's only handled in RequestMethods)? A painful break in compatibility but probably makes sense?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,87 @@
...
+        Additional paramaters to be passed to the encapsulated
+        :class:`.PoolManager` or :class:`.ProxyManager` instance.
+    """
+    def __init__(self, manager, cookie_jar=None, headers=None):
+        super(SessionManager, self).__init__(headers=headers)
+        self.manager = manager
+        self.context = Context(cookie_jar)
+
+    @classmethod
+    def with_defaults(cls, manager=None, cookie_jar=None, headers=None, **kw):
```
shazow: I'd cut this helper for now, maybe add it back later after we release if it's needed. Really we're talking about:

```python
SessionManager.with_defaults(somedefault=42)
# vs
SessionManager(PoolManager(somedefault=42))
```

The helper is not worth it yet, IMO.

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,87 @@
...
+        manager_class = kw.pop('manager_class', PoolManager)
+        manager = manager or manager_class(**kw)
+        return cls(manager, cookie_jar, headers=headers)
+
+    def set_cookie_policy(self, policy):
```
shazow: Are these one-liner context wrappers worth it vs calling them directly on the context property?

urllib3/util/retry.py
```diff
@@ -290,5 +289,17 @@ def __repr__(self):
                     cls=type(self), self=self)
 
 
+def redirect_method(method, status):
```
shazow: Would it make sense to have this live as a method/staticmethod/classmethod under the `Retry` class? This way people could override it with their own behaviour per-request.

shazow:
I really like this change, I feel it opens a lot of doors for improvements in urllib3. Thank you so much, @haikuginger.

A few moderate design-related change requests, but nothing huge.

I'd love to see some more lower-level tests (aka outside of with_dummyserver) that verify the new cookie-related behaviour/API promises we're introducing, such as making a fake Request and fake Response and validating that the Context object deals them it appropriately when applied.

Also I really want to see some feedback from @Lukasa @sigmavirus24 before we move forward since this is an important change.

urllib3/poolmanager.py
shazow: FWIW I think this is fine either way. :)

urllib3/util/context.py
```diff
@@ -0,0 +1,87 @@
...
+    Extensible class encapsulated by :class:`.SessionManager`; currently
+    used to manage cookies.
+
+    :param cookie_jar:
+        Used to pass a prebuilt :class:`http.cookiejar.CookieJar` into the
+        context to be used instead of an empty jar.
+    """
+
+    def __init__(self, cookie_jar=None):
+        # We unfortunately have to do it this way; empty cookie jars
```
krishnamannem: Not a pro at python but can we not minimize this by

```
def __init__(self, cookie_jar=CookieJar()):
      self.cookie_jar = cookie_jar
```

Lukasa: No. Default values in Python are evaluated and assigned once and re-used every time the method is called. That would change the behaviour here. =(

shazow: That would make every instance of Context use the same instance of CookieJar, and you'd get conflicting cookie states across requests. :)

See: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

urllib3/connectionpool.py
haikuginger: Not on `ConnectionPool`, but it inherits from `RequestMethods`, where redirect logic has been added. Basically, `.redirect()` is now a multilevel convenience method like `.request()` or `.request_encode_body()`. It should work fine, just like when the logic was embedded in the method, but can be accessed by any subclass of `RequestMethods`.

shazow: Oh woops, forgot that `HTTPConnectionPool` also inherits from `RequestMethods`, my bad.

urllib3/sessionmanager.py
haikuginger: Fair enough!

urllib3/sessionmanager.py
haikuginger: I think the big question is, "will there ever be a situation in which a SessionManager might need to choose between contexts to store information in?"

Right now, with only cookies being handled, we don't have a good grasp on what might happen next; the extraction and application is being handled wholly by the `Context`, but with n=1, it's hard to extrapolate. If we remove the convenience methods, then we're essentially making an API commitment of "There will always be one `Context` per SessionManager, which will do all processing of requests and responses."

The secondary, much less compelling, thing to think about is that by removing the wrapper methods, we have to go down to the Context level elsewhere, making for a slightly less friendly API.

urllib3/util/retry.py
haikuginger: I'm of a few minds on this.

Overriding could be a nice feature, and if we want to do that, then the Retry object (I think as an instance method) would be the place to do it. But, I'm not sure how much actual usage that'd get - does urllib3 have previous issues requesting that feature?

My preferred place would be to put it directly on the `Response` object, but I'm not sure if we can do that without patching into httplib private attributes on the response or passing the request method in anyway.

haikuginger:
>Maybe worth considering the following though?] Maybe urlopen needs to forget about redirects entirely (so that it's only handled in RequestMethods)? A painful break in compatibility but probably makes sense?

@shazow: I think that makes sense, given that we're moving the redirect handler onto RequestMethods. One caveat there is that, as far as I can tell, we'll only be able to send the instantiation of the `RequestMethods` object as `_pool` to `retries.increment()`, whereas currently we're able to send the `ConnectionPool` when using the PoolManager (as the SessionManager uses `PoolManager.urlopen()`, it already has this behavior).

test/test_sessionmanager.py
```diff
@@ -0,0 +1,32 @@
...
+class TestSessionManager(unittest.TestCase):
+
+    def setUp(self):
+        self.manager = SessionManager.with_defaults()
+
+    def test_set_cookie_policy(self):
+        policy = cookielib.DefaultCookiePolicy()
+        self.assertNotEqual(policy, self.manager.context.cookie_jar._policy)
+        self.manager.set_cookie_policy(policy)
+        self.assertEqual(policy, self.manager.context.cookie_jar._policy)
```
Lukasa: Should this be checking equality or identity?

test/test_sessionmanager.py
```diff
@@ -0,0 +1,32 @@
...
+        self.assertTrue(isinstance(manager.manager, ProxyManager))
+
+    def test_creates_pool_manager(self):
+        self.assertTrue(isinstance(self.manager.manager, PoolManager))
+
+    def test_with_external_jar(self):
+        this_policy = cookielib.DefaultCookiePolicy()
+        jar = CookieJar(policy=this_policy)
+        manager = SessionManager.with_defaults(cookie_jar=jar)
+        self.assertEqual(manager.context.cookie_jar, jar)
```
Lukasa: Same here, should this be identity or equality?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,87 @@
...
+        if not provided, a new one will be created.
+
+    :param \**pm_kwargs:
+        Additional paramaters to be passed to the encapsulated
+        :class:`.PoolManager` or :class:`.ProxyManager` instance.
+    """
+    def __init__(self, manager, cookie_jar=None, headers=None):
+        super(SessionManager, self).__init__(headers=headers)
+        self.manager = manager
+        self.context = Context(cookie_jar)
```
Lukasa: The `manager` kwarg isn't documented here.

Lukasa:
I've left some minor notes, but otherwise I really like this: great work @haikuginger!

test/test_sessionmanager.py
haikuginger: Will change to identity.

test/test_sessionmanager.py
haikuginger: See above.

urllib3/sessionmanager.py
haikuginger: I'll add documentation around it; I didn't update the docs after switching SessionManager to receive a PoolManager in init.

haikuginger:
Okay, I've made all the suggested non-design-related changes, added some testing, and made some low-level changes (non-API-breaking) in support of same.

Things that remain open for discussion:

- [ ] Discuss removing redirect handling from `urlopen()` across the board and make it the responsibility of `RequestMethods.request()`
- [ ] Pick a location for `redirect_method()`

Other than that, I think, we'll need @sigmavirus24 to lend his thoughts.

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,81 @@
...
+        self.context.apply_to(request)
+
+    def extract_context(self, response, request):
+        """
+        Take the given response and the request that produced it, and store
+        any context it produced for later retrieval.
+        """
+        self.context.extract_from(response, request)
+
+    def urlopen(self, method, url, redirect=True, retries=None, **kw):
```
shazow: I suggest adding a note here that it's just here for backwards-compatibility with urllib2-style code and that it's a lower-level method that should be avoided in favour of using .request(...)

haikuginger: If we're doing that on `SessionManager`, do you want similar notations on `PoolManager.urlopen()` and `ConnectionPool.urlopen()`?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,81 @@
...
+from .request import RequestMethods, Request
+from .util.context import Context
+
+
+class SessionManager(RequestMethods):
+    """
+    Allows arbitrary requests while maintaining session context across
+    those requests. Currently, that context consists of automatic
+    cookie storage and retrieval.
+
```
shazow: A basic usage example would go really well here for our docs.

urllib3/util/retry.py
shazow: I vote for the Retry object just because it's easiest to replace/override/pass in.

urllib3/sessionmanager.py
shazow: Hm yea I think that would help.

I see a lot of people "in the wild" trying to use urlopen by default and I keep having to point them to .request(...), so an extra note would be helpful. :)

shazow:
Also, I noticed I missed this reply earlier from @haikuginger:

> I think the big question is, "will there ever be a situation in which a SessionManager might need to choose between contexts to store information in?"
>
> Right now, with only cookies being handled, we don't have a good grasp on what might happen next; the extraction and application is being handled wholly by the Context, but with n=1, it's hard to extrapolate. If we remove the convenience methods, then we're essentially making an API commitment of "There will always be one Context per SessionManager, which will do all processing of requests and responses."
> 
> The secondary, much less compelling, thing to think about is that by removing the wrapper methods, we have to go down to the Context level elsewhere, making for a slightly less friendly API.

Is it true that we're making the commitment that "there will always be one Context"? I mean, the helpers as they are now are only being used within the SessionManager itself, right? That means the API is self-contained, so really by having the extra public helpers we're making a stronger API commitment vs not having them now and adding them later.

Specifically things like `apply_context` and `extract_context`.

For things like `set_cookie_policy`, should we even have this? Would it be better to have the user just pass in the Context (with the policy baked in) in the constructor instead of passing in the `cookie_jar` and baking the policy on their behalf?

I think that substantially improves configurability and reduces our API commitment and removes a bunch of extra logic. :)

haikuginger:
After giving it some more thought, I can't think of any use case for multiple Context objects that wouldn't be better handled by having a smarter Context (e.g., a setting that applies basic auth on a particular domain). So I'm good removing those accessor methods.

I think passing an optional `Context` into the SessionManager constructor does make sense, and definitely cleans up SM's API. If we're doing that, I think I'd like to do some work first to make `Context` more extensible and useful from a user-facing perspective; add a way to attach extraction/application methods, for example, hammer out a clear API for such attachments, and possibly add an example.

shazow:
@haikuginger Sounds good. Overall, I'd like to avoid as much redundancy as we can in the first version of the SessionManager. We can always add more helpers later, but removing them later is really difficult. I rather the base API be a little more explicit/onerous to start with and make it easier later than overcommit from the get-go.

haikuginger:
Makes sense; I'll strip it down, and start planning the layout of an extension system as a separate PR for sometime after the initial release.

urllib3/sessionmanager.py
haikuginger: All right; should we treat this note as the first step in moving redirect (and other) logic into `RequestMethods.request()`, or do you still want to think about doing that inside this PR?

My thought on that would be that this can serve as a notice that we're going to begin removing "smart" features like redirect from `urlopen`, and we can actually implement that stripping in the next major revision.

urllib3/util/retry.py
haikuginger: @Lukasa, @sigmavirus24, thoughts before I make that change?

Lukasa: I'm with @shazow.

urllib3/sessionmanager.py
shazow: Yea, let's worry about dumbing urlopen down later.

Not sure we need to be too explicit about that in the note, just saying that this is a lower-level method and that people should use the `Requests.request(...)` helper instead (with appropriate sphynx linking syntax) would be ideal.

urllib3/util/retry.py
haikuginger: 👍

haikuginger:
Okay, new changes!

I believe I've implemented all of the above. I've also substantially cut back the `Context` API and laid some basic groundwork for later extensibility.

urllib3/__init__.py
```diff
@@ -18,7 +18,8 @@
 from .util.url import get_host
 from .util.timeout import Timeout
 from .util.retry import Retry
-
+from .util.context import CookieJar, DefaultCookiePolicy, Context
```
shazow: I say we skip importing `CookieJar` and `DefaultCookiePolicy` into the root urllib3 namespace now that `Context` works by default without specifying them.

Let's add an example in the docs for how to do that, though (provide your own CookieJar and policy).

shazow:
@haikuginger Looks great! I suggest trimming down the root namespace imports a bit too and we should be good to go.

Also quick thought: Are we happy with `Context` as a name? would `SessionContext` be clearer? (I'm on the fence either way.)

urllib3/util/context.py
```diff
@@ -0,0 +1,89 @@
...
+            self.cookie_jar = cookie_jar
+        else:
+            self.cookie_jar = CookieJar()
+
+        # TODO: Right now, extract/apply methods are static, but in future,
+        # we'll support extending them at runtime. This is the beginning
+        # framework for that pivot. This behavior is subject to change, so
+        # don't base anything off it until there's an API.
+
+        self.extract_methods = (
```
shazow: IMO this is over-engineering. If we don't want people to depend on this, then we shouldn't provide it. :) Let's just hard-code the one line into the relevant apply_to/extract_from.

haikuginger:
I'd be happy with `SessionContext` as a name, especially given the point about `SSLContext`, and I'll make the other changes as well once we've got another +1 on `SessionContext`.

haikuginger:
Given no objection on `SessionContext`, all the above changes have been made.

docs/index.rst
```diff
@@ -142,10 +142,40 @@ sanity. To achieve this, the codebase is a collection of small reusable
 utilities and abstractions composed together in a few helpful layers.
 
 
+SessionManager
+--------------
+
+:doc:`SessionManager(...) <manager>` is currently urllib3's most abstract manager
```
shazow: Great first take, thank you! I might edit some of this later, hope you don't mind. :)

docs/index.rst
```diff
@@ -142,10 +142,40 @@ sanity. To achieve this, the codebase is a collection of small reusable
...
+    >>> cj = urllib3.util.context.CookieJar(policy=liberal_policy)
+    >>> context = urllib3.SessionContext(cookie_jar=cj)
+    >>> session = urllib3.SessionManager(urllib3.PoolManager(), context=context)
+
+
 PoolManager
 -----------
 
-The highest level is the :doc:`PoolManager(...) <managers>`.
+The highest HTTP level is the :doc:`PoolManager(...) <managers>`.
```
shazow: Why this change? It does HTTPS too. :) (People might be confused)

Could just get rid of this line altogether.

urllib3/sessionmanager.py
```diff
@@ -1,6 +1,6 @@
 from .util.retry import Retry
 from .request import RequestMethods, Request
-from .util.context import Context
+from .util.context import SessionContext
```
shazow: Should this be `util.session` or `util.sessioncontext` instead?

shazow:
Looks great, only a couple of minor non-blocking notes.

I'm going to let this stew for a few days in case anyone else has more notes to add, but otherwise I'm good to merge. Another +1 would be nice. :)

docs/index.rst
haikuginger: I wanted to make a shift in the documentation so it's clear that while PoolManager handles HTTP (or HTTPS) connections at a higher level than any other object, it's not necessarily the "highest-level" manager, now that we've got SessionManager coming in.

Wording definitely could use some work; I agree.

docs/index.rst
haikuginger: Not at all! It'll take time for me to figure out the "urllib3 writing style".

shazow: I really like having a first take on it because then my job becomes coming in and simplifying it rather than trying to come up with something from scratch. :)

urllib3/sessionmanager.py
haikuginger: Agreed. Will push that change briefly.

urllib3/connectionpool.py
```diff
@@ -448,7 +448,8 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
         """
         Get a connection from the pool and perform an HTTP request. This is the
         lowest level call for making a request, so you'll need to specify all
-        the raw details.
+        the raw details. To make HTTP calls, use :func:`urllib3.request.RequestMethods.request`
+        instead.
```
Lukasa: This reads weirdly to me. Maybe: "In general `request` should be preferred to this method."?

urllib3/poolmanager.py
Lukasa: I think it's fine too, just wanted to check we did it on purpose and that it wasn't left in from a previous round.

urllib3/sessionmanager.py
Lukasa: Bump.

Lukasa:
+1.

haikuginger:
This is a fairly large API addition and not-insubstantial backend change; should we look at cutting a 1.16 or 1.15.2 release before it gets merged?

shazow:
Good point! I'll try to cut a release soon.

shazow:
Although there is no compatibility breakage here so it'd be reasonably safe to include in v1.16 too. (Don't think we're going to do a v1.15.2).

haikuginger:
Makes sense. @Lukasa, @sigmavirus24, thoughts?

Lukasa:
I'm inclined to want to have this be a feature in a release more or less to itself. That'll let us consider it a big marquee feature and will also ensure that, if we've broken anything (god forbid, but it does happen), users will have another release they can use that is basically identical *sans* this feature, giving us time to fix things right rather than work under pressure to fix them fast.

haikuginger:
I agree with @Lukasa- the last thing we want is a `discoveryd`-style debacle without a clear interim solution. I'm unsure on what a concurrent-release version numbering scheme ought to look like- just 1.16 and 1.17?

shazow:
> I'm inclined to want to have this be a feature in a release more or less to itself.

Okely. v1.17 it is then, I'll probably cu tit around the same time as I cut v1.16 so people can roll back if they want. Though the code is fairly isolated, so I'm not too worried.

haikuginger:
@shazow - maybe just reserve the 1.16 version number, split a `1.16-release` branch off master before merging, and release 1.17? Then we can release 1.16 and unpublish 1.17 if it seems necessary.

Lukasa:
I don't see any reason not to just release the two together. I agree with @shazow, I don't think we're at any risk of having real problems, but at the same time: version numbers aren't exactly a finite quantity, so it doesn't hurt us to use one up to be on the safe side. :smile:

haikuginger:
Sounds good, then!

Lukasa:
Oh, and we should absolutely not unpublish a release. Good or bad, releases should stay forever. =)

haikuginger:
I figure I'll rebase this once #835 gets merged in, just in case of conflicts - make sense to everyone?

Lukasa:
That sounds good to me: it'll maybe give @shazow some impetus to actually merge #835. ;)

haikuginger:
I've rebased onto master, so this should be ready to be pulled in - taking a look at #835, I don't *think* there are any conflicts to worry about there.

shazow:
You're totes right, I'm just trying to decide whether we should continue with the v1.17 stand-alone release strategy or maybe start merging this into a persistent branch until we're ready to do one big release?

The stand-alone release thing doesn't feel quite right, especially since this functionality is still in flux and will continue breaking (when we add more autoconfig stuff like ssl, appengine, etc).

haikuginger:
I've added some stuff in support of #892. This has introduced some not-insubstantial architectural changes that could use review.

The big changes are breaking our `Request` object off from subclassing the stdlib's, and creating an architecture for dynamic insertion of different context handlers at runtime. This substantially increases the ease of, e.g., disabling cookie handling on a `SessionManager`.

haikuginger:
Closing in favor of #904, which pushes the approved changes onto their own branch. Will be opening up a new PR on that branch for review with the latest changes.

Issue #847, opened 2016-04-16
Retry backoff behavior inconsistent with documentation


appeltel:
The docstring for urllib3.util.retry.Retry indicates that "If the backoff_factor is 0.1, then :func:`.sleep` will sleep for [0.1s, 0.2s, 0.4s, ...] between retries."

That isn't consistent with the actual behavior on the first retry, as when `_observed_errors == 1` the sleep function returns 0, so there is no wait after the first retry. This behavior is specifically covered by a unit test requiring the first sleep to return 0. According to the actual behavior, the documentation should read "[0.0s, 0.2s, 0.4s, ...] between retries.", and probably also note that there is no wait after the first retry.

I'm not sure what the intended behavior is, or what it should be in principle. My vote would be for a wait after the first retry, and I would be happy to contribute a quick fix in either case.

shazow:
+1 to updating the docs, though I'd vote for keeping no-sleep for the immediate first retry. Vast vast majority of errors are resolved immediately by a second retry without any sleep, so I feel introducing an immediate sleep is more harm than good. Second try sleep makes sense to me.

PR #851, opened 2016-04-30
Correct docstring for Retry backoff factor

appeltel:
Correct the docstring for the backoff_factor argument to the Retry object, to reflect the fact that a wait is not applied before the second try, as per issue #847


shazow:
Thank you very much! :)

PR #853, opened 2016-05-03
Improve doc around Retry.status_forcelist and Retry.method_whitelist

rbcarson:
I got confused by some ambiguities in the doc for the Retry class, so I propose these improvements. I've backed it up with a few more tests.


shazow:
This is great, thank you! :custard:

PR #856, opened 2016-05-04
HTML5 non-ASCII File Names by Default

Spryttan:
These changes address one point of shazow/urllib3#303: encoding non-ASCII file names according to the HTML5 working draft by default, while still allowing file names to be encoded according to RFC 2231.


urllib3/fields.py
```diff
@@ -31,45 +31,93 @@ def format_header_param(name, value):
...
+        result = u'%s="%s"' % (name, value)
         try:
             result.encode('ascii')
         except (UnicodeEncodeError, UnicodeDecodeError):
             pass
         else:
             return result
-    if not six.PY3 and isinstance(value, six.text_type):  # Python 2:
+
+    if not six.PY3:  # Python 2:
```
Lukasa: Why has the `isinstance` check been removed here?

Lukasa:
Thanks @Spryttan!

I have three initial questions before delving into a big code review. Firstly: why the change to HTML5-by-default? That change is going to be frustrating for consumers like requests which need to defend against a sudden and somewhat subtle change that will not be clearly exposed but will affect the wire-form of their requests.

Secondly, why have you changed the return value of `format_header_param_*` to be unicode in all cases, rather than native strings?

Thirdly, it doesn't seem like the output of the encoder matches [the spec](https://www.w3.org/TR/html5/forms.html#multipart-form-data). My reading of the spec is that you encode a unicode code point (e.g. U+1234) to `&#1234;`. Your code doesn't do that. What have I failed to understand?

urllib3/fields.py
Spryttan: It was unnecessary, as the method doc requires that `value` be a unicode string. If we did allow non-unicode-data through here, we would need the encoding also, so it can be passed to `encode_rfc2231`. Right now, the formatted result is hardcoded to display `utf-8` as the encoding.

Spryttan:
Hi @Lukasa ,

I think there are two reasons to support HTML5 by default: The spec you linked to forbids the RFC2231 encoding: "User agents must not use the RFC 2231 encoding suggested by RFC 2388", and the HTML5-way seems to be more widely supported - but I have no data on that, just personal experience.

`format_header_param` used to return a `str` (bytes string) in Python 2 and a `str` (unicode string) in Python 3 - this was confusing. Also, it is easiest to manipulate string data if we keep it unicode all the time, and don't encode until as late as possible, when the request body is actually being built.

The part of the spec you linked to is referring to how to encode form data that is not expressible in the selected character set, which is a separate process from formatting the file name. The urllib3 "selected character coding" is hardcoded to utf-8 in filepost.py, line 11 `writer = codecs.lookup('utf-8')[3]`, and utf-8 can express all characters, so we should never see anything like  `&#1234;`.

Lukasa:
> The part of the spec you linked to is referring to how to encode form data that is not expressible in the selected character set, which is a separate process from formatting the file name. 

So, just to clarify, the net effect is that the 'encoding algorithm' used in HTML5 is just...send UTF-8?

Spryttan:
Yes, given that the selected character encoding is utf-8. And the file name just takes the same encoding as the selected character encoding, although there is leeway to approximate the precise name "if necessary".

urllib3/fields.py
```diff
@@ -19,7 +19,7 @@ def guess_content_type(filename, default='application/octet-stream'):
     return default
 
 
-def format_header_param(name, value):
```
shazow: Should we keep an alias in case someone is depending on the old name?

shazow: Ah we do, nm,

urllib3/fields.py
```diff
@@ -31,45 +31,93 @@ def format_header_param(name, value):
...
+    value = value.replace(u'\r', u' ').replace(u'\n', u' ')
+
+    return u'%s="%s"' % (name, value)
+
+
+format_header_param = format_header_param_rfc2231  # For backwards-compatibility
+
+
 class RequestField(object):
+    DEFAULT_FILENAME_ENCODING_STYLE = 'HTML5'
```
shazow: Ultimately this translates to which callable we use, right? How about just specifying the callable rather than mapping to it later from a string?

shazow: That is, something like:

```python
class RequestField(object):
    header_formatter = format_header_param_html5
```

I don't think this is the kind of thing that will be changing from request to request, so it's probably safe avoiding passing it in the constructors? Just let people override the static configuration-style variable in the class if they want to change the behaviour.

Lukasa:
@Spryttan So, just to be clear. The HTML 5 specification here can be summed up as: just encode the text don't worry about it? Wow, ok then.

I'm definitely nervous about swapping our default behaviour here, but @shazow can express a preference.

urllib3/fields.py
Spryttan: Just specifying the callable is a good idea.

`header_formatter` could change between requests, if you had to post to two different servers, which support different header param specs, within the same program. But, because this is probably an uncommon case, and in Python you can still set the `header_formatter` value per-instance anyways, I am not opposed to this static-config style.

Spryttan:
@Lukasa Yes.

urllib3/fields.py
shazow: Yea, I mean it's not common for it to be changed between requests so I rather make that logic explicit and overridable. This way people can bring their own formatter and we don't have to worry about setting up a lookup registry etc.

Spryttan:
About the failing build: `nose.proxy.OSError: [Errno 98] Address already in use` Is this just a transient error?

shazow:
> About the failing build: nose.proxy.OSError: [Errno 98] Address already in use Is this just a transient error?

Should be, restarted the build.

Lukasa:
@shazow Before merging, can we consider whether HTML5 or RFC 2231 should be the default?

sigmavirus24:
I'd also like a little time to research the relevant RFCs. There's a more recent RFC that makes reference to the HTML5 spec but updates the multipart/form-data spec. I think I passed it over to @Lukasa too.

shazow:
Sounds good.

It's probably the prudent thing to add a transition period with a release warning but if HTML5 turns out to be more compliant these days then I'm fine with just switching to it and seeing what happens. :)

singalen:
Came here from https://github.com/bottlepy/bottle/issues/852.
:~(

theacodes:
@singalen happy to accept this PR if you want to make it current!

Robbt:
So I managed to get this working on a personal fork and it solved a major headache I was having trying to upload files with utf8 filenames via POST to an outdated PHP server. I'm not sure if I should just open up a new PR with the merge I did, I didn't change much just updated a few config issues in the unit tests. But it did seem to work for me and I've noticed a lot of people are running into this problem with urllib3 and requests. I'd be happy to adopt the work that @Spryttan did or share my work back to his original PR. 

Here is a link to the branch where I did the merge to make it current https://github.com/robbt/urllib3/tree/Spryttan-master 

Let me know how to proceed, if I should open up a new PR or get this one re-opened.

theacodes:
I'd definitely be happy to review a PR for this against the current head.
Thank you!

PR #858, opened 2016-05-09
update timeout warning and default value on App Engine

snarfed:
motivated by sigmavirus24/requests-toolbelt#146. cc @jonparrott.

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -224,10 +224,11 @@ class AppEngineManager(RequestMethods):
         if timeout is Timeout.DEFAULT_TIMEOUT:
             return None  # Defer to URLFetch's default.
         if isinstance(timeout, Timeout):
-            if timeout._read is not timeout._connect:
+            if timeout._read is not None or timeout._connect is not None:
                 warnings.warn(
                     "URLFetch does not support granular timeout settings, "
-                    "reverting to total timeout.", AppEnginePlatformWarning)
+                    "reverting to total or default URLFetch timeout.",
+                    AppEnginePlatformWarning)
             return timeout.total
         return timeout
 
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -222,7 +222,7 @@ class AppEngineManager(RequestMethods):
 
     def _get_absolute_timeout(self, timeout):
         if timeout is Timeout.DEFAULT_TIMEOUT:
-            return 5  # 5s is the default timeout for URLFetch.
+            return None  # Defer to URLFetch's default.
         if isinstance(timeout, Timeout):
             if timeout._read is not timeout._connect:
                 warnings.warn(


snarfed:
CI failed on py26 only, and the failure looks like an unrelated...config issue? i don't see it in other recent builds though. hrmph.

```
ERROR: test suite for <class 'test.with_dummyserver.test_proxy_poolmanager.TestIPv6HTTPProxyManager'>

  File "/home/travis/build/shazow/urllib3/.tox/py26/lib/python2.6/site-packages/nose/suite.py", line 209, in run

    self.setUp()

  File "/home/travis/build/shazow/urllib3/.tox/py26/lib/python2.6/site-packages/nose/suite.py", line 292, in setUp

    self.setupContext(ancestor)

  File "/home/travis/build/shazow/urllib3/.tox/py26/lib/python2.6/site-packages/nose/suite.py", line 315, in setupContext

    try_run(context, names)

  File "/home/travis/build/shazow/urllib3/.tox/py26/lib/python2.6/site-packages/nose/util.py", line 471, in try_run

    return func()

  File "/home/travis/build/shazow/urllib3/dummyserver/testcase.py", line 164, in setUpClass

    app, cls.io_loop, cls.https_certs, 'https', cls.http_host)

  File "/home/travis/build/shazow/urllib3/dummyserver/server.py", line 213, in run_tornado_app

    sockets = bind_sockets(None, address=host)

  File "/home/travis/build/shazow/urllib3/dummyserver/server.py", line 199, in bind_sockets

    sock.bind(sockaddr)

  File "<string>", line 1, in bind

error: [Errno 98] Address already in use
```

Lukasa:
That test fails intermittently, I'll restart it.

urllib3/contrib/appengine.py
```diff
@@ -189,6 +189,10 @@ def _get_absolute_timeout(self, timeout):
                 warnings.warn(
                     "URLFetch does not support granular timeout settings, "
                     "reverting to total timeout.", AppEnginePlatformWarning)
+            if not timeout.total:
```
Lukasa: Should this be `if timeout.total is None`? It's not totally beyond possibility that someone will set the timeout here to `0`, or `0.0`, or something of that kind.

snarfed: sure! timeout 0 is generally either a special case or degenerate, but regardless, agreed. done.

shazow:
What's the net AppEngine effect if there is no total timeout specified? Is timeout disabled altogether? Or does it revert to the default?

jonparrott:
@shazow I don't know. Let me find out.

shazow:
@jonparrott Thanks! Whatever it is, maybe it's best to change the warning message so that it explains what will actually happen instead. :)

snarfed:
if timout total is unset, the net effect on app engine is to revert to urlfetch's default of 5s, https://cloud.google.com/appengine/docs/python/outbound-requests#request_timeouts . (that's what originally motivated me to dig into this and sigmavirus24/requests-toolbelt#146 originally.)

i've updated the warning message.

jonparrott:
I'll verify that `0` is treated the same as `None` by the urlfetch.

jonparrott:
Verified, `0`, `None`, and `False` as all treated as the default deadline which is 5 seconds unless changed by [`google.appengine.api.urlfetch. set_default_fetch_deadline`](https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch).

jonparrott:
I would prefer if this message matched the other messages in the module by explicitly mentioning URLFetch, e.g.: `URLFetch does not support granular timeout settings and no total timeout was set, so the current default URLFetch timeout of {} will be used.`.

You can get the current timeout values using `google.appengine.api.urlfetch.get_default_fetch_deadline`.

shazow:
It's generally bad form to have dynamic warning messages because then a bunch of warning-specific filtering helpers break (aka you can't set it to "only show warning once" etc).

shazow:
So when I pass in a `Timeout(total=None)`, I'm probably implying that I want no timeout, right? Should the message then be that URLFetch does not support unlimited timeouts?

shazow:
In fact, even using an empty default Timeout configuration (ie. `Timeout()`) implies `total=None`. Do we want this warning to come up anytime someone doesn't set a specific timeout? Seems extreme, no?

jonparrott:
> It's generally bad form to have dynamic warning messages

Good point. I just didn't want it to say `5s` as the user can change it.

snarfed:
@shazow we could warn only if `total` is unset and `_connect` or `_read` are set. that's what happened here, and it does imply that the user wants some sort of timeout, and doesn't understand app engine's constraints.

shazow:
@snarfed That makes sense, but I think that was the purpose of the warning just a few lines above the new one. Maybe we should fix the if-logic there to cover this case instead?

snarfed:
@shazow hmm, i think they're related but not quite the same. eg we want to show this warning even if `_connect` and `_read` are the same, which they were in my case. and it seems like we want to show the existing warning even if `total` is set.

i've updated this PR with that take. you all are the owners though, so i'll happily defer to you, just let me know what you settle on.

shazow:
I don't think it matters if they're the same so much as if they're not-default, right?

I suspect the original logic in that if-statement is incorrect to begin with. @jonparrott Am I mistaken?

jonparrott:
@shazow I've been wrong many times before. :)

I'm not sure my motivation for warning if `_read` and `_connect` aren't the same. IMO, this should *only* accept `total` and warn if `_read` and `_connect` are set to anything.

shazow:
I say we fix the existing if-statement, augment the message a bit, and add a comment for the intention so that it's clear for next time. :)

snarfed:
done! i omitted the comment since i think the code and log message make the intent clear now, but let me know if you disagree.

urllib3/contrib/appengine.py
```diff
@@ -185,10 +185,14 @@ def _get_absolute_timeout(self, timeout):
         if timeout is Timeout.DEFAULT_TIMEOUT:
             return 5  # 5s is the default timeout for URLFetch.
```
jonparrott: So, I know you didn't touch this code, but it seems that we shouldn't hard code this value. It should actually use the aforementioned method to get the real default.

jonparrott:
This LGTM with a tiny comment that can totally be ignored.

urllib3/contrib/appengine.py
snarfed: good call! done, via `None` instead of explicitly getting the default.

urllib3/contrib/appengine.py
```diff
@@ -185,10 +185,14 @@ def _get_absolute_timeout(self, timeout):
...
         if isinstance(timeout, Timeout):
-            if timeout._read is not timeout._connect:
+            if timeout._read is not None or timeout._connect is not None:
                 warnings.warn(
                     "URLFetch does not support granular timeout settings, "
                     "reverting to total timeout.", AppEnginePlatformWarning)
+                if timeout.total is None:
+                    warnings.warn(
```
shazow: I get that these two conditions are different, but is it worth drawing the distinction? I propose one of two options:

1. Keep it as just one message/branch and change it to "... reverting to total timeout (or default if not provided)."
2. If we want to keep two distinctive messages for the two cases, we should probably avoid printing 2 warnings in this scenario and catch that second branch as its own distinct scenario with a message that covers both conditions.

snarfed:
friendly ping! @shazow @jonparrott i'm happy to do whatever you all agree on here, if you're the two stakeholders.

Lukasa:
@snarfed I think @shazow has an outstanding comment on the diff.

snarfed:
@Lukasa true! he and @jonparrott have gone back and forth a bit though. i'd like to wait for them to agree first.

urllib3/contrib/appengine.py
jonparrott: I'm in favor of just one warning concerning timeouts.

jonparrott:
@snarfed if you're still interested in merging this, I'd recommend we go with one warning unless another maintainer disagrees.

jonparrott:
(& sorry I let this slip under my radar for so long).

snarfed:
@jonparrott sure! done.

jonparrott:
This change looks fine to me. @Lukasa do you want to do the final review?

Lukasa:
This looks great to me. It's a bit disappointing that we don't have any tests that rely on this change, but it's also understandable. Tests for this kind of thing would be greatly appreciated if we can figure out a way to test it, but for now we're ok.

Thanks @snarfed! :sparkles: :custard: :cookie:

snarfed:
understood, and agreed. thanks for merging!

jonparrott:
@Lukasa I can maybe look into improving tests for this sometime in the next couple of weeks. Though it's my ultimate goal to remove this module as soon as I can make App Engine less weird.

Lukasa:
@jonparrott That's fine, this is absolutely a best-effort thing.

PR #861, opened 2016-05-12
Hide a confusing "getresponse() got an unexpected keyword argument 'b…

bluetech:
…uffering'" error

Remove the TypeError from the exception chain in Python 3; otherwise it
looks like a programming error was the cause. A quick google search of
the above phrase shows I was not the only one.

Here is how a connection timeout looks like before this patch:

    urllib3.PoolManager().request('GET', 'google.com:81', retries=False, timeout=1)

    File "/home/ran/src/urllib3/urllib3/connectionpool.py", line 385, in _make_request
        httplib_response = conn.getresponse(buffering=True)
    TypeError: getresponse() got an unexpected keyword argument 'buffering'

    File "/home/ran/src/urllib3/urllib3/connectionpool.py", line 387, in _make_request
        httplib_response = conn.getresponse()
    File "/usr/lib/python3.5/http/client.py", line 1174, in getresponse
        response.begin()
    File "/usr/lib/python3.5/http/client.py", line 282, in begin
        version, status, reason = self._read_status()
    File "/usr/lib/python3.5/http/client.py", line 243, in _read_status
        line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
    File "/usr/lib/python3.5/socket.py", line 575, in readinto
        return self._sock.recv_into(b)
    socket.timeout: timed out

    [snip]

After, the first part is gone, and only the real cause is shown.

In Python2, exceptions are not chained automatically so there is no
difference.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -383,8 +383,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         try:
             try:  # Python 2.7, use buffering of HTTP responses
                 httplib_response = conn.getresponse(buffering=True)
-            except TypeError:  # Python 2.6 and older
-                httplib_response = conn.getresponse()
+            except TypeError:  # Python 2.6 and older, Python 3
+                try:
+                    httplib_response = conn.getresponse()
+                except Exception as e:
+                    # Remove the TypeError from the exception chain in Python 3;
+                    # otherwise it looks like a programming error was the cause.
+                    six.raise_from(e, None)
         except (SocketTimeout, BaseSSLError, SocketError) as e:
             self._raise_timeout(err=e, url=url, timeout_value=read_timeout)
             raise


bluetech:
The travis log is a bit hard to interpret, I assume it failed because the coverage for py27 and pypy falls from 100%? That's only natural, this part of the code is never executed in Python 2.7. Maybe someone can suggest how to handle that?

haikuginger:
You can add this to the end of the previous line to exclude coverage on Python 2:

    # Python 3:

bluetech:
@haikuginger huh, I hadn't realized these comments have a special meaning. I will try it, thanks.

haikuginger:
@bluetech: Yeah! Check out .coveragerc for more "special comments".

bluetech:
After two spurious unrelated failures, it now passed with the change @haikuginger suggested.

sigmavirus24:
I'm torn on this. On the one hand, this might help @Lukasa and I with the issue that never dies in requests. On the other, we're making urllib3 uglier and slightly more complex as part of it. Also, people writing code on Python 3 should really not be terribly surprised when seeing tracebacks like this. They're fairly common.

urllib3/connectionpool.py
```diff
@@ -383,8 +383,13 @@ def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
         try:
             try:  # Python 2.7, use buffering of HTTP responses
                 httplib_response = conn.getresponse(buffering=True)
-            except TypeError:  # Python 2.6 and older
-                httplib_response = conn.getresponse()
+            except TypeError:  # Python 2.6 and older, Python 3
+                try:
+                    httplib_response = conn.getresponse()
+                except Exception as e:
```
Lukasa: This is rather overbroad. It would be substantially better if we caught *and verified* that this exception is the given `TypeError` before hiding this part of the traceback.

bluetech: The `e` here is not the `TypeError`. I am not masking `e`, I just make it not chain to the `TypeError` but start a new chain instead.

Lukasa: Oh I apologise, I misread that. Nevermind. =)

Lukasa:
I'm inclined to want to take this: I think it provides a useful benefit at a fairly minor cost.

bluetech:
@sigmavirus24 I'll just note that in my system (using requests on python3) which accesses external services, timeout errors naturally happen a lot, and whenever someone sees a traceback in the logs like the one in the commit message, they immediately think it's a bug in urllib3, and waste some time investigating. So I finally decided to just fix it at the root.

shazow:
Agreed with what @lukasa said, it's a useful benefit and fairly minor cost. Let's do it!

Issue #850, opened 2016-04-21
Should reset connection pools after fork


tesseract2048:
Consider this scenario:

process A creates a connection pool and send a request to `url0`, and then fork twice. Then we got process B & C.
process B and C then send another request to `url0` at nearly the same moment, then the underlying connection they used would be the identical one, causing confused response.

Lukasa:
@tesseract2048 Yes, connection pools should be reset after forking, but I don't believe that urllib3 should be doing it.

This has been discussed elsewhere, but generally speaking we don't know the user's intent: do they want only the child to own the connections, or only the parent? IMO, if a user is going to fork, they should be maintaining discipline over their own state, and should clear the connection pools from whichever process they want to no longer have access to them.

shazow:
Right. I'm open to proposals that can reliably detect and warn (or even reset) when a fork happened, but that's borderline out of scope as urllib3 does not have any multiprocess-safe mechanics (we're merely threadsafe).

@tesseract2048 I'm closing this issue for now, but if you'd like to submit a PR to this effect I welcome it. :)

I would also appreciate any recipes you'd like to add to our documentation about how to use urllib3 properly in a multiprocess environment. This would be ideal, I think.

tesseract2048:
Thanks for response, @Lukasa @shazow .

I think adding description to the effect to document helps little because users may be just simply combining two libraries (i.e. python-requests and multi process thrift server), so connection polling and fork are done separately. Maybe they are not even aware of these implementation details before things actually break (and probably break with little chance and hard to repeat).

What do you say, if letting user specify whether they want to keep connections alive after fork (reset by default)? We could use POSIX calls such as `pthread_atfork` to detect a fork then handle the situation properly.

shazow:
@tesseract2048 Does that work on every platform (win/linux/mac/etc) without edge cases? I suspect it would be safer to provide a multi-process recipe in the docs than try to detect it automagically.

Lukasa:
As far as I know `pthread_atfork` isn't exposed by Python by default, so we'd need to add a dep to make that call.

tesseract2048:
Hi, @shazow @Lukasa . I cannot come up with any edge cases now but there could be many.
Another simpler idea would be associating a `pid` to each connection pool object, each process then checks its owner pid with associated pid (before using) to decide what to do. Does this sound portable?
We can introduce the mechanism and keep current behavior by default, then let other libraries to decide the policy.

shazow:
[...]

PR #862, opened 2016-05-13
late PR for issue #850

tesseract2048:


Lukasa:
Looks good to me!

shazow:
Rather than putting it in the README that is the frontpage of the library, maybe we should move it into a Recipes section in our docs that get rendered here http://urllib3.readthedocs.io/?

Lukasa:
+1 sounds good to me @shazow.

PR #866, opened 2016-05-17
Respect `release_conn=False` after retry

### LINKED ISSUE #651: Likely bug in retry/release-conn code, need tests

jmoldow:
Respect `ConnectionPool.urlopen(release_conn=False)`, making sure not to release the connection, even after a retry.

If a request failed but then succeeded after a retry, the connection would be released, contrary to what the user requested.

Fixes #651.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,9 @@ dev (master)
   
 * Fixed installing ``urllib3[socks]`` extra. (Issue #864)
 
+* Always respect ``ConnectionPool.urlopen(release_conn=False)``, even after a
+  retry. (Issue #651)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,11 +11,14 @@ dev (master)
 
 * Normalize scheme and host to lowercase for pool keys, and include
   ``source_address``. (Issue #830)
-  
+
 * Cleaner exception chain in Python 3 for ``_make_request``.
   (Issue #861)
-  
+
 * Fixed installing ``urllib3[socks]`` extra. (Issue #864)
+
+* Fixed signature of ``ConnectionPool.close`` so it can actually safely be
+  called by subclasses. (Issue #873)
 
 * Always respect ``ConnectionPool.urlopen(release_conn=False)``, even after a
   retry. (Issue #651)
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -168,6 +168,7 @@ In chronological order:
 
 * Jordan Moldow <https://github.com/jmoldow>
   * Fix low-level exceptions leaking from ``HTTPResponse.stream()``.
+  * Bugfix for ``ConnectionPool.urlopen(release_conn=False)``.
 
 * Predrag Gruevski <https://github.com/obi1kenobi>
   * Made cert digest comparison use a constant-time algorithm.
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,3 +7,4 @@ twine==1.5.0
 wheel==0.24.0
 tornado==4.2.1
 PySocks==1.5.6
+pkginfo>=1.0,!=1.3.0
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -1,6 +1,9 @@
 import unittest
+from collections import Iterator
+import itertools
 import socket
 import threading
+import time
 from nose.plugins.skip import SkipTest
 from tornado import ioloop, web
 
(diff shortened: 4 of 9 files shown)


Lukasa:
Thanks for this! This fix seems like a really good one.

However, the size of the diff for the tests is *massive*, and I think that's because you're using the test server to repro this. That's fine, but for a problem like this you may find it takes less code (and is more comprehensible) to write a socket level test. Have you considered doing that?

shazow:
> for a problem like this you may find it takes less code (and is more comprehensible) to write a socket level test. 

+1, I'd like to move as many tests as possible into socket-level tests as they're far less flaky and easier to debug.

jmoldow:
I tried writing a socket-level test, but I couldn't figure it out. Would one of you be able to write the socket-level test for this? @Lukasa @shazow

Lukasa:
I can do if that's something you want me to do. It'd basically just be a variation on `test_connection_closed_on_read_timeout_preload_false`. If you take a look at that test and still don't think you're comfortable writing that test let me know, and I'll write it myself.

jmoldow:
@Lukasa That was the test I was trying to copy. However, there some important differences:

- That test has the timeout on the `read`, but both `urlopen` calls succeed. My test doesn't involve `read` at all, and the timeout(s) are during `urlopen`.
- That test has one unsuccessful call followed by one successful call. I was hoping to have two unsuccessful calls, followed by one successful call. The reason being, I wanted to have one failed `urlopen(retries=0)`, and one successful (but with one retry) `urlopen(retries=1)`. The reason for the former, is to guarantee that the latter is in fact going through the retry logic.
- That test used a `threading.Event` to force the timeout. But we can't do that for the `urlopen(retries=1)`, since there's no straightforward way to set the event between retries.
  - In the testserver test I wrote, I introduced the concept of a server-side delay. So maybe you could do that here too, instead of using an `Event`.
  - I suppose you could write a `Retry` subclass, which sets the event in `Retry.increment`. I suppose you could also use that mechanism to assert that there was in fact a retry.

I tried to resolve those differences, but wasn't able to figure out how to do it (I think I don't fully understand the rules for how the `Response` object is generated from the socket).

jmoldow:
Also, if the size of the test diff is too much, we could remove some superfluous stuff. For example, don't technically need to make any assertions of `pool.num_connections`, since it isn't directly related to the functionality at hand. Similarly, can remove the final part of the tests that call `Response.release_conn()`, since that is testing different behavior. If you want me to remove those, it'd make the tests I wrote simpler to review.

Lukasa:
@jmoldow Ok, I've written what I believe is a correct socketlevel test in lukasa/urllib3@3d4d37e0b2b119766576363fa3f418c7aa959290, in the `issue/866` branch on my fork. Want to give that a try?

jmoldow:
Thanks @Lukasa . I confirmed that your test passes on this branch, and reproduces the issue on master. I've incorporated the test into this pull request, with some changes (such as the name of the test - the old name didn't accurately describe what was being tested).

I've also simplified the other tests I had written. They are now much smaller and more comprehensible.

Lukasa:
Ok, cool, this change looks good. I'm :+1:, but of course I'd like @shazow to review too. Bonus points for reviews from @sigmavirus24 and @haikuginger.

haikuginger:
It looks good to me. Logical change! 👍

shazow:
Do we still need the dummyserver changes/tests?

jmoldow:
I can remove the `test/with_dummyserver/test_connectionpool.py` / `dummyserver/testcase.py` changes if you'd like.

shazow:
If you're comfortable that the same functionality is exercised through socketlevel tests, I think that'd be best. My long-term goal is to get rid of the dummyserver so fewer things to prune later would be nice. :)

jmoldow:
@shazow done

shazow:
Excellent, thank you very much @jmoldow. :)

jmoldow:
You're welcome. Thank you for merging @shazow

shazow:
✨ :custard: ✨

Issue #873, opened 2016-05-24
`connectionpool.ConnectionPool.close()` has zero args


tharvik:
As written [here](https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L93), `close()` do not except any argument. I guess it was meant to have `self` (as overriden in `HTTPConnectionPool`). Or maybe having the `staticmethod` decorator. Or is it some weird corner case to keep compatibility?

Lukasa:
That looks like an error case to me. It doesn't matter, because that method never gets called: it's just used to define the expected interface. But still, it should probably be cleaned up. Want to provide a patch?

shazow:
Fixed now, thank you!

PR #874, opened 2016-05-24
fix connectionpool.ConnectionPool.close()

tharvik:
See #873 for reference.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -90,7 +90,7 @@ class ConnectionPool(object):
         # Return False to re-raise any potential exceptions
         return False
 
-    def close():
+    def close(self):
         """
         Close all pooled connections and disable the pool.
         """


Lukasa:
Hrrrm, those test failures *seem* to be unrelated to this change: something is going on with install. I'll investigate.

Lukasa:
Cool, we bumped into [this](https://bugs.launchpad.net/pkginfo/+bug/1585006). No point piling on, the reports are already there, so let's just work around.

Lukasa:
Fab, so I've proposed a fix that should resolve the busted tests in #875. When that (or something like it) gets merged, you'll need to rebase on top of it to solve the test failures.

Sorry about that inconvenience!

haikuginger:
While we're in here, should we switch from `pass` to a `NotImplemented` exception akin to `RequestMethods.urlopen()`?

tharvik:
> While we're in here, should we switch from pass to a NotImplemented exception akin to RequestMethods.urlopen()?

That would be even nicer, I'll do it when #875 is merged.

Lukasa:
@haikuginger @tharvik I don't actually think we want to do that. I think we shouldn't punish people who unconditionally call `super()` in `close()`, and it gives us the opportunity to be a bit more flexible later.

haikuginger:
I'll defer to @Lukasa's judgement on that, then; I'm 👍 on this change.

tharvik:
Okay, I won't add anything to the PR then.

Lukasa:
All better, mergy mergy mergy. Thanks @tharvik! :sparkles:

sigmavirus24:
@Lukasa you did the merge, don't forget the release notes ;)

haikuginger:
Thanks very much @tharvik! 🍰

PR #876, opened 2016-05-24
``PoolManager`` now references the gobal ``key_fn_by_scheme`` dictionary

jeremycline:
Initially for #830, the ``key_fn_by_scheme`` was copied for each
PoolManager instance. This is inconcistent with the
``pool_classes_by_scheme`` instance variable, which is a reference to
the global dictionary. This commit makes the two consistent by making
the ``key_fn_by_scheme`` instance variable a reference to the global
dictionary.

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -204,10 +204,9 @@ class TestPoolManager(unittest.TestCase):
         self.assertTrue(conn_pool is other_conn_pool)
 
     def test_default_pool_key_funcs_copy(self):
-        """Assert each PoolManager gets a copy of ``pool_keys_by_scheme``."""
+        """Assert each PoolManager gets a reference to ``key_fn_by_scheme``."""
         p = PoolManager()
-        self.assertEqual(p.key_fn_by_scheme, p.key_fn_by_scheme)
-        self.assertFalse(p.key_fn_by_scheme is key_fn_by_scheme)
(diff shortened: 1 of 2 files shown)


Lukasa:
Out of interest, why did you decide to go this way, rather than making `pool_classes_by_scheme` a copy?

jeremycline:
My interpretation of https://github.com/shazow/urllib3/pull/828#issuecomment-221049208 was that it would be nice for things to be different, but it might be a breaking change so waiting for a v2.0 would be better.

Lukasa:
So I think that's true, but I don't think that's a reason to change *this* to the bad case. We did one thing wrongly, but that's not a reason to do it wrongly a second time. =D That's just my opinion though, other contributors may have a different perspective, so let them weigh in too.

jeremycline:
Cool. I'm not sure if it's better to be consistent or to have the ``key_fn_by_scheme`` instance variable behave in a (to me, anyway) more expected way, so I could definitely be talked in either direction. :smile:

shazow:
@jeremycline Heh sorry for the confusing back and forth in the other thread. Altogether I think it's safer to make the breaking change in the original direction than in this direction.

jeremycline:
No problem! Do you want me to reopen the other PR, or just leave it as a wishlist item?

shazow:
I'll re-open the other one. :)

PR #879, opened 2016-05-25
Add ResponseCls class attr to HTTPConnectionPool

jmoldow:
This attribute defaults to `HTTPResponse`, but developers can subclass `HTTPConnectionPool` and substitute their own `HTTPResponse` subclass, as long as it has valid `from_httplib()` and `__init__()` methods, with compatible method signatures.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,9 @@ dev (master)
   
 * Retain ``release_conn` state across retries. (Issue #866)
 
+* Add customizable ``HTTPConnectionPool.ResponseCls``, which defaults to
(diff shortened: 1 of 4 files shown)


Lukasa:
I'm happy enough with this change.

shazow:
Great idea, thank you! Important step in the right direction.

jmoldow:
Can someone with permissions re-try the pypy tests? It looks like it failed a flaky test.

Lukasa:
ACK

shazow:
:shippedit:

haikuginger:
:boom:

Issue #867, opened 2016-05-19
add urllib3 to typeshed


tharvik:
Would that be okay if type annotations of urllib3 were added to the [typeshed repo](https://github.com/python/typeshed)?

It would be nice to have your explicit consent, as stated in [PEP 484](https://www.python.org/dev/peps/pep-0484/#the-typeshed-repo).

haikuginger:
I personally have no objection, although it'd probably be best for us to double-check the annotation when complete.

tharvik:
I'll write some and let you know when I open the PR, thanks!

Lukasa:
I also have no objection, though part of me wonders: if the type checker is now good enough, should we just have stub files here and make them part of our CI?

tharvik:
Function annotation are only available in python 3, you may want to keep compatibility with python 2 then.

Lukasa:
@tharvik That's why I'm suggesting stub files: stub files can be distributed with the library without breaking Python 2.

tharvik:
Ho, misread, yup, that would be better this way, I'll do you a pull request soon then (and by design of typeshed, also copy it there).

Lukasa:
My understanding of typeshed is that actually they'd rather we distributed it than they did. They shouldn't really need a copy if we ship one.

tharvik:
Correct, I'm reading PEP 484 again.

From [it](https://www.python.org/dev/peps/pep-0484/#storing-and-distributing-stub-files), we can either have stubs in `shared/typehints/python*` or alongside the existing python files. What do you prefer?

tharvik:
From what I just saw, it would be easier to split it by python version, because of much conditional imports and behaviour.

Lukasa:
I'm happy to split by Python version, at least in the first instance.

sigmavirus24:
mypy-lang supports using the magical comment annotations, why not use that instead?

sigmavirus24:
> if the type checker is now good enough

@Lukasa mypy-lang is good, yes. But the `typing` module is still as it was when it was introduced in PEP 484 (primitive).

shazow:
(No objections here)

tharvik:
> mypy-lang supports using the magical comment annotations, why not use that instead?

@sigmavirus24 we can also do that, but for now, mypy is still [quite bad](https://github.com/python/mypy/issues/1297) at handling conditional imports, so if we want it to be useful now, it would be better to have splitted stubs.

[...]

PR #880, opened 2016-05-26
Add typing to __init__ and _collections

tharvik:
So, here is a first iteration of adding types checking, as discuted in #867.

Having stubs files was harder to make it works, and in general, it would be best to keep related code and typing. So we will use magic comment as proposed by @sigmavirus24.

There is a bunch of TODO in code, with the corresponding bug report (or general issue).

I'll add some commit comments to explain choices made.

With the current state of mypy, there is a bunch of error, like
```
urllib3/_collections.py:22: error: Name 'OrderedDict' already defined
urllib3/_collections.py:27: error: Name 'RLock' already defined
urllib3/_collections.py:25: error: Name 'RLock' already defined
```
which are caused by mypy not handling conditionals, but, by avoiding theses paths while coding, we can have some type error remarks.

Overview
* add types to `urllib3/__init__.py` and `urllib3/_collections.py`
* adding `urllib3.util.typing` module to allow handling both python2 and python3 with an easier interface. We will add some classes during next iterations
* changing some import check based on `ImportError` to one using `sys.version_info`, as it is the recommanded way to do it in PEP 484


urllib3/__init__.py
```diff
@@ -89,6 +93,8 @@ def add_stderr_logger(level=logging.DEBUG):
 
 
 def disable_warnings(category=exceptions.HTTPWarning):
+    # type: (type) -> None
+    # TODO be more precise about which type is accepted
```
tharvik: I guess it will require either a typeshed or PEP change, `type` is working for now.

urllib3/_collections.py
```diff
@@ -1,21 +1,56 @@
...
+
+from .util.typing import (
+    Generic, TypeVar, List, Union, Tuple, Any,
+)
+from .util.typing import Callable  # noqa: unused in this module
+from .util.typing import Iterable  # noqa: unused in this module
+from .util.typing import Iterator  # noqa: unused in this module
+from .util.typing import MutableMapping  # noqa: unused in this module
+from .util.typing import Optional  # noqa: unused in this module
+from types import TracebackType  # noqa: unused in this module
```
tharvik: flake8 not handling use in comments? did I missed something?

urllib3/_collections.py
```diff
@@ -24,7 +59,7 @@ def __exit__(self, exc_type, exc_value, traceback):
 _Null = object()
 
 
-class RecentlyUsedContainer(MutableMapping):
+class RecentlyUsedContainer(collections.MutableMapping, Generic[_K, _V]):
```
tharvik: We can also inherit from `typing.MutableMapping[_K, _V]`, but it is not working just yet, require some change in the `typing` module.

urllib3/_collections.py
```diff
@@ -38,27 +73,32 @@ class RecentlyUsedContainer(MutableMapping):
...
+    # TODO be more precise about class type
+    ContainerCls = OrderedDict  # type: type
 
     def __init__(self, maxsize=10, dispose_func=None):
+        # type: (int, Optional[Callable[[_V], None]]) -> None
         self._maxsize = maxsize
         self.dispose_func = dispose_func
 
-        self._container = self.ContainerCls()
+        self._container = self.ContainerCls()  # type: OrderedDict[_K, _V]
```
tharvik: Yup, because using `popitem(last=...)` which is only defined by `OrderedDict` (in this case, why having a `ContainerCls`?).

urllib3/_collections.py
```diff
@@ -94,11 +138,12 @@ def clear(self):
                 self.dispose_func(value)
 
     def keys(self):
+        # type: () -> List[_K]
```
tharvik: Should return a valueview, a set like interface, so it is not really a replacement for a dict, maybe an issue here? We could also have it return a set handling duplicated values.

urllib3/_collections.py
```diff
@@ -144,20 +190,26 @@ def __init__(self, headers=None, **kwargs):
             self.extend(kwargs)
 
     def __setitem__(self, key, val):
+        # type: (str, str) -> Union[Tuple[str, str], List[str]]
         self._container[key.lower()] = (key, val)
         return self._container[key.lower()]
```
tharvik: `__setitem__` shouldn't return anything.

urllib3/_collections.py
```diff
@@ -284,38 +350,46 @@ def _copy_from(self, other):
...
+        # type: () -> Iterator[Tuple[str, str]]
         """Iterate over all headers, merging duplicate ones together."""
         for key in self:
             val = self._container[key.lower()]
-            yield val[0], ', '.join(val[1:])
+            # TODO remove ignore when python/mypy#1579
+            yield val[0], ', '.join(val[1:])  # type: ignore
 
     def items(self):
+        # type: () -> List[Tuple[str, str]]
```
tharvik: As `keys` before, It should return a set.

urllib3/_collections.py
```diff
@@ -284,38 +350,46 @@ def _copy_from(self, other):
...
         return list(self.iteritems())
 
     @classmethod
     def from_httplib(cls, message):  # Python 2
```
tharvik: As it is only python2, maybe having a `sys.version_info` check here and only define it for python2?

urllib3/__init__.py
```diff
@@ -22,11 +24,12 @@
...
 import logging
-try:  # Python 2.7+
+if sys.version_info >= (2, 7):
     from logging import NullHandler
-except ImportError:
-    class NullHandler(logging.Handler):
+else:
+    class NullHandler(logging.Handler):  # Python 2.6
         def emit(self, record):
+            # type: (logging.LogRecord) -> None
```
Lukasa: It's really disappointing that we'd have to change away from EAFP to LBYL in order to get the typing module to work properly. Is there really no alternative?

urllib3/_collections.py
Lukasa: Yeah, I don't think flake8 parses comments (and rightly so).

urllib3/__init__.py
tharvik: We can keep it EAFP, that shouldn't mather that much, mypy do not now handle it anyway.

Lukasa:
So right now I have some extremely mixed feelings about this. It seems like it's hurting us quite a lot to do this. =(

tharvik:
Why is it hurting? It doesn't break anything and IMO it is a nice feature for users to have. Also, it keeps compatibilty with python2, only enable checking if the `typing` module indeed exists (so no need for user to install), help devs finding some issues (#874 was found this way but that was more by reading code that mypy), ...

urllib3/_collections.py
shazow: As a rule, we prefer duck-typing over version-checking when possible because it's never that simple (there are tons of implementation of Python and they don't all agree).

shazow:
Now that we moved the type-specific imports into its own module, it's not *too* bad but I'd love to see some return on these annotations. At the very least, this process is helping us audit the code and find some subtle cleanup/improvements, right?

urllib3/util/typing.py
```diff
@@ -0,0 +1,57 @@
+try:
+    from typing import (
```
t-8ch: Will this work on python2? (which uses relative imports by default)

tharvik: The CI do not complain about it, also, there is not `typing` modules in python2 (mypy works from python3.2).

PR #882, opened 2016-05-29
Convert readthedocs link for their .org -> .io migration for hosted projects

adamchainz:
As per their email ‘Changes to project subdomains’:

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.

Test Plan: Manually visited all the links I’ve modified.


haikuginger:
Looks good to me. @shazow, any objections? Seems reasonable to make this change before cutting 1.16/1.17.

Lukasa:
Let's do it.

adamchainz:
You should also update the link on the github project page

haikuginger:
Thanks @adamchainz! If I'm not mistaken, @shazow will have to take care of that last bit. For right now, it hits a 301, so it should be fine.

shazow:
Excellent, thank you!

shazow:
Updated the project description too.

PR #893, opened 2016-06-14
updated SNIMissingWarning to address Jython

owns:
In my development environment, i need to disable SSL.  requests makes that easy by use using verify=False.  However, Jython doesn't allow SNI.  In order to disable it, I wrote the mentioned gist.  I'd be glad to make changes to it if anyone would find it helpful.  It was just a quick fix for me.


Lukasa:
@owns Wait, is it not simply enough to pass `verify=False` on Jython?

Lukasa:
Closing due to inactivity.

Issue #885, opened 2016-06-02
Expand RequestMethods optional arguments


haikuginger:
I've come across at least one example of an API where I need to both POST a body of some form and also have query parameters on a request, and this is something I'm accommodated so far in my [beekeeper](https://github.com/haikuginger/beekeeper) project - largely because I've been handling setting those items up in my own code.

I'm now migrating that to urllib3, and one snag I'm seeing is that `RequestMethods.request()` takes a single `fields` argument that turns into either a querystring or `x-www-urlencoded` depending on method- as far as I can tell, there's no way to choose to pass either one or the other explicitly.

Would there be any objection to expanding the `request()` signature to enable those items to be passed explicitly? The way I'm conceiving of it, `fields` would still be encoded according to the request method, while the type-specific arguments (`params` and `form`? Dunno - suggestions welcome) would always be added to the request in their specified format.

If this isn't an appropriate extension, it's not a big deal - I can keep doing it in my own code.

shazow:
Generally our pattern has been to be explicit about fields which are immediately used in this layer, and implicit about fields which get passed further down. I feel this has served us well thus far, but I'm very open to discussion about this. How does everyone else feel?

I feel that at least at the highest level (let's say #868), we should make more fields explicit just for better self-documenting goodness.

shazow:
Er I think I misread what you meant. Have you seen RequestMethods.request_encode_url and RequestMethods.request_encode_body? Are those what you're looking for?

haikuginger:
Right - so with those methods, you can specify a set of fields to be formatted - with `request_encode_url`, they'll all be formatted as a querystring, and with `request_encode_body`, they'll all be formatted as an `x-www-form-urlencoded` body. If you do a straight `.request`, then urllib3 picks which one to use based on the method.

What I'm suggesting is more along the lines of a method that has two explicit arguments that can be used simultaneously - one that specifies that a dict passed to it will be encoded to a querystring, and one that specifies that a dict passed to it will be encoded to a request body.

Lukasa:
I'd be open to having that. Requests does this with `params` vs `data`.

shazow:
Traditionally (when we would just deal with ConnectionPools and urlopen) I would say "lol just encode the url yourself!" but with our higher-level pools and RequestMethods, I think that's a fair point.

Right now we have `fields` which encodes based on the method, and we have `body` which skips encoding. We could have `url_fields` and `body_fields` to disambiguate?

[...]

PR #894, opened 2016-06-15
Override Content-Type when setting body fields

haikuginger:
Fixes an issue spotted while working on #885 - if the user passes a `Content-Type` header explicitly, then we'll leave it alone when setting up body-encoded fields. This corrects for that; we'll override with the appropriate header for what we implemented.

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -131,6 +131,10 @@ class TestingApp(RequestHandler):
     def upload(self, request):
         "Confirm that the uploaded file conforms to specification"
         # FIXME: This is a huge broken mess
+        content_type = request.headers.get('Content-Type', '')
+        if 'multipart/form-data' not in content_type:
+            return Response('Wrong content type: {}'.format(content_type),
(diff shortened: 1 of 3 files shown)


Lukasa:
I have no problem with this, though it's a moderately breaking change that needs to be documented in the changelog. Leaving this open for more review.

sigmavirus24:
There are people who intentionally want to send the wrong content-type (either because a service requires them to) or because they're actively trying to stress something. I don't think we should be changing this out under their feet. If urllib3 were more of an abstraction, I might argue against these kinds of footguns, but it isn't.

shazow:
It's fine if you want to send the wrong content type, the problem is that if you're using multipart encoding then the encoding is literally wrong. The delim declaration is missing.

shazow:
Only alternative I can think of is append/replace the delim declaration in whatever content type is given. It's more parsing/bug-prone but maybe safer? (Though I'm not convinced anyone is relying on this broken feature, I can't imagine any http parser implementing it this way as it's too ambiguous.)

PR #895, opened 2016-06-15
Allow nested Retry objects in Retry.total

jhermann:
I had the following exception in the wild:

    TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'

This happened while doing a `pip install` from a `devpi` that was down, causing a `Connection refused` and thus triggering the retry logic of `urllib3`. Seems someone has code like `Retry(r.increment())` or similar – instead of hunting that down, I found it easier and more robust to fix it at the source of the exception, especially since the fix is easy. YMMV.

### DIFF
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -88,6 +88,17 @@ class RetryTest(unittest.TestCase):
         self.assertTrue(Retry(0).raise_on_redirect)
         self.assertFalse(Retry(False).raise_on_redirect)
 
+    def test_retry_total_is_a_retry(self):
+        """ When retry.total is a Retry object, do not fail as before with a
+            "TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'"
+        """
+        retry = Retry(total=5)
+        retry = retry.increment()
+        self.assertEqual(retry.total, 4)
+        retry.total = Retry(total=5)
+        retry = retry.increment()
+        self.assertEqual(retry.total, 4)
+
     def test_retry_read_zero(self):
         """ No second chances on read timeouts, by default """
         error = ReadTimeoutError(None, "/", "read timed out")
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks for this patch!

I'm not sure what I think about this, though. There's simply no way around it: the code sample you provided is a failure to abide by the contract of the `Retry` object. `total` is fairly clearly documented to be an integer, so the fact that Python's dynamism allows passing a `Retry` object there doesn't justify doing it.

I'd actually be *more* inclined to want to put `isinstance` checks on `total` and friends to force this code to blow up in the misbehaving library than to add this fallback, especially as it has the somewhat dangerous possibility of looping indefinitely.

The final note is that anyone using Retries in this way is unlikely to be getting the behaviour they're expecting. For example, `read` and `connect` retry counts will not be propagated in a meaningful way. So I'm inclined to want to make this pattern more explicitly an error than to tolerate it.

Leaving this for @shazow and the team to opine on as well, though: they may disagree with me. =)

haikuginger:
I'm in agreement with @Lukasa- adding all the logic that would allow you to do `Retry(Retry())` in an acceptable way would basically duplicate the existing `.increment()` logic, but without the clarity of purpose. A PR putting strictures on the type of `total` et. al. would be a :+1: for me, though.

sigmavirus24:
So I unfortunately agree with @Lukasa and @haikuginger. I think your better bet is to find what pip is using that is causing this behaviour and fixing it there.

shazow:
Yea, `Retry(r.increment())` is clearly buggy code and coercing a fix for one person's buggy is going to impact _everyone_ in the long run. I appreciate the time you put into the PR with the test but I'm going to close this.

I can be convinced to add better errors for that scenario but generally I'm against supplanting Python's poor typing system with explicit type checks for debugging.

shazow:
Also thanks for everyone's input. :)

Issue #881, opened 2016-05-27
log server / server info on connectionpool


jvanasco:
The current implementation of connectionpool logs only the "local path" of a url - it does not log the server component.

https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L400-L401

this logging is not useful in multithreaded applications *if* there are as few as 2 concurrent connections. it also becomes increasingly confusing if there are redirects involved.  

logging would be much better if the server were included somewhere on that line.

haikuginger:
This doesn't sound unreasonable to me. It should be a relatively straightforward change to include `self.host` in the debug line, so I think it's more a matter of getting agreement on "should we".

shazow:
That log line has been there since ~2007. When I first wrote it, it was supposed to mimic what you'd see in a webserver log file. Given that a ConnectionPool only services a single host, it didn't make sense to put the host in there.

I'd be open to add more advanced logging into the PoolManager (or SessionManager), though.

Lukasa:
@shazow Other than trying to mimic web server logs is there any reason not to put the host in there? It kinda seems like it's the most natural place for it...

jvanasco:
I understand the need to not break existing code. Perhaps this info could be appended to the line, or there could be an argument/namespace variable to allow toggling between existing and "future" options?

As-is, the line is largely useless when used with concurrent threaded workers.  Logging often looked like this (loosely recreated):

	INFO:Starting new HTTP connection (1): example.com
	INFO:Starting new HTTP connection (1): aaa.example.com
	INFO:Starting new HTTP connection (1): bbb.example.com
	INFO:Starting new HTTP connection (1): ccc.example.com
	INFO:Starting new HTTP connection (1): ddd.example.com
	INFO:Starting new HTTP connection (1): eee.example.com
	DEBUG:"GET /path/1 HTTP/1.1" 404 56286
	DEBUG:"GET /path/2 HTTP/1.1" 404 56286
	DEBUG:"GET /path/3 HTTP/1.1" 404 56286
	DEBUG:"GET /path/4 HTTP/1.1" 404 56286
	DEBUG:"GET /path/5 HTTP/1.1" 404 56286
	INFO:Starting new HTTP connection (1): fff.example.com
	DEBUG:"GET /path/6 HTTP/1.1" 404 56286
	DEBUG:"GET /path/7 HTTP/1.1" 404 56286

That's 7 requests.  Imagine batch-processing in 100 threads :/

[...]

Issue #869, opened 2016-05-22
MaxRetryError when redirecting from http to https through PoolManager()


jayvdb:
As just an example http://en.wikipedia.org/wiki/Main_Page redirects to https://en.wikipedia.org/wiki/Main_Page , but requesting http://en.wikipedia.org/wiki/Main_Page via PoolManager fails. 

Using `urllib3.connection_from_url`:

```bash
python -c "import ssl, urllib3; r = urllib3.connection_from_url('https://en.wikipedia.org/wiki/Main_Page').request('GET', 'http://en.wikipedia.org/wiki/Main_Page', assert_same_host=False); print(r.status);"
urllib3/connectionpool.py:791: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
200
```

Using `urllib3.PoolManager().connection_from_url`

```bash
python -c "import ssl, urllib3; r = urllib3.PoolManager().connection_from_url('http://en.wikipedia.org/wiki/Main_Page').request('GET', 'http://en.wikipedia.org/wiki/Main_Page', assert_same_host=False); print(r.status);"
```

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "urllib3/request.py", line 69, in request
    **urlopen_kw)
  File "urllib3/request.py", line 90, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "urllib3/connectionpool.py", line 653, in urlopen
    release_conn=release_conn, **response_kw)
  File "urllib3/connectionpool.py", line 653, in urlopen
    release_conn=release_conn, **response_kw)
  File "urllib3/connectionpool.py", line 653, in urlopen
    release_conn=release_conn, **response_kw)
  File "urllib3/connectionpool.py", line 638, in urlopen
    retries = retries.increment(method, url, response=response, _pool=self)
  File "urllib3/util/retry.py", line 273, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='en.wikipedia.org', port=80): Max retries exceeded with url: https://en.wikipedia.org/wiki/Main_Page (Caused by ResponseError('too many redirects',))
```

Is that correct?  If it is correct, it feels a little confusing seeing `MaxRetryError` here.

haikuginger:
[...]

PR #897, opened 2016-06-17
add domain and method aware logging to connectionpool

birm:
One of my attempts at #881. I would consider at most one of the two (#898).

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -397,9 +397,9 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         # AppEngine doesn't have a version attr.
         http_version = getattr(conn, '_http_vsn_str', 'HTTP/?')
-        log.debug("%s:%s \"%s %s %s\" %s %s using scheme %s ", self.host, self.port,
+        log.debug("%s://%s:%s \"%s %s %s\" %s %s", self.scheme, self.host, self.port,
                   method, url, http_version, httplib_response.status,
-                  httplib_response.length, self.scheme)
+                  httplib_response.length)
 
         try:
             assert_header_parsing(httplib_response.msg)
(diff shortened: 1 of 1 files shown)


birm:
I don't really understand the pypy error here.

Lukasa:
The PyPy error is now magically gone. =D

I'd be happy with this change. Up to @shazow though!

haikuginger:
Looks good to me!

shazow:
Should we be including the scheme too, while we're at it?

haikuginger:
I vote yes for including scheme; wanting to know if your requests are happening securely is a valid use case for logging. But it should be based on the calling class scheme, rather than the URL's scheme; as seen in #869, the URL passed can be incorrect if `assert_same_host` is passed as False.

birm:
Okay, I agree, so I added the scheme. I'm not sure if the way I wrote it could be better though. thoughts?

Lukasa:
I think I'd like to skip `using scheme` if possible: it's nice to have logs that are relatively structured (e.g. can be parsed by grep). Either we should just find a place in there to put the string, or we should consider using `key=value` logic for some of it.

urllib3/connectionpool.py
```diff
@@ -397,8 +397,9 @@ def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
 
         # AppEngine doesn't have a version attr.
         http_version = getattr(conn, '_http_vsn_str', 'HTTP/?')
-        log.debug("\"%s %s %s\" %s %s", method, url, http_version,
-                  httplib_response.status, httplib_response.length)
+        log.debug("%s:%s \"%s %s %s\" %s %s using scheme %s ", self.host, self.port,
```
shazow: Would `https://127.0.0.1:8000` be better?

birm: That's much better, and I should have done that from the start.

shazow: 👍 no worries. :)

birm:
I now have scheme before the host; (scheme)://(host):(port) ...
Does this break anyone's workflow?

shazow:
Only one way to find out. :D

PR #898, opened 2016-06-17
add domain aware logging to poolmanager

birm:
this also involved making a str method for RecentlyUsedContainer, however I couldn't find an easy way to complete the debug log info as requested.

One of my attempts at #881. I would consider at most one of the two (#897).


Lukasa:
Thanks for this! For my part I'd prefer #897.

shazow:
Yea, I think I've been convinced that #897 is the way to go. :) Thank you for both proposals!

Issue #890, opened 2016-06-12
Tests fail on Win32


mindw:
Tested Python 2.7.11 32bit, 3.4.3 64bit, 3.5.1 64bit:
Also PySocks 1.5.7 is broken for Python 2.7.x on windows due to Anorov/PySocks#35.
**2.7.11:**
```
py27 runtests: commands[1] | nosetests
SSSINFO:urllib3.connectionpool:Starting new HTTPS connection (1): localhost
INFO:tornado.access:200 GET / (::1) 2.00ms
DEBUG:urllib3.connectionpool:"GET / HTTP/1.1" 200 13
......E..........E..S........................................................................................................E..........S..S..........................................SSS......S..............................................................................................................................................................................................................
======================================================================
ERROR: test_ca_dir_verified (test.contrib.test_pyopenssl.TestHTTPS)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "J:\dev\workspace\pythonxy-xy-27\src\python\urllib3\__workspace__\urllib3-git\test\__init__.py", line 66, in wrapper
    return test(*args, **kwargs)
  File "J:\dev\workspace\pythonxy-xy-27\src\python\urllib3\__workspace__\urllib3-git\test\with_dummyserver\test_https.py", line 101, in test_ca_dir_verified
    r = https_pool.request('GET', '/')
  File "J:\dev\workspace\pythonxy-xy-27\src\python\urllib3\__workspace__\urllib3-git\urllib3\request.py", line 69,
in request
    **urlopen_kw)
  File "J:\dev\workspace\pythonxy-xy-27\src\python\urllib3\__workspace__\urllib3-git\urllib3\request.py", line 90,
in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "J:\dev\workspace\pythonxy-xy-27\src\python\urllib3\__workspace__\urllib3-git\urllib3\connectionpool.py", line 621, in urlopen
    raise SSLError(e)
SSLError: ("bad handshake: Error([('PEM routines', 'PEM_read_bio', 'no start line'), ('x509 certificate routines',
'X509_load_cert_file', 'PEM lib'), ('SSL routines', 'ssl3_get_server_certificate', 'certificate verify failed')],)",)
-------------------- >> begin captured logging << --------------------
urllib3.connectionpool: INFO: Starting new HTTPS connection (1): localhost
urllib3.connectionpool: INFO: Starting new HTTPS connection (2): localhost
[...]

PR #899, opened 2016-06-18
on windows use wincertstore to simulate platform CA

mindw:
A proposal for fixing #890


Lukasa:
So, I don't think we want this. I'll elaborate more on #890.

PR #901, opened 2016-06-20
Remove absolute import.

Lukasa:
This absolute import breaks vendorizing urllib3: see kennethreitz/requests#3006.

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -60,7 +60,7 @@ try:  # Platform-specific: Python 2
     from socket import _fileobject
 except ImportError:  # Platform-specific: Python 3
     _fileobject = None
-    from urllib3.packages.backports.makefile import backport_makefile
+    from ..packages.backports.makefile import backport_makefile
 
 import ssl
 import select


haikuginger:
:+1:; I don't see a reason to not go ahead and merge.

shazow:
Good call, maybe we need somekind of test (integration or otherwise) to avoid this kind of error in the future.

Lukasa:
I think we can probably just write a really quick script for it: nothing in the urllib3 directory should have a file with the string "import urllib3" in it.

haikuginger:
Or `from urllib3`.

Lukasa:
Good call. It'd probably be pretty easy to write this as a little third-party Python tool that checks for this sort of thing if you were really interested, but otherwise it's basically just a `grep`.

haikuginger:
For right now, maybe skip the automation - those constructs are valid within comments/docstrings, so to identify them would be substantially more complex than a plain grep.

I ended up doing `grep -r -v '^#' urllib3 | grep -e 'import urllib3' -e 'from urllib3'` and did find one additional change that needed to be made, but other than that, the other item it found was just inside a docstring. It's quite likely that there's a way to exclude docstrings as well, but at the moment I don't have the regex magic necessary to make that happen.

PR #902, opened 2016-06-20
Removing absolute import in NTLMPool

haikuginger:
Continues work of #901.


Lukasa:
Ah, cool, so this isn't a heavily used contrib module, which is why I suspect it's been lurking so long. Looks great to me. :sparkles: :custard:

Issue #892, opened 2016-06-14
Planning for breaking changes on TRFKA1.17


haikuginger:
So, we're taking a step back and considering/refactoring all the breaking changes we want to make in the near future that'll break current API contracts on some level. This issue is basically a dump of my thoughts/plans so far, which are open for discussion to all comers.
- Method of SSL/TLS verification: There's been an interest in having platform-independent hostname verification pre-enabled for `SessionManager`. From talking to @Lukasa, there are two main ways to go about this - first, to either import or package certifi, or to require PyOpenSSL, certitude, and other binary packages to be able to reliably use the platform cert bundles. At this point, certifi is a more practical option, so we basically need to decide if we want to bundle it or require it.
- Scope of SSL/TLS verification: @shazow has indicated that he'd like autoconfig on `SessionManager`, as a high-level configuration object. Personally, I think that if we're putting in the effort to set up a secure default, all of our connection levels should be able to benefit from it - including `ConnectionPool` and `PoolManager`. This is up for discussion, and brings us to...
- Settings handling in general: I'd like to unify all sorts of autoconfiguration code onto a `Settings` object that can be passed into any `RequestMethods` subclass. This would include things like the `SSLContext` to be used, any platform-specific flags that need to be considered (e.g., GAE), or anything else. My vision would be to have any newly-created `Settings` object carry smart defaults, and any newly-created `RequestMethods` subclass get either a `Settings` object in arguments or one of a fresh instantiation of same or a link to a `urllib3`-wide `Settings` object. It'd look something like this:

``` python
import urllib3

# We can call like this...
sm = urllib3.SessionManager()

# Or like this...
insecure_settings = urllib3.Settings(verify_ssl=False)
sm = urlib3.SessionManager(settings=dumb_settings)

# Or you could even do this!
class InsecurePoolManager(urllib3.PoolManager):

    def __init__(self, *args, **kwargs):
        insecure_settings = Settings(verify_ssl=False)
        kwargs[settings] = insecure_settings
        super(InsecurePoolManager, self).__init__(*args, **kwargs)
```
- While we're on `RequestMethods`, support will be added for explicitly-passed params and body fields. (#885)
[...]

PR #903, opened 2016-06-21
Discrete field types in RequestMethods

haikuginger:
Implements #885, in support of #892.

### DIFF
--- a/urllib3/request.py
+++ b/urllib3/request.py
@@ -92,7 +92,7 @@ class RequestMethods(object):
         fields = fields or []
 
         if fields or form_fields:
-    
+
             content_type = None
 
             if body is not None:
@@ -129,7 +129,7 @@ class RequestMethods(object):
             querystring += urlencode(fields)
             if querystring and url_params:
                 querystring += '&'
-        
+
         if url_params:
             querystring += urlencode(url_params)
 
--- a/urllib3/request.py
+++ b/urllib3/request.py
@@ -50,7 +50,7 @@ class RequestMethods(object):
         raise NotImplemented("Classes extending RequestMethods must implement "
                              "their own ``urlopen`` method.")
 
-    def request(self, method, url, fields=None, headers=None, **urlopen_kw):
+    def request(self, method, url, fields=None, headers=None, body=None, **urlopen_kw):
         """
         Make a request using :meth:`urlopen` with the appropriate encoding of
         ``fields`` based on the ``method`` used.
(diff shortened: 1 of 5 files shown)


urllib3/request.py
```diff
@@ -61,91 +54,79 @@ def request(self, method, url, fields=None, headers=None, **urlopen_kw):
...
-        if method in self._encode_url_methods:
-            return self.request_encode_url(method, url, fields=fields,
-                                           headers=headers,
-                                           **urlopen_kw)
-        else:
-            return self.request_encode_body(method, url, fields=fields,
-                                            headers=headers,
-                                            **urlopen_kw)
-
-    def request_encode_url(self, method, url, fields=None, headers=None,
```
shazow: We probably want to keep backwards compat for these public methods?

haikuginger: Do you want to re-insert the existing code, or would shims that have the same behavior, using the new methods, work?

shazow: Shims please. :)

urllib3/request.py
```diff
@@ -61,91 +54,79 @@ def request(self, method, url, fields=None, headers=None, **urlopen_kw):
...
                 raise TypeError(
                     "request got values for both 'fields' and 'body', can only specify one.")
 
-            if encode_multipart:
+            if encode_multipart and method not in self._encode_url_methods:
+                fields = iter_field_objects(fields, form_fields)
                 body, content_type = encode_multipart_formdata(fields, boundary=multipart_boundary)
-            else:
-                body, content_type = urlencode(fields), 'application/x-www-form-urlencoded'
+            elif method not in self._encode_url_methods:
```
shazow: This code is almost identical to `encode_url`, maybe they should be normalized somehow?

I think it's probably not a great idea to expose `encode_url` as it is, but make it into a `_`-prefixed helper that can be used in both cases.

urllib3/request.py
```diff
@@ -61,91 +54,79 @@ def request(self, method, url, fields=None, headers=None, **urlopen_kw):
...
-        string can be explicitly set with the ``multipart_boundary`` parameter.
+        Encode and return a request body and headers to match
         """
-        if headers is None:
-            headers = self.headers
+        headers = headers or dict()
+        form_fields = form_fields or []
+        fields = fields or []
+
+        if fields or form_fields:
```
shazow: Since this if-statement is getting to be the main body of the flow, would be good to invert it and return early.

PR #904, opened 2016-06-22
Approved SessionManager changes

haikuginger:
Basic implementation of the SessionManager cookie-handling logic.

### DIFF
--- a/test/test_sessionmanager.py
+++ b/test/test_sessionmanager.py
@@ -1,0 +1,105 @@
+import unittest
+
+from urllib3.sessionmanager import SessionManager
+from urllib3.poolmanager import ProxyManager, PoolManager
+from urllib3.util.sessioncontext import DefaultCookiePolicy, CookieJar, SessionContext
+from urllib3.packages import six
+from urllib3.response import HTTPResponse
+from urllib3.request import Request
+
+cookielib = six.moves.http_cookiejar
+
+
+class TestSessionManager(unittest.TestCase):
+
+    def setUp(self):
+        self.manager = SessionManager(PoolManager())
+
+    def test_set_cookie_policy(self):
+        policy = cookielib.DefaultCookiePolicy()
+        self.assertNotEqual(policy, self.manager.context.cookie_jar._policy)
+        self.manager.context.cookie_jar.set_policy(policy)
+        self.assertTrue(policy is self.manager.context.cookie_jar._policy)
+
+    def test_create_proxy_manager(self):
+        manager = SessionManager(ProxyManager(proxy_url='http://none'))
+        self.assertTrue(isinstance(manager.manager, ProxyManager))
+
+    def test_creates_pool_manager(self):
+        self.assertTrue(isinstance(self.manager.manager, PoolManager))
+
+    def test_with_external_jar(self):
+        this_policy = cookielib.DefaultCookiePolicy()
+        jar = CookieJar(policy=this_policy)
+        context = SessionContext(cookie_jar=jar)
+        manager = SessionManager(PoolManager(), context=context)
+        print(dir(manager.context))
+        self.assertTrue(manager.context.cookie_jar is jar)
+
+
+class TestSessionContext(unittest.TestCase):
+
+    def setUp(self):
+        self.context = SessionContext()
+
+    def test_extract_cookie(self):
+        self.assertFalse(self.context.cookie_jar)
+        req = Request(method='GET', url='http://google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        self.assertTrue(self.context.cookie_jar)
+
+    def test_apply_cookie(self):
+        req = Request(method='GET', url='http://google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        req = Request(method='GET', url='http://google.com')
+        self.context.apply_to(req)
+        self.assertEqual(req.headers.get('Cookie'), 'cookiename=cookieval')
+
+    def test_no_apply_cookie(self):
+        req = Request(method='GET', url='http://google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        self.assertTrue(self.context.cookie_jar)
+        req = Request(method='GET', url='http://evil.com')
+        self.context.apply_to(req)
+        self.assertTrue(req.headers.get('Cookie', None) is None)
+
+    def test_unacceptable_cookie(self):
+        self.assertFalse(self.context.cookie_jar)
+        req = Request(method='GET', url='http://evil.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval; domain=.google.com'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        self.assertFalse(self.context.cookie_jar)
+
+    def test_parent_domain(self):
+        req = Request(method='GET', url='http://www.google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval; domain=.google.com'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        req = Request(method='GET', url='http://google.com')
+        self.context.apply_to(req)
+        self.assertEqual(req.headers.get('Cookie'), 'cookiename=cookieval')
+
+    def test_sibling_domain(self):
+        req = Request(method='GET', url='http://ww1.google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        req = Request(method='GET', url='http://ww2.google.com')
+        self.context.apply_to(req)
+        self.assertTrue(req.headers.get('Cookie') is None)
+
+    def test_sibling_domain_with_wildcard(self):
+        req = Request(method='GET', url='http://ww1.google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval; domain=.google.com'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        req = Request(method='GET', url='http://ww2.google.com')
+        self.context.apply_to(req)
+        self.assertEqual(req.headers.get('Cookie'), 'cookiename=cookieval')
--- a/test/test_sessionmanager.py
+++ b/test/test_sessionmanager.py
@@ -78,7 +78,7 @@ class TestSessionContext(unittest.TestCase):
         self.context.extract_from(resp, req)
         self.assertTrue(self.context.cookie_jar)
         for each in self.context.cookie_jar:
-            self.assertEqual(repr(each), repr(expected_cookie))
+            self.assertEqual(each.__dict__, expected_cookie.__dict__)
 
     def test_apply_cookie(self):
         """
--- a/test/test_sessionmanager.py
+++ b/test/test_sessionmanager.py
@@ -16,24 +16,38 @@ class TestSessionManager(unittest.TestCase):
         self.manager = SessionManager(PoolManager())
 
     def test_set_cookie_policy(self):
+        """
+        Test that we're able to set the policy on the SessionManager's CookieJar
+        """
         policy = cookielib.DefaultCookiePolicy()
         self.assertNotEqual(policy, self.manager.context.cookie_jar._policy)
         self.manager.context.cookie_jar.set_policy(policy)
         self.assertTrue(policy is self.manager.context.cookie_jar._policy)
 
     def test_create_proxy_manager(self):
-        manager = SessionManager(ProxyManager(proxy_url='http://none'))
-        self.assertTrue(isinstance(manager.manager, ProxyManager))
+        """
+        Make sure that when we pass a ProxyManager in, we use it.
+        """
+        pm = ProxyManager(proxy_url='http://none')
+        manager = SessionManager(pm)
+        self.assertTrue(manager.manager is pm)
 
     def test_creates_pool_manager(self):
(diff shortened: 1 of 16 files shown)


haikuginger:
@Lukasa, if you want to merge this, then I can start submitting PRs against that branch.

test/test_sessionmanager.py
```diff
@@ -0,0 +1,105 @@
...
+
+class TestSessionManager(unittest.TestCase):
+
+    def setUp(self):
+        self.manager = SessionManager(PoolManager())
+
+    def test_set_cookie_policy(self):
+        policy = cookielib.DefaultCookiePolicy()
+        self.assertNotEqual(policy, self.manager.context.cookie_jar._policy)
+        self.manager.context.cookie_jar.set_policy(policy)
```
Lukasa: So this seems *really* clunky: we're well past a Law of Demeter violation here. Perhaps the `SessionManager` should expose helpers for things like this, rather than expect users to patch this many objects deep?

test/test_sessionmanager.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def test_creates_pool_manager(self):
+        self.assertTrue(isinstance(self.manager.manager, PoolManager))
+
+    def test_with_external_jar(self):
+        this_policy = cookielib.DefaultCookiePolicy()
+        jar = CookieJar(policy=this_policy)
+        context = SessionContext(cookie_jar=jar)
+        manager = SessionManager(PoolManager(), context=context)
+        print(dir(manager.context))
```
Lukasa: This should probably be removed.

test/test_sessionmanager.py
```diff
@@ -0,0 +1,105 @@
...
+    def setUp(self):
+        self.context = SessionContext()
+
+    def test_extract_cookie(self):
+        self.assertFalse(self.context.cookie_jar)
+        req = Request(method='GET', url='http://google.com')
+        headers = {'Set-Cookie': 'cookiename=cookieval'}
+        resp = HTTPResponse(headers=headers)
+        self.context.extract_from(resp, req)
+        self.assertTrue(self.context.cookie_jar)
```
Lukasa: Ideally we'd have this test confirm not just that there is now something in the cookie jar, but what that is.

test/test_sessionmanager.py
```diff
@@ -0,0 +1,105 @@
...
+        print(dir(manager.context))
+        self.assertTrue(manager.context.cookie_jar is jar)
+
+
+class TestSessionContext(unittest.TestCase):
+
+    def setUp(self):
+        self.context = SessionContext()
+
+    def test_extract_cookie(self):
```
Lukasa: Can all of these tests have docstrings that explain what they're trying to do and what specifically they're aiming to test? As these test files get bigger and are written by increasingly disparate authors it will help us understand intent as well as execution =)

urllib3/util/retry.py
```diff
@@ -295,6 +294,20 @@ def __repr__(self):
...
+        """
+        Assuming we're doing a redirect, should we change HTTP methods?
+        """
+        if method == 'GET':
+            return method
+        if method == 'HEAD':
+            return method
+        if status == 303:
+            return 'GET'
+        if method == 'POST' and status in set([301, 302, 303]):
```
Lukasa: You could in principle remove 303 from this, though it doesn't matter much. It may be better, however, to hoist this set up to module scope (maybe as `_REDIRECT_STATUSES`?) rather than construct it each time we call this function.

urllib3/request.py
```diff
@@ -149,3 +187,24 @@ def request_encode_body(self, method, url, fields=None, headers=None,
...
+
+    def redirect(self, response, method, retries, **kwargs):
+        """
+        Abstracts the redirect process to be used from any :class:`RequestMethods` object
+        """
+        url = kwargs.pop('url', '')
+        redirect_location = urljoin(url, response.get_redirect_location())
+        method = retries.redirect_method(method, response.status)
+        try:
+            pool = kwargs.pop('pool', self)
```
Lukasa: Presumably this can't raise `MaxRetryError`, so let's remove it from the `try` block.

urllib3/request.py
```diff
@@ -1,13 +1,51 @@
...
 
 
-__all__ = ['RequestMethods']
+class Request(_Request):
+    """
+    Currently used as a shim to allow us to work with the stdlib cookie
+    handling, which expects a `urllib.request.Request`-like object.
+    """
+    def __init__(self, *args, **kwargs):
+        kwargs.pop('method', None)
```
Lukasa: Given that we're doing this only for its side effects, let's swap to `del kwargs['method']`

urllib3/response.py
```diff
@@ -169,6 +169,10 @@ def data(self):
     def connection(self):
         return self._connection
 
+    def info(self):
+        # Compatibility method for cookie handler
+        return self.headers
+
```
Lukasa: Can you turn this comment into a docstring that also includes an admonition not to use this method?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,77 @@
...
+        :class:`urllib3.poolmanager.ProxyManager` object
+        to handle HTTP requests for the SessionManager
+
+    :param context:
+        A predefined :class:`urllib3.util.context.SessionContext` object to use in the session;
+        if not provided, a new one will be created.
+
+    :param headers:
+        A set of headers to include with all requests, unless other
+        headers are given explicitly.
```
Lukasa: Maybe avoid the use of the word `set` here? It's also the name of a data type, and so this can confuse users who aren't thinking carefully about how this might work.

urllib3/util/sessioncontext.py
```diff
@@ -0,0 +1,75 @@
...
+        policy = PythonCookiePolicy.DomainStrict
+        kwargs.setdefault('strict_ns_domain', policy)
+        # Old-style class on Python 2
+        PythonCookiePolicy.__init__(self, *args, **kwargs)
+
+
+class CookieJar(PythonCookieJar):
+
+    def __init__(self, policy=None):
+        policy = policy or DefaultCookiePolicy()
```
Lukasa: We maybe want to use an explicit `is not None` check here, rather than rely on the truthiness of arbitrary objects.

urllib3/util/sessioncontext.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            self._policy._now = self._now = int(time.time())
+            cookies = self._cookies_for_request(request)
+            attrs = self._cookie_attrs(cookies)
+            # This is a modification; stdlib sets the entire cookie header
+            # and only if it's not there already. We're less picky.
+            if attrs:
+                request.add_cookies(*attrs)
+        finally:
+            self._cookies_lock.release()
```
Lukasa: Presumably this `cookies_lock` dance can be done with a `with` statement, instead of the `try...finally`?

Lukasa:
I have some feedback!

urllib3/sessionmanager.py
haikuginger: Can do!

urllib3/request.py
haikuginger: I believe `MaxRetryError` is actually what gets raised when we have too many redirects, so this block can raise that exception. Although, I'm not sure, in that case, if the connection still needs to be released or if it would have been closed already by a lower-level routine.

urllib3/util/retry.py
haikuginger: Sounds good.

test/test_sessionmanager.py
haikuginger: Ah. Yep.

urllib3/request.py
Lukasa: I'm not disputing that `MaxRetryError` can be raised, I'm saying that `dict.pop` can't raise it. =)

test/test_sessionmanager.py
haikuginger: I initially had a helper method at the `SessionManager` level to handle that; we had moved away from it as it too tightly coupled `SessionManager` to the lower levels. I do have a draft concept that may work here and maintain generic-ness, but I'd need to bring in all the commits from a later 
branch to provide the restructuring necessary. Would you prefer that on this PR, or would you prefer to merge this, and then see that draft?

Lukasa: Let's merge this first then. =)

urllib3/request.py
haikuginger: Gotcha. Can do!

urllib3/util/sessioncontext.py
haikuginger: I'll check that out to verify. This code block is largely copied from the stdlib, so I had avoided messing with it too much.

urllib3/util/sessioncontext.py
haikuginger: Sounds good!

urllib3/response.py
haikuginger: Yes!

urllib3/request.py
haikuginger: Sounds good.

test/test_sessionmanager.py
haikuginger: I'll look into that; not sure how much data is needed to construct a single cookie from scratch, but it should be doable.

test/test_sessionmanager.py
haikuginger: Agreed.

test/test_sessionmanager.py
haikuginger: So, for this one, the Cookie object doesn't have an equality operator, which makes it tricky. Best we can do is compare string representations using `__repr__`. Does that sound good?

urllib3/util/retry.py
haikuginger: I'm hoisting this to module scope, but leaving 303 in the set to help future readers. Explicit (having the code in the set) is better than implicit (skip it, because the logic flow prevents it from being a factor).

test/test_sessionmanager.py
Lukasa: Can we write our own comparator using the properties of the `Cookie` class?

haikuginger: We can; the `repr()` representation of `Cookie` already carries all the attribute values, though, so it might not give us too much of a benefit.

urllib3/util/sessioncontext.py
haikuginger: Check!

test/test_sessionmanager.py
haikuginger: Better: all the relevant bits are in `Cookie.__dict__`, which _does_ have an equality operator.

haikuginger:
@Lukasa, should be good if you want to take another look. 🙂

Lukasa:
Alright, I pronounce myself happy! I'd like a section in CHANGES.rst please, and then I'd also like any other contributor who is interested to please do a once-over review. If I don't hear anything else I'll aim to merge sometime this week.

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,77 @@
...
+
+    """
+    def __init__(self, manager, context=None, headers=None):
+        super(SessionManager, self).__init__(headers=headers)
+        self.manager = manager
+        self.context = context or SessionContext()
+
+    def urlopen(self, method, url, redirect=True, retries=None, **kw):
+        """
+        Same as :meth:`urllib2.poolmanager.PoolManager.urlopen` with added
```
sigmavirus24: urllib2 or urllib3? or perhaps urllib6?

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,77 @@
...
+        A predefined :class:`urllib3.util.context.SessionContext` object to use in the session;
+        if not provided, a new one will be created.
+
+    :param headers:
+        Headers to include with all requests, unless other
+        headers are given explicitly.
+
+    Example::
+
+        >>> manager = SessionManager(PoolManager())
```
sigmavirus24: I have mixed feelings about this. I feel like if this is how most people are going to use it, we shouldn't require it be provided.

urllib3/sessionmanager.py
haikuginger: urllib12 Gold Edition Epic Plus. Will fix.

urllib3/sessionmanager.py
haikuginger: I agree with you on that; this was the pattern we had originally arrived on eventually, though. Should we just default to `PoolManager`, or should we try to do something smarter?

haikuginger: Possibly something like this:

```python
class SessionManager(RequestMethods):

    manager_class = PoolManager

    def __init__(self, ..., manager=None, **manager_kw):
        if manager is not None:
            self.manager = manager
        else:
            self.manager = self.manager_class(**manager_kw)
        
```

That'd let us subclass with different types of managers (notably `ProxyManager`) pretty easily.

haikuginger: @sigmavirus24, how do you feel about the above pattern? I have it made, but I don't want to bother changing the tests again if it's not something we like. 😜

haikuginger:
Rebased from 4aee526.

haikuginger:
@Lukasa, @sigmavirus24, I've made a couple changes and believe this is ready to merge into the working branch. From there, we can get into exactly what it needs to look like before it merges to master.

Lukasa:
ACK. I have no time for a detailed review of this today, but I'll do it tomorrow or Monday. =)

Lukasa:
Ok, let's move this onward.

urllib3/sessionmanager.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        headers = kw.pop('headers', self.headers)
+
+        if not isinstance(retries, Retry):
+            retries = Retry.from_int(retries, redirect=redirect)
+
+        # Build a mock Request object to work with
+        request_object = Request(url=url, method=method, headers=headers)
+        self.context.apply_to(request_object)
+        modified_headers = request_object.get_all_headers()
```
Foorack, line 64: To anyone coming 5 years later and wants to do this, the "get_all_headers" function doesn't exist, and the cookies through "add_cookie_header" are not being applied to the "headers" array, but to a separate "unredirected_hdrs" array.

Issue #907, opened 2016-06-24
Chunked downloading is broken in >1.12


igungor:
Try this code snippet:

``` python
import urllib3

http = urllib3.PoolManager()
r = http.request('GET', 'http://www.nyaa.se/?page=download&tid=818605')
for chunk in r.stream():
    print len(chunk)
```

on urllib 1.12, the output is:

``` sh
(r)~  Δ python t.py
3862
5535
4150
3813
```

on urllib 1.13 and newer releases (every release, also master):

``` sh
(r)~  Δ python t.py
Traceback (most recent call last):
  File "t.py", line 7, in <module>
    for chunk in r.stream():
  File "/Users/ibrahim/.virtualenvs/r/lib/python2.7/site-packages/urllib3/response.py", line 353, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "/Users/ibrahim/.virtualenvs/r/lib/python2.7/site-packages/urllib3/response.py", line 502, in read_chunked
    self._update_chunk_length()
  File "/Users/ibrahim/.virtualenvs/r/lib/python2.7/site-packages/urllib3/response.py", line 448, in _update_chunk_length
    line = self._fp.fp.readline()
AttributeError: 'NoneType' object has no attribute 'readline'
```
##

sigmavirus24:
Woot thanks for the great reproduction case. So we're releasing the connection in urllib3.connectionpool because you're not specifying `preload_content=False`. I wonder what changed in 1.12 that caused that to fail like this.

sigmavirus24:
I bet we're preloading the content, but since we only check if the response is chunked, we don't ever expect the case where `preload_content=True`.

sigmavirus24:
And I bet this is all my fault. :(

sigmavirus24:
So we really do expect you to specify `preload_content=False` to stream a chunked response. That said, we can fix this, I'm just not sure we either *should* or will *want* to. A POC is up on #908.

igungor:
Alright, thanks for the quick reply and the suggestion. I've been bitten by this on a production code, which uses python-requests module (2.10.0), and it uses urlllib3 (1.15.1).

I guess it's fair to mention @kennethreitz about the issue. Should I file a bug for requests as well @kennethreitz ?

sigmavirus24:
[...]

PR #908, opened 2016-06-24
Fix chunked transfer-encoding with preload_content

sigmavirus24:
When we implemented our own chunked transfer-encoding logic, we broke
the use case (that we did not anticipate) of people not specifying
preload_content=False to stream their data. This provides a fix, sans
tests, for that use case.

Closes gh-907

---

**Note**:

- I'm not sure we want to fix this case (see the issue).
- I didn't write tests, I just wanted to see if I could work in a fix quickly before work.


shazow:
Hmm can you expand on the alternative? Is avoiding using chunked when not preloading allow all the necessary scenarios? (Keeping in mind that we do want to make preload_content=False by default sooner than later)

sigmavirus24:
@shazow so we have logic that checks `self.chunked` and doesn't check `self._body` so if you make a request to the site that is in gh-907, and instead do `r.data` (instead of `r.stream()`) then you'll get the same error. So what that says to me, is that if `preload_content` is True, we should probably set `self.chunked` to False because we're going to preload the content and make it not chunked. That alternative is valid regardless of the default value of preload content now or in the future. It's a preventative measure for a foot-gun.

shazow:
If I understand this correctly, you're saying that chunked responses should not be streamable? Or that the distinction of chunked vs not chunked is meaningless when we're streaming anyways?

sigmavirus24:
It's not just chunked versus not chunked when streaming. It also requires preload_content. If you make a request with the URL that was in the original issue, with our current implicit `preload_content=True` to a service that returns a chunked response, then we will happily preload that content for you. The problem is that if you use `r.stream()` (which people shouldn't if they don't pass `preload_content`) you get the AttributeError. I'm not sure if we want to refuse to stream in this case or if we want to never let `self.chunked` be True because that branch of logic is what causes this error.

shazow:
Er sorry I was using stream interchangeably with read-on-demand (as opposed to preloading).

If we can avoid the branching, all the better, but it's still not clear to me which scenario we're sacrificing (if any?).

sigmavirus24:
I think the confusion is generated by my original comment. Let me clarify with what I understand right now:

- We used to allow people to request a chunked URL without `preload_content=False` and allow them to use `response.stream()`
- We broke this because while we still pre-load the content, we also set `response.chunked = True` which causes us to try to read from the original response

It seems like this is *very* rare behaviour, but I don't think we want to continue having an AttributeError be raised when someone calls `response.stream()` in *some* cases. This PR as it is now does not really do what we want (I don't think). The question becomes: What do we want the behaviour to be?

shazow:
I agree that side-effect surprise exceptions are no good. What are the options for the behaviour?

Keeping in mind that ultimately we might want to get rid of preload_content altogether (and force False), especially once we have the contextmanager pattern which closes unconsumed responses on exit.

sigmavirus24:
> What are the options for the behaviour?

So we can:

- re-implement the old behaviour. This is as simple as setting `response.chunked = False` if `preload_content is True`.
- ignore preload content (like this PR presently does, although it does more than that) if we're dealing with a chunked response which might not be what the user actually wants
- catch the AttributeError in `response.stream` (or `read_chunked` or somewhere else) and raise a new exception instead that explains what might have happened?

shazow:
I honestly don't know what's best here. :) First option seems okay but they all seem okay (the third one seems saddest though).

Paging @Lukasa @jonparrott @haikuginger

Lukasa:
Let's do the first for now. We can then adjust that behaviour later down the line when `preload_context` gets nixed.

PR #911, opened 2016-07-01
Normalize the scheme and host in the URL parser

### LINKED ISSUE #833: PoolManagers are overly case-sensitive.
In a URI, both scheme and authority are case-insensitive: that is, the scheme, host, and port should all be considered case insensitively (and basically treated as lower-case).

Unfortunately, the PoolManager and ProxyManager do not consistently do that. The biggest problem is in the ProxyManager, where `connection_from_host` will incorrectly treat any HTTPS scheme with a capital letter in it like a HTTP connection (which isn't great), but the flaw is pretty common elsewhere in those objects (e.g. an uppercased host will get a different connection pool to a lowercased host).

At the very minimum we should be using a lowercased scheme at all points in the urllib3 stack. We should also consider whether we want to be case-sensitive about domains (I'm inclined to say that we do not).

Related: kennethreitz/requests#3075.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32609352-poolmanagers-are-overly-case-sensitive?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremycline:
I don't know that this is the most rigorous way to fix the problem of case-sensitivity, but it seemed like a nice centralized place to normalize things.

According to RFC 3986, the scheme and the host portion of the authority
in a URI are case-insensitive. ``urllib3`` uses the scheme in particular
as a key for various dictionaries, and oddly-cased schemes can break
this. This patch modifies the Url class to normalize both the scheme and
host.

fixes #833

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ dev (master)
 * Used URLFetch default timeout on AppEngine, rather than hardcoding our own.
   (Issue #858)
 
+* Normalize the scheme and host in the URL parser (Issue #833)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -77,13 +77,13 @@ class TestUtil(unittest.TestCase):
             'http://[2a00:1450:4001:c01::67]:80/test': ('http', '[2a00:1450:4001:c01::67]', 80),
 
             # More IPv6 from http://www.ietf.org/rfc/rfc2732.txt
-            'http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:8000/index.html': ('http', '[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]', 8000),
-            'http://[1080:0:0:0:8:800:200C:417A]/index.html': ('http', '[1080:0:0:0:8:800:200C:417A]', None),
+            'http://[fedc:ba98:7654:3210:fedc:ba98:7654:3210]:8000/index.html': ('http', '[fedc:ba98:7654:3210:fedc:ba98:7654:3210]', 8000),
+            'http://[1080:0:0:0:8:800:200c:417a]/index.html': ('http', '[1080:0:0:0:8:800:200c:417a]', None),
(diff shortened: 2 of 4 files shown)


haikuginger:
I'm 👍; RFC-7230 affirms that scheme and host are case-insensitive.

Lukasa:
This looks reasonable enough to me, but @sigmavirus24 is our resident URL expert, so I'd love it if he weighed in here before I merged this.

sigmavirus24:
@haikuginger the correct RFC to reference is 3986. That said, I'd feel more comfortable if we had tests affirming that the userinfo isn't changed ever. That part *is* case-sensitive. (I think urllib3 calls this part 'auth' as in authentication, not authority as you'll find the (userinfo@host:port) portion called in 3986.)

haikuginger:
> @haikuginger the correct RFC to reference is 3986.

Thanks! I think I might have confused the relationship of RFC-7230 to 3986 with the relationship of 3986 to 2396.

I agree that a test to ensure that other URL components aren't ever modified is a good idea.

jeremycline:
I shall make it so. Thanks for the feedback!

jeremycline:
@sigmavirus24 Does 90fa353 cover everything?

sigmavirus24:
Thanks @jeremycline! Looks good to me. :+1:

Lukasa:
Hurrah, thanks both! :sparkles: :cake: :sparkles: @jeremycline Do you mind updating CHANGES with this? Frustratingly I have to take an extra step to do it myself because I don't have admin on this repo, so it's easier if your PR does it for me. =D

jeremycline:
@Lukasa No problem. I've rebased, squashed the two commits for tests into one, and added a changelog entry. Let me know if the changelog isn't what you had in mind and I'll be happy to tweak it as necessary.

Lukasa:
Hurrah, thanks @jeremycline! :sparkles:

PR #915, opened 2016-07-04
Pointing flake8 specifically at the urllib3 package

haikuginger:
Fixes #914.

PR #916, opened 2016-07-05
Undo #915 and apply better solution

haikuginger:
This implements the change discussed in #915.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps=
     flake8
 commands=
     flake8 --version
-    flake8 urllib3
+    flake8 setup.py docs dummyserver urllib3
 
 [testenv:docs]
 basepython = python2.7


Lukasa:
I'm happy with this, but given that @sigmavirus24 is the expert, I'll lean on his review here. ;)

haikuginger:
We certainly wouldn't want to prematurely merge a solution without his advice! ;)

sigmavirus24:
I don't think this will fix your problem @haikuginger. Did you so something like:

```
$ git clone urllib3
$ cd urllib3
$ virtualenv .
$ . bin/activate
$ flake8
```

If so, you want `*site-packages*`. I can't really be certain what virtualenv you're trying to ignore here though. Are you trying to ignore the ones created by tox? In that case adding simply `.tox` to the end will work. I need more information before I can say whether this is the right tactic.

Lukasa:
@sigmavirus24 The goal is to ignore *all* virtualenvs that may be created. For example, I regularly have a virtualenv of my own installed in most of my project directories.

sigmavirus24:
@Lukasa the point is, some of these virtualenvs will have python scripts installed by pip in places other than `site-packages` which may cause flake8 to fail. `*site-packages*` will catch 70-80% of what you care about but not 100%

haikuginger:
@sigmavirus24: from what it sounds like you're saying, there might not be a comprehensive way to blacklist all virtualenvs from flake8, simply because of the possible amount of variation. If that's the case, should we attempt to cover as many edge cases as we can, or simply go with the "whitelist" that was the original plan of just running on the urllib3 package?

haikuginger:
@sigmavirus24, revisiting this atm.

My understanding from what I can find online is that we can't simply *activate* a Tox virtualenv. So something like that would allow us to immediately run any given command in that environment, but that's all. Which turns the current `make test-all` command into `tox -e venv make test-all`.

Now, given that Tox does create its own venvs for testing, it _may_ (and I'm very hesitant on that; it's for smarter people than me to say) be not entirely unreasonable to waive the "be inside a venv first" requirement from the makefile, in which case this is all rather superfluous, as there'd be no reason for a developer to have a venv inside the repo in the first place.

However.

If that's _not_ an option (and like I said, it's not for me to say), then we still want a solution, which will, by nature, either come through setting up explicitly allowed directories (ugly, but relatively easy; going through `flake8 -v`, that looks like `flake8 urllib3 dummyserver setup.py docs`), or by disallowing other ones (seems like `*site-packages*` would _generally_ work, but may miss some edge cases).

Thoughts?

Lukasa:
I definitely want to support the current workflow of having a custom virtualenv (or more than one) present outside the `.tox` directory.

haikuginger:
> I definitely want to support the current workflow of having a custom virtualenv (or more than one) present outside the .tox directory.

I would tend to agree. Should we go the ugly, working way, then?

sigmavirus24:
> I would tend to agree. Should we go the ugly, working way, then?

As long as the ugly working way is in line with how things worked before, and not just a narrow scope on urllib3, I can live with that.

haikuginger:
Should be all set.

Lukasa:
I'm definitely happy: @sigmavirus24?

sigmavirus24:
Thanks @haikuginger! :custard:

PR #920, opened 2016-07-07
Fix doc syntax in user-guide.rst

qwazerty:
This PR fixes a syntax error in `user-guide.rst` documentation.


Lukasa:
Thanks @qwazerty! :sparkles: :custard:

PR #922, opened 2016-07-11
Add support for IP address SAN fields.

### LINKED ISSUE #258: support ip address typed subject alterative names when connecting to an ip address using https
According to rfc6125 section 3.1.3.2 subjectAltName values of type iPAddress should be considered when the https connection is made directly to an ip address rather than a dns domain name. If a server uses such a certificate, urllib3 will fail to validate it, however other tools such as curl and go does. go even refuses to consider CN when verifying the certificate when connected by the 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1333451-support-ip-address-typed-subject-alterative-names-when-connecting-to-an-ip-address-using-https?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #258.

This is a complex and substantial change, and I'd like a lot of feedback from the other reviewers, as well as @shazow. There are several possible options for supporting this, and this PR is the most radical of the set. However, we may not want to commit to going this far: several of the partial solutions may provide enough value to be worthwhile.

The effect of this pull request is to ensure that, on all Python versions, IP addresses in the subject alternative name field of a TLS certificate can be used to validate the certificate. This is a stark improvement over the status quo, where this only works on Python 3.5 out-of-the-box. However, to achieve it we need to make the following changes:

- backport a newer copy of match_hostname from Python 3.5
- backport and vendor the ipaddress module
- adjust the backported copy of match_hostname to use our vendored copy of ipaddress if it can't find a proper one
- prefer our backported copy of match_hostname to the versions available in Python 2.7.9 through Python 3.4

This is pretty sweeping: in general we've aimed to use as much of the infrastructure that was present in Python as we could, but this change removes that again. It also means that we have not just backported modules but, in the case of match_hostname, are now carrying our patches to it (albeit [only a small one](https://github.com/shazow/urllib3/commit/704fcac89cb0026a290228013ccedbe8795c3f97)).

We could also do the following, less extensive things. In order from smallest to largest:

1. Change our logic to prefer the external match_hostname module from PyPI over the installed version on Python < 3.5, and then add `backports.ssl_match_hostname` and `ipaddress` to the `secure` extra.
2. Do 1, but also do a backport of the newer `match_hostname`. Don't backport `ipaddress`. This means that we still need people to install `ipaddress` to get the IP SANs, but means we don't have to vendor that 2k LoC module.
3. Do what I've done here.

I'm very much interested in what people think here. The version I've done is the most complete and has the effect of "just working" in all cases, but is also a really substantial change. I'd be interested in seeing people's opinions as to which of these approaches they like best.

One other note: because we're vendoring new packages, we probably need to alert any downstream Linux distributors because they may want to pull these packages out and make them hard dependencies instead. Do we know who our downstream repackagers are for urllib3?

/cc @sigmavirus24 @haikuginger @jonparrott @paultiplady @fasaxc


Lukasa:
Please all note, as well, that this doesn't work when PyOpenSSL is in use. I'll make appropriate patches to PyOpenSSL as-and-when we've sorted out how we're going to do this.

urllib3/packages/ssl_match_hostname/_implementation.py
```diff
@@ -64,6 +76,23 @@ def _dnsname_match(dn, hostname, max_wildcards=1):
...
+    """
+    # OpenSSL may add a trailing newline to a subjectAltName's IP address
+    # Divergence from upstream: ipaddress can't handle byte str
+    ip = ipaddress.ip_address(_to_unicode(ipname).rstrip())
+    return ip == host_ip
+
+
 def match_hostname(cert, hostname):
     """Verify that *cert* (in decoded format as returned by
     SSLSocket.getpeercert()) matches the *hostname*.  RFC 2818 and RFC 6125
```
haikuginger: IP addresses are accepted for `hostname` now! Huzzah!

haikuginger:
Having the feature "just work" across the board is huge, and overall this looks great. I haven't reviewed `ipaddress` since it's a straight vendor, but the modifications elsewhere all look fine to me.

On the matter of "how should we do this", I'm somewhat ambivalent between options 2 and 3. It looks like you've made some substantial modifications to the `ssl_match_hostname` module to make it work in this framework, so I think we should keep that vendored version.

urllib3/packages/ssl_match_hostname/__init__.py
```diff
@@ -1,5 +1,11 @@
+import sys
+
 try:
-    # Python 3.2+
+    # Our match_hostname function is the same as 3.5's, so we only want to
+    # import the match_hostname function if it's at least that good.
+    if sys.version_info < (3, 5):
```
haikuginger: Question here. This bypasses our modifications to `match_hostname` on 3.5 and higher. The two main features that we get from those modifications (as far as I can tell) are 1) ability to handle a missing `ipaddress` module and 2) ability to handle bytestring IP addresses.

Those *should* be non-issues, as >3.5 will never be missing the `ipaddress module` (it's stdlib) and only Python 2 will have bytestrings (Python 3 will always do Unicode strings).

Is that about the shape of it?

Lukasa: Not quite. We aren't carrying custom modifications to `match_hostname`, we're carrying basically the Python 3.5 `match_hostname` function with extra code to handle a missing `ipaddress` module. Basically, we're now behaving as though the Python 3.5 `match_hostname` function is the oldest version we will allow.

jonparrott:
Agreed with @haikuginger, having this just work is great. How we get this is a bit fuzzy.

I'm okay with adding a pure-python dependency the `secure` extra, so option 2 seems reasonable. The modifications to `match_hostname` seem significant enough to warrant not depending on the backported version.

haikuginger:
If we go with option 2, I'm not 100% familiar with the `backports` process; will we depend specifically on the backport as it exists right now, or will it be a `>=` depend?

Lukasa:
To be clear, in option 2 when I said we'd backport match_hostname, I meant "vendor". Sorry.

haikuginger:
Whoops, I didn't say what I meant to say - my question was specifically around best practices for the dependency on the backported `ipaddress` package.

Lukasa:
@haikuginger We'd probably want to check how it's versioned, and then make a decision. Ideally we want to allow bugfixes but nothing that changes the API.

jeremycline:
@ralphbean maintains urllib3 in Fedora, I believe.

For what it's worth, I lean towards 1 more than 2 or 3, but I do see the appeal of vendoring and having it Just Work.

Lukasa:
@ralphbean Do you have thoughts here?

ralphbean:
Hi, sorry I'm late to the party!  Downstream packager reporting for duty.

- Option 1 is easiest on me, although I acknowledge that its less that desirable for the vast majority of people using urllib3 off of pypi.  "Secure" should be default, not something users have to turn on.  However, if there is already a ``secure`` extra, then it wouldn't be too bad to add it there.
- Option 2 is definitely workable.
- Same with option 3.  The more new libraries that get vendored, the more work there is on downstream re-packagers to un-vendor them.  Besides which, we end up with confusion all around.

I guess option 2 is my favorite in terms of balance between UX and maintenance burden.

Lukasa:
Ok, here's my plan.

I am going to push a new urllib3 release sometime this week. I don't want to rush this in to that release, so I'll hold off until later. However, the overwhelming consensus seems to be that option 2 is the lesser of two evils, so when I swing back to this I'll implement option 2 and remove the vendored `ipaddress` module, while leaving in `match_hostname`.

setup.py
```diff
@@ -60,6 +60,9 @@
               'idna>=2.0.0',
               'certifi',
           ],
+          'secure:python_version <= "2.7"': [
```
sigmavirus24: Is it

```
secure:python_version
```

or

```
secure;python_version
```

I don't remember if it's a semi-colon or a colon.

Lukasa: Oh god, me either. @dstufft?

dstufft: For this syntax in ``extra_requires`` I believe it's a ``:``.

Lukasa: Ok, I think that setuptools is at fault here.

Lukasa:
Ok, this is finally in a mergeable state! Wooooo!

Contributors, @ralphbean, do you want to get a final look at this before I aim to merge early next week?

setup.py
```diff
@@ -60,6 +60,9 @@
               'idna>=2.0.0',
               'certifi',
           ],
+          'secure:python_version < "3.3"': [
```
sigmavirus24: A few things here:

1. The python version here does not match the one in setup.cfg
2. As you might note, in setup.cfg `secure` has a couple of `python_version` checks but we don't replicate them here (specifically because of concerns for older versions of setuptools (which is exactly what you ran into).
  Maybe we should not be catering to a slightly larger number of setuptools versions, but I'm not sure being usable on those older versions is a bad thing either.

Lukasa:
Ok, with no further feedback I'm going ahead and merging this. Obviously, feedback post-merge is also totally acceptable. Thanks all for all your hard work, I'm delighted to see this land!

PR #926, opened 2016-07-15
Update connectionpool.py

springjools:
Update debug print to proper DEBUG log level

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -210,7 +210,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         Return a fresh :class:`HTTPConnection`.
         """
         self.num_connections += 1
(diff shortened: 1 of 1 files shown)


haikuginger:
It looks good to me.

springjools:
Hmm, I just edited it in github editor, forgot that python is so picky with tabs. Should I fix and resubmit the PR ?

jeremycline:
It _is_ quite noisy if you're working with a server that doesn't do keep-alive. 

Personally I like it at the info level and it's helped me quickly recognize when a server isn't configured as it should be, but I'm okay turning on debug logs for this kind of this.

Lukasa:
@springjools There's no need to resubmit: you can just make further commits on this branch. If you find your branch on GitHub you can edit inline and just make a new commit to fix up the indentation. =)

Lukasa:
Closed in favour of #967.

Issue #928, opened 2016-07-17
steam(None) can result in an infinite loop


nateprewitt:
Currently calling `stream(None)` can result in an `HTTPResponse` trying to read a response forever. 

This occurs [here](https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L361-L365) because `is_fp_closed` won't evaluate to `True` in certain cases. The below test provided by @Lukasa shows the issue in our current implementation of `is_fp_closed`. Currently, in Python 3,  `_close_conn` does not call `super().close()`, which means that the `closed` attribute will never be set. We can fix this by checking `isclosed()` first, if it exists, which will appropriately evaluate to True when the contents of the stream have been completely read.

See related discussion in [kennethreitz/requests#3421](https://github.com/kennethreitz/requests/issues/3421)

```python
class TestStream(SocketDummyServerTestCase):
    def test_stream_none_unchunked_response_does_not_hang(self):
        done_event = Event()

        def socket_handler(listener):
            sock = listener.accept()[0]

            buf = b''
            while not buf.endswith(b'\r\n\r\n'):
                buf += sock.recv(65536)

            sock.send(
                b'HTTP/1.1 200 OK\r\n'
                b'Content-Length: 12\r\n'
                b'Content-type: text/plain\r\n'
                b'\r\n'
                b'hello, world'
            )
            done_event.wait(5)
            sock.close()

        self._start_server(socket_handler)
        pool = HTTPConnectionPool(self.host, self.port, retries=False)
        r = pool.request('GET', '/', timeout=1, preload_content=False)

        # Stream should read to the end.
        self.assertEqual([b'hello, world'], list(r.stream(None)))

        done_event.set()
```

tobinus:
Thank you for taking your time to address this, @nateprewitt and @Lukasa! :tada:

PR #929, opened 2016-07-17
fixing infinite loop when stream(None) called

nateprewitt:
Fixing the issue discussed in #928. Added in test from @Lukasa and reworked `is_fp_closed` to prefer `isclosed` first if available.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -25,6 +25,8 @@ dev (master)
   now use cryptography and idna, both of which are already dependencies of
   PyOpenSSL. (PR #930)
 
+* Fixed infinite loop in ``stream`` when amt=None. (Issue #928)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -208,6 +208,7 @@ In chronological order:
 
 * Nate Prewitt <nate.prewitt@gmail.com>
   * Ensure timeouts are not booleans and greater than zero.
+  * Fixed infinite loop in ``stream`` when amt=None.
 
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]


urllib3/util/response.py
```diff
@@ -12,18 +12,21 @@ def is_fp_closed(obj):
...
+    closed = getattr(obj, 'closed', None)
+    if closed is not None:
         # Check via the official file-like-object way.
-        return obj.closed
-    except AttributeError:
-        pass
+        return closed
 
-    try:
+    fp = getattr(obj, 'fp', False)
```
nateprewitt: This is a little ugly. Perhaps we should stick with the `try`/`except` method, but I was trying to adhere to the discussion @Lukasa and I had on a previous PR that `getattr` may be preferred to `try`/`except`.

urllib3/util/response.py
```diff
@@ -12,18 +12,21 @@ def is_fp_closed(obj):
...
+    if fp is not False:
```
nateprewitt: This is a little ugly. Perhaps we should stick with the try/except method, but I was trying to adhere to the discussion @Lukasa and I had on a previous PR that getattr may be preferred to try/except.

Lukasa: Oh sorry, I wasn't clear in the other issue. In the other issue I was nominating `getattr` because we were going multiple levels deep, which means that you can more easily get non-`AttributeError` exceptions. Here, please maintain the current style. =)

Lukasa: And as a future tip, it is almost always preferable, when adding function, to maintain the current style unless expressly directed not to. Style changes should ideally be done in PRs that don't change function, so that tests can accurately validate the safety of the change. =)

nateprewitt: Ok, good to know, thanks :) Reverting now.

Lukasa:
Thanks @nateprewitt! Two small notes and then we can merge this:

1. Can you add a line to CHANGES?
2. Can you, in the comment, refer to the urllib3 issue in question?

sigmavirus24:
Provided you address @Lukasa's review, this looks good to me.

nateprewitt:
Alright, should be good to go.

urllib3/util/response.py
```diff
@@ -13,6 +13,12 @@ def is_fp_closed(obj):
     """
 
     try:
+        # Check `isclosed()` first, in case Python3 doesn't set `closed`.
```
Lukasa: Sorry, can you update this comment to link to the relevant urllib3 issue as well?

nateprewitt: Apologies, thought you were referring to the commit message. Comment updated.

nateprewitt:
Looks like Travis may need a quick kick.

nateprewitt:
Another quick restart on Travis and this should be good to go.

Lukasa:
Sweet, looks good. Thanks @nateprewitt! :sparkles: :custard:

PR #930, opened 2016-07-19
Remove extra dependencies for PyOpenSSL.

Lukasa:
Currently to use the pyopenssl contrib module we need to install three dependencies: pyopenssl, ndg-httpsclient, and pyasn1. The latter two dependencies exist *only* to give us the ability to handle the Subject Alternative Name X.509 extension, and are otherwise entirely unrequired.

This PR changes that logic to use cryptography instead. Because PyOpenSSL already depends on cryptography, this has the effect of reducing our actual dependency graph. We still, at a top level, have three dependencies (annoyingly we need `idna`, blame @reaperhulk), but those three top-level dependencies are *already present* in our dependency graph transitively (`pyopenssl` requires `cryptography` requires `idna`), so this has the effect of reducing the number of packages we rely on directly. We also have much better ties to the maintainers of those packages than we do to the maintainers of ndg-httpsclient and pyasn1.

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -255,7 +255,7 @@ certificate verification on Python 2 will be installed::
 If you want to install the packages manually, you will need ``pyOpenSSL``,
 ``cryptography``, ``idna``, and ``certifi``.
 
-.. note:: If you are not using Mac OS X or Windows, note that `cryptography
+.. note:: If you are not using macOS or Windows, note that `cryptography
     <https://cryptography.io/en/latest/>`_ requires additional system packages
     to compile. See `building cryptography on Linux
     <https://cryptography.io/en/latest/installation/#building-cryptography-on-linux>`_
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -141,7 +141,7 @@ JSON
 ~~~~
 
 You can sent JSON a request by specifying the encoded data as the ``body``
-argument and setting the ``Content-Type`` header when calling 
+argument and setting the ``Content-Type`` header when calling
 :meth:`~poolmanager.PoolManager.request`::
 
     >>> import json
@@ -210,7 +210,7 @@ and most reliable method is to use the `certifi <https://certifi.io/en/latest>`_
 
 You can also install certifi along with urllib3 by using the ``secure``
 extra::
-    
+
     pip install urllib3[secure]
 
 .. warning:: If you're using Python 2 you may need additional packages. See the :ref:`section below <ssl_py2>` for more details.
@@ -253,11 +253,11 @@ certificate verification on Python 2 will be installed::
     pip install urllib3[secure]
 
 If you want to install the packages manually, you will need ``pyOpenSSL``,
-``ndg-httpsclient``, ``pyasn``, and ``certifi``.
+``cryptography``, ``idna``, and ``certifi``.
 
-.. note:: If you are not using Mac OS X or Windows, note that pyOpenSSL depends
-    on `cryptography <https://cryptography.io/en/latest/>`_ which requires
-    additional system packages to compile. See `building cryptography on Linux
+.. note:: If you are not using Mac OS X or Windows, note that `cryptography
+    <https://cryptography.io/en/latest/>`_ requires additional system packages
+    to compile. See `building cryptography on Linux
     <https://cryptography.io/en/latest/installation/#building-cryptography-on-linux>`_
     for the list of packages required.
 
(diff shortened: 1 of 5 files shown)


docs/user-guide.rst
```diff
@@ -253,11 +253,11 @@ certificate verification on Python 2 will be installed::
     pip install urllib3[secure]
 
 If you want to install the packages manually, you will need ``pyOpenSSL``,
-``ndg-httpsclient``, ``pyasn``, and ``certifi``.
+``cryptography``, ``idna``, and ``certifi``.
 
-.. note:: If you are not using Mac OS X or Windows, note that pyOpenSSL depends
-    on `cryptography <https://cryptography.io/en/latest/>`_ which requires
-    additional system packages to compile. See `building cryptography on Linux
+.. note:: If you are not using Mac OS X or Windows, note that `cryptography
```
haikuginger: Minor nit, but we should either use the present brand name (OS X), or the new brand name for this year (macOS). Probably the latter. Mac OS X was deprecated in 2012 with the Mountain Lion release.

Lukasa: Eh, yeah, let's update our docs to say macOS.

urllib3/contrib/pyopenssl.py
```diff
@@ -40,15 +43,12 @@
...
-    from ndg.httpsclient.subj_alt_name import SubjectAltName as BaseSubjectAltName
-except SyntaxError as e:
-    raise ImportError(e)
-
+import idna
 import OpenSSL.SSL
-from pyasn1.codec.der import decoder as der_decoder
-from pyasn1.type import univ, constraint
+from cryptography import x509
+from cryptography.haztmat.backends.openssl import backend as openssl_backend
```
haikuginger: `haztmat`? Typo?

urllib3/contrib/pyopenssl.py
```diff
@@ -58,16 +58,18 @@
...
+import sys
 
 from .. import util
 
 __all__ = ['inject_into_urllib3', 'extract_from_urllib3']
 
-# SNI only *really* works if we can read the subjectAltName of certificates.
-HAS_SNI = SUBJ_ALT_NAME_SUPPORT
+# SNI always works.
+HAS_SNI = True
```
haikuginger: Huzzah!

urllib3/contrib/pyopenssl.py
```diff
@@ -123,46 +128,68 @@ def extract_from_urllib3():
...
-                component = name.getComponentByPosition(entry)
-                if component.getName() != 'dNSName':
-                    continue
-                dns_name.append(str(component.getComponent()))
+    """
+    Given an PyOpenSSL certificate, provides all the subject alternative names.
+    """
+    # Pass the cert to cryptography, which has much better APIs for this.
+    # This is technically using private APIs, but should work across all
+    # relevant versions until PyOpenSSL gets something proper for this.
```
haikuginger: Does PyOpenSSL have an open issue asking for this feature? I know you're on the team over there, but it'd still be good to have some explicit tracking.

haikuginger:
After a twice-over, this looks pretty good to me! A couple minor nits, but fewer dependencies is always nice to have. 🙂

urllib3/contrib/pyopenssl.py
Lukasa: Yup, and a typo that meant that our tests weren't running on this code. Good spot!

urllib3/contrib/pyopenssl.py
Lukasa: Maybe pyca/pyopenssl#439, I need to check.

haikuginger: Sounds good! I just want to make sure we have a place to look to know where we stand. 👍

setup.py
```diff
@@ -55,8 +55,8 @@
       extras_require={
           'secure': [
               'pyOpenSSL>=0.13',
-              'ndg-httpsclient',
-              'pyasn1',
+              'cryptography>=1.3.4',
```
sigmavirus24: Did I miss it or did you not update `setup.cfg` with the metadata for this?

sigmavirus24:
Pending Travis, :+1:

haikuginger:
Does Travis seem to flake out a lot more than it used to? Seems like we have to do a re-test on pretty much every PR at this point.

Lukasa:
@haikuginger Maybe? Just means we need to spend some more time going through and fixing up events and stuff.

haikuginger:
I'm good with the current changeset; @Lukasa, do you want to add a line item to `CHANGES.rst`?

Lukasa:
@haikuginger Done.

haikuginger:
Thanks @Lukasa! ✨🍮✨

PR #935, opened 2016-07-26
Use LifoQueue, Empty, and Full from six

scop:


Lukasa:
Sorry I missed this!

Your imports aren't working because urllib3 doesn't depend on six, it vendors it. Can you change the imports to be package-relative?

scop:
I think they should work as-is because there's this above those import lines:
```python
from .packages import six
```

If it didn't work, I'd assume the test suite/CI build would break. Well it did actually break, but for some other reasons besides these changes I gather. Rebased against master and force-pushed, let's see how it goes now.

Lukasa:
As far as I know that's not how Python imports work: simply having brought a name into the namespace does not change its import path. You'll still need `from .packages.six import x`.

scop:
Ok, the tox dependency pulled `six` in, that's why it "worked".

The imports should be sorted out now, however builds fail because tests for some reason started to want to see more coverage for `http.client` and `queue` (`httplib` and `Queue` on py2) -- the stuff pulled in under the hood by `six.moves`. I don't know how to prevent that. Maybe a newer version of coverage would help?

Lukasa:
Hrm. I can't really work out why that happened, but why not add them to the `omit` section of `.coveragerc`?

scop:
What do you think should be put there? I gather `omit` is a list of paths, and where the affected core Python modules reside kind of depends... and even trying to specify paths to them feels something's not going the way it should be.

haikuginger:
If it works for @Lukasa, you could also create an `include` section in `.coveragerc`, and add `urllib3` to that. I've actually run into this before, and the above is part of the SessionManager branch I'm working on.

Lukasa:
@haikuginger Ah! I knew I'd seen us do that somewhere and I couldn't remember where it was. Good call, do that. =)

Lukasa:
Cool, looks good to me. Thanks @scop! :custard:

PR #936, opened 2016-07-26
Test with PyPy3 too

scop:


Lukasa:
Thanks! However, PyPy3 is Python 3.2, which we don't support.

scop:
PyPy3 v5.2* which this patch adds is Python 3.3, please reconsider.

scop:
Some green proof here: https://travis-ci.org/scop/urllib3/builds/147497199

Lukasa:
I apologise, you're quite right!

haikuginger:
PyPy3 5.2 is still in initial alpha; do we want to hold off on this (at least, publicizing compatibility) until it reaches release?

Lukasa:
Certainly while it's in alpha I'd like to be cautious.

Lukasa:
Let's hold off on this for now until we get to a fuller release.

PR #937, opened 2016-07-26
use dunder slots for Url class slots variable

kalefranz:
From the python [`namedtuple`](https://docs.python.org/2.7/library/collections.html#collections.namedtuple) documentation:

> The subclass shown above sets `__slots__` to an empty tuple. This helps keep memory requirements low by preventing the creation of instance dictionaries.

Was this the intent for the `slots` variable in `Url` class also?


Lukasa:
Yup, good catch!

PR #938, opened 2016-07-29
Use tox-travis for nightly, PyPy 5.3 and PyPy3 5.2

jayvdb:
Replaces custom PyPy 4.0.1 build with Travis supplied PyPy 5.3,
removing the need for two bash Travis specific scripts.

Adds Travis supplied PyPy3 5.2-alpha and nightly build jobs.

Uses tox-travis to simplify running the testenv only for the
specified Travis environment without needing to set TOXENV
and install specific versions of virtualenv.


jayvdb:
This is partially an alternative to #936 , but it also uses the existing Travis functionality to reduce the build times. e.g. without the two extra environments, the [build time] (https://travis-ci.org/jayvdb/urllib3/builds/148197967) is "Elapsed time 2 min 5 sec ; Total time 7 min 18 sec" , compared to master's "Elapsed time 2 min 52 sec; Total time 11 min 59 sec"

I'm happy to re-add 4.0.1 if it is a vital part of testing, however if so , I would suggest adding testing of PyPy 4.0.1 on Appveyor instead of Travis, as Appveyor doesnt have any PyPy so the install time is similar whether it is installing PyPy 4.0.1 or a different version.

Lukasa:
I'd rather we didn't use tox-travis: instead, it'd be better to just change the pyenv script to install the appropriate PyPy version. My question is: since when is Travis supplying pypy 5.3? If they are, we can remove the fallback: it was only present to cover the time when Travis were supplying ancient PyPy versions.

jayvdb:
> My question is: since when is Travis supplying pypy 5.3?

17 days ago
https://github.com/travis-ci/travis-ci/issues/6277#issuecomment-231856352

Travis also uses pyenv, so the local script is just replicating part of the Travis system.  The benefit of the local script is it can support any pyenv target version, while Travis only allows specific pyenv targets that they have pre-built and tested, so that it is fast.

Lukasa:
Ok, so in that case let's throw away our fallback. However, let's still not use tox-travis: let's just configure Travis appropriately so it gives us a useful PyPy.

PR #939, opened 2016-07-29
Use Travis supplied PyPy 5.3

jayvdb:
Replaces custom PyPy 4.0.1 build with Travis supplied PyPy 5.3,
removing the need for two bash Travis specific scripts.

Improve fetch_gae_sdk.py to simplify the invocation in .travis.yml


Lukasa:
Sweet, thanks so much! :custard:

PR #941, opened 2016-08-01
Use OS default certs when possible

krrr:
It seems that urllib3 used to use ssl.create_default_context when possible. Isn't it more convenient to have default certs?

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -11,7 +11,6 @@ from ..exceptions import SSLError, InsecurePlatformWarning, SNIMissingWarning
 
 SSLContext = None
 HAS_SNI = False
-create_default_context = None
 IS_PYOPENSSL = False
 
 # Maps the length of a digest to a possible hash function producing this digest
@@ -304,6 +303,9 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
             if e.errno == errno.ENOENT:
                 raise SSLError(e)
             raise
+    elif getattr(context, 'load_default_certs', None) is not None:
+        # try to load OS default certs; works well on Windows (require Python3.4+)
+        context.load_default_certs()
 
     if certfile:
         context.load_cert_chain(certfile, keyfile)


Lukasa:
Thanks for this!

I am in principle ok with doing this. However, right now it's overbroad. I have some notes:

1. You test for `OS_CERTS_AVAILABLE` by searching for `create_default_context`, but you don't use that function at all. Probably you want to search for `load_default_certs`.
2. The mere presence of `load_default_certs` doesn't actually signify that the OS certs are available. That works completely on only a partial set of *nix systems, and does not work for some Linuxen and most OS X systems. It partially works on Windows. That means we probably don't want to use that to force `CERT_REQUIRED`: it can lead to previously deliberately insecure connections being forced to be secure and fail, without easy recourse for the user. I'd be more inclined to want to call `load_default_certs` but add no further code to force validation.
3. Automatically loading the OS certs can lead to unexpected behaviour: in particular, if it's combined with a custom cert bundle, it can accidentally lead to *over*-trusting by trusting the OS certs as well as the ones provided by the user. For that reason, I'm inclined to want this code to actually go in to `ssl_wrap_socket` as an `elif` to `if ca_certs or ca_cert_dir`.

Do these sound reasonable to you?

krrr:
(1) and (2) are very reasonable. I haven't even considered OS other than Windows. But I can't understand (3), what is "over-trusting"?

Lukasa:
@krrr It's not a technical term, I just improvised it. ;) But it amounts to this:

Imagine creating a `PoolManager` and providing a `ca_certs=/path/to/bundle.pem` argument. That bundle contains one certificate, the root certificate used by your company. You then use `.request()` to make a web request. In the current code, the destination host *must* return a certificate chain that chains up to the cert in `bundle.pem`. In the change made here, that chain may *also* chain up to any root in the OS cert store. That's what I mean by "over-trusting": the user only wanted to trust the root in `bundle.pem`, but they ended up trusting everything their OS trusted. With TLS is it almost always better to fail closed in the face of ambiguity: I'd rather the request failed and the user had to debug why than have the request succeed when a user would have wanted it to fail.

Basically, loading root certs into OpenSSL is additive: they don't replace previously loaded roots. For that reason, I'd rather that root certs get loaded *only* when the user hasn't provided their own, at least in the default path. A user who wants that additive behaviour can call `create_urllib3_context` themselves and make the appropriate change.

krrr:
@Lukasa Thanks for quick and detailed reply! I have amended my commit

urllib3/util/ssl_.py
```diff
@@ -11,7 +11,6 @@
 
 SSLContext = None
 HAS_SNI = False
-create_default_context = None
```
sigmavirus24: Why is this removed?

krrr: @sigmavirus24 because it is not longer used?

do you mean this useless var should be kept for backward compatibility?

Lukasa:
Ok, so I'm broadly happy with this. It'd be good if you could rebase on the newest `master`, please, because that'll allow us to run our automated tests. Otherwise, once @sigmavirus24 is satisfied with your response, any contributor may merge this.

Can whatever contributor merges this please remember that we need a CHANGES file entry for this? Thanks! :custard:

urllib3/util/ssl_.py
Lukasa: @sigmavirus24 can you follow up on this? =)

sigmavirus24: I have a strong distaste for changes that include irrelevant changes in them, but looking over the file before this change, we're not using the constant anywhere anyway.

PR #942, opened 2016-08-01
Allow PyPy 5.3 to fail

sigmavirus24:
Since it's addition, our Travis builds have become very unstable and
this job seems to be implicated in most of those builds. Until we can
stabilize this particular build job, let's allow it to fail to avoid
having to continuously retry the build on Travis.


haikuginger:
Thanks @sigmavirus24! More ✅! Less ❌!

PR #946, opened 2016-08-12
Import more from six

scop:


sigmavirus24:
@scop have you found everything that can be imported from six?

@Lukasa @haikuginger if the answer to ^ is no, can we hold off until this is completed?

haikuginger:
I want this badly, but, like @sigmavirus24, I also want it done once and for all. @scop, if you haven't already, do you want to try searching the repo for any other instances of `except ImportError` that can be replaced with `six` helpers?

scop:
This commit was done by looking at the output of `git grep -B 4 ImportError`, I think I caught all of what was there. I just did another run with `git grep -B 4 except:` and there's one more in it, will force-push in a jiffy.

haikuginger:
Okay; in that case I'm 👍. @sigmavirus24, @Lukasa, any other thoughts?

@scop, if you could go ahead and rebase on master, that'd be great.

Lukasa:
I'm happy. We can merge when we get the rebase. =D

scop:
Rebased.

Lukasa:
Nice, thanks @scop! :custard:

PR #947, opened 2016-08-15
Add support for ChaCha20.

Lukasa:
ChaCha20 is coming in OpenSSL 1.1.0, and it's generally pretty excellent. OpenSSL kindly ignores requests for ciphers that it doesn't support, so we can unconditionally update our code to request ChaCha20 and expect OpenSSL to just quietly ignore the change when it doesn't know what ChaCha is.

This change has us prefer AES-GCM over ChaCha20. This decision is made mostly because newer processors have AES-NI hardware acceleration instructions, and right now we're not in a good place to query whether or not those instructions are present, so we just assume that they are. However, [CPython issue 27766](https://bugs.python.org/issue27766), in addition to making this change, also provides a private function that could be queried for AES-NI support. We may want to wait until CPython 3.6 drops to see if that function makes it in, because we could use it to be smarter about our cipher string.


Lukasa:
@haikuginger Done.

haikuginger:
Okay; I'm :+1:. @sigmavirus24?

sigmavirus24:
@Lukasa can you add a release note for this? Otherwise I'm 👍

Lukasa:
Release note added.

sigmavirus24:
Thanks! I'll merge once Travis is happy

haikuginger:
Whoops, didn't see you wanted to pull the trigger @sigmavirus24! Thanks very much @Lukasa! ✨🍻✨

Issue #723, opened 2015-10-19
Add a way to check if number of bytes received matches declared transfer length(s)


causton81:
This discussion originated on python requests (https://github.com/kennethreitz/requests/pull/2833). Basically, there are edge cases where a socket can appear to have closed normally (end-of-file) but was actually reset before sending/receiving all of a response. This may happen silently without raising any exceptions from urllib3.

A suggestion was made to allow users to enable a "strict transfer-length" check in urllib3. For what it's worth RFC2616 (http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4) describes when Content-Length should be used by servers. It seems like the chunked logic is probably safe from this, but probably want to confirm that too.

When I get a chance I'll try to submit a PR for comment.

shazow:
Thanks for transferring the discussion over!

Definitely open to suggestions/proposals to this. I wonder if we could keep this at a relatively higher-level abstraction, like an attribute on RequestMethods and an extra check before returning something.

Lukasa:
Keeping it high level sounds sensible. An extra check should be easy enough, especially as the underlying httplib object actually has exactly the attribute we want (`HTTPResponse.length`, for those interested, indicates how much data remains from the Content-Length header).

nateprewitt:
Hey @Lukasa, I've been playing around with this and `urllib3.response.HTTPResponse` doesn't directly have a `length` attribute. You can access the attr on `HTTPResponse.original_response` but that isn't populated if the `original_response` wasn't from httplib.

Would we want to move the `length` attr up to `response.HTTPResponse`, populating it if it exists on the `original_response`, or computing the value if it doesn't? If not, perhaps it would be better to always compute it and not rely on `length` at all?

Lukasa:
I'm open to always computing it.

nateprewitt:
I believe that this can be closed now that #949 has been merged.

PR #949, opened 2016-08-17
Add `enforce_content_length` for responses

nateprewitt:
So here's a pass at #723. This is kind of a weird edge case but it particularly prominent in the default configuration of Requests. Most calls performed by `urllib3` will raise a `IncompleteRead` error from `httplib` when the number of bytes in the body doesn't match the `Content-Length`. 
# The Skinny

`httplib` raises `IncompleteRead`s appropriately everywhere except on incrementally read data. This is the primary way Requests uses `urlopen` with `preload_content=False` and then reading with `iter_content()`. Retrieving data this way hits the flaw in `httplib`. I've added a flag to enable this functionality, so as not to break `stream(amt)` and `read(amt)` calls presently. In the next major release, I would advise the flag being removed to make all `read` operations uniform by default.

Notes:
- My unfamiliarity with the testing harness is definitely showing in `test_strict_content_length` but the test does prove the changes are working correctly. Tornado won't allow you to send uneven data, so this was the only other solution I could come up with. Any suggestions on alternative methods of simulating this problem would be appreciated.
- ~~I implemented `length` as a property to match the attribute nature of `httplib.HTTPResponse.length`. I realize an int that we modify may be preferred to a property, but felt it would be more likely to break if we implement int updates everywhere IO might happen in the code.~~

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -3,7 +3,9 @@ import unittest
 from io import BytesIO, BufferedReader
 
 from urllib3.response import HTTPResponse
-from urllib3.exceptions import DecodeError, ResponseNotChunked, ProtocolError
+from urllib3.exceptions import (
+    DecodeError, ResponseNotChunked, ProtocolError, InvalidHeader
+)
 from urllib3.packages.six.moves import http_client as httplib
 from urllib3.util.retry import Retry
 
@@ -376,6 +378,73 @@ class TestResponse(unittest.TestCase):
         stream = resp.stream(2, decode_content=False)
 
         self.assertRaises(StopIteration, next, stream)
+
+    def test_length_no_header(self):
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, preload_content=False)
+        self.assertEqual(resp.length_remaining, None)
+
+    def test_length_w_valid_header(self):
+        headers = {"content-length": "5"}
+        fp = BytesIO(b'12345')
+
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        self.assertEqual(resp.length_remaining, 5)
+
+    def test_length_w_bad_header(self):
+        garbage = {'content-length': 'foo'}
+        fp = BytesIO(b'12345')
+
+        resp = HTTPResponse(fp, headers=garbage, preload_content=False)
+        self.assertEqual(resp.length_remaining, None)
+
+        garbage['content-length'] = "-10"
+        resp = HTTPResponse(fp, headers=garbage, preload_content=False)
+        self.assertEqual(resp.length_remaining, None)
+
+    def test_length_when_chunked(self):
+        # This is expressly forbidden in RFC 7230 sec 3.3.2
+        # We fall back to chunked in this case and try to
+        # handle response ignoring content length.
+        headers = {'content-length': '5',
+                   'transfer-encoding': 'chunked'}
+        fp = BytesIO(b'12345')
+
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        self.assertEqual(resp.length_remaining, None)
+
+    def test_length_with_multiple_content_lengths(self):
+        headers = {'content-length': '5, 5, 5'}
+        garbage = {'content-length': '5, 42'}
+        fp = BytesIO(b'abcde')
+
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        self.assertEqual(resp.length_remaining, 5)
+
+        self.assertRaises(InvalidHeader, HTTPResponse, fp,
+                          headers=garbage, preload_content=False)
+
+    def test_length_after_read(self):
+        headers = {"content-length": "5"}
+
+        # Test no defined length
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, preload_content=False)
+        resp.read()
+        self.assertEqual(resp.length_remaining, None)
+
+        # Test our update from content-length
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        resp.read()
+        self.assertEqual(resp.length_remaining, 0)
+
+        # Test partial read
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        data = resp.stream(2)
+        next(data)
+        self.assertEqual(resp.length_remaining, 3)
 
     def test_mock_httpresponse_stream(self):
         # Mock out a HTTP Request that does enough to make it through urllib3's
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1076,3 +1076,68 @@ class TestStream(SocketDummyServerTestCase):
         self.assertEqual([b'hello, world'], list(r.stream(None)))
 
         done_event.set()
+
+class TestBadContentLength(SocketDummyServerTestCase):
+    def test_enforce_content_length_get(self):
+        done_event = Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            buf = b''
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+
+            sock.send(
+                b'HTTP/1.1 200 OK\r\n'
+                b'Content-Length: 22\r\n'
+                b'Content-type: text/plain\r\n'
+                b'\r\n'
+                b'hello, world'
+            )
+            done_event.wait(1)
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = HTTPConnectionPool(self.host, self.port, maxsize=1)
+
+        # Test stream read when content length less than headers claim
+        get_response = conn.request('GET', url='/', preload_content=False,
+                                    enforce_content_length=True)
+        data = get_response.stream(100)
+        # Read "good" data before we try to read again.
+        # This won't trigger till generator is exhausted.
+        next(data)
+        self.assertRaises(ProtocolError, next, data)
+
+        done_event.set()
+
+    def test_enforce_content_length_no_body(self):
+        done_event = Event()
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            buf = b''
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+
+            sock.send(
+                b'HTTP/1.1 200 OK\r\n'
+                b'Content-Length: 22\r\n'
+                b'Content-type: text/plain\r\n'
+                b'\r\n'
+            )
+            done_event.wait(1)
+            sock.close()
+
+        self._start_server(socket_handler)
+        conn = HTTPConnectionPool(self.host, self.port, maxsize=1)
+
+        #Test stream on 0 length body
+        head_response = conn.request('HEAD', url='/', preload_content=False,
+                                     enforce_content_length=True)
+        data = [chunk for chunk in head_response.stream(1)]
+        self.assertEqual(len(data), 0)
+
+        done_event.set()
--- a/urllib3/exceptions.py
+++ b/urllib3/exceptions.py
@@ -1,4 +1,7 @@
 from __future__ import absolute_import
+from .packages.six.moves.http_client import (
+    IncompleteRead as httplib_IncompleteRead
+)
 # Base Exceptions
 
 
@@ -191,6 +194,20 @@ class DependencyWarning(HTTPWarning):
 class ResponseNotChunked(ProtocolError, ValueError):
     "Response needs to be chunked in order to read it as chunks."
     pass
+
+
+class IncompleteRead(HTTPError, httplib_IncompleteRead):
+    """
+    Response length doesn't match expected Content-Length
+
+    Subclass of http_client.IncompleteRead to allow int value
+    for `partial` to avoid creating large objects on streamed
+    reads.
+    """
+    def __init__(self, partial, expected):
+        message = ('IncompleteRead(%i bytes read, '
+                   '%i more expected)' % (partial, expected))
+        httplib_IncompleteRead.__init__(self, message)
 
 
 class InvalidHeader(HTTPError):
--- a/urllib3/exceptions.py
+++ b/urllib3/exceptions.py
@@ -193,6 +193,11 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
+class InvalidHeader(HTTPError):
+    "The header provided was somehow invalid."
+    pass
+
+
 class ProxySchemeUnknown(AssertionError, ValueError):
     "ProxyManager does not support the supplied scheme"
     # TODO(t-8ch): Stop inheriting from AssertionError in v2.0.
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -9,7 +9,7 @@ from socket import error as SocketError
 from ._collections import HTTPHeaderDict
 from .exceptions import (
     ProtocolError, DecodeError, ReadTimeoutError,
-    ResponseNotChunked, InvalidHeader
+    ResponseNotChunked, IncompleteRead, InvalidHeader
 )
 from .packages.six import string_types as basestring, binary_type, PY3
 from .packages.six.moves import http_client as httplib
@@ -97,6 +97,10 @@ class HTTPResponse(io.IOBase):
     :param retries:
         The retries contains the last :class:`~urllib3.util.retry.Retry` that
         was used during the request.
+
+    :param enforce_content_length:
+        Enforce content length checking. Body returned by server must match
+        value of Content-Length header, if present. Otherwise, raise error.
     """
 
     CONTENT_DECODERS = ['gzip', 'deflate']
@@ -105,7 +109,7 @@ class HTTPResponse(io.IOBase):
     def __init__(self, body='', headers=None, status=0, version=0, reason=None,
                  strict=0, preload_content=True, decode_content=True,
                  original_response=None, pool=None, connection=None,
-                 retries=None, request_method=None):
+                 retries=None, enforce_content_length=False, request_method=None):
 
         if isinstance(headers, HTTPHeaderDict):
             self.headers = headers
@@ -117,6 +121,7 @@ class HTTPResponse(io.IOBase):
         self.strict = strict
         self.decode_content = decode_content
         self.retries = retries
+        self.enforce_content_length = enforce_content_length
 
         self._decoder = None
         self._body = None
@@ -383,6 +388,13 @@ class HTTPResponse(io.IOBase):
                     # no harm in redundantly calling close.
                     self._fp.close()
                     flush_decoder = True
+                    if self.enforce_content_length and self.length_remaining not in (0, None):
+                        # This is an edge case that httplib failed to cover due
+                        # to concerns of backward compatibility. We're
+                        # addressing it here to make sure IncompleteRead is
+                        # raised during streaming, so all calls with incorrect
+                        # Content-Length are caught.
+                        raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
 
         if data:
             self._fp_bytes_read += len(data)
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -2,17 +2,21 @@ from __future__ import absolute_import
 from contextlib import contextmanager
 import zlib
 import io
+import logging
 from socket import timeout as SocketTimeout
 from socket import error as SocketError
 
 from ._collections import HTTPHeaderDict
 from .exceptions import (
-    ProtocolError, DecodeError, ReadTimeoutError, ResponseNotChunked
+    ProtocolError, DecodeError, ReadTimeoutError,
+    ResponseNotChunked, InvalidHeader
 )
 from .packages.six import string_types as basestring, binary_type, PY3
 from .packages.six.moves import http_client as httplib
 from .connection import HTTPException, BaseSSLError
 from .util.response import is_fp_closed, is_response_to_head
+
+log = logging.getLogger(__name__)
 
 
 class DeflateDecoder(object):
@@ -100,7 +104,8 @@ class HTTPResponse(io.IOBase):
 
     def __init__(self, body='', headers=None, status=0, version=0, reason=None,
                  strict=0, preload_content=True, decode_content=True,
-                 original_response=None, pool=None, connection=None, retries=None):
+                 original_response=None, pool=None, connection=None,
+                 retries=None, request_method=None):
 
         if isinstance(headers, HTTPHeaderDict):
             self.headers = headers
@@ -136,6 +141,9 @@ class HTTPResponse(io.IOBase):
         encodings = (enc.strip() for enc in tr_enc.split(","))
         if "chunked" in encodings:
             self.chunked = True
+
+        # Determine length of response
+        self.length_remaining = self._init_length(request_method)
 
         # If requested, preload the body.
         if preload_content and not self._body:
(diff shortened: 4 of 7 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1076,3 +1076,49 @@ def socket_handler(listener):
...
+        def _make_connection(self):
+            self._start_server(socket_handler)
+            return HTTPConnectionPool(self.host, self.port, maxsize=1)
+
+        def test_get(self):
+            # Test stream read when content length less than headers claim
+            conn = _make_connection(self)
+            get_response = conn.request('GET', url='/', preload_content=False,
+                                        strict_content_length=True)
+            with self.assertRaises(ProtocolError):
```
nateprewitt: This is currently failing for python 2.6, but made me realize that we aren't catching streams where amt > resp.length. Working on a fix.

Lukasa:
Nate: while you're working on a fix for this I'll hold off from reviewing. No reason beyond the fact that I'm trying to manage my workload!

nateprewitt:
Totally! I think I've got the fix worked out, but doing some final testing. No rush on this at all.

test/test_response.py
```diff
@@ -377,6 +377,39 @@ def test_empty_stream(self):
...
+        resp = HTTPResponse(fp, preload_content=False)
+        resp.read()
+        assert resp.length == None
+
+        # Test our update from content-length
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        resp.read()
+        assert resp.length == 0
+
```
nateprewitt: Maybe test stream here for partial decrementing?

e.g.
```python
data = resp.stream(2)
#read 2 bytes
next(data)
assert resp.length == 3
```

Lukasa:
Restarted.

urllib3/response.py
```diff
@@ -93,14 +93,19 @@ class is also compatible with the Python standard library's :mod:`io`
...
     """
 
     CONTENT_DECODERS = ['gzip', 'deflate']
     REDIRECT_STATUSES = [301, 302, 303, 307, 308]
 
     def __init__(self, body='', headers=None, status=0, version=0, reason=None,
                  strict=0, preload_content=True, decode_content=True,
-                 original_response=None, pool=None, connection=None, retries=None):
+                 original_response=None, pool=None, connection=None,
+                 strict_content_length=False, retries=None, request_method=None):
```
nateprewitt: Should request_method be replaced with **response_kw so we're not defining minimal contact params?

urllib3/response.py
```diff
@@ -327,9 +360,15 @@ def read(self, amt=None, decode_content=None, cache_content=False):
                     # no harm in redundantly calling close.
                     self._fp.close()
                     flush_decoder = True
+                    if (self.strict_content_length and self.length not in (0, None)):
+                        # httplib requires an iterator/string instead of count of bytes
+                        data = [0 for x in range(self._fp_bytes_read)]
```
nateprewitt: This is awkward. This error is used everywhere else this issue occurs though, so it feels off to implement another error to avoid the need of this iterator.

nateprewitt:
Alright, I think this is ready for a glance whenever you have a spare moment @Lukasa.

This still has some rough edges, so I left a few comments inline, as well as my initial comment in the opening post.

Thanks!

urllib3/response.py
```diff
@@ -93,14 +93,19 @@ class is also compatible with the Python standard library's :mod:`io`
     :param retries:
         The retries contains the last :class:`~urllib3.util.retry.Retry` that
         was used during the request.
+
```
nateprewitt: Forgot to add docs for `length` attr. I'll do that in the next pass.

urllib3/response.py
nateprewitt: We might also create an empty shell object with the `__len__` value hardcoded to `self._fp_bytes_read` to avoid creating a possibly gigantic list on large streams? That's also an ugly solution.

Maybe the simplest thing is to just create a local, similarly or identically named, exception that takes two ints instead?

nateprewitt: Added custom Exception in 6321741, it has slightly different semantics (taking int instead of iterator as first arg). This seems like it might cause confusion but avoids creating a possibly huge list on large streams just to reuse the httplib function. I'm +.5 on this and open to alternatives I'm missing.

nateprewitt: Added custom exception in 2015f77, it has slightly different semantics (taking int instead of iterator as first arg). This seems like it might cause confusion but avoids creating a possibly huge list on large streams just to reuse the httplib exception. I'm +.5 on this and open to alternatives I'm missing.

test/test_response.py
```diff
@@ -377,6 +377,59 @@ def test_empty_stream(self):
...
+        resp = HTTPResponse(fp, headers=garbage, preload_content=False)
+        assert resp.length is None
+
+        garbage['content-length'] = "-10"
+        resp = HTTPResponse(fp, headers=garbage, preload_content=False)
+        assert resp.length is None
+
+    def test_length_when_chunked(self):
+        headers = {'content-length': '5',
+                   'transfer-encoding': 'chunked'}
```
Lukasa: This combination of headers is forbidden by [RFC 7230 Section 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2):

> A sender MUST NOT send a Content-Length header field in any message that contains a Transfer-Encoding header field.

test/test_response.py
```diff
@@ -377,6 +377,59 @@ def test_empty_stream(self):
...
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        resp.read()
+        assert resp.length == 0
+
+        # Test partial read
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        data = resp.stream(2)
+        next(data)
+        assert resp.length == 3
```
Lukasa: Given this behaviour, would it be more apt to call this field `length_remaining`?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1076,3 +1076,52 @@ def socket_handler(listener):
...
+        def test_head(test_obj):
+            #Test stream on 0 length body
+            conn = _make_connection(test_obj)
+            head_response = conn.request('HEAD', url='/', preload_content=False,
+                                         strict_content_length=True)
+            data = [chunk for chunk in head_response.stream(1)]
+            assert len(data) == 0
+
+        test_get(self)
+        test_head(self)
```
Lukasa: For the sake of clarity, I think I'd advise splitting these into two separate tests if possible.

urllib3/response.py
Lukasa: I don't think so. However, please put `strict_content_length` *after* `retries`.

urllib3/exceptions.py
```diff
@@ -193,6 +193,14 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
+class IncompleteRead(HTTPError):
```
Lukasa: This should probably inherit from `httplib`'s `IncompleteRead` error class, which we already sometimes raise.

urllib3/response.py
```diff
@@ -182,6 +192,38 @@ def tell(self):
...
+            status = int(self.status)
+        except:
+            status = 0
+
+        # Check for responses that shouldn't include a body
+        if (status in (204, 304) or 100 <= status < 200
+           or request_method == 'HEAD'):
+            length = 0
+
+        return length
```
Lukasa: For now this is fine. In the future, I wonder if we shouldn't be more aggressive about refusing to serve responses with bogus content lengths.

urllib3/response.py
```diff
@@ -182,6 +192,38 @@ def tell(self):
...
+    def _init_length(self, request_method):
+        """
+        Set initial length value for Response content.
+
+        Logic extracted from httplib.py in the standard Python library.
+        """
+        length = self.headers.get('content-length')
+        if length is not None and not self.chunked:
+            try:
+                length = int(length)
```
Lukasa: This is not quite good enough, as it leads us to subtly disregard [RFC 7230 section 3.3.2](https://tools.ietf.org/html/rfc7230#section-3.3.2):

> If a message is received that has multiple Content-Length header fields with field-values consisting of the same decimal value, or a single Content-Length header field with a field value containing a list of identical decimal values (e.g., "Content-Length: 42, 42"), indicating that duplicate Content-Length header fields have been generated or combined by an upstream message processor, then the recipient MUST either reject the message as invalid or replace the duplicated field-values with a single valid Content-Length field containing that decimal value prior to determining the message body length or forwarding the message. 

We should aim to handle this case *either* by rejecting all invalid content lengths, *or* by handling this case specially. Doing neither is pretty bad though. =(

Lukasa:
Cool, this is a really good patch so far! I've added some notes here for strictness and other things which I think are fairly important: let me know if you have thoughts!

test/test_response.py
nateprewitt: I added this test because the initial `httplib` length logic checks to make sure things aren't chunked. I view this test in the same vein as receiving a Content-Length of "foo". It shouldn't happen but do we want to actually return the content length in the event urllib3 happens to receive both? I'd say no, the length should be None because it can't be determined.

Alternatively, we could raise an exception here but I'm not sure if that would buy us anything useful other than aborting the operation.

test/test_response.py
nateprewitt: I'm fine with renaming it, I was just trying to keep this in sync with `httplib.HTTPResponse.length` so you can call `length` on either object and get back the same result.

test/with_dummyserver/test_socketlevel.py
nateprewitt: I'd definitely like to do that, I wasn't sure if there was a way to reuse the socket_handler without having to rewrite it for each test? I can if needed, I just didn't want to add redundant bloat if it wasn't needed.

urllib3/exceptions.py
nateprewitt: Ok, great.

urllib3/response.py
nateprewitt: By rejecting, do you mean raising an exception in that event instead of setting `length` to None?

If we want to handle the case (which I think would be better than the exception), something like:

```python
single_length = length.split(',')[0].strip()
length = int(single_length)
```
should work, right?

urllib3/response.py
nateprewitt: I'm ok with raising exceptions in the event of a header, but I think it should be done uniformly for the issues you brought up in the other comments. Returning `None` would only apply for Transfer-Encoding: chunked at that point. I guess as long as we leave the `strict_content_length` flag in place, this wouldn't be breaking.

nateprewitt: We should also definitely change the name, probably to your proposed `length_remaining` so people don't confuse it with `length` in `httplib`.

test/test_response.py
Lukasa: Yeah, this boils down to "how do we want to deal with this"? The options are as follows:

1. Raise an exception explaining what went wrong. This fails fast and clearly.
2. Fall back to no content-length. That means we'll treat the body as chunked. If it's not, we'll fail fast (IncompleteRead, usually). If it is, everything works.

I think that's probably ok in this case, but I'd like a comment explaining the rationale.

Lukasa: Possibly also a log at warning level to explain what we're doing.

test/test_response.py
Lukasa: Hrrrrm. I don't think there's any requirement for compatibility in this case. =)

test/with_dummyserver/test_socketlevel.py
Lukasa: Sure, you can hoist the function up to e.g. module scope.

urllib3/response.py
Lukasa: Yes, rejecting would be raising an exception. =)

As to that approach, no, not quite. That *also* violates RFC 7230 which only wants us to accept the header if the entire list is identical. That is, if:

```python
lengths = {val.strip() for val in header_val.split(',')}
if len(lengths) > 1:
    raise StupidHeaderBlockError("What are you thinking?")
length = int(lengths[0])
```

urllib3/response.py
Lukasa: I think in the first instance we should be tolerant of these mistakes, and add some function. Let's build this up out of small incremental steps. =)

urllib3/response.py
```diff
@@ -200,6 +202,17 @@ def _init_length(self, request_method):
         Logic extracted from httplib.py in the standard Python library.
         """
         length = self.headers.get('content-length')
+
+        if length is not None and self.chunked:
+            # This Response will fail with an IncompleteRead if it can't be
+            # received as chunked. This works as a fallback to at least attempt
```
nateprewitt: Could probably clarify "This works as a fallback to at least attempt to read" as "This method falls back to attempt reading..."

test/test_response.py
```diff
@@ -377,6 +379,74 @@ def test_empty_stream(self):
...
+        # handle response ignoring content length.
+        headers = {'content-length': '5',
+                   'transfer-encoding': 'chunked'}
+        fp = BytesIO(b'12345')
+
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        assert resp.length_remaining is None
+
+    def test_length_with_multiple_content_lengths(self):
+        headers = {'content-length': '5, 5.0'}
```
Lukasa: We shouldn't allow this. Floating point numbers in the content-length field aren't a good idea: we really only want base-10 integers. =)

urllib3/response.py
```diff
@@ -182,6 +195,53 @@ def tell(self):
...
+            # to read the response before raising an exception.
+            log.warning("Received response with both Content-Length and "
+                        "Transfer-Encoding set. This is expressly forbidden "
+                        "by RFC 7230 sec 3.3.2. Ignoring Content-Length and "
+                        "attempting to process response as Transfer-Encoding: "
+                        "chunked.")
```
Lukasa: Let's add an early return here? That lets us simplify up some later logic.

urllib3/response.py
```diff
@@ -182,6 +195,53 @@ def tell(self):
...
+
+        if length is not None and not self.chunked:
+            try:
+                lengths = set([int(float(val)) for val in length.split(',')])
+                if len(lengths) > 1:
+                    raise InvalidHeader("Content-Length contained multiple, "
+                                        "unmatching values (%s)" % length)
+                length = lengths.pop()
```
Lukasa: Any reason to use pop instead of `[0]`?

Lukasa:
Cool, getting much closer!

urllib3/response.py
nateprewitt: `set` doesn't support indexing but we know it's a single value at this point. We could do a dictionary too, I just figured a `set` made more sense since we're not utilizing the value portion of `dict`.

test/test_response.py
nateprewitt: That was my gut reaction at first, but after rereading RFC 7230, I wasn't sure since they chose to use the phrase "decimal value".
>If a message is received that has multiple Content-Length header fields with field-values consisting of the same decimal value, or a single Content-Length header field with a field value containing a list of identical decimal values (e.g., "Content-Length: 42, 42")

 I agree that something like 42.6 makes absolutely no sense, but I could see a server not converting a float to an int for some reason and returning 42.0.

I can strip out the `float` conversion for `lengths` though, and it should just fall back to `None`.

urllib3/response.py
nateprewitt: Rad, done and logic updated to remove unnecessary length declarations.

urllib3/response.py
Lukasa: Ah, yeah, good spot. Forgot sets don't let you index.

nateprewitt:
Alright, @Lukasa, things should be cleaned up with changes from last wave of comments.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1076,3 +1076,69 @@ def socket_handler(listener):
...
+            buf = b''
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+
+            sock.send(
+                b'HTTP/1.1 200 OK\r\n'
+                b'Content-Length: 22\r\n'
+                b'Content-type: text/plain\r\n'
+                b'\r\n'
+                b'hello, world'
```
Lukasa: Why is this actually sending a body?

nateprewitt: Because I copy pasted without thinking it through :) I'll strip out 1130.

urllib3/exceptions.py
```diff
@@ -193,6 +196,25 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
+class IncompleteRead(httplib_IncompleteRead):
```
Lukasa: Hrmmmmmm.

Sorry @nateprewitt, I know I should have dealt with this earlier, but I think that this should probably dual-inherit from both IncompleteRead and `HTTPError`, so that people indiscriminately catching all urllib3 errors catch this too.

urllib3/response.py
```diff
@@ -182,9 +195,53 @@ def tell(self):
...
+            log.warning("Received response with both Content-Length and "
+                        "Transfer-Encoding set. This is expressly forbidden "
+                        "by RFC 7230 sec 3.3.2. Ignoring Content-Length and "
+                        "attempting to process response as Transfer-Encoding: "
+                        "chunked.")
+            return None
+
+        if length is not None:
+            try:
+                lengths = set([int(val) for val in length.split(',')])
```
Lukasa: While we're tidying things up, can we add a comment here to explain what we're defending against and how this code manages to do it?

urllib3/response.py
```diff
@@ -182,9 +195,53 @@ def tell(self):
...
+                length = None
+            else:
+                if length < 0:
+                    length = None
+
+        # Convert status to int for comparison
+        # In some cases, httplib returns a status of "_UNKNOWN"
+        try:
+            status = int(self.status)
+        except:
```
Lukasa: Let's not use a blank `except` here: that's almost never what you actually want. Let's do `except ValueError` instead.

urllib3/response.py
```diff
@@ -327,9 +384,14 @@ def read(self, amt=None, decode_content=None, cache_content=False):
                     # no harm in redundantly calling close.
                     self._fp.close()
                     flush_decoder = True
+                    if (self.strict_content_length and
+                       self.length_remaining not in (0, None)):
+                        raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
```
Lukasa: So here's a note: this only fires for reads with lengths. If you call `read()`, this error will never fire. Should we enforce this for those reads too?

nateprewitt: This already fires for read() because it's [raised in httplib](https://github.com/python/cpython/blob/2.7/Lib/httplib.py#L597). This patch is covering the edge case of streamed data which httplib never implemented an error for due to wanting to avoid breaking backwards compatibility as noted [here](https://github.com/python/cpython/blob/2.7/Lib/httplib.py#L614).

Lukasa: Cool, in that case we should write this down in a comment too for when I inevitably forget that. ;)

nateprewitt:
Alright here's another handful of commits. Sorry the commit list is getting so long, it just seems easier to manage everything rather than rebasing each time. I'll squash things down into more comprehensive chunks when we're ready to merge.

Lukasa:
Ok, cool, I'm happy with this now I think. @haikuginger and @sigmavirus24, mind doing a review of this as well?

test/test_response.py
```diff
@@ -377,6 +379,73 @@ def test_empty_stream(self):
 
         self.assertRaises(StopIteration, next, stream)
 
+    def test_length_no_header(self):
+        fp = BytesIO(b'12345')
+        resp = HTTPResponse(fp, preload_content=False)
+        assert resp.length_remaining is None
```
sigmavirus24: We don't use pytest in urllib3. Please use the appropriate `assert*` methods, e.g.,

```py
self.assertIsNone(resp.length_remaining)
```

test/test_response.py
```diff
@@ -377,6 +379,73 @@ def test_empty_stream(self):
...
+
+    def test_length_w_valid_header(self):
+        headers = {"content-length": "5"}
+        fp = BytesIO(b'12345')
+
+        resp = HTTPResponse(fp, headers=headers, preload_content=False)
+        assert resp.length_remaining == 5
```
sigmavirus24: assertEqual

urllib3/response.py
```diff
@@ -327,9 +389,19 @@ def read(self, amt=None, decode_content=None, cache_content=False):
                     # no harm in redundantly calling close.
                     self._fp.close()
                     flush_decoder = True
+                    if (self.strict_content_length and
+                       self.length_remaining not in (0, None)):
```
sigmavirus24: This doesn't look to follow the style of the rest of the project and should be indented more. I suspect you're also following PEP8 with a line length of 79. That is not what this project follows.

sigmavirus24:
So I'm looking at our response class and we only initialize a decoder when the content-encoding specifies `gzip` or `deflate`. So we shouldn't always have a decoder to lean on. Why then is all of the logic in the DeflateDecoder class? Finally, a bit of a yak to shave: I'd prefer this flag be "enforce_content_length" rather than "strict_content_length". It's more descriptive.

nateprewitt:
@sigmavirus24 could you clarify this statement really quickly?
>So I'm looking at our response class and we only initialize a decoder when the content-encoding specifies gzip or deflate. So we shouldn't always have a decoder to lean on. Why then is all of the logic in the DeflateDecoder class?

I'm not sure what logic you're seeing added on the DeflateDecoder class. All of the major logic changes should either be in `_init_length` or `read` which are both part of HTTPResponse.

sigmavirus24:
Sorry for the confusion Nat. I didn't expand the diff enough and it looked as if you were adding to the DeflateDecoder class. Feel free to ignore that comment.

nateprewitt:
Thanks for the feedback Ian! Things should be updated.

urllib3/response.py
sigmavirus24: +1 on a comment

sigmavirus24:
@nateprewitt it looks like someone updated urllib3 to require your branch to be consistently rebased on top of master (easily one of GitHub's most annoying mistakes/misfeatures as it ties into other things people generally want). Can you rebase this as well please?

nateprewitt:
Ok cool, things are rebased onto current master. All of @sigmavirus24's proposed changes should be in place. Once @haikuginger gives the thumbs up, I'll squash things down to a more manageable commit list and update CHANGES.

urllib3/response.py
```diff
@@ -182,9 +195,58 @@ def tell(self):
...
+            # received as chunked. This method falls back to attempt reading
+            # the response before raising an exception.
+            log.warning("Received response with both Content-Length and "
+                        "Transfer-Encoding set. This is expressly forbidden "
+                        "by RFC 7230 sec 3.3.2. Ignoring Content-Length and "
+                        "attempting to process response as Transfer-Encoding: "
+                        "chunked.")
+            return None
+
+        if length is not None:
```
haikuginger: This conditional is entirely supplemental to the one above. Possibly do `elif length is not None`?

urllib3/response.py
```diff
@@ -327,9 +389,19 @@ def read(self, amt=None, decode_content=None, cache_content=False):
                     # no harm in redundantly calling close.
                     self._fp.close()
                     flush_decoder = True
+                    if (self.enforce_content_length and
+                            self.length_remaining not in (0, None)):
```
haikuginger: This is still kinda counter to urllib3's style. I would break the second statement out into a method (`self.has_remaining_data`, perhaps?) so we can fit everything on one line easily without supplemental parentheses. Also, as @sigmavirus24 mentioned, we go with a 100-character line here, rather than the 79-character line recommended by PEP-8, so I'm not sure if you even need to do any refactoring to fit it into one line.

haikuginger:
Couple minor nits left.

urllib3/response.py
nateprewitt: Cool, it's a length of 95 expanded, so I went with that approach. Adding the check as another method seems a bit excessive if we're only going to use it here.

nateprewitt:
Alright, @Lukasa, @sigmavirus24, @haikuginger, I think everything has been addressed and I squashed the commits down into two separate feature commits. One for the `length_remaining` attribute and then building on that to implement `enforce_content_length`.

Lukasa:
Cool, I still like this. @haikuginger @sigmavirus24, are you two happy?

sigmavirus24:
I've skimmed through (haven't downloaded it and played with it) but it :eyes:-only looks good.

haikuginger:
I'm on the same page as @sigmavirus24; haven't played with it, but it looks good.

Lukasa:
Ok, cool, I'm happy with this then. Thanks for the great work @nateprewitt!

Lukasa:
And thanks so much for the reviews @sigmavirus24 and @haikuginger, fantastic team job all around.

haikuginger:
Thanks for the excellent work, @nateprewitt! Way to drive this PR through. ✨🍻✨

urllib3/response.py
nateprewitt: @Lukasa are you seeing the more aggressive content-length validation as part of a urllib3 2.0/#someday feature, or something that would be integrated before a major (1.20+). It seems breaking, especially if people are using this for testing purposes. I wanted to record a few of the changes that were discussed here but passed on for later PRs.

nateprewitt: Nevermind, disregard this. The gain probably doesn't outweigh the loss. This would be breaking, it doesn't really provide significant benefit to the user, and will cripple some of the "do anything" functionality choices Kenneth wanted to preserve in Requests.

PR #953, opened 2016-08-18
Skip invalid headers

### LINKED ISSUE #950: Malformed header causes IndexError during header parsing
This response (note the space and quotes around `Content-type`):

```
$ curl --head https://consumersentinel.gov
HTTP/1.0 200 OK
 "Content-type": text/html
pragma: no-cache
retry-after: 600
Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
Server: BigIP
Connection: Keep-Alive
Content-Length: 2835
```

Causes this error in `urllib3`:

```
  File "/home/eric/.pyenv/versions/2.7.11/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 590, in urlopen
    **response_kw)
  File "/home/eric/.pyenv/versions/2.7.11/lib/python2.7/site-packages/requests/packages/urllib3/response.py", line 373, in from_httplib
    headers = HTTPHeaderDict.from_httplib(headers)
  File "/home/eric/.pyenv/versions/2.7.11/lib/python2.7/site-packages/requests/packages/urllib3/_collections.py", line 317, in from_httplib
    key, value = headers[-1]
IndexError: list index out of range
```

I'm using `urllib3` via `requests`, but it doesn't look like that's causing the issue.
##

haikuginger:
In cases where a header starts with whitespace, and that header does not immediately follow another header, drop that header. This imitates the treatment Python3 gives to these headers natively. Fixes #950.

### DIFF
--- a/test/test_collections.py
+++ b/test/test_collections.py
@@ -339,5 +339,34 @@ www-authenticate: bla
         self.assertEqual(d['www-authenticate'], 'asdf, bla')
         self.assertEqual(d.getlist('www-authenticate'), ['asdf', 'bla'])
 
+    def test_from_httplib_py2_invalid_first_header(self):
+        if six.PY3:
+            raise SkipTest("python3 has a different internal header implementation")
+        msg = """
+ Fake Header: I shouldn't be here
+Server: nginx
+Content-Type: text/html; charset=windows-1251
+Connection: keep-alive
+X-Some-Multiline: asdf
+ asdf
+ asdf
+Set-Cookie: bb_lastvisit=1348253375; expires=Sat, 21-Sep-2013 18:49:35 GMT; path=/
+Set-Cookie: bb_lastactivity=0; expires=Sat, 21-Sep-2013 18:49:35 GMT; path=/
+www-authenticate: asdf
+www-authenticate: bla
+
+"""
+        buffer = six.moves.StringIO(msg.lstrip().replace('\n', '\r\n'))
+        msg = six.moves.http_client.HTTPMessage(buffer)
+        d = HTTPHeaderDict.from_httplib(msg)
+        self.assertEqual(d['server'], 'nginx')
+        cookies = d.getlist('set-cookie')
+        self.assertEqual(len(cookies), 2)
+        self.assertTrue(cookies[0].startswith("bb_lastvisit"))
+        self.assertTrue(cookies[1].startswith("bb_lastactivity"))
+        self.assertEqual(d['x-some-multiline'].split(), ['asdf', 'asdf', 'asdf'])
+        self.assertEqual(d['www-authenticate'], 'asdf, bla')
+        self.assertEqual(d.getlist('www-authenticate'), ['asdf', 'bla'])
+
 if __name__ == '__main__':
     unittest.main()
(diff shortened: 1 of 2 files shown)


test/test_collections.py
```diff
@@ -339,5 +339,34 @@ def test_from_httplib_py2(self):
         self.assertEqual(d['www-authenticate'], 'asdf, bla')
         self.assertEqual(d.getlist('www-authenticate'), ['asdf', 'bla'])
 
+    def test_from_httplib_py2_invalid_first_header(self):
+        if six.PY3:
+            raise SkipTest("python3 has a different internal header implementation")
```
Lukasa: If we're implementing something that replicates Python 3's behaviour, shouldn't this test behave identically on Python 2 and Python 3?

test/test_collections.py
```diff
@@ -339,5 +339,34 @@ def test_from_httplib_py2(self):
...
+        msg = six.moves.http_client.HTTPMessage(buffer)
+        d = HTTPHeaderDict.from_httplib(msg)
+        self.assertEqual(d['server'], 'nginx')
+        cookies = d.getlist('set-cookie')
+        self.assertEqual(len(cookies), 2)
+        self.assertTrue(cookies[0].startswith("bb_lastvisit"))
+        self.assertTrue(cookies[1].startswith("bb_lastactivity"))
+        self.assertEqual(d['x-some-multiline'].split(), ['asdf', 'asdf', 'asdf'])
+        self.assertEqual(d['www-authenticate'], 'asdf, bla')
+        self.assertEqual(d.getlist('www-authenticate'), ['asdf', 'bla'])
```
Lukasa: It would be nice to turn this entire test into a socketlevel test that *actually* receives this response. Or maybe a new test that does that. Either way, it'd be good to get httplib involved in this processing.

urllib3/_collections.py
```diff
@@ -314,8 +314,11 @@ def from_httplib(cls, message):  # Python 2
 
         for line in message.headers:
             if line.startswith((' ', '\t')):
-                key, value = headers[-1]
-                headers[-1] = (key, value + '\r\n' + line.rstrip())
+                if headers:
+                # If the first header is an invalid one, then just drop it.
+                # This mimic Python 3.
```
Lukasa: Python 3 *specifically* drops the first header? Really?

haikuginger: That's what I can tell from testing when the first header has leading whitespace - it just disappears.

test/test_collections.py
haikuginger: We don't actually use this method on Python 3 at all - as far as I can tell, we're using it to mimic the Python 3 header data structure in Python 2.

Lukasa: Sure, but that's why I suggested rewriting the test to use the entire httplib stack: that way we can validate that we really do behave identically in both cases.

Are you comfortable writing that test @haikuginger?

haikuginger: Ahhh, gotcha. I'll muddle through somehow - I usually do.

Lukasa: Ok, shout if you have trouble. =)

konklone:
Looks like this got dropped...

Issue #952, opened 2016-08-18
urllib3 does not urlencode redirect targets on redirect.


eladbitton:
I am crawling pages and I'm getting `UnicodeEncodeError`. 

The problematic url is : http://www.efind.co.il/detailed/52133.html
I've isolated my code to this:
`url = "http://www.efind.co.il/detailed/52133.html"
manager = PoolManager()
response = manager.request('GET', url)`

This url is redirecting to this url: articles.efind.co.il/info/דגי-נוי-בבריכה-מאמר
which is not 'ascii'.

Because this is auto redirect i can't do anything about it. I am quoting the first url and giving it to the urllib, I don't have control over the redirected urls.

Anything i can do?
##

nateprewitt:
If you call request with `redirect=False`, it should halt the request before performing the next request to articles.efind.co.il/info/דגי-נוי-בבריכה-מאמר. You can probably handle the encoding of the `Location` header before calling request on the new URL.

eladbitton:
This is optional, but i think urllib should handle the redirects properly by itself.

nateprewitt:
I don't want to give incorrect information, so I'd wait for @Lukasa, @sigmavirus24 or @shazow to confirm this. 

To my understanding, unicode in URI/IRIs is supposed to be converted to UTF-8 and then % encoded before being sent across the wire. So the server emitting raw unicode in the Location header is a violation of [RFC3987](https://www.ietf.org/rfc/rfc3987.txt) and a compliance problem with the server, not urllib3.

sigmavirus24:
@nateprewitt this has nothing to do with 3987 and everything to do with the encoding that Python expects a header to use. I suspect (although @eladbitton didn't say this explicitly) that @eladbitton is using Python 3 which tries to coerce everything to `latin-1`.

@eladbitton could you share your *full* traceback and version information (of Python and urllib3).

That said, RFC 3987 *does describe* how to convert that particular URL, but most servers will send lots of junk in headers and expect clients to handle it.

eladbitton:
```
Traceback (most recent call last):
  File "/home/elad/Installed Applications/eclipse/plugins/org.python.pydev_5.1.2.201606231256/pysrc/pydevd.py", line 1530, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "/home/elad/Installed Applications/eclipse/plugins/org.python.pydev_5.1.2.201606231256/pysrc/pydevd.py", line 937, in run
[...]

PR #954, opened 2016-08-19
adding in rfc3986 to encode unicode URIs

nateprewitt:
This addresses #952 and adds urlencoding described in RFC3986 to `Url`s when they're created.

* I'd like to test that the URL used for `r` at the end of `test_unicode_location_url` is in fact '/取得する' but can't find a way to access the URL for the `Request`. Is there a way for me to check that from the `HTTPresponse` object?

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,3 +8,4 @@ wheel==0.24.0
 tornado==4.2.1
 PySocks==1.5.6
 pkginfo>=1.0,!=1.3.0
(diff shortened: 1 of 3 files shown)


sigmavirus24:
So this is only a fraction of #952 and doesn't really address the entire problem only a tiny bit. I'm working on the rest right now.

nateprewitt:
Okay, I'll close this out then if you're already working on it.

PR #955, opened 2016-08-20
retry: Respect Retry-After header

npcode:
Sleep for the seconds specifed by Retry-After response header before
retrying the request.

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -265,8 +265,9 @@ class TestingApp(RequestHandler):
 
     def retry_after(self, request):
         if datetime.now() - self.application.last_req < timedelta(seconds=1):
+            status = request.params.get("status", "429 Too Many Requests")
             return Response(
-                    status="429 Too Many Requests",
+                    status=status.decode('utf-8'),
                     headers=[('Retry-After', '1')])
 
         self.application.last_req = datetime.now()
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -11,6 +11,8 @@ import zlib
 
 from io import BytesIO
 from tornado.web import RequestHandler
+from datetime import datetime
+from datetime import timedelta
 
 from urllib3.packages.six.moves.http_client import responses
 from urllib3.packages.six.moves.urllib.parse import urlsplit
@@ -259,6 +261,16 @@ class TestingApp(RequestHandler):
         status = request.params.get("status", "200 OK")
 
         return Response(status=status)
+
+    def retry_after(self, request):
+        if datetime.now() - self.application.last_req < timedelta(seconds=1):
+            return Response(
+                    status="429 Too Many Requests",
+                    headers=[('Retry-After', '1')])
+
+        self.application.last_req = datetime.now()
+
+        return Response(status="200 OK")
 
     def shutdown(self, request):
         sys.exit()
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -11,6 +11,7 @@ import zlib
 
 from io import BytesIO
 from tornado.web import RequestHandler
+from tornado import httputil
 from datetime import datetime
 from datetime import timedelta
 
@@ -271,6 +272,18 @@ class TestingApp(RequestHandler):
         self.application.last_req = datetime.now()
 
         return Response(status="200 OK")
+
+    def redirect_after(self, request):
+        "Perform a redirect to ``target``"
+        date = request.params.get('date')
+        if date:
+            retry_after = str(httputil.format_timestamp(
+                    datetime.fromtimestamp(float(date))))
+        else:
+            retry_after = '1'
+        target = request.params.get('target', '/')
+        headers = [('Location', target), ('Retry-After', retry_after)]
+        return Response(status='303 See Other', headers=headers)
 
     def shutdown(self, request):
         sys.exit()
--- a/test/contrib/test_gae_manager.py
+++ b/test/contrib/test_gae_manager.py
@@ -18,7 +18,7 @@ from urllib3.util.url import Url
 from urllib3.util.retry import Retry, RequestHistory
 
 from test.with_dummyserver.test_connectionpool import (
-    TestConnectionPool, TestRetry)
+    TestConnectionPool, TestRetry, TestRetryAfter)
 
 
 # Prevent nose from running these test.
@@ -195,6 +195,17 @@ class TestGAERetry(TestRetry):
     # We don't need these tests because URLFetch resolves its own redirects.
     test_retry_redirect_history = None
     test_multi_redirect_history = None
+
+class TestGAERetryAfter(TestRetryAfter):
+    __test__ = True
+
+    # Magic class variable that tells NoseGAE to enable the URLFetch stub.
+    nosegae_urlfetch = True
+
+    def setUp(self):
+        # Disable urlfetch which doesn't respect Retry-After header.
+        self.manager = AppEngineManager(urlfetch_retries=False)
+        self.pool = MockPool(self.host, self.port, self.manager)
 
 
 if __name__ == '__main__':
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -811,13 +811,39 @@ class TestRetryAfter(HTTPDummyServerTestCase):
 
     def test_retry_after(self):
         # Request twice in a second to get a 429 response.
-        r = self.pool.request('GET', '/retry_after')
-        r = self.pool.request('GET', '/retry_after')
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': '429 Too Many Requests'},
+                retries=False)
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': '429 Too Many Requests'},
+                retries=False)
         self.assertEqual(r.status, 429)
 
         r = self.pool.request('GET', '/retry_after',
-                retries=Retry(status_forcelist=[429]))
+                fields={'status': '429 Too Many Requests'},
+                retries=True)
         self.assertEqual(r.status, 200)
+
+        # Request twice in a second to get a 503 response.
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': '503 Service Unavailable'},
+                retries=False)
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': '503 Service Unavailable'},
+                retries=False)
+        self.assertEqual(r.status, 503)
+
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': '503 Service Unavailable'},
+                retries=True)
+        self.assertEqual(r.status, 200)
+
+        # Ignore Retry-After header on status which is not defined in
+        # Retry.RETRY_AFTER_STATUS_CODES.
+        r = self.pool.request('GET', '/retry_after',
+                fields={'status': "418 I'm a teapot"},
+                retries=True)
+        self.assertEqual(r.status, 418)
 
     def test_redirect_after(self):
         r = self.pool.request('GET', '/redirect_after', retries=False)
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -836,5 +836,13 @@ class TestRetryAfter(HTTPDummyServerTestCase):
         delta = time.time() - t
         self.assertTrue(delta >= 1)
 
+        # Retry-After is past
+        t = time.time()
+        timestamp = t - 1
+        r = self.pool.request('GET', '/redirect_after?date=' + str(timestamp))
+        delta = time.time() - t
+        self.assertEqual(r.status, 200)
+        self.assertTrue(delta < 1)
+
 if __name__ == '__main__':
     unittest.main()
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -801,6 +801,15 @@ class TestRetry(HTTPDummyServerTestCase):
             (302, '/multi_redirect?redirect_codes=200')
         ])
 
+    def test_retry_after(self):
+        # Request twice in a second to get a 429 response.
+        r = self.pool.request('GET', '/retry_after')
+        r = self.pool.request('GET', '/retry_after')
+        self.assertEqual(r.status, 429)
+
+        r = self.pool.request('GET', '/retry_after',
+                retries=Retry(status_forcelist=[429]))
+        self.assertEqual(r.status, 200)
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -6,6 +6,9 @@ import unittest
 import time
 import warnings
 
+from datetime import datetime
+from datetime import timedelta
+
 import mock
 
 from .. import (
@@ -801,6 +804,11 @@ class TestRetry(HTTPDummyServerTestCase):
             (302, '/multi_redirect?redirect_codes=200')
         ])
 
+
+class TestRetryAfter(HTTPDummyServerTestCase):
+    def setUp(self):
+        self.pool = HTTPConnectionPool(self.host, self.port)
+
     def test_retry_after(self):
         # Request twice in a second to get a 429 response.
         r = self.pool.request('GET', '/retry_after')
@@ -810,6 +818,23 @@ class TestRetry(HTTPDummyServerTestCase):
         r = self.pool.request('GET', '/retry_after',
                 retries=Retry(status_forcelist=[429]))
         self.assertEqual(r.status, 200)
+
+    def test_redirect_after(self):
+        r = self.pool.request('GET', '/redirect_after', retries=False)
+        self.assertEqual(r.status, 303)
+
+        t = time.time()
+        r = self.pool.request('GET', '/redirect_after')
+        self.assertEqual(r.status, 200)
+        delta = time.time() - t
+        self.assertTrue(delta >= 1)
+
+        t = time.time()
+        timestamp = t + 2
+        r = self.pool.request('GET', '/redirect_after?date=' + str(timestamp))
+        self.assertEqual(r.status, 200)
+        delta = time.time() - t
+        self.assertTrue(delta >= 1)
 
 if __name__ == '__main__':
     unittest.main()
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -696,7 +696,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 release_conn=release_conn, **response_kw)
 
         # Check if we should retry the HTTP response.
-        if retries.is_forced_retry(method, status_code=response.status):
+        has_retry_after = bool(response.getheader('Retry-After'))
+        if retries.is_retry(method, response.status, has_retry_after):
             try:
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
@@ -707,7 +708,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                     raise
                 return response
             retries.sleep(response)
-            log.debug("Forced retry: %s", url)
+            log.debug("Retry: %s", url)
             return self.urlopen(
                 method, url, body, headers,
                 retries=retries, redirect=redirect,
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -686,6 +686,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                     raise
                 return response
 
+            retries.sleep_for_retry(response)
             log.debug("Redirecting %s -> %s", url, redirect_location)
             return self.urlopen(
                 method, redirect_location, body, headers,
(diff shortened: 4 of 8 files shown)


test/with_dummyserver/test_connectionpool.py
```diff
@@ -805,6 +805,14 @@ def test_multi_redirect_history(self):
             (302, '/multi_redirect?redirect_codes=200')
         ])
 
+    def test_retry_after(self):
+        r = self.pool.request('GET', '/retry_after')
+        r = self.pool.request('GET', '/retry_after')
+        self.assertEqual(r.status, 429)
```
Lukasa: Why are there two requests here?

dummyserver/handlers.py
```diff
@@ -264,6 +266,17 @@ def status(self, request):
...
         return Response(status=status)
 
+    def retry_after(self, request):
+        if hasattr(TestingApp, 'last_req') and \
+                datetime.now() - TestingApp.last_req < timedelta(seconds=1):
+            return Response(
+                    status="429 Too Many Requests",
+                    headers=[('Retry-After', '1')])
+
+        TestingApp.last_req = datetime.now()
```
Lukasa: I would vastly prefer not to patch this on to the class object: it is really fairly gross. Instead, it should be possible to set it on the `Application` object created in `server.py`: we can initialize it to zero in that code and then use `self.application` to query it, rather than patch onto the class.

Are you open to doing that?

Lukasa:
Thanks for this @npcode!

I certainly have no objection to supporting this behaviour: I'm happy to add support for it. I've left some notes inline, however, because I think the implementation needs a bit of work. Do you mind addressing those please?

test/with_dummyserver/test_connectionpool.py
npcode: To get a 429 response. The server responses 429 if a client requests twice in a second.

dummyserver/handlers.py
npcode: Thanks for your advice. I've updated it as you said.

test/with_dummyserver/test_connectionpool.py
Lukasa: Ah, I see. Can you note that in a comment please?

npcode: Done!

urllib3/contrib/appengine.py
```diff
@@ -184,6 +184,7 @@ def urlopen(self, method, url, body=None, headers=None,
             retries = retries.increment(
                 method, url, response=http_response, _pool=self)
             log.info("Forced retry: %s", url)
+            retries.sleep_for_retry_after(http_response)
             retries.sleep()
```
Lukasa: I think this logic will misbehave: you can sleep once for the retry after and then again for the exponential backoff. I don't think we want to do that: if we are respecting retry-after we shouldn't use our exponential backoff. The old logic here was better. =)

dummyserver/handlers.py
```diff
@@ -260,6 +262,17 @@ def status(self, request):
 
         return Response(status=status)
 
+    def retry_after(self, request):
+        if hasattr(self.application, 'last_req') and \
```
Lukasa: Rather than use this `hasattr`, let's clean it up and ensure that the application always has a `last_req` field by setting it ourselves in `server.py`.

npcode:
I have updated it again. Could you please take a look?

Lukasa:
Ok, I'm happy with this. @sigmavirus24 @haikuginger, thoughts?

haikuginger:
I think this makes sense. I think that we might not necessarily want to use the `status_forcelist` variable to determine if we ought to retry, though; that could lead to situations where we immediately retry because the status code is "right", but we don't have a value for the `Retry-After` header. We're also not handling the [redirect aspect of that header, or the possibility for a date value to be placed in that field.](https://tools.ietf.org/html/rfc7231#section-7.1.3)

Long story short, if we're implementing this feature, I really want to have it implemented according to spec. Otherwise, it makes more sense for downstream urllib3 consumers to implement a version for their own libraries that meets their particular needs.

Lukasa:
> We're also not handling the redirect aspect of that header, or the possibility for a date value to be placed in that field.

So the possibility of a date is a real concern, and we should probably aim to support that. I'm less worried about dealing with the redirect part at the exact same time: we can consider that as a separate enhancement.

As to the status forcelist, do you have a suggestion for what we should do @haikuginger?

haikuginger:
I think we should have a separate variable, named something like `retry_after_status_codes`, listing the valid options that, by RFC, can produce a `Retry-After` header. To my knowledge, that's 503, 429, and 413, as well as any redirect in the 3xx range, although @sigmavirus24 probably knows better than I.

Of course, the upstream user will be able to override it, but by default, we should stick to the RFC codes. And we should have explicit kwarg(s) that can either disable the feature altogether or set a maximum delay time (possibly with some sort of safe default)- we don't want to fall prey to a `Retry-After: 31536000` header with no recourse besides program termination.

Lukasa:
I'm happy enough with that idea: @sigmavirus24?

sigmavirus24:
I don't see anything objectionable. I have some questions, but they're not urgent or blocking. (My main concern is why `sleep` takes a `response` object. I suppose we're planning for some eventuality where it might need to look at other headers? What headers do we suppose it might need to introspect?

npcode:
@sigmavirus24 I passed `response` object only for `Retry-After` header. I guess it has a benefit to hide retrying logic, even the name of `Retry-After` header, from the caller. Currently I have no idea we will need to introspect another headers or status code in future.

urllib3/util/retry.py
```diff
@@ -194,12 +194,20 @@ def get_backoff_time(self):
         backoff_value = self.backoff_factor * (2 ** (consecutive_errors_len - 1))
         return min(self.BACKOFF_MAX, backoff_value)
 
-    def sleep(self):
+    def sleep(self, response=None):
         """ Sleep between retry attempts using an exponential backoff.
```
sigmavirus24: This docstring needs updating (now that I look at it). As it is currently written it implies it only uses an exponential back off. It should be updated to use a more vague title, e.g.,

```py
Sleep between retry attempts.

This method will respect a server's ``Retry-After`` response header and sleep the duration of the time requested. If that is not present, it will use an exponential backoff. By default, the backoff factor is 0 and this method will return immediately.
"""

Comment (npcode) [inline urllib3/util/retry.py]: @sigmavirus24 Thanks, I have updated the comment as you said.

**Comment by sigmavirus24** [comment]
👍 looks good to me

**Comment by Lukasa** [comment]
@npcode Are you interested in implementing Retry-After date handling and status codes as suggested by @haikuginger? If not, that's fine, we'll find someone to implement the other half.

**Comment by npcode** [comment]
@Lukasa Sure, I'll implement them.

**Comment by npcode** [comment]
gae test fails because `google.appengine.api.urlfetch` doesn't respsect `Retry-After` header when `follow_redirects` is enabled. Should we disable `follow_redirects` and implement redirection following for google appengine?

**Comment by Lukasa** [comment]
@jonparrott Is there any reason not to do that?

**Comment by jonparrott** [comment]
I would err against a sweeping change on how we call the underlying urlfetch service. It's a black box and sometimes the box doesn't do what we want. One immediate issue I can think of is that allowing urlfetch to deal with redirects reduces your quota (e.g. if something redirects you 10 times, having urlfetch handle it means you only use 1 request quota, but handling it within urllib3 means you used 10 request quota).

Instead, how about we add have an option to `AppEngineManager` on whether we should use urlfetch's redirect behavior or urllib3's?

**Comment by Lukasa** [comment]
@jonparrott I'm open to that. Are you willing to assist @npcode with prototyping that?

**Comment by jonparrott** [comment]
Sure. 

The general idea is that if you do:

```python
http = AppEngineManager(urlfetch_retries=True)
```

Then it'll work as it does today. However, if you do:

```python
http = AppEngineManager(urlfetch_retries=False)
```

Then it'll behave as the other managers. 

@npcode let me know if you need any help getting the implementation here to work. :)

npcode:
@jonparrott Thanks for your advice! I'll try it.

test/contrib/test_gae_manager.py
```diff
@@ -164,7 +164,7 @@ class TestGAERetry(TestRetry):
     nosegae_urlfetch = True
 
     def setUp(self):
-        self.manager = AppEngineManager()
+        self.manager = AppEngineManager(urlfetch_retries=False)
```
npcode: Do we need to test the case for `urlfetch_retires=True`?

jonparrott: I would say yes.

npcode:
I've updated this for `AppEngineManager`. Could you please review it again?

urllib3/connectionpool.py
```diff
@@ -695,7 +696,7 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
                 release_conn=release_conn, **response_kw)
 
         # Check if we should retry the HTTP response.
-        if retries.is_forced_retry(method, status_code=response.status):
+        if retries.is_retry(method, response.status, bool(response.getheader('Retry-After'))):
```
Lukasa: I don't think there's any need to squish all of this into one line, we can split the `bool` line out into a temporary variable.

urllib3/connectionpool.py
```diff
@@ -705,8 +706,8 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
                     response.release_conn()
                     raise
                 return response
-            retries.sleep()
-            log.debug("Forced retry: %s", url)
+            retries.sleep(response)
+            log.debug("Retry: %s", url)
```
Lukasa: Why are we removing the word `forced` from everything?

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
+                    redirect_url = redirect_location
+                return self.urlopen(
+                    method, redirect_url, body, headers,
+                    retries=retries, redirect=redirect,
+                    timeout=timeout, **response_kw)
 
         # Check if we should retry the HTTP response.
-        if retries.is_forced_retry(method, status_code=http_response.status):
+        if retries.is_retry(method, http_response.status,
+                            bool(http_response.getheader('Retry-After'))):
```
Lukasa: Same here, we don't need to squash this up.

urllib3/util/retry.py
```diff
@@ -194,12 +206,51 @@ def get_backoff_time(self):
...
 
-    def sleep(self):
-        """ Sleep between retry attempts using an exponential backoff.
+    def parse_retry_after(self, retry_after):
+        if retry_after.isdigit():
+            seconds = int(retry_after)
+        else:
+            retry_date_tuple = email.utils.parsedate(retry_after)
+            if retry_date_tuple is None:
+                raise ValueError("Invalid Retry-After header: %s" % retry_after)
```
Lukasa: Can we raise a urllib3 exception here? `ValueError` is going to be unexpected for a lot of users.

urllib3/util/retry.py
```diff
@@ -194,12 +206,51 @@ def get_backoff_time(self):
...
+        return seconds
+
+    def get_retry_after(self, response):
+        """ Get the value of Retry-After in seconds. """
+
+        retry_after = response.getheader("Retry-After")
 
-        By default, the backoff factor is 0 and this method will return
-        immediately.
+        if not retry_after:
```
haikuginger: I'd rather do an explicit `if retry_after is None` here. I also think we want an integer return value of `0` for cases where there isn't a valid `Retry-After` header; that'll let us skip a lot of branching in other logic, and just `sleep(x)` where `x == 0`.

urllib3/util/retry.py
```diff
@@ -194,12 +206,51 @@ def get_backoff_time(self):
...
+            return
+
+        return self.parse_retry_after(retry_after)
+
+    def sleep_for_retry(self, response=None):
```
haikuginger: Why do we need both `sleep_for_retry` and `sleep`?

urllib3/util/retry.py
```diff
@@ -194,12 +206,51 @@ def get_backoff_time(self):
...
+        will use an exponential backoff. By default, the backoff factor is 0 and
+        this method will return immediately.
         """
+
+        if response:
+            retry_after = self.get_retry_after(response)
+            if retry_after:
+                time.sleep(retry_after)
+                return
+
```
haikuginger: If we've got a `Retry-After` header, should we also be factoring in backoff calculations? The way I see it, the server already told us when it wants us to contact it again.

urllib3/util/retry.py
```diff
@@ -123,18 +124,28 @@ class Retry(object):
...
+    DEFAULT_RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
+
     #: Maximum backoff time.
     BACKOFF_MAX = 120
 
     def __init__(self, total=10, connect=None, read=None, redirect=None,
                  method_whitelist=DEFAULT_METHOD_WHITELIST, status_forcelist=None,
                  backoff_factor=0, raise_on_redirect=True, raise_on_status=True,
-                 history=None):
+                 history=None, status_retry_after_list=DEFAULT_RETRY_AFTER_STATUS_CODES):
```
haikuginger: I'd like an option to disable the Retry-After header respecting completely and in a way that makes sense - doing `status_retry_after_list = []` should work, but it's not ideal. Something like `respect_retry_after_headers=True` as a default would be ideal.

urllib3/connectionpool.py
npcode: Because this line runs if `retry.is_retry()` returns true. `retry.is_retry()` may return true even if the stauts code is not in `status_forcelist`.

urllib3/util/retry.py
npcode: @haikuginger Yes, that's why I used `ValueError`. It would be better if we have an exception for invalid HTTP header. I'll find better one or create new one.

nateprewitt: @npcode, I recently added an [InvalidHeader](https://github.com/shazow/urllib3/blob/master/urllib3/exceptions.py#L215-L217) exception which should work for this case.

npcode: @nateprewitt Thanks a lot! That's exactly what I want.

urllib3/util/retry.py
npcode: `sleep` tries to sleep with an exponential backoff if `Retry-After` doesn't exist while `sleep_for_retry` doesn't. We need `sleep_for_retry` for redirection which doesn't need to sleep with the backoff.

urllib3/util/retry.py
npcode: I think `get_retry_after` should return `None` instead of `0` to let we know the header doesn't exist.

urllib3/util/retry.py
npcode: Agreed. This implementation doesn't factor in backoff calculation in the case as you see.

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
+                r = urlparse(redirect_location)
+                if not r.netloc:
+                    u = urlparse(url)
+                    redirect_url = Url(scheme=u.scheme, host=u.netloc, path=r.path).url
+                else:
+                    redirect_url = redirect_location
+                return self.urlopen(
+                    method, redirect_url, body, headers,
+                    retries=retries, redirect=redirect,
+                    timeout=timeout, **response_kw)
```
Lukasa: This is a substantial new amount of redirect code: how do you feel here @jonparrott?

urllib3/contrib/appengine.py
```diff
@@ -131,16 +135,17 @@ def urlopen(self, method, url, body=None, headers=None,
...
                 url,
                 payload=body,
                 method=method,
                 headers=headers or {},
                 allow_truncated=False,
-                follow_redirects=(
-                    redirect and
-                    retries.redirect != 0 and
-                    retries.total),
+                follow_redirects=(self.urlfetch_retries and follow_redirects),
```
jonparrott: nit: parenthesis are unnecessary here.

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
+                raise MaxRetryError(self, url, "too many redirects")
+            else:
+                if http_response.status == 303:
+                    method = 'GET'
+
+                try:
+                    retries = retries.increment(method, url, response=http_response, _pool=self)
+                except MaxRetryError:
+                    if retries.raise_on_redirect:
+                        # Release the connection for this response, since we're not
```
jonparrott: There is no connection for `AppEngineManager`, so is this necessary?

urllib3/contrib/appengine.py
jonparrott: I would love to see an explicit test for `urlfetch_retries`.

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
+                        # returning it to be released manually.
+                        http_response.release_conn()
+                        raise
+                    return http_response
+
+                retries.sleep_for_retry(http_response)
+                log.info("Redirecting %s -> %s", url, redirect_location)
```
jonparrott: this should be `debug` to match `ConnectionPool`.

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
+                r = urlparse(redirect_location)
```
jonparrott: Lines 204-209 doesn't seem to be present in the other managers, can you add a comment as to why this is needed?

jonparrott:
It seems in general we've got a lot of repeated code around redirects and retrying. I can't help but wonder if that could be largely consolidated, but perhaps an exercise for another PR.

dummyserver/handlers.py
```diff
@@ -260,6 +263,29 @@ def status(self, request):
 
         return Response(status=status)
 
+    def retry_after(self, request):
+        if datetime.now() - self.application.last_req < timedelta(seconds=1):
+            status = request.params.get("status", "200 OK")
```
npcode: This should be "429 Too Many Requests". Thanks to let me know!

urllib3/contrib/appengine.py
npcode: Well, the lines are not necessary because they could be simply replaced with `urlparse.urljoin`.

urllib3/contrib/appengine.py
npcode: 5c729531137c836f7ec6ec325a423b40aae8742c changed the log level from `info` to `debug` for `ConnectionPool` but not for `AppEngineManager`. Thas was just a mistake?

urllib3/contrib/appengine.py
npcode: @jonparrott `TestGAERetry` tests the case of `urlfetch_retries=True`. Could you please give me an example or explanation of the tests you want to see if you think that is not enough?

urllib3/contrib/appengine.py
```diff
@@ -174,17 +179,46 @@ def urlopen(self, method, url, body=None, headers=None,
...
 
-        # Check for redirect response
-        if (http_response.get_redirect_location() and
-                retries.raise_on_redirect and redirect):
-            raise MaxRetryError(self, url, "too many redirects")
+        # Handle redirect?
+        redirect_location = redirect and http_response.get_redirect_location()
+        if redirect_location:
+            # Check for redirect response
+            if (self.urlfetch_retries and retries.raise_on_redirect):
```
npcode: I believe `raise_on_redirect` works. The `else` block here includes some codes, which is copied from `HTTPConnectionPool`, to raise the error.

urllib3/contrib/appengine.py
Lukasa: Yeah, I think that was just a mistake.

Lukasa:
Ok so @jonparrott mentioned a small number of changes, let's see if we can get those made and then I'll do another pass around.

urllib3/contrib/appengine.py
npcode: Thanks, I've removed this.

urllib3/contrib/appengine.py
npcode: Thanks, I've changed the log level.

Lukasa:
Ok @npcode, we've just shipped a new release with our bugfixes, so I'm happy to go ahead and merge this. Can you please update the branch to the new master?

npcode:
@Lukasa Okay, I did.

Lukasa:
Hrm, travis seems to be building the wrong commit.

Lukasa:
Seems a bit like we totally outfoxed Travis here, I have no idea why it's building the wrong commit.

sigmavirus24:
@Lukasa have you pulled the branch? GitHub doesn't display commits in the position they hold in the graph. They usually list them by date of creation so if @npcode reordered commits then that could be why it's building "the wrong commit"

Lukasa:
Aha, yup, the branch does seem right when I pull it. Goddamn it GitHub, get it together.

Lukasa:
Ok, I pronounce myself happy. Thanks so much for your work here @npcode, it's really fantastic stuff. It took a really long time to get done, and I appreciate your patience with this.

sigmavirus24:
@npcode You're awesome! 👍

PR #956, opened 2016-08-21
WIP: Handle Redirects in the PoolManager and Normalize URLs during redirects

sigmavirus24:
This could probably be two PRs, but they're deeply related in my mind so I'm starting by keeping the pieces of work together. I'll happily pull 2176c26 into its on PR if you want.

That said, I'm starting this WIP now because the previous implementation of Url was *not* rfc3986 compliant and I'm not sure if we want to keep that behaviour and if people are actually relying on it. Specifically doing

```py
from urllib3.util.url import parse_url

url = parse_url('google.com:80/foo')
```

Would give you an instance of `Url` where `url.hostname == 'google.com'`. [Section 3 of RFC3986](https://tools.ietf.org/html/rfc3986#section-3) however points out that you can only determine the authority (in which the hostname is contained) if it starts with `//`, e.g., a valid URI would be `//google.com:80` not `google.com:80`. So when `rfc3986.urlparse` sees this input, it (correctly per the RFC) parses that differently than `parse_url` used to.

Presently, we even have tests that assert this behaviour. Is this something anyone can rely on? As I understand it, you can't pass a URL without a scheme and hostname. I'm wondering what we needed this for.

I hate removing tests, but I think these should be removed. Thoughts @Lukasa @haikuginger ?


Lukasa:
For the moment I'm ACKing this, but I don't have time to review it right now: I'll need to take a look at it tomorrow. Just wanted to note that I hadn't missed it.

urllib3/poolmanager.py
```diff
@@ -254,10 +259,6 @@ def urlopen(self, method, url, redirect=True, **kw):
         if response.status == 303:
             method = 'GET'
 
-        retries = kw.get('retries')
-        if not isinstance(retries, Retry):
-            retries = Retry.from_int(retries, redirect=redirect)
-
```
Lukasa: What's the reasoning for moving this?

urllib3/util/url.py
```diff
@@ -1,37 +1,80 @@
...
         if path and not path.startswith('/'):
             path = '/' + path
-        if scheme:
-            scheme = scheme.lower()
-        if host:
-            host = host.lower()
-        return super(Url, cls).__new__(cls, scheme, auth, host, port, path,
-                                       query, fragment)
+            if parse_result is not None:
+                parse_result = parse_result.copy_with(path=path)
```
Lukasa: Do we want to move this logic into a custom constructor, rather than overload the default constructor?

haikuginger:
I have no additional substantive comments. One thought; would it be worth adding a method to do round-trip URL cleaning/validation?

haikuginger:
@sigmavirus24, the SessionManager branch moves the redirect logic to RequestMethods; would you be interested in taking a look at that branch and seeing if the logic you've added here can fit into it?

sigmavirus24:
@haikuginger yes. I'm just working on getting a new release of rfc3986 out the door first to handle a few things that I noticed urllib3 needs for these tests to pass

PR #958, opened 2016-08-24
Remove 3DES support.

### LINKED ISSUE #957: Remove 3DES from cipher suite.
Well, [3DES is broken now](https://sweet32.info/). This attack is difficult to mount against requests/urllib3, but not impossible, and so we should aim to remove 3DES support in the near future. We don't need to rush out a release for it though: the next release is fine.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37239034-remove-3des-from-cipher-suite?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
Resolves #957.

While I'm here, is there any reason we're still supporting HIGH? That will let 3DES in through the back door in older OpenSSL versions. I'm inclined to want to turn that off too.


Lukasa:
Ok, I just used the authority vested in me as maintainer to make the unilateral decision to remove HIGH. Contribs, you have a little bit of time to object, but I feel good about this.

haikuginger:
Do we want to remove the ciphers, or `!` them?

Lukasa:
Remove.

The only reason to ! them is if they would otherwise be included: for example, if we decided to leave HIGH in we'd have to !3DES. That's why we !MD5, for example: we disallow any construction of a cipher suite that involves MD5, in case some of our earlier suites might accidentally allow it.

haikuginger:
Okay; I'm 👍 then.

Lukasa:
<3 thanks @haikuginger!

sigmavirus24:
Belated 👍

PR #960, opened 2016-08-26
fix encode_multipart_formdata (TypeError: 'float' does not have the buffer interface)

kartiknagpal:
Allow request "fields" to have floating point values, otherwise an explicit cast to str is needed while making the request.

### DIFF
--- a/urllib3/filepost.py
+++ b/urllib3/filepost.py
@@ -77,7 +77,7 @@ def encode_multipart_formdata(fields, boundary=None):
         writer(body).write(field.render_headers())
         data = field.data
 
-        if isinstance(data, int):
+        if isinstance(data, int) or isinstance(data, float):
             data = str(data)  # Backwards compatibility
 
         if isinstance(data, six.text_type):


Lukasa:
This is an interesting PR. I'm not sure we want to support it, though, at least not in this form. Does it make sense to send non-integer values ever?

kartiknagpal:
Yes, it does makes sense whenever you need fractional values(eg. transaction amount, etc.). Floating point values exists for the same reason.

sigmavirus24:
I strongly feel that the representation of that floating point value should be determined by the user. The precision, rounding, etc. are things we absolutely should not be guessing.

kartiknagpal:
Sure, let the user determine. Let the user make that choice, whether to make use of say "1.5" or 1.5. The latter choice throws an exception, as of now.

Lukasa:
@sigmavirus24 is arguing that the current behaviour is preferable because it *forces* the user to make a choice.

haikuginger:
Also, it looks like the current statement only exists as a compatibility shim in the first place? If that's the case, I'd be against adding more features to something that's arguably already technical debt.

Lukasa:
I think that amounts to consensus here. Thanks @kartiknagpal, but I'm afraid we will not be accepting this PR. Sorry!

PR #964, opened 2016-09-05
PySocks 1.5.7 causes problems with IPv6.

Lukasa:
Just spent a few hours chasing this down, but basically the IPv6 support in PySocks 1.5.7 (which, to be clear, I wrote most of) is a bit busted. There's a fix in PySocks master for this, so we don't need to do any extra work, but we should add a blacklist marker for PySocks 1.5.7.


haikuginger:
Looks good to me. :+1:.

PR #934, opened 2016-07-26
Use "with" to close more files eagerly and also on error

scop:

PR #966, opened 2016-09-05
Extra with statements.

Lukasa:
Updates #934. Thanks to @scop for this work!


Lukasa:
My judgement is that we don't need a CHANGES update for this because it touches no product code.

PR #970, opened 2016-09-06
WIP HSTS-HPKP support

kislyuk:
**Do not merge yet**

This is a combination and update of #607 and #608. The two features are closely related, should share the same persistence interface and preload list functionality, and should have consistent interfaces/behaviors/naming. Here is the work that remains as per discussions in #607 and #608:

[ ] Persistence interface, possibly based on dogpile.cache - this should probably hash hostnames ssh-style, for privacy
[ ] Preload list interface and optional dependency
[ ] Move hpkp out of util as per prior comments
[ ] Close read of https://tools.ietf.org/html/rfc6797 and https://tools.ietf.org/html/rfc7469
[ ] RFC conformance tests
[ ] Thread safety / locks around writers
[ ] Write the docs


Lukasa:
@kislyuk Thanks for starting this! However, I strongly recommend doing one of these at a time (probably HSTS first): a patch that implements both at once is going to be almost impossible to effectively review.

sigmavirus24:
> a patch that implements both at once is going to be almost impossible to effectively review.

This patch already looks as if it **is** impossible to effectively review

PR #972, opened 2016-09-08
RFC: Transport Security Manager

kislyuk:
@Lukasa, can you please take a look at this and tell me what you think?

In the process of reviewing work from #607 and #608, and in light of comments in #970 about the complexity of the code, I've decided to try breaking down the work into smaller chunks.

Here is the first proposed chunk: two new stub classes, TransportSecurityManager and TransportSecurityStore, and minimal hooks for them.

If you agree with the layout/architecture of this, I can get it documented and further stubbed out, but the intent would be to first merge a stub that doesn't do any actual HSTS/HPKP, and then have a series of PRs to review/merge components that implement it.


urllib3/contrib/pyopenssl.py
```diff
@@ -413,6 +414,7 @@ def wrap_socket(self, sock, server_side=False,
 
 
 def _verify_callback(cnx, x509, err_no, err_depth, return_code):
+    cnx.certs.append(x509)
```
Lukasa: Let's leave this stuff out for now: it's logically part of HPKP.

urllib3/transport_security.py
```diff
@@ -0,0 +1,67 @@
...
+        log.debug("hsts validate: %s %s %s", scheme, conn.host, conn.port)
+
+    def validate_established_connection(self, conn):
+        """
+        Enforce HPKP or a custom certificate fingerprint.
+        """
+        if util.IS_PYOPENSSL:
+            log.debug("hpkp validate: %s", conn.connection.certs)
+        else:
+            log.debug("PyOpenSSL not available, hpkp validation disabled")
```
Lukasa: Do we need both of these methods?

urllib3/transport_security.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        try:
+            sts = _split_header(response.headers.get("strict-transport-security"))
+        except Exception as e:
+            sts = {}
+        try:
+            pins = _split_header(response.headers.get('public-key-pins'))
+        except Exception:
+            pins = {}
+        log.debug("tss enroll: %s %s", sts, pins)
```
Lukasa: Let's have this be really a stub with nothing in it.

Lukasa:
So *in general* the shape of this is good. However, I think the transport security manager should be owned at the PoolManager level, not the ConnectionPool level. Want to move that around?

urllib3/transport_security.py
kislyuk: Fundamentally, HSTS is concerned with examining the request before the connection is made (to prevent any unsecured content from going out over the wire), and HPKP is concerned with examining the completed TLS connection before request data is sent. So I think both these methods are necessary. We could make them into one method, but then it would have to be called from two places and adjust its behavior based on whether the connection is live.

Lukasa: HSTS is concerned with examining the request *and* the connection. This doesn't do that. =)

kislyuk: The connection object contains information about the scheme, host and port, doesn't it? What information from the request do we need to correctly do an HSTS check?

Lukasa: Ah, my apologies, you're quite right.

urllib3/transport_security.py
kislyuk: Done. (This was more to prove to myself that this has the information it needs to do its job)

urllib3/contrib/pyopenssl.py
kislyuk: Removed.

kislyuk:
@Lukasa, thanks for the review. I've added some docstrings and moved TSM initialization into `PoolManager`. I couldn't get rid of the import and initialization in `ConnectionPool`, though, since it can also get used directly without `PoolManager`, and `ConnectionPool` and `Connection` assume the presence of a TSM.

Please take a look - is that how you want it?

Lukasa:
@kislyuk I'll do a review tomorrow, but thanks for the work!

test/test_tsm.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+from urllib3.transport_security import parse_header, TransportSecurityManager, TransportSecurityStore
+
+
+class TSMTest(unittest.TestCase):
+
+    def test_parse_header(self):
+        self.assertEqual(parse_header("foo=1; bar=a"), dict(foo="1", bar="a"))
```
Lukasa: So I'd like to see this test fleshed out substantially with quite a few more input examples. =)

Lukasa: I'd also like some invalid inputs tested with the failures occurring appropriately.

urllib3/connectionpool.py
```diff
@@ -192,6 +194,8 @@ def __init__(self, host, port=None, strict=False,
         for _ in xrange(maxsize):
             self.pool.put(None)
 
+        self.transport_security_manager = transport_security_manager or TransportSecurityManager()
```
Lukasa: So I still think this is being initialized in the wrong place. If we're going to initialize it automatically at all, it should be at the PoolManager level or higher.

urllib3/connectionpool.py
```diff
@@ -166,6 +167,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
     def __init__(self, host, port=None, strict=False,
                  timeout=Timeout.DEFAULT_TIMEOUT, maxsize=1, block=False,
                  headers=None, retries=None,
+                 transport_security_manager=None,
```
Lukasa: I don't really love inserting new headers before others: people passing these headers positionally run into trouble.

urllib3/connection.py
```diff
@@ -244,10 +244,12 @@ class VerifiedHTTPSConnection(HTTPSConnection):
     ca_cert_dir = None
     ssl_version = None
     assert_fingerprint = None
+    transport_security_manager = None
 
     def set_cert(self, key_file=None, cert_file=None,
                  cert_reqs=None, ca_certs=None,
                  assert_hostname=None, assert_fingerprint=None,
+                 transport_security_manager=None,
```
Lukasa: Is there any reason we aren't inserting this header last?

Lukasa:
Nevermind, I reviewed it now! =D

urllib3/connectionpool.py
kislyuk: Because ConnectionPool can be instantiated without PoolManager (as the tests do, for example), removing this initialization means ConnectionPool and Connection have to be able to function without a TSM (with TSM set to None). This means adding a bunch of conditionals (`if self. transport_security_manager: ...`) instead of having a mandatory no-op TSM. Is that what you prefer?

urllib3/connectionpool.py
kislyuk: Fixed.

urllib3/connection.py
kislyuk: Fixed.

urllib3/connectionpool.py
kislyuk: I went ahead and removed the initialization, and changed the callers not to assume that `transport_security_manager` is set.

test/test_tsm.py
kislyuk: Added more tests.

kislyuk:
@Lukasa, I think I've addressed your latest round of comments, please take a look.

kislyuk:
@Lukasa, I think I've addressed your latest round of comments, please take a look.

Lukasa:
Hi @kislyuk, you weren't forgotten. ;) I was just taking the weekend off from major code reviews.

urllib3/connectionpool.py
Lukasa: Hrrrm. Ok, interesting point.

I'll review the code as-is, but I might end up giving feedback that tells you to go back. ;)

urllib3/connection.py
```diff
@@ -316,6 +318,9 @@ def connect(self):
             server_hostname=hostname,
             ssl_context=context)
 
+        if self.transport_security_manager:
```
Lukasa: `if self.transport_security_manager is not None:` is better here. =)

urllib3/poolmanager.py
```diff
@@ -112,6 +113,8 @@ class PoolManager(RequestMethods):
 
     def __init__(self, num_pools=10, headers=None, **connection_pool_kw):
         RequestMethods.__init__(self, headers)
+        if "transport_security_manager" not in connection_pool_kw:
+            connection_pool_kw["transport_security_manager"] = TransportSecurityManager()
```
Lukasa: So I think `transport_security_manager` is something that connection pools have to be keyed off: different transport security managers should cause different connection pools to be created.

Additionally, we should probably store the transport security manager we're using *on* the PoolManager itself: given that it's a global object, the correct way for users to get access to it will be from the pool manager.

Lukasa:
Cool, I think we're making good progress here. At this point I'd like to page @haikuginger in particular because this may intersect slightly with his `SessionManager` layer.

urllib3/connection.py
kislyuk: Fixed here and elsewhere.

urllib3/poolmanager.py
kislyuk: I've exposed `transport_security_manager` as a `PoolManager` attribute.

I'm not entirely sure what needs to be done to key connection pools off TSMs. It's already the case that a connection pool can only be associated with one TSM; a pool manager is also expected to only be associated with one TSM. Could you explain what I need to do?

Lukasa: To key a pool off `transport_security_manager`, it needs to be added to the `HTTPPoolKey` object. That should be sufficient, though you'll want to add a test like those in `test_poolmanager.py` to confirm that the keying is working appropriately.

What this means, essentially, is that if the user *changes* the TSM on the pool manager, all existing pools will no longer be usable for new connections.

kislyuk: Sounds good. OK, I added it to HTTPPoolKey and added a unit test. Please take a look.

Lukasa:
Ok, cool, I have no specific notes aside from wanting @haikuginger to take a look.

I do have an open-ended question though, @kislyuk. With HPKP, its purpose is soley to reject a connection based on bad pins. That's fine. With HSTS, however, do we want it to reject attempts to do HTTP connections for HSTS'd hosts, or would we rather automatically swap HSTS'd hosts to HTTPS transparently? If it's the second, I'm not sure that we're doing that here.

kislyuk:
@Lukasa, that's a great question. As I was implementing this, I thought the initially easier thing to do will be to reject such attempts, but down the line we will want to enable the automatic swapping. 

The RFC states that if an HSTS match is found, the UA must change the scheme to HTTPS, but says nothing more. https://tools.ietf.org/html/rfc6797#section-8.3

When I first looked at this a few days ago I was unsure how to implement the swapping. As I'm getting more comfortable with urllib3 internals, I now have an idea of how it might be done: HTTPConnectionPool.__init__ can raise an HSTSError if there is a match, and PoolManager.connection_from_host can catch that error, transform the request context, and try again. How does that sound?

Lukasa:
@kislyuk That's a good first idea, but doesn't cover the situation where we discover HSTS settings *after* we've initialized a ConnectionPool. I think that would mean that we want two behaviours:

1. ConnectionPools should throw errors if HTTP is being used to connect to a HSTS host. This would be done at request time, not `__init__`-time.
2. PoolManager should check before it selects a connection pool if the host is HSTS-enabled, and if it is it should automatically set the scheme to HTTPS.

That way, users that use the connectionpool directly get defended (they can't issue requests without disabling the TSM), while users that use our higher-level interfaces get the automatic switchover behaviour.

How does that sound?

kislyuk:
OK, I see the need to check against HSTS at connection time, not at init time. But I think the pool manager should still catch the HSTSError bubbling up from ConnectionPool at connection time, and then switch the pool. Doing a separate check in PoolManager seems like it would result in a race condition.

Lukasa:
It can't be at connection time, it has to be at request time. There is a race condition there, but it's non-permanent, so I think it's fine.

kislyuk:
OK, I get it now. So there's no avoiding putting the HSTS check in two different places.

I'll refactor the TSM hooks to reflect this design, and then I'd like to push for a merge.

kislyuk:
@Lukasa, I've simplified the TSM to only have one validation method for now, `validate_connection()`. The location in ConnectionPool where this check sits is per-request, not per-connection, so I think we're actually all set to follow your plan. The validate_connection method can have the following semantics:

* If the connection scheme is HTTP, do an HSTS assert
* If the connection scheme is HTTPS and the connection is established, do an HPKP check

There is a corresponding piece of logic that will need to go into PoolManager, to catch any HSTS assert errors and re-initialize the pool. I'd prefer to merge this initial architecture before I add that logic, since it will require adding more HSTS-specific bits into the stubs.

I don't think PoolManager needs to directly call TSM, because like I mentioned the check in ConnectionPool already happens on a per-request basis (and on a fresh ConnectionPool, happens before opening the socket). So I think it's simpler to handle this situation via exceptions.

urllib3/poolmanager.py
```diff
@@ -110,8 +116,14 @@ class PoolManager(RequestMethods):
 
     proxy = None
 
-    def __init__(self, num_pools=10, headers=None, **connection_pool_kw):
+    def __init__(self, num_pools=10, headers=None, transport_security_manager=None,
+                 **connection_pool_kw):
         RequestMethods.__init__(self, headers)
+
+        if transport_security_manager is None:
+            transport_security_manager = TransportSecurityManager()
```
haikuginger: How do we turn this off? _Can_ we turn this off?

kislyuk: The way to turn this off would be to pass a no-op subclass of TransportSecurityManager. Right now TransportSecurityManager is a no-op; in the next iteration I will include something like NullTransportSecurityManager, so:

```
from urllib import PoolManager
from urllib.transport_security import NullTransportSecurityManager

pool = PoolManager(transport_security_manager=NullTransportSecurityManager())
```

nateprewitt: Having to pass a no-op class seems like a somewhat painful/unpythonic solution, I feel like specifically setting a variable to `None` would suggest we wanted the feature disabled. The other uses of transport_security_manager do a check to verify it's not `None`. Could we possibly implement "on by default" with a different default or flag?

kislyuk: I can change this to transport_security_manager=True kwarg default, with None signifying to install no TSM.

kislyuk:
I think it would be great to have an event/callback pub-sub model, maybe something like `ConnectionPool.before_connect(callback)`, `ConnectionPool.after_connect(callback)`, `ConnectionPool.before_request(callback)`, `ConnectionPool.after_response(callback)` with a way to put those on PoolManager too. But that's beyond the scope of this PR and would be good for a future refactor.

urllib3/poolmanager.py
nateprewitt: I think that could work as long as the conditional is `if transport_security_manager is True` so we're not accidentally catching "truthy" objects. I'd wait for @Lukasa or @haikuginger to weigh in before making this change though.

urllib3/connection.py
```diff
@@ -316,6 +318,9 @@ def connect(self):
             server_hostname=hostname,
             ssl_context=context)
 
+        if self.transport_security_manager is not None:
+            self.transport_security_manager.validate_connection(self.sock)
```
Lukasa: I'm a little bit nervous about using the same method at two very different places in connection setup and management. Can we consider having two methods, with different names?

urllib3/poolmanager.py
Lukasa: So I'd like to suggest an alternative: a sentinel object. Generally speaking, if you want to have `None` do something sensible but have a "default" behaviour you'd create a module-scoped variable called something like `_DEFAULT_SENTINEL = object()`, and then set the default value of the kwarg to that. Then you can do identity tests to confirm whether or not the default value was used. I'd be happy to see that used here.

urllib3/connection.py
kislyuk: Renamed this one to `validate_ssl_connection`.

urllib3/poolmanager.py
kislyuk: @Lukasa, I've addressed your latest comments, please take a look.

urllib3/poolmanager.py
```diff
@@ -116,11 +119,11 @@ class PoolManager(RequestMethods):
 
     proxy = None
 
-    def __init__(self, num_pools=10, headers=None, transport_security_manager=None,
+    def __init__(self, num_pools=10, headers=None, transport_security_manager=_DEFAULT_TSM,
                  **connection_pool_kw):
         RequestMethods.__init__(self, headers)
 
-        if transport_security_manager is None:
+        if transport_security_manager == _DEFAULT_TSM:
```
Lukasa: This should be an `is` check, not an equality check.

urllib3/transport_security.py
```diff
@@ -28,11 +28,26 @@ def __init__(self, transport_security_store=None):
...
+        since new HSTS headers may be processed in the meantime.
+
         This is a stub, to be implemented later.
 
         :param conn:
             A :class:`urllib3.connection.HTTPConnection` instance.
+
+        """
+
+    def validate_ssl_connection(self, conn):
```
Lukasa: I don't love this name along with `validate_connection`: they're both fundamentally about validating TLS. I appreciate that naming here is really hard, but I'm still not sure about this.

Would it be better to have `validate_new_connection` and `validate_established_connection`? Or possibly just `validate_hpkp` and `validate_hsts`? At least those names are clear and descriptive: even if they're overly specific, we can always rename them later when we have a better feel for how these will fit together.

kislyuk: `validate_new_connection` would be misleading, because per our earlier discussion, we actually call it every time we make a request (as we should, in case new state is added from processing response headers).

I like `validate_hpkp` and `validate_hsts`. They're clear and concrete, but they might discourage implementers from extending the class. One other thing that logically fits into TSM is manual fingerprint pinning (`assert_fingerprint`), concurrent with HPKP. But anyway, I'll go ahead and rename to these for now.

kislyuk:
@Lukasa - Updated names, fixed identity check, please take another look.

urllib3/poolmanager.py
```diff
@@ -110,8 +119,14 @@ class PoolManager(RequestMethods):
 
     proxy = None
 
-    def __init__(self, num_pools=10, headers=None, **connection_pool_kw):
+    def __init__(self, num_pools=10, headers=None, transport_security_manager=_DEFAULT_TSM,
+                 **connection_pool_kw):
         RequestMethods.__init__(self, headers)
+
+        if transport_security_manager is _DEFAULT_TSM:
+            transport_security_manager = TransportSecurityManager()
```
Lukasa: I just noticed that this should probably have an `else` block that sets `transport_security_manager = None`: otherwise we'll pass the empty object down the stack.

urllib3/transport_security.py
```diff
@@ -0,0 +1,79 @@
+import logging
+from six.moves.http_cookiejar import split_header_words
+
+__all__ = ['TransportSecurityManager', 'TransportSecurityStore']
+
+
+log = logging.getLogger(__name__)
```
Lukasa: Let's remove this logger for now: I expect it to come back in in a future pull request, but that's ok.

urllib3/poolmanager.py
nateprewitt: Wouldn't that prevent the passing a custom `TransportSecurityManager()`? Either it's the default or it's `None`?

nateprewitt: Wouldn't that prevent passing a custom `TransportSecurityManager`? An `else` would imply if it's not the default, then it must be `None`, no?

kislyuk: Sorry, I'm not sure if I follow. If I call `PoolManager()`, `_DEFAULT_TSM` gets overwritten with an actual TSM. If I call `PoolManager(transport_security_manager=None)`, then `None` gets passed down the stack, disabling TSM. Any other value will also get passed down the stack and result in an error if it's not a TSM. Am I missing anything?

urllib3/transport_security.py
kislyuk: Fixed.

urllib3/poolmanager.py
Lukasa: Oh god, this is clearly why I shouldn't do code reviews from crowded conferences! Yeah, we don't need to do that, please ignore me. =P

urllib3/transport_security.py
```diff
@@ -0,0 +1,76 @@
+import logging
```
Lukasa: Can we remove this import as well?

kislyuk: Oops. Removed.

kislyuk:
Says here this branch is up-to-date with master. I last merged it in on Tuesday.

urllib3/poolmanager.py
```diff
@@ -33,12 +34,14 @@
...
+                                          'block', 'source_address',
+                                          'transport_security_manager')
 )
 HTTPSPoolKey = collections.namedtuple(
     'HTTPSPoolKey', HTTPPoolKey._fields + SSL_KEYWORDS
 )
 
+_DEFAULT_TSM = object()
 
 def _default_key_normalizer(key_class, request_context):
```
Lukasa: Looks like the builder says you need to add an extra blank line here to make PEP8 happy.

Lukasa:
Ok @kislyuk, I think this looks really good! What I'm going to do is to push a new branch up that contains this change on top of master. That will let us keep this work out of the main tree until it's providing some function, at which point we can merge it in one go.

Lukasa:
Ok, you can now see this in the branch `transport-security-manager`. Thanks for this start @kislyuk, you can now go ahead and start implementing HSTS! :custard: :sparkles: Well done!

PR #973, opened 2016-09-10
fixing incorrect message for IncompleteRead

nateprewitt:
I missed this during one of the final reviews when we changed urllib3's `IncompleteRead` from using `HTTPError` to httplib's `IncompleteRead` constructor. This is currently producing an incorrect length based on the `message` length rather than the ints we're passing in.

I updated the test to catch this and overwrote httplib's `__repr__` to create the correct message.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 dev (master)
 ------------
 
+* Fixed incorrect message for IncompleteRead exception. (PR #973)
+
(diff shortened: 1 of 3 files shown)


Lukasa:
Thanks @nateprewitt! Want to add a CHANGES.rst for this please?

nateprewitt:
Good to go.

Lukasa:
Thanks @nateprewitt! :custard:

Issue #977, opened 2016-09-11
HTTPResponse.closed consistency


nateprewitt:
I spent a while yesterday trying to figure out why responses were hitting logic requiring the file to be closed while still evaluating to `closed=False`. There's currently an inconsistency in how urllib3 handles closed responses and part of that is due to #929.

First off, I guess we need to clarify what "closed" means for the HTTPResponse. Does it mean the content has all been read, or that we've closed the file pointer? Perhaps it's both and we should be marking the response closed when all the content is read?

Background
--------------
Right now, we have two separate variables `isclosed` (Python 2&3) and `closed` (Python 3) on Python's `HTTPResponse` object. They both behave differently in what determines their boolean value. `isclosed` is defined by the `HTTPResponse`, whether that be in `httplib` (Python 2) or `http.client` (Python 3), while `closed` is defined by the `io` module that `http.client.HTTPResponse` inherits from in Python 3.

`isclosed` checks if the file pointer's content is `None` while the `io` module's `closed` attr only evaluates to True when set. This means we can read all of the bytes off the HTTPResponse object but still have an "open" file pointer with a length of 0 when using `io`'s `closed` attr. This won't change until `close()` is called on the file pointer,  but since Python 2 doesn't have a `closed` attr it drops through the first check in urllib3's `closed` method an evaluates as currently expected. Perhaps it's the Python 2 logic that's flawed though, and we shouldn't be leaving cleanup to the GC (assuming the response is ever dereferenced).

I put together 69018f7 to address the issue but didn't open the PR immediately because of the GC question and a side effect with urllib3's `close` method. Moving `isclosed` to a higher preference than `closed` causes `close` to no longer actually close the file pointer as it previously did in Python3. We can obviously implement something to address this but I wanted to get another pair of eyes on this before moving forward.

Example
-----------
This is a quick demo for Python 3 with the current master branch, as is the test included in the PR:
```python
from urllib3.connectionpool import HTTPConnectionPool
from urllib3.util.response import is_fp_closed

conn = HTTPConnectionPool('stackoverflow.com', maxsize=10)
get_response = conn.request('GET', url='/', preload_content=False)
data = [chunk for chunk in get_response.stream(1024)]
print(is_fp_closed(get_response._fp), get_response.closed)
```

Lukasa:
[...]

PR #978, opened 2016-09-12
HTTPResponse.closed consistency

nateprewitt:
Right now `urllib3.response.HTTPResponse.closed` returns different values based on Python version. This addresses the issue by preferring the more specific `isclosed` to `io`'s `closed` attr. See #977 for more details.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ dev (master)
 * Accept ``iPAddress`` subject alternative name fields in TLS certificates.
   (Issue #258)
 
+* Fixed consistency of ``HTTPResponse.closed`` between Python 2 and 3.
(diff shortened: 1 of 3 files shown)


nateprewitt:
Just to make a note of this, so we're not accidentally regressing #187. It looks like this helper was added to support `io` compatibility. Right now, `closed` doesn't function like a standard `io` module object in Python 2, or when calling `read()` with some `amt`. Given the inconsistency, and the fact that `http.client.HTTPResponse` also has caveats where it doesn't perform like a "proper" `io` object (read after close, etc.), I believe this is a better approach.

Lukasa:
Fab, this looks excellent. Great work @nateprewitt! :sparkles: :custard:

PR #980, opened 2016-09-18
Move to a more complex bit of idna handling.

### LINKED ISSUE #979: Usage of cryptography + idna breaks on wildcard certs
The changes in #930 seem to have broken with some wildcard certificates: https://stackoverflow.com/questions/39521147/why-is-urllib3-idna-complaining-about-a-wildcard-in-an-x509-cert-how-do-i-fix-i

I haven't attempted to reproduce this, but I thought recording it here would be of use.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37940366-usage-of-cryptography-idna-breaks-on-wildcard-certs?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Lukasa:
This resolves #979, according to @reaperhulk. I should aim to write a test for this!

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -138,7 +138,19 @@ def _dnsname_to_stdlib(name):
     then on Python 3 we also need to convert to unicode via UTF-8 (the stdlib
     uses PyUnicode_FromStringAndSize on it, which decodes via UTF-8).
     """
-    name = idna.encode(name)
+    def idna_encode(name):
+        """
+        Borrowed wholesale from the Python Cryptography Project. It turns out
+        that we can't just safely call `idna.encode`: it can explode for
+        wildcard names. This avoids that problem.
+        """
+        for prefix in [u'*.', u'.']:
+            if name.startswith(prefix):
+                name = name[len(prefix):]
+                return prefix.encode('ascii') + idna.encode(name)
+        return idna.encode(name)
+
+    name = idna_encode(name)
     if sys.version_info >= (3, 0):
         name = name.decode('utf-8')
     return name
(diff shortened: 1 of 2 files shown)


Lukasa:
I'd love some code review here! @haikuginger @sigmavirus24 @nateprewitt @reaperhulk?

urllib3/contrib/pyopenssl.py
```diff
@@ -138,7 +138,19 @@ def _dnsname_to_stdlib(name):
     then on Python 3 we also need to convert to unicode via UTF-8 (the stdlib
     uses PyUnicode_FromStringAndSize on it, which decodes via UTF-8).
     """
-    name = idna.encode(name)
+    def idna_encode(name):
```
sigmavirus24: Is there a specific reason for this being a closure?

Lukasa: Ah, no specific reason other than it can be. Would you like it hoisted?

reaperhulk:
LGTM

urllib3/contrib/pyopenssl.py
sigmavirus24: I'm not about to block this PR on this, but it does worry me that our pyopenssl integration has a very low bus factor. Tucking this in here doesn't seem entirely conducive to long-term maintenance but it's only used here so it's not presently very harmful either.

Lukasa: Let's leave it for now: I don't think it's useful anywhere else at this time, and we make very few API guarantees in the PyOpenSSL stuff so I don't think this will explode in any big way. We can always safely hoist it if we want to.

Lukasa: Can...can I impose on you to click the 'approve' button? I want all the green ticks. =P

PR #989, opened 2016-09-29
Disallow superscripts on digit checks

scop:
`isdigit()` lets other things besides 0-9 to pass, for example ².

An alternative to the fixes here would be to use e.g. `re.match(r"^[0-9]+$", ...)`, but that's slower.

With this change to `parse_retry_after`, a slight behavioral change is introduced: it starts to accept values surrounded by whitespace (not sure if they can end up here in usual use cases), and prefixed by `+` or `-`. Previously negative values raised `InvalidHeader`, now they get truncated to 0. Previously + prefixed values raised `InvalidHeader`, now they are accepted and the + ignored. If you feel these are issues that should be kept as they were, let me know and I'll adjust that change.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -10,6 +10,7 @@ from mock import patch, Mock
 
 from urllib3 import add_stderr_logger, disable_warnings
 from urllib3.util.request import make_headers
+from urllib3.util.retry import Retry
 from urllib3.util.timeout import Timeout
 from urllib3.util.url import (
     get_host,
@@ -28,6 +29,7 @@ from urllib3.exceptions import (
     InsecureRequestWarning,
     SSLError,
     SNIMissingWarning,
+    InvalidHeader,
 )
 from urllib3.util.connection import (
     allowed_gai_family,
@@ -527,3 +529,19 @@ class TestUtil(unittest.TestCase):
     def test_ip_family_ipv6_disabled(self):
         with patch('urllib3.util.connection.HAS_IPV6', False):
             self.assertEqual(allowed_gai_family(), socket.AF_INET)
+
+    def test_parse_retry_after(self):
+        invalid = [
+            "-1",
+            "+1",
+            "1.0",
+            six.u("\xb2"),  # \xb2 = ^2
+        ]
+        retry = Retry()
+
+        for value in invalid:
+            self.assertRaises(InvalidHeader, retry.parse_retry_after, value)
+
+        self.assertEqual(retry.parse_retry_after("0"), 0)
+        self.assertEqual(retry.parse_retry_after("1000"), 1000)
+        self.assertEqual(retry.parse_retry_after("\t42 "), 42)
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -34,6 +34,7 @@ from urllib3.util.connection import (
     _has_ipv6
 )
 from urllib3.util import is_fp_closed, ssl_
+from urllib3.packages import six
 
 from . import clear_warnings
 
@@ -95,6 +96,8 @@ class TestUtil(unittest.TestCase):
             'http://google.com:foo',
             'http://::1/',
             'http://::1:80/',
+            'http://google.com:-80',
+            six.u('http://google.com:\xb2\xb2'),  # \xb2 = ^2
         ]
 
         for location in invalid_host:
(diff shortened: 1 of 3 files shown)


scop:
Regarding re.match being slower, that's quite probably a non-issue in these code spots, so if you prefer that approach instead, let me know and I'll adjust.

scop:
(Travis failure is not related to this change.)

urllib3/util/url.py
```diff
@@ -192,7 +192,10 @@ def parse_url(url):
             # If given, ports must be integers.
             if not port.isdigit():
                 raise LocationParseError(url)
-            port = int(port)
+            try:
```
nateprewitt: If we're gonna put this in a try/except block, it probably is a bit redundant to keep the conditional above (lines 193-194), no?

nateprewitt: If we're gonna put this in a try/except block, it probably is a bit redundant to keep the conditional above (lines 193-194), no?

Lukasa: Yup, agreed. Let's remove the above.

test/test_util.py
```diff
@@ -95,6 +96,8 @@ def test_invalid_host(self):
             'http://google.com:foo',
             'http://::1/',
             'http://::1:80/',
+            'http://google.com:-80',
+            six.u('http://google.com:\xb2\xb2'),  # \xb2 = ^2
```
Lukasa: It seems like we don't have tests for this change in Retry-After: can we check that?

urllib3/util/url.py
Lukasa: Actually, I've changed my mind here. I agree ports need to be integers, so the `isdigit` check remains valuable.

nateprewitt: If `port` is not an int, it will fail in the try `port = int(port)`, and raise the same error in the except block. Isn't that the same functionality?

scop: It's getting pretty much impossible to follow all the comments here, but I'll try to summarize the parse_url change:

port.isdigit() will catch most non-integers, but it will let things such as ² (superscript 2) pass. Therefore the try...except is necessary around int(port), it will catch these. On the other hand, int(port) alone is not enough, because it will allow whitespace, plus and minus prefixes to pass. So the code should be correct as is. The alternative re.match approach I suggested would make the intent and code easier to understand though.

Lukasa: Swapping to regexes here doesn't really work because we allow both unicode and bytes URIs, so we'd need two branches with two different regexes. Given that, it might be better just to drop a comment here that explains the thinking.

test/test_util.py
scop: Sure, some retry-after checks could be added to test_util. But before doing that I'd like to know your thoughts on the exact behavior it should have regarding input such as negative integers, positive with + prefix, and surrounding whitespace (see my initial comment in this PR).

Lukasa: Surrounding whitespace is acceptable: everything else is not. Per [RFC 7231 Section 7.1.3](https://tools.ietf.org/html/rfc7231#section-7.1.3), the value is "is a non-negative decimal integer, representing time in seconds."

scop:
Revised commit force-pushed with requested changes: added comment to parse_url, and requested behavior and tests for parse_retry_after

urllib3/util/retry.py
```diff
@@ -205,7 +206,7 @@ def get_backoff_time(self):
         return min(self.BACKOFF_MAX, backoff_value)
 
     def parse_retry_after(self, retry_after):
-        if retry_after.isdigit():
+        if re.match(r"^\s*[0-9]+\s*$", retry_after):
```
Lukasa: Can we get a comment here that references the relevant RFC to explain why we allow whitespace?

scop: Actually in my opinion that method should be private and not allow whitespace. It's called only through get_retry_after which gets the Retry-After value from a response, and it is in my opinion the business of the response object to supply header values with nonsignificant whitespace stripped.

Lukasa: I'd rather that we have resilience here. There's no reason to have this implicit requirement (that the headers be stripped of extra whitespace) if we can just tolerate the whitespace.

scop:
Split to two commits and added the requested note.

PR #990, opened 2016-09-29
Make HTTPResponse.stream() work with _fp of non-HTTPResponse type (eg StringIO)

frankier:
Previously urllib's HTTPResponse would work correctly with any file like object. The 3rd party project requests_mock takes advantage of this by passing in a StringIO instance as the body (see: http://git.openstack.org/cgit/openstack/requests-mock/tree/requests_mock/response.py#n166 ). With the addition of chunked streaming in https://github.com/shazow/urllib3/commit/f21c2a2b73e4256ba2787f8470dbee6872987d2d stream() now assumes for chunked responses that if body is file-like then it is httplib.HTTPResponse-like breaking requests_mock.

This patch adds a guard in stream() to fall back to the old behaviour (delegate to the underlying file-like object as for unchunked responses) and another one to read_chunked() (fail early with an explicit error message).

Please let me know if there's any improvements I should make to get this merged or if you think this fix belongs elsewhere.

### DIFF
--- a/urllib3/exceptions.py
+++ b/urllib3/exceptions.py
@@ -196,6 +196,14 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
+class BodyNotHttplibCompatible(AssertionError, ValueError):
+    """
+    Body should be httplib.HTTPResponse like (have an fp attribute which
+    returns raw chunks) for read_chunked().
+    """
+    pass
+
+
 class IncompleteRead(HTTPError, httplib_IncompleteRead):
     """
     Response length doesn't match expected Content-Length
--- a/urllib3/exceptions.py
+++ b/urllib3/exceptions.py
@@ -196,7 +196,7 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
-class BodyNotHttplibCompatible(AssertionError, ValueError):
+class BodyNotHttplibCompatible(HTTPError):
     """
     Body should be httplib.HTTPResponse like (have an fp attribute which
     returns raw chunks) for read_chunked().
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -8,7 +8,7 @@ from socket import error as SocketError
 
 from ._collections import HTTPHeaderDict
 from .exceptions import (
-    ProtocolError, DecodeError, ReadTimeoutError,
+    BodyNotHttplibCompatible, ProtocolError, DecodeError, ReadTimeoutError,
     ResponseNotChunked, IncompleteRead, InvalidHeader
 )
 from .packages.six import string_types as basestring, binary_type, PY3
@@ -424,7 +424,7 @@ class HTTPResponse(io.IOBase):
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
-        if self.chunked:
+        if self.chunked and self.supports_chunked_reads():
             for line in self.read_chunked(amt, decode_content=decode_content):
                 yield line
         else:
@@ -515,6 +515,15 @@ class HTTPResponse(io.IOBase):
             b[:len(temp)] = temp
             return len(temp)
 
+    def supports_chunked_reads(self):
+        """
+        Checks if the underlying file-like object looks like a
+        httplib.HTTPResponse object. We do this by testing for the fp
+        attribute. If it is present we assume it returns raw chunks as
+        processed by read_chunked().
+        """
+        return hasattr(self._fp, 'fp')
+
     def _update_chunk_length(self):
         # First, we'll figure out length of a chunk and then
         # we'll try to read it from socket.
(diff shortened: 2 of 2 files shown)


frankier:
Good call. Done.

urllib3/response.py
```diff
@@ -515,6 +515,13 @@ def readinto(self, b):
             b[:len(temp)] = temp
             return len(temp)
 
+    def supports_chunked_reads(self):
+        """
+        Checks if the underlying file-like object looks like a
+        httplib.HTTPResponse object.
```
Lukasa: Can we elaborate on this docstring a little bit? The docstring and the name seem to be unrelated. Basically we can just use some of the terminology from `BodyNotHttplibCompatible`.

frankier:
Okay done. Pretty sure there should be no confusion for any readers who understand what HTTP chunking is now. [I hope ;)]

sigmavirus24:
Hi @frankier it looks like you need to rebase your PR and that you need to account for missed test coverage on Python 3.3.

frankier:
Rebased.

It says I'm missing test coverage in urllib3.util.retry but I didn't modify this module. See here: https://travis-ci.org/shazow/urllib3/jobs/164224223

Of course I could write a test if it seemed like the missing coverage was related to my patch, but in this case that doesn't immediately appear to be the case. Particularly odd is that it only fails for one Python version. Any ideas?

frankier:
Could someone with write access to the repository please trigger a rebuild of this on Travis? (I am unable to.) I think it's a transient error. All my previous versions worked fine and all other Python versions work fine.

haikuginger:
Okay, so I've got one concern on this - I'm not sure I like subclassing AssertionError. It feels a bit off. If @sigmavirus24 seconds me on that, let's go ahead and remove that; otherwise, I think this is about ready for merge.

urllib3/exceptions.py
```diff
@@ -196,6 +196,14 @@ class ResponseNotChunked(ProtocolError, ValueError):
     pass
 
 
+class BodyNotHttplibCompatible(AssertionError, ValueError):
```
sigmavirus24: Yeah, this should not inherit from either of these. Good catch @haikuginger

haikuginger: Hmmm - not sure on ValueError. It seems like we probably want something similar to the existing RequestNotChunked exception.

But we should certainly inheriting from some form of urllib3's `HTTPError`.

frankier: Okay, the new commit stops exceptions in urllib3.exceptions from participating in Python's built-in class hierarchy and adds a new base for urllib3 as @sigmavirus24 suggested.

Lukasa: Sorry, why does HTTPWarning need to inherit from Urllib3Exception? I think this change was overbroad. HTTPError is already the base exception class for urllib3, so the only change required was to not inherit from AssertionError or ValueError, and just to inherit from HTTPError.

frankier: Hey, whatever's most likely to get it merged. Never really been too sure how useful exception taxonomies are in the first place.

Lukasa: I'd like to see what @sigmavirus24's thinking was here: I'm wrong frequently enough that I like it when others confirm my thoughts. =D

sigmavirus24: HTTPError, to me, has a connotation that something happened during an actual connection (and most of the exceptions already in existence here confirm that). We can still use it as a base for most of our exceptions, but some of our exceptions don't inherit from that at all, so it's actually not our root exception and in my experience, users like having one last catch-all except block around a library (in case something funky goes wrong).

I think I was overzealous in asking HTTPWarning to inherit from Urllib3Exception. We don't raise that or it's sub-classes as exceptions so we can revert that.

Lukasa: Hrm. I understand that concern, but I think that frankly we have a class defined as the base exception for the library, and that we should use it as much as possible rather than add a new backward-compat version. The name isn't ideal, but I'm not too fussed about that tbh. =(

sigmavirus24: Well, the problem (as you can see from this PR after my requested changes) is that not every exception was inheriting from that to begin with. The pattern that @frankier chose to start with was already present (and kind of weird, at that).

We can't add HTTPError to those exception classes otherwise we'll marginally break people who have exceptions structured like


```py
try:
    # ...
except urllib3.exceptions.HTTPError:
    # ...
except urllib3.exceptions.ProxySchemeUnknown:
    # ...
```

So I do think we might want to restructure (although maybe right now is not the best time).

Lukasa: Let's avoid the scope creep here and restrict ourselves to the specific fix in question here. A separate issue to restructure the exception heirarchy is a good idea though.

PR #792, opened 2016-01-25
Cast OpenSSL errors to strings.

Lukasa:
Resolves #556.

I haven't written any tests for this because of #791. I'll start trying to work on #791 and, if we can get it ironed out quickly, we should block the merge of this behind #791 so that we can actually have a test for this.

Issue #556, opened 2015-02-28
Need to update error handling for pyOpenSSL


sigmavirus24:
If an `SSLError` comes from `pyOpenSSL`, we attempt to determine if it's an error due to request timeouts on python 2.6 being raised as `BaseSSLErrors` [here](https://github.com/shazow/urllib3/blob/29aa09bde9c42cc9a8d79aac47ee3d362b438cca/urllib3/connectionpool.py#L314). If the handshake ends unexpectedly under pyOpenSSL, then we get an `SSLError` that has a captured `SysCallError` which when you call `str(SSLError(msg, SysCallError(code, msg)))` you get

```py
TypeError: __str__ returned non-string (type SysCallError)
```

So we might want to update how we handle SSLErrors to accomodate pyOpenSSL. To be clear, I observed this on: python 2.7.9 with openssl 1.0.1l (from brew) and I was using requests to debug an issue with `https://apissl.cloudfactory.com`, e.g., `requests.get('https://apissl.cloudfactory.com')`

shazow:
Yea that whole process is voodoo. :/

Open to basically any kind of change. What do you have in mind?

sigmavirus24:
I don't know yet. =P

shazow:
Somedays I wonder if we should just fork out to golang. :P

Lukasa:
A note that pyOpenSSL is no longer required for SNI in Python 2.7.9. =) Obviously I don't think that removes it from the support matrix, but it's worth being aware of.

sigmavirus24:
Right. I was just trying to see if that would help with the problem the user was encountering with apissl.cloudfactory.com. Works just fine otherwise but it blows up here.

fuag15:
Blows up to other api's as well but I'm not at liberty to divulge a uri, would be very interested in a fix for this

lorenzog:
I had to fix this on a production server by editing `requests/packages/urllib3/connectionpool.py`. At line 315, before `if 'timed out' in str(err) or 'did not complete (read)' in str(err):` I inserted:

    if not isinstance(err, str):
            raise ProtocolError(self, url, "{}".format(err.strerror))

At least I get some inkling on what's going on - in my case this bubbles up to my application from `requests` as:

[...]

PR #992, opened 2016-09-30
String-ify more OpenSSL errors to fix TypeError: __str__ returned non-string (type Error)

sethp-jive:
Addresses the rather cryptic message:

```
  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 70, in get
  <...>
  File "/usr/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 355, in _make_request
    self._raise_timeout(err=e, url=url, timeout_value=conn.timeout)
  File "/usr/local/lib/python2.7/site-packages/requests/packages/urllib3/connectionpool.py", line 325, in _raise_timeout
    if 'timed out' in str(err) or 'did not complete (read)' in str(err):  # Python 2.6
TypeError: __str__ returned non-string (type Error)
```

Which can arise when using the `pyopenssl` contrib module and setting bad verification paths (e.g. file not found, invalid format, etc.).

An easy way to reproduce:

```
$ pip install requests ndg-httpsclient
$ python -c "import requests; requests.get('https://python.org', verify='/no/file')"
```

I think this is a similar issue as to #556? At least, I started down this rabbit hole from kennethreitz/requests#2524, but #792 wasn't quite enough to address it. With this change, I see the much better (though not quite good) error:

```
requests.exceptions.SSLError: ('bad ca_certs ("/no/file"): Error([(\'system library\', \'fopen\', \'No such file or directory\'), (\'BIO routines\', \'BIO_new_file\', \'no such file\'), (\'x509 certificate routines\', \'X509_load_cert_crl_file\', \'system lib\')],)',)
```

Feedback appreciated!

### DIFF
--- a/test/contrib/test_pyopenssl.py
+++ b/test/contrib/test_pyopenssl.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 import unittest
+import urllib3
 
 from nose.plugins.skip import SkipTest
+from urllib3.exceptions import SSLError
 from urllib3.packages import six
 
 try:
@@ -55,3 +57,15 @@ class TestPyOpenSSLHelpers(unittest.TestCase):
         expected_result = '*.xn--p1b6ci4b4b3a.xn--11b5bs8d'
 
         self.assertEqual(_dnsname_to_stdlib(name), expected_result)
+
+class TestPyOpenSSLErrorHandling(unittest.TestCase):
+    """
+    Tests for wrapping OpenSSL errors in urllib3 errors
+    """
+    def test_invalid_ca_certs(self):
+        """
+        Checks that we get an urllib3.exceptions.SSLError and not openssl.Error with missing ca_certs.
+        """
+        with self.assertRaises(SSLError):
+            http = urllib3.PoolManager(ca_certs='/no/path', cert_reqs='CERT_REQUIRED')
+            http.request('GET', 'https://python.org')
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -386,7 +386,10 @@ class PyOpenSSLContext(object):
             cafile = cafile.encode('utf-8')
         if capath is not None:
             capath = capath.encode('utf-8')
-        self._ctx.load_verify_locations(cafile, capath)
+        try:
+            self._ctx.load_verify_locations(cafile, capath)
+        except OpenSSL.SSL.Error as e:
+            raise ssl.SSLError('bad ca_certs ("%s"): %r' % (cafile or capath, e))
         if cadata is not None:
             self._ctx.load_verify_locations(BytesIO(cadata))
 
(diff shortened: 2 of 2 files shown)


Lukasa:
Can you please confirm that your problem is reproducible on the latest Requests? We should have fixed it.

sethp-jive:
I have confirmed that this is happening in requests 2.11.1 (the latest I can get with `pip install requests`), but only when `ndg-httpsclient` is also installed (otherwise, the pyopenssl contrib module isn't loaded due to the [guarded import](https://github.com/shazow/urllib3/blob/1.16/urllib3/contrib/pyopenssl.py#L48-L52)).

I haven't tried urllib3 v1.18, but since it looks like the code is still propagating the non-stringifiable OpenSSL `Error` directly (see change), I would expect the problem to still present, albeit triggered by a different set of installed packages.

sethp-jive:
@Lukasa not to pester, but step 4 of https://urllib3.readthedocs.io/en/latest/contributing.html suggests I should gently remind you to come back around to this :)

haikuginger:
@sethp-jive, @Lukasa is on vacation at the moment; he'll get back to you when he gets back.

Lukasa:
@sethp-jive Thanks for pestering! Sorry, it's hard to effectively notify people when maintainers are on a temporary absence. =(

So while this fix is good, it will interestingly not help the next release of requests. This is because we have changed the code path that urllib3 uses, which means that the except block in question is not used. You can see this by running the urllib3-specific version of this code:

```python
import urllib3

http = urllib3.PoolManager(ca_certs='/no/path', cert_reqs='CERT_REQUIRED')
http.request('GET', 'https://python.org')
```

On my machine, this raises a `OpenSSL.SSL.Error`, which shows that we have lost our except block that is supposed to wrap this error. For that reason, I'd actually like an alternative patch. Would you be willing to move this except block into `load_verify_locations` and apply the fix there? You can then remove the except block from `ssl_wrap_socket` in that file. The fix will then apply in new versions of Requests, which will use the new PyOpenSSL code path.

Is that likely to be suitable?

sethp-jive:
No worries – I hope you had a good break!

I've moved the catch/rethrow block as you requested, and it appears to be working fine on my machine. Does the new patch look OK to you?

urllib3/contrib/pyopenssl.py
```diff
@@ -386,7 +386,10 @@ def load_verify_locations(self, cafile=None, capath=None, cadata=None):
             cafile = cafile.encode('utf-8')
         if capath is not None:
             capath = capath.encode('utf-8')
-        self._ctx.load_verify_locations(cafile, capath)
+        try:
+            self._ctx.load_verify_locations(cafile, capath)
+        except OpenSSL.SSL.Error as e:
+            raise ssl.SSLError('bad ca_certs ("%s"): %r' % (cafile or capath, e))
```
Lukasa: Let's show both strings, inputs, rather than just cafile or capath.

sethp-jive:
Yup, I'll write a test. The failure in CI seems spurious (address already in use), have you seen that one before or should I dig in to it?

Lukasa:
Yeah, that'll be spurious.

test/contrib/test_pyopenssl.py
```diff
@@ -55,3 +57,15 @@ def test_dnsname_to_stdlib_leading_splat(self):
...
+    """
+    Tests for wrapping OpenSSL errors in urllib3 errors
+    """
+    def test_invalid_ca_certs(self):
+        """
+        Checks that we get an urllib3.exceptions.SSLError and not openssl.Error with missing ca_certs.
+        """
+        with self.assertRaises(SSLError):
+            http = urllib3.PoolManager(ca_certs='/no/path', cert_reqs='CERT_REQUIRED')
+            http.request('GET', 'https://python.org')
```
Lukasa: This is a good effort, but it would be better if this could go in `test_https` and could attempt to contact the test server. It generally works a bit better to avoid the possibility of making outbound connections if at all possible.

sethp-jive: It won't get that far – the exception is thrown before the connection is made. It just has to resolve – would `https://localhost` be better?

PR #993, opened 2016-10-01
Transport Security Manager HSTS support

kislyuk:
@Lukasa please review.

This is a follow-on to #972. It establishes HSTS support with an in-memory only store. I'll introduce preload list ingestion (via an optional external package) and a persistent store later.

### DIFF
--- a/urllib3/transport_security.py
+++ b/urllib3/transport_security.py
@@ -112,7 +112,8 @@ class TransportSecurityStore(object):
                                  expires=expires)
 
     def invalidate_host(self, host):
-        del self._store[host]
+        if host in self._store:
+            del self._store[host]
 
     def get_pins(self, host):
         raise NotImplementedError()
--- a/urllib3/transport_security.py
+++ b/urllib3/transport_security.py
@@ -43,7 +43,7 @@ class TransportSecurityManager(object):
         """
         if not isinstance(conn, HTTPSConnection):
             if self._tss.requires_https(conn.host):
-                msg = 'Host {} has set an HSTS policy preventing plain HTTP connections'
+                msg = 'Host {0} has set an HSTS policy preventing plain HTTP connections'
                 raise HSTSError(msg.format(conn.host))
 
     def validate_hpkp(self, conn):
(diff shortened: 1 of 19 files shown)


kislyuk:
ping @Lukasa

haikuginger:
Hey @kislyuk, @Lukasa's on vacation at the moment, but I'm sure he'll be glad to take a look when he gets back.

kislyuk:
OK, thanks @haikuginger.

urllib3/transport_security.py
```diff
@@ -46,30 +57,112 @@ def validate_hpkp(self, conn):
...
     def get_pins(self, host):
-        raise NotImplementedError("Must be overridden.")
+        raise NotImplementedError()
 
     def requires_https(self, host):
-        raise NotImplementedError("Must be overridden.")
+        parts = host.split(".")
+        for i in range(len(parts)-1):
+            superdomain = ".".join(parts[i:])
+            if superdomain in self._store:
```
Lukasa: I'd rather see a `try...except KeyError: continue` than this `if` statement. =)

urllib3/transport_security.py
```diff
@@ -46,30 +57,112 @@ def validate_hpkp(self, conn):
...
                    max_age=None):
-        raise NotImplementedError("Must be overridden.")
+        max_age = min(int(max_age), self.MAX_AGE_LIMIT)
+        expires = int(time.time()) + max_age
+        self._store[host].update(force_https=force_https, include_subdomains=include_subdomains,
+                                 expires=expires)
 
     def invalidate_host(self, host):
-        raise NotImplementedError("Must be overridden.")
+        del self._store[host]
```
Lukasa: I think this isn't right. Firstly, `invalidate_host` right now is called when HSTS sets `max-age=0`. That's fine, but when the host record is extended to include HPKP records this won't be right: simply turning off HSTS is not a reason to remove HPKP. Secondly, it makes `force_https` a redundant field: it can only ever be `True`, because `False` is equivalent to "there is no record".

That means I think we need two changes. Firstly, we need to be able to distinguish "there is no record" from "there is a record with HSTS not enabled". That's required for HPKP, and since we know it's coming we should build it in here.

Secondly, we should add newer methods for switching the status of a host. `store_host` should be able to store only the relevant fields for a given property (HSTS or HPKP).

Lukasa:
@kislyuk Looks like you accidentally merged master into this PR. It'd be better to merge master into the feature branch than to merge it in here.

kislyuk:
@shazow, I don't have access to the feature branch, so please merge master into it.

Lukasa:
@kislyuk I was going to suggest you just rebase the merge out of this branch and then force push.

kislyuk:
OK, I will do that.

jonparrott:

PR #995, opened 2016-10-09
flake8 fixes

scop:
The import order ones are triggered by the flake8-import-order extension, others plain flake8 (3.0.4, pycodestyle 2.0.0)

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -135,9 +135,9 @@ class AppEngineManager(RequestMethods):
 
         try:
             follow_redirects = (
-                    redirect and
-                    retries.redirect != 0 and
-                    retries.total)
+                redirect and
+                retries.redirect != 0 and
+                retries.total)
             response = urlfetch.fetch(
                 url,
                 payload=body,
--- a/urllib3/contrib/socks.py
+++ b/urllib3/contrib/socks.py
@@ -33,9 +33,7 @@ except ImportError:
         'SOCKS support in urllib3 requires the installation of optional '
         'dependencies: specifically, PySocks.  For more information, see '
         'https://urllib3.readthedocs.io/en/latest/contrib.html#socks-proxies'
-        ),
-        DependencyWarning
-    )
+    ), DependencyWarning)
     raise
 
 from socket import error as SocketError, timeout as SocketTimeout


haikuginger:
I agree with @sigmavirus24 on all points.

scop:
Considering that the vast majority of imports in urllib3 are actually ordered, I think it's a fair assumption to make that that's desirable. flake8-import-order dependency added.

The other changes in the "flake8 fixes" commit are not related to any new plugins and I have no idea why those issues would not be caught by the flake8-py3 tox env run. They certainly are for my local flake8 runs, with apparently the same flake8 and friends versions as the tox env has, but FWIW I'm not running them within tox.

sigmavirus24:
@scop look at the flake8 job now that you've added the import order plugin. Far more files in `urllib3` are reporting ordering errors. I don't think that the "vast majority" are ordered appropriately. Even if they were, I'd question whether they were intentionally ordered according to a specific style.

> I have no idea why those issues would not be caught by the flake8-py3 tox env run.

If you can discover why they aren't caught by that, that would be immensely helpful.

scop:
I still think that the vast majority of imports in urllib3 are ordered. It's just that the import-order plugin also checks for the order of names imported like "from foo import name1, name2, ...", they're not currently ordered, and those are the new ones reported now. Anyway, I'd appreciate if you could state with certainty whether this is desirable or not, and I'll keep in or remove that part accordingly.

Lukasa:
I have no objection to adding an import-order check. However, if we're going to do it, we need to do it properly: please fix up the failing build. =)

scop:
Ok, some serious confusion here. First, I now see a lot of messages from the import-order plugin that I didn't notice before. Don't know how I managed to miss those. Second, I took a look at fixing those, but ran into a bug in import-order so I decided to forget about that. Third, the non-import-order related errors show up only if the import-order plugin is installed, even if the messages have nothing to do with it, and its own messages are disabled.

So... I know how to reproduce a setup where those couple of messages occur, and they seem legit, but I don't have an explanation why exactly they occur only when the seemingly unrelated import-order plugin is installed. I left them there anyway in case you want to merge them, and I don't plan to add any more stuff to this PR for the time being.

Lukasa:
M'kay, I think it's only worth merging this if we can commit to keeping things cleaned up. If we can't enforce them at merge time, I'm pretty unconvinced that the churn is worthwhile. Thanks for the work any @scop, the thought was a good one!

Issue #797, opened 2016-01-27
Implement Happy Eyeballs to support IPv4/IPv6 combination


xsuchy:
In case host have IPv4 and IPv6 address but the route for one of them is broken, then user using urllib3 cannot connect to that URL.

This can be avoided (beside fixing the route, which is not always possible) by implementing Happy Eyeballs specification:
http://tools.ietf.org/html/rfc6555

Lukasa:
Implementing Happy Eyeballs is really tricky for us, and potentially impossible, because urllib3 has a policy of not spawning its own threads. Because of the unpleasant way in which Python threads interact with the interpreter, libraries that spawn their own threads can have all kinds of unexpected effects on user code. This is a bad time, so we simply don't do it.

We could *in principle* use happy eyeballs temporarily switching the sockets to non-blocking mode, issuing all the connect requests at once, selecting on them all, and then switching them back to blocking once we know which one we want to use. That's worth considering, I suppose, but it'll be quite a chunk of work.

Lukasa:
Ok, I actually believe we can do this.

The code is going to be crazy complex, and at this point we really will want to factor it out into a helper, but it's definitely do-able. I've written a version that isn't entirely spec-compliant, in part because it connects to every single DNS result at once (not entirely web-friendly), but it's do-able.

However, this is sufficiently complex that I'd want us to really factor this out aggressively. There's a lot of complexity involved in doing this properly: we'll want to cache successful results, we'll want to properly handle the case where the first two addresses we find both fail to connect, and we need to make sure that error reporting continues to function properly.

I think this is a good long term goal, but I also think it's one that is best suited to a BountySource style approach. I've uploaded my initial test implementation [here](https://github.com/Lukasa/urllib3/tree/happy-eyeballs): anyone wanting to take this work on is welcome to use it as a basis for it.

Alternatively, we can hope that I'll eventually get around to this!

graingert:
ah I now see this is discussed here: https://github.com/urllib3/urllib3/pull/1000

pquentin:
Removed the bounty per https://github.com/urllib3/urllib3/pull/3112#issuecomment-1702173568. We can revisit this issue after async support. #3038 is the best way to move this forward.

pemensik:
Since #3038 is completed, is there possible way forward?

pemensik:
[...]

PR #998, opened 2016-10-18
Add poll, epoll, kevent/kqueue/Replace select.select with util functions

tbsf:
This is an update to PR #685 where I implement kevent/kqueue as well as poll/epoll. Feel free to refactor or integrate into that PR.

PR #999, opened 2016-10-18
Happy Eyes Protocol

SethMichaelLarson:
Hello, this PR is for issue #797 and adds upon the work and takes the recommendations of @Lukasa in order to implement a subset of the Happy Eyes Protocol ([RFC 6555](https://tools.ietf.org/html/rfc6555)) to decrease the performance hit that dual-stack clients feel when connecting to a server with advertised IPv6 addresses but for whatever reason cannot be connected to via IPv6. Without this protocol in place, clients attempting to connect to a service with broken IPv6 will see a long timeout for IPv6 before their system will try to connect via IPv4 which punishes dual-stack users.

The suggestion from @Lukasa was to break this into it's own helper which this PR accomplishes. This PR also has support for caching Happy Eyes results for faster use.

Currently there are two tests failing on my machine (Receving a ProxyError rather than a ConnectionTimeoutError) but because I don't have very much knowledge about urllib3's code-base I needed some help from other maintainers debugging them. 

I currently have not added any tests to this PR because I want to know the best way to add tests that are for an optional protocol for creating a connection such as this, with suggestions on the best way to test this protocol as well as the `create_connection` method without the Happy Eyes protocol enabled. Criticism welcome.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@ dev (master)
 
 * Disallow superscripts and other integerish things in URL ports. (Issue #989)
 
+* Allow urllib3's HTTPResponse.stream() method to continue to work with
+  non-httplib underlying FPs. (Pull #990)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 


SethMichaelLarson:
This PR would also be improved by using the changes in #998 as currently it only uses `select` as opposed to `poll`, `epoll`, `kqueue` or `devpoll`. I will monitor that PR and add changes here if that PR is merged.

PR #1000, opened 2016-10-19
Happy Eyeballs Algorithm

SethMichaelLarson:
Hello, this PR is for issue #797 and adds upon the work and takes the recommendations of @Lukasa in order to implement a subset of the Happy Eyeballs Algorithm ([RFC 6555](https://tools.ietf.org/html/rfc6555)) to decrease the performance hit that dual-stack clients feel when connecting to a server with advertised IPv6 addresses but for whatever reason cannot be connected to via IPv6. Without this protocol in place, clients attempting to connect to a service with broken IPv6 will see a long timeout for IPv6 before their system will try to connect via IPv4 which punishes dual-stack users.

The suggestion from @Lukasa was to break this into its own helper which this PR accomplishes. This PR also has support for caching Happy Eyes results for faster use.

Currently, there are two tests failing on my machine (Receiving a ProxyError rather than a ConnectionTimeoutError) but because I don't have very much knowledge about urllib3's code-base I needed some help from other maintainers debugging them. 

I currently have not added any tests to this PR because I want to know the best way to add tests that are for an optional protocol for creating a connection such as this, with suggestions on the best way to test this protocol as well as the `create_connection` method without the Happy Eyes protocol enabled. Criticism welcome.


Lukasa:
So, let's begin by flagging this as dependent on the result of #1001. If we're going to have multiple parts of the codebase using the best available selector we should make sure that we make that available first.

I'll otherwise head through and do a review of this patch, excluding the use of `select`.

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+def _connect_happy_eyes(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
+                        source_address=None, socket_options=None):
+    """ Implements the Happy Eyes protocol (RFC 6555) which allows
+    multiple sockets to attempt to connect from different families
+    for better connect times for dual-stack clients where server
+    IPv6 service is advertised but broken. """
+    err = None
+
+    if address in HAPPY_EYES_CACHE:
+        family, proto, expires = HAPPY_EYES_CACHE[address]
```
Lukasa: `urllib3` needs to be thread-safe, but this use of dictionaries isn't. We may need to lock around this cache.

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+                if timeout is not socket._GLOBAL_DEFAULT_TIMEOUT:
+                    sock.settimeout(timeout)
+                if source_address:
+                    sock.bind(source_address)
+
+                sock.connect(address)
+                return sock
+            except socket.error as e:
+                err = e
+                sock.close()
```
Lukasa: Let's try to factor out this code block. It appears in a few places now, let's make it a helper function.

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+                sock = socket.socket(family, socket.SOCK_STREAM, proto)
+
+                _set_socket_options(sock, socket_options)
+
+                if timeout is not socket._GLOBAL_DEFAULT_TIMEOUT:
+                    sock.settimeout(timeout)
+                if source_address:
+                    sock.bind(source_address)
+
+                sock.connect(address)
```
Lukasa: I don't think this code logic is right. We still need to call `getaddrinfo` on the address. So let's try to reduce the cyclomatic complexity of this code a bit by using the getaddrinfo block below: the Happy Eyes cache should just cache family information.

This has the advantage of falling into line with RFC 6555, which says that implementations should cache what family is working, but not the specific result of the DNS lookup.

urllib3/util/connection.py
```diff
@@ -142,3 +278,6 @@ def _has_ipv6(host):
     return has_ipv6
 
 HAS_IPV6 = _has_ipv6('::1')
+HAS_HAPPY_EYES = select and HAS_IPV6
+HAPPY_EYES_CACHE = {}
+HAPPY_EYES_CACHE_TIME = 600
```
Lukasa: Let's specify that this is ten minutes, per RFC 6555.

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+    found = None
+    while socks:
+        select_timeout = timeout
+        if timeout is socket._GLOBAL_DEFAULT_TIMEOUT:
+            select_timeout = None
+
+        # Use select to detect which sockets are connected.
+        _, connected, _ = select([], socks, [], select_timeout)
+
+        # If no sockets are ready, then we timed out.
```
Lukasa: This statement is not actually unequivocally true on Python 2, because EINTR can cause this statement to return early. The `selectors` module would handle this.

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+            sock.settimeout(0)  # Make the socket non-blocking to use select.
+
+            # We should catch all errors except all non-blocking error codes.
+            errcode = sock.connect_ex(sa)
+            if errcode not in [errno.EINPROGRESS,
+                               errno.EAGAIN,
+                               errno.ENOTCONN]:
+                raise socket.error(errcode)
+
+            socks.append(sock)
```
Lukasa: We may need more complex logic than this.

RFC 6555 says a few things of note. The first thing it says is in Section 5.4, where it points out that:

> It is possible that a DNS query for an A or AAAA resource record will return more than one A or AAAA address.  When this occurs, it is RECOMMENDED that a Happy Eyeballs implementation order the responses following the host's address preference policy and then try the first address.  If that fails after a certain time (see Section 5.5), the next address SHOULD be the IPv4 address.
>
> If that fails to connect after a certain time (see Section 5.5), a Happy Eyeballs implementation SHOULD try the other addresses returned; the order of these connection attempts is not important.

In Section 5.5, it says:

> It is RECOMMENDED that connection attempts be paced to give connections a chance to complete.  It is RECOMMENDED that connection attempts be paced 150-250 ms apart to balance human factors against network load.

So what we may want to do is to take this result and iterate over it, opening a connection, adding it to the selector pool, and then waiting for select to return with a timeout of 200ms before we create the next connection. We wouldn't remove the other connection from the pool, of course: the pool would just get slowly bigger until we finally connect.

This would require some careful math to make sure we don't exceed the total connection timeout, but I think it'd provide a much better behaviour.

Are you up for implementing this?

urllib3/util/connection.py
```diff
@@ -108,6 +116,134 @@ def _set_socket_options(sock, options):
...
+    multiple sockets to attempt to connect from different families
+    for better connect times for dual-stack clients where server
+    IPv6 service is advertised but broken. """
+    err = None
+
+    if address in HAPPY_EYES_CACHE:
+        family, proto, expires = HAPPY_EYES_CACHE[address]
+
+        # If the cache entry is expired, don't use it.
+        if time.time() < expires:
```
Lukasa: As always, the use of `time` instead of `monotime` is frustrating, but stupid old Python versions don't have a better function to call. =(

SethMichaelLarson: We can use `monotime` if it's available otherwise stick with `time`? Is that sufficient?

Lukasa: I think that would be a better experience, yeah.

urllib3/util/connection.py
SethMichaelLarson: Ah, sure I can reorder this logic to only cache the family.

urllib3/util/connection.py
SethMichaelLarson: I sure am. I'll take my best stab at it and put it up for review again.

urllib3/util/connection.py
SethMichaelLarson: Another good reason to vendor `selectors`. Perhaps we should think about doing this.

urllib3/util/connection.py
Lukasa: Well, technically the dictionary is thread-safe. The problem here is the LBYL logic in place here. Ideally we should just refactor this to a `try...except`.

urllib3/util/connection.py
SethMichaelLarson: Sure I'll add a comment.

SethMichaelLarson:
Thank you @Lukasa for the great review. I'll work on this more today.

urllib3/util/connection.py
Lukasa: We should either vendor `selectors`, or if we're really worried about the overhead (which we might be), we should start from `selectors` and rip out the guts of that module to reduce it to just what we need. I'd accept either model.

SethMichaelLarson: What I currently am doing is the 'guts' of the module but of course rip the guts out of anything it looks terrible as the current code I have does. I think just using `selectors` at this point is probably our cleanest and safest bet.

SethMichaelLarson: I'll do the speed comparison as well, as you know I'm a benchmark guy. ;)

Lukasa: Yup, sounds good. =D

SethMichaelLarson: I'll also note that I've done work on the vendored `selectors` library for asyncio to improve the handling fds for the `SelectSelector`, as Windows is stuck with it for urllib3 it may be a good idea to port those changes here as well.

Lukasa: @SethMichaelLarson So the backport of selectors on PyPI seems to be stuck at Python 3.4 level. I wonder if we should ask the maintainers to update to a newer version?

SethMichaelLarson: My guess is they don't want to handle the whole issue with interrupts that you described earlier. Perhaps we should vendor in and simply default to `SelectSelector` for Python <3.3 with restarts on `InterruptError` when they happen during selecting?

Lukasa: Do we think that's it? I wonder. I feel like we should discuss this with upstream first.

Lukasa: Or alternatively, backport selectors from Python 3.5/3.6 ourselves instead.

SethMichaelLarson: Rather, it handles the `EINTR` but doesn't do what we want it to do which is retry the select.

Lukasa: What does it do with the `EINTR`?

SethMichaelLarson: Just returns that no fds are ready for reading or writing. This is not the same behavior that Python takes with [PEP 475](https://www.python.org/dev/peps/pep-0475/).

Lukasa: Gah, annoying. Ok.

So let me define the constraints of what I want here. I'd like us to have a library that provides an abstraction over the best `select` or `select`-like APIs available on a platform that is suitable both for the HappyEyeballs case and the PyOpenSSL contrib module case. It should handle `EINTR`. It should not bring in a bunch of unimportant code. And it should be maintainable.

If this can't be a straight backport or a third-party dependency, we need to treat it as product code. That means a full suite of tests, 100% test coverage, and all that jazz. We can derive this from selectors if that's a useful starting off point.

SethMichaelLarson: Hard part of getting 100% coverage on code like this is that `kqueue` and `devpoll` are BSD and Solaris only. One of may have to install those OSes to test the code in that case.

Lukasa: Well, that's not the real problem. The *real* problem is that we mandate the 100% coverage in CI. So having OS-specific code for OSes we can't CI on is going to be a problem.

Now, TravisCI has macOS, so that might cover `kqueue`. Not sure about `devpoll` though.

SethMichaelLarson: Could always not support `devpoll` and force Solaris to use `epoll` which is about as fast but is much more widespread.

Lukasa: I'm fine with dropping `devpoll` support. Oracle can give us a Jenkins if they want it back.

Lukasa: Yup, I got a goahead from @shazow. So let's start doing that.

Wow @SethMichaelLarson, you do have a way of stumbling into the biggest possible work items in a project, don't you!

SethMichaelLarson: @Lukasa What can I say, I'm a glutton for punishment (and helping out!) ;)

ashfall:
In hopes that this doesn't come across as too nitpicky, there's a major naming flaw that I would love to see addressed before I dig into reviewing further:

s/Eyes/Eyeballs
s/protocol/algorithm

Subtle difference, but different still :). Wikipedia says that you can call it a "fast fallback algorithm" also (https://en.wikipedia.org/wiki/Happy_Eyeballs).

SethMichaelLarson:
@ashfall Good point, I'll change those things in the next commit cycle.

ashfall:
Thanks! Could you please add some tests? Here are another implementation's unit tests for this algorithm that you might find useful as a reference: https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/test/test_endpoints.py#L1573

I'll also say that this not trivial to implement, and might need to go through some review cycles before it can be merge-ready, but it is an awesome algorithm and worth all of the effort! Thanks for all your work :)

SethMichaelLarson:
@ashfall Yes I will add some tests, thank you for the reference!
Thanks, I'm aware of the work load, but when it's done it'll feel great. I'm ready for a long haul for sure, especially because I'm also implementing features that are blocking this feature! :P

SethMichaelLarson:
@Lukasa What is the proper TimeoutError to raise when a socket times out during a connection?

SethMichaelLarson:
So I've done a lot of work here, it will take a while to review and I definitely need to add many test cases from @ashfall's reference. I'm back to the same two failing tests.  NOTE: I added a temporary selectors.py file so I can work on this while selectors is still awaiting merge. I will do a proper merge of the files once #1001 is merged.

Lukasa:
The appropriate error to raise, @SethMichaelLarson, is going to be `socket.timeout`. =)

SethMichaelLarson:
@Lukasa Okay just making sure there's not some higher level exception that other code expects me to throw :)

SethMichaelLarson:
Now that selectors are mostly finalized I'm going to start work again on this PR. :)

SethMichaelLarson:
Just putting this here but the "fail" from Travis is not quite a fail, the CI failed on flake (simple fix) and for Mac OS + Python 3.3, but all other Mac OS builders passed 100% and it looks like a random fail :) I still haven't added custom tests for the PR yet but that will come soon. :)  I hope this is headed in the direction you wanted it to go @Lukasa ?

SethMichaelLarson:
Soft ping :)

Lukasa:
Want to resolve the conflicts before I do a big review?

SethMichaelLarson:
@lukasa Sure thing!

test/with_dummyserver/test_connectionpool.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import with_statement
```
Lukasa: We don't appear to need this.

tox.ini
```diff
@@ -53,4 +53,4 @@ commands=
     make -C {toxinidir}/docs html
 whitelist_externals=
     make
-    rm
+    rm
```
Lukasa: Newline. =)

urllib3/util/connection.py
```diff
@@ -1,16 +1,19 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, with_statement
```
Lukasa: Once again, we don't appear to need this.

urllib3/util/connection.py
```diff
@@ -57,6 +61,12 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
     # The original create_connection function always returns all records.
     family = allowed_gai_family()
 
+    # If IPv6 and selectors are available, use the Happy Eyes algorithm.
+    # (RFC 6555 https://tools.ietf.org/html/rfc6555)
+    if HAS_IPV6 and HAS_SELECT and ENABLE_HAPPY_EYEBALLS:
+        return happy_eyeballs_algorithm((host, port), timeout,
+                                        source_address, socket_options)
```
Lukasa: It's a little bit disappointing that this code is in two separate places. It runs the risk of causing us to have these two blocks of code fall out of step. =(

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+    IPv6 service is advertised but broken. """
+    result = None  # This is the actual connected socket eventually.
+    err = None
+    family = 0
+    cached = False
+
+    # We need to keep track of the time so we don't exceed timeout.
+    start_time = monotonic()
+    if timeout is None:
+        timeout_time = None
```
Lukasa: Test coverage please. :smile:

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+        timeout_time = start_time + timeout
+
+    # Check the cache to see if our address is already there.
+    if address in _HAPPY_EYEBALLS_CACHE:
+        family, expires = _HAPPY_EYEBALLS_CACHE[address]
+
+        # If the cache entry is expired, don't use it.
+        if start_time > expires:
+            del _HAPPY_EYEBALLS_CACHE[address]
+            family = 0
```
Lukasa: Test coverage. :smile:

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+
+        dns_results = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+        dns_results_len = len(dns_results)
+
+        for i in range(dns_results_len):
+            af, socktype, proto, canonname, sa = dns_results[i]
+
+            # We only care about IPv4 and IPv6 addresses.
+            if af not in _IP_FAMILIES:
+                continue
```
Lukasa: Coverage. :smile:

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+    with DefaultSelector() as selector:
+        # Perform a DNS lookup for the address for IPv4 or IPv6 families.
+        if not family:
+            family = socket.AF_UNSPEC
+
+        dns_results = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+        dns_results_len = len(dns_results)
+
+        for i in range(dns_results_len):
+            af, socktype, proto, canonname, sa = dns_results[i]
```
Lukasa: Uh...why are we using old-school for loops here?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+            if af not in _IP_FAMILIES:
+                continue
+
+            sock = None
+            try:
+                sock = socket.socket(af, socktype, proto)
+                if timeout_time is socket._GLOBAL_DEFAULT_TIMEOUT:
+                    sock_timeout = sock.gettimeout()
+                    if sock_timeout is None:
+                        timeout_time = None
```
Lukasa: This will override the `_GLOBAL_DEFAULT_TIMEOUT` value, which may not be ideal. One way or another we should comment that we're doing it.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+                    else:
+                        timeout_time = start_time + sock_timeout
+
+                if socket_options:
+                    for opt in socket_options:
+                        sock.setsockopt(*opt)
```
Lukasa: Let's try to continue to use the utility function we wrote for this in `util.connection`. It may need to be hoisted into this module, but that's ok.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+                    sock.bind(source_address)
+
+                # Set non-blocking for selecting.
+                sock.settimeout(0.0)
+
+                # Connect to the host.
+                errcode = sock.connect_ex(sa)
+                if errcode and errcode not in _ASYNC_ERROR_NUMBERS:
+                    err = socket.error(errcode)
+                    continue
```
Lukasa: Ideally we'd like to call `close` here.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+
+                errcode = sock.getsockopt(socket.SOL_SOCKET, socket.SO_ERROR)
+                if errcode and errcode not in _ASYNC_ERROR_NUMBERS:
+                    err = socket.error(errcode)
+                    continue
```
Lukasa: Same here.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+
+                # Register this new socket with the selector.
+                socks.append(sock)
```
Lukasa: This comment is misplaced.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+        if err:
+            raise err
+
+        # Otherwise if there's no other errors we either timed out or were refused connection.
+        if (timeout is None or
+                timeout is socket._GLOBAL_DEFAULT_TIMEOUT or
+                start_time + timeout < monotonic()):
+            raise socket.timeout()
+        else:
+            raise socket.error(errno.ECONNREFUSED)
```
Lukasa: This method *desperately* needs refactoring: it's a giant chain of logic that makes it really hard to follow what's going on. =)

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+                        err = socket.error(errcode)
+                        socks.remove(sock)
+                        selector.unregister(sock)
+
+                # If we don't have any sockets, don't try selecting.
+                if not socks:
+                    # If there are no more sockets to look through, then
+                    # we should give up immediately with a non-timeout error.
+                    if i == dns_results_len - 1:
+                        err = socket.error(errno.ECONNREFUSED)
```
Lukasa: Rather than use `ECONNREFUSED`, it would be better to use the last error we got from any of our sockets.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,241 @@
...
+                if sock is not None:
+                    sock.close()
+
+            # This is where the selecting logic occurs. Once all sockets
+            # have been created and are added to the selector, they should
+            # be continually selected until one of them works.  If there are
+            # still sockets to add to the selector, then select for only 200
+            # ms before adding the next socket to the selector.
+            reattempt_select = True
+            while reattempt_select:
```
Lukasa: I have my suspicions about the correctness of this loop, but I can't easily tell if they're valid because the logic of this method is so complex.

urllib3/util/rfc6555.py
Lukasa: Once again, `ECONNREFUSED` doesn't seem right here.

urllib3/util/selectors.py
```diff
@@ -521,4 +521,4 @@ def close(self):
     def no_selector(_):
         raise ValueError("Platform does not have a selector")
     DefaultSelector = no_selector
-    HAS_SELECT = False
+    HAS_SELECT = False
```
Lukasa: Newline please! :sparkles:

SethMichaelLarson:
Thanks for the huge review @Lukasa I totally agree that this function needs some major refactoring, I tried a few times but it felt strange passing along so much state in the function to it's helpers.... I'll give it another shot. :)

Lukasa:
I thought you might feel concerned about the state passing, so here's an idea: consider making a class object. =) Classes are a good place to hold state, and that will allow you to pin some state on the class as you go.

SethMichaelLarson:
@Lukasa Ah okay, I'll try that out and see how it feels, can't be any worse than how it is now. ;)

Lukasa:
To be clear, you can consider the class an implementation detail: that is, the function just creates a class and then calls a method on it and returns the value. Users still just call a function, but the class is created under the hood.

SethMichaelLarson:
Yeah I can do that, that's definitely the best way to solve this.  This method will be looking a lot better after that. :)

urllib3/util/connection.py
SethMichaelLarson: Should all of this be brought into the same file? I knew this functionality would have a lot to it so I made it separate, if we want it all together I can do that. :)

urllib3/util/rfc6555.py
SethMichaelLarson: Because the index is used to detect the 'last' socket to try and if it is the last socket to just block the remaining time in the timeout.

Lukasa: Sad. =(

SethMichaelLarson:
So I moved to an object as was discussed, I believe it looks a lot better but still maybe a little hard to understand exactly what's going on without knowing the RFC. Would like some another quick review if possible @Lukasa :)

SethMichaelLarson:
Ignore the `last_socket` variable in `_wait_for_connection`, I think I'm going to drop that logic entirely and try to treat every socket the same whether it's the first or the last. That logic is a relic from the old implementation.

urllib3/util/connection.py
```diff
@@ -34,7 +37,8 @@ def is_connection_dropped(conn):  # Platform-specific
 # One additional modification is that we avoid binding to IPv6 servers
 # discovered in DNS if the system doesn't have IPv6 functionality.
 def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                      source_address=None, socket_options=None):
+                      source_address=None, socket_options=None,
+                      enable_happy_eyeballs=True):
```
nateprewitt: Sorry if missing something here, but is there a reason for both the global `ENABLE_HAPPY_EYEBALLS` and the `create_connection` param `enable_happy_eyeballs`? It currently looks like only the global is being used.

SethMichaelLarson: Good catch! This was added a long time ago before I realized that create_connection is buried really deep in the API, I wanted to give a way for users to disable the algorithm if they need to but I don't think it's worth it to expose it at a high level. I'll remove the param. :)

test/with_dummyserver/test_connectionpool.py
Lukasa: We still don't need this.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -618,6 +620,8 @@ def test_dns_error(self):
         self.assertRaises(MaxRetryError, pool.request, 'GET', '/test', retries=2)
 
     def test_source_address(self):
+        connection.ENABLE_HAPPY_EYEBALLS = False
+        self.addCleanup(setattr, connection, "ENABLE_HAPPY_EYEBALLS", True)
```
Lukasa: Let's put a comment here saying why we need happy eyeballs off.

urllib3/util/connection.py
Lukasa: Still don't need this.

urllib3/util/connection.py
Lukasa: I'm assuming this is getting addressed in the next round of changes.

urllib3/util/connection.py
Lukasa: I'm not sure about bringing it into the same file, I'm just suggesting that we should take a very close look to see if there is common code that can be factored out or used in both places.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+_HAPPY_EYEBALLS_CACHE_TIME = 60 * 10  # 10 minutes according to RFC 6555
+_ASYNC_ERROR_NUMBERS = set([errno.EINPROGRESS,
+                            errno.EAGAIN,
+                            errno.EWOULDBLOCK])
+if hasattr(errno, "WSAEWOULDBLOCK"):  # Platform-specific: Windows
+    _ASYNC_ERROR_NUMBERS.add(errno.WSAEWOULDBLOCK)
+
+_IP_FAMILIES = set([socket.AF_INET])
+if hasattr(socket, "AF_INET6"):
+    _IP_FAMILIES.add(socket.AF_INET6)
```
Lukasa: Here's a real question: is there any point in doing Happy Eyeballs if we don't have IPv6?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        self._source_address = source_address
+        self._socket_options = socket_options
+        self._sockets = []
+        self._selector = DefaultSelector()
+        self._start_time = None
+
+    def connect(self):
+        """ Given the setup of the algorithm, attempt to
+        use a cached value and if it doesn't work then try
+        to use a """
```
Lukasa: Try to use a what?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+                            errno.EWOULDBLOCK])
+if hasattr(errno, "WSAEWOULDBLOCK"):  # Platform-specific: Windows
+    _ASYNC_ERROR_NUMBERS.add(errno.WSAEWOULDBLOCK)
+
+_IP_FAMILIES = set([socket.AF_INET])
+if hasattr(socket, "AF_INET6"):
+    _IP_FAMILIES.add(socket.AF_INET6)
+
+
+class _HappyEyeballs(object):
```
Lukasa: We need an explanatory docstring for this class that indicates that it's a one-use-only thing that is mostly used to store a lot of common state that needs to be moved around.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+                # Invalidate the cache if it's expired.
+                if expires > current_time():
+                    del _HAPPY_EYEBALLS_CACHE[self._address]
+                else:
+                    try:
+                        sock = self._algorithm(family)
+                    # Invalidate the cache if connecting doesn't work.
+                    except (OSError, socket.error):
+                        del _HAPPY_EYEBALLS_CACHE[self._address]
```
Lukasa: This whole `if` block is a logical function, let's split it out.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+            if sock is None:
+                sock = self._algorithm()
+
+                # Add a new entry to the cache if successful.
+                cache_entry = (sock.family,
+                               current_time() + _HAPPY_EYEBALLS_CACHE_TIME)
+                _HAPPY_EYEBALLS_CACHE[self._address] = cache_entry
```
Lukasa: This is a logical function, let's split it out.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+            return sock
+        finally:
+            self._cleanup()
+
+    def _algorithm(self, family=None):
```
Lukasa: I don't love this function name.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        """ Executes the algorithm for a given family
+        or no family if IPv4 or IPv6 are allowed. """
+        if family is None:
+            family = socket.AF_UNSPEC
```
Lukasa: I don't think there's any reason to have the default of family be `None`: let's just set it to `socket.AF_UNSPEC` and save ourselves an if block.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        host, port = self._address
+        dns_results = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+        dns_results = list(filter(lambda *args: args[0] not in _IP_FAMILIES,
+                                  dns_results))
```
Lukasa: This is much clearer as a list comprehension:

```python
dns_results = [result for result in dns_results if result[0] in _IP_FAMILES]
```

I should note that while writing this as a list comprehension I spotted that your conditional appears to be backwards. :wink:

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+        result = None
+        for family, socktype, proto, _, sa in dns_results:
+            self._create_socket(family, socktype, proto, sa)
+            result = self._wait_for_connection(False)
+            if result:
+                break
+            self._timeout_check()
+
+        if result is None:
```
Lukasa: We can be a bit clever here: `if result is None` amounts to exiting the loop without hitting the break statement, so this `if` statement can actually be rewritten as an `else` clause to the `for` loop. Saves us a comparison and lets us use this little-used but occasionally helpful syntactic feature.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+            if self._error:
+                raise self._error
+            else:
+                raise socket.error(errno.ECONNREFUSED)
```
Lukasa: Still don't love this fallback to `ECONNREFUSED`.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # Restore the old timeout value and remove it from cleanup.
+        result.settimeout(self._timeout)
+        self._sockets.remove(result)
+        self._selector.unregister(result)
```
Lukasa: These three statements are better done inside `_wait_for_result`.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+        return result
+
+    def _wait_for_connection(self, last_socket=False):
+        """ Does the actual selecting on the sockets waiting
+        for one to be ready to be written to. """
+        # Make sure we don't have any errored sockets.
+        for sock in self._sockets:
+            self._socket_check(sock)
```
Lukasa: I think it wouldn't hurt to write a function called `_check_all_sockets` that just has this as a loop body.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # If there are no sockets to select on, then don't try.
+        if not self._sockets:
+            return None
```
Lukasa: This should come first in the function, as there's no point running the loop if there are no sockets.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # If it's the last socket then we can block until timeout
+        if last_socket:
+            if self._timeout is None:
+                select_time = None
+            else:
+                select_time = max(0.0, self._timeout - (current_time() - self._start_time))
+        # Otherwise block for 200ms like the RFC suggests.
+        else:
+            select_time = self._get_select_time()
```
Lukasa: Maybe change `get` to `calculate`?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+            sock = key.fileobj
+
+            # Successfully found a proper connected socket.
+            if self._socket_check(sock):
+                return sock
+
+        return None
+
+    def _create_socket(self, family, socktype, proto, sa):
+        """ Tries it's best to create a socket and register
```
Lukasa: Nit: s/it's/its/

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        the new socket with the selector. If it fails, no
+        problem, the algorithm continues as usual. """
+        sock = None
+        try:
+            sock = socket.socket(family, socktype, proto)
+            if self._timeout is socket._GLOBAL_DEFAULT_TIMEOUT:
+                self._timeout = sock.gettimeout()
```
Lukasa: I'm still not really sure why we're doing this: what's the thinking here?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+            else:
+                self._sockets.append(sock)
+                self._selector.register(sock, EVENT_WRITE)
+        except (socket.error, OSError) as e:
+            self._error = e
+            if sock is not None:
+                try:
+                    sock.close()
+                except (OSError, socket.error):
+                    pass
```
Lukasa: This block of code occurs three times: let's hoist it into a helper function.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _socket_check(self, sock):
```
Lukasa: I'd like this name to be a bit more expressive.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        """ Checks a socket to see if it is in an
+        error state, if so deal with cleaning it up
```
Lukasa: Missing full stop.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+    def _timeout_check(self):
+        """ Checks to see if the algorithm has timed out
+        and if it has, raise a timeout error. """
+        if self._has_timed_out():
+            raise socket.timeout()
+
+    def _has_timed_out(self):
+        """ Checks to see if the algorithm has timed out. """
+        if self._timeout is socket._GLOBAL_DEFAULT_TIMEOUT:
+            return False
```
Lukasa: This is a weird block; what's it doing here?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,223 @@
...
+        else:
+            return current_time() > self._start_time + self._timeout
+
+    def _cleanup(self):
+        for sock in self._sockets:
+            try:
+                sock.close()
+            except (OSError, socket.error):
+                pass
+        self._selector.close()
```
Lukasa: We should consider turning this object into a context manager and using it that way, to ensure that we always clean up all our sockets and selectors regardless of how we escape the lookup.

urllib3/util/rfc6555.py
SethMichaelLarson: Left as an exercise to the reader. ;) Kidding, I'll fix it :)

urllib3/util/rfc6555.py
SethMichaelLarson: Neither do I :P I'll think of a better one :)

urllib3/util/rfc6555.py
SethMichaelLarson: Good idea

urllib3/util/rfc6555.py
SethMichaelLarson: I thought it was backwards too, but all the tests pass with it like that and fail with it not like that and I was getting really annoyed because it's not right and shouldn't work. I'll use listcomp, I should just banish `filter` and `map` from my mind.

urllib3/util/rfc6555.py
SethMichaelLarson: You're right, I will use that. :)

urllib3/util/rfc6555.py
SethMichaelLarson: Is there a better error to raise as a default 'couldn't connect' error? I don't think it usually gets to that point.

urllib3/util/rfc6555.py
SethMichaelLarson: Yeah they are, I will move them.

urllib3/util/rfc6555.py
SethMichaelLarson: If the timeout is equal to the global default socket timeout value then we need to get that value from an actual socket after it's created so we have a usable value later on in case it's been changed.

urllib3/util/rfc6555.py
Lukasa: So we need to decide whether we believe this branch is reachable or not. If it isn't, then we should change to simply assert that `self._error` is always truthy.

If it *is* possible to reach here, then we need to understand under what circumstances that happens so get a better idea of what error to raise. However, I doubt it'll be `ECONNREFUSED`, because that implies active action on the part of the remote peer to say "No", and that probably isn't how we get to this branch.

urllib3/util/rfc6555.py
SethMichaelLarson: I had a question about this, the only two errors I'm aware of from closing something like a socket are interrupts and errors when closing an already closed socket.  Is that something we can ignore or should we still wrap closes like this?

urllib3/util/rfc6555.py
Lukasa: Ok, so I think we need to do that as early as possible and make it clear *why* we're doing it. I was going to suggest creating a temporary socket to do that but the overhead is probably a bit much. Regardless, we want to do it as early as we can and also add a bunch of comments everywhere to say why this isn't a problem and why we're doing it this way.

urllib3/util/rfc6555.py
Lukasa: We should wrap closes. There's no reason to become wrong in our assumption here and fail when we can just as easily say "oh well" and move on.

SethMichaelLarson: Okay that's the feeling I had as well. :)

urllib3/util/rfc6555.py
SethMichaelLarson: So this code isn't actually unreachable, there's at least one scenario I can think of. The call to `getaddrinfo` returns only non-IPv4 and non-IPv6 names there will be no `gaierror` and no sockets for the algorithm to try so no errors and then it'll raise this error because there's also no result.
That error should probably happen earlier, like right after the dns_results are filtered.

urllib3/util/rfc6555.py
SethMichaelLarson: Sure thing!

urllib3/util/rfc6555.py
Lukasa: Yup, and that would be inappropriate to return `ECONNREFUSED` for as well.

SethMichaelLarson:
Hello all! Been a busy few days, currently celebrating an anniversary so I will return post-weekend with a fresh set of commits. :)

SethMichaelLarson:
So some tests should definitely fail with this commit, I just want to see which tests fail on which platforms. Hopefully they're consistent with what I'm getting on Windows. :)

SethMichaelLarson:
Looks like tests are passing on Linux, having a little trouble on Mac OS... hm... will take more looks at this tomorrow as well.

SethMichaelLarson:
Looks like Mac OS, builders are timing out while connecting to localhost... I might see if I can fix this without touching any tests.

SethMichaelLarson:
This PR is ready for another round of review I believe. :)

test/with_dummyserver/test_connectionpool.py
```diff
@@ -35,6 +35,7 @@
 from urllib3.packages.six.moves.urllib.parse import urlencode
 from urllib3.util.retry import Retry, RequestHistory
 from urllib3.util.timeout import Timeout
+from urllib3.util import connection
```
Lukasa: This import is unnecessary.

urllib3/util/connection.py
```diff
@@ -1,16 +1,19 @@
 from __future__ import absolute_import
 import socket
+from urllib3.util.rfc6555 import (
+    happy_eyeballs_algorithm
+)
 from .wait import wait_for_read
 from .selectors import HAS_SELECT, SelectorError
 
+ENABLE_HAPPY_EYEBALLS = True
```
Lukasa: This doesn't seem to get used anywhere: it's just a global variable that stays True all the time. What's it for?

urllib3/util/connection.py
```diff
@@ -77,7 +86,6 @@ def create_connection(address, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
             err = e
             if sock is not None:
                 sock.close()
-                sock = None
 
     if err is not None:
```
Lukasa: So this PR seems to hurt our coverage here: we'll need to fix that at some point.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,244 @@
...
+                            errno.EAGAIN,
+                            errno.EWOULDBLOCK])
+if hasattr(errno, "WSAEWOULDBLOCK"):  # Platform-specific: Windows
+    _ASYNC_ERROR_NUMBERS.add(errno.WSAEWOULDBLOCK)
+
+_IP_FAMILIES = set([socket.AF_INET, socket.AF_INET6])
+
+
+def _safely_close_socket(sock):
+    """ Close a socket an guarantee it won't
```
Lukasa: "and"

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,244 @@
...
+        # If there are no sockets to select on, then don't try.
+        if not self._sockets:
+            return None
+
+        ready = self._selector.select(self._calculate_select_time())
+
+        for key, _ in ready:
+            sock = key.fileobj
+
+            # Successfully found a proper connected socket.
```
Lukasa: This comment doesn't seem quite right.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,244 @@
...
+
+            if self._source_address:
+                sock.bind(self._source_address)
+
+            # Set the socket to non-blocking so we can select it.
+            sock.settimeout(0.0)
+
+            error_code = sock.connect_ex(sa)
+            if error_code and error_code not in _ASYNC_ERROR_NUMBERS:
+                self._error = socket.error(error_code)
```
Lukasa: Missing some test coverage here.

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,244 @@
...
+            return current_time() > self._start_time + self._timeout
+
+    def _cleanup(self):
+        for sock in self._sockets:
+            _safely_close_socket(sock)
+        self._selector.close()
+
+    def _calculate_select_time(self):
+        if self._timeout is socket._GLOBAL_DEFAULT_TIMEOUT:
+            return 0.2
```
Lukasa: Is this branch reachable?

urllib3/util/rfc6555.py
```diff
@@ -0,0 +1,244 @@
...
+        self._source_address = source_address
+        self._socket_options = socket_options
+        self._sockets = []
+        self._selector = DefaultSelector()
+        self._start_time = None
+
+    def connect(self):
+        """ Given the setup of the algorithm, attempt to
+        use a cached value for socket family.  If no cached
+        value exists or the cached value fails follow RFC 6555
```
nateprewitt: nit: this might read a bit clearer with a comma after fails.

urllib3/util/connection.py
SethMichaelLarson: It's used in `create_connection`, I wanted to give users the ability to disable Happy Eyeballs if they wanted.  Should that be a parameter instead?

urllib3/util/connection.py
SethMichaelLarson: Why does it hurt coverage? The `sock = None` does nothing as `sock` isn't used anywhere after that so it's not needed.

urllib3/util/rfc6555.py
SethMichaelLarson: Nice catch :)

urllib3/util/rfc6555.py
SethMichaelLarson: Yeah I agree, I will make this change. :)

urllib3/util/rfc6555.py
SethMichaelLarson: I'll change the comment. :)

urllib3/util/rfc6555.py
SethMichaelLarson: Yes there are a few places missing coverage, I will take a look at coverage tonight!

SethMichaelLarson:
Thank you @Lukasa and @nateprewitt for the review! :) I'm glad that it's starting to really take shape.

Regarding @ashfall's comment: I have no reason to doubt that transforming this code to be more of a state machine would be beneficial for functionality, perhaps even readability.  The current approach IMO has good readability for how cyclic this algorithm is.  As long as I have reviewers willing to help me with this implementation and everyone else believes it is the way to go then I will do the work. :)

Also which IRC is this that I'm missing out on? :) If it would serve this project well that I participate there as well I would like to join!

urllib3/util/connection.py
Lukasa: If they need to disable happy eyeballs we have failed miserably. Regardless, we should make it possible. Let's make it private (`_ENABLE_HAPPY_EYEBALLS`) and put a comment above it indicating why it's there.

Lukasa:
@SethMichaelLarson Heh, @ashfall and I sit in several IRC channels together due to our work on Twisted and other things. In this instance we were just having an informal chat in the context of whether Twisted's HappyEyeballs implementation should move to more of an explicit state machine.

urllib3/util/connection.py
SethMichaelLarson: Oooh! Sorry Github review comments only show the line your comment was on, I will look at this. :) Thanks!

SethMichaelLarson:
@Lukasa Ah okay fair enough. :) Thanks for copying this info over here too :)

ashfall:
@SethMichaelLarson FWIW, this discussion took place in `#twisted-dev` on Freenode. You are very welcome to join us! :)

As for the state machine, this is the ticket, for context: https://twistedmatrix.com/trac/ticket/6698

SethMichaelLarson:
I added all the comments from above, now I'm going to work on full coverage.

dsully:
+1 for a Happy Eyeballs implementation.

SethMichaelLarson:
I'm pretty sure but just checking, this PR will still be useful in v2 @Lukasa ? Do you have any suggestions on how to handle Mac OS builders apparently timing out when connecting to localhost? Perhaps we should disable happy eyeballs if the timeout value is extremely small?

Lukasa:
This PR will absolutely be useful in v2, so we should keep focusing on it.

I have no real idea of what to do for the Mac thing. I'd rather keep running the tests on the Mac platform if we can.

SethMichaelLarson:
I'd like to keep tests on Mac OS as well. Perhaps disable happy eyeballs when connecting to localhost?

Lukasa:
> Perhaps disable happy eyeballs when connecting to localhost unless it's specifically enabled?

I suspect this is generally a good idea anyway, so we may want to do that (connections to localhost should be as good for AF_INET as AF_INET6, so who cares).

SethMichaelLarson:
So I've updated the `creation_connection` function a bit, it'll now run a `socket.getaddrinfo` first before deciding to use Happy Eyeballs or not and saves the results for future use if needed. Is this suitable for detecting a loopback address @Lukasa ?

SethMichaelLarson:
Latest commit seems to have decreased the actual test failures down to a single transient Mac OS failure. Definitely messed up coverage though, will have to do something about that.

SethMichaelLarson:
So the latest build all Mac OS fail due to `too many files` being open during the selectors test cases, not actually due to Happy Eyeballs tests. Are sockets somehow leaking from happy eyeballs? I'll have to check this out...

Lukasa:
@SethMichaelLarson I don't think this is HappyEyeballs related: I've seen it locally during tests for the v2 work. So I think it may be a separate issue, though I haven't dived into it yet (wasn't immediately relevant).

SethMichaelLarson:
@Lukasa Should I open a PR to make that test skip if it encounters this error?

TobiX:
Sorry for digging up this old pull request: I was looking into getting Happy Eyeballs support into urllib3 and found this (stalled?) pull request. It seems this was already very far along in getting the feature in... Why was it closed? Did it just get stale and forgotten? If no one objects, I'd like to revive this branch and move this feature forward...

sethmlarson:
Stalled due to v2, you're welcome to give it another swing! :)

PR #1002, opened 2016-10-20
Add support for OS X.

Lukasa:
This is based on the work done for PyCA cryptography.


SethMichaelLarson:
Never understood why still to this day it's impossible to test Python on Mac OS with Travis.

Lukasa:
It's not impossible, but it requires much more manual work. Still, this isn't too bad.

SethMichaelLarson:
Oh definitely not impossible, but not as easy as using Linux.

Lukasa:
Sure, but just you wait until you want some Windows.

SethMichaelLarson:
I've actually had good experiences with AppyVeyor :P

Lukasa:
Sure, it's fine, but it's not *less* work. =D

SethMichaelLarson:
Yeah it's really not haha, gotta write your own Python installer script if you want any versions besides 2.7.9 and 3.5 I believe.

SethMichaelLarson:
Looks like PyPy is failing a single SSL test, so close!

Lukasa:
Worse, it's failing a single SSL test *on OS X*. This is almost certainly because of `c_rehash`, which I *hate*.

SethMichaelLarson:
Let the hate flow through you! :) You got this

Lukasa:
Hrm, that doesn't seem to be it, the newest OpenSSL doesn't have a different rehash policy.

SethMichaelLarson:
Is it something to do with `DEFAULT_CA_DIR`? I know the CA stuff is tougher to figure out in Mac OS/Windows. Sorry I can't be of more help in this.

Lukasa:
Should be: the cert validation is *failing*, which is usually the result of failing to find the right cert, and having DEFAULT_CA_DIR set wouldn't lead to that.

Lukasa:
But as a good first step, I can reproduce this failure locally. So: progress!

SethMichaelLarson:
Huzzah to that! That's great news!

Lukasa:
Ok, this error seems to be the result of PyPy being linked against the ancient OpenSSL in macOS. For whatever reason, that OpenSSL doesn't seem to parse the cadir properly. Given that I can reproduce the problem using the command-line version of openssl on my own mac, I'm just going to throw my hands in the air for now and try to avoid getting blocked behind that.

Let's remove the PyPy support for now. We can consider adding it again later.

_travis/install.sh
```diff
@@ -0,0 +1,57 @@
...
+    pip install --user virtualenv
+else
+    pip install virtualenv
+fi
+
+pip install tox
+
+if [[ "${TOXENV}" == "gae" && ! -d ${GAE_PYTHONPATH} ]]; then
+  python _travis/fetch_gae_sdk.py ;
+fi
```
haikuginger: No \n makes me sad. 😜

SethMichaelLarson:
@Lukasa Okay, sounds like a good way to go for now.

_travis/install.sh
```diff
@@ -0,0 +1,57 @@
...
+            ;;
+        py34)
+            pyenv install 3.4.5
+            pyenv global 3.4.5
+            ;;
+        py35)
+            pyenv install 3.5.2
+            pyenv global 3.5.2
+            ;;
+        pypy*)
```
haikuginger: Remove this if we're not testing macOS with PyPy?

SethMichaelLarson: We may in the future though, there's just an issue with PyPy's OpenSSL linking.

Lukasa: Yeah, on balance I'm kinda inclined to leave it.

haikuginger: 👍

SethMichaelLarson:
Of course the build fails randomly right after passing. :P Can something be done about that Address already in use error? I've seen it a few times in automated builds randomly.

Lukasa:
We covered it in most places, but it has clearly not totally gone away. These are usually timing issues.

SethMichaelLarson:
@Lukasa Could you restart the py27 build now, then it'll be done before the Mac OS builds :P

Lukasa:
Alright, cool, that's OS X. Next up: coverage reports.

PR #1008, opened 2016-10-25
Add NameResolutionError for failed DNS lookups

### LINKED ISSUE #1003: DNS lookup error exception

```
```

##

techtonik:
Fixes #1003 for Windows. Test case for:

```
from urllib3.util.connection import create_connection
create_connection(('wowsucherror', 80))
```

On Windows, it gives:

```
  File "test.py", line 3, in <module>
    create_connection(('wowsucherror', 80))
  File "E:\urllib3\urllib3\util\connection.py", line 81, in create_connection
    raise NameResolutionError(host, e)
urllib3.exceptions.NameResolutionError: address 'wowsucherror' not found ([Errno 11001] getaddrinfo failed)
```

### DIFF
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -31,6 +31,7 @@ except NameError:  # Python 2:
 from .exceptions import (
     NewConnectionError,
     ConnectTimeoutError,
+    NameLookupError,
     SubjectAltNameWarning,
     SystemTimeWarning,
 )
@@ -124,6 +125,9 @@ class HTTPConnection(_HTTPConnection, object):
     def _new_conn(self):
         """ Establish a socket connection and set nodelay settings on it.
 
+        :raises ConnectTimeoutError:
+        :raises NewConnectionError:
+        :raises NameLookupError:
         :return: New socket connection.
         """
         extra_kw = {}
@@ -145,6 +149,9 @@ class HTTPConnection(_HTTPConnection, object):
         except SocketError as e:
             raise NewConnectionError(
                 self, "Failed to establish a new connection: %s" % e)
+
+        except socket.gaierror as e:
+            raise NameLookupError(self.host, e)
 
         return conn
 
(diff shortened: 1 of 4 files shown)


Lukasa:
Thanks, this looks great! Can we please get a test for this? I'd like us to make sure we keep track of the regression steps here.

techtonik:
This breaks tests. I don't know - probably the change is already covered and is not backward compatible.

Lukasa:
I think the specific issue is that urlopen previously caught this error, and now does not. Looks like the new exception would need to be added [here](https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L632).

techtonik:
Why repeat name resolution? I believe it is a long process, so I'd bail out immediately if name is not found.

Lukasa:
For the same reason as you'd repeat anything else: because transient network errors happen all the time.

In particular, urllib3 uses an exponential backoff system whereby each retry is delayed by a multiple of the previous. That means that transient network errors can, in cases where long delays are tolerated, be ridden out.

Either way, our retry logic needs to handle all errors in connection setup, so that block still has to catch this error.

techtonik:
@Lukasa `util.connection.create_connection` now throws new exception in addition to undocumented `SocketTimeout` and `SocketError`.

1. Should the exception handling be moved to `connection._new_conn`?
2. Should `urllib3` version be bumped for new API?

Lukasa:
Yeah, try moving that exception handling and see if it improves things.

Don't worry about urllib3 version bumps, that's part of the release process so I'll do it when we next ship a release.

techtonik:
@Lukasa I am failing to debug why `localhost` can not be found https://travis-ci.org/shazow/urllib3/jobs/174207669#L257 Do you have an IDE setup to see what's going on there?

Lukasa:
My best guess is that we're attempting to bind an IPv6  address but without IPv6 kernel module support. My guess is that a non `socket.gaierror` exception fired there. Try changing the code to print out all exceptions from `getaddrinfo` and then reraise: that should help isolate the problem.

jonparrott:

PR #1010, opened 2016-10-27
Fix PyOpenSSL with OpenSSL 1.1.0 (and CVE-2016-9015)

Lukasa:
This PR provides a number of fixes to the PyOpenSSL contrib module to allow that module to work with OpenSSL 1.1.0.

While working on this PR I stumbled across a security vulnerability that was introduced in urllib3 1.17 (by me 😥). Happily, this vulnerability affects only those using the PyOpenSSL contrib module with OpenSSL 1.1.0. Unhappily, it's still a real vulnerability. This fix was backported to 1.18 and shipped in 1.18.1 as commit c32cdbc16a9634fa0f8c829d1270301570158715. It is present in this PR as well but in a more comprehensive form along with a number of other testing fixes. The vulnerability has been given the identifier CVE-2016-9015. A lengthy explanation of the vulnerability can be found on the [oss-security mailing list](http://www.openwall.com/lists/oss-security/2016/10/27/6).

I've coordinated with our downstream repackagers, who are not affected by this vulnerability, and with a representative of the Requests team (in this case, myself), who are also not affected by this vulnerability: all of those constituencies are using urllib3 1.16. I have also coordinated with the OpenSSL team.

This patch requires a few more things before it can be merged into the master branch. Specifically, it requires enhancement of our Travis environment to test with OpenSSL 1.1.0. Probably this is most easily done on the OS X builders, which can have Homebrew install the newer OpenSSL version so we can link PyOpenSSL against it.

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -471,16 +471,22 @@ class TestHTTPS_TLSv1(HTTPSDummyServerTestCase):
         self._pool = HTTPSConnectionPool(self.host, self.port)
 
     def test_set_ssl_version_to_sslv3(self):
-        self._pool.ssl_version = ssl.PROTOCOL_SSLv3
-        self.assertRaises(SSLError, self._pool.request, 'GET', '/')
+        self.assertEqual(
+            ssl.PROTOCOL_SSLv3,
+            util.ssl_.resolve_ssl_version(ssl.PROTOCOL_SSLv3)
+        )
 
     def test_ssl_version_as_string(self):
-        self._pool.ssl_version = 'PROTOCOL_SSLv3'
-        self.assertRaises(SSLError, self._pool.request, 'GET', '/')
+        self.assertEqual(
+            ssl.PROTOCOL_SSLv3,
+            util.ssl_.resolve_ssl_version("PROTOCOL_SSLv3")
+        )
 
     def test_ssl_version_as_short_string(self):
-        self._pool.ssl_version = 'SSLv3'
-        self.assertRaises(SSLError, self._pool.request, 'GET', '/')
+        self.assertEqual(
+            ssl.PROTOCOL_SSLv3,
+            util.ssl_.resolve_ssl_version("SSLv3")
+        )
 
     def test_discards_connection_on_sslerror(self):
         self._pool.cert_reqs = 'CERT_REQUIRED'
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -375,7 +375,8 @@ class PyOpenSSLContext(object):
     @verify_mode.setter
     def verify_mode(self, value):
         self._ctx.set_verify(
-            _stdlib_to_openssl_verify[value], _verify_callback
+            _stdlib_to_openssl_verify[value],
+            _verify_callback
         )
 
     def set_default_verify_paths(self):
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -88,12 +88,15 @@ try:
 except AttributeError:
     pass
 
-_openssl_verify = {
+_stdlib_to_openssl_verify = {
     ssl.CERT_NONE: OpenSSL.SSL.VERIFY_NONE,
     ssl.CERT_OPTIONAL: OpenSSL.SSL.VERIFY_PEER,
     ssl.CERT_REQUIRED:
         OpenSSL.SSL.VERIFY_PEER + OpenSSL.SSL.VERIFY_FAIL_IF_NO_PEER_CERT,
 }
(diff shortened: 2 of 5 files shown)


urllib3/contrib/pyopenssl.py
```diff
@@ -88,12 +88,15 @@
...
-_openssl_verify = {
+_stdlib_to_openssl_verify = {
     ssl.CERT_NONE: OpenSSL.SSL.VERIFY_NONE,
     ssl.CERT_OPTIONAL: OpenSSL.SSL.VERIFY_PEER,
     ssl.CERT_REQUIRED:
         OpenSSL.SSL.VERIFY_PEER + OpenSSL.SSL.VERIFY_FAIL_IF_NO_PEER_CERT,
 }
+_openssl_to_stdlib_verify = dict(
+    (v, k) for k, v in _stdlib_to_openssl_verify.items()
+)
```
haikuginger: Any particular reason to write it this way as opposed to `{v: k for k, v in...}`? Both are perfectly valid, but one is a bit more clear.

urllib3/contrib/pyopenssl.py
```diff
@@ -367,11 +370,13 @@ def options(self, value):
...
     @property
     def verify_mode(self):
-        return self._ctx.get_verify_mode()
+        return _openssl_to_stdlib_verify[self._ctx.get_verify_mode()]
 
     @verify_mode.setter
     def verify_mode(self, value):
-        self._ctx.set_verify(value, _verify_callback)
+        self._ctx.set_verify(
+            _stdlib_to_openssl_verify[value], _verify_callback
```
haikuginger: Line break between the variables, please.

SethMichaelLarson:
We could also have it tested on Linux by having a single Linux tester use `sudo: true` and installing the latest OpenSSL if you wanted to be double rigorous. :)

test/with_dummyserver/test_https.py
```diff
@@ -471,16 +471,22 @@ def setUp(self):
...
+            util.ssl_.resolve_ssl_version("PROTOCOL_SSLv3")
+        )
 
     def test_ssl_version_as_short_string(self):
-        self._pool.ssl_version = 'SSLv3'
-        self.assertRaises(SSLError, self._pool.request, 'GET', '/')
+        self.assertEqual(
+            ssl.PROTOCOL_SSLv3,
+            util.ssl_.resolve_ssl_version("SSLv3")
+        )
```
haikuginger: Do we want to completely get rid of the existing tests? It seems as though they're testing something different here that might be useful to keep around.

urllib3/contrib/pyopenssl.py
Lukasa: Dict comprehensions were added in 2.7, and we support 2.6.

test/with_dummyserver/test_https.py
Lukasa: As best as I could see, this is actually the only bit of functionality that the tests were executing.

As to "do we want to get rid of it", I think we have to: they rely on the notion that SSLv3 is actually present in the OpenSSL on the system, and that's frequently untrue.

urllib3/contrib/pyopenssl.py
haikuginger: That'd do it! I knew that in the back of my head somewhere, but it slipped away - thanks for the reminder!

urllib3/contrib/pyopenssl.py
Lukasa: done.

test/with_dummyserver/test_https.py
haikuginger: Okay, sounds good!

haikuginger:
All right; I'm 👍 to go ahead and merge.

Lukasa:
Cool, I'll wait for the builds to complete.

t-8ch:
We should probably also have some tests for this.

Lukasa:
We do. =) We run all the HTTPS tests under PyOpenSSL as well as the stdlib. We just don't have coverage for the contrib module at this time.

If you're willing to confirm that we have 100% test coverage and write the tests we need, I'd be happy to enable coverage for the contrib module.

t-8ch:
Hm, I can't follow.
Ignoring the coverage the test should have could this issue, no?

Lukasa:
@t-8ch The tests did catch the issue. You should read my [entire writeup](http://www.openwall.com/lists/oss-security/2016/10/27/6).

t-8ch:
@Lukasa Oh, nice.

*pats testsuite on the back and offers it some sugar*

test/with_dummyserver/test_https.py
t-8ch: This nevertheless breaks for me as `AttributeError: module 'ssl' has no attribute 'PROTOCOL_SSLv3'`

PR #1013, opened 2016-10-27
Introduce IPv6 zone identifier support.

cfs-pure:
In addition to stripping IPv6 address literals in connection pools, process IPv6 address literals which might have URI encoded zone identifiers specified in RFC-6874.

See also kennethreitz/requests#1985

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -49,6 +49,16 @@ class TestConnectionPool(unittest.TestCase):
             ('http://google.com/', 'http://google.com:80/abracadabra'),
             ('https://google.com:443/', 'https://google.com/abracadabra'),
             ('https://google.com/', 'https://google.com:443/abracadabra'),
+            ('http://[2607:f8b0:4005:805::200e%25eth0]/',
+             'http://[2607:f8b0:4005:805::200e%eth0]/'
+            ),
+            ('https://[2607:f8b0:4005:805::200e%25eth0]:443/',
+             'https://[2607:f8b0:4005:805::200e%eth0]:443/'
+            ),
+            ('http://[::1]/', 'http://[::1]'),
+            ('http://[2001:558:fc00:200:f816:3eff:fef9:b954%lo]/',
+             'http://[2001:558:fc00:200:f816:3eff:fef9:b954%25lo]'
+            ),
         ]
 
         for a, b in same_host:
@@ -70,6 +80,9 @@ class TestConnectionPool(unittest.TestCase):
             ('https://google.com:80', 'http://google.com'),
             ('https://google.com:443', 'http://google.com'),
             ('http://google.com:80', 'https://google.com'),
+            # Zone identifiers are unique connection end points and should
+            # never be equivalent.
+            ('http://[dead::beef]', 'https://[dead::beef%en5]/'),
         ]
 
         for a, b in not_same_host:
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -67,13 +67,7 @@ class ConnectionPool(object):
         if not host:
             raise LocationValueError("No host specified.")
 
-        # httplib doesn't like it when we include brackets in ipv6 addresses
-        # Specifically, if we include brackets but also pass the port then
-        # httplib crazily doubles up the square brackets on the Host header.
-        # Instead, we need to make sure we never pass ``None`` as the port.
-        # However, for backward compatibility reasons we can't actually
-        # *assert* that.
-        self.host = host.strip('[]')
+        self.host = _ipv6_host(host)
         self.port = port
 
     def __str__(self):
@@ -437,6 +431,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         # TODO: Add optional support for socket.gethostbyname checking.
         scheme, host, port = get_host(url)
+
+        host = _ipv6_host(host)
 
         # Use explicit default port for comparison when none is given
         if self.port and not port:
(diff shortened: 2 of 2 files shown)


SethMichaelLarson:
There should definitely be at least one test for this.  Does the `%25` never occur when not within `[]` braces, or can it occur when not in the braces as well?

cfs-pure:
I agree, I'm working on some useful test cases (I'm thinking minimally a test_same_host case that uses 'http://[2607:f8b0:4005:805::200e%eth0]/' and 'http://[2607:f8b0:4005:805::200e%25eth0]/')

Regarding %25 outside of square brackets:  I was being somewhat conservative. If a consumer isn't processing host argument by stripping the square brackets, then I think it is safe to assume it didn't attempt to URI decode the host argument.  Certainly, I'm not against host.replace() being done blindly if that is not the case.

cfs-pure:
`HTTPConnectionPool.is_same_host()` does not do the same processing on host that `ConnectionPool.__init__()` does, so I broke out that IPv6 massaging into its own method.

urllib3/connectionpool.py
```diff
@@ -67,14 +67,28 @@ def __init__(self, host, port=None):
         if not host:
             raise LocationValueError("No host specified.")
 
+        if host[0] == '[' and host[-1] == ']':
+            host = host.replace('%25', '%').strip('[]')
```
Lukasa: This stanza seems to execute twice, because `_ipv6_host` includes it as well.

urllib3/connectionpool.py
```diff
@@ -438,6 +452,8 @@ def is_same_host(self, url):
         # TODO: Add optional support for socket.gethostbyname checking.
         scheme, host, port = get_host(url)
 
+        host = self._ipv6_host(host)
```
Lukasa: Good spot! We *definitely* need a test that exercises this change.

cfs-pure: No problem on the first request, the second one has a minimal test now (see test/test_connectionpool.py)

cfs-pure: Updates pushed to my fork.

test/test_connectionpool.py
```diff
@@ -49,6 +49,7 @@ def test_same_host(self):
             ('http://google.com/', 'http://google.com:80/abracadabra'),
             ('https://google.com:443/', 'https://google.com/abracadabra'),
             ('https://google.com/', 'https://google.com:443/abracadabra'),
+            ( 'http://[2607:f8b0:4005:805::200e%25eth0]/', 'http://[2607:f8b0:4005:805::200e%eth0]/'),
```
Lukasa: Can we extend this test a fair bit? Let's add a few other IPv6 address formats to confirm that they are all processed sensibly.

urllib3/connectionpool.py
```diff
@@ -67,14 +67,26 @@ def __init__(self, host, port=None):
...
-        # *assert* that.
-        self.host = host.strip('[]')
-        self.port = port
+        # *assert* that.  See http://bugs.python.org/issue28539
+        #
+        # Also if an IPv6 address literal has a zone identifier, the
+        # percent sign might be URIencoded, convert it back into ASCII
+        if host[0] == '[' and host[-1] == ']':
+            host = host.replace('%25', '%').strip('[]')
+        return host
```
Lukasa: Argh, sorry, should have noticed this earlier, but this method doesn't actually use `self` anywhere. That's usually a good sign that this is a function pretending to be a method, so we should hoist it off the class.

test/test_connectionpool.py
cfs-pure: Added a few more cases and a negative case (which compares that a IPv6 address literal with the same destination address is _different_ from one with a zone identifier).

cfs-pure:
Yay flakey unit tests!

briggr1:
Wondering (hoping) if these ipv6 changes will be included in the urllib3 that is embedded in requests?

SethMichaelLarson:
@briggr1 That would take at least until the next release of urllib3 and then the next release of requests. @Lukasa usually gives that a time frame of about 3 months.

Lukasa:
Sorry folks, I lost track of this issue. @cfs-pure can you bring it back up to date with the current master?

briggr1:
Giving this issue a friendly poke again ...

cfs-pure:
I've resolved the conflict, but the a unit test unrelated to my changes seems to have failed.

nateprewitt:
@cfs-pure, I believe the issue you're encountering is because the `host` variable was previously being lower cased with `lower()`. You haven't included that same functionality in `_ipv6_host` which is why the test is failing.

cfs-pure:
Pushed a fixup to lower case the host after it's dry cleaned.

cfs-pure:
@Lukasa this change doesn't remove the zone identifier, it just partially URI decodes it.  You don't want this done if the host arguments don't contain square brackets because you don't know if it is URI encoded or not.  Either way, the zone identifier gets passed down to the HTTPConnection class and a socket can be created.

Lukasa:
Yes, sorry, I didn't mean remove. ;)

So yeah, I am aware of the limitation: I just wanted to make sure we all agree it's an acceptable one.

Lukasa:
Thanks @cfs-pure!

briggr1:
thank you folks!  looking forward to using this once it makes it into requests/urllib3.

We have lot's of current needs for IPv6 using requests, and as time goes on and IPv4 address space runs out, more and more folks will, too.

briggr1:
Just wanted to say thank you to all involved for getting this change in and completed.  I picked up 2.17.3 today, and IPv6 requests using my zone identifier is working!

first, test from curl using my primary interface, en016780032:

[root@rbriggs-rhel72]# curl https://fe80:0:0:0:2b2:198d:d39e:441c%eno16780032 -k
<html><head><meta http-equiv="refresh" content="0;URL='/rest/v1'" /></head><a href="/rest/v1">/rest/v1</a></html>

Now, try the same thing from requests:

>>> resp = re.get('https://[fe80:0:0:0:2b2:198d:d39e:441c%eno16780032]', verify=False)
>>> resp
<Response [200]>
>>> resp.content
'<html><head><meta http-equiv="refresh" content="0;URL=\'/rest/v1\'" /></head><a href="/rest/v1">/rest/v1</a></html>\n'

Thank you all again!

PR #1016, opened 2016-10-30
The ``connection_from_*`` methods now accept pool_kwargs

jeremycline:
This adds the ability to override the connection pool keyword arguments
used when constructing new pools in the `connection_from_*` methods. A
new keyword argument, `pool_kwargs`, is accepted by methods in
`PoolManager`. This is nice because without it, users that want to
change these arguments (when, say, adjusting SSL configurations) need to
make sure they update the keyword arguments and get a new pool in a
thread-safe manner.

This is a somewhat more invasive version of the patch on https://github.com/kennethreitz/requests/issues/3633 since people (reasonably) were surprised to see the instance's `connection_pool_kw` modified. I'm still not completely pleased with this change for a few reasons:
- If I provide some override `pool_kwargs` and a new pool doesn't need to be created (the `pool_kwargs` aren't fields in the pool key), they will be ignored. That seems like it will surprise users.
- A huge number of methods grew a new kwarg that gets passed down the chain of function calls, which is a little sad.

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,14 +1,9 @@
-import functools
-from collections import namedtuple
 import sys
 
 from urllib3.poolmanager import (
-    _default_key_normalizer,
-    HTTPPoolKey,
-    HTTPSPoolKey,
+    PoolKey,
     key_fn_by_scheme,
     PoolManager,
-    SSL_KEYWORDS,
 )
 from urllib3 import connection_from_url
 from urllib3.exceptions import (
@@ -138,30 +133,9 @@ class TestPoolManager(unittest.TestCase):
         )
         self.assertTrue(
             all(
-                isinstance(key, HTTPPoolKey)
+                isinstance(key, PoolKey)
                 for key in p.pools.keys())
         )
-
-    def test_http_pool_key_extra_kwargs(self):
-        """Assert non-HTTPPoolKey fields are ignored when selecting a pool."""
-        p = PoolManager()
-        self.addCleanup(p.clear)
-        conn_pool = p.connection_from_url('http://example.com/')
-        p.connection_pool_kw['some_kwarg'] = 'that should be ignored'
-        other_conn_pool = p.connection_from_url('http://example.com/')
-
-        self.assertTrue(conn_pool is other_conn_pool)
-
-    def test_http_pool_key_https_kwargs(self):
-        """Assert HTTPSPoolKey fields are ignored when selecting a HTTP pool."""
-        p = PoolManager()
-        self.addCleanup(p.clear)
-        conn_pool = p.connection_from_url('http://example.com/')
-        for key in SSL_KEYWORDS:
-            p.connection_pool_kw[key] = 'this should be ignored'
-        other_conn_pool = p.connection_from_url('http://example.com/')
-
-        self.assertTrue(conn_pool is other_conn_pool)
 
     def test_https_pool_key_fields(self):
         """Assert the HTTPSPoolKey fields are honored when selecting a pool."""
@@ -204,19 +178,9 @@ class TestPoolManager(unittest.TestCase):
         self.assertTrue(all(pool in conn_pools for pool in dup_pools))
         self.assertTrue(
             all(
-                isinstance(key, HTTPSPoolKey)
+                isinstance(key, PoolKey)
                 for key in p.pools.keys())
         )
-
-    def test_https_pool_key_extra_kwargs(self):
-        """Assert non-HTTPSPoolKey fields are ignored when selecting a pool."""
-        p = PoolManager()
-        self.addCleanup(p.clear)
-        conn_pool = p.connection_from_url('https://example.com/')
-        p.connection_pool_kw['some_kwarg'] = 'that should be ignored'
-        other_conn_pool = p.connection_from_url('https://example.com/')
-
-        self.assertTrue(conn_pool is other_conn_pool)
 
     def test_default_pool_key_funcs_copy(self):
         """Assert each PoolManager gets a copy of ``pool_keys_by_scheme``."""
@@ -265,7 +229,7 @@ class TestPoolManager(unittest.TestCase):
 
         self.assertEqual(1, len(p.pools))
         self.assertTrue(pool is other_pool)
-        self.assertTrue(all(isinstance(key, HTTPSPoolKey) for key in p.pools.keys()))
+        self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
     def test_https_connection_from_host_case_insensitive(self):
         """Assert scheme case is ignored when getting the https key class."""
@@ -276,7 +240,7 @@ class TestPoolManager(unittest.TestCase):
 
         self.assertEqual(1, len(p.pools))
         self.assertTrue(pool is other_pool)
-        self.assertTrue(all(isinstance(key, HTTPSPoolKey) for key in p.pools.keys()))
+        self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
     def test_https_connection_from_context_case_insensitive(self):
         """Assert scheme case is ignored when getting the https key class."""
@@ -289,7 +253,7 @@ class TestPoolManager(unittest.TestCase):
 
         self.assertEqual(1, len(p.pools))
         self.assertTrue(pool is other_pool)
-        self.assertTrue(all(isinstance(key, HTTPSPoolKey) for key in p.pools.keys()))
+        self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
     def test_http_connection_from_url_case_insensitive(self):
         """Assert scheme case is ignored when pooling HTTP connections."""
@@ -299,7 +263,7 @@ class TestPoolManager(unittest.TestCase):
 
         self.assertEqual(1, len(p.pools))
         self.assertTrue(pool is other_pool)
-        self.assertTrue(all(isinstance(key, HTTPPoolKey) for key in p.pools.keys()))
+        self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
     def test_http_connection_from_host_case_insensitive(self):
         """Assert scheme case is ignored when getting the https key class."""
@@ -310,7 +274,7 @@ class TestPoolManager(unittest.TestCase):
 
         self.assertEqual(1, len(p.pools))
         self.assertTrue(pool is other_pool)
-        self.assertTrue(all(isinstance(key, HTTPPoolKey) for key in p.pools.keys()))
+        self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
     def test_http_connection_from_context_case_insensitive(self):
         """Assert scheme case is ignored when getting the https key class."""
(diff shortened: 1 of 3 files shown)


urllib3/poolmanager.py
```diff
@@ -129,22 +129,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
         """
-        Create a new :class:`ConnectionPool` based on host, port and scheme.
+        Create a new :class:`ConnectionPool` based on host, port, scheme, and
+        any additional pool keyword arguments.
 
-        This method is used to actually create the connection pools handed out
-        by :meth:`connection_from_url` and companion methods. It is intended
-        to be overridden for customization.
+        If `pool_kwargs` is provided, it overrides the instance's
+        `connection_pool_kw` variable. This method is used to actually create
```
Lukasa: I don't think we should override the instance value, I think we should *merge* it. That is, if a key is specified in both dictionaries, the one from `pool_kwargs` should dominate the one from `self.connection_pool_kw`, but it should be possible to leave some unspecified.

This may require a custom method, because we'll probably want to allow *unsetting* some of these values by setting their value to `None`.

urllib3/poolmanager.py
```diff
@@ -155,46 +158,55 @@ def clear(self):
...
         """
 
         if not host:
             raise LocationValueError("No host specified.")
 
-        request_context = self.connection_pool_kw.copy()
+        if pool_kwargs:
+            request_context = pool_kwargs.copy()
+        else:
+            request_context = self.connection_pool_kw.copy()
```
Lukasa: Same here, we probably want to merge rather than override.

urllib3/poolmanager.py
```diff
@@ -204,34 +216,41 @@ def connection_from_pool_key(self, pool_key):
...
+        need to be created for the request, the provided ``pool_kwargs`` are
+        not used.
         """
         u = parse_url(url)
-        return self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
+        return self.connection_from_host(u.host, port=u.port, scheme=u.scheme,
+                                         pool_kwargs=pool_kwargs)
 
-    def urlopen(self, method, url, redirect=True, **kw):
+    def urlopen(self, method, url, redirect=True, pool_kwargs=None, **kw):
```
Lukasa: I'm nervous about us passing this through from `urlopen`. Do we need to do this?

jeremycline: I don't believe we need it, no. I'm fine with dropping it from ``urlopen``. If we suddenly find a reason we do need it, it's easy to add later.

urllib3/poolmanager.py
jeremycline: I was on the fence about merging vs overriding and went with overriding because it seemed like a simple way to avoid the unsetting problem. Is it more of a burden on the user to make sure they have all the values they need in the override dictionary, or to make sure they correctly unset values we're merging in? I honestly don't know, so if merging is more appealing to you, I can change it to merge.

jeremycline:
Sorry for letting this languish. The tests are now passing and I think I've addressed all the initial comments!

urllib3/poolmanager.py
```diff
@@ -155,46 +156,53 @@ def clear(self):
...
+        provided, it is merged with the instance's ``connection_pool_kw``
+        variable and used to create the new connection pool, if one is
+        needed.
         """
 
         if not host:
             raise LocationValueError("No host specified.")
 
-        request_context = self.connection_pool_kw.copy()
+        request_context = self._merge_pool_kwargs(pool_kwargs)
```
Lukasa: Do we need to do a merge here? It's doubly confusing because we end up passing pool_kwargs twice, once to the request context and once to the separate `pool_kwargs` argument.

urllib3/poolmanager.py
```diff
@@ -204,22 +212,46 @@ def connection_from_pool_key(self, pool_key):
...
+        if override:
+            for key, value in override.items():
+                if value is None:
+                    try:
+                        del base_pool_kwargs[key]
+                    except KeyError:
+                        pass
+                else:
+                    base_pool_kwargs[key] = value
+        return base_pool_kwargs
```
Lukasa: Should this method come *after* `connection_from_url`, just to keep the `connection_from_*` methods in a visual block?

urllib3/poolmanager.py
jeremycline: You're right, it's very confusing. I realized that if we _do_ merge here, though, there's no reason to continue passing the pool_kwargs dictionary around. I've added a new commit that keeps this merge and stops passing the pool_kwargs around. Hopefully that's a clearer way to do this. If not I can back that commit out and mull on it some more :smile:

I regret adding ``connection_from_context`` and ``connection_from_pool_key`` since it only makes it more confusing and I don't think they add a lot of value. Oh well.

urllib3/poolmanager.py
jeremycline: I have made it so

Lukasa:
Fab, thanks for re-evaluating! I'll re-review on Monday. =D

urllib3/poolmanager.py
```diff
@@ -129,22 +129,32 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        kwargs = self.connection_pool_kw
+        if request_context is None:
+            request_context = self.connection_pool_kw.copy()
+
+        # Although the context has everything necessary to create the pool,
+        # this function has historically only used the scheme, host, and port
+        # in the positional args. When an API change is acceptable these can
+        # be removed.
+        for key in ('scheme', 'host', 'port'):
+            request_context.pop(key, None)
```
Lukasa: What are we trying to do here? Why do we need to strip these three values from the context if they're present?

urllib3/poolmanager.py
```diff
@@ -186,9 +199,9 @@ def connection_from_context(self, request_context):
         pool_key_constructor = self.key_fn_by_scheme[scheme]
         pool_key = pool_key_constructor(request_context)
 
-        return self.connection_from_pool_key(pool_key)
+        return self.connection_from_pool_key(pool_key, request_context=request_context)
```
Lukasa: Is there a good reason that the request context is not part of the pool key?

jeremycline: The pool key is a subset of values from the context at the moment. The HTTP pool key is on the following keywords from the ``request_context`` dictionary:

* scheme
* host
* port
* timeout
* retries
* strict
* block
* source_address

However, in addition to those fields the context may also have one or more key-value pairs meant for the ``HTTPConnectionPool`` class which accepts all those keyword arguments (except for scheme) and:

* maxsize
* headers
* _proxy
* _proxy_header

On top of that, any additional keywords provided to ``HTTPConnectionPool`` get handed on to ``HTTPConnection``, which accepts ``(*args, **kwargs)``, but looks for:

* source_address
* socket_options

It has just occurred to me that since we don't key on ``maxsize``, for example, I think changing this after a pool has been created won't cause a new pool with a different ``maxsize`` to be created. It's not something I would have considered doing, but I guess that would be surprising to someone. Maybe we _should_ key on the entire context.

What might make this all cleaner is to have an object that contains what ``context`` currently holds and also knows how to derive a pool key from it so that we aren't handing both the pool key and the context around.

urllib3/poolmanager.py
jeremycline: I didn't want to pass the scheme, host, or port from the context to the pool class since this function is still accepting those arguments as positional and the pool classes accept the host key as position and everything else as keyword arguments. They were [added to the context](https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L170) since the pool key uses them. It is awkward to do here, though.

urllib3/poolmanager.py
Lukasa: Yeah, to be clear, I think we should key on the entire context. I don't think there is anything in the context that shouldn't be part of the pool key.

jeremycline: Okay, that does seem like the safest approach. So we'd have something similar to what you proposed in https://github.com/shazow/urllib3/pull/751?

Lukasa: Yeah, I think so. Clearly I keep rehashing the same ideas over and over. ;)

jeremycline:
Okay, I've got it keying the pools on the entire context. I left all the old stuff there for the moment so you can still define custom key functions and the namedtuples still exist. It looks like cryptography is failing to link for Python 2.6 on OSX, though, which is why Travis is unhappy.

urllib3/poolmanager.py
```diff
@@ -40,38 +39,42 @@
...
-    context = {}
-    for key in key_class._fields:
-        context[key] = request_context.get(key)
+    # Since we mutate the dictionary, make a copy first
+    context = request_context.copy()
     context['scheme'] = context['scheme'].lower()
     context['host'] = context['host'].lower()
-    return key_class(**context)
+
+    return _frozenset_from_dictionary(context)
```
Lukasa: Why are we moving to frozenset instead of namedtuple?

jeremycline: I assumed we wanted something like #751 proposed so I went that direction. I suppose it could still be namedtuple classes, but wouldn't they have to be created dynamically to handle the various keyword arguments? Would we nest namedtuples for nested dictionary keyword arguments? I didn't feel like defining a namedtuple with a static set of fields made much since nowhere else enforces a restriction on keyword arguments you can pass to the PoolManager/ConnectionPool.

Lukasa: I mean, the restriction is implicit eventually because kwargs that are unrecognised at the pool level are passed to the connection, at which point they are either accepted or not.

However, I take your point. Hrm. I do still feel like the namedtuple is the right way to go, but the headers kwarg does rerepsent a problem.

jeremycline: Yup, it just felt weird to have a restriction on keyword arguments, then a set of APIs that took anything even though, as you said, it'll eventually run into trouble.

We could accept that oddity and have a nametuple that just handles the keywords we know to be dictionaries explicitly by using a frozenset in the tuple.

Lukasa: I think I'm ok with accepting that oddity, at least long enough to see a patch using it.

jeremycline: Okay, I poked at this a while last weekend, but I'm not very happy with a few problems I've encountered with the namedtuples.

The first problem I hit is that namedtuples [can't have field names that start with underscores](https://docs.python.org/dev/library/collections.html#collections.namedtuple) and most unfortunately, there are several keyword arguments that [start with underscore](https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L144). This means we have to [translate these special cases](https://github.com/shazow/urllib3/pull/1016/commits/e7920be73daa459caa9d183c3d0dd8212859374d#diff-0401b3595c5b19c0b003447baaaf35d0R102) (although admittedly that code could probably be made prettier).

The other problem I hit was that there are [other connection types that expect kwargs that I didn't account for](https://github.com/shazow/urllib3/blob/master/urllib3/contrib/socks.py#L64). This could be an additional thing other PoolManagers need to handle (by altering the default pool key I guess), but I think it would break anything that's already out there.

Both these problems could be worked around, but it'll add complexity. I wanted to see what you thought before I went too far down this path.

Lukasa: Arg, that's a real annoyance. To the second issue, I'm ok with requiring that other PoolManagers extend their pool key: that seems fine to me.

The other stuff is worse. Having to mutate in this way is really tricky, and I'm not entirely sure how best to handle it. Now, we can define a unique mapping in such a way that we prepend a meaningful string to the front of the kwargs, but it's pretty silly.

Still, that potentially makes a compelling argument for using frozenset instead of namedtuple. 😞  There are just a few too many nasty hacks involved in using the namedtuple.

jeremycline: Yeah. Prepending a string sounds like the safest, most reliable approach. I'll see what that looks like, but I suspect the simplicity of the frozenset approach will be more appealing than having namedtuples.

jeremycline:
Okay, I prepended 'key_' to all the pool key fields. It doesn't look _that_ bad, but it's a little odd. I think the Python 2.6 OSX build just needs a kick to pass since it passed on [my fork](https://travis-ci.org/jeremycline/urllib3/jobs/210062748).

Sorry for the long delay, I was traveling every weekend for the past month :(

urllib3/poolmanager.py
```diff
@@ -39,10 +39,38 @@
     'HTTPSPoolKey', HTTPPoolKey._fields + SSL_KEYWORDS
 )
```
Lukasa: Are we using these any more?

jeremycline: Nope. I was going to leave them because I thought removing them would be API-breaking since they've been in a release, but if you're comfortable with them gone I'm happy to yank them out.

jeremycline:
> But I'd like to get some other eyes on it because tbh I think I have tunnel vision on this patch at this point: I've spent so long looking at it I'm not sure whether I've just convinced myself that this is the right way to go!

Me too 😄 - more eyes are always good.

haikuginger:
I'll put eyes on it sometime this week.

urllib3/poolmanager.py
Lukasa: At the very least we need to comment them out.

urllib3/poolmanager.py
```diff
@@ -52,26 +80,43 @@ def _default_key_normalizer(key_class, request_context):
...
+    # These are both dictionaries and need to be transformed into frozensets
+    for key in ('headers', '_proxy_headers', '_socks_options'):
+        if key in context and context[key] is not None:
+            context[key] = frozenset(context[key].items())
+
+    # Map the kwargs to the names in the namedtuple - this is necessary since
+    # namedtuples can't have fields starting with '_'.
+    for key in list(context.keys()):
+        context['key_' + key] = context[key]
+        context.pop(key)
```
nateprewitt: Since we're using `pop`, is there a reason not to do `content['key_' + key] = context.pop(key)`?

urllib3/poolmanager.py
```diff
@@ -129,22 +174,32 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    def _new_pool(self, scheme, host, port, request_context=None):
         """
-        Create a new :class:`ConnectionPool` based on host, port and scheme.
+        Create a new :class:`ConnectionPool` based on host, port, scheme, and
+        any additional pool keyword arguments.
 
-        This method is used to actually create the connection pools handed out
-        by :meth:`connection_from_url` and companion methods. It is intended
-        to be overridden for customization.
+        If `request_context` is provided, it is provided as keyword arguments
```
nateprewitt: Nit: We seem to be using double ticks (\`\`) in all the other docstrings in this file. May want to move \`request_context\` inline with that.

test/test_poolmanager.py
```diff
@@ -323,20 +287,84 @@ def test_http_connection_from_context_case_insensitive(self):
...
+            'http://example.com', pool_kwargs={'key': 'other'})
+        pool3 = p.connection_from_url(
+            'http://example.com', pool_kwargs={'key': 'value', 'x': 'y'})
 
         self.assertEqual(2, len(p.pools))
+        self.assertTrue(pool1 is pool3)
+        self.assertFalse(pool1 is pool2)
+
+    def test_override_pool_kwargs_url(self):
+        """Assert overriding pool kwargs works with connection_from_url"""
```
nateprewitt: This one's another silly nit, but all of the other docstrings end with periods.

Lukasa:
@nateprewitt, just a quick note: mind doing a re-review when you have some time to make sure your concerns were appropriately addressed?

Lukasa:
Ok, this looks reasonable to me then. :sparkles: Let's do it.

jeremycline:
@Lukasa Sure. Do you want me to squash the commits down as well, or do you prefer keeping them all as-is?

Lukasa:
Let's squash them down, why not.

jeremycline:
Okay, they're all squashed down and the changelog is updated.

Lukasa:
Nicely done @jeremycline!

jeremycline:
Thanks! I couldn't have done it without all the reviews and feedback, so many thanks to everyone for those.

mbohlool:
is `assert_hostname` (`key_assert_hostname`) a valid key to pass to poolmanager? It is not part of that list and our client is failing because of that.

PR #1017, opened 2016-10-31
Fix empty filenames in multipart headers

### LINKED ISSUE #1015: Multipart request headers do not work properly for values of empty string
Continuing the discussion from https://github.com/sigmavirus24/requests-toolbelt/issues/162, attempting to create a `RequestField` which is then made multipart via `make_multipart` does not work properly if the filename given is an empty string.

urllib3 test code:

```
from urllib3.fields import RequestField
field = RequestField(name="somename", data="somedata", filename="")
field.make_multipart(content_type="application/octet-stream")
print(field.headers)
```

Expected output:

```
{'Content-Type': 'application/octet-stream', 'Content-Location': None, 'Content-Disposition': 'form-data; name="somename"; filename=""'}
```

Actual output:

```
{'Content-Type': 'application/octet-stream', 'Content-Location': None, 'Content-Disposition': 'form-data; name="somename"'}
```
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38799361-multipart-request-headers-do-not-work-properly-for-values-of-empty-string?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

wchill:
Fixes #1015.

I'm not too familiar with using Tox for testing, and I got one error when running the test suite, but it was in some unrelated code (SSL) so I'll just submit this for review for the time being.

### DIFF
--- a/test/test_fields.py
+++ b/test/test_fields.py
@@ -36,6 +36,15 @@ class TestRequestField(unittest.TestCase):
             'Content-Location: /test\r\n'
             '\r\n')
 
+    def test_make_multipart_empty_filename(self):
+        field = RequestField('somename', 'data', '')
+        field.make_multipart(content_type='application/octet-stream')
+        self.assertEqual(
+            field.render_headers(),
+            'Content-Disposition: form-data; name="somename"; filename=""\r\n'
+            'Content-Type: application/octet-stream\r\n'
+            '\r\n')
+
     def test_render_parts(self):
         field = RequestField('somename', 'data')
         parts = field._render_parts({'name': 'value', 'filename': 'value'})
--- a/urllib3/fields.py
+++ b/urllib3/fields.py
@@ -130,7 +130,7 @@ class RequestField(object):
             iterable = header_parts.items()
 
         for name, value in iterable:
-            if value:
+            if value is not None:
                 parts.append(self._render_part(name, value))
 
         return '; '.join(parts)


test/test_fields.py
```diff
@@ -36,6 +36,14 @@ def test_make_multipart(self):
...
             '\r\n')
 
+        field = RequestField('somename', 'data', '')
+        field.make_multipart(content_type='application/octet-stream')
+        self.assertEqual(
+            field.render_headers(),
+            'Content-Disposition: form-data; name="somename"; filename=""\r\n'
+            'Content-Type: application/octet-stream\r\n'
+            '\r\n')
+
```
Lukasa: Thanks for writing this test! Can you split it into its own test function please?

urllib3/fields.py
```diff
@@ -130,7 +130,7 @@ def _render_parts(self, header_parts):
             iterable = header_parts.items()
 
         for name, value in iterable:
-            if value:
+            if value or value == '':
```
Lukasa: I think presumably what we want to do is to exclude `None` as a value. So let's change this check to `is not None` instead.

wchill: I think it's more accurate if it's not a falsy value in general (other than empty string), since dealing with an empty list, set, etc. doesn't really make sense.

Lukasa: The problem with that logic is that you need to handle *at least* the two empty string types (`b''` and `u''`). More generally, I think the current behaviour (silent non-emission of data for falsy non-strings) is worse than the new failure mode we'd introduce (exceptions thrown for falsy non-strings). We don't accept *non-empty* lists or sets, so I don't see why we should tolerate *empty* lists or sets.

wchill: All right, works for me.

wchill:
Don't merge yet, I messed something up with my rebase.

test/test_fields.py
wchill: This is done

Lukasa:
Excellent, this looks great, thanks so much @wchill! :custard:

PR #1021, opened 2016-11-01
Cause a Travis failure if there are warnings in the docs

alexwlchan:
I’ve tried to set up a docs job on Travis that will fail if there are problems in the docs build.

There weren’t any errors already (hooray!) so I’ve introduced a deliberate consistency bug. Hopefully the build will fail!

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -43,7 +43,7 @@ basepython = python2.7
 deps=
     -r{toxinidir}/docs/requirements.txt
 commands=
-    rm -r {toxinidir}/docs/_build
+    rm -rf {toxinidir}/docs/_build
     make -C {toxinidir}/docs html
 whitelist_externals=
     make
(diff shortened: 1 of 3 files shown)


alexwlchan:
It failed! =D

After a week where CI has been breaking for me consistently even though I don’t want it to, it’s somewhat gratifying to know that I can also break the build when I really mean to.

Can I get a review from somebody please? Then I’ll squash and throw away the deliberate docs bug.

tox.ini
```diff
@@ -43,7 +43,7 @@ basepython = python2.7
 deps=
     -r{toxinidir}/docs/requirements.txt
 commands=
-    rm -r {toxinidir}/docs/_build
+    rm -rf {toxinidir}/docs/_build
```
alexwlchan: So it turns out this works as long as you’ve ever done a docs build before.

If you checkout the codebase and run this command, it errors trying to clean up a directory that it hasn’t been able to create yet.

alexwlchan:
Removed the bug and rebased, I have.

PR #1025, opened 2016-11-04
Correct the import of urljoin for Python 3

quite:
At least in Python 3.5.2, this caused cascading breakage down to requests-toolbelt,
used by vdirsyncer. On Py3, urljoin is at urllib.parse.

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -42,7 +42,7 @@ from __future__ import absolute_import
 import logging
 import os
 import warnings
-from urlparse import urljoin
+from ..packages.six.moves.urllib.parse import urljoin
 
 from ..exceptions import (
     HTTPError,


Lukasa:
To be clear, this problem occurred for you when working with AppEngine, right?

quite:
Actually, only trying to install latest vdirsyncer...

Lukasa:
So this module should only be imported on AppEngine or for use with AppEngine. I think we should still take this fix, but vdirsyncer probably shouldn't be unconditionally using this contrib module.

untitaker:
I'm fairly certain I don't import that module, please wait a few hours until I can investigate

Lukasa:
@untitaker It's entirely possible that it's *transitively* imported via the toolbelt.

Lukasa:
Thanks for this patch! :sparkles: :cake: :sparkles:

untitaker:
I use FingerprintAdapter from toolbelt, but I can't see how it would import the appengine stuff from that.

PR #1028, opened 2016-11-05
Update RECENT_DATE to 2016

### LINKED ISSUE #1005: Add unit test for RECENT_DATE
I've got a feeling this value could be easily forgotten to be updated over time. Could we perhaps add a unit test that will fail if it gets to be too far away from the current date.  It seems like it was last modified in mid 2014 but wasn't updated since then, should this value be updated now?

Link to the blame: https://github.com/shazow/urllib3/blame/master/urllib3/connection.py#L59
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/38578956-add-unit-test-for-recent_date?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F192525&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SethMichaelLarson:
This PR resolves #1005 by adding a unit test to allow `RECENT_DATE` to have a maximum of two years of difference between the current date and it's last value.  I also updated `RECENT_DATE` to January 1st, 2016 to refresh the value to a more recent date.

My rationale for these changes are:
- When `RECENT_DATE` was added it was August in 2014 so I figured jumping to 2016 made sense.
- Because this change is coming more than two years after the original `RECENT_DATE` check was added, it should be safe to assume this will not break anyone who's system times fit the original set of rules which would require not more than two-three years difference in system clock time.

The unit test will also make it harder to forget that this value should be updated, hopefully more often than once every two years but I think this is a good compromise, I'm open to suggestions if this should be more/less frequent.

### DIFF
--- a/test/test_connection.py
+++ b/test/test_connection.py
@@ -52,6 +52,8 @@ class TestConnection(unittest.TestCase):
     def test_recent_date(self):
         # This test is to make sure that the RECENT_DATE value
         # doesn't get too far behind what the current date is.
+        # When this test fails update urllib3.connection.RECENT_DATE
+        # according to the rules defined in that file.
         two_years = datetime.timedelta(days=365 * 2)
         self.assertGreater(RECENT_DATE, (datetime.datetime.today() - two_years).date())
 
--- a/test/test_connection.py
+++ b/test/test_connection.py
@@ -1,6 +1,6 @@
 import datetime
 import sys
-if sys.version_info > (2, 6):
+if sys.version_info >= (2, 7):
     import unittest
 else:
     import unittest2 as unittest
@@ -53,7 +53,7 @@ class TestConnection(unittest.TestCase):
         # This test is to make sure that the RECENT_DATE value
         # doesn't get too far behind what the current date is.
         two_years = datetime.timedelta(days=365 * 2)
-        self.assertTrue(RECENT_DATE > (datetime.datetime.today() - two_years).date())
+        self.assertGreater(RECENT_DATE, (datetime.datetime.today() - two_years).date())
 
 
 if __name__ == '__main__':
--- a/test/test_connection.py
+++ b/test/test_connection.py
@@ -1,3 +1,4 @@
+import datetime
 import unittest
 
 import mock
@@ -6,6 +7,7 @@ from urllib3.connection import (
     CertificateError,
     VerifiedHTTPSConnection,
     _match_hostname,
+    RECENT_DATE
 )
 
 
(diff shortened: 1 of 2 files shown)


urllib3/connection.py
```diff
@@ -56,7 +56,7 @@ class ConnectionError(Exception):
     'https': 443,
 }
 
-RECENT_DATE = datetime.date(2014, 1, 1)
+RECENT_DATE = datetime.date(2016, 1, 1)
```
Lukasa: Can we get a comment above here that explains our "two-year" rule? What is the latest this should be? What is the earliest it should be?

test/test_connection.py
```diff
@@ -43,6 +49,12 @@ def test_match_hostname_mismatch(self):
             )
             self.assertEqual(e._peer_cert, cert)
 
+    def test_recent_date(self):
+        # This test is to make sure that the RECENT_DATE value
+        # doesn't get too far behind what the current date is.
```
Lukasa: Let's add a comment here expressly saying what has to be done to fix this test when it fails.

test/test_connection.py
```diff
@@ -1,11 +1,17 @@
-import unittest
+import datetime
+import sys
+if sys.version_info > (2, 6):
+    import unittest
+else:
```
Lukasa: Do we need unittest2? What function from there are we using?

sigmavirus24: I have the same question as @Lukasa here. unittest2 causes far more headaches when used like this than not (in my experience).

urllib3/connection.py
SethMichaelLarson: Sure I can add a comment above with our rules. :)

test/test_connection.py
SethMichaelLarson: So the reason I added this was to use `assertGreater` which apparently wasn't in `unittest` for Python 2.6. I now realize why it didn't work on CI. I'll push a change that properly chcks `sys.version_info`.

SethMichaelLarson:
Ready again for review @lukasa :)

urllib3/connection.py
```diff
@@ -56,7 +56,12 @@ class ConnectionError(Exception):
     'https': 443,
 }
 
-RECENT_DATE = datetime.date(2014, 1, 1)
+# When updating RECENT_DATE be sure to not
+# update the value to be later than two years
+# in the past and no earlier than 6 months
+# in the past to stay consistent with the
+# history of this value.
```
Lukasa: Woah, lots of negatives in here. I think we can make this comment clearer:

```python
# When updating RECENT_DATE, move it to
# within two years of the current date, and no
# earlier than 6 months ago.
```

That should be fine. =)

SethMichaelLarson: Thanks, yeah that was my third revision of that comment before I just decided to let it be reviewed and hopefully come up with a better way of putting it. :) I'll change this!

SethMichaelLarson:
Okay @Lukasa I added your comment. :)

SethMichaelLarson:
:tada:

Issue #624, opened 2015-05-15
Twice memory usage in `encode_multipart_formdata`


Axik:
Basically, if you wants to upload file like 1GB it will cost you 2GB RAM.

```
Filename: /Users/kyrylo/.virtualenvs/drsdk/lib/python2.7/site-packages/requests/packages/urllib3/filepost.py

Line #    Mem usage    Increment   Line Contents
================================================
    59    981.3 MiB      0.0 MiB   @profile
    60                             def encode_multipart_formdata(fields, boundary=None):
    61                                 """
    62                                 Encode a dictionary of ``fields`` using the multipart/form-data MIME format.
    63                             
    64                                 :param fields:
    65                                     Dictionary of fields or list of (key, :class:`~urllib3.fields.RequestField`).
    66                             
    67                                 :param boundary:
    68                                     If not specified, then a random boundary will be generated using
    69                                     :func:`mimetools.choose_boundary`.
    70                                 """
    71    981.3 MiB      0.0 MiB       body = BytesIO()
    72    981.3 MiB      0.0 MiB       if boundary is None:
    73    981.3 MiB      0.0 MiB           boundary = choose_boundary()
    74                             
    75   1920.7 MiB    939.4 MiB       for field in iter_field_objects(fields):
    76    981.3 MiB   -939.4 MiB           body.write(b('--%s\r\n' % (boundary)))    77                             
    78    981.3 MiB      0.0 MiB           writer(body).write(field.render_headers())
    79    981.3 MiB      0.0 MiB           data = field.data
    80                             
    81    981.3 MiB      0.0 MiB           if isinstance(data, int):
    82                                         data = str(data)  # Backwards compatibility
    83                             
    84    981.3 MiB      0.0 MiB           if isinstance(data, six.text_type):
    85                                         writer(body).write(data)
    86                                     else:
    87   1920.7 MiB    939.4 MiB               body.write(data)
    88                             
    89   1920.7 MiB      0.0 MiB           body.write(b'\r\n')
    90                             
    91   1920.7 MiB      0.0 MiB       body.write(b('--%s--\r\n' % (boundary)))
    92                             
    93   1920.7 MiB      0.0 MiB       content_type = str('multipart/form-data; boundary=%s' % boundary)
    94                             
    95   2000.1 MiB     79.4 MiB       return body.getvalue(), content_type
```

P.S. using this profiler https://pypi.python.org/pypi/memory_profiler

shazow:
[...]

Issue #51, opened 2012-02-26
Support for posting big files/streaming objects


allover:
Currently, it is not possible to post a 4 GB file using urllib3 since that requires reading the entire file content into a buffer before sending. (unless I'm missing something?)
It would be nice if we could pass file-like objects (objects with a read() method or something) as post variables, which will be read from on demand and perform well on huge files.
The only python script capable of this that I'm aware of is http://atlee.ca/software/poster/, which is a urllib2 opener class.
##

shazow:
This would be nice to have. :/ (bump)

SethMichaelLarson:
This is being worked on in #1030

gerazenobi:
Hi, I was wondering if this was still an issue or not since the pull request #1030 was actually closed?

Cheers.

PR #1030, opened 2016-11-07
Streaming File Uploads

SethMichaelLarson:
Took the work that was done a long time ago on the branch `filepost-stream` and brought it up to speed with the current API for filepost and fields modules, made it Python 2 and 3 compatible, added a test for memory usage and converted the current tests to use the new filepost iterator.

More tests to exercise each possible type of file-like or streamable object that can be passed into filepost will be created after this type of structure is approved and still wanted in urllib3.

I don't know if this behavior is still desired within urllib3 or as @sigmavirus24 has mentioned using code from requests-toolbelt.

Relevant issues: #51 #624

(CI failed because of random read timeout on Mac OS + Python 3.3)

### DIFF
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -32,7 +32,7 @@ class TestMultipartEncoderGenerator(unittest.TestCase):
         data = b'x' * MEGABYTE  # 1 MB of data
         start_memory = psutil.Process().memory_info().rss
 
-        encoded, _ = encode_multipart_formdata([(b'k', data)], boundary=BOUNDARY)
+        encoded, _ = encode_multipart_formdata([('k', data)], boundary=BOUNDARY)
         end_memory = psutil.Process().memory_info().rss
         self.assertTrue(end_memory <= start_memory + MEGABYTE)
 
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -334,7 +334,7 @@ class TestStreamingUploads(unittest.TestCase):
         except StopIteration:
             self.assertEqual(i, len(expected) - 1)
 
-    def test_chunking_remaining(self):
+    def test_iter_chunking_remaining(self):
         fields = [('k', ('somefile.txt', _ReadableObject(b'v' * 1024), 'image/jpeg'))]
 
         for chunk_size in range(1, 128):
@@ -348,3 +348,15 @@ class TestStreamingUploads(unittest.TestCase):
             for chunk in encoded:
                 if chunk not in non_chunks:
                     self.assertTrue(len(chunk) == chunk_size or len(chunk) == 1024 % chunk_size)
+
+    def test_read_chunking_remaining(self):
+        for chunk_size in range(2, 128):
+            fields = [('k', ('somefile.txt', io.BytesIO(b'v' * 1024), 'image/jpeg'))]
+            encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY, chunk_size=chunk_size)
+            to_read = 1169
+
+            while to_read > 0:
+                read_size = min(chunk_size - 1, to_read)
+                chunk = encoded.read(read_size)
+                self.assertEqual(len(chunk), read_size)
+                to_read -= len(chunk)
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -1,4 +1,5 @@
 import unittest
+import psutil
 
 from urllib3.filepost import (
     encode_multipart_formdata,
@@ -25,6 +26,15 @@ class TestMultipartEncoderGenerator(unittest.TestCase):
             encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY)
             encoded = b''.join(encoded)
             self.assertEqual(len(encoded), len(MultipartEncoderGenerator(fields, boundary=BOUNDARY)))
+
+    def test_memory_usage(self):
+        MEGABYTE = 1024 * 1024
+        data = b'x' * MEGABYTE  # 1 MB of data
+        start_memory = psutil.Process().memory_info().rss
+
+        encoded, _ = encode_multipart_formdata([(b'k', data)], boundary=BOUNDARY)
+        end_memory = psutil.Process().memory_info().rss
+        self.assertTrue(end_memory <= start_memory + MEGABYTE)
 
 
 class TestIterfields(unittest.TestCase):
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -15,16 +15,15 @@ BOUNDARY = '!! test boundary !!'
 class TestMultipartEncoderGenerator(unittest.TestCase):
     def test_len(self):
         fieldsets = [
-            #[('k', 'v'), ('k2', 'v2')],
-            #[('k', b'v'), (u('k2'), b'v2')],
-            #[('k', b'v'), (u('k2'), 'v2')],
+            [('k', 'v'), ('k2', 'v2')],
+            [('k', b'v'), (u('k2'), b'v2')],
+            [('k', b'v'), (u('k2'), 'v2')],
             [('foo', b'a'), ('foo', b'b')],
         ]
 
         for fields in fieldsets:
             encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY)
             encoded = b''.join(encoded)
-
             self.assertEqual(len(encoded), len(MultipartEncoderGenerator(fields, boundary=BOUNDARY)))
 
 
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -1,6 +1,10 @@
 import unittest
 
-from urllib3.filepost import encode_multipart_formdata, iter_fields
+from urllib3.filepost import (
+    encode_multipart_formdata,
+    iter_fields,
+    MultipartEncoderGenerator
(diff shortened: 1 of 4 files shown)


Lukasa:
So @sigmavirus24 is much more familiar with the weird edge cases and problems that can arise in this area so I'd like to let him review this if possible. =)

SethMichaelLarson:
Sure thing, thanks @Lukasa ! :)

SethMichaelLarson:
For Turkey Day 🦃 here's one juicy PR with tests, 100% coverage and ready for review! @sigmavirus24 cc @Lukasa :)

test/test_filepost.py
```diff
@@ -131,3 +168,195 @@ def test_request_fields(self):
...
+        encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY, chunk_size=1)
+        non_chunks = [b'--' + b(BOUNDARY) + b'\r\n',
+                      b'Content-Disposition: form-data; name="k"; filename="somefile.txt"\r\n'
+                      b'Content-Type: image/jpeg\r\n\r\n',
+                      b'\r\n',
+                      b'--' + b(BOUNDARY) + b'--\r\n']
+
+        for chunk in encoded:
+            if chunk not in non_chunks:
+                self.assertIs(chunk, unknown_object)
```
Lukasa: I don't know that this test is well-crafted. It might be better to use equality the whole way through and place the unknown object in the right place.

urllib3/filepost.py
```diff
@@ -56,39 +53,165 @@ def iter_fields(fields):
...
+            if hasattr(data, '__len__'):
+                size += len(encode(data))
+            elif isinstance(data, six.integer_types):
+                size += len(str(data))
+            elif hasattr(data, 'seek'):
+                size += file_size(data)
+            elif hasattr(data, 'read'):
+                size += len(encode(data.read()))  # This is undesired
+            elif hasattr(data, '__iter__'):
+                size += sum(len(encode(chunk)) for chunk in data)  # This is also undesired
```
Lukasa: Do these two blocks of code work? It seems like they would move the file pointer forward and never move it back.

urllib3/filepost.py
```diff
@@ -56,39 +53,165 @@ def iter_fields(fields):
...
+            else:
+                size += len(encode(data))  # Hope for the best
+
+        empty_multipart = MultipartEncoderGenerator(empty_fields, boundary=self.boundary)
+        empty_size = sum(encode(len(chunk)) for chunk in empty_multipart)
```
Lukasa: There has *got* to be a better way to get this value than this.

urllib3/filepost.py
```diff
@@ -56,39 +53,165 @@ def iter_fields(fields):
...
+                while True:
+                    chunk = data.read(self.chunk_size)
+                    if not chunk:
+                        break
+                    yield encode(chunk)
+
+            elif hasattr(data, '__iter__'):
+                # Stream from an iterator
+                for chunk in data:
+                    yield encode(chunk)
```
Lukasa: I note that there is a lot of assumption that iterators will return text. That's probably not right. The text/bytes handling stuff should probably be factored out. It may be sensible to try to nest iterators here so that we can factor out the encode/decode logic.

urllib3/filepost.py
```diff
@@ -56,39 +53,165 @@ def iter_fields(fields):
...
+
+            else:
+                # Hope for the best
+                yield encode(data)
+
+            yield encode('\r\n')
```
Lukasa: There is exactly zero reason to do this: just use a bytes literal and save ourselves the function calls.

urllib3/filepost.py
```diff
@@ -56,39 +53,165 @@ def iter_fields(fields):
...
+                count += len(chunk)
+        except StopIteration:
+            pass
 
-        body.write(b'\r\n')
+        if count > size:
+            self.leftover = data[size:]
 
-    body.write(b('--%s--\r\n' % (boundary)))
+        return data[:size]
```
Lukasa: It might be nice if this had `str`/`unicode` methods so that it could behave like a string in circumstances where people want to use it like one. Just a thought.

test/test_filepost.py
```diff
@@ -1,15 +1,49 @@
...
+            self.assertEqual(len(encoded), len(MultipartEncoderGenerator(fields, boundary=BOUNDARY)))
+
+    def test_memory_usage(self):
+        MEGABYTE = 1024 * 1024
+        data = b'x' * MEGABYTE  # 1 MB of data
+        start_memory = psutil.Process().memory_info().rss
+
+        encoded, _ = encode_multipart_formdata([('k', data)], boundary=BOUNDARY)
+        end_memory = psutil.Process().memory_info().rss
+        self.assertTrue(end_memory <= start_memory + MEGABYTE)
```
haikuginger: Yeah, I would tend to agree. In cases where memory usage currently balloons, it's not like we don't know where the usage is, nor that the Python interpreter isn't freeing up unreferenced data; it's simply that we hold onto the reference for two sets of the data at once. Perhaps we could check that the data in the streamer is the same by identity? That'd test that we're not copying the data by mistake.

test/test_filepost.py
```diff
@@ -131,3 +168,195 @@ def test_request_fields(self):
...
+        fields = [('k', ('somefile.txt', _IterableObject(b'v'), 'image/jpeg'))]
+
+        encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY)
+        self.assertEqual(len(encoded), 146)
+
+    def test_length_unknown_object(self):
+        fields = [('k', ('somefile.txt', object(), 'image/jpeg'))]
+
+        encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY)
+        self.assertRaises(TypeError, encoded.__len__)
```
haikuginger: I believe that you can do `self.assertRaises(TypeError, len, encoded)`. Personal preference, though.

test/test_filepost.py
```diff
@@ -1,15 +1,49 @@
...
+            [('k', 'v'), ('k2', 'v2')],
+            [('k', b'v'), (u('k2'), b'v2')],
+            [('k', b'v'), (u('k2'), 'v2')],
+            [('foo', b'a'), ('foo', b'b')],
+        ]
+
+        for fields in fieldsets:
+            encoded, _ = encode_multipart_formdata(fields, boundary=BOUNDARY)
+            encoded = b''.join(encoded)
+            self.assertEqual(len(encoded), len(MultipartEncoderGenerator(fields, boundary=BOUNDARY)))
```
haikuginger: Isn't this basically testing against itself? I'd prefer to have fixed length numbers to test against.

urllib3/filepost.py
haikuginger: Agreed.

SethMichaelLarson:
Thank you @haikuginger and @Lukasa for the great first round of reviews! :) I will implement these changes post-Turkey day!

test/test_filepost.py
SethMichaelLarson: I think I'll just drop this test, maybe I can come up with a better way to test that the data is not copied or duplicated.

test/test_filepost.py
SethMichaelLarson: That's definitely more natural.

urllib3/filepost.py
SethMichaelLarson: That's exactly what they do, hence the undesired behavior comments ;) These were a part of the old stream that this work is based on so I copied them over. I was thinking we could hold the data inside of a buffer after getting the length but I wanted to get it looked at before I made that change.

test/test_filepost.py
SethMichaelLarson: I'll use your suggestion.

urllib3/filepost.py
SethMichaelLarson: Done!

urllib3/filepost.py
SethMichaelLarson: Thing is we have no idea what the iterators will return, the original implementation of this code seemed very optimistic about what will be passed into fields. I don't quite understand what you're asking for though, could you clarify? :)

urllib3/filepost.py
SethMichaelLarson: I will try to do the math directly.

SethMichaelLarson:
So this patch has 100% coverage and passing test suite. What further steps should I take or is this delayed for v2? Also should I still wait for sigma to review? :)

Lukasa:
Yup, @sigmavirus24 has some thoughts on this percolating somewhere inside his absurdly busy and overstretched brain. =)

SethMichaelLarson:
Awesome! I'll defer this until he has time! :)

SethMichaelLarson:
@Lukasa This PR will probably have to be redone in v2 right @Lukasa ?

Lukasa:
Yeah, it may be worth trying to defer it for v2.

SethMichaelLarson:
Closing this to defer to v2

spacether:
@sethmlarson 
> Closing this to defer to v2
Where is v2?

sethmlarson:
@spacether  It's in development on the main branch, not yet released. This feature is being worked on in  #2331 currently.

PR #1036, opened 2016-11-13
Differentiate socks5h from socks5 and socks4a from socks4

### LINKED ISSUE #1035: Differentiate socks5h from socks5 and socks4a from socks4 when handling proxy string
In a proxy string, socks5h:// and socks4a:// mean that the hostname is
resolved by the SOCKS server.  socks5:// and socks4:// mean that the
hostname is resolved locally.  socks4a:// means to use SOCKS4a, which is
an extension of SOCKS4.  Let's make urllib3 honor it.

ijcoe6ru:
In the proxy string, socks5h:// and socks4a:// mean that the hostname is
to be resolved by the socks server.  socks5:// and socks4:// mean the
hostname is to be resolved locally.

Fix #1035

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -1,4 +1,4 @@
-import unittest
+import sys
 import socket
 import threading
 from nose.plugins.skip import SkipTest
@@ -12,6 +12,11 @@ from dummyserver.server import (
 )
 from dummyserver.handlers import TestingApp
 from dummyserver.proxy import ProxyHandler
+
+if sys.version_info >= (2, 7):
+    import unittest
+else:
+    import unittest2 as unittest
 
 
 def consume_socket(sock, chunks=65536):
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -236,12 +236,7 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
             handler = handle_socks5_negotiation(sock, negotiate=False)
             addr, port = next(handler)
 
-            # We're not using assertIn because the tests need to pass under the
-            # old Python 2.6
-            if not addr in ['127.0.0.1', '::1']:
-                raise self.failureException(
-                    "Requested address is %s, 127.0.0.1 or ::1 expected" %
-                    safe_repr(addr))
+            self.assertIn(addr, ['127.0.0.1', '::1'])
             self.assertTrue(port, 80)
             handler.send(True)
 
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -236,7 +236,12 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
             handler = handle_socks5_negotiation(sock, negotiate=False)
             addr, port = next(handler)
 
-            self.assertIn(addr, ['127.0.0.1', '::1'])
+            # We're not using assertIn because the tests need to pass under the
+            # old Python 2.6
+            if not addr in ['127.0.0.1', '::1']:
+                raise self.failureException(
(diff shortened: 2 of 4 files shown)


test/contrib/test_socks.py
```diff
@@ -229,6 +229,37 @@ def request_handler(listener):
...
         self.assertEqual(response.headers['Server'], 'SocksTestServer')
 
+    def test_local_dns(self):
+        def request_handler(listener):
+            sock = listener.accept()[0]
+
+            handler = handle_socks5_negotiation(sock, negotiate=False)
+            addr, port = next(handler)
+
+            self.assertIn(addr, ['127.0.0.1', '::1'])
```
Lukasa: You will need to start using `unittest2` in these test cases to use `assertIn`: it's not present on Python 2.6.

ijcoe6ru:
I've added tests for doing DNS lookup locally and changed the tests that
use example.com to use socks4a and socks5h.

As of the concern that some code may break, they'll probably be fine
since some well-known software, such as curl, use this convention.

test/contrib/test_socks.py
ijcoe6ru: Excuse me.  Why are the tests run in Python 2.6 instead of 2.7?

SethMichaelLarson: We have to support Python 2.6 because our upstreams require it.

Lukasa: We don't just run in Python 2.6. We run in Python 2.6, Python 2.7, Python 3.3, Python 3.4, Python 3.5, and PyPy 2, across two different operating systems. =) The tests, and the code, must work on all of those platforms.

test/contrib/test_socks.py
```diff
@@ -236,7 +236,12 @@ def request_handler(listener):
             handler = handle_socks5_negotiation(sock, negotiate=False)
             addr, port = next(handler)
 
-            self.assertIn(addr, ['127.0.0.1', '::1'])
+            # We're not using assertIn because the tests need to pass under the
+            # old Python 2.6
+            if not addr in ['127.0.0.1', '::1']:
+                raise self.failureException(
+                    "Requested address is %s, 127.0.0.1 or ::1 expected" %
+                    safe_repr(addr))
```
Lukasa: There's no need to do this. You can leave the `self.assertIn` statement and then just edit `dummyserver/testcase.py` to change `import unittest` to

```python
import sys

if sys.version_info >= (2, 7):
    import unittest
else:
    import unittest2 as unittest
```

That should resolve the problem.

ijcoe6ru:
Thanks.  I hadn't noticed that unittest2 was used elsewhere, so I thought it would've introduced new dependencies.

dummyserver/testcase.py
```diff
@@ -1,4 +1,8 @@
-import unittest
+import sys
+if sys.version_info >= (2, 7):
+    import unittest
+else:
+    import unittest2 as unittest
```
Lukasa: So to make flake8 happy, let's move this block down to line 19.

ijcoe6ru:
I don't know why the test has failed. Can anyone help me?

Lukasa:
It's just a transient failure, I've restarted that test.

SethMichaelLarson:
To see details about which tests failed and on what platform, you can click on the Travis details in the required checks box here at the bottom of the PR.  It looks like the test that failed was a randomly failure.

salehe:
This is a very useful feature. Looking forward for it to be merged

ijcoe6ru:
Why hasn't this been merged yet?

SethMichaelLarson:
Huzzah! :D

PR #1039, opened 2016-11-14
Rewind Seekable Body on Retry/Redirect

nateprewitt:
This is picking up where #459 left off. There are two commits in the first pass on this PR, one (40a110c) with a bare bones (unsafe) fix for the issue that @joneskoo was hitting, and another (ae7d799) with a fuller port of Requests implementation.

I'm on the fence about whether the second is overkill. If we decide to go down that route though, I still need to add individual testing for `rewind_body`.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -10,7 +10,7 @@ from itertools import chain
 from mock import patch, Mock
 
 from urllib3 import add_stderr_logger, disable_warnings
-from urllib3.util.request import make_headers, rewind_body
+from urllib3.util.request import make_headers, rewind_body, _FAILEDTELL
 from urllib3.util.retry import Retry
 from urllib3.util.timeout import Timeout
 from urllib3.util.url import (
@@ -274,14 +274,16 @@ class TestUtil(unittest.TestCase):
         body.read()  # Consume body
 
         # Simulate failed tell()
-        self.assertRaises(UnrewindableBodyError, rewind_body, body, object())
+        body_pos = _FAILEDTELL
+        self.assertRaises(UnrewindableBodyError, rewind_body, body, body_pos)
 
     def test_rewind_body_bad_position(self):
         body = io.BytesIO(b'test data')
         body.read()  # Consume body
 
         # Pass non-integer position
-        self.assertRaises(UnrewindableBodyError, rewind_body, body, None)
+        self.assertRaises(ValueError, rewind_body, body, None)
+        self.assertRaises(ValueError, rewind_body, body, object())
 
     def test_rewind_body_failed_seek(self):
         class BadSeek():
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -1,6 +1,7 @@
 import hashlib
 import warnings
 import logging
+import io
 import unittest
 import ssl
 import socket
@@ -9,7 +10,7 @@ from itertools import chain
 from mock import patch, Mock
 
 from urllib3 import add_stderr_logger, disable_warnings
-from urllib3.util.request import make_headers
+from urllib3.util.request import make_headers, rewind_body
 from urllib3.util.retry import Retry
 from urllib3.util.timeout import Timeout
 from urllib3.util.url import (
@@ -31,6 +32,7 @@ from urllib3.exceptions import (
     SSLError,
     SNIMissingWarning,
     InvalidHeader,
+    UnrewindableBodyError,
 )
 from urllib3.util.connection import (
     allowed_gai_family,
@@ -255,6 +257,39 @@ class TestUtil(unittest.TestCase):
         self.assertEqual(
             make_headers(disable_cache=True),
             {'cache-control': 'no-cache'})
+
+    def test_rewind_body(self):
+        body = io.BytesIO(b'test data')
+        self.assertEqual(body.read(), b'test data')
+
+        # Assert the file object has been consumed
+        self.assertEqual(body.read(), b'')
+
+        # Rewind it back to just be b'data'
+        rewind_body(body, 5)
+        self.assertEqual(body.read(), b'data')
+
+    def test_rewind_body_failed_tell(self):
+        body = io.BytesIO(b'test data')
+        body.read()  # Consume body
+
+        # Simulate failed tell()
+        self.assertRaises(UnrewindableBodyError, rewind_body, body, object())
+
+    def test_rewind_body_bad_position(self):
+        body = io.BytesIO(b'test data')
+        body.read()  # Consume body
+
+        # Pass non-integer position
+        self.assertRaises(UnrewindableBodyError, rewind_body, body, None)
+
+    def test_rewind_body_failed_seek(self):
+        class BadSeek():
+
+            def seek(self, pos, offset=0):
+                raise IOError
+
+        self.assertRaises(UnrewindableBodyError, rewind_body, BadSeek(), 2)
 
     def test_split_first(self):
         test_cases = {
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -873,7 +873,7 @@ class TestRetryAfter(HTTPDummyServerTestCase):
 
 class TestRetryWithTimeout(HTTPDummyServerTestCase):
     def setUp(self):
-        self.pool = HTTPConnectionPool(self.host, self.port, timeout=0.001)
+        self.pool = HTTPConnectionPool(self.host, self.port, timeout=0.1)
 
     def test_retries_put_filehandle(self):
         """HTTP PUT retry with a file-like object should not timeout"""
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -1,4 +1,5 @@
 import errno
+import io
 import logging
 import socket
 import sys
@@ -869,6 +870,26 @@ class TestRetryAfter(HTTPDummyServerTestCase):
         delta = time.time() - t
         self.assertEqual(r.status, 200)
         self.assertTrue(delta < 1)
+
+class TestRetryWithTimeout(HTTPDummyServerTestCase):
+    def setUp(self):
+        self.pool = HTTPConnectionPool(self.host, self.port, timeout=0.001)
+
+    def test_retries_put_filehandle(self):
+        """HTTP PUT retry with a file-like object should not timeout"""
+        retry = Retry(total=3, status_forcelist=[418])
+        # httplib reads in 8k chunks; use a larger content length
+        content_length = 65535
+        uploaded_file = io.BytesIO(b'A' * content_length)
+        headers = {'test-name': 'test_put_fileobj',
+                   'Content-Length': str(content_length)}
+        resp = self.pool.urlopen('PUT', '/successful_retry',
+                                 headers=headers,
+                                 retries=retry,
+                                 body=uploaded_file,
+                                 assert_same_host=False, redirect=False)
+        self.assertEqual(resp.status, 200)
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -29,6 +29,7 @@ from urllib3.exceptions import (
     ReadTimeoutError,
     ProtocolError,
     NewConnectionError,
+    UnrewindableBodyError,
 )
 from urllib3.packages.six import b, u
 from urllib3.packages.six.moves.urllib.parse import urlencode
@@ -686,8 +687,6 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             self.assertEqual(http.pool.qsize(), http.pool.maxsize)
 
 
-
-
 class TestRetry(HTTPDummyServerTestCase):
     def setUp(self):
         self.pool = HTTPConnectionPool(self.host, self.port)
@@ -871,7 +870,7 @@ class TestRetryAfter(HTTPDummyServerTestCase):
         self.assertEqual(r.status, 200)
         self.assertTrue(delta < 1)
 
-class TestRetryWithTimeout(HTTPDummyServerTestCase):
+class TestFileBodiesOnRetryOrRedirect(HTTPDummyServerTestCase):
     def setUp(self):
         self.pool = HTTPConnectionPool(self.host, self.port, timeout=0.1)
 
@@ -880,8 +879,9 @@ class TestRetryWithTimeout(HTTPDummyServerTestCase):
         retry = Retry(total=3, status_forcelist=[418])
         # httplib reads in 8k chunks; use a larger content length
         content_length = 65535
-        uploaded_file = io.BytesIO(b'A' * content_length)
-        headers = {'test-name': 'test_put_fileobj',
+        data = b'A' * content_length
+        uploaded_file = io.BytesIO(data)
+        headers = {'test-name': 'test_retries_put_filehandle',
                    'Content-Length': str(content_length)}
         resp = self.pool.urlopen('PUT', '/successful_retry',
                                  headers=headers,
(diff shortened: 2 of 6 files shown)


nateprewitt:
@jonparrott, sorry to page you in here.

I've added what appears to be the first test using a file-like body to `TestConnectionPool` and GAE isn't having it. I did some additional testing, and it looks like appengine (at least if you enter at `urlfetch`) assumes the payload is a string. Is that correct? I'm not sure if this is an oversight with testing from urllib3's side, or me just be making a mess of things.

@sigmavirus24 was there a design choice to not include file-like bodies in the TestConnectionPool tests?

### Related Tracebacks
One for `io.BytesIO`, one for a standard file.

```
  File "/home/travis/build/nateprewitt/urllib3/test/with_dummyserver/test_connectionpool.py", line 719, in test_redirect_with_failed_tell
    body=body)
  File "/home/travis/build/nateprewitt/urllib3/test/contrib/test_gae_manager.py", line 44, in urlopen
    return self.manager.urlopen(method, url, *args, **kwargs)
  File "/home/travis/build/nateprewitt/urllib3/urllib3/contrib/appengine.py", line 149, in urlopen
    validate_certificate=self.validate_certificate,
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch.py", line 271, in fetch
    return rpc.get_result()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub_map.py", line 613, in get_result
    return self.__get_result_hook(self)
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch.py", line 378, in _get_fetch_result
    rpc.check_success()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub_map.py", line 579, in check_success
    self.__rpc.CheckSuccess()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_rpc.py", line 157, in _WaitImpl
    self.request, self.response)
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub.py", line 131, in MakeSyncCall
    method(request, response)
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch_stub.py", line 244, in _Dynamic_Fetch
    validate_certificate=validate_certificate)
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch_stub.py", line 366, in _RetrieveURL
    escaped_payload = payload.encode('string_escape')
AttributeError: '_io.BytesIO' object has no attribute 'encode'
```

```
  File "/home/travis/build/nateprewitt/urllib3/test/with_dummyserver/test_connectionpool.py", line 694, in test_redirect_with_Bytes_Object
    body=body)
  File "/home/travis/build/nateprewitt/urllib3/test/contrib/test_gae_manager.py", line 44, in urlopen
    return self.manager.urlopen(method, url, *args, **kwargs)
  File "/home/travis/build/nateprewitt/urllib3/urllib3/contrib/appengine.py", line 149, in urlopen
    validate_certificate=self.validate_certificate,
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch.py", line 271, in fetch
    return rpc.get_result()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub_map.py", line 613, in get_result
    return self.__get_result_hook(self)
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch.py", line 378, in _get_fetch_result
    rpc.check_success()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub_map.py", line 579, in check_success
    self.__rpc.CheckSuccess()
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_rpc.py", line 157, in _WaitImpl
    self.request, self.response)
  File "/home/travis/.cache/google_appengine/google/appengine/api/apiproxy_stub.py", line 108, in MakeSyncCall
    if request.ByteSize() > self.__max_request_size:
  File "/home/travis/.cache/google_appengine/google/appengine/api/urlfetch_service_pb.py", line 436, in ByteSize
    if (self.has_payload_): n += 1 + self.lengthString(len(self.payload_))
TypeError: object of type 'file' has no len()
```

jonparrott:
@nateprewitt never feel bad about summoning me. It's what I'm here for.

I think you're right - App Engine isn't going to play well with anything other than str/bytes. It's probably something we should explicitly raise for in the App Engine manager.

nateprewitt:
Great, thanks for confirming @jonparrott!

nateprewitt:
In regards to transient failures, tox runs clean locally but I'm seeing failure in at least one node on 'test_timeout' on each push. Restarting the node generally fixes the issue but it's odd that it seems so "consistently inconsistent". This same error has been cropping up on other PRs for at least the last few weeks but I'm not sure if I've exacerbated it.

Lukasa:
Almost any test that involves a timeout gets really tricky on the CI runs because they run virtualized, which means they're subject to all kinds of weird timing variations *and* run remarkably slowly. This causes a real exacerbation of timing issues.

nateprewitt:
Yeah, there's definitely something not quite right. I've spent the last few days testing this and am at a bit of a loss. Each individual piece I've changed with the others mocked won't cause this spike in `timeout_test` failures on Travis, but when you put them together they dramatically raise the rate of occurrence. I'll keep digging. I'm wondering if it's somehow increasing the execution time of `urlopen` just enough to push the timeout over the threshold often.

nateprewitt:
So... I think these errors were brought on by some kind of shared connection/connection state between test cases? That or some voodoo inside of Travis' networking.

Anyway, moving `test_redirect_with_failed_tell` into a different test class fixes the spiked failure rate. This may be because it's not manually closing the connection when it raises the Exception? Is there more teardown I should be handling here?

Lukasa:
I think each test class runs a single server, so it's possible there is some lingering state there.

urllib3/util/request.py
```diff
@@ -70,3 +71,27 @@ def make_headers(keep_alive=None, accept_encoding=None, user_agent=None,
         headers['cache-control'] = 'no-cache'
 
     return headers
+
+
+def rewind_body(body, pos):
+    """
+    Attempt to rewind body to a certain position.
+    Primarily used for request redirects and retrys.
```
nateprewitt: One day I'll English pretty too. s/retrys/retries/g on the next push.

nateprewitt:
Ok @Lukasa, I think this ones ready for a peek whenever you find yourself hankering to do a review.

urllib3/connectionpool.py
```diff
@@ -576,6 +577,19 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
+        # value stored, let's try to get one for future use.
+        if body_pos is not None:
+            rewind_body(body, body_pos)
+        elif getattr(body, 'tell', None) is not None:
+            try:
+                body_pos = body.tell()
+            except (IOError, OSError):
+                # This differentiates from None, allowing us to catch
+                # a failed `tell()` later when trying to rewind the body.
+                body_pos = object()
```
Lukasa: While we're here, I'd rather we used a sentinel object for this than just any instance of `object()`.

urllib3/exceptions.py
```diff
@@ -239,3 +239,8 @@ class HeaderParsingError(HTTPError):
     def __init__(self, defects, unparsed_data):
         message = '%s, unparsed data: %r' % (defects or 'Unknown', unparsed_data)
         super(HeaderParsingError, self).__init__(message)
+
+
+class UnrewindableBodyError(HTTPError):
+    "Requests encountered an error when trying to rewind a body"
```
Lukasa: Requests?

test/with_dummyserver/test_connectionpool.py
```diff
@@ -870,5 +870,65 @@ def test_redirect_after(self):
...
+        content_length = 65535
+        uploaded_file = io.BytesIO(b'A' * content_length)
+        headers = {'test-name': 'test_redirect_put_fileobj_timeout',
+                   'Content-Length': str(content_length)}
+        resp = self.pool.urlopen('PUT', '/redirect?target=/successful_retry',
+                                 headers=headers,
+                                 retries=retry,
+                                 body=uploaded_file,
+                                 assert_same_host=False, redirect=True)
+        self.assertEqual(resp.status, 200)
```
nateprewitt: I'm actually not sure how we'd get that information with the current testing structure. The dummyserver doesn't return our read request data and we can't access the request from the response object.

I've tried sending the redirect to /put but test aborts on ```TypeError: put() takes exactly 1 argument (2 given)``` because Tornado is passing more params than we're expecting. I can't find any tests calling '/put' or '/post' so I'm not sure they've ever been used?

I've also tried '/echo' which would normally be what we want but it just passes us back an empty string. I thought we possibly weren't rewinding, but even passing a file-like object directly to '/echo' without a redirect will still return ''. It looks like Tornado is consuming the file objects and then not passing that data on to the handler.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -870,5 +870,65 @@ def test_redirect_after(self):
...
+
+    def test_redirect_with_failed_tell(self):
+        """Abort request if failed to get a position from tell()"""
+        class BadTellObject():
```
nateprewitt: Nope, that's what I did originally but misinterpreted httplib using BytesIO's unimplemented `fileno` as being a result of the broken tell(). Turns out I just needed to set the content length.

urllib3/connectionpool.py
nateprewitt: So just to clarify, you're talking about moving lines 589-591 into utils/requests.py, yes? I'm not sure we want to move much more than that out of `urlopen`.

Lukasa: Nope, I want to move the entire logical block: lines 580-591.

test/with_dummyserver/test_connectionpool.py
Lukasa: If we can't get the data out of the testing harness, that just means we need to enhance the testing harness. =)

nateprewitt: Alright, let's do it!

nateprewitt:
Things should be tidied and ready for another glance on Monday, @Lukasa. Have a good weekend!

test/test_util.py
```diff
@@ -256,6 +258,39 @@ def test_make_headers(self):
...
+        # Rewind it back to just be b'data'
+        rewind_body(body, 5)
+        self.assertEqual(body.read(), b'data')
+
+    def test_rewind_body_failed_tell(self):
+        body = io.BytesIO(b'test data')
+        body.read()  # Consume body
+
+        # Simulate failed tell()
+        self.assertRaises(UnrewindableBodyError, rewind_body, body, object())
```
Lukasa: This should probably now use the sentinel object directly.

test/test_util.py
```diff
@@ -256,6 +258,39 @@ def test_make_headers(self):
...
+
+    def test_rewind_body_bad_position(self):
+        body = io.BytesIO(b'test data')
+        body.read()  # Consume body
+
+        # Pass non-integer position
+        self.assertRaises(UnrewindableBodyError, rewind_body, body, None)
```
Lukasa: For invalid positions, should we throw `UnrewindableBodyError`, or simply a `ValueError`? No strong guidance here, just a question work thinking about.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -870,5 +870,58 @@ def test_redirect_after(self):
...
+        content_length = 65535
+        data = b'A' * content_length
+        uploaded_file = io.BytesIO(data)
+        headers = {'test-name': 'test_redirect_put_file',
+                   'Content-Length': str(content_length)}
+        resp = self.pool.urlopen('PUT', '/echo',
+                                 headers=headers,
+                                 retries=retry,
+                                 body=uploaded_file,
+                                 assert_same_host=False, redirect=True)
```
Lukasa: So this new test is good, but does it..uh, does it ever actually rewind the body?

urllib3/util/request.py
```diff
@@ -70,3 +72,45 @@ def make_headers(keep_alive=None, accept_encoding=None, user_agent=None,
...
+def rewind_body(body, pos):
+    """
+    Attempt to rewind body to a certain position.
+    Primarily used for request redirects and retries.
+
+    :param body:
+        File-like object that supports seek.
+
+    :param pos:
+        Position to seek to in file. An `object()` value
```
Lukasa: No it isn't.

test/with_dummyserver/test_connectionpool.py
nateprewitt: Nope! Stripped the redirect while testing and forgot to add it back in.

test/test_util.py
nateprewitt: I think I originally entertained that idea with Requests implementation, but then we made `_body_position` "private". The common case we're catching here with an invalid position is the `_FAILEDTELL` sentinel object, which the user isn't setting. It made more sense to hide that abstraction from people not looking to go digging through the code for what `ValueError: object() is not a valid file position.`

It may make more sense here since this is a public parameter to generalize the error, I don't have strong feelings either way. This was just legacy logic from my original approach.

nateprewitt:
Forgot to leave a note, but comment and sentinel object are addressed. I expanded `rewind_body` to raise a `ValueError` when appropriate too, since `body_pos` is a bit more user facing here.

urllib3/util/request.py
```diff
@@ -70,3 +72,47 @@ def make_headers(keep_alive=None, accept_encoding=None, user_agent=None,
...
+        Position to seek to in file.
+    """
+    body_seek = getattr(body, 'seek', None)
+    if body_seek is not None and isinstance(body_pos, integer_types):
+        try:
+            body_seek(body_pos)
+        except (IOError, OSError):
+            raise UnrewindableBodyError("An error occured when rewinding request "
+                                        "body for redirect/retry.")
+    elif body_pos == _FAILEDTELL:
```
Lukasa: This should be an identity check (`if body_pos is _FAILEDTELL`).

nateprewitt:
Test amended, should be ready for a glance from any other contributors that want to take a look.

haikuginger:
@Lukasa I unfortunately don't have the bandwidth for this at the moment; very exciting though!

test/with_dummyserver/test_connectionpool.py
```diff
@@ -870,5 +870,59 @@ def test_redirect_after(self):
...
+    def test_redirect_with_failed_tell(self):
+        """Abort request if failed to get a position from tell()"""
+        class BadTellObject(io.BytesIO):
+
+            def tell(self):
+                raise IOError
+
+        body = BadTellObject(b'the data')
+        url = '/redirect?target=/successful_retry'
+        headers = {'Content-Length': '8'}
```
Lukasa: Fair enough. Might be good to have a comment to that effect here.

nateprewitt:
Yep, that seems completely reasonable. Comment updated 😊

nateprewitt:
3.4 will need a quick kick.

urllib3/connectionpool.py
```diff
@@ -449,7 +450,7 @@ def is_same_host(self, url):
     def urlopen(self, method, url, body=None, headers=None, retries=None,
                 redirect=True, assert_same_host=True, timeout=_Default,
                 pool_timeout=None, release_conn=None, chunked=False,
-                **response_kw):
+                body_pos=None, **response_kw):
```
jonparrott: Add this new argument to the doc string?

nateprewitt:
@Lukasa are we waiting a bit longer for further input, or should I rebase this onto the new master?

Lukasa:
Yeah, I was planning to wait until Friday. Note that I can update your branch as well, so there's no need for you to worry about rebasing.

nateprewitt:
Cool, sounds good :)

Issue #694, opened 2015-08-19
Timeout calculation vulnerable to clock changes


willnewton:
The calculation of timeouts in util/timeout.py uses time.time() which is the wall clock time which may be adjusted via various mechanisms such as NTP. This means that an NTP adjustment can make the timeout unduly short or long. Python 3.3 adds time.monotonic() which is a clock that is unaffected by clock changes so will provide more predictable timeouts.
##

Lukasa:
This is a really tricky area because the monotonic clock does not exist in the stdlib for versions of Python earlier than 3.3.

There is a PyPI module, [monotime](https://pypi.python.org/pypi/Monotime), but that doesn't support Windows. There's another module, [python-monotonic](https://pypi.python.org/pypi/monotonic/), which does support Windows, but then requires ctypes which is a bit terrible to have as a dependency.

My recommendation here is in two steps:

1. We should use `time.monotonic()` if it's available.
2. We could consider adding `python-monotonic` as an optional dependency à la `pyopenssl`, and if it's present we can prefer it.

I think doing just 1 by itself would be an improvement, but 2 would probably be better if `python-monotonic` actually works as suggested.

shazow:
How about we move `urllib3.util.timeout.current_time` into `urllib3.util.timeout.Timeout.current_time`, then people can replace it with whatever callable they want?

SethMichaelLarson:
So I was thinking about adapting the python-monotonic approach to suite our own needs. Basically use ctypes as late as possible and if it's not available or fails then we use `time.time` in it's place because we have no other option.  Is anyone opposed to this or is there a reason not to do this?

Lukasa:
The biggest problem with ctypes, IIRC, is that it performs pretty poorly on PyPy. I'll double check though.

SethMichaelLarson:
We could always skip anything with ctypes on PyPy and use `time.time` just in case if we want to be extra careful.  Or if it doesn't work on a certain PyPy+Platform combo we can always avoid that. :) Eager to hear what you find!

Lukasa:
I mean, it'll work on PyPy, it'll just be dog slow. But we can try it, I guess, and see how it goes. =)

SethMichaelLarson:
I'll give it a try when I get home. :) I'll put it in urllib3/utils and you can review it tomorrow morning (my time)

sethmlarson:
We now only use `time.monotonic` when it's available. Closing this issue.

PR #1041, opened 2016-11-15
Use time.monotonic if available

SethMichaelLarson:
Did the minimum suggested in #694 and started using `time.monotonic` if it's available. Discussion about whether we should vendor any library like `python-monotime` or `monotonic`, but this is a great first step.

### DIFF
--- a/urllib3/util/timeout.py
+++ b/urllib3/util/timeout.py
@@ -11,11 +11,10 @@ from ..exceptions import TimeoutStateError
 _Default = object()
 
 
-def current_time():
-    """
-    Retrieve the current time. This function is mocked out in unit testing.
-    """
-    return time.time()
+# Use time.monotonic if available.
+current_time = time.time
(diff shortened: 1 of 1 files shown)


urllib3/util/timeout.py
```diff
@@ -11,11 +11,10 @@
...
 
 
-def current_time():
-    """
-    Retrieve the current time. This function is mocked out in unit testing.
-    """
-    return time.time()
+# Use time.monotonic if available.
+current_time = time.time
+if hasattr(time, "monotonic"):
```
sigmavirus24: Why not do something like:

```py
current_time = getattr(time, 'monotonic', time.time)
```

This avoids using hasattr, and keeps the logic nice and simple

SethMichaelLarson: Good idea, I'll do this. :)

SethMichaelLarson:
Ready for review. :)

Issue #1044, opened 2016-11-21
Old "cryptography" package causes crash


aseering:
Cross-posting this from kennethreitz/requests#3714 ; folks there seemed to think that this would be best addressed within urllib3:

As of `requests==2.12.1` \[EDIT: and whatever version of urllib3 it embeds\], `requests[security]` depends on `cryptography>=1.3.4`.  That's fine.  However, `requests` (without `[security]`) crashes if `cryptography==1.0.2` is installed.  This is a perfectly valid configuration of packages according to `pip`; as a user, I would expect it to behave as if I had not specified `[security]` (because I didn't specify it).

The specific failure is discussed a bit more here:

http://discuss.flexget.com/t/unhandled-error-in-plugin-configure-series-solved/2900

The crash was resolved, also as described at the above link, by running:

```
pip install --upgrade cryptography
```

Uninstalling `cryptography` would probably have worked fine too.

The crash appears to be caused because `requests` internally tests for `[security]` by importing a thing which imports its bundled `pyopenssl` which imports `cryptography`; if that succeeds, the extra is used.  But with a too-old version of `cryptography`, it imports just fine but it constructs objects that are missing required fields.

I can think of a couple approaches here:

* Catch this failure and print an error message that more-helpfully explains that your `cryptography` package needs updating
* Validate the version/functionality of the `cryptography`package more carefully, and don't enable the `[security]` extra's features if an old version happens to be installed

I'd be happy to post a PR for either approach.  Right now I'm looking for feedback on whether it's worth addressing this (I'd certainly appreciate it) and on what approach you'd prefer.

Lukasa:
Hi again!

So, urllib3 *also* has an appropriate extra that *also* enforces this constraint, so we want to worry about it here too. I think we need to adjust `inject_into_urllib3` to do some feature detection. I'd be happy to review a pull request that adds this kind of checking.

Lukasa:
Resolved in #1045.

PR #1045, opened 2016-11-21
Require an up-to-date 'cryptography' module for urllib3 injection

aseering:
This fixes Issue #1044 for me.

### DIFF
--- a/test/contrib/test_pyopenssl_dependencies.py
+++ b/test/contrib/test_pyopenssl_dependencies.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+import unittest
+
+from nose.plugins.skip import SkipTest
+
+try:
+    from urllib3.contrib.pyopenssl import (inject_into_urllib3,
+                                           extract_from_urllib3)
+except ImportError as e:
+    raise SkipTest('Could not import PyOpenSSL: %r' % e)
+
+from mock import patch
+
+class TestPyOpenSSLInjection(unittest.TestCase):
+    """
+    Tests for error handling in pyopenssl's 'inject_into urllib3'
+    """
+    def test_inject_validate_fail(self):
+        """
+        Injection should not be supported if we are missing required dependencies.
+        """
+        successfully_injected = False
+        try:
+            with patch("cryptography.x509.extensions.Extensions") as mock:
+
+                # The following two lines are what this test intends to test.
+                # The remainder of this function is setup and clean-up logic.
+                del mock.get_extension_for_class
+                self.assertRaises(ImportError, inject_into_urllib3)
+
+                successfully_injected = True
+        finally:
+            if successfully_injected:
+                # `inject_into_urllib3` is not supposed to succeed.
+                # If it does, this test should fail, but we should
+                # clean up so that subsequent tests are unaffected.
+                extract_from_urllib3()
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -110,6 +110,8 @@ log = logging.getLogger(__name__)
 def inject_into_urllib3():
     'Monkey-patch urllib3 with PyOpenSSL-backed SSL-support.'
 
+    _validate_dependencies_met()
+
     util.ssl_.SSLContext = PyOpenSSLContext
     util.HAS_SNI = HAS_SNI
     util.ssl_.HAS_SNI = HAS_SNI
@@ -125,6 +127,18 @@ def extract_from_urllib3():
     util.ssl_.HAS_SNI = orig_util_HAS_SNI
     util.IS_PYOPENSSL = False
     util.ssl_.IS_PYOPENSSL = False
+
+
+def _validate_dependencies_met():
+    """
+    Verifies that PyOpenSSL's package-level dependencies have been met.
+    Throws `ImportError` if they are not met.
+    """
+    # Method added in `cryptography==1.1`; not available in older versions
+    from cryptography.x509.extensions import Extensions
+    if not hasattr(Extensions, "get_extension_for_class"):
+        raise ImportError("'cryptography' module missing required functionality.  "
+                          "Try upgrading to v1.3.4 or newer.")
 
 
 def _dnsname_to_stdlib(name):


test/contrib/test_pyopenssl.py
```diff
@@ -55,3 +57,13 @@ def test_dnsname_to_stdlib_leading_splat(self):
...
         self.assertEqual(_dnsname_to_stdlib(name), expected_result)
+
+    def test_inject_validate_fail(self):
+        """
+        Injection should not be supported if we are missing required dependencies
+        """
+        with patch("cryptography.x509.extensions.Extensions") as mock:
+            del mock.get_extension_for_class
+            with self.assertRaises(ImportError):
+                inject_into_urllib3()
```
aseering: This test passes for me locally. At least, I'm pretty sure it does. I actually get a lot of failures when I try to run this suite locally, though I don't think this test is among them. How do y'all run the test suite locally? Or do you just always use Travis-CI?

aseering:
I have a script on my machine that reproduces this on a more real-world (not monkeypatched) environment.  However, it's a shell script that runs some specific `pip install` commands.  Let me know if it would be useful; I can post it somewhere.

SethMichaelLarson:
@aseering Definitely would be useful, you should include it somewhere in this thread.

urllib3/contrib/pyopenssl.py
```diff
@@ -127,6 +129,17 @@ def extract_from_urllib3():
...
 
+def _validate_dependencies_met():
+    """
+    Verifies that PyOpenSSL's package-level dependencies have been met.
+    Throws `ImportError` if they are not met.
+    """
+    # Method added in `cryptography==1.1`; not available in older versions
+    from cryptography.x509.extensions import Extensions
+    if not hasattr(Extensions, "get_extension_for_class"):
+        raise ImportError("'cryptography' module missing required function (try upgrading the module?): 'cryptography.x509.extensions.Extensions.get_extension_for_class'")
```
nateprewitt: This is failing in the flake8 tests because it's over the 99 character line length maximum. You'll need to line wrap the error message down to something like:

```python
raise ImportError("'cryptography' module missing required function (try upgrading the "
                  "module?): 'cryptography.x509.extensions.Extensions."
                  "get_extension_for_class'")
```

It may also be worth consolidating the error to a more generic "cryptography doesn't meet minimum requirements of x.x". That would allow the ImportError to encapsulate a wider range of dependency issues.

nateprewitt: As another side note, I think this only shows that cryptography meets version 1.1. Is there any way we can modify the validation so we meet the 1.3.4 requirement in setup.py? That way we don't have a weird limbo state in 1.2-1.3.3 were this passes but we may still be missing dependencies.

test/contrib/test_pyopenssl.py
```diff
@@ -55,3 +57,13 @@ def test_dnsname_to_stdlib_leading_splat(self):
         expected_result = '*.xn--p1b6ci4b4b3a.xn--11b5bs8d'
 
         self.assertEqual(_dnsname_to_stdlib(name), expected_result)
+
+    def test_inject_validate_fail(self):
+        """
+        Injection should not be supported if we are missing required dependencies
```
nateprewitt: minor nit: maybe a period here to match the other docstrings?

urllib3/contrib/pyopenssl.py
aseering: Regarding validating the 1.2-1.3.3 range, actually, does anyone know why that requirement was put into place?  I could easily just validate the version-string directly if that's the preference.  But I feel like the spirit of duck typing would want this test to check for the particular functionality in that release urllib3 depends on.

aseering: Regarding your other comments, thanks!  I'll update my PR accordingly the next time I have a chance to sit down at a computer and make the changes.

nateprewitt: @aseering, I definitely agree with checking for functionality rather than hardcoding a version check. @Lukasa will likely have a better guess as to what feature(s) make 1.3.4 the min requirement.

aseering:
Here's a reproducer shell script (with some embedded Python):

```
#!/bin/sh

# Get a checkout of urllib3 to test against, if we don't already have one
if [ ! -d urllib3 ]; then
    git clone https://github.com/shazow/urllib3.git
fi
export PYTHONPATH="$PWD/urllib3"

virtualenv repro_env
source repro_env/bin/activate

pip install --upgrade pip
pip install PyOpenSSL==0.15 cryptography==1.0.2

python <<EOF

import urllib3
import certifi
http = urllib3.PoolManager(cert_reqs='CERT_REQUIRED',
                           ca_certs=certifi.where())
r = http.request('GET', 'https://github.com')
EOF

EXIT_CODE="$?"

if [ $EXIT_CODE -eq 0 ] ; then
    echo "Bug did not reproduce"
else
    echo "Bug reproduced"
fi

rm -rf repro_env

exit $EXIT_CODE
```

test/contrib/test_pyopenssl.py
```diff
@@ -55,3 +57,11 @@ def test_dnsname_to_stdlib_leading_splat(self):
...
 
         self.assertEqual(_dnsname_to_stdlib(name), expected_result)
+
+    def test_inject_validate_fail(self):
+        """
+        Injection should not be supported if we are missing required dependencies.
+        """
+        with patch("cryptography.x509.extensions.Extensions") as mock:
+            del mock.get_extension_for_class
+            self.assertRaises(ImportError, inject_into_urllib3)
```
Lukasa: So this test is going to be a bit problematic because the module is *already* injected into urllib3 thanks to `setup_module` in this file. You may want to pull this out into a different test file (possibly a brand new one).

urllib3/contrib/pyopenssl.py
```diff
@@ -127,6 +129,18 @@ def extract_from_urllib3():
...
+    """
+    Verifies that PyOpenSSL's package-level dependencies have been met.
+    Throws `ImportError` if they are not met.
+    """
+    # Method added in `cryptography==1.1`; not available in older versions
+    from cryptography.x509.extensions import Extensions
+    if not hasattr(Extensions, "get_extension_for_class"):
+        raise ImportError("'cryptography' module missing required functionality.  "
+                          "Try upgrading to v1.3.4 or newer.")
+
```
Lukasa: I think this looks like a good approach to this problem to me.

test/contrib/test_pyopenssl.py
aseering: Ok, I've pulled this out into a new file.

test/contrib/test_pyopenssl_dependencies.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Tests for error handling in pyopenssl's 'inject_into urllib3'
+    """
+    def test_inject_validate_fail(self):
+        """
+        Injection should not be supported if we are missing required dependencies.
+        """
+        with patch("cryptography.x509.extensions.Extensions") as mock:
+            del mock.get_extension_for_class
+            self.assertRaises(ImportError, inject_into_urllib3)
```
Lukasa: Can this test make sure it adds, as a cleanup step, `extract_from_urllib3`? That way if this test fails it doesn't dirty up other test runs.

aseering: Hm...  Is `extract_from_urllib3` supposed to always be safe to run even if `inject_into_urllib3` has not been called?  I wasn't assuming that was part of the API's contract -- I assumed it was more likely that `inject_into_urllib3` would correctly handle repeated calls because developers would be more likely to call it by mistake, hence putting this test in the other file.

I can add a try/except/finally block that cleans up only if no assertion is raised.  That would be less than elegant, but it would arguably be more correct...

Lukasa: Might be the better way to go though. =)

The issue isn't "can we call `inject_into_urllib3` more than once", it's that if we call into it and it succeeds, and we run this test before we run any of the others, all the other tests are now running against PyOpenSSL instead of the stdlib. That's not a good outcome.

aseering: Yeah -- "can we call `inject_into_urllib3` more than once" was the issue that you raised prior to this one :-)  Anyway, I'll make the try/finally change.  (Upon further thought, I think I can clean it up a bit and skip the `except` clause.)

aseering:
@Lukasa -- I think I've made your requested changes; at your convenience could you let me know what you think?

Also -- one of the test runs seems to have failed.  But that has been true for each of my test runs, and it's always a different test (and always on just one of the target Python versions, and never the same one) and never obviously related to my code.  So I'm not really sure how to tell when this code is passing tests and ready to merge.  Any thoughts?

test/contrib/test_pyopenssl_dependencies.py
```diff
@@ -0,0 +1,37 @@
...
+
+                # The following two lines are what this test intends to test.
+                # The remainder of this function is setup and clean-up logic.
+                del mock.get_extension_for_class
+                self.assertRaises(ImportError, inject_into_urllib3)
+
+                successfully_injected = True
+        finally:
+            if successfully_injected:
+                # `inject_into_urrlib3` is not supposed to succeed.
```
Lukasa: Nit: `urllib3`.

aseering:
@Lukasa -- d'oh!  Fixed.

aseering:
Thanks @Lukasa for the detailed review!

PR #1048, opened 2016-11-22
Check `getresponse` signature on import

ncoghlan:
`HTTPConnection.getresponse` only accepts a `buffering`
argument on Python 2.7.

Moving the check to import time means that calls  in
`urllib3.connectionpool` will always use the correct
signature, regardless of Python version.


Lukasa:
Thanks @ncoghlan! It'd be really good if we could test this in some way. Are you up for writing a test?

ncoghlan:
(Note: CI failure is due to flake8, I'll look more closely at the formatting tomorrow)

As far as testing goes, I was treating this more as a refactoring than anything else - if the existing tests pass across all supported runtimes, then I didn't break anything and the replacement code is functionally equivalent to the old code.

What would you be looking for a dedicated test to actually test? The one thing that comes to mind for me would be a test that mocks out `urllib3.connectionpool._get_httplib_response` to ensure that it's actually being called as expected.

Lukasa:
I guess my biggest concern is to confirm that the functionality is really working as we expect, but it's possible that treating this as a refactor is good enough.

SethMichaelLarson:
I'd second a test, any time you do introspection like this things can easily break on regression later.

urllib3/connectionpool.py
```diff
@@ -98,6 +98,23 @@ def close(self):
 # This is taken from http://hg.python.org/cpython/file/7aaba721ebc0/Lib/socket.py#l252
 _blocking_errnos = set([errno.EAGAIN, errno.EWOULDBLOCK])
 
+# CPython 2.7 supports an opt-in parameter for response buffering
+# in HTTPConnection.getresponse
+# Rather than checking on each request, we inspect the code object
+# (if available) at import and only pass the argument if it appears
+# in the list of variable names
+def _request_httplib_response_buffering():
```
sigmavirus24: This should use the inspect module. The only fault here is that you need conditions to handle differences between Python 3.3 and beyond and earlier. See [this](https://github.com/PyCQA/flake8/blob/master/src/flake8/utils.py#L293..L306) for an idea of how this could be done.

ncoghlan: I was trying to avoid that, since `inspect` is a relatively heavy import (not as large as `urllib3` itself, but not small either - on CPython 3.5 it would add 10 net new modules beyond what "import urllib3" already brings in). The theory behind the current approach is that we're *not* trying to detect a documented Python feature - we're trying to detect an undocumented implementation-specific behaviour of CPython 2.7. That means that an attribute error on `im_func` (indicating Python 3's lack of unbound functions), an attribute error on `func_code` (indicating a 2.x implementation without CPython-style code objects), or a lack of `buffering` in `co_varnames` (indicating a method implementation without a `buffering` local variable, and hence definitely no such parameter) all indicate "don't pass `buffering=True`)

You also reminded me that I never followed through on http://bugs.python.org/issue27172 (the `getfullargspec()` deprecation is being reversed, since it now builds on `inspect.signature()` internally, and doesn't lose information the way `getargspec()` does). If the release manager approves, I'll aim to get that addressed for 3.6.0 (and backported to 3.5).

Lukasa: So @ncoghlan it should be noted that if we want to continue with this current mechanism (which is fine by me), we should absolutely reproduce your logic above in a code comment. This is going to be tricky code to maintain, particularly for anyone less versed in Python internals than you (which is to say 98% of Python programmers of any stripe), so understanding what is being done here and the various ways in which this can return the correct result are really important.

ncoghlan:
While adding the comments @Lukasa requested, I realised that a really robust guard against a mistake in the import time introspection would be to still trap the possible `TypeError` when passing `buffering` explicitly.

However, as I added that guard, I also realised that given the combination of a helper function and the `TypeError` guard, it becomes possible to replace the helper at runtime the first time it fails, and the import time introspection can be discarded completely.

Essentially it becomes a module level implementation of the strategy pattern - we start with the CPython 2.7 strategy (i.e. pass `buffering=True`), but if that isn't compatible with the current runtime, we replace it with an approach we know will work.

urllib3/connectionpool.py
```diff
@@ -99,6 +99,30 @@ def close(self):
...
+        pass
+    # This isn't CPython 2.7, so don't even try to pass "buffering"
+    # on future calls, and instead invoke the connection method directly
+    _replace_get_httplib_response()
+    # Retry the original failed call without the "buffering" argument
+    return conn.getresponse()
+
+
+def _replace_get_httplib_response():
+    """Use HTTPConnection.getresponse directly as _get_hittplib_response"""
```
Lukasa: Nit: `hittplib` is not the name of that module.

urllib3/connectionpool.py
```diff
@@ -99,6 +99,30 @@ def close(self):
...
+# as an implementation detail of CPython 2.7, so other versions and other
+# implementations don't support it.
+def _get_httplib_response(conn):
+    """Get HTTP connection response with buffering explicitly enabled"""
+    try:
+        # CPython 2.7 supports this call, while other versions and
+        # implementations may not do so (e.g. 3.x doesn't)
+        return conn.getresponse(buffering=True)
+    except TypeError:
+        pass
```
Lukasa: Rather than `except...pass` it might be better just to bring the rest of the code inside the except block.

ncoghlan: This layout is to avoid Python 3's exception chaining if/when the conn.getresponse() call fails. However, I can bring the `_replace_get_httplib_response()` call inside the except block, and only have the `conn.getresponse()` call outside it, which will eliminate the need for `pass`.

ncoghlan:
With the latest changes, I think I'm seeing some sensible test cases as well:

1. The already suggested case of mocking out `_get_httplib_response` and ensuring it gets called when expected.
2. I need to tweak the patch so that the default `_get_httplib_response` implementation remains available as `_get_buffered_httplib_response` even if it gets replaced. Then the tests can make sure that `conn.getresponse(buffering=True)` is called as expected by default *and* that the implementation gets replaced if that call throws TypeError.

Lukasa:
Uh...github seems *quite* confused about the diff.

ncoghlan:
Heh, I guess I pushed the online editor to its breaking point :)

I'll switch to editing locally, and not push any new updates until I have the tests written as well.

ncoghlan:
Starting to write the test case made it clear that the import-time check suggested in this version of the PR is fundamentally flawed, as it doesn't account for `ConnectionCls` potentially being set on subclasses, so I'm going to close this PR.

However, it should still be possible to only do the try-and-fail-with-TypeError once per connection pool instance, rather than on every attempt to read a response from a connection object, so I'm going to see if I can make a PR that works that way instead.

Issue #1046, opened 2016-11-22
Necessary to cleanup DefaultSelector in def _wait_for_io_events() ?


aniapte:
Hello,

We are hitting the error "filedescriptor out of range in select()"  when creating too many new connections. We are using urllib3 as a part of requests. In order to get around it we are waiting for https://github.com/shazow/urllib3/pull/1001 to get released. The questions I have are:

1. When will https://github.com/shazow/urllib3/pull/1001 be released as part of a new urllib3 release?
2. I was looking at the patch and was wondering if any cleanup is required for the DefaultSelector() object that is created in the util._wait_for_io_events() function. See https://github.com/shazow/urllib3/blob/master/urllib3/util/wait.py#L22. The epoll selector class does have the close() method where it closes the platform epoll object - but is the close() called from anywhere?

regards,
aniruddha

Lukasa:
1. As it's been merged, it'll be part of the next release. There is no schedule for that release, though. Such a release also has to propagate into requests, which can usually mean a wait time of around 3 months.

2. Yeah, probably that should use the selector like a context manager. The explicit cleanup isn't entirely necessary, the gc will eventually get everything, but there's no reason not to use a context manager here. I'll welcome a patch for this. /cc @SethMichaelLarson

SethMichaelLarson:
I'm on it!

PR #1049, opened 2016-11-22
Use DefaultSelector as context manager.

SethMichaelLarson:
PR for #1046, uses the selector as a context manager to ensure that it is properly cleaned up.

### DIFF
--- a/test/test_selectors.py
+++ b/test/test_selectors.py
@@ -536,22 +536,7 @@ class BaseSelectorTestCase(unittest.TestCase, AlarmMixin, TimerMixin):
         after_fds = len(proc.open_files())
         self.assertEqual(before_fds, after_fds)
 
-    def test_wait_io_close_is_called(self):
-        selector = self.SELECTOR()
-        self.addCleanup(selector.close)
-
-        def fake_constructor(*args, **kwargs):
-            return selector
-
-        old_selector = wait.DefaultSelector
-        wait.DefaultSelector = fake_constructor
-        self.addCleanup(setattr, wait, "DefaultSelector", old_selector)
-
-        rd, wr = self.make_socketpair()
-        wait.wait_for_write([rd, wr], 0.001)
-        self.assertIs(selector._map, None)
-
-    def test_wait_selector_error(self):
+    def test_selector_error(self):
         err = selectors.SelectorError(1)
         self.assertEqual(err.__repr__(), "<SelectorError errno=1>")
         self.assertEqual(err.__str__(), "<SelectorError errno=1>")
@@ -560,7 +545,7 @@ class BaseSelectorTestCase(unittest.TestCase, AlarmMixin, TimerMixin):
 class BaseWaitForTestCase(unittest.TestCase, TimerMixin, AlarmMixin):
     SELECTOR = selectors.DefaultSelector
 
-    def patch_wait_selector(self):
+    def setUp(self):
         old_selector = wait.DefaultSelector
         wait.DefaultSelector = self.SELECTOR
         self.addCleanup(setattr, wait, "DefaultSelector", old_selector)
@@ -583,26 +568,21 @@ class BaseWaitForTestCase(unittest.TestCase, TimerMixin, AlarmMixin):
         return s
 
     def test_wait_for_read_single_socket(self):
-        self.patch_wait_selector()
         rd, wr = self.make_socketpair()
         self.assertEqual([], wait.wait_for_read(rd, timeout=SHORT_SELECT))
 
     def test_wait_for_read_multiple_socket(self):
-        self.patch_wait_selector()
         rd, rd2 = self.make_socketpair()
         self.assertEqual([], wait.wait_for_read([rd, rd2], timeout=SHORT_SELECT))
 
     def test_wait_for_read_empty(self):
-        self.patch_wait_selector()
         self.assertEqual([], wait.wait_for_read([], timeout=SHORT_SELECT))
 
     def test_wait_for_write_single_socket(self):
-        self.patch_wait_selector()
         wr, wr2 = self.make_socketpair()
         self.assertEqual([wr], wait.wait_for_write(wr, timeout=SHORT_SELECT))
 
     def test_wait_for_write_multiple_socket(self):
-        self.patch_wait_selector()
         wr, wr2 = self.make_socketpair()
         result = wait.wait_for_write([wr, wr2], timeout=SHORT_SELECT)
         # assertItemsEqual renamed in Python 3.x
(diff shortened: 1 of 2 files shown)


Lukasa:
Thanks! Can we get a test, please? =)

SethMichaelLarson:
Sure I can add a test to ensure no fd leakage.

SethMichaelLarson:
Added the test, seems to be passing on my machine but let's wait for CI. :)

test/test_selectors.py
```diff
@@ -613,6 +612,34 @@ def test_leaking_fds(self):
         after_fds = len(proc.open_files())
         self.assertEqual(before_fds, after_fds)
 
+    @skipIf(sys.platform == "win32", "psutil.Process.open_files() is unstable on Windows.")
+    def test_wait_read_leaking_fds(self):
+        old_selector = wait.DefaultSelector
+        wait.DefaultSelector = self.SELECTOR
```
Lukasa: Is there any reason we need to do this patching?

test/test_selectors.py
```diff
@@ -613,6 +612,34 @@ def test_leaking_fds(self):
...
+        self.addCleanup(setattr, wait, "DefaultSelector", old_selector)
+
+        proc = psutil.Process()
+        rd, wr = self.make_socketpair()
+        before_fds = len(proc.open_files())
+        self.assertEqual(1, len(wait.wait_for_write([wr], 0.001)))
+        after_fds = len(proc.open_files())
+        self.assertEqual(before_fds, after_fds)
```
Lukasa: These tests may be simpler if you just check whether or not `close` got called.

test/test_selectors.py
SethMichaelLarson: Otherwise each selector type won't be tested, only whatever the default is times the number of available selectors.

test/test_selectors.py
SethMichaelLarson: I can add a test for that as well, I figured check the 'bad outcome' we're trying to avoid. :)

test/test_selectors.py
Lukasa: Is there any reason, then, that `wait_for_x` shouldn't be tested across the whole range of selectors, as we do with the selectors module itself?

SethMichaelLarson:
Is this the kind of test you were looking for @Lukasa?

test/test_selectors.py
SethMichaelLarson: That's very true, should I implement that as well?

Lukasa:
Yup, that would be fine.

SethMichaelLarson:
TIL it's really annoying to inject into an object that's being used as a context manager with Mock. I'll remove  the mock import.

SethMichaelLarson:
Now testing each selector available for `wait_for_*` tests as well. :) All the tests!!!

SethMichaelLarson:
This PR might be ready for merge. :) 🎉

test/test_selectors.py
Lukasa: The issue is that they are flaky: if anything else in the test runner manipulates FDs for any reason then they will start to misbehave.

test/test_selectors.py
```diff
@@ -584,7 +512,7 @@ def test_select_interrupt_exception(self):
         class AlarmInterrupt(Exception):
             pass
 
-        def alarm_exception(*args):
+        def alarm_exception(*_):
```
SethMichaelLarson: I can revert it :)

test/test_selectors.py
```diff
@@ -613,6 +541,111 @@ def test_leaking_fds(self):
...
 
+    @skipIf(sys.platform == "win32", "psutil.Process.open_files() is unstable on Windows.")
+    def test_wait_read_leaking_fds(self):
+        self.patch_wait_selector()
+        proc = psutil.Process()
+        rd, wr = self.make_socketpair()
+        before_fds = len(proc.open_files())
+        self.assertEqual(0, len(wait.wait_for_read([rd], 0.001)))
+        after_fds = len(proc.open_files())
+        self.assertEqual(before_fds, after_fds)
```
SethMichaelLarson: There is a test to check whether close was called below this one, I'll remove the fd tests in favor of two tests (`wait_for_read` and `wait_for_write`) for if the selector is properly closed. :)

SethMichaelLarson:
Ready for review again :)

test/test_selectors.py
```diff
@@ -119,86 +118,15 @@ def assertTakesTime(self, lower=None, upper=None):
...
-        rd, wr = self.make_socketpair()
-
-        self.set_alarm(SHORT_SELECT, lambda *args: wr.send(b'x'))
-
-        with self.assertTakesTime(lower=SHORT_SELECT, upper=SHORT_SELECT):
-            self.assertEqual([rd], wait_for_read(rd, timeout=LONG_SELECT))
-        self.assertEqual(rd.recv(1), b'x')
-
-
-@skipUnless(selectors.HAS_SELECT, "Platform doesn't have a selector")
```
Lukasa: So I understand why you moved all this into the `BaseSelectorTestCase`, but I think in this instance I'd rather have some copy-paste code to get the "runs with every selector" behaviour. I quite liked the logical separation of the tests before.

SethMichaelLarson: Sure, I can do this. :)

SethMichaelLarson:
Got an unrelated fail on Python 3.4. This should be ready to be reviewed again :) Happy holidays @Lukasa !

test/test_selectors.py
```diff
@@ -613,6 +536,116 @@ def test_leaking_fds(self):
...
+        def fake_constructor(*args, **kwargs):
+            return selector
+
+        old_selector = wait.DefaultSelector
+        wait.DefaultSelector = fake_constructor
+        self.addCleanup(setattr, wait, "DefaultSelector", old_selector)
+
+        rd, wr = self.make_socketpair()
+        wait.wait_for_write([rd, wr], 0.001)
+        self.assertIs(selector._map, None)
```
Lukasa: Any reason this test isn't in `BaseWaitForTestCase`?

test/test_selectors.py
```diff
@@ -613,6 +536,116 @@ def test_leaking_fds(self):
...
+
+    def test_wait_selector_error(self):
+        err = selectors.SelectorError(1)
+        self.assertEqual(err.__repr__(), "<SelectorError errno=1>")
+        self.assertEqual(err.__str__(), "<SelectorError errno=1>")
```
Lukasa: This test isn't obviously anything to do with `wait`, so we should probably remove it from the name of the test.

test/test_selectors.py
```diff
@@ -613,6 +536,116 @@ def test_leaking_fds(self):
...
+
+
+class BaseWaitForTestCase(unittest.TestCase, TimerMixin, AlarmMixin):
+    SELECTOR = selectors.DefaultSelector
+
+    def patch_wait_selector(self):
+        old_selector = wait.DefaultSelector
+        wait.DefaultSelector = self.SELECTOR
+        self.addCleanup(setattr, wait, "DefaultSelector", old_selector)
```
Lukasa: Given that we always call this, shouldn't this method be a `setUp`?

SethMichaelLarson: Very very true I will change this. :)

test/test_selectors.py
SethMichaelLarson: Because I missed it! I'll move it to where it belongs :)

test/test_selectors.py
SethMichaelLarson: Yeah.... haha I'll do this right now :)

PR #1050, opened 2016-11-23
Changes for 1044

Lukasa:


Lukasa:
Well I see Travis is feeling particularly slow today.

SethMichaelLarson:
@Lukasa Mac OS has been way over-worked as of late on Travis and it doesn't seem like they're adding any new Mac OS workers to the pool... It looks like the tests are now running though!

PR #1052, opened 2016-11-25
dummyserver: raise syn backlog

etbuira:
If the client thread called connect() by the time the server thread
actually accept()ed connections, the client got a connection refused
error.


Lukasa:
This seems reasonable to me. :)

PR #1054, opened 2016-11-26
tests: Q&D skip network dependant tests when network is unavailable

etbuira:
A proper fix would be to bring the socket.{gai,}error in
NewConnectionError, but i'm not willing to undertake this refactoring.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -114,6 +114,10 @@ def requires_network(test):
             raise
     return wrapper
 
+def get_exception_message(e):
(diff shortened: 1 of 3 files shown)


Lukasa:
So I'm *really* nervous about adding these bits of logic.

I'd rather see one of two other fixes. Either rewrite these tests to avoid their needing access to the actual network, or let's test for the network once, explicitly, and then use that to provide a decorator that skips tests that need the network.

PR #1055, opened 2016-11-27
Stop trying 'buffering' arg after TypeError

### LINKED ISSUE #1047: "buffering" arg support is checked on every call to `ConnectionCls.getresponse()` in HTTPConnectionPool
On Python 3.5 with urllib3 1.15, I just hit the issue discussed in https://github.com/kennethreitz/requests/issues/1289#issuecomment-31294851 where urllib3 is checking for `buffering` parameter support first, and then falling back to calling `getresponse()` without any arguments, creating a misleading stack trace when the latter call fails.

Thanks to https://github.com/shazow/urllib3/commit/7fa026c37a5326dcaf97cb9c66ecd15df9c1a194 more recent versions of urllib3 won't show the confusing chained traceback when a connection fails.

However, it's also the case that on every current Python version *except* Python 2.7, the first call will always fail (and even on Python 2.7 it's relying on an undocumented parameter, so not all Python implementations will support it).

This problem could be eliminated by checking for the relevant signature information once at import time rather than on every call to `getresponse`:

    def _request_response_buffering():
        try:
            names = HTTPConnection.getresponse.im_func.func_code.co_varnames
        except AttributeError:
            return False # Either Python 3 or not CPython
        return "buffering" in names

    if _request_response_buffering():
        def _getresponse(conn):
            return conn.getresponse(buffering=True)
    else:
        _getresponse = HTTPConnection.getresponse

And then calling `_getreponse(conn)` instead of `conn.getresponse()`

If such an approach would be acceptable, then I'd be happy to create a PR for it.

ncoghlan:
If `conn.getresponse(buffering=True)` throws TypeError once,
it's always going to throw it on all future calls.

The new HTTPConnectionPool._getresponse helper method added in
this patch instead replaces itself on the instance with the
underlying ConnectionCls method so all future calls on that
particular pool instance will use that instead of the pool
level helper method.

A runtime sanity check is included to detect the case where
a connection that isn't an instance of ConnectionCls is somehow
accessed through the pool.

Closes #1047

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -8,9 +8,9 @@ from urllib3.connectionpool import (
     HTTPConnectionPool,
     HTTPSConnectionPool,
 )
-from urllib3.response import httplib, HTTPResponse
+from urllib3.response import httplib
 from urllib3.util.timeout import Timeout
-from urllib3.packages.six.moves.http_client import HTTPException, HTTPMessage
+from urllib3.packages.six.moves.http_client import HTTPException
 from urllib3.packages.six.moves.queue import Empty
 from urllib3.packages.ssl_match_hostname import CertificateError
 from urllib3.exceptions import (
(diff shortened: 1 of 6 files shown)


test/test_connectionpool.py
```diff
@@ -389,6 +389,96 @@ def _make_request(self, *args, **kwargs):
...
+            ResponseCls = CustomResponse
+
+            def _validate_conn(self, conn):
+                self.ConnectionCls = HTTPConnection
+                return True
+        bad_pool = BadConnectionPool(host='localhost', maxsize=1, block=True)
+        error_structure = 'expected.*HTTPConnection.*got.*CustomConnection'
+        with self.assertRaisesRegex(TypeError, error_structure):
+            bad_pool.request('GET', '/', redirect=False, retries=False)
+
```
Lukasa: This test is extremely complex, mostly because it defines a few helper objects inline which makes them unavailable for reuse in granular tests. It might be better to split this test out to a new file, define the helpers at module scope, and then write a collection of smaller tests.

test/test_connectionpool.py
```diff
@@ -389,6 +389,96 @@ def _make_request(self, *args, **kwargs):
...
+            def _validate_conn(self, conn):
+                return True
+
+        # Set up a custom pool that lets us test the getresponse() handling
+        pool = CustomConnectionPool(host='localhost', maxsize=1, block=True)
+        def _make_new_request(**request_kwds):
+            return pool.request('GET', '/',
+                                redirect=False,
+                                retries=False,
+                                **request_kwds,
```
Lukasa: This is a syntax error.

ncoghlan: Oops, only tested locally on 3.5 (which allows it).

test/test_connectionpool.py
Lukasa: Sounds good to me!

ncoghlan:
Looking at the CI failure, there seems to be something going fairly fundamentally wrong on Python 2.x, so I'll get that working locally and see if it also fixes the problem with Python 3.3.

ncoghlan:
It looks like at least some of the Python 2.7 failures were real incompatibilities relating to the behaviour of old-style classes and unbound methods.

There were also some compatibility issues specifically in the new tests.

I've pushed fixes for those.

ncoghlan:
Oops, in switching over to `unittest2` on 2.x, I inadvertently dropped the plain import of `unittest` on 3.x.

ncoghlan:
The remaining Travis failures and the CodeCov change didn't look related to this patch, so I've rebased on master and pushed the update in order to trigger a fresh CI run.

Lukasa:
So, the Python 3.3 tests are timing out, but I cannot *for the life of me* get them to pass by restarting the build. I wonder if something got severely slowed down, or if Travis is just repeatedly using the same overloaded build machine.

ncoghlan:
Last remaining failure is this on Python 3.3:

```
FAIL: test_timeout (test.with_dummyserver.test_connectionpool.TestConnectionPoolTimeouts)

  File "/home/travis/build/shazow/urllib3/test/with_dummyserver/test_connectionpool.py", line 123, in test_timeout

    self.assertTrue(delta < LONG_TIMEOUT, "timeout was pool-level LONG_TIMEOUT rather than request-level SHORT_TIMEOUT")

nose.proxy.AssertionError: False is not true : timeout was pool-level LONG_TIMEOUT rather than request-level SHORT_TIMEOUT
```

I'm not sure of any way the changes in this PR could be triggering that without also triggering it on other views. Any suggestions for further investigation, @Lukasa?

Lukasa:
Nope, I think it just has to be environmental.

ncoghlan:
Huh, I at least figured out why so many of the tests were failing for me locally: https://github.com/shazow/urllib3/issues/1070

Given that, I can now confirm the tests are passing locally for me under 3.3, with a single failure under 2.7 relating to source address confusion between IPv4 and IPv6.

nateprewitt:
I hit this as well in #1039 where 3.3 and occasionally 3.4 would fail for long periods of time, but only on Travis. I had left a couple [comments](https://github.com/shazow/urllib3/pull/1039#issuecomment-261264193) on what I thought *might* be the problem but never pinned it down. There seems to be something off either with how we're testing the connection pool, or how this specific test is interacting in Travis' environment.

ncoghlan:
I just pushed a commit trying the hacky option of increasing the long timeout in those tests to 20 ms rather than the current 10 ms. If it is just a process scheduling issue in Travis, the extra 10 ms may be enough to make failures even less likely.

sigmavirus24:
Well at least now Python 3.4 is failing too?

ncoghlan:
It occurred to me to try using `assertLess` so we can see what value the tests are actually getting for `delta`.

ncoghlan:
While I don't see why it would be version dependent, I figured it couldn't hurt to try using `time.monotonic()` for the relative time measurements when available. I also fixed the connection pool timeout tests to use `unittest2` on 2.x, so `self.assertLess` will actually be available (I had only tested my previous commit on 3.5).

Lukasa:
🙁 It's a bit frustrating that this patch has gotten so big. I wonder how necessary it will be with the v2.0 work ongoing.

ncoghlan:
Assuming I can actually get the CI tests to pass reliably, it would probably make sense to separate out the test suite adjustments into a separate PR - many of them aren't specifically related to the change, they're about making the timeout tests more reliable and debuggable.

ncoghlan:
Aside from the incidental cross-version errors caused by relying on the `assertLess` method, the new failures show that the tests are seeing deltas in the 25-45 ms bracket, even though the nominal short timeout is only 1 ms.

ncoghlan:
Another timeout investigation experiment pushed, this one based on the `urllib3.connectionpool: DEBUG: Resetting dropped connection: localhost` that appears in the logs of the failed tests.

jonparrott:

PR #1056, opened 2016-11-27
Fix tests where no network is available

etbuira:


test/__init__.py
```diff
@@ -100,18 +100,17 @@ def wrapper(*args, **kwargs):
         msg = "Can't run {name} because the network is unreachable".format(
             name=test.__name__)
         try:
-            return test(*args, **kwargs)
+            sock = socket.create_connection((TARPIT_HOST, 80), 0.0001)
+        except socket.timeout:
+            pass
```
sigmavirus24: If it times out creating a socket, why not use similar code to the other two exceptions? (I'm afraid that without a complimentary bug report explaining why, and without commit messages that explain *why* you'll be answering a lot of *why* questions in this review.)

test/__init__.py
```diff
@@ -100,18 +100,17 @@ def wrapper(*args, **kwargs):
         msg = "Can't run {name} because the network is unreachable".format(
             name=test.__name__)
         try:
-            return test(*args, **kwargs)
+            sock = socket.create_connection((TARPIT_HOST, 80), 0.0001)
```
sigmavirus24: Why are we assigning to sock here but not using it?

test/test_connectionpool.py
```diff
@@ -275,11 +275,6 @@ def test_contextmanager(self):
         self.assertRaises(ClosedPoolError, pool._get_conn)
         self.assertRaises(Empty, old_pool_queue.get, block=False)
 
-    def test_mixed_case_url(self):
-        pool = HTTPConnectionPool('Example.com')
-        response = pool.request('GET', "http://Example.com")
```
sigmavirus24: Why are you moving this test to use the dummy server? Why not have both?

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -7,7 +7,9 @@
 from dummyserver.testcase import HTTPDummyProxyTestCase, IPv6HTTPDummyProxyTestCase
 from dummyserver.server import (
     DEFAULT_CA, DEFAULT_CA_BAD, get_unreachable_address)
-from .. import TARPIT_HOST
+from .. import (TARPIT_HOST,
```
sigmavirus24: This does not follow the style of any of the other imports. Please either do

```py
from .. import TARPIT_HOST, requires_network
```

or

```py
from .. import (
    TARPIT_HOST, requires_network)
```

test/__init__.py
etbuira: Agreed, feel free to add sock.close() afterward.

test/__init__.py
etbuira: Because a timeout means the OS found a route, and a route is what we are looking for here.

test/test_connectionpool.py
etbuira: The commit message reads 'make mixed_case_url TC network independant'

test/with_dummyserver/test_proxy_poolmanager.py
etbuira: It's never too late to adopt a diff-friendly style for selective imports :)

test/test_connectionpool.py
Lukasa: So what's not necessarily entirely clear is that this is a follow-on PR to #1054, where I suggested that rather than dirty up all the tests that currently use the network we should move to make them either use localhost (if possible), or have a single decorator that we can use to skip tests that need the network when the network is unavailable.

test/with_dummyserver/test_proxy_poolmanager.py
Lukasa: Nevertheless, it is better to match the pre-existing style when asked. =)

test/__init__.py
```diff
@@ -100,18 +100,17 @@ def wrapper(*args, **kwargs):
...
             # This test needs an initial network connection to attempt the
             # connection to the TARPIT_HOST. This fails if you are in a place
             # without an Internet connection, so we skip the test in that case.
             if _is_unreachable_err(e):
                 raise SkipTest(msg)
             raise
-        except MaxRetryError as e:
-            if _is_unreachable_err(e.reason):
-                raise SkipTest(msg)
-            raise
```
Lukasa: So, let's ask this question: why was this decorator not *already sufficient* for our cases? What did it not handle correctly?

PR #1057, opened 2016-11-27
Fix tests where no network is available

etbuira:
This follows PR #1056

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -100,7 +100,10 @@ def requires_network(test):
         msg = "Can't run {name} because the network is unreachable".format(
             name=test.__name__)
         try:
-            return test(*args, **kwargs)
+            sock = socket.create_connection((TARPIT_HOST, 80), 0.0001)
+            sock.close()
+        except socket.timeout:
+            pass
         except socket.error as e:
             # This test needs an initial network connection to attempt the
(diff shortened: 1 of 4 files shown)


Lukasa:
Ok, at this point I have to ask you to stop closing PRs and opening new ones. If you simply push new commits, or rebase your current ones, on the same branch the pull request will automatically update. It becomes much easier to track work if you do it that way around. =)

test/__init__.py
```diff
@@ -100,18 +100,18 @@ def wrapper(*args, **kwargs):
...
             # connection to the TARPIT_HOST. This fails if you are in a place
             # without an Internet connection, so we skip the test in that case.
             if _is_unreachable_err(e):
                 raise SkipTest(msg)
             raise
-        except MaxRetryError as e:
-            if _is_unreachable_err(e.reason):
-                raise SkipTest(msg)
-            raise
+        return test(*args, **kwargs)
```
Lukasa: So we haven't answered my question on the previous PR: *why* is the current decorator insufficient?

etbuira: This is explained in the commit message

Lukasa: Ok, so in that case I'd like to change the logic here. I'd like to do this check only once and remember the result. =) Save ourselves from doing this too often.

PR #1061, opened 2016-12-01
work around a problem with an unlucky six/future interaction

megies:
see kennethreitz/requests#3742
see obspy/obspy#1599

### DIFF
--- a/test/test_queue_monkeypatch.py
+++ b/test/test_queue_monkeypatch.py
@@ -1,10 +1,11 @@
 from __future__ import absolute_import
 
+import mock
 import unittest
 
 import urllib3
 from urllib3.exceptions import EmptyPoolError
-import Queue
+from urllib3.packages.six.moves import queue
 
 class BadError(Exception):
     """
@@ -12,17 +13,17 @@ class BadError(Exception):
     """
     pass
 
-Queue.Empty = BadError
 
-
-class TestConnectionPool(unittest.TestCase):
+class TestMonkeypatchResistance(unittest.TestCase):
     """
+    Test that connection pool works even with a monkey patched Queue module,
+    see obspy/obspy#1599, kennethreitz/requests#3742, shazow/urllib3#1061.
     """
     def test_queue_monkeypatching(self):
-        http = urllib3.HTTPConnectionPool(host="localhost", block=True)
-        first_conn = http._get_conn(timeout=1)
-        with self.assertRaises(EmptyPoolError):
-            second_conn = http._get_conn(timeout=1)
+        with mock.patch.object(queue, 'Empty', BadError):
+            http = urllib3.HTTPConnectionPool(host="localhost", block=True)
+            first_conn = http._get_conn(timeout=1)
+            self.assertRaises(EmptyPoolError, http._get_conn, timeout=1)
 
 
 if __name__ == '__main__':
--- a/test/test_queue_monkeypatch.py
+++ b/test/test_queue_monkeypatch.py
@@ -1,0 +1,29 @@
+from __future__ import absolute_import
+
+import unittest
+
+import urllib3
+from urllib3.exceptions import EmptyPoolError
+import Queue
+
+class BadError(Exception):
+    """
+    This should not be raised.
+    """
+    pass
+
+Queue.Empty = BadError
+
+
+class TestConnectionPool(unittest.TestCase):
+    """
+    """
+    def test_queue_monkeypatching(self):
+        http = urllib3.HTTPConnectionPool(host="localhost", block=True)
+        first_conn = http._get_conn(timeout=1)
+        with self.assertRaises(EmptyPoolError):
+            second_conn = http._get_conn(timeout=1)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -25,7 +25,7 @@ from .exceptions import (
 )
 from .packages.ssl_match_hostname import CertificateError
 from .packages import six
-from .packages.six.moves.queue import LifoQueue, Empty, Full
+from .packages.six.moves import queue
 from .connection import (
     port_by_scheme,
     DummyConnection,
@@ -62,7 +62,7 @@ class ConnectionPool(object):
     """
 
     scheme = None
-    QueueCls = LifoQueue
+    QueueCls = queue.LifoQueue
 
     def __init__(self, host, port=None):
         if not host:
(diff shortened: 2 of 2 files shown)


Lukasa:
So I'm inclined to merge this, if for no better reason than problems with people monkeypatching the `queue` module breaking urllib3 have happened a few times in the past. There's no sign that they will *stop* happening, so we should avoid bringing the objects directly into our namespace: it just totally wrecks the exception handling.

@megies It would be fantastic if you could add a regression test for this: it should be sufficient to monkeypatch `queue.Empty` and confirm that some basic connectionpool operations continue to function.

megies:
> @megies It would be fantastic if you could add a regression test for this

Sorry, I've tried to reconstruct this situation in a unit test but couldn't come up with it..

Lukasa:
Did you try monkeypatching `queue.Empty`?

megies:
Yes, I tried, but I couldn't reproduce the situation where the above fix makes a difference..

Lukasa:
This does it:

```python
import urllib3
import Queue

class BadError(Exception):
    """
    This should not be raised.
    """
    pass

Queue.Empty = BadError

http = urllib3.HTTPConnectionPool(host="example.com", block=True)
first_conn = http._get_conn(timeout=1)
second_conn = http._get_conn(timeout=1)
```

megies:
Alright, this works for me locally. When on master it raises the monkey patched `BadError` and when on this branch it raises `EmptyPoolError` for me.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,29 @@
...
+from urllib3.exceptions import EmptyPoolError
+import Queue
+
+class BadError(Exception):
+    """
+    This should not be raised.
+    """
+    pass
+
+Queue.Empty = BadError
```
Lukasa: So this monkeypatch needs to be temporary. We have the `mock` module as a test dependency, so we probably want to use `with mock.patch('Queue.Empty', BadError)` in the test instead, to make sure we don't accidentally break other tests.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import absolute_import
+
+import unittest
+
+import urllib3
+from urllib3.exceptions import EmptyPoolError
+import Queue
```
Lukasa: This is probably going to break on Python 3 because the Queue module moved. You'll need to import it from six.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,29 @@
...
+    This should not be raised.
+    """
+    pass
+
+Queue.Empty = BadError
+
+
+class TestConnectionPool(unittest.TestCase):
+    """
+    """
```
Lukasa: Let's either have no docstring or one that explains what's going on, rather than empty docstring.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,29 @@
...
+class BadError(Exception):
+    """
+    This should not be raised.
+    """
+    pass
+
+Queue.Empty = BadError
+
+
+class TestConnectionPool(unittest.TestCase):
```
Lukasa: I think we could have a better name for this test case here: `class TestMonkeypatchResistance` maybe?

test/test_queue_monkeypatch.py
megies: That's what I tried initially, to import from six and use mock, but I can't get it right, somehow.. the following doesn't raise BadError on unpatched master:
```python

import unittest

import mock

import urllib3
from urllib3.exceptions import EmptyPoolError
from six.moves import queue

class BadError(Exception):
    """ 
    This should not be raised.
    """
    pass


class TestMonkeypatchResistance(unittest.TestCase):
    """ 
    Test that connection pool works even with a monkey patched Queue module,
    see obspy/obspy#1599, kennethreitz/requests#3742, shazow/urllib3#1061.
    """
    def test_queue_monkeypatching(self):
        with mock.patch('queue.Empty', BadError):
            http = urllib3.HTTPConnectionPool(host="localhost", block=True)
            first_conn = http._get_conn(timeout=1)
            with self.assertRaises(EmptyPoolError):
                second_conn = http._get_conn(timeout=1)


if __name__ == '__main__':
    unittest.main()
```

Lukasa: Try changing the first argument to `patch` to `'six.moves.queue.Empty`.

megies: Ah wait, I got it.. need to use `mock.patch.object`, of course. Stupid me..

megies:
This should do it. Hmm, but probably should use `six` from `urllib3.packages` in the test case?

Lukasa:
Yeah, I think so.

megies:
Ok, last changes I hope. For me locally the test case passes on this branch and fails on master.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,32 @@
...
+class TestMonkeypatchResistance(unittest.TestCase):
+    """
+    Test that connection pool works even with a monkey patched Queue module,
+    see obspy/obspy#1599, kennethreitz/requests#3742, shazow/urllib3#1061.
+    """
+    def test_queue_monkeypatching(self):
+        with mock.patch.object(queue, 'Empty', BadError):
+            http = urllib3.HTTPConnectionPool(host="localhost", block=True)
+            first_conn = http._get_conn(timeout=1)
+            with self.assertRaises(EmptyPoolError):
```
Lukasa: I'm afraid the `with` form of this doesn't work with Python 2.6, so you'll need to conditionally import `unittest2` as you can see done in the `selectors` module.

megies: Alright. Not used to Py2.6 compatibility anymore, therefore the oversight. Thanks for the heads up.

sigmavirus24: Given that we don't use `second_conn` why can't we just do:

```py
self.assertRaises(EmptyPoolError, http._get_conn, timeout=1)
```

This will avoid the unittest2 shenanigans altogether

Lukasa: Yeah, that's probably the better thing to do.

megies: :+1: done and force-pushed.

Lukasa:
Fab, assuming all the tests pass I'll merge this. =)

megies:
FYI we can work around this within obspy (obspy/obspy#1612, obspy/obspy#1614), but it might still be good to solve this in requests/urllib3, mostly in case other packages use `future` and requests/urllib3.

test/test_queue_monkeypatch.py
```diff
@@ -0,0 +1,29 @@
...
+from urllib3.packages.six.moves import queue
+
+class BadError(Exception):
+    """
+    This should not be raised.
+    """
+    pass
+
+
+class TestMonkeypatchResistance(unittest.TestCase):
```
Lukasa: You're not importing `unittest` anywhere.

megies: Ah.. oops, didn't notice..

Lukasa:
Thanks @megies!

PR #1063, opened 2016-12-05
Verify that pyOpenSSL is sufficiently new before injecting

reaperhulk:
`cryptography` is being checked, but pyOpenSSL is not, which results in issues like https://github.com/kennethreitz/requests/issues/3701

This approach does require doing an allocation of an `X509` object to verify that the expected private attribute is available.

I also modified the tests as the previous attempt to check `successfully_injected` would not work as it was never set to True (when the assertion failed it would jump immediately into the `finally` block). It isn't unsafe to always call `extract_from_urllib3` so now it always does.

### DIFF
--- a/urllib3/contrib/pyopenssl.py
+++ b/urllib3/contrib/pyopenssl.py
@@ -140,6 +140,14 @@ def _validate_dependencies_met():
         raise ImportError("'cryptography' module missing required functionality.  "
                           "Try upgrading to v1.3.4 or newer.")
 
+    # pyOpenSSL 0.14 and above use cryptography for OpenSSL bindings. The _x509
+    # attribute is only present on those versions.
+    from OpenSSL.crypto import X509
+    x509 = X509()
+    if not hasattr(x509, "_x509"):
+        raise ImportError("'pyOpenSSL' module missing required functionality. "
+                          "Try upgrading to v0.14 or newer.")
+
 
 def _dnsname_to_stdlib(name):
     """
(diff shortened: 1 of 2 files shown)


urllib3/contrib/pyopenssl.py
```diff
@@ -140,6 +140,14 @@ def _validate_dependencies_met():
         raise ImportError("'cryptography' module missing required functionality.  "
                           "Try upgrading to v1.3.4 or newer.")
 
+    # pyOpenSSL 0.14 and above use cryptography for OpenSSL bindings. The _x509
+    # attribute is only present on those versions.
+    from OpenSSL.crypto import X509
+    x509 = X509()
+    if not hasattr(x509, "_x509"):
```
sigmavirus24: Should we instead use `getattr(x509,  '_x509', None) is not None`?

reaperhulk: I always forget why `hasattr` is considered bad practice. Remind me? Also, should I also change the one right above this that does the cryptography check?

sigmavirus24: https://hynek.me/articles/hasattr/ is the reasoning, tl;dr

`hasattr` is slower, invokes side-effects of properties and shadows (hides) them.

And I'd be in favor of fixing up both iff @Lukasa agrees.

Lukasa: Let's do it. =)

Lukasa:
Woo!

rayluo:
@Lukasa Any plan on when to include this patch in an upcoming urllib3 release, and then a requests release?

Lukasa:
@rayluo There is no current plan on this, which means you should expect a lead time of at least a month. Sorry! =(

Issue #1059, opened 2016-11-29
Read timeout causes retry despite HTTP method used


alexdias:
Hello,

We've encountered an issue in which a read timeout will always cause a retry (unless the *read* parameter to the *Retry* class is set to 0), no matter what HTTP method is used to perform the request. Part of the code path that is followed seems to be https://github.com/shazow/urllib3/blob/7002c1347c677b2743244bf36bd05aa3625f57f1/urllib3/connectionpool.py#L642.
To give more context, we're encountering retries on POST requests (which are not indempotent) when a server's response time causes a read timeout.

Is this known behaviour? We were expecting the *method_whitelist* parameter to apply to any type of errors.

Thanks.

Lukasa:
Yup. It looks like this was simply not put in place, or a refactor broke it at some point.

Anyone who would like to can undertake fixing it.

alexdias:
Thanks Cory, I'll see if I can submit a patch in the next few days.

posener:
@Lukasa , hello, seems like read error also ignores the retry method_whitelist attribute. is it intentional?

Lukasa:
@posener I suspect it's the exact same as the timeout problem discussed in this issue: it just got lost.

posener:
Great, @alexdias, are you going to fix this too? Need some help here?
Cheers,
Eyal

alexdias:
@posener Hi Eyal, I've only done the read timeout thus far but I can do the read error tomorrow. Thanks!

posener:
@alexdias , Hi, Just to be clear,
I'm also talking about the read timeout error,
The thing is, that, for example if we currently get a read timeout on a POST method, and the POST method is not in the `method_whitelist`, the retry still make another attempt.
My sense tells that it shouldn't. at least that is what the api that the object exposes suggests.

MANY THANKS!

alexdias:
@posener Ah yes, that is the exact thing I am fixing :)

Lukasa:
Resolved by #1066.

PR #1066, opened 2016-12-06
Don't retry on timeout if method is not in whitelist (#1059)

alexdias:
This should take care of #1059.

I wasn't sure if accessing the _method_whitelist_ attribute of the Retry object was ok, but as it was not named privately I assumed it would be ok (rather than creating a new method for the check).

Let me know if there's anything I should edit here or if there's any comments.

PS: Apologies for the second (empty) commit, Github was down on the exact moment I pushed the first one.

### DIFF
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -225,7 +225,7 @@ class RetryTest(unittest.TestCase):
             self.assertEqual(str(e.reason), 'conntimeout')
 
     def test_history(self):
-        retry = Retry(total=10)
+        retry = Retry(total=10, method_whitelist=frozenset(['GET', 'POST']))
         self.assertEqual(retry.history, tuple())
         connection_error = ConnectTimeoutError('conntimeout')
         retry = retry.increment('GET', '/test1', None, connection_error)
@@ -239,3 +239,13 @@ class RetryTest(unittest.TestCase):
         self.assertEqual(retry.history, (RequestHistory('GET', '/test1', connection_error, None, None),
                                          RequestHistory('POST', '/test2', read_error, None, None),
                                          RequestHistory('GET', '/test3', None, 500, None)))
+
+    def test_retry_method_not_in_whitelist(self):
+        error = ReadTimeoutError(None, "/", "read timed out")
+        retry = Retry()
+        try:
+            retry.increment(method='POST', error=error)
+        except ReadTimeoutError:
+            pass
+        else:
+            self.fail('Request did not raise expected ReadTimeoutError')
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -18,7 +18,7 @@ class RetryTest(unittest.TestCase):
         retry = Retry()
         self.assertEqual(str(retry), 'Retry(total=10, connect=None, read=None, redirect=None)')
         for _ in range(3):
-            retry = retry.increment()
+            retry = retry.increment(method='GET')
         self.assertEqual(str(retry), 'Retry(total=7, connect=None, read=None, redirect=None)')
 
     def test_retry_both_specified(self):
@@ -45,9 +45,9 @@ class RetryTest(unittest.TestCase):
         """ A lower read timeout than the total is honored """
         error = ReadTimeoutError(None, "/", "read timed out")
         retry = Retry(read=2, total=3)
-        retry = retry.increment(error=error)
-        retry = retry.increment(error=error)
-        self.assertRaises(MaxRetryError, retry.increment, error=error)
+        retry = retry.increment(method='GET', error=error)
+        retry = retry.increment(method='GET', error=error)
+        self.assertRaises(MaxRetryError, retry.increment, method='GET', error=error)
 
     def test_retry_total_none(self):
         """ if Total is none, connect error should take precedence """
@@ -63,9 +63,9 @@ class RetryTest(unittest.TestCase):
 
         error = ReadTimeoutError(None, "/", "read timed out")
         retry = Retry(connect=2, total=None)
-        retry = retry.increment(error=error)
-        retry = retry.increment(error=error)
-        retry = retry.increment(error=error)
+        retry = retry.increment(method='GET', error=error)
+        retry = retry.increment(method='GET', error=error)
+        retry = retry.increment(method='GET', error=error)
         self.assertFalse(retry.is_exhausted())
 
     def test_retry_default(self):
@@ -93,7 +93,7 @@ class RetryTest(unittest.TestCase):
         error = ReadTimeoutError(None, "/", "read timed out")
         retry = Retry(read=0)
         try:
-            retry.increment(error=error)
+            retry.increment(method='GET', error=error)
             self.fail("Failed to raise error.")
         except MaxRetryError as e:
             self.assertEqual(e.reason, error)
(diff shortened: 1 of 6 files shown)


Lukasa:
@alexdias It'd be fab if you could force push that empty commit away. 😊

urllib3/connectionpool.py
```diff
@@ -648,6 +648,9 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
                 e = ProxyError('Cannot connect to proxy.', e)
             elif isinstance(e, (SocketError, HTTPException)):
                 e = ProtocolError('Connection aborted.', e)
+            elif isinstance(e, TimeoutError):
+                if not retries.method_whitelist or method.upper() not in retries.method_whitelist:
+                    raise e
```
Lukasa: I think that you should actually aim to put this inside `retries.increment`. Generally speaking that seems to have most of the relevant logic.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -166,6 +166,68 @@ def test_connect_timeout(self):
...
+        delta = time.time() - now
+
+        self.assertTrue(delta < LONG_TIMEOUT, "timeout was pool-level LONG_TIMEOUT rather than request-level SHORT_TIMEOUT")
+        block_event.set() # Release request
+
+        # Timeout int/float passed directly to request and _make_request should
+        # raise a request timeout
+        wait_for_socket(ready_event)
+        self.assertRaises(ReadTimeoutError, pool.request, 'POST', '/', timeout=SHORT_TIMEOUT)
+        block_event.set() # Release request
```
Lukasa: Does this test add anything new? I feel like it's a replica of a preexisting test.

urllib3/connectionpool.py
alexdias: @Lukasa Ah yes, there's a check (right inside `retries.increment`) in which it re-raises the error; I'll create a similar one but with the logic in the diff above.

test/with_dummyserver/test_connectionpool.py
alexdias: @Lukasa Do you mean the whole test or this specific assert on line 199? The whole test is very similar to `test_timeout`, but it differs in that it passes `retries=True` to the pool class, and uses POST. Thus it tests that it does not retry on POST (which means it should raise a ReadTimeoutError). Do you think it should be more explicit, such as for instance having a specific method whitelist and Retry object?

sigmavirus24: @alexdias honestly, I don't think you need most of this test. Everything after L181 seems to be covered by other tests and the behaviour it's testing (using the request timeout value instead of pool value) is tested elsewhere and irrelevant to your particular bug fix. You can likely trim the test, but I'd like to know if the test fails for you before you applied your patch.

alexdias: @sigmavirus24 Trimming it sounds good.

Removing the patch on connectionpool.py, here's part of the output for `make test`:
`
ERROR: test_timeout_method_not_in_whitelist (test.with_dummyserver.test_connectionpool.TestConnectionPoolTimeouts)
  File "/vagrant/urllib3/test/with_dummyserver/test_connectionpool.py", line 180, in test_timeout_method_not_in_whitelist
    self.assertRaises(ReadTimeoutError, pool.request, 'POST', '/')
  File "/opt/smarkets/python/3.5.1/lib/python3.5/unittest/case.py", line 727, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/opt/smarkets/python/3.5.1/lib/python3.5/unittest/case.py", line 176, in handle
    callable_obj(*args, **kwargs)
  File "/vagrant/urllib3/urllib3/request.py", line 70, in request
    **urlopen_kw)
  File "/vagrant/urllib3/urllib3/request.py", line 148, in request_encode_body
  File "/vagrant/urllib3/urllib3/connectionpool.py", line 682, in urlopen
    **response_kw)
  File "/vagrant/urllib3/urllib3/connectionpool.py", line 653, in urlopen
  File "/vagrant/urllib3/urllib3/util/retry.py", line 363, in increment
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=34364): Max retries exceeded with url: / (Caused by ReadTimeoutError("HTTPConnectionPool(host='localhost', port=34364): Read timed out. (read timeout=0.001)",))
-------------------- >> begin captured logging << --------------------
urllib3.util.retry: DEBUG: Converted retries value: True -> Retry(total=True, connect=None, read=None, redirect=None)
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): localhost
urllib3.util.retry: DEBUG: Incremented Retry for (url='/'): Retry(total=0, connect=None, read=None, redirect=None)
urllib3.connectionpool: WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None)) after connection broken by 'ReadTimeoutError("HTTPConnectionPool(host='localhost', port=34364): Read timed out. (read timeout=0.001)",)': /
urllib3.connectionpool: DEBUG: Starting new HTTP connection (2): localhost
--------------------- >> end captured logging << ---------------------
`

Would you prefer if I caught the MaxRetryError and explicitly failed the test, instead of letting it fail due to the unexpected raise?

Lukasa: I think if you're expecting the error (which it seems like you are) then you should aim to explicitly catch it, and fail if it's not appropriate.

alexdias:
@Lukasa @sigmavirus24 Hey guys, I've made a few changes, do you mind taking a look? Thanks!

Lukasa:
@alexdias If you want to force a read timeout, don't use the dummyserver. Instead, write a socketlevel test: see `test_socketlevel.py` for examples.

test/test_retry.py
```diff
@@ -239,3 +239,13 @@ def test_history(self):
...
+
+    def test_retry_method_not_in_whitelist(self):
+        error = ReadTimeoutError(None, "/", "read timed out")
+        retry = Retry()
+        try:
+            retry.increment(method='POST', error=error)
+        except ReadTimeoutError:
+            pass
+        else:
+            self.fail('Request did not raise expected ReadTimeoutError')
```
Lukasa: This whole block can just be replaced with `self.assertRaises(ReadTimeoutError, retry.increment, method="POST", error=error)`

alexdias: @Lukasa That's much better, thanks! I'm not sure if I have time today to change this and write the other test, so I might only get to this next week as I'm away for the next few days.

SethMichaelLarson: If you allow maintainers to write to your branch this could get done while you're away. :)

alexdias: @SethMichaelLarson That setting is enabled I believe :)

Lukasa: There is also no rush on this: urllib3 isn't going anywhere in the next 7 days. ;)

alexdias:
@Lukasa I've made some changes, could I have another review please?

Thanks!

urllib3/util/retry.py
```diff
@@ -330,7 +338,7 @@ def increment(self, method=None, url=None, response=None, error=None,
 
         elif error and self._is_read_error(error):
             # Read retry?
-            if read is False:
+            if read is False or (method is not None and not self._is_method_retryable(method)):
```
Lukasa: Hrm. In what circumstance do we have `method=None` here?

alexdias: @Lukasa I don't think it's actually `None` anywhere (except some tests), but the default value for the argument is None (which I assumed meant it was a valid value to pass) so I added that check.

Lukasa: Hrm. I don't think I think we should allow `None` here: it's not clear to me that it's every actually going to be passed. You raise a fair point about the default value though. I'm inclined to say let's just remove the check.

alexdias: @Lukasa That seems ok to me! I'm thinking of just adding `method='GET'` (or equivalent) to the tests that need it, is that what you have in mind?

Lukasa: Yeah, I think so. =)

alexdias: @Lukasa Given I'm at it, should I just add the method value to all uses of `retry.increment` in the tests (I currently have only added them to the tests that use a read error)?

Lukasa: Almost certainly, yes: I suspect it's relevant. ;)

Lukasa: Ah, actually, there is one kind of test that isn't relevant: connection timeouts, or any error during connection setup (e.g. TLS).

So I'd only do it if it's retrying something that isn't occurring during connection setup.

alexdias:
@Lukasa Could you take another look please? :)

I wasn't sure about passing the method argument on tests that for instance only tested the string representation or the backoff factor, but I guess it's consistent if it's always passed except for tests with connection setup errors.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -152,6 +152,24 @@ def socket_handler(listener):
 
         self.assertEqual(http.pool.qsize(), http.pool.maxsize)
 
+    def test_read_timeout_dont_retry_method_not_in_whitelist(self):
+        timed_out = Event()
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+            while not sock.recv(65536):
```
Lukasa: Hrm. This is a slightly odd construction. This loop will only execute once, and in the event the connection comes up and is then closed it'll loop a second time and then IOError. What were you aiming for here?

urllib3/util/retry.py
```diff
@@ -273,10 +273,18 @@ def _is_read_error(self, err):
...
+    def _is_method_retryable(self, method):
+        """ Is this method retryable? (Based on method whitelist)
+        """
+        if self.method_whitelist and method.upper() not in self.method_whitelist:
+            return False
+
+        return True
+
     def is_retry(self, method, status_code, has_retry_after=False):
         """ Is this method/status code retryable? (Based on method/codes whitelists)
```
Lukasa: We're going to need to clear up the fact that these two methods have very similar docstrings.

alexdias: @Lukasa True, how about for `_is_method_retryable`: """ Checks if a given HTTP method should be retried upon, depending if it is included on the method whitelist."""

And for `is_retry`: """ Is this method/status code retryable? (Based on whitelists)"""

I could also change the one for `is_retry` to include the fact that it also checks variables such as `total`, `respect_retry_after_header`, and `has_retry_after`.

test/with_dummyserver/test_socketlevel.py
Lukasa: Yeah, a single `recv` is almost certainly fine: just enough to ensure the request doesn't get blocked behind the socket queue.

urllib3/util/retry.py
Lukasa: Yeah, I think all of these ideas are good, tbh.

alexdias:
@Lukasa Could I have a review please? Thanks!

Lukasa:
@alexdias Yup, sure thing. An FYI: I'm now technically on holiday, so my latency between reviews will have gone up a bit. =(

alexdias:
No worries :) Enjoy your holidays!

alexdias:
@Lukasa No problem, thanks for the reviews!

Issue #776, opened 2016-01-05
v2.0 Wishlist


shazow:
Not sure we'll do this, but if we ever get a v2.0 release, some major changes that would be nice to see:

- Deprecate outdated APIs
  - `pool.urlopen(...)`
  - `response.data` (in favour of `.read()`)
  - Default `response.read(preload_content=False)`
  - others?
- Request history
  - We'll need a pre-configurable request object (instead of using a few dozen keyword params).
- Autoconfig for AppEngine/Proxy settings/SSL (certifi)/others?
  - We'll need a Session-like class in front of PoolManager that is more implicit. Ideally any new feature we add which can get auto-configured based on env should happen magically here.

shazow:
Any other ideas/requests? @Lukasa @sigmavirus24 @dstufft (feel free to cc others for feedback)

Lukasa:
I think for my part I'd like a declarative TLS API: something that expresses what you want in abstract terms and maps it to a real configuration.

How about, if we're going really "pie in the sky", I add: asynchronous by default? ;)

shazow:
> How about, if we're going really "pie in the sky", I add: asynchronous by default? ;)

I'll add to the long-shot list. Any ideas on what would need to change for this to be possible/what it would look like?

ml31415:
About this httplib replacement. As I had a look into this lately, I'd say replacing it wouldn't be too complicated. In the end, it's only three classes in there. As a starter, it would be enough to copy the imported parts from httplib, make it compatible with six, and then start simplifying and refactoring under the hood, step by step. I saw the header management has some overlap and could be simplifyed. You probably find some other spots. There is also a bunch of weird stuff from ancient python times in there, like using a dict without setting values, instead of a set etc. 

All in all, the httplib replacement would sound more like a WIP thingy for me, as I guess it wouldn't break any public API. The question remains why? In the end httplib is in the standard library. So even if it's a bit ugly itself, I'm not sure if it's worth reinventing the wheel.

shazow:
@ml31415 There is tons of code in urllib3 working around httplib's shortcomings, things like mangling headers and bugs in the state machine and others. Especially differences in the py2 vs py3 versions. I wouldn't want to necessarily fork/rewrite httplib as a vendored lib, but moreso switch to a different implementation of the parser/state machine (maybe even @Lukasa's hyper, when it's ready).

jonparrott:
> We'll need a Session-like class in front of PoolManager that is more implicit.

[...]

PR #1067, opened 2016-12-08
[WIP] Remove httplib.

Lukasa:
This branch contains the most substantial chunk of work I have ever done on urllib3: the beginnings of an effort to remove our dependency on httplib.

Removing our dependency on httplib has been a long-held desire of the urllib3 project. In fact, it's so desired that #776 (our wishlist of features for v2.0) includes in a section entitled "unlikely longshots" the phrase "Get rid of httplib dependence".

This branch does exactly that, by bringing the `http.client` implementation from Python 3.5 that we were already used to interacting with into our codebase and then ripping its guts out and replacing them with @njsmith's [h11](https://github.com/njsmith/h11) library.

As you can see from the commit log, my work on this began back in October, but I've sat on it for the past two months in order to get it to a place where I had something worth showing. Now I finally do. On my machine, a run of `tox -e py27` passes all tests, albeit with incomplete test coverage:

```
py27 runtests: commands[1] | nosetests
SSS....................S....................................................................................................................................S.....................................................S........................................................................................................................................................................................................................SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS...............................S..S..............................S................................................................................
Name                      Stmts   Miss  Cover   Missing
-------------------------------------------------------
urllib3                      20      0   100%   
urllib3._collections        148      0   100%   
urllib3.connection          515    103    80%   80-82, 158, 162, 175, 229, 261, 285, 295-300, 306-338, 341-364, 368-372, 375, 385-392, 396-399, 408, 449, 453, 471, 567, 571-572, 602-611, 615, 651-652, 657, 729-732, 736-740, 755, 781, 830, 855-856, 868-869, 905-906, 972-973, 985, 1005
urllib3.connectionpool      240      1    99%   256
urllib3.contrib               0      0   100%   
urllib3.contrib.socks        35      0   100%   
urllib3.exceptions           73      2    97%   216, 219
urllib3.fields               62      0   100%   
urllib3.filepost             32      0   100%   
urllib3.poolmanager         124      0   100%   
urllib3.request              32      0   100%   
urllib3.response            232      1    99%   453
urllib3.util                  1      0   100%   
urllib3.util.connection      47      0   100%   
urllib3.util.request         37      0   100%   
urllib3.util.response        19      5    74%   54-65
urllib3.util.retry          121      0   100%   
urllib3.util.selectors      295     45    85%   101, 360-423
urllib3.util.ssl_            64      0   100%   
urllib3.util.timeout         48      0   100%   
urllib3.util.url             97      0   100%   
urllib3.util.wait            14      1    93%   21
-------------------------------------------------------
TOTAL                      2256    158    93%   
Ran 604 tests in 13.646s

OK (SKIP=42)
___________________________________ summary ____________________________________
  py27: commands succeeded
  congratulations :)
```

Initially, my goal was to do this in the least-disruptive way possible: that's why I began by trying to replicate the logic of httplib inside urllib3 but on top of h11. I would have succeeded in that goal if I'd been able to avoid rewriting any tests in any substantive way.

Unfortunately, as you can see from this change, I did not succeed at that. In particular, it was simply not possible to continue with some behaviours that urllib3 had previously promised, such as maintaining header case. I also had to bend over backwards to preserve some other behaviours urllib3 users have expected, such as reporting chunk boundaries, which is probably not something we should ever have promised to do.

From my perspective, then, I think that if we want to continue this work it needs to be considered a breaking change for urllib3. Any change this substantial simply cannot be done in a non-breaking way. A huge number of things will change: our tolerance of errors (we're much stricter now because h11 is much more insistent about being spec-compliant), our wire format (h11 does some case normalization on header field names), and maybe even some of our output formats.

This is also still very much a work in progress. There are TODO statements everywhere. The code is poorly factored: it's a wacky in-between state between trying to maintain the httplib abstraction and a total rejection of it in favour of something much clearer. A whole lot of our code that worked around httplib flaws is no longer necessary and can be removed, but hasn't yet been. A better underlying implementation that doesn't perform socket writes would also be possible, as would some refactoring of code I wrote to pull out common features. And the Python 3 tests don't pass, mostly because of the fact that right now this new branch emits headers as bytestrings rather than unicode strings.

However, at this point it's no longer feasible for me to work on this branch alone. Further progress requires actual governance decisions, and I am not qualified to make those on my own, nor should I. urllib3 has a great collection of core maintainers and regular contributors whose opinions and work I value, and would like to solicit to help me. As I see it, we need to answer the following questions:

1. Is this worth doing? Do we want to be rid of httplib so badly that we will perform a *massive* codebase rewrite to get rid of it?
2. If so, are we happy with using h11?
    1. If we are, do we depend on it using pip, or vendor it? How does our decision affect Requests?
3. Should this be the beginning of a v2.0 branch that will contain other bits of work from #776?
    1. If it should, how many breakages do we want to roll into one change? Do we want to go all in and leave a *massive* codebase change?
4. What pieces of functionality do we want to retain from the old codebase? In particular, do we want:
    1. To still have `stream()` expose chunk boundaries?
    2. To return `str` header field names and values on Python 3?
    3. To return *different types* for header field names and values on Python 3?
    4. To implement buffered I/O in the low-level primitives?
    5. To maintain the weird two-level Response structure where we have a urllib3 `Response` object that contains a httplib `Response` object within it?
5. How would we want to manage a rollout of this functionality?
6. How do we want to review it?
7. What kind of timeline do we want for releasing it?
8. If we decide to go ahead with a v2.0, should we freeze the current master branch for everything except bug/security fixes until we complete v2.0?

When @shazow originally made me interim lead maintainer, he told me that I had complete authority because he could always undo any change I made that he didn't like. I think this particular change is an exception: I don't think I have any degree of "lead maintainer" authority on this. We're going to need @shazow to express some opinions too.

While I'm here, I'd also like to tag a few community members that are particularly likely to be interested in this proposal:

/cc @njsmith @durin42 @dstufft


SethMichaelLarson:
When we move to 2.0 can't we make a lot of decisions about what APIs we expose? If we increment the version number we can get away with changing a few parts of the interface if it means a better integration with h11.

Lukasa:
@SethMichaelLarson So, yes, we can. But the more things we change the more inertia there is to avoid upgrading to 2.0. So we need to balance our freedom to break with our desire to make upgrading as smooth as possible.

SethMichaelLarson:
@Lukasa I'm finding myself agreeing with a lot of what you're saying. I think a branch freeze while 2.0 is being worked on is a good idea.  Don't take the route of Bootstrap and try to maintain two completely different branches.  Focus on moving forward. :)

shazow:
> When @shazow originally made me interim lead maintainer, he told me that I had complete authority because he could always undo any change I made that he didn't like. 

That's not true. I said you had complete authority because I trust you. I had no plans to undo any changes just because I disagreed. :)

For urllib3 v2.0, I'm happy with making all the breaking changes we want especially if this means simplifying the codebase and user flows.

sigmavirus24:
> Don't take the route of Bootstrap and try to maintain two completely different branches. Focus on moving forward. :)

I don't think anyone is suggesting that. Everyone here has enough experience to know that we're all too overcomitted to do that. On top of that, a feature freeze on master is fine with me. What's unacceptable is a bug fix freeze.

Lukasa:
Yup, I have no interest in doing a bugfix freeze. Bug and security fixes would continue to be accepted to the v1 tree until v2 was released, and we'd want to consider security fixes for v1 even into the lifetime of v2.

SethMichaelLarson:
If this is all starting to happen now should the larger PRs that I have open currently be saved for v2?

Lukasa:
Right now I don't think we have anything like a consensus view on what's actually happening here. I think we need to take a few days to think about how we'd like this to go so that everyone can get a handle on how to do this.

I have been thinking about this for two months and I'm extremely unsure about how we should progress, so let's *for the moment* continue to treat `master` as an active development branch until such time as we've committed to doing a v2.

However, we should make sure that anything that reaches into or relies on httplib internals is treated with a grain of salt: they may want to go away over time.

dstufft:
I guess my biggest question is, have we asked @njsmith if he's OK with h11 more or less getting installed on every single Python install and used by like every single Python developer? I suspect the set of Python developers who do not directly or indirectly use one of {pip, requests, urllib3} is very small and that tends to mean that rare edge cases become common edge cases and increases the "cost" of maintainership.

Lukasa:
Another very good question to add the pile. :smile:

SethMichaelLarson:
I'm sure if we became an h11 upstream it would be easier to answer questions and fix things compared to the standard library httplib.  Our involvement would help him, he wouldn't be alone hopefully.

Lukasa:
I should note that there's an open invitation to bring h11 into the Hyper project should @njsmith ever want to do that. That's very much his call though.

dstufft:
Also h11 currently only supports 2.7 and 3.3+. For my own personal use cases of pip, we're going to be killing 2.6 support in the next major release so that's not a big deal. I don't know if urllib3/requests plan on dropping 2.6 when pip does or not, but it's another consideration (also whether @njsmith is OK with not being able to drop older versions as readily as he otherwise could).

Lukasa:
I think if we went to v2 we'd drop 2.6 support.

shazow:
Some Q&A&Q:

> Is this worth doing? Do we want to be rid of httplib so badly that we will perform a massive codebase rewrite to get rid of it?

How much work do you estimate remains here? What is going to be the hardest part?

> If so, are we happy with using h11?

If h11 is actively maintained (seems like it) and brings a lot of quality to the table (also seems like it), then I'm +1. Also what @dstufft said, would love to hear what @njsmith thinks. :)

> If it should, how many breakages do we want to roll into one change? Do we want to go all in and leave a massive codebase change?

Ideally v2.0 should be The Big Breaking change, but once we start getting a better idea of what will break then we can start deprecating things in 1.x ahead of the release if that makes us feel better.

> What pieces of functionality do we want to retain from the old codebase? In particular, do we want:
> To still have stream() expose chunk boundaries?

What does this mean? Does it split stream blocks based on chunks? If so, I think we can live without it.

> To return str header field names and values on Python 3?
> To return different types for header field names and values on Python 3?
> To implement buffered I/O in the low-level primitives?

Don't think these are very important, maybe nice-to-have at best. I'm comfortable with launching v2.0 without a lot of little things like this and add them back later if there's still demand. It's always hard to know which features are actually needed after the fact without the opportunity to remove them. :)

> To maintain the weird two-level Response structure where we have a urllib3 Response object that contains a httplib Response object within it?

Would love to get rid of this if we can.

> How would we want to manage a rollout of this functionality?
> How do we want to review it?
> What kind of timeline do we want for releasing it?

I think this is a "person writing the code gets to choose" situation.

Would be ideal if we can get a few committed "customers" to help beta-test v2.0 as we get close (hi @dstufft).

> If we decide to go ahead with a v2.0, should we freeze the current master branch for everything except bug/security fixes until we complete v2.0?

Sounds good.

Additional questions:

How does this relate to adding http/2 support?

How do we ensure that we make the most of this massive refactor opportunity to make all the changes we need to support http/2 mechanics in the future?

SethMichaelLarson:
Dropping Python 2.x is on the v2 wishlist currently, we don't have to drop 2.7 but I think dropping 2.6 is reasonable for v2.

durin42:
For hg, we could probably use httplib on 2.6 until we drop that (haven't
even started that discussion, sigh), but 2.7 will be a must for us to be
interested for something like 5 years or so (guessing).

shazow:
> Dropping Python 2.x is on the v2 wishlist currently, we don't have to drop 2.7 but I think dropping 2.6 is reasonable for v2.

I would go as far as to say: I'd support working on the v2 branch in pure Py3, then backporting it to Py2 last minute if we still think it's a good idea.

SethMichaelLarson:
@shazow That sounds like a lot more work than having Python 2.x in mind during v2 development.

shazow:
@SethMichaelLarson I'm worried that Py2's limitations will inform and restrict our design decisions too much, especially with many iterations that happen with these kinds of big endeavours. I'd prefer for urllib3 to take full advantage of Py3 and be somewhat limited in Py2 rather than the other way around.

SethMichaelLarson:
@shazow That's definitely true, I know you're a big advocate for Python 3.x only and I feel the exact same way about dropping 2.x, I just don't want us to undertake this journey and then in or near the end decide to roll back and support 2.7

dstufft:
> Would be ideal if we can get a few committed "customers" to help beta-test v2.0 as we get close (hi @dstufft).

Absolutely, it shouldn't be hard to, at a minimum, have a PR to pip that uses it and we could possibly even cut a beta release ourselves and publish it to PyPI to get more testing.

> I would go as far as to say: I'd support working on the v2 branch in pure Py3, then backporting it to Py2 last minute if we still think it's a good idea.

I would say that like @durin42, dropping 2.7 for pip is currently not going to be feasible for at least several years. I would not be super thrilled if I ended up needing two copies of urllib3 inside of pip, one that is older for Python 2.7 and a newer one for 3.x+. I don't have much of an opinion on whether to design it for 3.x and then adjust it to work on 2.x at the end of have it be 2.x + 3.x the whole time, but if the final result doesn't support 2.7 It'll make pip's life significantly harder.

haikuginger:
@shazow @SethMichaelLarson, what does dropping 2.x get us that dropping 2.6 alone doesn't get us, aside from the obvious removal of bytes/string ambiguity?

shazow:
@SethMichaelLarson What do you mean by roll back? The worst case I'm envisioning is we finish the urllib3 rewrite in Py3, and we ended up using a lot of cool Py3 features and we decide it's not worth backporting it to Py2, so Py3 users end up getting the cool new version of urllib3 and Python 2 users are stuck with urllib3 v1.x until they upgrade to Py3.

More realistic case is the Py2 version ends up being more limited than the Py3 version.

But speaking from personal experience, making sure everything always works in Py2 and Py3 is a _huge_ overhead during development. I'd estimate maybe as much as half of the overall effort goes into that.

SethMichaelLarson:
I meant roll back on our decision, not anything code related. :) I'd love for v2 to be Python 3.x only, it just makes life harder for our upstreams.  I'm aware that dual support makes things much harder than going with one language.

Because I'm not a contributor to any upstream I'm +0.5 on Python 3.x only, I'd love to have a 3.x-only urllib3.  I would do as much as I could to help regardless of that decision. :)

Lukasa:
Ah, thanks for chiming in @durin42, that reminds me: a nice carrot for this work would be that if we can land it then along with @SethMichaelLarson's selectors work it enables us to support 100-Continue gracefully, which gives mercurial the opportunity to use something better than httplib and helps us continue urllib3's ongoing world domination by getting it into mercurial as well. A nice advantage!

As to @shazow's questions: 

> How much work do you estimate remains here? What is going to be the hardest part?

A bunch. I haven't audited the tests to confirm that they're actually testing what they think they are. We also have to make decisions about header field names/values and types, which probably requires a bunch of work. Then also cleanup, logical improvements, decisions about vendoring/dependency. This is all a big chunk of stuff.

Happily, it isn't all going to sit on me. Some of the core changes should probably rest with me because I've been closest to the work, but other people can definitely get involved in some of the other stuff, and also consider interactions with other v2 features.

TL;DR: I think there's a few months work on this, but happily I have that time and currently consider this to be my first priority.

> What does this mean? Does it split stream blocks based on chunks? If so, I think we can live without it.

Yeah. Right now I bullied @njsmith into adding some extra data to allow us to do that, but I agree with him that it's dumb, and it also doesn't work with buffering, so my preferred approach would be to just lose that functionality entirely.

> > To return str header field names and values on Python 3?
> >To return different types for header field names and values on Python 3?
> > To implement buffered I/O in the low-level primitives?
>
> Don't think these are very important, maybe nice-to-have at best. I'm comfortable with launching v2.0 without a lot of little things like this and add them back later if there's still demand. It's always hard to know which features are actually needed after the fact without the opportunity to remove them. :)

I'm inclined to want to say that we should drop all of them: keep header field names/values as bytes by default in v2.0 (which is definitely the *right* thing to do) and then let user complaints guide us for v2.1 onwards.

> > To maintain the weird two-level Response structure where we have a urllib3 Response object that contains a httplib Response object within it?
> 
> Would love to get rid of this if we can.

Deal.

> How does this relate to adding http/2 support?
> 
> How do we ensure that we make the most of this massive refactor opportunity to make all the changes we need to support http/2 mechanics in the future?

Ah, yes, the elephant in the room. I'd want to take this opportunity to add a third question: how do we open the door for an easy adoption of async semantics in the future? This is because these two questions are pretty heavily related. http2 is pretty gross unless you have some degree of asynchrony in your code, because it needs to be able to react quite quickly to changes in connection state.

That would give us a few options:

1. Have urllib3 change to using threads internally, and move all I/O to background threads. We can do this: the backport of the selectors module would even make it kinda graceful. But ultimately I don't know it's a good idea: threads in Python can be a bit gross. Worth considering though.

2. Have urllib3 do a substantial refactor to start passing around Request and Response objects. These objects get built by the outer layers and then passed into the `PoolManager`, which then does one call into a `ConnectionPool` to get a `Connection`, and then one call into a `Connection` with the `Request` object to get a `Response` object.

    This second mechanism, ultimately, is about enforcing our abstractions internally. That makes it much easier to then retrofit the HTTP/2 implementation into those abstractions. Simultaneously it reduces the number of points in the codebase that know about I/O or that call anything that knows about I/O, which opens the door to support asynchrony down the road.

Further discussion on these points is definitely worth having.

SethMichaelLarson:
@Lukasa Glad my selectors backport is going to be put to good use. :)  Option 2 is the route I went with stormhttp and seemed to work somewhat well.

sigmavirus24:
> Yeah. Right now I bullied @njsmith into adding some extra data to allow us to do that, but I agree with him that it's dumb, and it also doesn't work with buffering, so my preferred approach would be to just lose that functionality entirely.

Didn't we just bend over backwards to add that for some people? Are we going to remove allowing people to view each chunk again?

Lukasa:
We did. I think we shouldn't have, in hindsight. ;)

shazow:
> @lukasa: That would give us a few options: ...

I've been scarred by threads-in-libraries so I'm inherently unfavourable to that option.

Adding boundary objects that describe Requests and Responses is something we've wanted for a long time (also helps with request/response histories that people have wanted), so that makes sense. And I agree it's a natural place to maintain this compatibility layer.

What about async/await in Py3?

sigmavirus24:
Also that two-level Response hierarchy is what Requests relies on for cookies. That said, I'm happy with us figuring out shims to make that work in Requests because I'd love to tear out all of Requests cookie nonsense.

SethMichaelLarson:
@shazow If we make our library as IO-agnostic as possible then it would be effortless for users of the library to do their own async/await.

Lukasa:
> What about async/await in Py3?

If we do a big refactor to have a very small number of places do I/O related things, then we should be able to write equivalent versions of those that use, say, `asyncio`. That would allow us to have *two* APIs: one synchronous, one async, which share the majority of their code. We can conditionally install the async one only on Python 3, and everything should be fine.

Most importantly, though, this allows us to ship a synchronous-only v2 and then slowly write the async functions to add into a neat future release. It avoids us doing too much boiling of the ocean in this change.

> Also that two-level Response hierarchy is what Requests relies on for cookies. That said, I'm happy with us figuring out shims to make that work in Requests because I'd love to tear out all of Requests cookie nonsense.

We'll probably have to, because ultimately using cookielib requires things that look like what httplib provides, and we're planning to give up on that. ;)

nateprewitt:
> Have urllib3 do a substantial refactor to start passing around Request and Response objects.

While this may be a lot of additional work, I think having this abstraction layer will simplify a lot of the internals for urllib3. Especially if it allows us to support async while avoiding having any substantial threading done inside of urllib3.

sigmavirus24:
> We'll probably have to, because ultimately using cookielib requires things that look like what httplib provides, and we're planning to give up on that. ;)

Excellent. I'll move forward with the work on the implementation of cookie semantics we'd talked about a while back then. :)

haikuginger:
My answers to Corey's questions:

1. >Is this worth doing? Do we want to be rid of httplib so badly that we will perform a *massive* codebase rewrite to get rid of it?
    
    I think it is, as part of a sprint towards urllib3 2.0.
2. >If so, are we happy with using h11?

    I'd need to read through it, but a perusal of the readme shows the author is coming from a good place.
    1. >If we are, do we depend on it using pip, or vendor it? How does our decision affect Requests?

        I'd tend towards vendoring.
3. >Should this be the beginning of a v2.0 branch that will contain other bits of work from #776?

    Yes.
    1. > If it should, how many breakages do we want to roll into one change? Do we want to go all in and leave a *massive* codebase change?

    I think it's important to get all the things we _really want_ into 2.0. However, I also think that we should try to preserve the current API to the extent possible.
4. > What pieces of functionality do we want to retain from the old codebase? In particular, do we want:
    1. > To still have `stream()` expose chunk boundaries?
        Not familiar with this use case.
    2. > To return `str` header field names and values on Python 3?
        No opinion on this.
    3. > To return *different types* for header field names and values on Python 3?
        See above.
    4. > To implement buffered I/O in the low-level primitives?
        I think for this one, it's more important to make it easy to patch on than to actually have it.
    5. > To maintain the weird two-level Response structure where we have a urllib3 `Response` object that contains a httplib `Response` object within it?
        Kill it with fire.
5. > How would we want to manage a rollout of this functionality?
    We'd obviously want to keep in close contact with the people who vendor us. Beyond that... deploy it to pypi, I guess?
6. > How do we want to review it?
    It might be worth having an downstream thumbs-up on any PRs to the 2.0 feature branch to ensure that we're not accidentally breaking something important to pip/requests/etc. Beyond that, I think our standard requirement of 100% coverage is a good start.
7. > What kind of timeline do we want for releasing it?
    We should release it when it's done. 😜
8. If we decide to go ahead with a v2.0, should we freeze the current master branch for everything except bug/security fixes until we complete v2.0?
    Yes. However, given that urllib3 features are driven substantially by standards, we might want to establish some sort of a metric to determine which requests are "features", even if they're documented by a relevant standard that urllib3 isn't in strict compliance with. Hopefully the previous sentence makes sense.

njsmith:
I'm not too worried about urllib3 depending on h11 from a maintenance point of view. I mean, obviously I'm not going to guarantee any particular response time, and as usual you should have some plan in mind in case I wander off and abandon it, but -- it's a fairly small project with very well-defined scope. You'll also want to look at: https://github.com/njsmith/h11/issues/17

I'm sure there will be a few other things that need fixes/enhancements in h11, e.g. https://github.com/njsmith/h11/issues/7 is probably a blocker for you.

Python version support: trying to support 2.6 would probably be rather horrible do to the lack of support for `b""`; OTOH I guess maybe pip will have dropped 2.6 by the time this matters? We can dream... 2.7 support is already there, though. (In fact I followed @shazow's preferred approach, and wrote the whole thing as 3.5 only and then got @Lukasa to backport it ;-).) Actually 3.4 is in some ways more annoying than 2.7, because 3.4 [doesn't have `bytes.__mod__`](https://github.com/njsmith/h11/blob/1df9935a763f38418fc58e1ab47ac308b5759110/h11/_writers.py#L19-L27).

We probably should move the repo over to hyper... as far as I'm concerned that's just a "when we get around to it" item.

For reference, the actual codebase is (`cd h11 && cloc . --exclude-dir tests`):
```
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                          10            264            881            883
```
and the test suite (`cloc h11/tests`):
```
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Python                          10            309            197           1623
```

It's possible that by the time you get down refactoring, urllib3+h11 could be smaller than urllib3 :-).

I would definitely suggest planning on vendoring h11 for the first few releases, just to make sure everything is settled down API-wise before risking version skew issues. This is based on the intuition that the next day after urllib3-on-h11 is released you'll probably get some weird bug report that requires a breaking API change in h11, and that's a lot easier if we can keep h11's API slushy for a bit longer.

durin42:
> On Dec 8, 2016, at 5:46 PM, Nathaniel J. Smith <notifications@github.com> wrote:
> 
> Python version support: trying to support 2.6 would probably be rather horrible do to the lack of support for b""; OTOH I guess maybe pip will have dropped 2.6 by the time this matters?

Er, Python 2.6 is fine with b”” literals?

njsmith:
> Er, Python 2.6 is fine with b”” literals?

Oh, well, the grammar documented in the 2.6 language reference isn't :-). You can see how much time I'm spending on 2.6 these days...

Lukasa:
Cool, ok.

So the TL;DR seems to be that everyone is in favour of pursuing this, and the consensus is generally that we should consider coming up to a feature freeze on the master branch very shortly.

My proposal then is that I should try and do a bit more of the monolithic refactoring work now, but then turn this into a 2.0 branch that we can start collaborating on. It'd be nice to free up the possibility of people addressing some of my TODOs or generally helping with the ongoing work, as well as making it a bit easier for people to dive in to some of the other things that were on the list for #776.

Lukasa:
I'm beginning to get a bit more aggressive about removing things we no longer need. That'll help with a future refactoring effort: we won't spend our time trying to refactor in a content-length check (for example) that is no longer required because h11 does it for us.

Lukasa:
So, as I've been cleaning up the builds, I've noticed an issue: AppEngine. One major side-effect of this refactor is that we lose the tight integration with GAE that has been used in the past, because we are no longer using httplib. On top of that, we've generally be in favour of drastically changing the API we work with, meaning that it wouldn't even necessarily be that easy to get GAE working.

I think, for this reason, I'd like to tag @jonparrott to weigh in here, as well as the wider community. How do we forsee our GAE integration working in the future? Will we want to try to factor our code in such a way that it becomes possible to drop back in a httplib-style backend like GAE?

Lukasa:
So for the short term I'm going to mark the GAE builds as "allowed to fail" on this branch, because I'd like us to come back to them but I don't want them standing in the way right now. =D

Lukasa:
Woo, ok, tests are green!

So, I have opened up a [project](https://github.com/shazow/urllib3/projects/1) and shoved a whole lot of cards in there for things that people can do if they want to start getting involved. I'm going to continue pushing forward on some of the major refactoring work, but at this point now that the tests are passing (albeit with missing coverage) we have a decent basis to start work with.

For that reason, I'm going to move this to a `v2` branch on this repository and encourage people who want to get involved to start making pull requests against that branch. Please create issues and move project cards around to make it clear when something is being worked on by someone to avoid duplicating too much effort. I'm going to continue focusing on excising code that we don't need and trying to complete some of the refactoring so that I can reduce the feature surface, and then I'll look at the internal refactor to try to separate the common protocol code and the bits that are to do with "synchronous I/O", with the goal of enabling future work to support async code.

Lukasa:
Closing in favour of #1068.

shazow:
Re: GAE, Jon can definitely add more context, but my understanding is that GAE doesn't actually need httplib. [Even today](https://github.com/shazow/urllib3/blob/master/urllib3/contrib/appengine.py), it just straight-up uses `urlfetch` when available (and would have to fall back to plain sockets if not). If we can have a place where we bypass anything that touches a socket and pass it straight into AppEngine's `urlfetch` when enabled, then that's all it will take. This would mean we'd only support a subset of features, but that's already true today.

jonparrott:
Echoing what @shazow said: Today we're basically just adapting App Engine's urlfetch API to match urllib3's. As long as that remains a viable option, we can continue to support GAE. Also note that I'm more than willing to help with that.

To add some more context to this - I've been continuously pushing for App Engine standard to be "less weird". Some of that is coming to bear fruit, but App Engine's support for sockets is still "weird". I would expect to see App Engine standard's Python runtime become less weird over time. Also keep in mind that App Engine standard is currently limited to Python 2.7 only, and any Python 3 support on standard would likely look closer to "vanilla" Python than "App Engine" Python.

All this is to say - don't sweat App Engine that much.  As long as the API allows the right level of abstraction we can use the same approach we use today.

PR #1068, opened 2016-12-12
Work for v2 of urllib3.

Lukasa:
This is a follow-on from #1067. This meta-PR is used primarily to allow a single, easy place to see what work has been done and is being done for v2.

Smaller work items:

- Remove support for respecting chunked bodies as part of `stream`. (#1069)


test/test_selectors.py
```diff
@@ -16,7 +16,10 @@
 except ImportError:
     from time import time as get_time
 
-import resource
+try:  # Windows doesn't have the resource module.
+    import resource
```
SethMichaelLarson: How did this work in my PR adding Windows support?

Lukasa: The `try...except` exists in master but uses Python 2.6 as the justification for its presence. I removed that block a while back because v2 doesn't support 2.6.

SethMichaelLarson: Makes sense!

Lukasa:
Ok folks, so @haikuginger and @SethMichaelLarson have expressed interest in picking up some v2 work at some point, and additionally others might want to. The next major chunk of work is related to refactoring `urlopen` in the `PoolManager` and `ConnectionPool`. If people want to start picking this stuff up before I get there, I've outlined what work I think we need [here](https://files.lukasa.co.uk/notes/urllib3_refactor_request_response/).

njsmith:
@Lukasa: Looking at your notes there, I notice one of the things you're grappling with is how to structure the code to allow for mixed sync/async APIs. I had a thought here that might make this simpler, which outgrew this comment box so: https://gist.github.com/njsmith/7af696f98272f21294ebd0a7dbf268d4

I'd be interested to hear what you think.

Lukasa:
So the idea is an interesting one. It certainly seems like it could plausibly work, and would likely vastly reduce code duplication compared to the current plan. 

However, I have some concerns. My biggest ones are:

1. It makes Python 2 very much a second class citizen. The Python 2 code becomes an invisible cousin of the actual code in the tree. This makes it harder to support and substantially more brittle. 
2. It's extremely non-obvious, and therefore new-contributor-unfriendly. It would unquestionably require a section of the documentation explaining how it works and why we're doing it, as well as providing guidance on what contributors should (and shouldn't) do. This is also somewhat true of a "two separate trees" thing which I currently have in my head, but that has the advantage of only causing code duplication (which is simple enough for most programmers to deal with).
3. It's clever. Very clever. This is not normally a problem for urllib3 (it has a lot of clever code), but it is notable.
4. Testing is a concern. Because the tests also need to be processed accordingly, the tests are now written and then manipulated. This allows errors in code generation in product code to be matched by errors in generation in test code, and so to potentially be lost or overlooked. I have no real idea of how to deal with this issue. 

However, there are some big advantages to this kind of approach, particularly if we're willing to consider Python 2 a bit of an evolutionary dead end. If the rest of the contributors are open to pursuing this idea we can definitely spike it out and see if it works.

njsmith:
I agree with all that – though hopefully it should help with the testing issue that any black box tests of the synchronous public API will be unaffected by the de-coloring transformation.

Doh. I just realized I keep making these awkward phrases like "de-coloring transformation", when English has a perfectly good word for this! My proposal is that for Python 2 we use a script to *bleach* the code 😁

Lukasa:
> I agree with all that – though hopefully it should help with the testing issue that any black box tests of the synchronous public API will be unaffected by the de-coloring transformation.

Well, this rather begs the question, doesn't it? (Grammatically speaking no, not quite, but close!)

The synchronous tests will be unaffected by the *bleaching* only if they are written synchronously to begin with. That would require that our code deduplication process be applied only to the product code (and maybe some subset of the tests), still requiring that we write all our tests twice. Does that sound right?

njsmith:
Hmm, good point, this is a little subtle so let me think it through in more detail.

So the original proposal assumes that we ultimately want to expose a set of mostly-duplicate public APIs (one for each backend), and proposes to do that by having a fixed set of as-small-as-possible shims around a core library written using async/await and an IO abstraction, and then automatically generate a second py2-compatible version of the core library. So in the original proposal, we'd only be bleaching the core library and any tests written directly against it; it doesn't affect the shim layer or the public APIs it exposes. (For concreteness: In the gist, `abstract_echo_client` is what I'm referring to as the "core library", and `blocking_echo_client`/`trio_echo_client` are the "shim layer public APIs".)

So we have two problems:
* We have a bunch of almost-but-not-quite-duplicate public APIs that we need to test somehow. This problem isn't created by the bleaching proposal; it's just intrinsic to the kind of API we ultimately want to expose. But the basic bleaching proposal doesn't help either. And given that there probably will be differences between the APIs (e.g. timeout handling in the gist), we may have to accept some amount of test duplication here.
* We can write generic tests for the core library API (and we can even use sham implementations of `AbstractNetworkIO` if they want to, i.e. we can treat the core library as a sans-IO library for testing!). But these tests will need to be bleached to run on py2, some there's some added uncertainty about if they're really testing the right thing.

I guess a tempting option would be to minimize test duplication by writing most of the tests against the core library, and then just have a few high-level end-to-end tests for each of the shim layer public APIs, primarily to exercise the shim layer itself (but also as end-to-end tests). That's probably what I'd do if we were only targeting python 3 and didn't have to worry about bleaching. I think it still makes sense? Obviously there are some risks here, but I think I could probably convince myself to be comfortable with them given:

* the extreme simplicity of this particular code transformation
* that we would still have some untransformed end-to-end tests running against the synchronous shim API, which should exercise enough of the library to at least catch any gratuitous bugs caused by the code transformation

I've been trying to think of a case where removing the `async` and `await` keywords from a test produces a new test that accepts behavior it shouldn't. There are trivial cases like "bug in code transformer accidentally deletes the whole test", but that's why we'd have tests for the code transformer :-). And beyond that I've haven't been able to come up with much... I'm curious if anyone else can think of an example, even a contrived one, just to help guide our intuition.

Gathering coverage statistics for the transformed code seems like another way to get some confidence that the tests are doing something reasonable.

Lukasa:
While we are here, a few further concerns have occurred to me. The one that leaps out to me the most is that having an actually-synchronous-but-marked-async backend is a bit of an attractive nuisance on Python 3, so we'd probably also want to do some code gen on that platform to remove that backend entirely. This means that neither Py2 nor Py3 would use the code "as written", at which point we may as well save ourselves the code gen and just write the synchronous backend directly. Does that make sense?

njsmith:
> Does that make sense?

'fraid not :-). What problem are you trying to solve? What do you mean by "attractive nuisance"? Remember that the point of the code gen is to allow us to get away with maintaining only a single copy of all the substantive HTTP logic (i.e., the vast majority of the code in urllib3). Even if we didn't want to use the actual synchronous backend on py3 (and why wouldn't we?) and instead have some synchronous shim that allocates an asyncio loop and uses it (I'm guessing is what you might have in mind instead?), then this seems orthogonal to the code gen idea?

Lukasa:
What I'm getting at is that on Python 3 you'd have a "synchronous" API that nonetheless is made up of `async` functions that must be awaited. That is an attractive nuisance because if you don't think carefully and dig into the internals it's possible to *assume* you're using an asyncio-compatible async module but actually block your event loop, and have no easy visual way to spot that.

Worse, it means that to write a Python 2 + 3 compatible code base that uses urllib3 you'd have to also do code generation.

njsmith:
> if you don't think carefully and dig into the internals it's possible to assume you're using an asyncio-compatible async module but actually block your event loop,

Well, I mean the fact that you had to type
```python
pool = urllib3._abstractapi.AbstractPoolManager(urllib3._backends.sync.SynchronousIO())
```
might give you a few clues that things aren't going to go well?

A similar problem exists for distinguishing 
`urllib3.asyncio.PoolManager` vs `urllib3.curio.PoolManager`. If you mix those up it'll fail with some incomprehensible error instead of blocking, but either way I don't think it's a likely mistake.

> Worse, it means that to write a Python 2 + 3 compatible code base that uses urllib3 you'd have to also do code generation.

Huh? The synchronous public API is there on both versions, so 2+3 compatible code can still use that just like it does now. There are duplicate public APIs for the different backends, so if you want to write some duplicate logic to wrap them that option still exists (that's the main alternative you're contemplating, right?). And then on top of this downstreams also get the *option* of doing code generation to efficiently wrap all the APIs with a single code base (if you choose to also make the abstract interface a public API).

I genuinely don't know if this idea is a good one, but I'm not worried about these particular issues :-)

Lukasa:
> Well, I mean the fact that you had to type
> 
> ```python
> pool = urllib3._abstractapi.AbstractPoolManager(urllib3._backends.sync.SynchronousIO())
> ```
> 
> might give you a few clues that things aren't going to go well?

At that specific point in the code, sure, but not anywhere else. Everywhere else is decorated with the keyword `await`.

> Huh? The synchronous public API is there on both versions, so 2+3 compatible code can still use that just like it does now. There are duplicate public APIs for the different backends, so if you want to write some duplicate logic to wrap them that option still exists (that's the main alternative you're contemplating, right?)

Hang on, I thought that we had *one* `AbstractPoolManager`, decorated with `async` keywords, that was stripped of those keywords on Python 2. That would mean that the signature of, say, `AbstractPoolManager.request` would have to be `def request(self, *args, **kwargs):` on Python 2 and `async def request(self, *args, **kwargs):` on Python 3. Those two functions require different calling conventions, and so are not source-compatible: you need to code generate between them.

njsmith:
> At that specific point in the code, sure, but not anywhere else. Everywhere else is decorated with the keyword `await`.

OK, but this requires that someone find this weird internal API, explicitly put it into synchronous mode, and then use it in asynchronous code? I feel like this is like saying that `aiohttp` is an attractive nuisance because someone might import it and then forget it said "aio" in the name and try using it inside a twisted `ensureDeferred` function.

> Hang on, I thought that we had one `AbstractPoolManager`, decorated with `async` keywords, that was stripped of those keywords on Python 2. That would mean that the signature of, say, `AbstractPoolManager.request` would have to be `def request(self, *args, **kwargs):` on Python 2 and `async def request(self, *args, **kwargs):` on Python 3. Those two functions require different calling conventions, and so are not source-compatible: you need to code generate between them.

Sure, but `AbstractPoolManager` is primarily an internal API; the main public API is the shim layer on top of it. `urllib3.sync.SyncPoolManager` (or whatever) would exist and its methods would have the same signatures on both py2 and py3. That's the whole point :-).

With the bleaching approach we have the *option* to make `AbstractPoolManager` a kind of "hazmat" public-but-you-better-know-what-you're-doing API, which would *allow* libraries like `requests` to *also* use the bleaching approach by writing a single code base against the abstract API and then bleaching it to generate a py2 version. But if you compare two strategies that have been proposed for multi-backend urllib3 – code-duplication vs bleaching – then bleaching provides exactly the same public APIs as code-duplication, plus optionally this extra abstract API. So any strategy for calling urllib3-using-code-duplication will also work to call urllib3-using-bleaching; it's just that with the bleaching approach there's this extra strategy potentially available for downstream projects if they find it attractive.

Lukasa:
> `urllib3.sync.SyncPoolManager` (or whatever) would exist and its methods would have the same signatures on both py2 and py3. That's the whole point :-).

Ok, so your mental model of this is not `class SyncPoolManager(AbstractPoolManager)`, but instead just `class SyncPoolManager` with a note that this has the same general interface as `AbstractPoolManager` but with no `async` declarations on the class methods?

njsmith:
@Lukasa: doh, I think the word "abstract" is what's throwing us. Sorry. I just meant "abstract" as like "not tied to any particular backend" – it's still a concrete interface and nothing inherits from it. For what I had in mind for `AbstractPoolManager` and `SyncPoolManager`, see the relationship between `abstract_echo_client` and `blocking_echo_client` in the [gist](https://gist.github.com/njsmith/7af696f98272f21294ebd0a7dbf268d4).

Abstract is definitely the wrong word to use here. s/abstract/flexible/g, maybe?

namexuhuan:
Sgit

PR #1069, opened 2016-12-12
Stop supporting read_chunked and friends.

Lukasa:
This patch removes support for `read_chunked` and all its supporting methods. We agreed to remove these because supporting the functionality whereby we'd expose chunked bodies to the user is not all that useful.

We should consider whether `stream` should swap over to using `read1` instead of `read`, but that's a different patch: see https://github.com/shazow/urllib3/projects/1#card-1081219.


Lukasa:
Also: just *look* at this giant negative diff!

sigmavirus24:
For reference, the issue was https://github.com/shazow/urllib3/issues/550

sigmavirus24:
So I think the heart of that issue is:

> upper layers, `requests`, could process those encoded chunks and developers wouldn't need to decode those chunks.

So we don't need to continue to provide access to the chunks, but we do need to respect a user's decision to decode those (which we apparently weren't doing before?). Can we add tests to ensure that even with this PR, we're retaining that behaviour that we once had?

Lukasa:
Interestingly, I think I disagree. Quoting the original original issue, kennethreitz/requests#2449, with emphasis added by me:

> Right now if I do `response.iter_lines()` I get access to chunks part where **length of the chunk is encoded with optional chunk extension**. It would be awesome if requests would provide something like `iter_chunks` where I would get access to **the chunk data only**.

Reading this now with the benefit of hindsight, this issue seemed to be originally about the fact that the chunk extension was confusing the httplib that was being used there. Our custom code handled the extensions better, so the issue went away.

In terms of worry about decompression of chunks, I think this PR makes it a non-issue, because both chunked and non-chunked bodies now take the same route through the codebase. The differences between the two are handled entirely by h11.

Lukasa:
\o/ Thanks @sigmavirus24! :cake:

PR #1072, opened 2016-12-14
remove _UNKNOWN

nateprewitt:
This will remove `_UNKNOWN`, a sentinel object used in httplib, out of the `OldHTTPResponse`. All of these variables, except `length` are initialized in `begin` so their starting state of `None` is fine. I can't find anything that is actually touching `length` in `OldHTTPResponse` now.

### DIFF
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -122,8 +122,7 @@ class DummyConnection(object):
     pass
 
 
-# TODO: This is a holdover from httplib, do we need it?
-_UNKNOWN = object()
+_DEFAULT_SOCKET_OPTS = object()
 
 
 # TODO: This is a holdover from httplib, do we need it?
@@ -167,12 +166,12 @@ class OldHTTPResponse(io.BufferedIOBase):
         self.headers = self.msg = None
 
         # from the Status-Line of the response
-        self.version = _UNKNOWN
-        self.status = _UNKNOWN
-        self.reason = _UNKNOWN
+        self.version = None
+        self.status = None
+        self.reason = None
 
-        self.length = _UNKNOWN          # number of bytes left in response
-        self.will_close = _UNKNOWN      # conn will close at end of response
+        self.length = None          # number of bytes left in response
+        self.will_close = None      # conn will close at end of response
 
     def _read_response(self):
         """
@@ -470,7 +469,7 @@ class HTTPConnection(object):
     response_class = OldHTTPResponse
 
     def __init__(self, host, port, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                 source_address=None, socket_options=_UNKNOWN):
+                 source_address=None, socket_options=_DEFAULT_SOCKET_OPTS):
 
         # TODO: Do we need this? I think we might not: urllib3 may not ever
         # provide host and port in one string like the stdlib allows.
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Is there a specific reason behind this change?

SethMichaelLarson:
Kills that `TODO`? Unless that is meant to be for the whole class.

nateprewitt:
Yeah this addresses the project list item "remove _UNKNOWN for v2. Haven't found a way to link to those.

SethMichaelLarson:
@nateprewitt I don't think there's a way to link project notes to Issues / PRs

haikuginger:
https://github.com/shazow/urllib3/projects/1#card-1080206

nateprewitt:
Thanks @haikuginger!

Lukasa:
Yeah, the TL;DR high level overview of why I want changes like this is that I want to reduce the surface area of OldHTTPResponse to make factoring it away easier. If I can more easily see what we don't need, then it is easier to do that.

urllib3/connection.py
```diff
@@ -167,12 +163,12 @@ def __init__(self, sock, state_machine, method=None, url=None):
...
-        self.version = _UNKNOWN
-        self.status = _UNKNOWN
-        self.reason = _UNKNOWN
+        self.version = None
+        self.status = None
+        self.reason = None
 
-        self.length = _UNKNOWN          # number of bytes left in response
-        self.will_close = _UNKNOWN      # conn will close at end of response
+        self.length = None          # number of bytes left in response
```
Lukasa: Yeah, this can almost certainly go away too. Another PR for that would be much appreciated.

urllib3/connection.py
```diff
@@ -470,7 +466,7 @@ class HTTPConnection(object):
     response_class = OldHTTPResponse
 
     def __init__(self, host, port, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                 source_address=None, socket_options=_UNKNOWN):
+                 source_address=None, socket_options=default_socket_options):
```
Lukasa: Let's change this patch. For various boring reasons it's a bad idea to use mutable objects as the default values of function arguments: it doesn't do what you expect and can lead to really weird bugs. Better to use an `if is not None` check rather than throw it away entirely.

nateprewitt: So I went with `None` originally for this, and the problem is that `None` has different semantics than `_UNKNOWN`.

Passing `None` was previously equivalent to `self.socket_options = None`, but 
```python
self.socket_options = (
             socket_options if socket_options is not None		
             else self.default_socket_options		
)
```
means that that it falls to the else statement and `self.socket_options = self.default_socket_options`. 

I think we'll either need another sentinel object for this specific scenario, or change how we signify "No Socket Options".

SethMichaelLarson: Perhaps sentinel for this one option?

Lukasa:
Looks like we might need a sentinel, unless this is duplicated by our other logic for socket options.

nateprewitt:
Yeah, I took a deeper dive yesterday and I don't think we can avoid the sentinel here. I added a more specific one for this case.

sigmavirus24:
Unrelated to the code in this PR, but I'd really prefer if people create issues from project cards so that they're reference-able. On top of that, more descriptive commit messages would be *wonderful*. As it stands "remove _UNKNOWN" tells me nothing more than reading the patch does.

nateprewitt:
Totally, +1 on that. I'll reword this with an extended commit message. Perhaps a smaller umbrella issue of "simplifying OldHTTPResponse for v2 refactoring" so we're not creating a new ticket for every small removal with the same purpose?

nateprewitt:
Alright 3.4 needs a quick kick, but this should be set for whenever. Don't feel compelled to review @Lukasa, I'm happy to let this sit until after the holidays  :)

sigmavirus24:
Restarted 3.4 on OSX for you @nateprewitt

PR #1074, opened 2016-12-23
Draft pipeline architecture

haikuginger:
@Lukasa and I had discussed a possible architecture alteration for V2. Instead of having `PoolManager` encapsulate `ConnectionPool`, `SessionManager` encapsulate `PoolManager`, and so on, each of these high-level objects would be built as a pipeline of vastly-simplified steps, each of which would be entirely self-contained.

This pull request is a draft of a first step towards doing that. It adds [a pipeline architecture I've been working on](https://github.com/haikuginger/httppipeline), and refactors `PoolManager` into a set of individual pipeline elements.

As this is quite speculative, I haven't written tests for it yet; this is more of a proof-of-concept for discussion.

For right now, this isn't integrated with `RequestMethods` (which would be broken down into a couple pipeline elements of its own); to test it, do the following:

```python
from urllib3.pipeline.poolmanager import PoolManagerPipeline
pm = PoolManagerPipeline()
```

From there, doing `pm.request()` should behave similarly to `PoolManager.urlopen()`, with the exception of not taking positional arguments. Positional arguments could be added with an API-level pipeline element.

Future benefits include the ability to build small, easily-reusable modules. For example, we could add a `CookiePipelineElement` that sat near the base level, handling cookies on each request that passes through and response that passes back. We'd also be able to test individual elements as units much more easily without substantial mocking.

Pinging @jonparrott, @Lukasa, @SethMichaelLarson, @shazow, and @sigmavirus24 to take a look and drop their opinions. If people are happy with the overall architecture, then I'll continue working on it.

### DIFF
--- a/urllib3/pipeline/base.py
+++ b/urllib3/pipeline/base.py
@@ -4,9 +4,6 @@ import six
 
 from urllib3.pipeline.context import Context, ContextWrapper
 
-def has_callable_attr(obj, attr_name):
-    return hasattr(obj, attr_name) and hasattr(getattr(obj, attr_name), '__call__')
-
 
 class HttpPipelineElement(object):
     
@@ -16,14 +13,6 @@ class HttpPipelineElement(object):
         that can be used to store state in a context object.
         """
         self.id = uuid4().hex
-
-    def save_context(self, context, key, value):
-
-        context.setdefault(self.id, {})
-        context[self.id][key] = value
-
-    def get_context(self, context, key, default=None):
-        return context.get(self.id, {}).get(key, default)
 
     def apply(self, context, **kwargs):
         return kwargs
--- a/urllib3/pipeline/context.py
+++ b/urllib3/pipeline/context.py
@@ -1,3 +1,5 @@
+_DEFAULT = object()
+
 class ContextWrapper(object):
 
     def __init__(self, pipeline_element_id, context):
@@ -7,10 +9,10 @@ class ContextWrapper(object):
     def save(self, key, value):
         self.context._save(self.element_id, key, value)
 
-    def get(self, key, default=None):
+    def get(self, key, default=_DEFAULT):
         return self.context._get(self.element_id, key, default)
 
-    def pop(self, key, default=None):
+    def pop(self, key, default=_DEFAULT):
         return self.context._pop(self.element_id, key, default)
 
 class Context(dict):
@@ -19,14 +21,18 @@ class Context(dict):
         self.setdefault(pipeline_element_id, {})
         self[pipeline_element_id][key] = value
 
-    def _get(self, pipeline_element_id, key, default=None):
-        if pipeline_element_id in self:
-            return self[pipeline_element_id].get(key, default)
+    def _get(self, pipeline_element_id, key, default=_DEFAULT):
+        self.setdefault(pipeline_element_id, {})
+        context = self[pipeline_element_id]
+        if default is _DEFAULT:
+            return context.get(key)
         else:
-            return default
+            return context.get(key, default)
 
-    def _pop(self, pipeline_element_id, key, default=None):
-        if pipeline_element_id in self:
-            return self[pipeline_element_id].pop(key, default)
+    def _pop(self, pipeline_element_id, key, default=_DEFAULT):
+        self.setdefault(pipeline_element_id, {})
+        context = self[pipeline_element_id]
+        if default is _DEFAULT:
+            return context.pop(key)
         else:
-            return default
+            return context.pop(key, default)


SethMichaelLarson:
Has there been any thought into how this would affect our decision about h2? Are we planning that for v2?  HTTP/2 doesn't lend itself as easily to a pipeline architecture compared to HTTP/1.1

Lukasa:
> HTTP/2 doesn't lend itself as easily to a pipeline architecture compared to HTTP/1.1

Sure it does. What makes you think it doesn't?

urllib3/pipeline/base.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        self.id = uuid4().hex
+
+    def save_context(self, context, key, value):
+
+        context.setdefault(self.id, {})
+        context[self.id][key] = value
+
+    def get_context(self, context, key, default=None):
+        return context.get(self.id, {}).get(key, default)
```
Lukasa: This logic makes it impossible to store a `None` value in the context. Do we want that limitation?

shazow:
I'd love to see a code sketch of what the end goal is with this refactor compared to what we have today. Specifically, I'd want to see examples of what the strengths of this design would be and where our current design falls short.

One thing I worry about is that we're accomplishing decoupling by replacing native Python features (objects, arguments, embedding/inheritance, etc) with a catch-all arbitrary dict of key-values ("context"). I think this is a fairly big sacrifice: we do get a lot of clarity and benefits from native Python features, and while the "Everything is data" mentality of functional languages is appealing I'm afraid that in the context of Python it might turn into something unmanageable.

Note that I'm not arguing against pipeline architectures in general, but just concerns with this specific approach. :) I'm sure there's a sweet spot somewhere, and I want to see if this is it.

SethMichaelLarson:
@Lukasa Because HTTP/2 is a lot more than send request and receive data back which is basically all that HTTP/1.1 is. I'll stop arguing it because I know for a fact you know HTTP/2 and what's going in v2 more than me. :)

urllib3/pipeline/base.py
haikuginger: Not sure I follow - it could lead to confusion if the upstream user is trying to determine if a value they stored is `None`, while the default use will return `None` if nothing's stored, but that could be circumvented by providing an alternate default value.

Would you prefer an implementation that raises a KeyError if a value doesn't exist and a default value isn't explicitly provided? E.g., `self.get_context(context, 'my_key')` would raise, while `self.get_context(context, 'my_key', None)` would return `None`.

Lukasa: I'd suggest we have both methods, if possible.

haikuginger:
@shazow, like I mentioned at PyConCA, a lot of this alternate architecture came about as a result of frustration with trying to shoehorn the cookie-handling aspects of `SessionManager` into the existing architecture.

One example is redirect logic - because `SessionManager` was operating "above" `PoolManager`, but still had to collect cookies on a redirect, we had to move the redirect logic to a new method on RequestMethods, as well as force `PoolManager` to disable redirects.

In comparison, with a pipeline-based architecture, we can insert the cookie handling logic exactly where it needs to go (near the bottom of the stack) without any vast rearchitecture - we just create a `SessionManagerPipeline` object that has a `CookiePipelineElement` inserted at the appropriate place. We can also make strong API claims about both the pipeline as a whole and individual pipeline elements that users can assemble for their own purposes.

I understand your concern about the `Context` object. I'm less worried about it, because the way it's architected means that keys can only be accessed by the PipelineElement that created them in the first place, and because context values are semantically different from instance attributes - instance attributes are shared across all requests, while context values are locked to a particular request.

We're not giving up complex data structures where we really need them; `Context` just allows us to say, "here's an important piece of information about this request that I might need later on" - and it can be in the form of a complex data structure if it needs to be. Rather than a part of the overall stored state, `Context` should be thought of as a way for us to maintain the same set of variables across multiple entries into a given pipeline element. In other words, the following two blocks should be thought of the same way:

```python
def run():
    x = 1
    do_other_thing()
    print(x)

class PrintsOnePipelineElement(HttpPipelineElement):

    def apply(self, context, **kwargs):
        context.save('x', 1)

    def resolve(self, context, response):
        print(context.get('x'))
```

I'm sure there's a better way to present it from a user-facing API perspective (that's where you fine people come in), but I'm not sure there's a way to create a reliable pipeline framework without the ability to stick arbitrary bits of context somewhere for use later on- whether that's by glomming onto the request or somewhere else.

On a separate note: I read through the HTTP/2 spec last night and wondered if we might want to normalize on a `urllib3.Request` object that's isomorphic with an HTTP/2 request - have attributes for the pseudo-headers like `:authority`, `:method`, etc., while having other headers in another attribute, and request body in a final one. As for pipelining HTTP/2, I don't see a specific problem with it either; the big question is if we want urllib3 to accept `PUSH_PROMISE` frames, and if so, how we should present them in the API. For the rest of the API, it seems like the concept of "selecting a connection" should still work - we'll just be picking an HTTP/2 stream instead.

haikuginger:
@Lukasa, I've done some cleanup and removed stuff that wasn't being used. I stayed with a single method for get, pop, etc, but I modified it so that we fall through to the behavior we'd see if doing `dict.get()` or `dict.pop` directly.

Lukasa:
So, here's a question: can/should `Context` be replaced by `Request`/`Response` objects? That is, should all of our pipeline operate on semantic objects, rather than catch-all key-value-pair objects?

nateprewitt:
> So, here's a question: can/should Context be replaced by Request/Response objects? That is, should all of our pipeline operate on semantic objects, rather than catch-all key-value-pair objects?

Personally, I feel like at least having a distinction between `Request` and `Response` is important. Amorphous super objects can be easy to confuse/debug when they require additional interaction to determine what it's purpose is. If a separation is possible without significant work, I'm +1 on it.

That said, this is really exciting stuff @haikuginger! Excited to see where this goes.

haikuginger:
> can/should `Context` be replaced by `Request`/`Response` objects?

I don't think so, because I believe they serve different purposes. `Context` is a bundle of state that's specific to a particular request at a particular place in the pipeline, and shouldn't be mutated except by that pipeline element, when working with that particular request.

In contrast, a `Request` object (which I'm absolutely in favor of) can and should be mutated at pretty much every step of the way; no pipeline element should ever consider it safe to store state on a request directly for later retrieval, because any other pipeline element could modify it further down the line.

I think my ideal "outgoing" method signature for a pipeline element looks a bit like this:

```python
def apply(self, request, context, **kwargs):
    pass
```

In this model, there are four places to put state, each with a particular scope:

- `self`, or the pipeline element itself, is where global settings can be stored. For example, on a `RetryPipelineElement`, you might store `default_max_retries`. Another example would be a `CookiePipelineElement`, which harvests `Set-Cookie` headers from each request, and applies them to all other requests that fit the bill.

- `request` is a strictly-defined HTTP request object. It contains a set of well-defined fields with specific semantic meaning - for example, `headers`, `body`, etc. These fields can and should be modified by any pipeline element that the request passes through; the whole pipeline put together is in the business of creating a complete request that meets the user's intent. For example, a `BasicAuthPipelineElement` would do `request.headers['Authorization'] = 'Basic: xxxxxxxxx'.

- `context` is a mushy bag of k/v pairs, as everyone's already noticed. ;) But the purpose is quite well-defined. Fields in `context` are necessary to allow each pipeline element to save state about a particular request that can't be modified by other pipeline elements. For example, a `RedirectPipelineElement` might set an `original_request` variable to allow it to hold onto information about whether a particular request ought to be redirected when relevant or not.

- `**kwargs` is also, obviously, a mushy bag of k/v pairs. What it allows us to do is take arguments that don't strictly belong in the request, and pass them through until they reach the pipeline element to which they're directly relevant. And, it allows higher-level pipeline elements to send information to lower-level elements. For example, in the existing code, the `RedirectPipelineElement` generates a `Retry` object to save in its context and use to determine whether to allow a particular redirect. `**kwargs` allows it to pass that same object down the pipeline to a lower-level `RetryPipelineElement` (not currently implemented; it's just using ConnectionPool.urlopen()` at the moment) that can use the same `Retry` object, even though the concept of a retry isn't strictly related to the request itself (in that it's not something that determines socket-level behavior).

In other words, what I'm trying to get at is that there's a place for both strictly-defined semantic objects, where possible, and arbitrary k/v stores.

Lukasa:
> `Context` is a bundle of state that's specific to a particular request at a particular place in the pipeline, and shouldn't be mutated except by that pipeline element, when working with that particular request.

So, here's a question: why have `Context` at all?

Each pipeline element can be an instance of a class, and so can store its own state however it chooses. Those that want to be stateless can be, and those that want to store state can store it on themselves. Why pass that state around at all? Why not make it the responsibility of the pipeline elements to persist state as they see fit?

haikuginger:
> Why pass that state around at all?

So, this is a valid question. An alternate structure could assign UUIDs to each request, rather than to each pipeline element - then, pipeline elements could lookup data on themselves based on that UUID, rather than look up data tied to the request based on their own UUID. So that's not out of the question at all, although, from an intuitive level, it seems like it might not be as good from a thread safety perspective.

> Why not make it the responsibility of the pipeline elements to persist state as they see fit?

Because we're not just making tools for us to use ourselves here; PipelineElements, if agreed upon, will likely become _the_ canonical method of extending the capabilities of urllib3. By clearly defining what the different canonical scopes of state are, and by providing well-implemented, thoroughly-reviewed (not as they are at the moment; obviously this is still a WIP) methods to use those scopes, we improve the usability of our own API, and can help to prevent errors from being made while building case-specific state storage mechanisms.

Lukasa:
Does it make sense for pipeline elements to maintain per-request state? What pipeline elements would need it? I can see them needing to maintain per-pipeline state, but why per-request state?

haikuginger:
> What pipeline elements would need it?

One example would be a CookiePipelineElement - one instance of per-request state that needs to be kept around is the request domain, so that if a response's `Set-Cookie` header specifies a domain to set a cookie for, we can determine whether the host we reached is allowed to set a cookie on that domain.

Lukasa:
Sure, but can't we do that by saying that when a response is received we pass it back up the chain along with the request that initiated it?

haikuginger:
That's certainly an option, but the request might be mutated by downstream elements - for example, (and this is just a hypothetical), a ProxyPipelineElement would modify a request's URL; if the CookiePipelineElement is "above" the ProxyPipelineElement, then reading the URL of the request would give the false impression that the request was sent to `proxy.com`, when it was actually a request to `realsite.com` via `proxy.com`.

And there are other cases where the data doesn't fit neatly in the paradigm of a Request object - for example, we'd want to keep per-request state about the number of redirects and retries that have been performed for a particular request, but that data doesn't fit neatly in the concept of a Request object being a representation of the decisions we make about choosing sockets and writing data over them - in other words, `url` _belongs_ on a Request object, but `retries_performed` doesn't so much.

Lukasa:
> but `retries_performed` doesn't so much.

So, this is the most compelling argument so far that we need *something* more, but I don't know that it needs to be a massive state object. What we're discussing here is fundamentally the idea of a "transaction": a series of requests that are logically related in some way. Is there any reason we shouldn't just generate a "transaction ID" when we begin and pass that to all objects, thereby continuing to push this out of the core logic into the helpers?

haikuginger:
Results of IRC conversation: something like the current `Context` object might not actually be the _worst_ thing ever, but we should reframe it so that its purpose is narrower and more clearly defined.

jonparrott:

PR #1077, opened 2016-12-23
Add Python 3.6 support.

Lukasa:


SethMichaelLarson:
Quick question @Lukasa, any reason why we don't test against the 'nightly' Python build?

Lukasa:
Honestly, I think the answer is because we don't care? It'd have to go in "allowed failures", because upstream do break their master branch from time to time, and at that point we don't notice when things go wrong very much.

PR #1081, opened 2017-01-03
Cleanup request methods.

Lukasa:
This patch removes `putrequest`/`putheader`/`endheaders`, all of which were superfluous to our actual needs. Given that we plan to use a static Request object at some point in the future, we have no need to allow these incremental request building methods, and so we can avoid them entirely and remove several bits of state tracking they enforced.

It also removes a bunch of code. Yay!

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -623,7 +623,7 @@ class TestProxyManager(SocketDummyServerTestCase):
         base_url = 'http://%s:%d' % (self.host, self.port)
 
         # Define some proxy headers.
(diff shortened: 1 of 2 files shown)


Lukasa:
Ok, so this patch has revealed that internationalized domain names are currently broken in the v2 branch. That's fine, I think: urllib3 hasn't ever really thought about them in a sensible way, so we should take this opportunity to decide whether we want to support them directly or whether they should be our user's problem (or indeed a Requests issue).

For now, however, I'm removing httplib's terrible IDNA support from this branch, because it's half-assed and doesn't work for proxies anyway.

Lukasa:
Ok, I'm going to go ahead and merge this. We're making forward progress on code coverage too. :sparkles:

PR #1083, opened 2017-01-04
Remove getheader/getheaders from OldHTTPResponse

Lukasa:
These are vestigial and we have our own implementation higher up the tree anyway.


Lukasa:
It's important to set the base branch correctly. Whoops.

PR #1084, opened 2017-01-05
Refactor common code from reading loops

Lukasa:
This pulls out a bunch of common code, trying to strip down the number of places we touch h11 and remove duplicate code. To get both `read()` and `read1()` working sensibly is a bit of a pain in the neck, so right now we're using this wacky generator-pipeline construction to pull it off.

Ultimately, I think this is mostly temporary, and we'll move to a new I/O model that removes a bunch of this complexity, but right now we're not there.


Lukasa:
Hah, look at all that improved coverage! Very nice. =D

PR #1086, opened 2017-01-05
Remove BufferedReader from OldHTTPResponse

Lukasa:
Frankly, this subclass is just cluttering up the codebase. It adds a bunch of methods we don't need, and requires that we do some weird workarounds like the `isclosed()` method. Let's abandon that, and delete a whole bunch more code.

OldHTTPResponse is getting pretty small now!


Lukasa:
Awesome, down to 28 uncovered lines of code!

PR #1095, opened 2017-01-18
Add fast path for system call wrapper on Python 3.5+

SethMichaelLarson:
I've updated my selectors2 project to 1.1.0 and the main feature added is a fast path the system call wrapper on Python 3.5+. This wrapper can be substantially smaller and faster because in Python 3.5+ PEP 475 is implemented which automatically retries system calls in the case of an interrupt. I figured I would propagate these changes here to make sure urllib3 is getting everything out of its selectors upgrade. :tada:

Overall the wrapper performs about **126% faster** with a happy path and **36% faster** with an error path.

This version also catches and correctly handles an error on Windows platforms when trying to call `BaseSelector.unregister` on a closed socket where the previous version did not.

### DIFF
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -8,9 +8,11 @@
 import errno
 import math
 import select
+import socket
+import sys
 from collections import namedtuple, Mapping
-
 import time
+
 try:
     monotonic = time.monotonic
 except (AttributeError, ImportError):  # Python 3.3<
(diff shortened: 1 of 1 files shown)


urllib3/util/selectors.py
```diff
@@ -8,9 +8,11 @@
 import errno
 import math
 import select
+import socket
+import sys
 from collections import namedtuple, Mapping
-
 import time
```
Lukasa: If we're rearranging imports it might be better to move `import time` above the `from` import.

urllib3/util/selectors.py
```diff
@@ -50,60 +52,75 @@ def _fileobj_to_fd(fileobj):
...
-        # select.error is thrown by select.poll.poll
-        # Aren't we thankful for Python 3.x rework for exceptions?
+            return func(*args, **kwargs)
         except (OSError, IOError, select.error) as e:
-            # select.error wasn't a subclass of OSError in the past.
             errcode = None
             if hasattr(e, "errno"):
                 errcode = e.errno
             elif hasattr(e, "args"):
                 errcode = e.args[0]
```
Lukasa: Looks like we have a problem with reachability here. When would we expect to have an error with args here, and can we trigger it in testing?

SethMichaelLarson: I noticed this too, I'll do some more digging.

SethMichaelLarson: So it looks like the `.args` attribute is only an attribute of `select.error` before Python 3.3 ([PEP 3151](https://www.python.org/dev/peps/pep-3151)) and is just an alias of `OSError` in 3.3+ meaning it only has an `.errno` attribute. I'll submit a commit removing the dead code.

SethMichaelLarson:
Only fail was a transient failure on Mac OS. Ready for re-review @Lukasa :)

PR #1098, opened 2017-01-19
Skip test_select_many_events if OSError(24) is thrown on Travis.

SethMichaelLarson:
This error has been seen on Mac OS builders with the past few PRs, wondering if it's a Travis Mac OS error only and isn't anything wrong with selectors. If so, this will disable that test on Travis Mac OS builders when the error occurs. Should not hurt coverage at all.

@Lukasa said something about investigating, if you find anything that would cause this besides Travis just close this. :)

### DIFF
--- a/test/test_selectors.py
+++ b/test/test_selectors.py
@@ -364,7 +364,14 @@ class BaseSelectorTestCase(unittest.TestCase, AlarmMixin, TimerMixin):
         readers = []
         writers = []
         for _ in range(32):
-            rd, wr = self.make_socketpair()
+            try:
+                rd, wr = self.make_socketpair()
+            except OSError as e:
+                if e.errno == 24 and 'TRAVIS' in os.environ:
+                    self.skipTest(('Could not open enough sockets. This occurs '
+                                   'sometimes on Mac OS Travis builders.'))
+                else:
+                    raise
             readers.append(rd)
             writers.append(wr)
             s.register(rd, selectors.EVENT_READ)


sigmavirus24:
I'd much rather not skip something out of convenience. OS X has different constraints than some Operating Systems and can cause *very* weird interactions. I'd instead encourage you to treat this as an early warning that something is wrong with selectors. These never happened previously.

SethMichaelLarson:
@sigmavirus24 This also didn't happen when selectors were finalized and merged. It's only come up the last few days and @Lukasa mentioned investigating it because he was seeing the troubles in his v2 branch as well. I'm not suggesting merging this immediately, but if the problem is with Travis I don't see why we shouldn't skip the test.

sigmavirus24:
@SethMichaelLarson because it's highly unlikely it's only happening on Travis. If it can happen in a CI environment it can happen in a user environment and if we're triggering it in a test run, then it can probably just as easily be triggered in real usage.

SethMichaelLarson:
@sigmavirus24 This is why I stated in the PR that I'm deferring this to @Lukasa's investigation as I do not own a Mac OS in which to reproduce this problem on. He and others will certainly know more about Mac OS more than I do.

Lukasa:
Yeah, so I'm inclined to want to work out what is going on here. It is *probably* not more than just having a ulimit that is too low, but we should consider this a problem worth addressing.

Lukasa:
So, this is most easily detectable on PyPy on my machine, which seems to show that this is ultimately a collection issue. What's not entirely clear to me is why PyPy is using up so many more FDs than the rest of the test cases. I can't reproduce this if I *just* run the selectors test case, and forcing a gc cleanup before the `many_selectors` test doesn't resolve it either.

So what I think I'm going to try to do is to forcibly hang the tests when that OSError gets raised an introspect the process to see what FDs it has open.

Lukasa:
So a quick look at the list of FDs that are open shows that there are mostly connections lying around from previous tests. These connections aren't even necessarily closed, and it's entirely possible that PyPy's GC is unwilling to clean them up yet.

Happily, this is ultimately just the result of the default soft ulimit on macOS. By default it allows only 256 open files: running `ulimit -n 1024` on my machine before running the tests basically resolves the issue for me. So the fastest fix is to do that in our Travis script.

It'd be nice to know why forcing PyPy to GC doesn't clean these up, but I do know that PyPy won't automatically GC unless it's under memory pressure: exhaustion of FDs doesn't trigger a PyPy GC.

The other option is to ensure that our tests do more aggressive cleanup, but that's a quite substantial chunk of work.

Lukasa:
As I look through the open files I note that there are a lot of KQUEUE objects still open:

```
pypy    5312 cory    5u  KQUEUE                                       count=4, state=0x8
pypy    5312 cory    8u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory   18u  KQUEUE                                       count=18, state=0x8
pypy    5312 cory   23u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory   30u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory   47u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory   52u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory   62u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory   90u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory  108u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory  118u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory  122u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory  131u  KQUEUE                                       count=1, state=0x8
pypy    5312 cory  158u  KQUEUE                                       count=7, state=0x8
pypy    5312 cory  172u  KQUEUE                                       count=13, state=0x8
pypy    5312 cory  177u  KQUEUE                                       count=11, state=0x8
pypy    5312 cory  204u  KQUEUE                                       count=2, state=0x8
pypy    5312 cory  212u  KQUEUE                                       count=0, state=0x8
```

I don't think we should be leaking any of these in the current master branch. So it's a bit nerve wracking that they're still being held around. Many of these seem to have quite a few file descriptors still registered with them as well.

While I'm here, the count of open objects for that PyPy process at the time was:

```
   3 CHR (a character special file, probably stdin/stdout/stderr)
   1 DIR (a directory handle)
  24 IPv4 (an IPv4 socket)
 147 IPv6 (an IPv6 socket)
  18 KQUEUE (a kqueue object)
  34 PIPE (a...uh...pipe)
  15 REG (a regular file)
   2 systm (not sure what these are but there aren't enough to care)
  28 unix (UNIX domain sockets)
```

The unix domain sockets almost invariably have the highest FDs, which is a bit odd: I don't know why so many of those got created late.

Anyway, we have 171 sockets floating around in this PyPy process, which feels like it's a bit much.

SethMichaelLarson:
Does `select.kqueue.close` not actually close a Kqueue file? Outside of test cases the only place where selectors get used is as a context manager. :/

Lukasa:
It certainly seems to, according to `dtruss` output from a quick test script:

```
kqueue(0x102F2E8D8, 0x102D13B30, 0x2000)		 = 3 0
close(0x3)		 = 0 0
```

SethMichaelLarson:
Dang, well then the stdlib selectors has the same problem we're having. Is it just KqueueSelector that has this problem? Maybe we'll have to close the fd manually rather than using the kqueue's interface.

Lukasa:
I don't think closing the FD manually will help: the kqueue selector *does* close it. You can see the FD being closed directly there.

SethMichaelLarson:
Sorry, thought you meant "seems that it doesn't close".

Lukasa:
Yeah, I worked that out once I saw your reply. ;)

SethMichaelLarson:
It's way too early for technical things here but I know this is the best time for both of our times to overlap. Do you have any ideas why it's holding on to these Kqueue objects? I'm really struggling to think of why they'd not be closing.

Lukasa:
@SethMichaelLarson I'm not sure I'm immediately worried about those kqueue objects yet. I'm much more concerned by the hundreds of sockets that aren't closed. Even if those kqueue objects were keeping their associated sockets open, and I don't know that they are, that doesn't come close to accounting for the 170-odd open and/or CLOSE_WAIT sockets that are dangling around.

Lukasa:
Put another way: I think whatever is keeping those sockets alive is likely to be keeping kqueue objects alive too. Annoyingly, there aren't dtrace probes available for PyPy, so I can't use dtrace to work out where FDs are being allocated from and not being freed. That's a pain in the neck. I suppose I can try to keep track of allocation/freeing of kqueue objects to see if I can work out where they're going.

SethMichaelLarson:
If I can be of any help I am here. :) I'll try my best!

Lukasa:
Agh, now I've lost my repro. That rather suggests that this is timing dependent somehow, though I'm really not sure how.

Lukasa:
Oh lol, no I didn't, I just forgot that I raised my ulimit in my shell. ;) Sometimes I wonder how people mistake me for a professional software developer. :grin:

SethMichaelLarson:
@Lukasa I actually thought to say something and ask you if you had forgotten to lower your ulimit but I didn't want to bug you. :)

Lukasa:
Ok, adding some basic logic to selectors to track when close is called shows that all the kqueue selectors are actually being closed. So that's not the issue.

SethMichaelLarson:
@Lukasa *Sigh of relief that I didn't necessitate a bug release branch*

Lukasa:
I mean, I think the core issue here is just that necessarily urllib3 has a tendency to "leak" sockets from most of the tests because the tests don't attempt to close their connection pools or pool managers. I haven't worked out why the process seems to be retaining so many objects though: even if I repeatedly call gc.collect() I can only free up about 18 FDs.

SethMichaelLarson:
@Lukasa Should an effort be made to clean up as many tests as possible that leak fds?

Lukasa:
Well, certainly some of the worst offenders could be cleaned up. Adding a test helper to get a connection pool or pool manager with cleanup wouldn't hurt.

SethMichaelLarson:
I'll see what I can do if you have other tasks to take on today?

Lukasa:
Go for it. =)

Lukasa:
Let's close this for now then: either raising the ulimit or doing #1100 would be preferable to this.

Issue #1100, opened 2017-01-20
Tests leak FDs which leads to test failures on machines with low ulimits.


SethMichaelLarson:
As was discussed a lot in #1098 many tests open sockets and don't provide a proper clean-up which results in many leaking FDs during a test execution. This leads to problems on machines with low soft ulimits such as Mac OS.

I'll be making an effort to change these tests to not leak FDs such as unclosed connections, pools, etc.

alexwlchan:
One observation from my testing: I’ve seen some FDs that leak intermittently. This can make it harder to tell if a patch succeeded in cleaning it up, or that was just a lucky run.

SethMichaelLarson:
I think this issue is mostly finished. Closing for now. Thanks @alexwlchan!

PR #1102, opened 2017-01-21
Add cleanup to pools and managers

SethMichaelLarson:
Work for #1100 by adding `.close` for all ConnectionPools and `.clear` for all PoolManagers.

Also noticed that in `TestUtil` there were a few duplicate keys in `parse_url_host_map` so I changed it to a list of tuples to make sure the duplicates weren't silently discarded.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -29,7 +29,6 @@ from urllib3.exceptions import (
     LocationParseError,
     TimeoutStateError,
     InsecureRequestWarning,
-    SSLError,
     SNIMissingWarning,
     InvalidHeader,
     UnrewindableBodyError,
@@ -84,8 +83,10 @@ class TestUtil(unittest.TestCase):
             'http://[2a00:1450:4001:c01::67]:80/test': ('http', '[2a00:1450:4001:c01::67]', 80),
 
             # More IPv6 from http://www.ietf.org/rfc/rfc2732.txt
-            'http://[fedc:ba98:7654:3210:fedc:ba98:7654:3210]:8000/index.html': ('http', '[fedc:ba98:7654:3210:fedc:ba98:7654:3210]', 8000),
-            'http://[1080:0:0:0:8:800:200c:417a]/index.html': ('http', '[1080:0:0:0:8:800:200c:417a]', None),
+            'http://[fedc:ba98:7654:3210:fedc:ba98:7654:3210]:8000/index.html': (
+                'http', '[fedc:ba98:7654:3210:fedc:ba98:7654:3210]', 8000),
+            'http://[1080:0:0:0:8:800:200c:417a]/index.html': (
+                'http', '[1080:0:0:0:8:800:200c:417a]', None),
             'http://[3ffe:2a00:100:7031::1]': ('http', '[3ffe:2a00:100:7031::1]', None),
             'http://[1080::8:800:200c:417a]/foo': ('http', '[1080::8:800:200c:417a]', None),
             'http://[::192.9.5.5]/ipng': ('http', '[::192.9.5.5]', None),
@@ -120,8 +121,10 @@ class TestUtil(unittest.TestCase):
             '173.194.35.7': ('http', '173.194.35.7', None),
             'HTTP://173.194.35.7': ('http', '173.194.35.7', None),
             'HTTP://[2a00:1450:4001:c01::67]:80/test': ('http', '[2a00:1450:4001:c01::67]', 80),
-            'HTTP://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:8000/index.html': ('http', '[fedc:ba98:7654:3210:fedc:ba98:7654:3210]', 8000),
-            'HTTPS://[1080:0:0:0:8:800:200c:417A]/index.html': ('https', '[1080:0:0:0:8:800:200c:417a]', None),
+            'HTTP://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:8000/index.html': (
+                'http', '[fedc:ba98:7654:3210:fedc:ba98:7654:3210]', 8000),
+            'HTTPS://[1080:0:0:0:8:800:200c:417A]/index.html': (
+                'https', '[1080:0:0:0:8:800:200c:417a]', None),
         }
         for url, expected_host in url_host_map.items():
             returned_host = get_host(url)
@@ -131,7 +134,8 @@ class TestUtil(unittest.TestCase):
         """Assert parse_url normalizes the scheme/host, and only the scheme/host"""
         test_urls = [
             ('HTTP://GOOGLE.COM/MAIL/', 'http://google.com/MAIL/'),
-            ('HTTP://JeremyCline:Hunter2@Example.com:8080/', 'http://JeremyCline:Hunter2@example.com:8080/'),
+            ('HTTP://JeremyCline:Hunter2@Example.com:8080/',
+             'http://JeremyCline:Hunter2@example.com:8080/'),
             ('HTTPS://Example.Com/?Key=Value', 'https://example.com/?Key=Value'),
             ('Https://Example.Com/#Fragment', 'https://example.com/#Fragment'),
         ]
@@ -154,7 +158,9 @@ class TestUtil(unittest.TestCase):
         ('#?/!google.com/?foo#bar', Url(path='', fragment='?/!google.com/?foo#bar')),
         ('/foo', Url(path='/foo')),
         ('/foo?bar=baz', Url(path='/foo', query='bar=baz')),
-        ('/foo?bar=baz#banana?apple/orange', Url(path='/foo', query='bar=baz', fragment='banana?apple/orange')),
+        ('/foo?bar=baz#banana?apple/orange', Url(path='/foo',
+                                                 query='bar=baz',
+                                                 fragment='banana?apple/orange')),
 
         # Port
         ('http://google.com/', Url('http', host='google.com', path='/')),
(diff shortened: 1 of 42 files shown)


Lukasa:
Hrm, looks like we have some problems with the queue module here. Want to look into them?

SethMichaelLarson:
@Lukasa Yeah I'll look at them unless you have time today. Won't have much time for open source today.

Lukasa:
There's no rush, it's a weekend. Take time for yourself. =D

SethMichaelLarson:
I think I figured out the `queue` issue with those two tests. Also added `unittest2.TestCase` substitution for all the tests that didn't have the substitution already and required `.addCleanup`. It would be great if you could run your FD test again on this @Lukasa if the CI passes to see if this helps the leaking situation at all? :)

SethMichaelLarson:
So I'm having trouble with GAE as the `import urllib3` comes inside the testcase which is what is causing the `queue` issue but it doesn't seem like I can just move it outside the testcase as then importing `google.appengine.api.prospective_search.prospective_search_stub` from within `nosegae` fails.

test/test_util.py
```diff
@@ -138,34 +139,34 @@ def test_parse_url_normalization(self):
...
+        ('http://google.com/', Url('http', host='google.com', path='/')),
+        ('http://google.com:80/', Url('http', host='google.com', port=80, path='/')),
+        ('http://google.com:80', Url('http', host='google.com', port=80)),
 
         # Auth
-        'http://foo:bar@localhost/': Url('http', auth='foo:bar', host='localhost', path='/'),
-        'http://foo@localhost/': Url('http', auth='foo', host='localhost', path='/'),
-        'http://foo:bar@baz@localhost/': Url('http', auth='foo:bar@baz', host='localhost', path='/'),
-        'http://@': Url('http', host=None, auth='')
-    }
```
Lukasa: Why are we changing this?

Lukasa:
So *why* is the late import causing the `queue` issue, sorry?

test/test_util.py
```diff
@@ -48,6 +48,7 @@
 # numbers used for timeouts
```
sigmavirus24: Every change in this file is completely unrelated to the rest of the PR. Please revert these.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -837,9 +856,9 @@ def socket_handler(listener):
                        ':9A:8C:B6:07:CA:58:EE:74:5E')
 
         def request():
+            pool = HTTPSConnectionPool(self.host, self.port,
```
sigmavirus24: Why is there so much code re-organization in this pull request? It doesn't belong.

test/with_dummyserver/test_no_ssl.py
```diff
@@ -8,21 +8,21 @@
 from dummyserver.testcase import (
         HTTPDummyServerTestCase, HTTPSDummyServerTestCase)
 
+import urllib3
```
sigmavirus24: This migration should happen in a separate PR

test/test_util.py
SethMichaelLarson: There's two keys with different values for `http://google.com/mail` and `http://google.com/` and that will silently drop one of the test cases.

test/with_dummyserver/test_no_ssl.py
SethMichaelLarson: If this import is inside the testcase there is an error on cleanup.

test/test_util.py
SethMichaelLarson: See my reply to @Lukasa above

SethMichaelLarson:
@Lukasa I don't know why importing urllib3 within the testcase makes the `queue` module within `urllib3` turn to `None` by clean-up time. I'll investigate further.

test/test_util.py
sigmavirus24: There are still spurious and unrelated changes in this file besides turning the dictionary into a list of tuples and removing calls to `dict.items()`

SethMichaelLarson: As I mentioned above, if you look inside that dictionary there is two duplicated keys with different values within that dictionary so for the entire time the test has been around two of the values were not being tested and silently discarded.

```python
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
>>> a = {'a': 1, 'a': 2}
>>> a
{'a': 2}
```

By changing this to a list of tuples it preserves both test cases and changing to a list of tuples necessitates removing the `.items()` the bottom loop where these values are used.

I'm open to moving this to it's own PR but as it stands currently this test is dropping some of it's test cases silently, something I figured we'd want to avoid.

Lukasa: Ian is saying that there are spurious changes here *beyond* the change to tuple, which he's just OK'd.

sigmavirus24: That's still unrelated to ensuring a pool manager is closed, but I'm less averse to it being included here for expediency. It also isn't explained in any of the commits or their messages (all the commits are a single line according to the GitHub UI).

So future developers will find this and have no clear understanding of the *why* unless they very closely read the diff. It should at least have been split into its own commit for documenting in our history.

SethMichaelLarson: The only change beyond the tuple change and the `.items()` change for this file is adding a line of whitespace on L51 and deleting a line of whitespace on L374. Could you point me to what spurious changes in this file you want to see reverted?

test/with_dummyserver/test_socketlevel.py
SethMichaelLarson: This is done because there's a chance that pool is not defined by the time we get outside of the `try-catch` block (eg If an exception occurs during the creation of the `HTTPSConnectionPool`). I can revert this change if that's what's wanted.

test/test_util.py
sigmavirus24: > And I'm sorry for not including the change as a separate commit. I will try to be better.

You can still split it out into its own commit.

> what spurious changes in this file you want to see reverted?

The whitespace changes. They're stylistic and have no bearing on this PR. If you want to submit yet another PR to re-flake8 this project, that's fine, but that should be separate.

Lukasa: @SethMichaelLarson There's no need to feel bad about not splitting the commit out. Things are overlooked all the time, and that's why we do code review. =) And remember: @sigmavirus24 and I have exactly this kind of experience on the other side of code review all the time (hit me up on Twitter and I'll point out my current OpenSSL code review saga).

SethMichaelLarson: I'll just revert this whole file and open a separate PR for the dictionary issue. If flaking the unit tests is something that is desired I will put all the work into one PR. I find it difficult to work on the tests and find actual deficiencies when my IDE is lit up like a Christmas tree with linting errors.

Lukasa: By all means, file a PR with the linting errors. Merging things like that is usually a box-ticking exercise that I can do with half of my brain turned off, so I'm not too worried about it. =D

test/with_dummyserver/test_socketlevel.py
Lukasa: Yeah, this one has to stay.

test/test_util.py
SethMichaelLarson: Okay, as long as it's no trouble. I'll dedicate some time to that task as well as cleaning up this PR.

Lukasa: No trouble at all.

Lukasa:
So, an FYI: this has become more important in the v2 branch as I refactor to use h11 more closely, because all our connection objects now use a selector and a socket. This doubles the FD use per connection, which means that I can now reproduce the "too many files" error consistently in all tests.

SethMichaelLarson:
I'll get time tonight to work on this branch @Lukasa :) And I'm glad the selectors are useful as well. It would be nice if somehow all connections within a pool could share a selector! :D

Lukasa:
> It would be nice if somehow all connections within a pool could share a selector! :D

So while they *could*, it gets tricky fast: you need to use semaphores to manage the selector because connection pools are thread-safe, and so their shared selector would need to be as well. This is not totally trivial, and incurs a massive performance cost (locking in Python is very slow).

A potential optimisation would be to have one selector per thread in thread-local storage, which would vastly reduce the overhead. That is something we should look into as v2 progresses.

SethMichaelLarson:
Excited to have selectors serve a higher purpose than just waiting for a single socket! :)

Lukasa:
To be clear, the reason we're using selectors for this is twofold. Firstly, and most importantly in terms of user annoyance, they allow us to *finally* consider implementing "total" timeouts in a way that makes sense (that is, "don't wait more than `n` seconds for the entire response").

But the most important reason for me is that they fix a number of problems relating to the fact that the server can respond to us before we're finished uploading. This allows us to send data while we watch for the possibility that the server has already responded. We've not been able to do this before, and it's a huge win. It also opens other use-cases like this, such as support for Expect: 100-continue.

SethMichaelLarson:
@Lukasa I will say the biggest annoyance I have with Python 2.x is that everything that should have a `timeout` parameter (`threading.Lock.acquire`, `subprocess.Popen.communicate`, ...) doesn't, and it makes it really annoying to implement those actual timeouts.

Lukasa:
Hrm, Travis seems to have lost this commit.

Lukasa:
Looks like we have some flake8 problems. =D

SethMichaelLarson:
@Lukasa I'll fix those later today :)

What are your thoughts on leaving the GAE tests as they were? When I pull the import outside the test function then nosegae has trouble importing and when I have the import inside the test function and do cleanup then we run into the `queue` module turning to `None` before cleanup. :/

Lukasa:
We can probably start by leaving the GAE tests as they are for now. Let's make progress as gently as we can. =)

SethMichaelLarson:
Looks like (transient?) failures on Python 3.4 Linux and Python 3.3 OSX. I haven't seen the OSX error before but it looks like a read timeout on localhost?

SethMichaelLarson:
Python 3.3 OSX still failing, another transient issue. :/

SethMichaelLarson:
Yay! Our CI has come through in the end :)

sigmavirus24:
I'm not happy because I think this is still indicative of a problem that can and will occur on a user's system, but I have no objection to the changes in this pull request.

Lukasa:
> I'm not happy because I think this is still indicative of a problem that can and will occur on a user's system, but I have no objection to the changes in this pull request.

So, to be clear, it *absolutely is*. This is a problem we've known about for a very long time, namely that because we pool connections we necessarily keep them alive until we're GC'd and rely on the GC to do that cleaning up.

We could avoid this in some cases by making things context managers, but ultimately that only solves it for well-bounded lifetimes. Otherwise, the problem is present everywhere. However, it is worse here because we create a *bunch* of connections in a tight loop during a test case.

PR #1103, opened 2017-01-23
Fix all flake/linter errors within tests

SethMichaelLarson:
Work so far on fixing all flake / lint errors in the `test/` directory. Also adds that directory to the list of locations that are checked by `flake8` by the CI.

This PR does not yet check for line too long because of how many instances there are of that error, I will work on that tonight.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,7 +6,7 @@ cover-package=urllib3
 cover-erase=true
 
 [flake8]
-exclude=./docs/conf.py,./test/*,./urllib3/packages/*
+exclude=./docs/conf.py,./urllib3/packages/*
 max-line-length=99
 
 [wheel]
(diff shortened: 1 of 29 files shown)


SethMichaelLarson:
The tests should fail, I'm going to make the line-to-long commit last because it's most likely to garner some comments. 

Also the dictionary duplicated keys is a flake error too apparently so I think it'd be okay to fix it in this PR?

SethMichaelLarson:
Would like a review on the last commit about E501. I didn't change any code functionality but some people have strong opinions how lines should wrap. :)

SethMichaelLarson:
That should be the last commit for this PR. :)

Lukasa:
Awesome, LGTM. Let's merge this.

Issue #1104, opened 2017-01-25
AttributeError: 'module' object has no attribute 'epoll'


neiljerram:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/poll.py", line 115, in wait
    listener.cb(fileno)
  File "/usr/local/lib/python2.7/dist-packages/eventlet/greenthread.py", line 214, in main
    result = function(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/networking_calico/agent/dhcp_agent.py", line 553, in loop
    super(SubnetWatcher, self).loop()
  File "/usr/local/lib/python2.7/dist-packages/networking_calico/logutils.py", line 21, in wrapped
    return fn(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/networking_calico/etcdutils.py", line 250, in loop
    response = self.wait_for_etcd_event()
  File "/usr/local/lib/python2.7/dist-packages/networking_calico/etcdutils.py", line 344, in wait_for_etcd_event
    timeout=self.etcd_timeout)
  File "/usr/local/lib/python2.7/dist-packages/etcd/client.py", line 562, in read
    timeout=timeout)
  File "/usr/local/lib/python2.7/dist-packages/etcd/client.py", line 820, in wrapper
    params=params, timeout=timeout)
  File "/usr/local/lib/python2.7/dist-packages/etcd/client.py", line 890, in api_execute
    preload_content=False)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/request.py", line 66, in request
    **urlopen_kw)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/request.py", line 87, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/poolmanager.py", line 244, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 588, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/usr/local/lib/python2.7/dist-packages/urllib3/connectionpool.py", line 241, in _get_conn
    if conn and is_connection_dropped(conn):
  File "/usr/local/lib/python2.7/dist-packages/urllib3/util/connection.py", line 27, in is_connection_dropped
    return bool(wait_for_read(sock, timeout=0.0))
[...]

PR #1105, opened 2017-01-25
Selectors detect a monkey-patched select module

SethMichaelLarson:
PR to fix issue when the `select` module is patched after compile time by gevent, eventlet, etc as found in #1104 This patch doesn't contain any tests but passes all current tests, will add some later today. I'm sure there'll be some coverage issues as well that will be resolved. :)

This patch definitely makes `selectors.py` a bit of an abomination, open to better suggestions. :) cc: @Lukasa

### DIFF
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -13,6 +13,9 @@ import sys
 import time
 from collections import namedtuple, Mapping
 
+# Set a canary to detect if the select module is monkey patched.
+select.__patch_canary = True
+
 try:
     monotonic = time.monotonic
 except (AttributeError, ImportError):  # Python 3.3<
@@ -23,6 +26,10 @@ EVENT_WRITE = (1 << 1)
 
 HAS_SELECT = True  # Variable that shows whether the platform has a selector.
 _SYSCALL_SENTINEL = object()  # Sentinel in case a system call returns None.
+
+
+def _no_selector(_):
+    raise ValueError("Platform does not have a selector")
 
 
 class SelectorError(Exception):
@@ -282,8 +289,9 @@ class BaseSelector(object):
 
 
 # Almost all platforms have select.select()
+SelectSelector = None
 if hasattr(select, "select"):
-    class SelectSelector(BaseSelector):
+    class _SelectSelector(BaseSelector):
         """ Select-based selector. """
         def __init__(self):
             super(SelectSelector, self).__init__()
@@ -331,9 +339,12 @@ if hasattr(select, "select"):
                     ready.append((key, events & key.events))
             return ready
 
+    SelectoSelector = _SelectSelector
 
+
+PollSelector = None
 if hasattr(select, "poll"):
-    class PollSelector(BaseSelector):
+    class _PollSelector(BaseSelector):
         """ Poll-based selector """
         def __init__(self):
             super(PollSelector, self).__init__()
@@ -384,9 +395,12 @@ if hasattr(select, "poll"):
 
             return ready
 
+    PollSelector = _PollSelector
 
+
+EpollSelector = None
 if hasattr(select, "epoll"):
-    class EpollSelector(BaseSelector):
+    class _EpollSelector(BaseSelector):
         """ Epoll-based selector """
         def __init__(self):
             super(EpollSelector, self).__init__()
@@ -451,9 +465,12 @@ if hasattr(select, "epoll"):
             self._epoll.close()
             super(EpollSelector, self).close()
 
+    EpollSelector = _EpollSelector
 
+
+KqueueSelector = None
 if hasattr(select, "kqueue"):
-    class KqueueSelector(BaseSelector):
+    class _KqueueSelector(BaseSelector):
         """ Kqueue / Kevent-based selector """
         def __init__(self):
             super(KqueueSelector, self).__init__()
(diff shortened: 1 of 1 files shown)


urllib3/util/selectors.py
```diff
@@ -534,20 +551,73 @@ def close(self):
...
+    global DefaultSelector
+    global KqueueSelector
+    global EpollSelector
+    global PollSelector
+    global SelectSelector
+
+    global HAS_SELECT
+
+    patched = False
+    if not hasattr(select, '__patch_canary'):  # Platform-specific: Gevent, Eventlet
```
SethMichaelLarson: Definitely should remove this `#Platform-specific` comment when new tests are written. :)

urllib3/util/selectors.py
```diff
@@ -534,20 +551,73 @@ def close(self):
             self._kqueue.close()
             super(KqueueSelector, self).close()
 
+    KqueueSelector = _KqueueSelector
+
+
+def _detect_select_patch(*args, **kwargs):
+    # If monkey-patching occurs then the canary won't be
```
SethMichaelLarson: Should move this comment to a more appropriate location.

urllib3/util/selectors.py
```diff
@@ -534,20 +551,73 @@ def close(self):
...
+    # defined within the new select module. Because the
+    # module was patched we now have to re-evaluate which
+    # selectors are available.
```
Lukasa: I don't think this is true: I don't think the module gets replaced, I think the attributes get deleted.

SethMichaelLarson: Ah! I was misinterpreting how eventlet works, looking at https://github.com/eventlet/eventlet/blob/master/eventlet/patcher.py#L290 confirms that you're correct. Perhaps I'll have to set this attribute on `select.select` instead of just `select`. If `select.select` isn't available then there's no selector anyway so no reason to check for patching.

Lukasa: I'm *extremely* nervous about trying to solve a problem with monkeypatching by doing our own monkeypatching.

SethMichaelLarson: This is true, is there any alternative except do nothing?

SethMichaelLarson: I guess the alternative is determining `DefaultSelector` on each call as you mentioned in the issue.

Lukasa: We could maybe do some descriptor magic to try to lazily initialize this module? Or turn all the selectors into factory functions that do lazy initialization.

SethMichaelLarson: I could try the factory approach and see what that looks like.

sigmavirus24:
What if we simply declare a no-op interface if `select` has had these things removed? (Also, instead of using `hasattr` using `getattr`)

Lukasa:
> What if we simply declare a no-op interface if select has had these things removed?

When do we detect the removal?

All of this difficulty is boiling down to the fact that we presumably don't want to do `hasattr`/`getattr` checks each time we create a selector. If we do want that this problem is trivially solved. If we don't, the question we have to answer is *when* do we do them.

sigmavirus24:
So I'd argue that it should be done each time because at one point we may be fine and at another gevent may have kicked in. It's a little unreasonable to say that users have to be sure to invoke gevent patching prior to ever importing requests/urllib3 and it's best if we can handle both cases.

To that effect, I'd argue that `HAS_SELECT` should instead be a function called `can_use_select` or something that does the checks for us. Then in `util.wait` we use that to determine if we can or cannot use `util.selectors`. I'm also not thrilled about `HAS_SELECT` being used outside of `wait` but whatever. Given the design, I'd have expected that attributes in `util.selectors` would have been relatively private to `util` and things in `util.wait` would be the internal public API. Something like `can_wait()` would be exposed in that case. That's a separate complaint though

Lukasa:
> So I'd argue that it should be done each time because at one point we may be fine and at another gevent may have kicked in.

Some relevant background here: right now I am sitting on a v2 branch that creates a selector for every connection. That means we'll create selectors at least once per connection, and when using PyOpenSSL we'll do it even more frequently. Are we happy bearing this overhead each time?

sigmavirus24:
> Are we happy bearing this overhead each time?

What kind of penalty is there? I'm not intimately familiar with the penalty of creating selectors each time.

Lukasa:
The penalty is the cost of the hasattr chain, I think. Assuming we define the classes at import time and then only use them based on the hasattr result, that's the overhead.

SethMichaelLarson:
@Lukasa If we do go that route we could have clever boolean logic to determine which selector to use to reduce the number of `hasattr` checks based on previous runs through `DefaultSelector`? For instance if the platform doesn't have poll we can assume it also won't have kqueue or epoll because it's probably Windows or a patched `select`.

sigmavirus24:
So what if we do something where we store off the results of `getattr(select, ..., None)`. Once `gevent` has monkey-patched things, we'll still have those references. Could we/would it make sense to use those stored references instead of `select.<...>`? Alternatively, we could use those stored references to reduce the number of things to check. If `select_epoll` is not None, we should check that it's still not `None` prior to continuing. That means fewer checks each time, yes?

Lukasa:
> Once gevent has monkey-patched things, we'll still have those references. Could we/would it make sense to use those stored references instead of select.<...>?

No. gevent is patching these to prevent us from impeding with their I/O loop. If we use them, we'll break the monkey patch and stop working on gevent. That's bad.

>  If select_epoll is not None, we should check that it's still not None prior to continuing. That means fewer checks each time, yes?

I think so, yes.

SethMichaelLarson:
Yeah we can't hold the references, I had that thought at first and then realized it would break monkey patching even more than it is currently.

sigmavirus24:
> Yeah we can't hold the references

We can't hold and then call them. We can hold them for the purposes of intelligently knowing when we can use them though.

Also to clarify my statement, if we have a selector that's not None and it becomes None (or we choose to use a better sentinel for "non-existent") then we should clear that reference to have fewer checks on future calls. Agreed? I feel like checking each time and verifying we can still use a selector is the better method. Granted it may add time to go through some conditionals but I think it's worth it.

Not every user will be able to guarantee that they can monkey-patch with gevent prior to using urllib3/requests.

Lukasa:
I can get behind that, sure.

SethMichaelLarson:
We could simplify this current PR a lot more if it were also allowed to not have the same behaviour of `selectors.py` which is to not define selectors that can't be used. This PR doesn't break that promise because upon realising that a selector shouldn't be available it will remove it from the module. If instead all selectors are always defined and then `DefaultSelector` does the choosing this PR could be greatly simplified.

Lukasa:
Let's abandon that promise. It's silly. urllib3 will never use anything but DefaultSelector except for testing purposes anyway.

SethMichaelLarson:
@Lukasa Yeah, I realize this now after wondering for a bit why I kept that promise haha

SethMichaelLarson:
I'm closing this in favor of #1128

PR #1107, opened 2017-02-01
Add support for X509 and PKey for load_cert_chain instead of just a filepath

poerhiza:
Add support for X509 and PKey for load_cert_chain instead of just a filepath


Lukasa:
Thanks for this patch!

I am inclined to say that we shouldn't merge this. I don't really want to add API surface that allows users to write code that is *dependent* on having PyOpenSSL injected into urllib3. Instead, I recommend that you use `create_urllib3_context` to create a PyOpenSSL SSLContext, and then use the methods directly on it to load the certificates you need to load.

PR #1108, opened 2017-02-01
Break firmly away from httplib

Lukasa:
Ok, this new major commit series breaks firmly away from httplib. It introduces a brand new connection class that is vastly stripped down compared to what httplib offered, including a very limited set of functionality. It rewrites that connection to use selectors rather than blocking socket I/O, enabling us to abort request sending for early responses (a regularly requested feature).

It also allows us to begin blasting through the code removing stuff, which is a huge win.

This is something that I'd like people to look over at a high level. Obviously, detailed review is difficult, but another set of eyes on this would be really helpful.


Lukasa:
As a point of pride I'd like to note that this commit manages to barely introduce any lines of code. This is a pretty big win!

urllib3/base.py
```diff
@@ -0,0 +1,106 @@
...
+        self.headers = HTTPHeaderDict(headers)
+
+        #: The request body. This is allowed to be one a few kind of objects:
+        #:    - A byte string.
+        #:    - A "readable" object.
+        #:    - An iterable of byte strings.
+        #:    - A text string (not recommended, auto-encoded to UTF-8)
+        self.body = body
+
+    def add_host(self, host, port, scheme):
```
SethMichaelLarson: Might be useful to add a default value to scheme to inherit from `self.target`.

Lukasa: `target` should almost never be a full URL.

urllib3/response.py
```diff
@@ -394,20 +401,23 @@ def getheader(self, name, default=None):
...
-        elif hasattr(self._fp, 'closed'):
-            return self._fp.closed
+        elif hasattr(self._fp, 'complete'):
+            return self._fp.complete
         else:
-            return True
+            return False
 
     def fileno(self):
+        # TODO: Do we need this?
```
SethMichaelLarson: Having `fileno()` makes it selectable. :)

urllib3/base.py
SethMichaelLarson: Ah, nevermind then!

urllib3/sync_connection.py
```diff
@@ -0,0 +1,556 @@
...
+    """
+    # TODO: revisit
+    for name, value in headers:
+        if isinstance(name, six.text_type):
+            name = name.encode('ascii')
+
+        if isinstance(value, six.text_type):
+            value = value.encode('latin-1')
+        elif isinstance(value, int):
+            value = str(value).encode('ascii')
```
SethMichaelLarson: I'm sure this is just for consistency but `b'%d' % value` also works here.

urllib3/sync_connection.py
```diff
@@ -0,0 +1,556 @@
...
+        Close this connection, suitable for being re-added to a connection
+        pool.
+        """
+        if self._sock is not None:
+            sock, self._sock = self._sock, None
+            sock.setblocking(1)
+            sock.close()
+
+        if self._selector is not None:
+            selector, self._selector = self._selector, None
```
SethMichaelLarson: Never seen this construction but this is to both ensure that the selector is collected as well as raising any exception that comes out of `selector.close()`?

urllib3/base.py
```diff
@@ -0,0 +1,106 @@
...
+
+        #: The request body. This is an iterable of bytes, and *must* be
+        #: iterated if the connection is to be preserved.
+        self.body = body
+
+        #: The HTTP version of the response. Stored as a bytestring.
+        self.version = version
+
+    @property
+    def complete(self):
```
nateprewitt: Sorry that I haven't had a moment to pull this down yet to check, but is `body` guaranteed to have `complete`? The BytesIO objects used in the tests don't seem to. It looks like you're catching for `AttributeError` in certain places for it but not everywhere.

haikuginger:
Heading on vacation tomorrow, so I'll have a chance to go over this (and also attend to my own work) at that point in time.

urllib3/response.py
Lukasa: It makes it *misleadingly* selectable. I don't think we want users to select on our response object, do we?

urllib3/sync_connection.py
Lukasa: It's actually more than that, it's to ensure that the selector is removed from the object so that it cannot subsequently be accessed. It essentially guarantees that once we've closed a connection there is a defined state.

urllib3/base.py
Lukasa: No, it's absolutely not guaranteed to have complete. However, note that complete is present one level up as well, which may explain the capturing.

Whether we want to tolerate BytesIO is an ongoing question in my mind.

urllib3/connectionpool.py
```diff
@@ -279,15 +294,11 @@ def _put_conn(self, conn):
         if conn:
             conn.close()
 
-    def _validate_conn(self, conn):
+    def _start_conn(self, conn, connect_timeout):
```
sigmavirus24: Should be a staticmethod, no?

urllib3/connectionpool.py
```diff
@@ -742,82 +751,76 @@ def __init__(self, host, port=None,
...
-        return conn
-
-    def _prepare_proxy(self, conn):
-        """
-        Establish tunnel connection early, because otherwise httplib
-        would improperly set Host: header to proxy's IP:port.
-        """
-        conn.set_tunnel(self.host, self.port, self.proxy_headers)
-        conn.connect()
+    def _build_context(self, context, keyfile, certfile, cert_reqs,
```
sigmavirus24: Also static, yes?

urllib3/response.py
sigmavirus24: > I don't think we want users to select on our response object, do we?

No I don't think we do. Especially not since we're selecting on the underlying parts of this.

urllib3/sync_connection.py
sigmavirus24: Not on all supported versions of Python 3 iirc. I'm fairly certain that was only added back in 3.5

urllib3/response.py
SethMichaelLarson: No we definitely don't because we're already doing the lifting for them.

urllib3/sync_connection.py
SethMichaelLarson: Yep totally forgot about that! I'm so spoiled by 3.5 :)

urllib3/connectionpool.py
Lukasa: Doesn't look like it: it uses `self` a bunch.

urllib3/connectionpool.py
Lukasa: This one should absolutely be, except I don't believe in staticmethods so we'll just make it a function instead.

SethMichaelLarson:
:tada:

Lukasa:
Look at all that *code coverage*. Woooooo!

So, when I merge this I have a big, big OmniFocus project full of cleanups, but this puts us in a much nicer place.

My goals include breaking up a bunch of functions so they can easily be overridden with async versions (i.e. moving most of the work out to helper functions), propagating the Request/Response objects higher up the call stack (e.g. having them eventually be created by the PoolManager), and then we get onto the meaty stuff. Most notably, I want to reconsider our use of "timeouts" to have them instead be "deadlines": that is, we wait no more than *n* seconds to complete a connect or a read, regardless of how many I/O ops it takes.

But yes, it's 🎉  time!

SethMichaelLarson:
@Lukasa Sounds like a good plan. It would be nice if certain parts of this could be turned into issues for the community to tackle but I understand that these types of changes need to be fairly monolithic. :)

Lukasa:
I think they absolutely can. I take it it was the timeout bit that caught your attention? ;)

SethMichaelLarson:
Oh nothing in particular! Just don't want you to feel like you have to do all the work yourself. :) I've got plenty just trying to complete my already open PRs here in addition to all the other OS I do.

PR #1115, opened 2017-02-12
Fix Python 2.6 builds

Lukasa:
So, cryptography currently doesn't have a wheel build for Python 2.6, which is causing our master branch tests to fail. This is because we're setting CFLAGS and LDFLAGS for the original build, which are unfortunately saved and re-used *badly* for subsequent pip builds.

I'm going to try to fix it.

### DIFF
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -10,12 +10,6 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     brew outdated openssl || brew upgrade openssl
     brew install openssl@1.1
 
-    # We can't use OpenSSL@1.1 for the standard library, because it's
-    # unsupported, but we want it for PyOpenSSL. So use regular openssl here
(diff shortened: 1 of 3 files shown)


sigmavirus24:
I suspect this will work just fine, my only request is that the commit message be updated to explain why this is okay (e.g., that the wheels don't need these environment variables set and they break things on OSX Py2.6)

Lukasa:
Yeah, right now this is mostly exploratory: I have my suspicions that this *won't* work and then we'll want to investigate why. But then yes, I'm happy to go back and fix that.

Lukasa:
And indeed it does not work.

Lukasa:
@sigmavirus24 Mind doing a review here?

Lukasa:
(Or indeed anyone else who has nothing to do of a Sunday.)

PR #1116, opened 2017-02-13
Find supported TLS version in tests.

### LINKED ISSUE #1114: Reduce reliance on specific TLS/SSL constants
There are a bunch of tests in `test.with_dummyserver.with_https` that rely on having specific OpenSSL constants present (e.g. Some of the ones in TestHTTPS_TLSv1. Ideally these tests would be refactored to be more tolerant of not having some constants around by picking the ones they can get, rather than being specific. 

Related: #1113

Lukasa:
Resolves #1114.

This should be a solid forward-looking revision to these tests to ensure that they can run on a variety of systems without difficulty. We shouldn't have tests fail just because OpenSSL has been compiled slightly weirdly, especially as urllib3 itself has no reliance on these constants being present.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -41,6 +41,7 @@ from urllib3.util import is_fp_closed, ssl_
 from urllib3.packages import six
 
 from . import clear_warnings
+from .tls_helpers import TLS_VERSION, TLS_VERSION_STRING
 
 # This number represents a time in seconds, it doesn't mean anything in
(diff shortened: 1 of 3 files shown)


Lukasa:
Further consideration suggests this might be a bad idea.

Ultimately, without more collusion from the web server involved a whole bunch of tests will fail, because they rely on a certain shape of Client Hello packet that must have TLSv1 in the front.

For this reason, I'm pretty inclined to just leave these tests alone.

PR #1119, opened 2017-02-15
Re-add PyOpenSSL testing.

Lukasa:
This slipped through code review of #1103.


SethMichaelLarson:
Damn it. Sorry @Lukasa, this is what happens when I trust my tools more than use my brain to think about what's happening. I'll look over #1103 again for any other cases like this.

Lukasa:
Heh, no need to apologise. It's easily done, and importing just for side-effects is not generally a good practice.

haikuginger:
Thanks @Lukasa!

✨🍰✨

Issue #367, opened 2014-04-03
Differences between standard SSL and Pyopenssl (ZeroReturnError, SysCallError)


mthurman:
We use urllib3 via requests with pyopenssl so we can have SNI support in python 2.7.3.

Since we switched, we've noticed more errors from requests. Specifically:
- ZeroReturnError
- SysCallError: (104, 'Connection reset by peer')
- SysCallError: (-1, 'Unexpected EOF')

It seems like these should all be caught somehow in contrib/pyopenssl.py so at the requests level, we don't have to worry about pyopenssl being different than the standard ssl module.

Here's a few of the ideas I'm considering:
- in `fileobject.read`, everywhere we are catching `WantReadError` also catch `ZeroReturnError` and make sure that data is an empty string and continuing
- In `inject_into_urllib3`, setting `connection.BaseSSLError = OpenSSL.SSL.Error`

Based on my limited understanding of pyopenssl (and python's standard _ssl.c) I'm probably missing something in trying to unify these interfaces. If somebody with more familiarity with this stuff can point me in the right direction I can try to formalize this stuff into a pull request.

Here's a gist of how I started investigating the `ZeroReturnError` specifically: https://gist.github.com/mthurman/9963741
##

lukas-gitl:
Any progress on this?

SethMichaelLarson:
@lukas-gitl Considering it's been 3 years since this was originally reported it may be useful to open up a new issue and then link to this issue if you're experiencing this. :)

IvanLauLinTiong:
pyOpenSSL is deprecated and will be removed in future release version 2.x (https://github.com/urllib3/urllib3/issues/2691).

PR #1125, opened 2017-02-25
Wrap OpenSSL SysCallError on send as well as recv

patstew:
I was having a similar issue as described in #367, occasionally getting "SysCallError: (104, 'Connection reset by peer')" on sending through the requests library. As I understand it, it should be intercepted and wrapped here.


Lukasa:
This looks reasonable enough to me. It'd be nice to have a test for this, but we don't have a good framework for testing this module right now and I am not going to ask you to start one for this PR. 

Thanks! 👍

PR #1128, opened 2017-03-06
Detect monkey-patched select module

SethMichaelLarson:
This is another implementation for solving issue #1104. #1105 is too complex because it was trying to keep promises of the Python stdlib selectors modules that aren't relevant to urllib3.
This change is simpler and should make for a more comfortable change. :)

This change doesn't detect monkey-patches that occur after the first call to `DefaultSelector` but the major problem seems to be patches during import time so I think this is a good compromise.

Should hopefully make it into next requests release (?) as I've seen more than one project blacklisting upgrading requests until issue #1104 is resolved even though it's not really our fault.

### DIFF
--- a/test/test_selectors.py
+++ b/test/test_selectors.py
@@ -723,8 +723,8 @@ class TestUniqueSelectScenarios(BaseSelectorTestCase):
     def test_select_module_defines_does_not_implement_poll(self):
         # This test is to make sure that if a platform defines
         # a selector as being available but does not actually
-        # implement it (kennethreitz/requests#3906) does not
-        # make DefaultSelector() fail.
+        # implement it (kennethreitz/requests#3906) then
+        # DefaultSelector() does not fail.
 
         # Reset the _DEFAULT_SELECTOR value as if using for the first time.
         selectors._DEFAULT_SELECTOR = None
--- a/test/test_selectors.py
+++ b/test/test_selectors.py
@@ -2,6 +2,7 @@ from __future__ import with_statement
 import errno
 import os
 import psutil
+import select
 import signal
 import socket
 import sys
@@ -120,7 +121,6 @@ class TimerMixin(object):
 @skipUnless(selectors.HAS_SELECT, "Platform doesn't have a selector")
 class BaseSelectorTestCase(unittest.TestCase, AlarmMixin, TimerMixin):
     """ Implements the tests that each type of selector must pass. """
-    SELECTOR = selectors.DefaultSelector
 
     def make_socketpair(self):
         rd, wr = socketpair()
@@ -135,9 +135,17 @@ class BaseSelectorTestCase(unittest.TestCase, AlarmMixin, TimerMixin):
         return rd, wr
 
     def make_selector(self):
-        s = self.SELECTOR()
+        s = selectors.DefaultSelector()
         self.addCleanup(s.close)
         return s
+
+    def setup_select_module(self, selector):
+        selectors._DEFAULT_SELECTOR = None
+        for s in ['select', 'poll', 'epoll', 'kqueue']:
+            if s != selector and hasattr(select, s):
+                old_selector = getattr(select, s)
+                delattr(select, s)
+                self.addCleanup(setattr, select, s, old_selector)
 
     def standard_setup(self):
         s = self.make_selector()
@@ -681,24 +689,93 @@ class ScalableSelectorMixin(object):
         self.assertEqual(limit_nofile // 2, len(s.select()))
 
 
+@skipUnless(selectors.HAS_SELECT, "Platform doesn't have a selector.")
+class TestUniqueSelectScenarios(BaseSelectorTestCase):
+    def test_select_module_patched_after_import(self):
+        # This test is to make sure that after import time
+        # calling DefaultSelector() will still give a good
+        # return value. This issue is caused by gevent, eventlet.
+
+        # Now remove all selectors except `select.select`.
+        self.setup_select_module('select')
+
+        # Make sure that the selector returned only uses the selector available.
+        selector = self.make_selector()
+        self.assertIsInstance(selector, selectors.SelectSelector)
+
+    @skipUnless(hasattr(errno, 'ENOSYS'), 'Platform does not have errno.ENOSYS.')
+    def test_select_module_defines_does_not_implement_poll(self):
+        # This test is to make sure that if a platform defines
+        # a selector as being available but does not actually
+        # implement it (kennethreitz/requests#3906) does not
+        # make DefaultSelector() fail.
+
+        # Reset the _DEFAULT_SELECTOR value as if using for the first time.
+        selectors._DEFAULT_SELECTOR = None
+
+        had_poll = hasattr(select, 'poll')
+
+        # Remove all selectors except `select.select`.
+        self.setup_select_module('select')
+
+        # Now we're going to patch in a bad `poll`.
+        class BadPoll(object):
+            def poll(self, timeout):
+                raise OSError(errno.ENOSYS)
+
+        select.poll = BadPoll
+        if not had_poll:
+            self.addCleanup(delattr, select, 'poll')
+
+        selector = self.make_selector()
+        self.assertIsInstance(selector, selectors.SelectSelector)
+
+    @skipUnless(hasattr(errno, 'ENOSYS'), 'Platform does not have errno.ENOSYS.')
+    def test_select_module_defines_does_not_implement_epoll(self):
+        # Same as above test except with `select.epoll`.
+
+        # Reset the _DEFAULT_SELECTOR value as if using for the first time.
+        selectors._DEFAULT_SELECTOR = None
+
+        had_epoll = hasattr(select, 'epoll')
+
+        # Remove all selectors except `select.select`.
+        self.setup_select_module('select')
+
+        # Now we're going to patch in a bad `epoll`.
+        def bad_epoll(*args, **kwargs):
+            raise OSError(errno.ENOSYS)
+
+        select.epoll = bad_epoll
+        if not had_epoll:
+            self.addCleanup(delattr, select, 'epoll')
+
+        selector = self.make_selector()
+        self.assertIsInstance(selector, selectors.SelectSelector)
+
+
 @skipUnless(hasattr(selectors, "SelectSelector"), "Platform doesn't have a SelectSelector")
 class SelectSelectorTestCase(BaseSelectorTestCase):
-    SELECTOR = getattr(selectors, "SelectSelector", None)
+    def setUp(self):
+        self.setup_select_module('select')
 
 
 @skipUnless(hasattr(selectors, "PollSelector"), "Platform doesn't have a PollSelector")
 class PollSelectorTestCase(BaseSelectorTestCase, ScalableSelectorMixin):
-    SELECTOR = getattr(selectors, "PollSelector", None)
+    def setUp(self):
+        self.setup_select_module('poll')
 
 
 @skipUnless(hasattr(selectors, "EpollSelector"), "Platform doesn't have an EpollSelector")
 class EpollSelectorTestCase(BaseSelectorTestCase, ScalableSelectorMixin):
-    SELECTOR = getattr(selectors, "EpollSelector", None)
+    def setUp(self):
+        self.setup_select_module('epoll')
 
 
 @skipUnless(hasattr(selectors, "KqueueSelector"), "Platform doesn't have a KqueueSelector")
 class KqueueSelectorTestCase(BaseSelectorTestCase, ScalableSelectorMixin):
-    SELECTOR = getattr(selectors, "KqueueSelector", None)
+    def setUp(self):
+        self.setup_select_module('kqueue')
 
 
 @skipUnless(hasattr(selectors, "SelectSelector"), "Platform doesn't have a SelectSelector")
--- a/test/test_selectors.py
+++ b/test/test_selectors.py
@@ -50,6 +50,11 @@ TOLERANCE = 0.75
 TRAVIS_CI = "TRAVIS" in os.environ
 
 
+skipUnlessHasSelector = skipUnless(selectors.HAS_SELECT, "Platform doesn't have a selector")
+skipUnlessHasENOSYS = skipUnless(hasattr(errno, 'ENOSYS'), "Platform doesn't have errno.ENOSYS")
+skipUnlessHasAlarm = skipUnless(hasattr(signal, 'alarm'), "Platform doesn't have signal.alarm()")
+
+
 class AlarmThread(threading.Thread):
     def __init__(self, timeout):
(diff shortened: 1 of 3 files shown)


Lukasa:
While we're here waiting for builds, do you mind adding a changelog note?

SethMichaelLarson:
Sure, I've probably got to add some more `#Platform-specific` comments within `_default_selector` as well to get 100% coverage.

SethMichaelLarson:
So codecov is showing the incorrect PR for some reason. Should be showing 100%: https://codecov.io/gh/shazow/urllib3/pull/1128
CI seems to have failed with a transient failure.

urllib3/util/selectors.py
```diff
@@ -535,19 +535,30 @@ def close(self):
...
+        selector_type = KqueueSelector
+    elif hasattr(select, 'epoll'):  # Platform-specific: Linux
+        selector_type = EpollSelector
+    elif hasattr(select, 'poll'):  # Platform-specific: Linux
+        selector_type = PollSelector
+    elif hasattr(select, 'select'):  # Platform-specific: Windows
+        selector_type = SelectSelector
+    else:  # Platform-specific: AppEngine
+        raise ValueError('Platform does not have a selector')
+    DefaultSelector = selector_type
```
haikuginger: Do we only want to use this function the first time we initialize a selector, or every time? What if a selector gets used before the monkey patch takes effect? What sort of performance costs would we incur by simply leaving DefaultSelector set to execute this function, rather than attaching to a concrete selector class?

haikuginger:
I have one question, but the overall setup looks good to me.

urllib3/util/selectors.py
SethMichaelLarson: The problem we're trying to mitigate is basically this one:
```python
import urllib3
from eventlet import monkey
monkey.patch_all()
```

We discussed having it run every time, I could run a test to see the performance hit that running each time versus once. In v2 there's going to be a selector per `Connection` so that's a lot of calls.

SethMichaelLarson: Running on Python 3.5.2 I received the following results from timeit:

Here was my setup:
```bash
python -m timeit -s 'from urllib3.util import selectors' 'selectors.DefaultSelector().close()'
```

Checking only once:
```bash
100000 loops, best of 3: 5.93 usec per loop
```

Checking each call:
```bash
100000 loops, best of 3: 8.61 usec per loop
```

Not really a big difference. I'm willing to say we could call it each time.

haikuginger: > We discussed having it run every time

Ah, sorry to drag it back to a previous discussion, then. :) Unfortunately I haven't been able to spend as much time as I'd like on open source lately.

SethMichaelLarson: Oh no worries I'm just telling you for reference if you wanted to read back on our discussion. :) I'm the first person that should be pointed fingers at for taking breaks from my PRs haha

There's no numbers or tests in that discussion so it's nice to pin down the performance difference here.

haikuginger: Okay! Based on the above, I'd tend to lean towards dynamically picking the selector type every time - a difference of 3µs per selector is relatively minimal compared to the amount of time even a single network hop can take. But, of course, it's up to @Lukasa at the end of the day, and I'm happy either way.

urllib3/util/selectors.py
```diff
@@ -535,19 +535,44 @@ def close(self):
...
+    """ Checks that select structs can be allocated
+    by the underlying operating system, not just
+    advertised by the select module. We don't check
+    select() because we'll be hopeful that most 
+    platforms that don't have it available will not
+    advertise it. (ie: GAE) """
+    try:
+        getattr(select, struct)().close()
+        return True
+    except (OSError, AttributeError):
```
Lukasa: Given we catch `AttributeError` here, do we need the `hasattr` checks below?

urllib3/util/selectors.py
```diff
@@ -535,19 +535,44 @@ def close(self):
...
+    HAS_SELECT = False
+
+
+def _can_allocate(struct):
+    """ Checks that select structs can be allocated
+    by the underlying operating system, not just
+    advertised by the select module. We don't check
+    select() because we'll be hopeful that most 
+    platforms that don't have it available will not
+    advertise it. (ie: GAE) """
```
Lukasa: This paragraph doesn't have to be so tightly bound. Use all 79 characters if you want!

SethMichaelLarson: Haha you can tell when I'm working on the GitHub web platform versus a real IDE. I'm sitting in my bed after just woken up making these changes and I have to run to work soon. :)

Lukasa: Hah, nice. There's no rush. :smile:

urllib3/util/selectors.py
SethMichaelLarson: This is very true. I'll remove the `hasattr` checks.

SethMichaelLarson:
Latest change adds detecting if the `select` module is incorrectly advertising a service not supported by the underlying OS as per @programatix's problem in kennethreitz/requests#3906, #1131, and #1132.
I'll add some comments in the requests issue as well to point to this issue.

So @haikuginger this next change will definitely not allow for checking each time. :/

urllib3/util/selectors.py
SethMichaelLarson: Yeah, I've discovered that around this time is one of the better times to get comments from maintainers on `urllib3` :)

urllib3/util/selectors.py
```diff
@@ -535,19 +535,44 @@ def close(self):
...
+    try:
+        getattr(select, struct)().close()
```
SethMichaelLarson: So this definitely won't work, will have to detect if the selector has a `close()` function and use it then. (`select.poll` has no `close()`).
Also might have to add a scenario for `select.poll.poll()` as it won't throw the error until it's actually used.

urllib3/util/selectors.py
Lothiraldan: It seems that currently this function is called every time. I think the reference `urllib3.util.selectors.DefaultSelector` is correctly replaced but not the `DefaultSelector` reference in the wait.py file.
I don't know if it's the expected behavior.

SethMichaelLarson: It's only called once currently. Within `_default_selector()` right at the very end `DefaultSelector` is changed to the result of the first call of `_default_selector()`.

SethMichaelLarson:
@Lukasa So we're missing coverage inside of `_can_allocate` because how how selector tests choose the selector to use. Should I rewrite `make_selector()` in the test cases to use the `_default_selector()` route rather than marking the function to skip coverage?

Lukasa:
Go for it. 👍

urllib3/util/selectors.py
Lothiraldan: I don't think so, with this script and having add `print('calling _default_selector')` in the `_default_selector` function:

```
import urllib3.util.selectors

connection = PoolManager()
print('DefaultSelector', urllib3.util.selectors.DefaultSelector)

import gevent.monkey
gevent.monkey.patch_all()

connection.urlopen('GET', 'http://google.fr')
print('DefaultSelector', urllib3.util.selectors.DefaultSelector)
connection.urlopen('GET', 'http://google.fr')
print('DefaultSelector', urllib3.util.selectors.DefaultSelector)
connection.urlopen('GET', 'http://google.fr')
print('DefaultSelector', urllib3.util.selectors.DefaultSelector)
```

I get this output:

```
('DefaultSelector', <function _default_selector at 0x101700ed8>)
('DefaultSelector', <function _default_selector at 0x101700ed8>)
calling _default_selector
calling _default_selector
('DefaultSelector', <class 'urllib3.util.selectors.SelectSelector'>)
calling _default_selector
calling _default_selector
('DefaultSelector', <class 'urllib3.util.selectors.SelectSelector'>)
```

As wait.py https://github.com/shazow/urllib3/blob/master/urllib3/util/wait.py#L3 use directly the DefaultSelector reference, it isn't replaced when setting the global `DefaultSelector`. Checkout this slides https://gitpitch.com/Lothiraldan/python-production-monkey-patching/master?grs=github&t=sky for more explanation (6 slides to the right sorry couldn't get the right sharing url).

SethMichaelLarson: Ah good catch! I'll deal with this as well in my next patch.

SethMichaelLarson:
So now the `test_selectors` uses the `DefaultSelector()` pathway that would be used during normal usage of the module. I also added test cases for the unique situations that we are making this patch for.

test/test_selectors.py
```diff
@@ -681,24 +689,93 @@ def test_above_fd_setsize(self):
...
+        # Remove all selectors except `select.select`.
+        self.setup_select_module('select')
+
+        # Now we're going to patch in a bad `epoll`.
+        def bad_epoll(*args, **kwargs):
+            raise OSError(errno.ENOSYS)
+
+        select.epoll = bad_epoll
+        if not had_epoll:
+            self.addCleanup(delattr, select, 'epoll')
```
Lukasa: So this patching and unpatching is a really common pattern in these tests: it might be good to factor it out to a helper method like `patch_selector_implementation('epoll', bad_epoll)`

test/test_selectors.py
```diff
@@ -681,24 +689,93 @@ def test_above_fd_setsize(self):
...
 
 
 @skipUnless(hasattr(selectors, "EpollSelector"), "Platform doesn't have an EpollSelector")
 class EpollSelectorTestCase(BaseSelectorTestCase, ScalableSelectorMixin):
-    SELECTOR = getattr(selectors, "EpollSelector", None)
+    def setUp(self):
+        self.setup_select_module('epoll')
 
 
 @skipUnless(hasattr(selectors, "KqueueSelector"), "Platform doesn't have a KqueueSelector")
```
Lukasa: Does this test work correctly now?

test/test_selectors.py
```diff
@@ -681,24 +689,93 @@ def test_above_fd_setsize(self):
...
+                raise OSError(errno.ENOSYS)
+
+        select.poll = BadPoll
+        if not had_poll:
+            self.addCleanup(delattr, select, 'poll')
+
+        selector = self.make_selector()
+        self.assertIsInstance(selector, selectors.SelectSelector)
+
+    @skipUnless(hasattr(errno, 'ENOSYS'), 'Platform does not have errno.ENOSYS.')
```
sigmavirus24: You can create new decorators by doing

```
skipUnlessHasEnoSys = skipUnless(hasattr(errno, 'ENOSYS'), 'Platform does not have errno.ENOSYS.')


     @skipUnlessHasEnoSys
     def test_...
```

That way you're not copying it around several places and others who add tests won't be either.

SethMichaelLarson: Good idea, I addressed this in my latest.

test/test_selectors.py
SethMichaelLarson: Sorry, I'm not sure what this comment is about?

test/test_selectors.py
SethMichaelLarson: Yes it is! I changed the function `setup_select_module` to be slightly different and to do two things at once. Please tell me if I'm being too smart for our collective good with the function. I feel like it's within the realm of okay but close to the edge of being too much.

test/test_selectors.py
```diff
@@ -135,10 +135,18 @@ def make_socketpair(self):
         return rd, wr
 
     def make_selector(self):
-        s = self.SELECTOR()
+        s = selectors.DefaultSelector()
         self.addCleanup(s.close)
         return s
 
+    def setup_select_module(self, selector):
```
SethMichaelLarson: True, I renamed it to `patch_select_module` and added a docstring.

test/test_selectors.py
Lukasa: My question is, does having a KqueueSelector attribute correctly define whether we should run these tests, or do we need to change this logic?

SethMichaelLarson: So we don't need to change it but we could? `selectors.py` checks to see if `select` has `kqueue` before defining `KqueueSelector`.

test/test_selectors.py
```diff
@@ -681,24 +705,81 @@ def test_above_fd_setsize(self):
...
+        # Make sure that the selector returned only uses the selector available.
+        selector = self.make_selector()
+        self.assertIsInstance(selector, selectors.SelectSelector)
+
+    @skipUnlessHasENOSYS
+    def test_select_module_defines_does_not_implement_poll(self):
+        # This test is to make sure that if a platform defines
+        # a selector as being available but does not actually
+        # implement it (kennethreitz/requests#3906) does not
+        # make DefaultSelector() fail.
```
Lukasa: This comment has got some confusing grammar in it. It should probably end with `does not actually implement it (kennethreitz/requests#3906) then DefaultSelector() does not fail`.

SethMichaelLarson: Okay I'll fix this comment. :)

SethMichaelLarson:
Thank you @Lukasa, if you see anything just point it out. I'll do another once-over as well. :)

SethMichaelLarson:
Can't merge this myself @Lukasa, don't have write privileges for this repo :)

haikuginger:
Thanks, @SethMichaelLarson!

✨🍮✨

urllib3/util/selectors.py
```diff
@@ -535,19 +536,46 @@ def close(self):
...
-elif 'PollSelector' in globals():  # Platform-specific: Linux
-    DefaultSelector = PollSelector
-elif 'SelectSelector' in globals():  # Platform-specific: Windows
-    DefaultSelector = SelectSelector
-else:  # Platform-specific: AppEngine
-    def no_selector(_):
-        raise ValueError("Platform does not have a selector")
-    DefaultSelector = no_selector
-    HAS_SELECT = False
+def DefaultSelector(*_):
```
sigmavirus24: Why are we using `*_` to collect arguments here?

SethMichaelLarson: Without it you receive `TypeError: DefaultSelector() takes 0 positional arguments but 1 was given` during the `wait_for_*` functions but normal calls to it you receive no positional arguments. I can only assume it's passing in an argument because it's treating it like a class?

Lukasa: To be fair, we wrote the `wait_for_*` functions, so we can potentially just fix those.

SethMichaelLarson: So for some reason the `TestCase` is being passed in as an argument (`self`) for the `wait_for_io` tests. :/ I can now see why this is happening. I'll open a separate PR to fix this issue.

SethMichaelLarson: To be clear this was occurring in the past set of tests for adding selectors as well, each selector was receiving a `unittest.TestCase` as `self` but it was silently working fine.

SethMichaelLarson: See #1134.

PR #1130, opened 2017-03-08
Change HTTPHeaderDict vals intertal representation to list

yetingsky:
Just using list will make the code clear and not break the API

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -144,7 +144,7 @@ class HTTPHeaderDict(MutableMapping):
             self.extend(kwargs)
 
     def __setitem__(self, key, val):
-        self._container[key.lower()] = (key, val)
+        self._container[key.lower()] = [key, val]
         return self._container[key.lower()]
 
     def __getitem__(self, key):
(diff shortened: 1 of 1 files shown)


urllib3/_collections.py
```diff
@@ -278,10 +268,7 @@ def __repr__(self):
     def _copy_from(self, other):
         for key in other:
             val = other.getlist(key)
-            if isinstance(val, list):
-                # Don't need to convert tuples
-                val = list(val)
-            self._container[key.lower()] = [key] + val
+            self._container[key.lower()] = [key] + list(val)
```
sigmavirus24: So I would rather be defensive here. The other object *might not* be a HTTPHeaderDict and could potentially not return a list. If it returns, for example, a string then we'll be doing:

```
>>> list('str')
['s', 't', 'r']
```

yetingsky: I think here `list(val)` will just be fine. [getlist](https://github.com/yetingsky/urllib3/blob/master/urllib3/_collections.py#L257) will always return list, according to docstring and implementation.
Actually, `isinstance(val, list)` may never reach

yetingsky:
when a head only has one value, using tuple. but when a head has many value(like cookie header), using list because tuple can not append... This does not make sense to me... Maybe be consistent will be better.

urllib3/_collections.py
sigmavirus24: @yetingsky a `HTTPHeaderDict` is not the only thing we might be copying from though. I don't trust other implementations to be 100% correct.

yetingsky:
@sigmavirus24 :+1: you are right, l have leaved that part unchanged

PR #1131, opened 2017-03-08
DefaultSelector fallback to less efficient methods

programatix:
Test the selector before making it the default selector. This is due to issue reported by myself at https://github.com/kennethreitz/requests/issues/3906

### DIFF
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -539,14 +539,26 @@ if hasattr(select, "kqueue"):
 # kqueue == epoll > poll > select. Devpoll not supported. (See above)
 # select() also can't accept a FD > FD_SETSIZE (usually around 1024)
 if 'KqueueSelector' in globals():  # Platform-specific: Mac OS and BSD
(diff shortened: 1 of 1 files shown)


Lukasa:
Hello again! :sparkles:

So I think this is generally useful, but we should write tests for it if we can. We also need to consider whether this should be merged or coordinated with @SethMichaelLarson's patch in #1128.

programatix:
I'll attempt to write tests for it, but suggestion would be appreciated. The change made will be executed during `import`, thus if there is any error in it, it would cause exception during `import`. Any reference on how to test such condition?

As for manual test, I can confirm that at least the `EpollSelector` falling-back to 'PollSelector' part works on my router.

Lukasa:
@programatix Yeah, that's the trickier bit, and it's why I want to wait to see what @SethMichaelLarson has to say here. At the very least we need to factor this out into a function, but his patch has already done that so we may be able to make this a bit easier for you by merging these in an order that lets you piggy-back on his work.

PR #1132, opened 2017-03-08
Additional check before declaring a Selector

programatix:
Ignore Selector(s) which are unsupported by underlying OS.

I think this is better than https://github.com/shazow/urllib3/pull/1131. However for some weird reason, even though I just changed the indexing of the code, the github compare makes it looks like I've changed the whole file.

### DIFF
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -336,7 +336,9 @@ if hasattr(select, "poll"):
     try:
         # Test if poll is supported by underlying OS
         select.poll()
-
+    except OSError:
+        pass
+    else:
         class PollSelector(BaseSelector):
             """ Poll-based selector """
             def __init__(self):
@@ -388,15 +390,14 @@ if hasattr(select, "poll"):
 
                 return ready
 
-    except OSError:
-        pass
-
 
 if hasattr(select, "epoll"):
     try:
         # Test if epoll is supported underlying OS
         select.epoll()
-
+    except OSError:
+        pass
+    else:
         class EpollSelector(BaseSelector):
             """ Epoll-based selector """
             def __init__(self):
@@ -462,15 +463,14 @@ if hasattr(select, "epoll"):
                 self._epoll.close()
                 super(EpollSelector, self).close()
 
-    except OSError:
-        pass
-
 
 if hasattr(select, "kqueue"):
     try:
         # Test if kqueue is supported underlying OS
         select.kqueue()
-
+    except OSError:
+        pass
+    else:
         class KqueueSelector(BaseSelector):
             """ Kqueue / Kevent-based selector """
             def __init__(self):
@@ -551,9 +551,6 @@ if hasattr(select, "kqueue"):
             def close(self):
                 self._kqueue.close()
                 super(KqueueSelector, self).close()
-
-    except OSError:
-        pass
 
 
 # Choose the best implementation, roughly:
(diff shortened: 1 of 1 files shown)


urllib3/util/selectors.py
```diff
@@ -333,206 +333,227 @@ def select(self, timeout=None):
...
+                        ready.append((key, events & key.events))
 
-                key = self._key_from_fd(fd)
-                if key:
-                    ready.append((key, events & key.events))
+                return ready
 
-            return ready
+    except OSError:
+        pass
```
Lukasa: These except blocks are way too big. It'll be better to use `try...except...else` and place the class declaration in the `else` statement.

Lukasa:
Ok, so this could work, but the diff is *massive* for what is actually a very small change. I wonder if we can handle this by defining all the classes up front and then setting boolean variables if the underlying selector actually works. That way we'll only end up with a small diff.

Is there any reason to entirely prevent the definition of the class if the selector can't be called?

urllib3/util/selectors.py
programatix: Applied `try...except...else` as suggested.

programatix:
From the code, I understand that it uses the `hasattr(select, "xxx")` to determine whether the selector exists, which resulted in the class being defined, else it would be skipped. It then decides which selector to be assigned to `DefaultSelector` by checking whether the specific selector exists in globals().

So, I thought that since if the `hasattr(select, "xxx")` returns false and the class would not be defined, why not prevent the class being defined if the additional check failed?

On "defining all the classes up front", it will work. It can be done by removing the `if hasattr(select, "xxx")` line and replace the `if 'xxxSelector' in globals():` with it. The additional check can be done there. But this too will cause the massive diff since the indexing of the classes need to be changed.

Lukasa:
Yeah, so as far as I can tell the reason the `hasattr` check is present is not so much for urllib3's sake as it is for the original Python selectors module. We could easily move to using boolean flags without causing ourselves any trouble.

However, ultimately I think we should hold off and let @SethMichaelLarson weigh in. This is really mostly his work, and he'll have a better idea of what looks reasonable.

programatix:
I'm fine with anything as long as the next `requests` release works on my router without the need to update the script, else thanks to you, at least I now know how to workaround it.

I've reported the problem with the firmware dev but it seems to be ignored.

Lukasa:
Well that's disappointing. 😞 

Regardless, we definitely consider this a bug and will aim to come up with a good fix for it.

SethMichaelLarson:
Hello @programatix! Could you do a quick test and try using the Python 3 `selectors` module on your same system? If it doesn't work you might also open up an issue on CPython. (Or I can open the issue for you and mention your problem.) because the `select` module shouldn't advertise availability of any function which isn't supported by the underlying OS.

Back to urllib3, I'm going to make a revision to #1128 to make sure that each structure can actually be allocated.

programatix:
Hi @SethMichaelLarson,

I performed the test as requested and the following is the result,
```
Python 3.6.0 (default, Feb 17 2017, 13:43:51)
[GCC 5.4.0] on linux
>>> import selectors
>>> selectors.DefaultSelector()
  File "/opt/lib/python3.6/selectors.py", line 399, in __init__
    self._epoll = select.epoll()
OSError: [Errno 89] Function not implemented
```
It would be great if you could help in opening an issue at CPython as you're more well versed on this issue than me.

Lukasa:
So I think we're closing this in favour of #1128.

PR #1136, opened 2017-03-23
Remove support for sending unicode request bodies.

Lukasa:
This has been bothering me for a while: the auto-encoding of unicode bodies is a really good source of subtle bugs, and I'd argue it violates the Zen of Python. This change removes that support from the v2 branch.

### DIFF
--- a/test/with_dummyserver/test_chunked_transfer.py
+++ b/test/with_dummyserver/test_chunked_transfer.py
@@ -65,10 +65,8 @@ class TestChunkedTransfer(SocketDummyServerTestCase):
         self._test_body(b'thisshouldbeonechunk\r\nasdf')
 
     def test_unicode_body(self):
-        # Define u'thisshouldbeonechunk\r\näöüß' in a way, so that python3.1
-        # does not suffer a syntax error
         # Unicode bodies are not supported.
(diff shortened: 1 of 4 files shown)


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -66,11 +67,12 @@ def test_bytestring_body(self):
     def test_unicode_body(self):
         # Define u'thisshouldbeonechunk\r\näöüß' in a way, so that python3.1
```
nateprewitt: More of a though/optional comment: Since we no longer support 3.1, and are using unicode strings in other tests, is there a reason not to define this as `u'thisshouldbeonechunk\r\näöüß'`. It's a bigger surface area change on the test but seems more explicit in what we're trying to accomplish here.

Lukasa:
I *think* Requests would choose to handle that encoding automatically. Doing it at the "more helpful" level in Requests is probably better. Requests can also, in many cases, set appropriate HTTP headers automatically.

test/with_dummyserver/test_chunked_transfer.py
Lukasa: Uh, no, good spot.

Lukasa: Fixed. :sparkles:

Lukasa:
Alright let's do it. Thanks for the reviews folks!

sigmavirus24:
🍰

PR #1139, opened 2017-03-31
Reduce selector overuse in v2 branch.

Lukasa:
Right now the code will create new selectors to check for EOF, even though each connection now has one. We should probably stop doing that. :D


SethMichaelLarson:
:tada:

PR #1140, opened 2017-03-31
Remove chunked kwarg.

Lukasa:
It doesn't do anything anymore, so let's be done with it.


sigmavirus24:
Does this not cause a merge conflict with #1141?

Lukasa:
Yup, it does, but I'll fix this up when we merge that.

PR #1006, opened 2016-10-24
Add Windows CI support with AppVeyor

SethMichaelLarson:
Recently @Lukasa has added Mac OS support using Travis in order to fully test the new selectors addition.  One of the main projects that requires decent selectors support is the Happy Eyeballs algorithm (RFC 6555) but the algorithm also requires the use of Windows error constants in order to correctly work on all platforms.

I decided to go ahead and use a good template for testing Python with AppVeyor found at: https://github.com/ogrisel/python-appveyor-demo  I've changed it from a primarily Cython-focused project template to a tox-focused project and I've gotten AppVeyor to run all our tests on all our supported Python versions.  This will give us a lot less guessing when implementing or using features that are for platforms that are not available to all maintainers.  I have not tested the codecov section of the code, but I'm sure it's not too far off if at all.  I've already got all tests running. I'm sure this will take a repository owner to fully integrate AppVeyor support.

From my own trial runs, there are a few tests that fail on Windows already, especially in Python 2.6 with many TLS/SSL errors, there is also another test failing on all Python versions.

PR #1142, opened 2017-03-31
Address issues with testing on Windows

SethMichaelLarson:
This PR comes after #1006 has been merged into master. We knew there would probably be some issues, and lo-and-behold there are. Only issues seem to be with testing using `socket.socketpair` and incompatibilities with our `socketpair` helper for Python 2.

### DIFF
--- a/test/socketpair_helper.py
+++ b/test/socketpair_helper.py
@@ -3,7 +3,10 @@ import socket
 try:
     _CONNECT_ERROR = (BlockingIOError, InterruptedError)
 except:
-    _CONNECT_ERROR = (Exception,)
+    try:
+        _CONNECT_ERROR = (WinError, OSError, socket.error)
+    except:
+        _CONNECT_ERROR = (OSError, socket.error)
     
 if hasattr(socket, 'socketpair'):
     # Since Python 3.5, socket.socketpair() is now also available on Windows
--- a/test/socketpair_helper.py
+++ b/test/socketpair_helper.py
@@ -1,10 +1,16 @@
 import socket
 
+# Figuring out what errors could come out of a socket. There are three
+# different situations. Python 3 post-PEP3151 will define and use
+# BlockingIOError and InterruptedError from sockets. For Python pre-PEP3151
+# both OSError and socket.error can be raised except on Windows where
+# WindowsError can also be raised. We want to catch all of these possible
+# exceptions so we catch WindowsError if it's defined.
 try:
     _CONNECT_ERROR = (BlockingIOError, InterruptedError)
 except NameError:
     try:
-        _CONNECT_ERROR = (WinError, OSError, socket.error)  # noqa: F821
+        _CONNECT_ERROR = (WindowsError, OSError, socket.error)  # noqa: F821
     except NameError:
         _CONNECT_ERROR = (OSError, socket.error)
 
--- a/test/socketpair_helper.py
+++ b/test/socketpair_helper.py
@@ -4,10 +4,10 @@ try:
     _CONNECT_ERROR = (BlockingIOError, InterruptedError)
 except:
     try:
-        _CONNECT_ERROR = (WinError, OSError, socket.error)
+        _CONNECT_ERROR = (WinError, OSError, socket.error)  # noqa: F821
     except:
         _CONNECT_ERROR = (OSError, socket.error)
-    
+
 if hasattr(socket, 'socketpair'):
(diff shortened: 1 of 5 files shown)


SethMichaelLarson:
I'll note here that `continuous-integration/appveyor/py` is currently not a required commit status for branch protection. Probably should be added.

While we are at it should also add `codecov/project` as being required now that the commit status has been used since activating Codecov.

I don't know if you can do this @Lukasa but I think this is something @shazow will also have to do.

haikuginger:
@SethMichaelLarson, I noticed while reading this that we lifted some code from https://gist.github.com/geertj/4325783, and noted that it's public domain? I don't anticipate any issues, but do we have a written source on the public domain claim? I don't see a license on that gist.

There's also a [backport package](https://github.com/mhils/backports.socketpair) that says it's lifted from Python 3.5, but again, without an explicit license given.

If we can't verify the public domain status, then I think we probably want to do our own backport from the Python 3.5 source code and include the PSF license header in that module.

SethMichaelLarson:
I don't remember noting it as public domain, the only stuff that's public domain are the PowerShell scripts. The socketpair helper code is from the Python 3.5+ Windows socketpair implementation (taken from that gist originally) but asyncio also uses it.

test/socketpair_helper.py
```diff
@@ -1,5 +1,10 @@
 import socket
 
+try:
+    _CONNECT_ERROR = (BlockingIOError, InterruptedError)
+except:
```
Lukasa: Better to be explicit about what we expect here.

SethMichaelLarson:
I found where it is listed as public domain (apparently). I probably took it directly from asyncio as I was working on that project a lot at the time I made this change. https://github.com/python/asyncio/blob/master/asyncio/windows_utils.py#L37

test/socketpair_helper.py
SethMichaelLarson: Sure, will be an additional `try` and `catch` as non-Windows platforms don't define `WinError`.

Lukasa: That's fine, I don't mind that. Better safe than sorry. :grin:

shazow:
> I'll note here that `continuous-integration/appveyor/py` is currently not a required commit status for branch protection. Probably should be added.
> 
> While we are at it should also add `codecov/project` as being required now that the commit status has been used since activating Codecov.
> 
> I don't know if you can do this @Lukasa but I think this is something @shazow will also have to do.

@SethMichaelLarson Done!

SethMichaelLarson:
So now a single TLSv1 test in Python 2.6 is failing because of errors? I haven't seen this error before but I'll look at it. Any ideas?

Lukasa:
Probably this is an issue with Python 2.6 on Windows having a bad OpenSSL library.

SethMichaelLarson:
Should we skip on Python 2.6 and Windows?

Lukasa:
Might be the way to go, yeah.

SethMichaelLarson:
Looks like a selector test failed on AppVeyor. Might have to go the same direction we did with Travis and increase the tolerance for timing on AppVeyor so we don't have random failures.

Lukasa:
Sounds reasonable.

test/socketpair_helper.py
Lukasa: These are still naked except statements.

SethMichaelLarson: Oh sorry! I thought you meant that you didn't want `_CONNECT_ERROR = (Exception,)`. I'll fix this.

test/socketpair_helper.py
```diff
@@ -1,5 +1,13 @@
 import socket
 
+try:
+    _CONNECT_ERROR = (BlockingIOError, InterruptedError)
+except NameError:
+    try:
+        _CONNECT_ERROR = (WinError, OSError, socket.error)  # noqa: F821
+    except NameError:
+        _CONNECT_ERROR = (OSError, socket.error)
```
Lukasa: Ok cool, can we get a comment above this block explaining why we're doing this dance? :grin:

SethMichaelLarson: Sure! I'll add that.

SethMichaelLarson:
Forgot to whitelist `APPVEYOR` environment variable in tox. I've added that to the tox.ini

Lukasa:
Look at *all those green tests*. Very nicely done!

SethMichaelLarson:
There are quite a lot now aren't there! :) Huzzah to better coverage!

PR #1143, opened 2017-04-03
Factor our some duplicate code in v2 branch.

Lukasa:


Lukasa:
So this PR is failing right now because AppVeyor expects to run. I'll bring in the master branch to v2 once we've resolved #1142.

PR #1145, opened 2017-04-05
Rewrite Response I/O in terms of stream.

Lukasa:
Now that I/O is done in terms of iterating a base connection, rather than in terms of calling read() on a file-like object, stream() becomes the more natural method to use to do I/O than read() does. This has meant both that there is a bunch of awkward code duplication in read() *and* that stream() became very inefficient, as it was doing buffered I/O.

This change adjusts stream so that it no longer does buffered I/O, and rewrites read() to be in terms of stream. This should help clean things up for writing an async response object in future. Along the way I've made some miscellaneous smaller changes.

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -233,6 +233,35 @@ class TestResponse(unittest.TestCase):
         resp.close()
         self.assertRaises(StopIteration, next, resp.stream())
 
+    def test_close_midstream(self):
+        # A mock fp object that wraps a list and allows closing.
+        class MockFP(object):
+            self.list = None
+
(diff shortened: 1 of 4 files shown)


test/test_response.py
```diff
@@ -202,29 +202,76 @@ def test_io_bufferedreader(self):
 
         # This is necessary to make sure the "no bytes left" part of `readinto`
         # gets tested.
+        print("First read")
```
nateprewitt: These two print statements probably aren't needed.

shazow:
:clap:

PR #1146, opened 2017-04-10
Move redirect handling entirely to PoolManager

Lukasa:
Redirect handling code was duplicated in urllib3, being present in two places: in the ConnectionPool, and the PoolManager. This centralises all redirect handling code into the PoolManager, where it needs to be to handle cross-host redirects. ConnectionPool objects are no longer capable of doing redirects.

After this I'll provide a PR that moves all retry handling up to PoolManager as well, at which point ConnectionPool will have been greatly simplified.


haikuginger:
Will review in the next couple days.

sigmavirus24:
Am I understanding correctly that this loses the redirect history since we put the onus of managing redirects on the Retry object?

Lukasa:
I don't think so, the tests for RedirectHistory continue to pass.

Lukasa:
Ok, tests are green, this is good to review.

urllib3/connectionpool.py
```diff
@@ -533,11 +521,7 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
             headers = self.headers
 
         if not isinstance(retries, Retry):
-            retries = Retry.from_int(retries, redirect=redirect, default=self.retries)
-
-        # Check host
-        if assert_same_host and not self.is_same_host(url):
-            raise HostChangedError(self, url, retries)
+            retries = Retry.from_int(retries, default=self.retries)
```
haikuginger: Perhaps an explicit `redirect=False`?

urllib3/poolmanager.py
```diff
@@ -301,8 +302,12 @@ def urlopen(self, method, url, redirect=True, **kw):
         u = parse_url(url)
         conn = self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
 
-        kw['assert_same_host'] = False
-        kw['redirect'] = False
+        # Rewind body position, if needed. Record current position
```
haikuginger: ^update docstring, please.

Lukasa: What docstring update is needed? I feel like I don't understand. 😕

haikuginger: The PoolManager.request method docstring says something about "custom cross-host redirect logic"; this is now the only redirect logic.

Lukasa: Aha, I see. =) Yeah, I can fix that.

Lukasa:
Nits addressed.

PR #1150, opened 2017-04-24
Test Python 3.6 on Appveyor

Lukasa:
We should probably increase our test matrix to test as many versions on as many platforms as possible.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,6 +30,11 @@ environment:
       PYTHON_ARCH: "64"
       TOXENV: "py35"
 
+    - PYTHON: "C:\\Python36-x64"
+      PYTHON_VERSION: "3.6.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py36"
+
 install:
   # Install Python (from the official .msi of http://python.org) and pip when
   # not already installed.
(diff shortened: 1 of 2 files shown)


Lukasa:
Well...that's a surprising test failure. I have just validated that `datetime.datetime.fromtimestamp(0)` does in fact return `OSError` on Windows under Python 3.6. I don't yet know why.

Lukasa:
Ah, it's [Python issue 29097](https://bugs.python.org/issue29097).

SethMichaelLarson:
You could add `3.7-dev` to Travis as well. :)

Lukasa:
I'm a bit less interested in doing that in the CI, mostly because it's often not useful to block on bugs introduced upstream.

sigmavirus24:
> I'm a bit less interested in doing that in the CI, mostly because it's often not useful to block on bugs introduced upstream.

You could make it an allowed failure though. That would mean people would have to check it to report issues upstream, but *shrug*

Lukasa:
Yeah, so I've gone pretty cold on allowed-failures. *Generally speaking* it's been my experience that people don't check them, so they end up red and stay red.

PR #1157, opened 2017-04-25
Add assert_hostname to list of valid keys for PoolManager

dims:
In Kubernetes' client-python, we ran into this problem with
1.21:
https://travis-ci.org/kubernetes-incubator/client-python/jobs/225699267

We add the assert_hostname to our PoolManager here:
https://github.com/kubernetes-incubator/client-python/blob/192b67c4466c4e1c45f05eb4a1c6a6899e1c279d/kubernetes/client/rest.py#L107

I can't see any other workaround at the moment, please let us know is there's a better way to do this.

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -277,6 +277,16 @@ class TestPoolManager(unittest.TestCase):
         self.assertTrue(pool is other_pool)
         self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
+    def test_assert_hostname_and_fingerprint_flag(self):
+        """Assert that pool manager can accept hostname and fingerprint flags."""
+        fingerprint = '92:81:FE:85:F7:0C:26:60:EC:D6:B3:BF:93:CF:F9:71:CC:07:7D:0A'
+        p = PoolManager(assert_hostname=True, assert_fingerprint=fingerprint)
+        self.addCleanup(p.clear)
+        pool = p.connection_from_url('https://example.com/')
+        self.assertEqual(1, len(p.pools))
+        self.assertTrue(pool.assert_hostname)
+        self.assertEqual(fingerprint, pool.assert_fingerprint)
+
     def test_http_connection_from_context_case_insensitive(self):
         """Assert scheme case is ignored when getting the https key class."""
         p = PoolManager()
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ dev (master)
 * Fixed regression in 1.21 that threw exceptions when users passed the
   ``socket_options`` flag to the ``PoolManager``.  (Issue #1165)
 
+* Fixed regression in 1.21 that threw exceptions when users passed the
+  ``assert_hostname`` or ```assert_fingerprint`flag to the ``PoolManager``.
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -45,6 +45,8 @@ _key_fields = (
     'key__proxy_headers',  # dict
     'key_socket_options',  # list of (level (int), optname (int), value (int or str)) tuples
     'key__socks_options',  # dict
+    'key_assert_hostname',  # bool or string
+    'key_assert_fingerprint',  # str
 )
 
 #: The namedtuple class used to construct keys for the connection pool.


haikuginger:
@dims, what's your goal in trying to pass `assert_hostname` to a PoolManager? It makes sense in the context of a ConnectionPool, which has a single host it connects to, but a PoolManager, by design, is not tied to any specific pool.

dims:
@haikuginger Jesse, All we are doing here is make sure we can set the assert_hostname in the HTTPSConnectionPool (https://github.com/shazow/urllib3/blob/0cb7780a0613baa518be26df4b623ecb0ab3cbe5/urllib3/connectionpool.py#L740) and we are possibly doing it wrong. what's the right way to do it?

Here's how we are using PoolManager (https://github.com/kubernetes-incubator/client-python/search?utf8=%E2%9C%93&q=PoolManager&type=), please note that this is not new code.

Lukasa:
For this argument you should really use `PoolManager.connection_from_host` to pass the `assert_hostname` argument. All the rest are fine, but that one should be passed that way.

However, I acknowledge that this *used* to work, and a case could be made that #1016 regressed this. Do we think that we should go back and re-add this, or do we think it was just not really a sensible thing to do?

haikuginger:
@dims, if you do not have a very specific purpose in mind, then it is not a meaningful operation to set `assert_hostname` for the child `ConnectionPools` of a `PoolManager`.

`assert_hostname` means that a `ConnectionPool` will reject requests that are targeted to a different hostname than the `ConnectionPool` was created with.

In the context of `PoolManager`, if a request is made to a different host than a `ConnectionPool` was created with, then the request will be directed either to a different existing `ConnectionPool`, or a new `ConnectionPool` will be created to send the request, and the `assert_hostname` logic on the previous `ConnectionPool` _will never be executed_.

If you have a specific reason for creating a `ConnectionPool` with `assert_hostname` set from a `PoolManager`, you can use [`PoolManager.connection_from_host()`](https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L198) and pass in a `pool_kwargs` dictionary containing `assert_hostname`.

Lukasa:
@haikuginger It should be noted that `assert_hostname` can also be set to be a string, in which case it will override the literal host we connected to as the name for validating the remote cert. It is *also* senseless to pass this to a PoolManager (as it by definition makes it only valid for one host), but should be noted.

dims:
@haikuginger we were using `request` method in PoolManager instance itself, looks like we have to rewrite stuff to use connection_from_host

dims:
@Lukasa : for us it was a blanket off switch to turn off assert_hostname for everything...

haikuginger:
@dims, if you're using `PoolManager().request()`, what specific undesired behavior are you seeing which you want to counteract?

If, as I believe from taking a look at the history of your codebase, you want to disable certificate hostname verification for SSL connections despite the fact that your PoolManager is capable of performing that verification, then the correct thing to do is to pass `cert_reqs='CERT_NONE'`.

dims:
@haikuginger : y, i don't remember exactly why we added that, @mbohlool has a PR to get rid of it. will ping you all back if we need something more. https://github.com/kubernetes-incubator/client-python/pull/200

Thanks,
Dims

mbohlool:
@haikuginger I believe we used `assert_hostname` for ssl host verifications. If `ipaddress` package exists, our ssl host verification passes. but if `ipaddress` package does not exists and we don't pass `assert_hostname` to connection pool, it will fail with a something like this:

```
hostname 'x1.x2.x3.x4' doesn't match either of 'x1.x2.x3.x4', '10.0.0.1', 'kubernetes', 'kubernetes.default', 'kubernetes.default.svc', 'kubernetes.default.svc.cluster.local', 'kubernetes-master'
```

some of our users may have this running right now without having `ipaddress` package and it will break if we remove this option.

Lukasa:
@mbohlool Is there a reason you don't add a dependency on `ipaddress`? That will resolve your problem in 100% of cases without disabling cert validation.

dims:
@Lukasa @haikuginger Cory, Jesse, i was looking at the other option you mentioned `PoolManager.connection_from_host()`, guess that won't work as well unless i am doing something wrong, please see below:

```
Python 2.7.13 (default, Feb  7 2017, 13:20:44)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)] on darwin
>>> import urllib3
>>> urllib3.__version__
'1.21'
>>> manager = urllib3.PoolManager()
>>> manager.connection_from_host('google.com', scheme='https')
<urllib3.connectionpool.HTTPSConnectionPool object at 0x10fd2ef50>
>>> manager.connection_from_host('google.com', scheme='https', pool_kwargs={'assert_hostname': True})
  File "/Users/dims/opensource/client-python/.tox/py27/lib/python2.7/site-packages/urllib3/poolmanager.py", line 219, in connection_from_host
  File "/Users/dims/opensource/client-python/.tox/py27/lib/python2.7/site-packages/urllib3/poolmanager.py", line 230, in connection_from_context
  File "/Users/dims/opensource/client-python/.tox/py27/lib/python2.7/site-packages/urllib3/poolmanager.py", line 95, in _default_key_normalizer
    return key_class(**context)
TypeError: __new__() got an unexpected keyword argument 'key_assert_hostname'
>>>
```

Lukasa:
Yeah, so I think that represents a breakage.

As discussed above, I think realistically either you don't want a PoolManager but instead you want a ConnectionPool, or we should treat it as reasonable to pass `assert_hostname` and `assert_fingerprint` at `PoolManager` scope. I lean towards "it used to be possible so it should stay possible", in which case I'd also want us to widen this PR to include `assert_fingerprint` (which is similarly broken).

Any contributors disagree?

haikuginger:
That makes sense to me. I think the correct way to do this is probably to have a collection of field names that cannot be included in the creation of a pool key, and simply pop them out of `context` in [`_default_field_normalizer`](https://github.com/shazow/urllib3/blob/master/urllib3/poolmanager.py#L55).

jeremycline:
I'm also of the opinion that it should stay possible since it used to be possible.

dims:
@haikuginger @Lukasa : since i am a n00b here, please feel free to use any part of this PR to fix this problem up.

If it is just adding `'key_assert_fingerprint'`, i can easily do that. please let me know.

Lukasa:
That's my suspicion, though @haikuginger proposed an alternative that might be worth looking at. @haikuginger, given that these fields are properties of the ConnectionPool, what's the reasoning for not making them part of the pool key?

haikuginger:
@Lukasa, I suppose adding them wouldn't hurt. It might be more reliable as well, if the `pool_kwargs` attribute on the PoolManager changes in the interim.

Lukasa:
Ok, @dims want to add `key_assert_fingerprint`?

dims:
@Lukasa @haikuginger : Done! will watch the CI jobs.

test/test_poolmanager.py
```diff
@@ -276,6 +276,14 @@ def test_http_connection_from_host_case_insensitive(self):
         self.assertTrue(pool is other_pool)
         self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
+    def test_assert_hostname_and_fingerprint_flag(self):
+        """Assert scheme case is ignored when getting the https key class."""
+        fingerprint = '92:81:FE:85:F7:0C:26:60:EC:D6:B3:BF:93:CF:F9:71:CC:07:7D:0A'
+        p = PoolManager(assert_hostname=True, assert_fingerprint=fingerprint)
+        self.addCleanup(p.clear)
+        p.connection_from_url('https://example.com/')
+        self.assertEqual(1, len(p.pools))
```
Lukasa: Can we extend this test to validate that the values are passed properly to the connection pool, and that they can be overridden?

dims: Ack, updated with additional asserts for the params we pass in.

test/test_poolmanager.py
```diff
@@ -276,6 +276,16 @@ def test_http_connection_from_host_case_insensitive(self):
         self.assertTrue(pool is other_pool)
         self.assertTrue(all(isinstance(key, PoolKey) for key in p.pools.keys()))
 
+    def test_assert_hostname_and_fingerprint_flag(self):
+        """Assert scheme case is ignored when getting the https key class."""
```
Lukasa: This docstring isn't quite right, do you mind rewording it to explain what is going on in this test?

dims: oops. done @Lukasa

SethMichaelLarson:
Changelog entry? Or is this minor enough.

Lukasa:
Oh yes, well noted. Can you bring yourself up-to-date with the current master and add a changelog entry to CHANGES?

Lukasa:
Don't forget to add the changelog entry! =)

polyzen:
Any chance this fix could be merged and released soon? It's a showstopper for Vdirsyncer. Thank you. :)

Lukasa:
@polyzen Not until the code review feedback is addressed, no. 😉

dims:
@Lukasa @haikuginger Cory, Jesse, apologies for the delay, i've updated the CHANGES.rst.

dims:
Thanks @SethMichaelLarson @Lukasa @haikuginger

mbohlool:
Good job @dims

haikuginger:
Thanks very much for the contribution, @dims! ✨🍰✨

Issue #1158, opened 2017-04-26
Retry when status is in status_forcelist and don't use the same connection


masteralves:
Hey folks,

I am trying to use urllib connectionpool.request so that when we get a status code that is in the status_forcelist, the same connection isn't reused for the retries. I tried using `release_conn=False` and then calling `.release_conn()`, but by inspecting the code, it looks like this is only releasing the last used connection. 
Assuming that we fail the 1st try and then we succeed the 2nd, the 1st connection is never released. Then I imagine we would have a problem of leaking connections.
Is this a bug, or is there something I'm missing here?

Cheers

Lukasa:
So, I wrote a test script:

```python
import urllib3

http = urllib3.HTTPConnectionPool(host='httpbin.org')
retry = urllib3.Retry(total=1, status=1, status_forcelist=[404], raise_on_status=False)

resp = http.request('GET', '/status/404', retries=retry)
print(resp.status)
```

Produces the following tcpdump output:

```
12:45:22.795720 IP 192.168.1.3.50182 > 50.16.195.223.80: Flags [S], seq 2772615110, win 65535, options [mss 1460,nop,wscale 5,nop,nop,TS val 494752945 ecr 0,sackOK,eol], length 0
12:45:22.896051 IP 50.16.195.223.80 > 192.168.1.3.50182: Flags [S.], seq 156185054, ack 2772615111, win 26847, options [mss 1452,sackOK,TS val 26566565 ecr 494752945,nop,wscale 8], length 0
12:45:22.896170 IP 192.168.1.3.50182 > 50.16.195.223.80: Flags [.], ack 1, win 4140, options [nop,nop,TS val 494753044 ecr 26566565], length 0
12:45:22.896310 IP 192.168.1.3.50182 > 50.16.195.223.80: Flags [P.], seq 1:75, ack 1, win 4140, options [nop,nop,TS val 494753044 ecr 26566565], length 74: HTTP: GET /status/404 HTTP/1.1
12:45:22.988213 IP 50.16.195.223.80 > 192.168.1.3.50182: Flags [.], ack 75, win 105, options [nop,nop,TS val 26566588 ecr 494753044], length 0
12:45:22.992870 IP 50.16.195.223.80 > 192.168.1.3.50182: Flags [P.], seq 1:260, ack 75, win 105, options [nop,nop,TS val 26566589 ecr 494753044], length 259: HTTP: HTTP/1.1 404 NOT FOUND
[...]

PR #1162, opened 2017-04-27
Add alternative FIFO queue as a data structure for the connection pool

masteralves:
This comes from issue #1158 
I am suggesting this change because for some systems it is
better to use all the connections from the pool equally, rather than the
same ones all the time, which is the default case with LIFO.

In our case, it improves reliability because our servers are behind an
ELB. Once there is a failure and we have to retry the request, we don't
want to reuse the same connection, but rather start a different connection
in the hope that the ELB will choose a different machine to respond.

Furthermore, an online search leads me to believe that FIFO is a commonly
used data structure in connection pools:
https://en.wikipedia.org/wiki/Object_pool_pattern#Examples
The C# example uses List.Remove(0) and List.Add(...)

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -230,5 +230,8 @@ In chronological order:
 * Tuukka Mustonen <tuukka.mustonen@gmail.com>
   * Add counter for status_forcelist retries.
 
+* João Alves <joao.moa@gmail.com>
+  * Add fifo queue as an alternative for connection pool data structure.
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -417,6 +417,10 @@ class TestConnectionPool(unittest.TestCase):
                                 preload_content=False)
         self.assertTrue(isinstance(response, CustomHTTPResponse))
 
+    def test_fifo_queue(self):
+        pool = HTTPConnectionPool(host='localhost', fifo_queue=True, maxsize=1, block=True)
+        self.addCleanup(pool.close)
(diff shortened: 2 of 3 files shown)


masteralves:
Well... You would need both `HTTPConnectionPool` and `HTTPSConnectionPool` custom subclasses and this looks like something that should be common in any pool. The data structure is a core component of the pool afterall. Also, I didn't see any other custom subclasses, so I guess creating one only for this small change would be overkill...

Lukasa:
I don't mean merging the custom subclasses into urllib3, I mean overriding them in your own codebase. Should be about 12 LoC total to make this change.

masteralves:
Yeah, we could do that, the reason I proposed this change is because I thought it could be useful for other people as well and overriding is kind of a workaround. ELBs are common as well as FIFO pools, more people can have this problem. If you think we should do that, then I'm happy to change only our own codebase and forget about this pull request.

Lukasa:
Overriding isn't a workaround, this is an entirely intentional mechanism for hooking into urllib3. We expect people to use it: that's why we put it there. :smile:

masteralves:
What I mean by workaround is "a less ideal solution than setting a boolean :)"

Lukasa:
Hah, yes, that's true. But ultimately I'm not yet convinced that the additional code complexity is worthwhile across our entire userbase. If others report similar concerns we can worry about it then, but for now I think a better approach is for you to use our extension mechanisms to resolve this in your codebase.

masteralves:
Ok that's fair. Thanks for your help :+1:

Lukasa:
No problem, thanks for your contribution!

yunstanford:
actually, i have the same issue here, when we do rolling deployment for instances behind elb.

Issue #1160, opened 2017-04-27
Abandon nosetests.


Lukasa:
Sadly, nosetests appears to be abandonware according [to their documentation](https://nose.readthedocs.io/en/latest/):

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership.

Given that we will now need to migrate our testing infrastructure away from nose, we should consider where we need to move. My suggestion would be to move to `py.test`: writing tests for `py.test` is generally a nicer experience than writing traditional unittest style tests, and it has fewer requirements to do weird unittest importing. However, I'm happy to get feedback from the other contributors about what they'd like to do.

The work required to complete this would be as follows:

- [ ] Initial PR to move from nosetests as the test runner to the new runner, with only the changes to the testing we actually need to get those tests to run.
- [ ] Subsequent PRs to incrementally migrate our test suite to the style of the new runner, instead of unittest style, if applicable.

Thoughts?

SethMichaelLarson:
I would love to see a move to `py.test`. Seems to be the best way to go.

haikuginger:
I'm a fan of `py.test`. It makes it a lot easier to just `assert var is None` instead of having to figure out the specific assertion method to call.

nateprewitt:
I'm definitely in favor of moving to `py.test` and would be happy to help with any of the migration work that needs to be done.

SethMichaelLarson:
The nice thing about this migration is it can be done in chunks because `py.test` supports `unittest` and `pytest` styled tests.

SethMichaelLarson:
So far the biggest blocker I've found is our use of `nose.plugins.skip.SkipTest` at the module level, something that `pytest` I don't think supports in a 1-to-1 fashion.

alexwlchan:
> So far the biggest blocker I've found is our use of nose.plugins.skip.SkipTest at the module level, something that pytest I don't think supports in a 1-to-1 fashion.

I think you can use:

```python
pytestmark = pytest.mark.skipif(reason="foo")
```

JordanP:
py.test is nice but did we consider using the good old `unittest` package ? It's part of the stdlib, brings 0 external lib and should be known by most (all ?) Python developer. 

py.test has some learning curve. Just asking, what is in py.test that can't be done with unittest ? With setup() and setupClass(), and 2/3 helper functions, my experience tells me we can do a lot.

Lukasa:
[...]

PR #1163, opened 2017-04-27
Switch from Nose to Pytest Runner

SethMichaelLarson:
The first step in the effort to switch to `py.test` as opposed to `nose`/`unittest` (#1160).
So far I've hit two problems: 
- Apparently `py.test` reloads the entire module per-test and emits the "No true SSLContext object available" warning for each test which luckily only affects Python 2.6 `test_https.TestHTTPS.test_ssl_wrong_system_time` and `test_https.TestHTTPS.test_ssl_correct_system_time` tests. Id' advocate just skipping these tests on Python versions without a correct SSLContext.

- Google App Engine, I don't know much about it or how testbeds work so I could use some direction here.

### DIFF
--- a/test/appengine/nose.cfg
+++ b/test/appengine/nose.cfg
@@ -1,4 +1,0 @@
-[nosetests]
-cover-min-percentage=0
-with-gae=1
-gae-application=test/appengine/app.yaml
--- a/test/contrib/test_gae_manager.py
+++ b/test/contrib/test_gae_manager.py
@@ -29,6 +29,7 @@ def setup_module(module):
     if not HAS_APPENGINE_SDK:
         pytest.skip('Tests require Google AppEngine SDK.')
 
+
 # This class is used so we can re-use the tests from the connection pool.
 # It proxies all requests to the manager.
 class MockPool(object):
--- a/test/contrib/test_gae_manager.py
+++ b/test/contrib/test_gae_manager.py
@@ -30,6 +30,21 @@ def setup_module(module):
         pytest.skip('Tests require Google AppEngine SDK.')
 
 
+def activate_urlfetch():
+    from google.appengine.ext import testbed
+
+    bed = testbed.Testbed()
+    bed.activate()
+    bed.init_urlfetch_stub()
+
+    return bed
+
+
+def deactivate_urlfetch(bed):
+    if bed is not None:
+        bed.deactivate()
+
+
 # This class is used so we can re-use the tests from the connection pool.
 # It proxies all requests to the manager.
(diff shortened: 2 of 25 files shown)


test/appengine/nose.cfg
```diff
@@ -1,4 +0,0 @@
-[nosetests]
-cover-min-percentage=0
-with-gae=1
-gae-application=test/appengine/app.yaml
```
Lukasa: Yeah, these look like things that will require manual intervention to reproduce.

test/contrib/test_gae_manager.py
```diff
@@ -57,9 +61,6 @@ def _absolute_url(self, path):
 class TestGAEConnectionManager(TestConnectionPool):
     __test__ = True
 
-    # Magic class variable that tells NoseGAE to enable the URLFetch stub.
-    nosegae_urlfetch = True
```
Lukasa: Do we need an equivalent to this for pytest_gae?

SethMichaelLarson: I'm reading the docs as we speak. :)

SethMichaelLarson: Okay I think I'm going to switch from pytest_gae which looks like a unmaintained project (?) to pytest_beds which at least mentions urlfetch in it's docs. Will do some more research.

SethMichaelLarson:
I've switched from `warnings.catch_warnings()` to using `pytest.warns(None)`, this might resolve our issue with Python 2.6 but we shall see.

SethMichaelLarson:
Nope! Python 2.6 still throwing that SSLContext warning. I'll skip on those Python versions.

SethMichaelLarson:
Hitting one of the pain points of `py.test`-style tests with GAE tests. Fixtures and test inheritance don't mix well. This is the only thing I dislike about `py.test` compared to `unittest`/`nose`. :cry:

SethMichaelLarson:
Alright I'm stuck for now, I don't have an AppEngine environment to manually track down why it's not calling `google.appengine.api.urlfetch.fetch` in `TestHTTP.test_urlfetch_called_with_http`. I think this is the only issue that's remaining for this PR.

SethMichaelLarson:
Pretty sure you didn't mean to approve these changes @Lukasa, would be a train wreck to merge now. :wink:

SethMichaelLarson:
I don't have enough time to get this PR 100% ready to merge as I'm stuck on GAE and don't have time to create my own local environment to test things.  Seems as though @alexwlchan is putting this work in so I'll close this.

PR #1168, opened 2017-05-01
Add JythonSelectSelector to urllib3/util/selectors.py [WIP]

SethMichaelLarson:
As reported by @LordGaav Jython apparently over-rides the default behavior of `socket.socket.fileno()` and instead returns a `socket`. The current implementation of `BaseSelector` relies heavily on the hash-ability of the return value of `fileobj.fileno()`.  This is just a quick implementation, there may be issues with it. I would also like to actually test it on Jython and if that's not possible in our CI setup perhaps make some `mock` tests that simulate a `socket` that returns itself from `fileno()`.

References:
- kennethreitz/requests#3992
- http://bugs.jython.org/issue1678
- https://wiki.python.org/jython/NewSocketModule#socket.fileno.28.29_does_not_return_an_integer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,9 @@ env:
 # https://github.com/travis-ci/travis-ci/issues/4794
 matrix:
   include:
+    - python: 2.7
+      env:
+        - TOXENV=jython JYTHON=org.python:jython-installer:2.7.0
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -8,6 +8,9 @@ if [ -n "$JYTHON" ]; then
     jip install $JYTHON
     OLD_VIRTUAL_ENV=$VIRTUAL_ENV
     java -jar $OLD_VIRTUAL_ENV/javalib/jython-installer-2.7.0.jar -s -d $HOME/jython
+
+    # Required for --distribute option.
+    pip install virtualenv==1.9.1
     virtualenv --distribute -p $HOME/jython/bin/jython $HOME/jenv
 fi
 
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -13,9 +13,8 @@ if [ -n "$JYTHON" ]; then
     pip install virtualenv==1.9.1
     virtualenv --distribute -p $HOME/jython/bin/jython $HOME/jenv
     source $HOME/jenv/bin/activate
-fi
 
-if [[ "$(uname -s)" == 'Darwin' ]]; then
+elif [[ "$(uname -s)" == 'Darwin' ]]; then
     sw_vers
     brew update || brew update
 
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -6,9 +6,8 @@ set -x
 if [ -n "$JYTHON" ]; then
     pip install jip
     jip install $JYTHON
-    _JYTHON_BASENAME=${NON_GROUP_ID/:/-}
     OLD_VIRTUAL_ENV=$VIRTUAL_ENV
-    java -jar $OLD_VIRTUAL_ENV/javalib/${_JYTHON_BASENAME}.jar -s -d $HOME/jython
+    java -jar $OLD_VIRTUAL_ENV/javalib/jython-installer-2.7.0.jar -s -d $HOME/jython
     virtualenv --distribute -p $HOME/jython/bin/jython $HOME/jenv
 fi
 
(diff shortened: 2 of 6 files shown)


LordGaav:
Great work, let me know if there is something I need to do / test.

SethMichaelLarson:
So I have tested this locally on my own copy of Jython 2.7.0 and it works correctly now. I'm just trying to get a build in our CI with Jython to see if I can skirt writing intense mocking tests.

SethMichaelLarson:
Starting to think this might not be feasible if we can't install `tox` in Jython.

Lukasa:
Sure? tox shouldn't need to be installed in a Jython env.

SethMichaelLarson:
Nevermind you can install tox even in Jython, we won't be able to get coverage though because `coverage` doesn't support Jython. Still good to test functionality.

SethMichaelLarson:
Jython's implementation of sockets and/or select is just wrong. Sockets with no data to read sometimes appear as readable to `select.select`.

```python
from test.socketpair_helper import socketpair
import select
r, w = socketpair()
select.select([r], [], [], 1)  # This returns: ([<_socket._socketobject object at 0x19b>], [], [])
r.recv(1)  # This returns: ''
```

For reference, in CPython `select.select` returns three empty lists and `r.recv(1)` blocks indefinitely.
Is this something we actually want to pursue? Because the selector does work in Jython and it seems like it works for pip and requests when I sub in `urllib3/util/selectors.py` but writing tests for this is a nightmare because of how flaky and strange Jython acts with select and sockets.

Lukasa:
Wait, Jython reads the empty string on a non-EOF condition?

SethMichaelLarson:
Seems to, but not always. Most of the time it behaves correctly.

SethMichaelLarson:
From my brief test it appears to occur about 25% of the time:
```python
from test.socketpair_helper import socketpair
readers = []
for _ in range(100):
    readers.append(socketpair()[0])

import select
r, _, _ = select.select(readers, [], [], 1)

print(len(r)) # 26
```

SethMichaelLarson:
The same goes for writability, sockets that should always be writable appear to be not writable:

```python
from test.socketpair_helper import socketpair
writers = []
for _ in range(100):
    writers.append(socketpair()[0])

import select
_, w, _ = select.select([], writers, [], 1)

print(len(w)) # 58
```

Lukasa:
Ok, this feels like a weird behaviour that suggests that Jython sockets really don't behave as I'd expect. If you force the socket mode to non-blocking does that help?

SethMichaelLarson:
```
>>> from test.socketpair_helper import socketpair
>>> writers = []
>>> for _ in range(100):
...     _, wr = socketpair()
...     wr.setblocking(False)
...     writers.append(wr)
... 
>>> import select
>>> _, w, _ = select.select([], writers, [], 1)
>>> len(w)
69
```

Same result it seems for writing.

SethMichaelLarson:
FYI I believe each socket in Jython uses a [`io.netty.channel.nio.NioEventLoopGroup`](https://github.com/jythontools/jython/blob/master/Lib/_socket.py#L51) which I think uses a Thread Pool for... something? I know this because if I try to allocate a ton of sockets Jython will complain about not being able to allocate another thread pool.

SethMichaelLarson:
Okay then. Jython implements a function called `select.cpython_compatible_select` :scream:. * Apparently * does what it says on the tin, still not completely right though.

```
>>> from test.socketpair_helper import socketpair
>>> from select import cpython_compatible_select as select
>>> writers = []
>>> for _ in range(100):
...     writers.append(socketpair()[0])
... 
>>> _, w, _ = select([], writers, [], 1)
>>> len(w)
91
```

SethMichaelLarson:
What are your thoughts on this @Lukasa? Would like some comments on direction before continuing work on this.

Lukasa:
Soooooo...I'm -1 if we can't test it. If we *can* test it I'm probably +0? I don't think it hurts us to get Jython support back, so if we can get test coverage I'm happy with it.

haikuginger:
I'm probably in a similar place to @Lukasa, although it grosses me out that it's necessary. If Jython's low-level API is that different from CPython's, it's hard to imagine being able to build portable code anyway - although urllib3 being portable is probably the first step.

SethMichaelLarson:
Forgot this PR was still open. I added a `JythonSelectSelector` to selectors2 a while ago. It'll be there as a sort of reference implementation if this is desired in the future.

PR #1170, opened 2017-05-02
Add TLS 1.3 cipher suites

tiran:
Constants were taken from OpenSSL master and https://tools.ietf.org/html/draft-ietf-tls-tls13-18#page-103

@Lukasa I don't have resources to test ST patch.


Lukasa:
That's ok, Travis will test SecureTransport for us. :D

urllib3/contrib/securetransport.py
```diff
@@ -91,6 +91,9 @@
 # individual cipher suites. We need to do this becuase this is how
 # SecureTransport wants them.
 CIPHER_SUITES = [
+    SecurityConst.TLS_AES_128_GCM_SHA256,
+    SecurityConst.TLS_AES_256_GCM_SHA384,
+    SecurityConst.TLS_CHACHA20_POLY1305_SHA256,
```
Lukasa: Can we keep this list in the same order for both OpenSSL and SecureTransport, please?

urllib3/util/ssl_.py
```diff
@@ -61,13 +61,17 @@ def _const_compare_digest_backport(a, b):
 # - https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
 #
 # The general intent is:
+# - Enable TLS 1.3 cipher suites
```
Lukasa: Let's have this say "prefer", not "enable". =)

PR #1171, opened 2017-05-06
Fix deadlocks with pool_block=True and retries

mattbillenstein:
Discard connection on retry/redirect before issuing another request to prevent deadlocks with pool_block=True

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -687,11 +687,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
                 if retries.raise_on_redirect:
-                    # Release the connection for this response, since we're not
-                    # returning it to be released manually.
-                    response.release_conn()
+                    # Discarding the connection for this response, since we're
+                    # not returning it to be released manually.
+                    response.discard_conn()
                     raise
                 return response
+
+            response.discard_conn()
 
             retries.sleep_for_retry(response)
             log.debug("Redirecting %s -> %s", url, redirect_location)
@@ -710,11 +712,14 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
                 if retries.raise_on_status:
-                    # Release the connection for this response, since we're not
+                    # Discarding the connection for this response, since we're not
                     # returning it to be released manually.
-                    response.release_conn()
+                    response.discard_conn()
                     raise
                 return response
+
+            response.discard_conn()
+
             retries.sleep(response)
             log.debug("Retry: %s", url)
             return self.urlopen(
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -178,6 +178,13 @@ class HTTPResponse(io.IOBase):
         self._pool._put_conn(self._connection)
         self._connection = None
 
+    def discard_conn(self):
+        if not self._pool or not self._connection:
+            return
+
+        self._connection = self._connection and self._connection.close()
+        self._pool._put_conn(self._connection)
+
     @property
     def data(self):
         # For backwords-compat with earlier urllib3 0.4 and earlier.


mattbillenstein:
We discard the connection when not returning the immediate response -- also after the try/except there is a call to discard it before we recurse -- the main problem was when we would recurse, we still held onto the connection, so we would grab another one on recurse and exhaust the pool causing a deadlock.

Lukasa:
My specific concern is why we needed a new "discard" method. I think the `release_conn` method was absolutely fine: what makes you think it wasn't?

mattbillenstein:
I did try a version using release_conn, but I would later get connection errors when the connection was reused as it seems  just releasing the connection can put it back in the pool in a bad state:

      File "/Users/mattb/src/foo/lib/python2.7/site-packages/gevent/greenlet.py", line 536, in run
        result = self._run(*self.args, **self.kwargs)
      File "./retry_pool_hang_gevent.py", line 48, in worker
        res = session.get(url, proxies=proxies)
      File "/Users/mattb/src/foo/lib/python2.7/site-packages/requests/sessions.py", line 520, in get
        return self.request('GET', url, **kwargs)
      File "/Users/mattb/src/foo/lib/python2.7/site-packages/requests/sessions.py", line 507, in request
      File "/Users/mattb/src/foo/lib/python2.7/site-packages/requests/sessions.py", line 628, in send
      File "/Users/mattb/src/foo/lib/python2.7/site-packages/requests/adapters.py", line 502, in send
    ConnectionError: HTTPConnectionPool(host='local.umt.io', port=80): Max retries exceeded with url: /404 (Caused by ProtocolError('Connection aborted.', ResponseNotReady()))
    Sun May  7 13:20:12 2017 <Greenlet at 0x102b49550: worker> failed with ConnectionError

So discard_conn closes it, and removes references to it both in the response object and the pool.

Lukasa:
So that should not be happening. If we can reproduce that bug, we should aim to fix it, but I'd like to see a test that reproduces it before we leap to any conclusions. 😁

mattbillenstein:
I'm using this testcase to test:  https://github.com/mattbillenstein/requests-retry-pool-hang  -- the gevent version will raise a Deadlocked exception when it fails; the threaded version will just hang forever.

Installing https://github.com/mattbillenstein/requests with this patch will run fine, changing discard_conn to release_conn emits the ConnectionError exception above.

Lukasa:
So it'd be much better to try to reduce that down to a small test-case that doesn't have confounding factors. In this case, that'd mean removing Requests from the case altogether and then trying to boil down to the essence of the problem.

In this case, the essence of the problem seems to be: if you have a `Retry` object with `raise_on_redirect` set to `True`, and it raises, the connection is not torn down if there is a response body being sent (which it should be).

This should be very simple to test: a simple HTTP server should be enough to provide an example (or even just a single socket repeatedly vomiting out the same HTTP response), and a few lines of test code setting up a `Retry` object that allows zero redirects and that raises on redirect. If we can demonstrate the problem in that form then we'll be in a much better place to move forward.

I realise that I'm not being clear about why I'm pushing back like this. The TL;DR is that I don't want to add a `discard_conn` method: it's an attractive nuisance and it makes manual a step that has previously been automatic. If there is a flaw in the logic used in `discard_conn` we should fix it, rather than patch over the problem by using a different method internally.

mattbillenstein:
raise_on_redirect is not really the case I'm having a problem with or testing -- again, this is with pool_block=True and retries.  Connections are not being freed before recursing on the retry, so the pool is being exhausted.

Here is that same testcase without requests:  https://github.com/mattbillenstein/urllib3-retry-pool-hang

Lukasa:
Raise on redirect may not be the problem you're having, but this patch includes a change to it, so I want to understand the motivation behind that change. :smile:

mattbillenstein:
I guess consistency -- in my testing, I couldn't simply release the connection back to the pool since there may be some response content still in the connection.

It seems that in the retry/redirect case that either the response should be returned with a reference to the connection, or the connection should be closed and freed entirely.

And I'm not married to this pull request -- it just seems like a deadlock in a library as prominent as requests with fairly conventional usage as the testcase has demonstrated is a pretty major problem.  This PR is just something that works for me and as I understand the code seems to fix connection leaks which go unnoticed due to the "leaky bucket" abstraction with pool_block=False.  If you have a better understanding of this and a better idea of how it should be fixed, I'll gladly review/test that.

Additionally regarding connection-pool churn --  since the current code holds onto connections during the retry, it connects to the server again if the pool is full and then discards that new connection if the pool remains full.  So this PR just discards the failed connection before the retry, creates a new connection, and then has a chance to put that connection to the pool given we just freed a slot.  We're also not holding connection state during the retry if we close it first.  So, it seems no worse, and possibly better since we replaced a possibly bad connection with a new one instead of leaving a bad connection in the pool and creating and discarding a new one.

Lukasa:
> I guess consistency -- in my testing, I couldn't simply release the connection back to the pool since there may be some response content still in the connection.

Sure, again, I have no problem with that. :smile:

Again, let me be clear. I have an objection to adding `discard_conn`, for two reasons:

1. It's an extra method in our public API that we don't need.
2. It causes us to be inefficient in connection re-use.

Wherever possible we should aim to preserve connections, rather than throw them away. That very much includes cases like this, where `pool_block` is set to `True` presumably to limit load on the server, and so throwing away connections on redirect and status-based retry is a waste. The connection is absolutely fine: the only reason we're throwing it away is because there may be unconsumed bytes on the wire.

I'd vastly prefer we resolve this properly, by consuming any unconsumed bytes there may be. This is pretty easily done by adding this stanza before the releasing:

```python
if not response.closed:
    response.read()
```

This will cause us to block to consume whatever data is still waiting in the pipe and then move on, which is vastly preferable than throwing the connection away and doing another TCP (+TLS) handshake.

> And I'm not married to this pull request -- it just seems like a deadlock in a library as prominent as requests with fairly conventional usage as the testcase has demonstrated is a pretty major problem. This PR is just something that works for me and as I understand the code seems to fix connection leaks which go unnoticed due to the "leaky bucket" abstraction with pool_block=False. If you have a better understanding of this and a better idea of how it should be fixed, I'll gladly review/test that.

I'm sorry if it seems like I don't care or have been pushing back too hard. The reason I'm doing that is precisely *because* Requests is prominent: the first idea that jumps into our head that fixes the problem is rarely the proper long-term sustainable fix, and I'd rather we delay fixing it to get it right than shove in the first thing that comes along. This is because fixes like these have a tendency to stick around: once the problem no longer manifests, people have a tendency to forget to come back to do it properly. This goes double for any short-term fix that adds to the API surface.

So my suggestion would be, at this time, to write tests that exercise these code-paths. This shouldn't be too hard: the socketlevel tests in the repository are a good place to start. Writing one of those that never accepts a new socket, and so must *always* use just one connection, and that triggers a retry with a blocking connection pool would be a good start. This should let us get to the place where these retries can fire and re-use the same connection, which is what we'd like them to do.

Once we do that you can add in the calls to `read()` and we can discuss if a refactor is needed. At some point we probably also want to deal with the fact that `PoolManager` has similar redirect logic in it that may also need enhancing.

Does that sound reasonable to you?

mattbillenstein:
Alright, I like that fix better -- opened a new PR for that one.

PR #1173, opened 2017-05-08
Rebuild sync connection loop.

### LINKED ISSUE #1149: sync_connection does not correctly handle SSLWantRead and SSLWantWrite
If either of `SSLWantRead` or `SSLWantWrite` are raised in the context of `sync_connection._receive_bytes()` on the V2 branch, the error is raised, rather than being interpreted and handled with appropriate action.

Lukasa:
This resolves #1149.

Sadly, it turns out that resolving #1149 required a pretty substantial rethink of the underlying sending loop in the v2 branch. A lot of the code was subtly mishandling expectations: for example, it's not at all unreasonable for a TLS connection to become readable during upload, and that's not an error: it is just triggering a TLS renegotiation. So we needed to become a bit smarter about error handling there.

Ultimately, this has required a big rewrite and the addition of a brand new bit of testing harness. We now have a sort-of DSL for testing low-level socket activity in the sync_connection, which allows us to tightly control exactly what happens in the low-level sockets and selectors. That transitions us into a purely-deterministic world in which nothing unexpected happens. Ultimately this may be a better testing approach than even the socket-level tests, at the cost of being divorced from "actual network behaviour".

### DIFF
--- a/test/contrib/test_securetransport.py
+++ b/test/contrib/test_securetransport.py
@@ -7,6 +7,8 @@ try:
 except ImportError as e:
     raise SkipTest('Could not import SecureTransport: %r' % e)
 
+raise SkipTest('SecureTransport currently not supported in v2!')
+
 from ..with_dummyserver.test_https import TestHTTPS, TestHTTPS_TLSv1  # noqa: F401
 from ..with_dummyserver.test_socketlevel import (  # noqa: F401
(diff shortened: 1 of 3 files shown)


haikuginger:
Will review this week.

Lukasa:
:heart:

PR #1174, opened 2017-05-08
Fix deadlocks with pool_block=True and retries

mattbillenstein:
Replaces #1171 

I'll look at including a test this week.


mattbillenstein:
Ah, yes, updated the PR.

urllib3/connectionpool.py
```diff
@@ -687,12 +687,17 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
                 if retries.raise_on_redirect:
-                    # Release the connection for this response, since we're not
-                    # returning it to be released manually.
+                    # Drain and release the connection for this response, since
+                    # we're not returning it to be released manually.
+                    response.read()
                     response.release_conn()
```
Lukasa: Cool, so with all that code duplication about it'd be nice to pull this out to a handy-dandy helper function. Declaring it inline in `urlopen` would probably be enough.

mattbillenstein: Hmm, yeah, I was sorta thinking it would be nice if the response did this for us -- it wouldn't make sense to release a connection back to the pool with data still in the pipe and it seems a bit awkward that a consumer of this lib would have to write the same pair of calls in their code.

Lukasa: @mattbillenstein Yeah, so I had that thought too. I think it's possible that `release_conn` should check connection state. However, I think for `release_conn` the calculation is different: if the response is not complete when `release_conn` is called, probably the connection should be *closed*, not drained. Draining the connection here is still a specific-to-this-use-case operation, because here we have confidence that the connection is fine. However, when a user calls `release_conn` for any reason, we have no idea what's going on, and don't have any good way to keep track of what we think is going on.

So, I'd consider the enhancement to `release_conn` to be a separate concern, IMO: in this case we'll still want to drain the connection.

Lukasa: I'm disinclined to extend the API surface of `Response` for this. It represents a particularly unusual use-case.

mattbillenstein: response.release_conn(drain=True) ?

Lukasa: Nope. =) This still represents an API extension, and additionally represents that kind of API wart which is a method that totally varies in its behaviour based on a boolean flag. Always better to have two methods than a method like that.

mattbillenstein:
Added tests for this -- seems to be a single test failure on python 3.3 -- could you take a look?

urllib3/connectionpool.py
```diff
@@ -666,6 +666,10 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
                                 release_conn=release_conn, body_pos=body_pos,
                                 **response_kw)
 
+        def drain_and_release_conn(response):
+            response.read()
```
Lukasa: Here's a question: should we catch exceptions here? I feel like the answer is probably yes: they'll be ultimately a bit misleading if we hit them.

mattbillenstein: Maybe yes.

In the case of a redirect, the entire response has probably already been read off the wire and is sitting in a buffer.  And this is probably likely for many of the non-200 cases as well?

I'm going to push this to my production systems and see if I find any exceptions in this.

Lukasa: The general concern is that we have, in the past, bumped into malformed responses in the past, including on redirects.

mattbillenstein: Wouldn't that raise before getting here?  We've parsed the redirect headers; this is just throwing away the response body; but I haven't traced through any of the code yet.

In any case, to be safe, happy to catch and discard the conn, but I'm not completely sure which types of exceptions we should try to catch here.

Lukasa: Not necessarily: the body may be chunk encoded, or overlong, or a decent number of other possible wacky framing errors.

mattbillenstein: How do you guys wish to proceed here?

Lukasa: I think we should catch exceptions, and if we hit it then we want to discard the connection instead of re-use it.

mattbillenstein: Which exception types?  The handler above catches:

except (TimeoutError, HTTPException, SocketError, ProtocolError,
                BaseSSLError, SSLError, CertificateError) as e:

Lukasa: "The handler above"?

mattbillenstein: The exception handler directly above this function...

Lukasa: Ah, I see. Yeah, we probably need most of those. `CertificateError` we don't need, but the rest seem like good things to catch.

mattbillenstein: Ok, so that leads us back to how to properly discard the connection without modifying the Response interface?

Lukasa: Nope. =) `Response` will handle this appropriately. In fact, we can even remove the explicit call to `release_conn` here as `Response` will do that automatically once the entire body is read.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -987,5 +987,32 @@ def tell(self):
             self.assertTrue('Unable to record file position for' in str(e))
 
 
+class TestRetryPoolSize(HTTPDummyServerTestCase):
+    def setUp(self):
+        retries = Retry(
+            total=3,
+            backoff_factor=0.01,
```
Lukasa: Do we need a backoff factor?

test/with_dummyserver/test_connectionpool.py
```diff
@@ -987,5 +987,32 @@ def tell(self):
...
+        self.addCleanup(self.pool.close)
+
+    def test_pool_size_retry(self):
+        self.pool.urlopen('GET', '/not_found', preload_content=False)
+        self.assertEquals(self.pool.num_connections, 1)
+
+
+class TestRedirectPoolSize(HTTPDummyServerTestCase):
+    def setUp(self):
+        self.pool = HTTPConnectionPool(self.host, self.port, maxsize=10, retries=5, block=True)
```
Lukasa: It might be worth adding some more verbose retry setup here to guarantee that redirects are in fact turned on, just in case this API changes for any reason.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -1006,7 +1005,13 @@ def test_pool_size_retry(self):
 
 class TestRedirectPoolSize(HTTPDummyServerTestCase):
     def setUp(self):
-        self.pool = HTTPConnectionPool(self.host, self.port, maxsize=10, retries=5, block=True)
+        retries = Retry(
+            total=3,
+            raise_on_status=False,
+            status_forcelist=[404],
+        )
```
Lukasa: This needs to set `redirect=True` to avoid breaking if we ever change the kwarg defaults.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -987,5 +987,38 @@ def tell(self):
...
+        self.assertEquals(self.pool.num_connections, 1)
+
+
+class TestRedirectPoolSize(HTTPDummyServerTestCase):
+    def setUp(self):
+        retries = Retry(
+            total=3,
+            raise_on_status=False,
+            status_forcelist=[404],
+            redirect = True,
```
Lukasa: Mind removing the two spaces? Sorry. :cry:

urllib3/connectionpool.py
```diff
@@ -666,6 +666,15 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
                                 **response_kw)
 
+        def drain_and_release_conn(response):
+            try:
+                # discard any remaining response body, the connection will be
+                # released back to the pool once the entire response is read
+                response.read()
+            except (TimeoutError, HTTPException, SocketError, ProtocolError,
+                    BaseSSLError, SSLError) as e:
+                pass
```
Lukasa: We're going to need a test that covers this sensibly: that'll basically mean sending back a malformed HTTP response to trigger this error, or at least triggering a timeout.

mattbillenstein: Is there an example of how to do something like this in one of the existing tests?

Lukasa: Sure, take a look at [this one](https://github.com/shazow/urllib3/blob/master/test/with_dummyserver/test_socketlevel.py#L282).

mattbillenstein: Added a test and verified it raises ProtocolError in that function...

test/with_dummyserver/test_connectionpool.py
```diff
@@ -987,5 +987,64 @@ def tell(self):
...
+            total=3,
+            raise_on_status=False,
+            status_forcelist=[404],
+        )
+        pool = HTTPConnectionPool(self.host, self.port, maxsize=10,
+                                       retries=retries, block=True)
+        self.addCleanup(pool.close)
+
+        pool.urlopen('GET', '/not_found', preload_content=False)
+        self.assertEquals(pool.num_connections, 1)
```
Lukasa: Can this be moved to `test_socketlevel.py` please?

Lukasa:
You've got a flake8 violation: mind resolving it? :grin:

mattbillenstein:
Yeah, sorry, keep forgetting to check that -- should be clean now.

Lukasa:
Still failing. :wink:

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1393,3 +1393,36 @@ def socket_handler(listener):
         self.assertEqual(len(data), 0)
 
         done_event.set()
+
+
+class TestRetryPoolSizeDrainFail(SocketDummyServerTestCase):
+
+    def test_pool_size_retry_drain_fail(self):
+        def socket_handler(listener):
+            while 1:
```
Lukasa: We don't want this to run forever: can we make sure it only runs the number of times we expect it to? (Should be twice)

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1399,7 +1399,7 @@ class TestRetryPoolSizeDrainFail(SocketDummyServerTestCase):
 
     def test_pool_size_retry_drain_fail(self):
         def socket_handler(listener):
-            while 1:
+            for _ in xrange(2):
```
Lukasa: `xrange` isn't present on Python 3. We don't mind an intermediate list being created on Py 2 because it's only two elements long, so regular `range` is fine.

PR #1177, opened 2017-05-15
fix infinity recursion error when using with gevent.

luxrck:


Lukasa:
I do not believe we should bend over backwards to appease code that monkeypatches. When using gevent, just apply the monkeypatch before you import anything else. This is good general practice, and will save you from pain in the future.

Issue #1112, opened 2017-02-08
Bug: No Retry for SSL Error


lukas-gitl:
That following code should fire off a `MaxRetryError`, but instead doen't retry and throws an `SSLError`.

**Test case:**
_server.py_
```python
import time
import socket

s = socket.socket()
s.bind(('', 4433))
s.listen(10)

while True:
    news, _ = s.accept()
    time.sleep(0.5)
    news.close()
```
_client.py_
```python
import urllib3
http = urllib3.PoolManager()
http.request('GET', 'https://localhost:4433/', retries=5)
```

<sup>*Disclaimer:* Test case was created by [**Lukasa**](https://github.com/kennethreitz/requests/issues/3845#issuecomment-278273421)</sup>

SethMichaelLarson:
Doesn't look like you're setting up HTTPS on your server. It's just a regular non-SSL socket.

Lukasa:
Ok, so my analysis of this bug is just that we do not retry on `SSLError`, and we probably should. The code block in `ConnectionPool.urlopen` is the one that makes this call, and I think it's just not right. I'm inclined to say that the blocks for `BaseSSLError`, `CertificateError`, and `SSLError` should all be rolled down into the block below.

Thoughts? This is something that would be good for one of the regular contributors to pick up if they want to. =)

piquadrat:
I'm currently trying to implement this as proposed by @Lukasa, basically like this:

    diff --git a/urllib3/connectionpool.py b/urllib3/connectionpool.py
    index b4f1166..031df4c 100644
    --- a/urllib3/connectionpool.py
    +++ b/urllib3/connectionpool.py
    @@ -622,26 +622,16 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 # Timed out by queue.
                 raise EmptyPoolError(self, "No pool connections are available.")

    -        except (BaseSSLError, CertificateError) as e:
    -            # Close the connection. If a connection is reused on which there
    -            # was a Certificate error, the next request will certainly raise
    -            # another Certificate error.
    -            clean_exit = False
    -            raise SSLError(e)
    -
    -        except SSLError:
    -            # Treat SSLError separately from BaseSSLError to preserve
    -            # traceback.
    -            clean_exit = False
    -            raise
    -
    -        except (TimeoutError, HTTPException, SocketError, ProtocolError) as e:
    +        except (TimeoutError, HTTPException, SocketError, ProtocolError,
[...]

PR #1178, opened 2017-05-15
Retry SSL Errors

beniwohli:
This is a fix for #1112. Instead of immediately raising SSLErrors, they are handled by the retry mechanism.

The change introduced ~22 test failures, all of them due to the fact that requests failed due to SSL errors now raise a `MaxRetryError`, and not an `SSLError`. I adapted these tests, also checking that  `MaxRetryError.reason` is an `SSLError`.

As mentioned in #1112, this is somewhat backwards incompatible: if clients explicitly handle `SSLError` exceptions, they now need to catch `MaxRetryError` instead and check if `reason` is an `SSLError`.

I didn't add an explicit test, since the new behaviour is tested implicitly in a lot of existing tests. Is that ok?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changes
 dev (master)
 ------------
 
+* Made the connection pool retry on ``SSLError``.  The original ``SSLError``
+  is available on ``MaxRetryError.reason``. (Issue #1112)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -36,7 +36,7 @@ from urllib3.exceptions import (
     InsecureRequestWarning,
     SystemTimeWarning,
     InsecurePlatformWarning,
-)
+    MaxRetryError)
 from urllib3.packages import six
 from urllib3.util.timeout import Timeout
 import urllib3.util as util
@@ -171,10 +171,11 @@ class TestHTTPS(HTTPSDummyServerTestCase):
         try:
             https_pool.request('GET', '/')
             self.fail("Didn't raise SSL invalid common name")
-        except SSLError as e:
+        except MaxRetryError as e:
+            self.assertIsInstance(e.reason, SSLError)
             self.assertTrue(
-                "doesn't match" in str(e) or
-                "certificate verify failed" in str(e)
+                "doesn't match" in str(e.reason) or
+                "certificate verify failed" in str(e.reason)
             )
 
     def test_verified_with_bad_ca_certs(self):
@@ -186,10 +187,11 @@ class TestHTTPS(HTTPSDummyServerTestCase):
         try:
             https_pool.request('GET', '/')
             self.fail("Didn't raise SSL error with bad CA certs")
-        except SSLError as e:
-            self.assertTrue('certificate verify failed' in str(e),
+        except MaxRetryError as e:
+            self.assertIsInstance(e.reason, SSLError)
+            self.assertTrue('certificate verify failed' in str(e.reason),
                             "Expected 'certificate verify failed',"
-                            "instead got: %r" % e)
+                            "instead got: %r" % e.reason)
 
     def test_verified_without_ca_certs(self):
         # default is cert_reqs=None which is ssl.CERT_NONE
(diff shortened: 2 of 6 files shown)


Lukasa:
> I didn't add an explicit test, since the new behaviour is tested implicitly in a lot of existing tests. Is that ok?

I think I'd like an explicit test that proves that there is a second connection occurring. This would probably have to be done down at the `test_socketlevel` level.

piquadrat:
@Lukasa done in #6dee0b6 (sort of). I mostly cargo-culted the test by re-using parts from other tests in the same TestCase class and from `test_retry.py`.

test/with_dummyserver/test_https.py
```diff
@@ -36,7 +36,7 @@
     InsecureRequestWarning,
     SystemTimeWarning,
     InsecurePlatformWarning,
-)
+    MaxRetryError)
```
Lukasa: Can you keep the parenthesis style the same?

test/with_dummyserver/test_https.py
```diff
@@ -330,17 +335,22 @@ def test_assert_invalid_fingerprint(self):
         https_pool.assert_fingerprint = 'AA:AA:AA:AA:AA:AAAA:AA:AAAA:AA:' \
                                         'AA:AA:AA:AA:AA:AA:AA:AA:AA'
 
-        self.assertRaises(SSLError, https_pool.request, 'GET', '/')
+        def _test_request():
```
Lukasa: I don't love closing over the `https_pool`: can we make it an argument instead?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1010,14 +1012,64 @@ def request():
...
+    def test_retry_ssl_error(self):
+        def socket_handler(listener):
+            # first request, trigger an SSLError
+            sock = listener.accept()[0]
+            sock2 = sock.dup()
+            ssl_sock = ssl.wrap_socket(sock,
+                                       server_side=True,
+                                       keyfile=DEFAULT_CERTS['keyfile'],
+                                       certfile=DEFAULT_CERTS['certfile'],
+                                       ca_certs=DEFAULT_CA)
```
Lukasa: We probably don't need the CA certs

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1010,14 +1012,64 @@ def request():
...
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += ssl_sock.recv(65536)
+
+            # Deliberately send from the non-SSL socket to trigger an SSLError
+            sock2.send((
+                           'HTTP/1.1 200 OK\r\n'
+                           'Content-Type: text/plain\r\n'
+                           'Content-Length: 4\r\n'
+                           '\r\n'
+                           'Fail').encode('utf-8'))
```
Lukasa: The indentation here is a bit odd...

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1010,14 +1012,64 @@ def request():
...
+            sock2.close()
+            ssl_sock.close()
+
+            # retried request
+            sock = listener.accept()[0]
+            ssl_sock = ssl.wrap_socket(sock,
+                                       server_side=True,
+                                       keyfile=DEFAULT_CERTS['keyfile'],
+                                       certfile=DEFAULT_CERTS['certfile'],
+                                       ca_certs=DEFAULT_CA)
```
Lukasa: Same note about the CA certs.

piquadrat:
Thanks for the comments @Lukasa! All of them should be addressed in d443b37 :)

piquadrat:
Done. I kept it as short as possible to stay in style with the other change logs, hope that's ok.

Lukasa:
Sorry, it looks like this slipped off my plate and I lost track of it. Do you mind resolving the conflicts that got introduced while this languished?

beniwohli:
No reason to be sorry at all, hope PyCon was a blast :)

I rebased the branch on top of master. Not sure why Github thought the merge commit was necessary. Are you squashing the commits before merging? Otherwise I'll try to clean up the commits a bit before merge.

CHANGES.rst
```diff
@@ -6,6 +6,9 @@ dev (master)
 
 * ... [Short description of non-trivial change.] (Issue #)
 
+* Made the connection pool retry on ``SSLError``.  The original ``SSLError``
+  is available on ``MaxRetryError.reason``. (Issue #1112)
+
```
Lukasa: This has ended up in the wrong place: it should be above the placeholder, not below.

beniwohli:
Ordering in `CHANGES.rst` is fixed and commits are cleaned up :)

PR #1181, opened 2017-05-26
fix 1 lgtm.com alert

jhelie:
Hi,
Just wanted to quickly fix this alert flagged up on lgtm.com.

`UnicodeError` is a subclass of `ValueError` so the ordering of the except blocks would have never allowed to reach it (arguably it had limited impact since the same action is to be taken in both blocks).

Hope this helps!

A total of 14 alerts have been flagged up by lgtm.com, you can enable pull request integration for fully automated PR reviews that will flag these in the future so that they don't get past code review.
https://lgtm.com/projects/g/shazow/urllib3

### DIFF
--- a/urllib3/packages/ssl_match_hostname/_implementation.py
+++ b/urllib3/packages/ssl_match_hostname/_implementation.py
@@ -108,13 +108,13 @@ def match_hostname(cert, hostname):
     try:
         # Divergence from upstream: ipaddress can't handle byte str
         host_ip = ipaddress.ip_address(_to_unicode(hostname))
-    except ValueError:
-        # Not an IP address (common case)
-        host_ip = None
     except UnicodeError:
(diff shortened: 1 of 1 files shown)


SethMichaelLarson:
If `UnicodeError` is the same as `ValueError` and both `except` blocks do the same thing is there any reason to not just remove the `except UnicodeError` block?

jhelie:
none that I can see. At the moment just deleting the `UnicodeError` would result in exactly the same behaviour.

Lukasa:
Thanks, but this is a backport vendored with minimal changes. That means it'd be better to propose this fix upstream: otherwise we'll just lose it when we vendor an updated version.

jhelie:
@Lukasa apologies I hadn't realised it came from https://bitbucket.org/brandon/backports.ssl_match_hostname !

Lukasa:
No problem!

PR #1187, opened 2017-05-29
Swap out nose for pytest in the test runner

alexwlchan:
The smallest possible step in the direction of #1160. All this does is start running the tests under pytest instead of nose – I was then thinking of doing small patches to gradually remove the nose-specific bits, so that dependency can eventually be removed.

(I have seen #1163, but (1) it appears to be a bit stalled, and (2) I usually prefer smaller patches, and getting the test runner out of the way paves the path for more small patches later.)

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,4 +9,5 @@ tornado==4.2.1
 PySocks==1.5.6
 pkginfo>=1.0,!=1.3.0
 psutil==4.3.1
+pytest-cov==2.5.1
 pytest==3.1.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -9,3 +9,4 @@ tornado==4.2.1
 PySocks==1.5.6
 pkginfo>=1.0,!=1.3.0
 psutil==4.3.1
+pytest==3.1.0
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ setup(name='urllib3',
       tests_require=[
           # These are a less-specific subset of dev-requirements.txt, for the
           # convenience of distro package maintainers.
+          'pytest',
           'nose',
           'mock',
           'tornado',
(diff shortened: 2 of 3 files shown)


sigmavirus24:
As this has not stalled and it is a small, easily consumable patch, I'm wholeheartedly 👍 on this. It does look as though we might need a bit more for this to be useful though. The `gae` environment, for example, is still using `nosetests`. Is there a chance you can also convert that here?

Lukasa:
@SethMichaelLarson Happy to build your patch on top of this work?

SethMichaelLarson:
Of course, biggest staller (as I suspected it would be) is GAE because there's nothing that does GAE for pytest like there is for nose.

alexwlchan:
Ah, this is a small patch, but a bit _too_ small – I don’t think it’s creating any coverage data for codecov. Let me have a quick fiddle.

Lukasa:
Come on Travis, do better.

Lukasa:
So...what's the plan about GAE?

alexwlchan:
> what's the plan about GAE?

I don’t know much about GAE or its interactions with pytest, tbh. I’m not even really sure what Nose-GAE provides yet.

My preferred approach would be merging this patch now (which opens the door to removing nose from the rest of the test suite), and GAE being a separate patch – as we see in the other patch, that seems to be a bit trickier.

Lukasa:
I think that's a reasonable note. As seems to have been my theme this weekend I'm going to flag @jonparrott so he knows that this is something we need to keep an eye on, but I think moving forward with this is a good idea.

jonparrott:
I'm +1 on dropping nose in favor of pytest. The same `gcp-devrel-py-tools` that I used to fix the GAE builds has the equivalent of nosegae for pytest in it, so I can handle making the app engine tox environment work with pytest. I filed #1191 for myself. I'll take care of it during my block of hours on Friday for OSS. :)

PR #1188, opened 2017-05-29
Improve SNI control

vtemian:
Hello!

Thank you for your work and dedication!

This patch should allow clients like `requests` to control the hostname used for TLS connections.
My use-case is not that complicated. I want to use the `Host` header as hostname, via `requests`.

### DIFF
--- a/test/test_filepost.py
+++ b/test/test_filepost.py
@@ -1,4 +1,4 @@
-import unittest
+import pytest
 
 from urllib3.filepost import encode_multipart_formdata, iter_fields
 from urllib3.fields import RequestField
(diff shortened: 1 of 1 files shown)


Lukasa:
Thanks for opening this patch!

While this is a worthy use-case to support, I wonder whether we shouldn't re-evaluate how this is supported. Ultimately what this boils down to is a desire to override the DNS resolution process: that is, to make a hostname resolve to an IP address other than the one that DNS would resolve it to (which includes potentially overriding an NXDOMAIN result).

For this reason, I wonder if we should look more to curl for inspiration. curl's command-line API has this flag:

```
--resolve <host:port:address>
      Provide  a  custom  address  for  a specific host and port pair.
      Using this, you can make the curl requests(s)  use  a  specified
      address  and  prevent the otherwise normally resolved address to
      be used. Consider it a sort of /etc/hosts  alternative  provided
      on  the  command line. The port number should be the number used
      for the specific protocol the host will be used  for.  It  means
      you  need several entries if you want to provide address for the
      same host but different ports.
```

I wonder if we should add some kind of way to provide this kind of mapping, instead. This would be set at a fairly high level (probably the `PoolManager`) and used to override DNS when needed. This is a substantially more flexible tool than just the SNI change provided here, while also resolving this use-case neatly.

What do we think?

vtemian:
I'll take a close look at the current codebase and try to come with a working patch.
Your idea is certainly more elegant :+1:

PR #1189, opened 2017-05-30
Add cleanup code to the dummyserver tests

alexwlchan:
Running the tests locally was throwing up a lot of warnings of the form:

    ResourceWarning: unclosed <ssl.SSLSocket fd=16,
        family=AddressFamily.AF_INET6, type=SocketKind.SOCK_STREAM, proto=6,
        laddr=('::1', 61145, 0, 0), raddr=('::1', 61134, 0, 0)>

This patch tries to reduce the number of `ResourceWarning`s that are emitted by the `with_dummyserver` tests. I get 37 less warnings on this branch compared to the current master (c82139e).

Longer term, it might be nice to switch to using `HTTPConnectionPool` as a context manager, which also reduces the specific ties to unittest, a la #1160 – although I’m less sure what I‘d do about the `HTTPConnection` objects. (Context manager omitted from this patch because the extra indentation makes the diff significantly messier.)

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -181,12 +181,14 @@ class TestConnectionPoolTimeouts(SocketDummyServerTestCase):
         pool = HTTPConnectionPool(host, port, timeout=timeout)
         self.addCleanup(pool.close)
         conn = pool._get_conn()
+        self.addCleanup(conn.close)
         self.assertRaises(ConnectTimeoutError, pool._make_request, conn, 'GET', '/')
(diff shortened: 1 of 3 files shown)


alexwlchan:
There are still a few `ResourceWarning`s being thrown by these tests – I think the `cls.server_thread` isn’t being cleaned up correctly in `SocketDummyServerTestCase` (the warning seems to depend on which test was the last to run) – I haven‘t had time to dig into that properly.

alexwlchan:
(Related to #1100, I think.)

PR #1190, opened 2017-05-30
Remove dead branch from TestHTTPS.test_ca_dir_verified

### LINKED ISSUE #1164: TestHTTPS.test_ca_dir_verified skipped on Python 2.7.8 or older
[`TestHTTPS.test_ca_dir_verified`](https://github.com/shazow/urllib3/blob/master/test/with_dummyserver/test_https.py#L150) has the `@onlyPy279OrNewer` decorator to skip on Python 2.7.8 or older. However the test contains logic which checks the Python version and then does different verifications based on whether `sys.version` is greater or equal to `(2, 7, 9)`. See below:

```python
@onlyPy279OrNewer
@notSecureTransport
def test_ca_dir_verified(self):

    # ...

    if sys.version_info >= (2, 7, 9):
        self.assertFalse(warn.called, warn.call_args_list)
    else:  # <--- This branch is never taken?
        self.assertTrue(warn.called)
        call, = warn.call_args_list
        error = call[0][1]
        self.assertEqual(error, InsecurePlatformWarning)
```

Was it intentional to keep this branch?

The decorator for the test was added in commit https://github.com/shazow/urllib3/commit/e15ea73b23574b6f7ce5867d4f8d25aed2718b3f which came shortly after the test was created.

alexwlchan:
Resolves #1164.

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -160,14 +160,7 @@ class TestHTTPS(HTTPSDummyServerTestCase):
         with mock.patch('warnings.warn') as warn:
             r = https_pool.request('GET', '/')
             self.assertEqual(r.status, 200)
-
-            if sys.version_info >= (2, 7, 9):
-                self.assertFalse(warn.called, warn.call_args_list)
-            else:
-                self.assertTrue(warn.called)
-                call, = warn.call_args_list
-                error = call[0][1]
-                self.assertEqual(error, InsecurePlatformWarning)
+            self.assertFalse(warn.called, warn.call_args_list)
 
     def test_invalid_common_name(self):
         https_pool = HTTPSConnectionPool('127.0.0.1', self.port,


alexwlchan:
So the `@onlyPy279OrNewer` decorator should be removed?

Lukasa:
This is a tricky situation. Basically, the branch is dead in the sense that the ability to specify a CA directory does not exist prior to 2.7.9, so the whole test cannot run without that feature backported. While it's possible that a distro has backported the feature, I personally doubt it.

nateprewitt:
I think what @alexwlchan is pointing out is this else branch is unreachable, no matter the python branch, due to the decorator. This test is only executed on Python 2.7.9+.

Lukasa:
@nateprewitt That's not *entirely* accurate: it's possible that downstream repackagers are doing something wacky. But as I said, I doubt it. :wink:

alexwlchan:
Rebased against master, and I’ve amended the commit message to make it clearer why I think this code is dead.

PR #1192, opened 2017-05-30
Rewrite test_response.py and test_retry.py to be pytest-style

alexwlchan:
Making some incremental progress on #1160. Such lovely and clean unittest-free code!

And as a bonus, you no longer get `ResourceWarning`s emitted from `test_response.py` if all the tests succeed.

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -181,7 +181,8 @@ class TestResponse(unittest.TestCase):
 
         # Try closing with an `httplib.HTTPResponse`, because it has an
         # `isclosed` method.
-        hlr = httplib.HTTPResponse(socket.socket())
+        sock = socket.socket()
+        hlr = httplib.HTTPResponse(sock)
         resp2 = HTTPResponse(hlr, preload_content=False)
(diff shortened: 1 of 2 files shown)


test/test_response.py
```diff
@@ -156,79 +161,82 @@ def test_chunked_decoding_gzip(self):
...
-        self.assertEqual(resp3.closed, True)
-        self.assertRaises(IOError, resp3.fileno)
+        assert resp3.closed
+        with pytest.raises(IOError):
+            resp3.fileno()
 
-    def test_io_closed_consistently(self):
-        hlr = httplib.HTTPResponse(socket.socket())
+    def test_io_closed_consistently(self, sock):
+        hlr = httplib.HTTPResponse(sock)
```
alexwlchan: This `socket.socket()` object was never closed, so the test would have a `ResourceWarning`. I’ve switched to a fixture so it always gets cleaned up properly.

alexwlchan:
I’m going to close this and break it out as two separate patches – one for the pytest rewrite, one for the leaked sockets.

PR #1199, opened 2017-06-03
Rewrite test_compatibility.py to be pytest-style

alexwlchan:
Related to #1160.

### DIFF
--- a/test/test_compatibility.py
+++ b/test/test_compatibility.py
@@ -1,10 +1,11 @@
-import unittest
 import warnings
+
+import pytest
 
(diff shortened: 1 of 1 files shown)


Lukasa:
I wonder what codecov is up to.

Lukasa:
This just seems stuck. It can't have affected coverage I don't think, so I'm gonna merge it.

Lukasa:
Ah, crap, I can't. Sorry @alexwlchan, but can you open a new PR? I think we broke this one. 🙁

PR #1203, opened 2017-06-03
Rewrite test_compatibility.py to be pytest-style

alexwlchan:
Attempting to get #1199 to talk to coverage.

### DIFF
--- a/test/test_compatibility.py
+++ b/test/test_compatibility.py
@@ -1,10 +1,11 @@
-import unittest
 import warnings
+
+import pytest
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
It doesn't seem this is enough.

alexwlchan:
I’m a little bit flummoxed. This test class isn’t imported or used anywhere else, and I don’t get any warnings about leaked FDs. Not sure how the two could possibly be related. I’ll rebase against master and give it another go.

Lukasa:
Goddamn it appveyor.

PR #1204, opened 2017-06-05
Rewrite test_exceptions.py to be pytest-style

alexwlchan:
Related to #1160.

### DIFF
--- a/test/test_exceptions.py
+++ b/test/test_exceptions.py
@@ -26,7 +26,8 @@ class TestPickle(object):
         ReadTimeoutError(HTTPConnectionPool('localhost'), '/', None),
     ])
     def test_exceptions(self, exception):
-        assert pickle.loads(pickle.dumps(exception))
+        result = pickle.loads(pickle.dumps(exception))
+        assert isinstance(result, type(exception))
 
 
 class TestFormat(object):
--- a/test/test_exceptions.py
+++ b/test/test_exceptions.py
@@ -1,5 +1,6 @@
-import unittest
 import pickle
+
+import pytest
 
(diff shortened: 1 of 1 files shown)


test/test_exceptions.py
```diff
@@ -1,53 +1,37 @@
...
+        HTTPError('foo', IOError('foo')),
+        MaxRetryError(HTTPConnectionPool('localhost'), '/', None),
+        LocationParseError('fake location'),
+        ClosedPoolError(HTTPConnectionPool('localhost'), None),
+        EmptyPoolError(HTTPConnectionPool('localhost'), None),
+        HostChangedError(HTTPConnectionPool('localhost'), '/', None),
+        ReadTimeoutError(HTTPConnectionPool('localhost'), '/', None),
+    ])
+    def test_exceptions(self, exception):
+        assert pickle.loads(pickle.dumps(exception))
```
Lukasa: Looking at this, one wonders whether there was supposed to be a more concrete assertion here other than just "does not return `None`".

alexwlchan: Perhaps? I tried changing it to

```python
assert pickle.loads(pickle.dumps(exception)) == exception
```

but that turns out to be False. Maybe that’s indicative of a bug?

Lukasa: /me shrugs.

Ultimately I have no idea.

Lukasa: I doubt equality is implemented for these exceptions though, so probably it doesn't matter.

alexwlchan: Yeah, looks like they don’t have equality helpers. I did make it the test a little stronger anyway, but I don’t think there’s more that can be easily checked beyond that.

PR #1207, opened 2017-06-06
Rewrite test_no_ssl.py to be pytest-style

alexwlchan:
Part of #1160.

### DIFF
--- a/test/test_no_ssl.py
+++ b/test/test_no_ssl.py
@@ -6,10 +6,8 @@ Test what happens if Python was built without SSL
 """
 
 import sys
-if sys.version_info >= (2, 7):
-    import unittest
-else:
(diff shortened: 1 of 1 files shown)


test/test_no_ssl.py
```diff
@@ -65,28 +63,22 @@ def pop(self):
 module_stash = ModuleStash('urllib3')
 
 
-class TestWithoutSSL(unittest.TestCase):
```
Lukasa: Surprise this was imported elsewhere in the tests! Can't change the name without also changing those other tests.

alexwlchan: Blergh, and one of those is in the AppEngine tests. I’ll leave this for now.

PR #1208, opened 2017-06-06
Rewrite test_poolmanager.py to be pytest-style

alexwlchan:
Part of #1160.

Note: I took out all the cleanup code because I don't get any `ResourceWarning`s without it when running locally. I'd like to double-check that CI doesn't think differently before this is merged.

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,5 +1,6 @@
 import socket
-import sys
+
+import pytest
 
 from urllib3.poolmanager import (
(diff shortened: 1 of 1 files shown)


alexwlchan:
> The CI does. ;)

Can you give me a pointer please? I can see a bunch of FD leaks in `test/with_dummyserver/test_[proxy_]poolmanager.py`, but that’s unrelated to this test. I can’t see any leaks coming from this file specifically.

PR #1211, opened 2017-06-07
OSError does not take keyword arguments

rhettlunn:
This PR ensures that the intended OSError is raised instead of a TypeError.

### DIFF
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -112,7 +112,7 @@ else:
                     if expires is not None:
                         current_time = monotonic()
                         if current_time > expires:
-                            raise OSError(errno=errno.ETIMEDOUT)
+                            raise OSError(errno.ETIMEDOUT)
                         if recalc_timeout:
                             if "timeout" in kwargs:
                                 kwargs["timeout"] = expires - current_time
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -72,24 +72,13 @@ else:
         """ Wrapper function for syscalls that could fail due to EINTR.
         All functions should be retried if there is time left in the timeout
         in accordance with PEP 475. """
-        # FIXME: We are not checking (testing) that select call args
-        #        are being passed to us as (rlist, wlist, xlist[, timeout])
-        #        Currently we're sometimes being passed (rlist, wlist, timeout)
-        #        (Other times it's being passed in kwargs, which should work fine)
 
-        # timeout may be positional or a keyword argument
-        if "timeout" in kwargs:
-            timeout = kwargs["timeout"]
-        elif len(args) > 2:
-            # FIXME: Is this true for the other syscalls we wrap?
-            #        Answer: no - select is the only one with a timeout
-            #        (poll, epoll, and kqueue do not have timeouts)
-            #        therefore our true test should be to see if the func
-            #        we're being passed is select.select, and only then
-            #        determine positionally the timeout
-            timeout = args[2]
-        else:
-            timeout = None
+        if recalc_timeout and "timeout" not in kwargs:
+            raise ValueError(
+                "Timeout must be in kwargs to be recalculated")
+
+        # timeout for recalcultion must be a keyword argument
+        timeout = kwargs.get("timeout", None)
 
         # Based on the timeout, determine call expiry time
         if timeout is None:
@@ -103,9 +92,6 @@ else:
                 expires = monotonic() + timeout
 
         args = list(args)
-        if recalc_timeout and timeout is None:
-            raise ValueError(
-                "Timeout must be in args or kwargs to be recalculated")
 
         result = _SYSCALL_SENTINEL
         while result is _SYSCALL_SENTINEL:
@@ -133,13 +119,7 @@ else:
                         if current_time > expires:
                             raise OSError(errno.ETIMEDOUT, "Connection timed out")
                         if recalc_timeout:
-                            # FIXME: we're now not updating the timeout argument in kwargs
-                            #        (We never updated it in args)
-                            #        Note that we're not currently testing to see if
-                            #        a passed-in timeout is reaching the select.select
-                            #        (I strongly suspect it is not in some cases)
-                            if timeout is not None:
-                                timeout = expires - current_time
+                            kwargs["timeout"] = expires - current_time
                     continue
                 if errcode:
                     raise SelectorError(errcode)
@@ -341,7 +321,7 @@ if hasattr(select, "select"):
             timeout = None if timeout is None else max(timeout, 0.0)
             ready = []
             r, w, _ = _syscall_wrapper(self._select, True, self._readers,
-                                       self._writers, timeout)
+                                       self._writers, timeout=timeout)
             r = set(r)
             w = set(w)
             for fd in r | w:
@@ -534,7 +514,7 @@ if hasattr(select, "kqueue"):
             ready_fds = {}
 
             kevent_list = _syscall_wrapper(self._kqueue.control, True,
-                                           None, max_events, timeout)
+                                           None, max_events, timeout=timeout)
 
             for kevent in kevent_list:
                 fd = kevent.ident
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -91,8 +91,6 @@ else:
             else:
                 expires = monotonic() + timeout
 
-        args = list(args)
-
         result = _SYSCALL_SENTINEL
         while result is _SYSCALL_SENTINEL:
             try:
--- a/urllib3/util/selectors.py
+++ b/urllib3/util/selectors.py
@@ -72,7 +72,26 @@ else:
         """ Wrapper function for syscalls that could fail due to EINTR.
         All functions should be retried if there is time left in the timeout
         in accordance with PEP 475. """
-        timeout = kwargs.get("timeout", None)
+        # FIXME: We are not checking (testing) that select call args
+        #        are being passed to us as (rlist, wlist, xlist[, timeout])
+        #        Currently we're sometimes being passed (rlist, wlist, timeout)
+        #        (Other times it's being passed in kwargs, which should work fine)
+
+        # timeout may be positional or a keyword argument
+        if "timeout" in kwargs:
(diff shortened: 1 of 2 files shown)


SethMichaelLarson:
None off the top of my head. I'm scouring the tests right now.

SethMichaelLarson:
[Looks like codecov isn't registering this line as required](https://codecov.io/gh/shazow/urllib3/src/master/urllib3/util/selectors.py#L115)?

SethMichaelLarson:
Looks like not a single `raise` statement is receiving coverage. If you look at the PR referencing this one's [coverage in codecov](https://codecov.io/gh/shazow/urllib3/pull/1212/changes) you can see that there are a lot of `raise` statements aren't covered, not just in selectors.

Lukasa:
Hrm, odd. In that case, can we aim to get a test for this behaviour?

SethMichaelLarson:
Don't worry about removing `.coveragerc` in this PR @rhettlunn, that'll be handled in another PR.

Since this is kind of a tough space to get test coverage I would suggest creating a mock test in the "SpecialSelectorScenerios" test case and mock the low-level `select()` function to wait for a long time (to ensure a timeout) and then raise an `OSError` with a `errno.EINTR` and assert that `OSError(errno.ETIMEOUT)` is raised from the high-level `SelectSelector.select()` call.

Does that make sense?

rhettlunn:
I've written the new test and it's passing (I put it into TestUniqueSelectScenarios).

I also needed to make another change to selectors.py - _syscall_wrapper wasn't able to properly recalculate the timeout when it wasn't passed as a keyword argument (and it wasn't raising on that before, though it is now).

I'm pretty new to Github, so give me a few minutes to figure out how to update the PR with the new code.

rhettlunn:
Aha! I was wondering if this would be an issue - The Python 3.5+ version of _syscall_wrapper catches all (OSError, IOError, select.error) and reraises as SelectorError, but the lower versions can raise OSError, or whatever error it received (if it's not an interrupt and doesn't have a truthy errcode).

I don't know if that's the behaviour we want, so I'll wait for feedback.

Lukasa:
I'm hopig @SethMichaelLarson will have an insight here.

SethMichaelLarson:
Back from vacation, I'm looking at the diffs right now. Thanks for the work on this, @rhettlunn!

urllib3/util/selectors.py
```diff
@@ -133,13 +119,7 @@ def _syscall_wrapper(func, recalc_timeout, *args, **kwargs):
...
                             raise OSError(errno.ETIMEDOUT, "Connection timed out")
                         if recalc_timeout:
-                            # FIXME: we're now not updating the timeout argument in kwargs
-                            #        (We never updated it in args)
-                            #        Note that we're not currently testing to see if
-                            #        a passed-in timeout is reaching the select.select
-                            #        (I strongly suspect it is not in some cases)
-                            if timeout is not None:
-                                timeout = expires - current_time
+                            kwargs["timeout"] = expires - current_time
```
SethMichaelLarson: This should probably be `timeout = expires - current_time` so the value is updated to be used in the next attempt at the syscall.

urllib3/util/selectors.py
```diff
@@ -99,7 +104,7 @@ def _syscall_wrapper(func, recalc_timeout, *args, **kwargs):
             except (OSError, IOError, select.error) as e:
                 # select.error wasn't a subclass of OSError in the past.
                 errcode = None
-                if hasattr(e, "errno"):
+                if hasattr(e, "errno") and e.errno is not None:
```
SethMichaelLarson: Curious what scenarios `e.errno` is `None`?

SethMichaelLarson:
So I've actually changed how this is handled in the latest release of the module `selectors2` and begun just raising whatever error we get unwrapped as long as it's not `errno.EINTR` or `errno.WSAEINTR`.
I can't actually remember why I decided to wrap the exception. Because it's low-level and won't be seen by any interface except for our own I don't think it really matters? I'm in favor of not using `SelectorError` and just raising whatever we get if it's not an interrupt. If @Lukasa agrees we can change that part of the code.

Secondly, about the `**kwargs` nonsense with getting the timeout value. You're completely right, I just missed using `kwargs` in those two uses of system calls.

urllib3/util/selectors.py
```diff
@@ -112,10 +117,9 @@ def _syscall_wrapper(func, recalc_timeout, *args, **kwargs):
                     if expires is not None:
                         current_time = monotonic()
                         if current_time > expires:
-                            raise OSError(errno=errno.ETIMEDOUT)
+                            raise OSError(errno.ETIMEDOUT, "Connection timed out")
                         if recalc_timeout:
-                            if "timeout" in kwargs:
-                                kwargs["timeout"] = expires - current_time
+                            kwargs["timeout"] = expires - current_time
```
SethMichaelLarson: This should be `timeout = expires - current_time`?

urllib3/util/selectors.py
rhettlunn: e.errno is None if OSError or IOError are raised without a second parameter. I don't know if that is done by anything we're calling, but this way if it is we'll fallback to e.args[0]

urllib3/util/selectors.py
rhettlunn: We need to update the timeout that we're passing to func, so the the underlying call gets the updated timeout value.

SethMichaelLarson: That's true, I'm wondering how this wasn't caught by testing. I know timeout testing is disabled on CI but I should have seen this locally. Might have to write more tests.

urllib3/util/selectors.py
```diff
@@ -84,9 +92,6 @@ def _syscall_wrapper(func, recalc_timeout, *args, **kwargs):
                 expires = monotonic() + timeout
 
         args = list(args)
```
SethMichaelLarson: This can probably be removed? Not sure why this is even here.

rhettlunn: Agreed.

urllib3/util/selectors.py
rhettlunn: The old code was covering that (there's an intermediate commit in my fork that I was going to post for context and ask questions, but then figured out those answers myself).

All I changed at that snippet was to remove the check for `"timeout" in kwargs`, as that is now covered by the updated `recalc_timeout and "timeout" not in kwargs` check.

SethMichaelLarson: I'm trying to write a test locally right now to reproduce this issue, I'll reply back with my results.

SethMichaelLarson:
Alright I wrote a test that should be added to this PR. Add it under the unique scenarios test case. It fails with the latest master but shouldn't fail with your changes. You'll also have to import `mock` at the top of `test_selectors.py`.

```python
    def test_timeout_is_recalculated_after_interrupt(self):
        selectors._DEFAULT_SELECTOR = None

        mock_socket = mock.Mock()
        mock_socket.fileno.return_value = 1

        class InterruptingSelect(object):
            """ Helper object that imitates a select that interrupts
            after sleeping some time then returns a result. """
            def __init__(self):
                self.call_count = 0
                self.calls = []

            def select(self, *args, **kwargs):
                self.calls.append((args, kwargs))
                self.call_count += 1
                if self.call_count == 1:
                    time.sleep(0.1)
                    error = OSError()
                    error.errno = errno.EINTR
                    raise error
                else:
                    return [1], [], []

        mock_select = InterruptingSelect()

        patch_select_module(self, select=mock_select.select)

        selector = self.make_selector()
        selector.register(mock_socket, selectors.EVENT_READ)

        result = selector.select(timeout=1.0)

        # Make sure the mocked call actually completed correctly.
        self.assertEqual(len(result), 1)
        self.assertEqual(result[0][0].fileobj, mock_socket)
        self.assertEqual(result[0][1], selectors.EVENT_READ)

        # There should be two calls to the mock_select.select() function
        self.assertEqual(mock_select.call_count, 2)

        # Timeout should be less in the second call.
        # The structure of mock_select.calls is [(args, kwargs), (args, kwargs)] where
        # args is ([r], [w], [x], timeout).
        self.assertLess(mock_select.calls[1][0][3], mock_select.calls[0][0][3])
```

urllib3/util/selectors.py
SethMichaelLarson: I could reproduce, see the comment below with the new test.

rhettlunn:
Your new test is passing on 2.7, but (of course) it's currently failing on 3.5 with:
```
  File "E:\urllib3-master\urllib3\util\selectors.py", line 64, in _syscall_wrapper
    return func(*args, **kwargs)
  File "E:\urllib3-master\urllib3\util\selectors.py", line 312, in _select
    return select.select(r, w, [], timeout)
  File "E:\urllib3-master\test\test_selectors.py", line 760, in select
    raise error
OSError

  File "E:\urllib3-master\test\test_selectors.py", line 771, in test_timeout_is_recalculated_after_interrupt
    result = selector.select(timeout=1.0)
  File "E:\urllib3-master\urllib3\util\selectors.py", line 322, in select
    self._writers, timeout=timeout)
  File "E:\urllib3-master\urllib3\util\selectors.py", line 69, in _syscall_wrapper
    raise SelectorError(errcode)
urllib3.util.selectors.SelectorError: <SelectorError errno=4>
```
...due to us not catching/retrying EINTR on 3.5

I've added an exclusion to that test like to the other new one, unless you think we should do something more complex.

SethMichaelLarson:
Yep that's good, I think just skipping in Python 3.5+ is fine.

urllib3/util/selectors.py
```diff
@@ -510,7 +512,7 @@ def select(self, timeout=None):
             ready_fds = {}
 
             kevent_list = _syscall_wrapper(self._kqueue.control, True,
```
SethMichaelLarson: So we're going to have to make a wrapper for `select.kqueue.control` as it [doesn't take key-word arguments](https://travis-ci.org/shazow/urllib3/jobs/244134449#L1750) and `timeout` is always passed in as a key-word argument in the `_syscall_wrapper`. Look at the `SelectSelector` for an example of this.

SethMichaelLarson:
This is fixed in #1343.

PR #1212, opened 2017-06-07
Test removing raise from .coveragerc

SethMichaelLarson:
Testing a theory in #1211


Lukasa:
Might be good to try to get coverage on these if we can. This may be a worthy target for a contributor who feels like making a tangible improvement to the codebase.

SethMichaelLarson:
Test confirmed, closing this.

PR #1215, opened 2017-06-09
Rewrite test_utils.py to be pytest-style

alexwlchan:
Related to #1160.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -2,12 +2,12 @@ import hashlib
 import warnings
 import logging
 import io
-import unittest
 import ssl
 import socket
 from itertools import chain
 
 from mock import patch, Mock
+import pytest
 
 from urllib3 import add_stderr_logger, disable_warnings
 from urllib3.util.request import make_headers, rewind_body, _FAILEDTELL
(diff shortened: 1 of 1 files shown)


test/test_util.py
```diff
@@ -48,106 +48,96 @@
...
-                'http+unix', '%2fvar%2frun%2fSOCKET', None),
-        }
-        for url, expected_host in url_host_map.items():
-            returned_host = get_host(url)
-            self.assertEqual(returned_host, expected_host)
-
-    def test_parse_url_normalization(self):
+class TestUtil(object):
+
+    url_host_map = {
```
Lukasa: It doesn't seem like this needs to be a dictionary, and the dictionary nature of it makes it really easy for us to accidentally miss a test.

test/test_util.py
```diff
@@ -48,106 +48,96 @@
...
-            'http://::1/',
-            'http://::1:80/',
-            'http://google.com:-80',
-            six.u('http://google.com:\xb2\xb2'),  # \xb2 = ^2
-        ]
-
-        for location in invalid_host:
-            self.assertRaises(LocationParseError, get_host, location)
-
-    def test_host_normalization(self):
```
Lukasa: What happened to these tests?

test/test_util.py
```diff
@@ -191,141 +181,133 @@ def test_parse_url_normalization(self):
...
-        for url, expected_Url in chain(self.parse_url_host_map,
-                                       self.non_round_tripping_parse_url_host_map.items()):
-            returned_Url = parse_url(url)
-            self.assertEqual(returned_Url, expected_Url)
-
-    def test_unparse_url(self):
-        for url, expected_Url in self.parse_url_host_map:
-            self.assertEqual(url, expected_Url.url)
+    @pytest.mark.parametrize('url, expected_Url', parse_url_host_map)
+    def test_unparse_url(self, url, expected_Url):
```
Lukasa: Let's fix up the weird casing here while we're making changes.

test/test_util.py
```diff
@@ -191,141 +181,133 @@ def test_parse_url_normalization(self):
...
+    }
 
-        }
+    @pytest.mark.parametrize(
+        'url, expected_Url',
+        chain(parse_url_host_map, non_round_tripping_parse_url_host_map.items())
+    )
+    def test_parse_url(self, url, expected_Url):
+        returned_Url = parse_url(url)
+        assert returned_Url == expected_Url
```
Lukasa: And here.

test/test_util.py
```diff
@@ -191,141 +181,133 @@ def test_parse_url_normalization(self):
         # Empty Port
         'http://google.com:': Url('http', host='google.com'),
         'http://google.com:/': Url('http', host='google.com', path='/'),
+    }
```
Lukasa: Again, probably no need for a dict here.

test/test_util.py
```diff
@@ -191,141 +181,133 @@ def test_parse_url_normalization(self):
...
+        'http://google.com/mail': '/mail',
+        'http://google.com/mail/': '/mail/',
+        'http://google.com/': '/',
+        'http://google.com': '/',
+        '': '/',
+        '/': '/',
+        '?': '/?',
+        '#': '/',
+        '/foo?bar=baz': '/foo?bar=baz',
+    }
```
Lukasa: Again, probably no need for a dict here.

test/test_util.py
```diff
@@ -191,141 +181,133 @@ def test_parse_url_normalization(self):
...
+    def test_request_uri(self, url, expected_request_uri):
+        returned_url = parse_url(url)
+        assert returned_url.request_uri == expected_request_uri
+
+    url_netloc_map = {
+        'http://google.com/mail': 'google.com',
+        'http://google.com:80/mail': 'google.com:80',
+        'google.com/foobar': 'google.com',
+        'google.com:12345': 'google.com:12345',
+    }
```
Lukasa: And here.

test/test_util.py
alexwlchan: What used to be `test_host_normalization` got rolled into `url_host_map`, as the body of the two tests were identical.

test/test_util.py
alexwlchan: The casing isn’t entirely without merit – it’s looking for instances of the `Url` class.

PR #1216, opened 2017-06-09
Clean up a few more leaked FDs in tests

alexwlchan:
Related to #1100.

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -184,6 +184,7 @@ class TestResponse(unittest.TestCase):
         sock = socket.socket()
         self.addCleanup(sock.close)
         hlr = httplib.HTTPResponse(sock)
+        self.addCleanup(hlr.close)
         resp2 = HTTPResponse(hlr, preload_content=False)
         self.assertEqual(resp2.closed, False)
         resp2.close()
@@ -203,6 +204,7 @@ class TestResponse(unittest.TestCase):
         sock = socket.socket()
         self.addCleanup(sock.close)
         hlr = httplib.HTTPResponse(sock)
+        self.addCleanup(hlr.close)
         hlr.fp = BytesIO(b'foo')
         hlr.chunked = 0
         hlr.length = 3
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -181,7 +181,9 @@ class TestResponse(unittest.TestCase):
 
         # Try closing with an `httplib.HTTPResponse`, because it has an
         # `isclosed` method.
-        hlr = httplib.HTTPResponse(socket.socket())
+        sock = socket.socket()
+        self.addCleanup(sock.close)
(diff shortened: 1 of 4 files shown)


test/test_response.py
```diff
@@ -198,7 +200,9 @@ def test_io(self):
         self.assertRaises(IOError, resp3.fileno)
 
     def test_io_closed_consistently(self):
-        hlr = httplib.HTTPResponse(socket.socket())
+        sock = socket.socket()
+        self.addCleanup(sock.close)
```
Lukasa: Closing the response would be better than closing the socket.

alexwlchan: I don’t think that works – if I replace these three lines with

```python
hlr = httplib.HTTPResponse(socket.socket())
self.addCleanup(hlr.close)
```

I get two `ResourceWarning`s when I run the tests. If you look at the [source code for HTTPResponse](https://github.com/python/cpython/blob/a632d00a1cbb6f4f8978cb45dd9789658cceb7c8/Lib/http/client.py#L217), it creates a file object associated with the socket, and the file object gets closed when you call `hlr.close()`, but the socket itself is left open.

Lukasa: That just shouldn't happen: there should be no other references to the socket keeping it alive. =( Any idea where the other reference is?

alexwlchan: I’ve been able to reduce it to the following minimal case of FD leakage:

```python
import socket
import unittest

import http.client as httplib


class TestResponse(unittest.TestCase):
    def test_io_closed_consistently(self):
        hlr = httplib.HTTPResponse(socket.socket())


if __name__ == '__main__':
    unittest.main()
```

If you stick that in `urllib3/test` and run it, you get the FD leakage. If you run it in a clean environment, you don’t – seems like something in the urllib3 repo is causing us to hold on to the socket (which might explain leaks elsewhere).

alexwlchan: Ah, so the urllib3 test suite uses pytest==3.1.0, which shows the warning. The latest version of pytest==3.1.2 doesn’t show the warning, but running under unittest drops the warning on both occasions. 🤔

alexwlchan: Reviewing the [docs on socket programming](https://docs.python.org/3.5/howto/sockets.html#disconnecting):

> Python takes the automatic shutdown a step further, and says that when a socket is garbage collected, it will automatically do a `close` if it’s needed. But relying on this is a very bad habit. If your socket just disappears without doing a `close`, the socket at the other end may hang indefinitely, thinking you’re just being slow. _Please_ `close` your sockets when you’re done.

So maybe have an explicit `close` on both resources? Would that be acceptable @Lukasa?

Lukasa: Yeah, that works

alexwlchan: Okay, done.

PR #1229, opened 2017-07-17
Fix compatibility for cookiejar

ml31415:
py3 cookiejar expects `get_all()` from header objects and `info()` from response objects. This PR adds the necessary compatibility methods and a test for that in `test_compatibility`. `getlist()` needs to get a default argument, to serve as a dropin for the required `get_all()`. I opted for extending `getlist` instead of reinventing the wheel by creating a separate `get_all`, not sure if you agree here.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -13,6 +13,8 @@ dev (master)
 * Drain and release connection before recursing on retry/redirect.  Fixes
   deadlocks with a blocking connectionpool. (Issue #1167)
 
+* Fixed compatibility for cookiejar. (Issue #1229)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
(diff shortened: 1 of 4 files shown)


ml31415:
Just saw #904 which already addresses the same issue amongst some major changes. So this may serve as a temporary fix, until #904 is accepted.

Lukasa:
Is this compatibility for python 3, or for both? My Python 2.7 barfs on `extract_cookies` as well.

ml31415:
For the header dict, py2 was taken care of already with `getheaders` and `iget`, so this part is py3 only. `info` for the response object was already implemented at some point I  guess, not sure where it went, but for me the missing `info` is also causing py2 to fail right now.

ml31415:
The travis fail seems unrelated.

ml31415:
Done.

PR #1230, opened 2017-07-18
Stamp out a bunch of warnings in the tests

alexwlchan:
The test suite emits a lot of warnings (63 on [this recent Travis run](https://travis-ci.org/shazow/urllib3/jobs/254519138)).

These warnings are mostly benign, but they’re mixed in with real problems, such as #1100. This patch causes the test suite to emit substantially fewer warnings, meaning the interesting ones will be easier to spot. It also cleans up another two leaked FDs.

### DIFF
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -1,10 +1,13 @@
 import threading
 import socket
 
+from urllib3 import disable_warnings
 from urllib3.contrib import socks
-from urllib3.exceptions import ConnectTimeoutError, NewConnectionError
+from urllib3.exceptions import (
+    ConnectTimeoutError, InsecureRequestWarning, NewConnectionError,
+)
 
-from dummyserver.server import DEFAULT_CERTS
+from dummyserver.server import DEFAULT_CERTS, NoIPv6Warning
 from dummyserver.testcase import IPV4SocketDummyServerTestCase
 
 from nose.plugins.skip import SkipTest
@@ -199,6 +202,8 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
     Test the SOCKS proxy in SOCKS5 mode.
     """
     def test_basic_request(self):
+        disable_warnings(NoIPv6Warning)
+
         def request_handler(listener):
             sock = listener.accept()[0]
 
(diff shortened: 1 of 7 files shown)


alexwlchan:
Just as a point of comparison, the [equivalent Travis run](https://travis-ci.org/shazow/urllib3/jobs/254732776) for this branch emits just 17 warnings, and they all point to leaked FDs.

test/contrib/test_socks.py
```diff
@@ -199,6 +206,8 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
     Test the SOCKS proxy in SOCKS5 mode.
     """
     def test_basic_request(self):
+        disable_warnings(NoIPv6Warning)
```
Lukasa: This repeated warning disabling is potentially a problem. Does py.test restore the warnings filters after every test run? If not, we're just repeatedly disabling the `NoIPv6Warning`.

Lukasa: As a follow-on question: can we just remove this warning entirely? What's the point of having it and always turning it off, given that we raise it from our own test infrastructure anyway.

alexwlchan: py.test restores the default warnings filters after every run. Here’s a minimal example:

```python
import warnings

from urllib3 import disable_warnings
from urllib3.exceptions import InsecureRequestWarning


def test1():
    print('test1 before', len(warnings.filters))
    disable_warnings(InsecureRequestWarning)
    print('test1 after', len(warnings.filters))


def test2():
    print('test2 before', len(warnings.filters))
    disable_warnings(InsecureRequestWarning)
    print('test2 after', len(warnings.filters))
```

which has output:

```
('test1 before', 9)
('test1 after', 10)
('test2 before', 9)
('test2 after', 10)
```

If you print the filters rather than their length, you can see the `InsecureRequestWarning` filter is purged between runs.

As to ignoring them entirely – I wonder if putting them in `conftest.py` would do it? I’ll try that.

alexwlchan: Hmm. Neither `conftest.py`, nor `__init__.py`, nor sticking it at the top-level of the test file do the trick. I could put it in the `setUp()` call? Thoughts, @Lukasa?

Lukasa: Any reason we can't make it an automatic py.test fixture?

jonparrott:

PR #1231, opened 2017-07-19
Use vendored version of ``six`` in pyopenssl.

erikrose:
Fixes import errors like https://github.com/certbot/certbot/issues/4886 when importing `requests`. Everywhere else in `contrib` vendors properly; it looks like we just missed a spot.

If this looks good, could we please have a 1.21.2 release soon so Let's Encrypt doesn't have to work around this in a zillion OS packages? I expect this problem is widespread, since any import of `requests` without a suitable `six` also installed could trigger it. [pip may even be hitting it](https://github.com/certbot/certbot/issues/2902), though I haven't dug into that deeply. Many thanks!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ dev (master)
 
 * Fixed compatibility for cookiejar. (Issue #1229)
 
+* pyopenssl: Use vendored version of ``six``.
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,7 @@ dev (master)
 
 * Fixed compatibility for cookiejar. (Issue #1229)
 
-* pyopenssl: Use vendored version of ``six``.
+* pyopenssl: Use vendored version of ``six``. (Issue #1231)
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -15,6 +15,8 @@ dev (master)
 
 * Fixed compatibility for cookiejar. (Issue #1229)
 
+* pyopenssl: Use vendored version of ``six``.
```
Lukasa: Can you add the PR number here?

erikrose: Gladly!

erikrose:
1.22 is certainly better than nothing. The trouble is that the latest `requests` brings in `'urllib3>=1.21.1,<1.22'`, so it will continue to fail until it, too, is updated.

Lukasa:
Yeah, I'm aware of that problem. Sadly there isn't much we can do about that other than aim to ship a new Requests as well that supports this urllib3 version.

erikrose:
I suppose we'd have had to get the OS packages for urllib3 updated on all our distros anyway. Thanks again for the quick attention!

Lukasa:
Ok I've just shipped urllib3 1.22.

erikrose:
Thank you! :-D

PR #1235, opened 2017-07-21
Release connection on chunked HEAD response.

### LINKED ISSUE #1234: HEAD response with chunked Transfer-Encoding does not release the connection
Amazon S3 sets "Transfer-Encoding: chunked" in the response even for a HEAD request. For example:
```
DEBUG:botocore.vendored.requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): s3.amazonaws.com
send: 'HEAD /timur-test/garbage HTTP/1.1\r\nHost: s3.amazonaws.com\r\nAccept-Encoding: identity\r\n\r\n'
reply: 'HTTP/1.1 301 Moved Permanently\r\n'
header: x-amz-bucket-region: us-west-2
header: x-amz-request-id: 4E83FBED76CBB9BB
header: x-amz-id-2: dxdu+gwK6fdzdK2CaA/0yAcBxO7wdSmzRDwo9JRYA7QeMEk8xMkuE5VJ2AiabWWSqpi+IHweXqA=
header: Content-Type: application/xml
header: Transfer-Encoding: chunked
header: Date: Fri, 21 Jul 2017 06:41:02 GMT
header: Server: AmazonS3
```
This causes the connection to not be returned to the pool after calling `stream()` or `read_chunked()` (which `stream()` is a thin wrapper for in this case). This is due to the following conditional in the `read_chunked()` method: https://github.com/shazow/urllib3/blob/master/urllib3/response.py#L592.

I believe moving this conditional inside the `error_catcher` context manager is the right resolution to this (or calling `release_conn()` explicitly). Will submit a PR for this.

timuralp:
AWS S3 servers set "Transfer-Encoding: chunked" even for a HEAD
response, which results in urllib3 not returning the connection to the
pool after calling stream() or read_chunked().

The patch moves the HEAD request check into the error_catcher context
manager, which ensures to release the connection to the pool upon
completion.

Fixes #1234

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Changes
 dev (master)
 ------------
 
+* Put the connection back in the pool when calling stream() or read_chunked() on
+  a chunked HEAD response. (Issue #1234)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -3,6 +3,7 @@ import socket
 from io import BytesIO, BufferedReader
 
 import pytest
+import mock
 
 from urllib3.response import HTTPResponse
 from urllib3.exceptions import (
@@ -605,6 +606,22 @@ class TestResponse(object):
         r.chunk_left = None
         resp = HTTPResponse(r, preload_content=False, headers={'transfer-encoding': 'chunked'})
         assert stream == list(resp.stream())
+
+    def test_chunked_head_response(self):
+        r = httplib.HTTPResponse(MockSock, method='HEAD')
+        r.chunked = True
+        r.chunk_left = None
+        resp = HTTPResponse('',
+                            preload_content=False,
+                            headers={'transfer-encoding': 'chunked'},
+                            original_response=r)
+        assert resp.chunked is True
+
+        resp.supports_chunked_reads = lambda: True
+        resp.release_conn = mock.Mock()
+        for _ in resp.stream():
+            continue
+        resp.release_conn.assert_called_once_with()
 
     def test_get_case_insensitive_headers(self):
         headers = {'host': 'example.com'}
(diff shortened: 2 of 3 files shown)


test/test_response.py
```diff
@@ -606,6 +606,29 @@ def test_chunked_response_with_extensions(self):
...
+        resp = HTTPResponse(MockChunkedEncodingResponse(''), preload_content=False,
+                            headers={'transfer-encoding': 'chunked'},
+                            original_response=r)
+        assert resp.chunked is True
+
+        class TrackRelease(object):
+            def __init__(self):
+                self.released = False
+
+        track_release = TrackRelease()
```
Lukasa: Rather than use this inner class we can save ourselves some time and just store the calls in a list. Alternatively, you can use the mock module directly, which may be the best approach.

Lukasa:
Can you also add a CHANGES.rst entry for this?

timuralp:
Done -- didn't notice the mock module being used earlier. That's definitely the cleaner way to test it.

CHANGES.rst
```diff
@@ -17,6 +17,9 @@ dev (master)
 
 * ... [Short description of non-trivial change.] (Issue #)
 
+* Put the connection back in the pool when calling stream() or read_chunked() on
+  a chunked HEAD response. (Issue #1234)
+
```
Lukasa: Can you put this above the placeholder instead of below?

timuralp:
Done -- rebased and moved the CHANGES.rst entry.

sigmavirus24:
@timuralp say "hi" to the folks at SwiftStack for me. Haven't spoken to any of them in a long time. :smile:

Lukasa:
Thanks for this @timuralp!

peterbe:
For the record, I spent an afternoon debugging requests and urllib3 and tonnes of googling trying to figure out why new HTTPS connections had to be set up after a `session.head(url)` when the URL was an S3 URL that always 404s. 

This PR totally fixes. I've compared (roughly 300 requests) using `urllib3==1.22` versus `urllib3@de81455` and the median time it takes to `asset session.head(url).status_code==404` went from 0.55 seconds to 0.13 seconds here on my home broadband (which is pretty fast). 

So yay!!

PR #1236, opened 2017-07-21
Fix compatibility with tornado 4.3+

felixonmars:
`HTTPHeaders` is not JSON serializable, so make it a dict first.


sigmavirus24:
Preliminary 👍

felixonmars:
codecov seems stuck :/

felixonmars:
OK, a branch update fixed codecov :)

sigmavirus24:
Thanks @felixonmars !

PR #1237, opened 2017-07-30
Let connectionpool dispose themselves

Ngo-The-Trung:
As mentioned in [this stackoverflow post](https://stackoverflow.com/questions/45369131/seeing-closedpoolerror-when-using-urllib3-poolmanager), there's an issue with the current PoolManager API which can close ConnectionPool objects before they are even used.

I don't think this is easily fixed at the PoolManager level because the API  doesn't give rooms for a way to notify that a ConnectionPool is being used (or is to be released). As such I propose adding a different set of data structures with different APIs to fix this.

I would like some review/discussion in terms of direction before I move on to writing tests and making final adjustments.

@Lukasa @shazow

### DIFF
--- a/urllib3/_collections.py
+++ b/urllib3/_collections.py
@@ -1,7 +1,8 @@
 from __future__ import absolute_import
 from collections import Mapping, MutableMapping
+from contextlib import contextmanager
 try:
-    from threading import RLock
+    from threading import RLock, Condition
 except ImportError:  # Platform-specific: No threads available
     class RLock:
         def __enter__(self):
@@ -18,7 +19,8 @@ except ImportError:
 from .packages.six import iterkeys, itervalues, PY3
 
 
-__all__ = ['RecentlyUsedContainer', 'HTTPHeaderDict']
+__all__ = ['RecentlyUsedContainer', 'BlockingRecentlyUsedContainer',
+           'HTTPHeaderDict']
 
 
 _Null = object()
@@ -96,6 +98,129 @@ class RecentlyUsedContainer(MutableMapping):
     def keys(self):
         with self.lock:
             return list(iterkeys(self._container))
+
+
+class BlockingRecentlyUsedContainer(object):
+    """
+    Provides a thread-safe container which maintains up to ``maxsize`` keys.
+    Key access is done through acquiring leases, which have to be released
+    afterwards. Inserting keys beyond ``maxsize`` throw away old keys, but not
+    before ensuring that all leases on such keys have been released.
+
+    Note: eviction/disposal is triggered on inserting new keys, not on
+    releasing of the last lease. In the case of connection pooling, this means
+    we will maintain the connections as long as possible.
+
+    This does not expose any means to clear the pool, because it would involve
+    waiting for all leases to be released. If there are concurrent acquisition
+    of leases, the clear method may block indefinitely.
+
+    :param maxsize:
+        Maximum number of recent elements to retain.
+
+    :param dispose_func:
+        Every time an item is evicted from the container,
+        ``dispose_func(value)`` is called.  Callback which will get called
+    """
+    ContainerCls = OrderedDict
+
+    def __init__(self, maxsize=10, dispose_func=None):
+        self._maxsize = maxsize
+        self._dispose_func = dispose_func
+
+        self._container = self.ContainerCls()
+        self.lock = RLock()
+
+        self._freelist = self.ContainerCls()
+        self._nwaiters = 0
+        self._container_not_full = Condition(self.lock)
+        self._victim = None
+
+    def acquire(self, key, constructor):
+        """
+        Acquires a lease on ``key`` inside this pool, if it has already
+        existed. If it hasn't, ``constructor`` is called to generate the value
+        and it is inserted into the pool beforehand.
+
+        If the insertion is done while the pool is full, another key is
+        evicted in a FIFO order, but not before all leases on the key have been
+        released.
+
+        :param key:
+            Key to acquire lease upon.
+
+        :param constructor:
+            Function-like, used to construct the new item to lease on if it
+            hasn't existed yet.
+
+        :returns value:
+            Value associated with key, or the value created by ``constructor``
+            if no such item can be found.
+        """
+        with self.lock:
+            result = self._container.get(key, _Null)
+
+            if result is not _Null:
+                refcount, value = self._container.pop(key)
+                self._container[key] = [refcount + 1, value]
+                _ = self._freelist.pop(key, _Null)
+                return value
+            else:
+                item = constructor()
+
+                if len(self._container) < self._maxsize:
+                    self._container[key] = [1, item]
+                else:
+                    evicted_value = _Null
+                    if len(self._freelist) > 0:
+                        _, evicted_value = self._freelist.popitem(last=False)
+                    else:
+                        self._nwaiters += 1
+                        self._container_not_full.wait()
+                        _, evicted_value = self._container.pop(self._victim)
+                        self._nwaiters -= 1
+
+                    self._container[key] = [1, item]
+                    if self._dispose_func and evicted_value is not _Null:
+                        self._dispose_func(evicted_value)
+
+                return item
+
+    def release(self, key, silent=False):
+        """
+        Release the lease on ``key``.
+        If other threads are waiting for for an eviction victim, notify them.
+
+        :param key:
+            Key to release lease from
+
+        :param silent:
+            If True, may raise RuntimeError if user release more times than
+            the number of times this key is acquired.
+        """
+        with self.lock:
+            refcount, item = self._container[key]
+            refcount -= 1
+            self._container[key][0] = refcount
+
+            if refcount == 0:
+                if self._nwaiters > 0:
+                    self._victim = key
+                    self._container_not_full.notify()
+                else:
+                    # Mark key as free for future eviction
+                    self._freelist[key] = item
+            elif refcount < 0 and not silent:
+                raise RuntimeError("Release of unacquired key")
+
+    @contextmanager
+    def acquire_then_release(self, key, constructor, silent=False):
+        value = self.acquire(key, constructor)
+        try:
+            yield value
+        except Exception:
+            raise
+        self.release(key, silent)
 
 
 class HTTPHeaderDict(MutableMapping):


Lukasa:
Hrm, I'm not sure how I feel about this approach. I think a better idea is going to be to reference count the underlying connectionpool objects: if they're being used, they must not be cleared. Is there a reason not to choose to use that approach instead of adding this extra complexity?

Ngo-The-Trung:
I actually kind of considered that, didn't feel confident enough to take it on.

I'll redo this later today using that.

Ngo-The-Trung:
@Lukasa how about now

Ngo-The-Trung:
How so? `PoolManager` can no longer close `ConnectionPool` objects since the `dispose_func` isn't passed to the container anymore. Each `ConnectionPool` manages its own closure when the last reference to it disappears.

Lukasa:
That's true, but we are now in a position where it's not possible to do eager closure of the pools: if you try to clear a poolmanager your outstanding sockets will get non-deterministically cleaned up. I think we need to maintain an actual reference count ourselves.

Ngo-The-Trung:
That leads to another question however: eviction should block until the evicted victim cleans up, shouldn't it? Otherwise, as it is right now, they simply "float" somewhere (we don't keep track of them). If the goal of closure is to ensure that all connections are cleaned up, it would be necessary to prevent these floating connections too, which requires us to block during eviction.

Another question: My previous [implementation](https://github.com/Ngo-The-Trung/urllib3/commit/3a9affc) has some refcounting/blocking mechanisms but you did say that it was too complex. Did I misunderstand this? (by "complexity" did you mean the API change? The refcount/blocking code? I suppose we can do without changing the API change)

Lukasa:
> That leads to another question however: eviction should block until the evicted victim cleans up, shouldn't it? 

Not necessarily, no. So long as we attempt to close things whenever we use them, it's fine.

But you've somewhat convinced me to reconsider the context manager based approach. I think we can make that work more simply though: want to push it back up and we can take another look at it?

Ngo-The-Trung:
Pushed.

One thing I haven't mentioned about this implementation is that it doesn't allow easy clear() either.

If concurrent threads perform additional acquire() while another is doing a clear(), it can keep blocking forever.

I considered making another class that wraps this BlockingRecentlyUsedContainer. When clear() is called we swap it out for a new container, so subsequent acquire() will go to the new container and doesn't prevent the old container from draining. But context managers belonging to the old container calls release() on the new container, it's a source for error.

I just decided not to make the clear() method in that case.

urllib3/_collections.py
```diff
@@ -98,6 +100,129 @@ def keys(self):
...
+
+                if len(self._container) < self._maxsize:
+                    self._container[key] = [1, item]
+                else:
+                    evicted_value = _Null
+                    if len(self._freelist) > 0:
+                        _, evicted_value = self._freelist.popitem(last=False)
+                    else:
+                        self._nwaiters += 1
+                        self._container_not_full.wait()
```
Lukasa: This doesn't work. The wait on the condition variable is done while holding `self.lock`. However, the *notify* of the condition variable requires holding `self.lock`, which means if anyone is waiting for a connection to be returned to this queue (in order to evict it) they will block all further queue activity forever. This is a pretty classic deadlock.

In this instance, I also think it's a bit overkill. There's no reason to block threads on this. Instead, we should just keep track of how many connection pools we have created. If that number goes above the "size" of this pool, then if there are any connections in the pool that have not been extracted they should be removed. If all connections from this pool have been removed, then they should be evicted in the order they're returned to the pool until the count drops below a certain level.

This means we don't block people forever waiting for connectionpools to be returned. It does allow a slight overcommitment of resources, but if people don't want that then they should probably just use a semaphore or something similar to manage how many outstanding requests they have.

Lukasa: Ah, sorry, I missed that in the constructor. However, I think my argument about this being overkill stands: we don't need the added complexity here.

Ngo-The-Trung:
Github is lagging and refusing to let me reply ...

The common case here is that most connectionpools are only acquired and released shortly after. The `_freelist` should almost always be populated.

The condition variable is only there to solve the rare scenario where all the poolmanager are being used. This most likely be because they configured the number of pools to be too low. This ensures that what the user configured (number of pools, size of each pool) is honored. We can also tune the overcommit ... but honestly I don't see the point, the user can reconfigure their parameter if they see this rare case appearing too often.

Lukasa:
> The condition variable is only there to solve the rare scenario where all the poolmanager are being used.

Sure, I get that, but that's part of why I don't think it's worth having it. It exists to provide blocking behaviour in an uncommon case. My suggestion is to say that in that uncommon case we should just keep making forward progress, and then throw connection pools away when they're returned later on. It saves us a bunch of book-keeping overhead, avoids a few confusing branches of code, and generally makes everyone's life easier.

Ngo-The-Trung:
What about the bit about predictability? If a user configures 5 pools with 5 connections per pool personally I'd expect the library not create more than 25 connections.

If the connections don't return sufficiently quickly, theoretically there isn't an upper bound.

Lukasa:
> If a user configures 5 pools with 5 connections per pool personally I'd expect the library not create more than 25 connections.

Then urllib3 will surprise you, because by default `ConnectionPool` objects are leaky: that is, they allow you to create more than the pool size of connections. They only *store* up to the pool size.

In this instance I don't think any particular expectation is more appropriate than the other: they're just both an assumption that needs to be validated by looking at the code. *In general* most users assume that if they call `PoolManager.urlopen` that we're going to service that request as swiftly as possible. Users that do not want that are likely to look at the documentation and code, at which point they'll discover the `block` parameter to the `ConnectionPool` and handle it appropriately.

I am open to adding blocking behaviour to the `PoolManager`, but I think I'd need to see some argument that it's actually necessary. We've never received a formal feature request for it, so to begin with I think we should take the simplest route to correctness, and then worry about whether we need to extend the approach to handle the additional features.

Ngo-The-Trung:
Perhaps our expectations are different here. I'm used to seeing other connection pools block on get(), example:

- [HikariCP](https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/pool/HikariPool.java#L154)
- [ActiveRecord connectionpool](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/ConnectionPool.html#method-i-checkout) (they don't use the word "block" but I suppose the "timeout" parameter is a hint enough, you can also verify [at the source](https://github.com/rails/rails/blob/3d00452f1f736efcdf074e782ac4b291e7bb2b5e/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L725))
- [another ruby connection_pool gem](https://github.com/mperham/connection_pool/blob/master/lib/connection_pool/timed_stack.rb#L64)

As for why I prefer that ConnectionPool does the book keeping for me (ensuring that the number of connections stay under a limit) is because it makes managing resources simpler. If server A is talking to server B using a connection pool with a 1000 connection limit makes for an easy calculation. This means if I spawn 9 more copies of server A, I know that B can't have more than 10K connections from As at any time.

If B is sufficiently slow, connections can linger for quite a long time before dying, so the max number of concurrent connections from A to B becomes harder to analyse. I'll have to set the sizes much lower, but even then it doesn't provide any sort of guarantee to me under different loads. Having a hard upper bound helps, which is why I trust that a ConnectionPool offer me this sort of hard guarantee.

I wouldn't mind if it *does* leak a few connections, but if the number of leaked connections scale with the latency of the destination host and the number of threads using the pool, it can be very tricky.

Lukasa:
> Perhaps our expectations are different here.

They are indeed. That's why I said "In this instance I don't think any particular expectation is more appropriate than the other: they're just both an assumption that needs to be validated by looking at the code." :smile:

> As for why I prefer that ConnectionPool does the book keeping for me (ensuring that the number of connections stay under a limit) is because it makes managing resources simpler.

So I understand that. In fact, `ConnectionPool` *does* enable this behaviour, for exactly this reason: managing the number of TCP connections between two hosts is useful. You can set the `block` parameter to `True` and get the behaviour you want.

But we're not talking about `ConnectionPool`, we're talking about `PoolManager`, which is a pool of pools. This makes the situation trickier. If you set a `PoolManager` with size 10, made up of connection pools of size 10, all of which block, how many connections can you have open at once? The answer is "unclear, somewhere between 10 and 100". If you only ever contact 10 hosts, then you'll trend towards 100 connections. If you contact lots of hosts, you'll trend towards 10 connections.

This means that a `PoolManager` size sets a hard cap on resource usage, but does not necessarily allow your code to actually reach that hard cap in regular usage. In many cases it won't even get close, e.g. if you're only using a few threads it's unlikely you'll *ever* maximally use a `ConnectionPool` so you'll end up *much* closer to the `PoolManager` limit than the limit `PoolManager x ConnectionPool`.

If we swap `PoolManager` to block by default when there are many connection pools in use, in practice this means blocks will be entirely unpredictable and will be predicated entirely on the usage model. Making many connections to few hosts will rarely block: making the same number of connections to many hosts will frequently block. This is a completely non-intuitive behaviour that is hard to reason about.

The reality is that there is no one perfect connection pooling strategy. There are many possible strategies, each of which at least one programmer thinks of as the "obvious" strategy, but in reality we need to accept trade-offs here. We definitely need to fix the issue where a `ConnectionPool` is closed while in use: that's a bug. Beyond that, we should only add support for what our users actually need, and right now I still don't see anyone complaining that they need a blocking `PoolManager`. If they need it, we can come back and add it, but for now we should aim to do the simplest thing that resolves the issue and maintains the previous behaviour.

Ngo-The-Trung:
Alright. I was expecting the PoolManager to compose the underlying pools into another pool-like construct, giving me the same set of guarantee (connection limit).

As of now it's a bit tough to use it properly.

Lukasa:
> Alright. I was expecting the PoolManager to compose the underlying pools into another pool-like construct, giving me the same set of guarantee (connection limit).

Yup, that's an understandable desire. Unfortunately it's just not the way urllib3 works today. If we were going to rewrite urllib3 from the start today the distinction between `PoolManager` and `ConnectionPool` would want to be rethought, but as it stands they work the way they do.

Sorry that we couldn't get closer to what you were looking for here. =(

Ngo-The-Trung:
No problem, I appreciate this conversation. 

but I have a few questions/feature requests if you don't mind. If changing the PoolManager (or creating an equivalent class) isn't desirable, because the class in itself wasn't meant to be a pool, how about we create a new ConnectionPool object that has built in hostname/context based connection multiplexing? I like this feature from PoolManager, but since it isn't completely composable with ConnectionPool objects, how about we bring it to the ConnectionPool objects themselves?

This pool would provide a strong upper bound on the number of connections. It can also adapt better to different workload. Say, if we we set the max num of connections to 1000, we can expect the pool to shift all connections from host A to host B depending on the workload at time. This can be more resource-wise than, say, a PoolManager of 100 pools, each hostname limited to 10 connections. It could also have parameters to fine-tune the limit, in case hammering any host at any time isn't a good idea.

Not related, but also a feature request: can there be an optional parameter for the PoolManager to resolve the hostname before converting to a key? At the moment if the set of destinations consist of many domains pointing to the same IP address (i.e. the domain has a lot of subdomains) the connections are not as freely reusable as they might be.

I'm willing to submit patches if you find either idea feasible.

Lukasa:
> If changing the PoolManager (or creating an equivalent class) isn't desirable, because the class in itself wasn't meant to be a pool, how about we create a new ConnectionPool object that has built in hostname/context based connection multiplexing? 

So I'm definitely *open* to this, but I think I need to stress how substantial a chunk of work this is. If we were going to do it it would amount to nothing short of a majority rewrite of the codebase: more than 50% of the code relies on the abstraction layer between `PoolManager` and `ConnectionPool`, and changing it would basically require redesigning the architecture of the library. 

That is not a "we don't want this": I think having different connection pool logic would likely be a good thing, and would certainly be a refactor worth investigating. But it'd be a big breaking change that would require a lot of work, and I think we'd need to see the whole patch before we decided to go forward on it. 

> an there be an optional parameter for the PoolManager to resolve the hostname before converting to a key? At the moment if the set of destinations consist of many domains pointing to the same IP address (i.e. the domain has a lot of subdomains) the connections are not as freely reusable as they might be.

Do you mind clarifying why exactly this is a problem? What is it about the DNS-based load balancing that interacts poorly with the host based pooling? Because it seems to me that pooling based on hostname should be completely legitimate.

Ngo-The-Trung:
Alright I'll make an attempt at the first idea.

About the second problem, I have a use case where I pool a lot of connections to various subdomains of a single website. All subdomains have the same IP address, so I suppose any TCP connections pointing to a subdomain is reusable by another. But without option to resolve the IP address first, the PoolManager will create different connectionpool objects for each subdomain, reducing the reusability of the connections in each pool.

Lukasa:
> But without option to resolve the IP address first, the PoolManager will create different connectionpool objects for each subdomain, reducing the reusability of the connections in each pool.

Ah, so this raises a really interesting question: is it acceptable to pool connections across *different* hostnames to the same IP address?

The answer here depends on your use-case. Famously browsers *do not* do this unconditionally for a number of reasons mostly relating to what's known as "ambient authority". It's also a bit difficult because it interacts awkwardly with TLS, e.g. you may be served a TLS cert valid for only one of the hostnames that resolve to a given IP, so it's not acceptable to coalesce them.

Generally speaking browsers treat hostnames as defining trust boundaries, and resources like connections should not be shared across these trust boundaries. I'm inclined to do the same here: the subtle risks of sharing connections to different hostnames is sufficiently risky that it seems generally unwise.

*However*, you may be able to achieve the effect you want by using hostnames in your URLs directly and setting host headers. Does that get you to where you want to go?

Ngo-The-Trung:
I probably should have considered TLS playing a role in this.

You're right setting the headers manually will solve my issue. I was wondering if it's better if the library does it for me, but considering what you mentioned it's probably better left to the user of the library.

PR #1240, opened 2017-08-06
Remove all uses of nose from the codebase

alexwlchan:
Related: #1160.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ clean:
 
 .PHONY: test
 test: requirements
-	nosetests
+	tox
 
 .PHONY: test-all
 test-all: requirements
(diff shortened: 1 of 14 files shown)


Lukasa:
Ok, Appveyor 3.3 is failing for shitty Appveyor reasons. Is there any reason for us to keep supporting 3.3?

Lukasa:
I mean, Requests supports it, but...probably it shouldn't.

Lukasa:
Sadly I may have to fix this. 🙁

alexwlchan:
> Is there any reason for us to keep supporting 3.3?

🔥 🔥 🔥

Lukasa:
Aha, [this](https://cryptography.io/en/latest/changelog/#v2-0) is the problem we're encountering. To maintain support for Python 3.3 is possible so long as we cap the cryptography version, which I don't think we should do, so I think we should begin sunsetting Python 3.3 support.

alexwlchan:
Woohoo, a green build!

PR #1241, opened 2017-08-08
Add 'other' counter

tuukkamustonen:
Similar to https://github.com/shazow/urllib3/issues/1147, would you be willing to merge this?

Honestly, I'm no longer sure this makes sense. My original goal was to retry different sorts of errors (read, connect, status) once each, the idea being something like: "it's not worth retrying connect timeout multiple times - if it doesn't answer right away, it probably doesn't answer after 5 retries..." but at the same time "let's allow retrying _different_ types or errors on the same go, because maybe proxy just disconnected us, and after re-initializing connection it might throw timeout, that we can also retry once"...

And actually, `ProxyError` (for example) doesn't fall into any category currently, so `other` counter would catch that. However, there are gazillions of potential exceptions in `requests.exceptions`, and this will regard all of them as one, so my idea doesn't quite realize here. Thus, I wonder if this makes any sense...

In the end, it doesn't really do any harm if I just `Retry(3)`... maybe not perfect, but it works and is more flexible...

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -228,7 +228,7 @@ In chronological order:
   * Stop buffering entire deflate-encoded responses.
 
 * Tuukka Mustonen <tuukka.mustonen@gmail.com>
-  * Add counter for status_forcelist retries.
+  * Add 'status' and 'other' counters.
 
 * Erik Rose <erik@mozilla.com>
   * Bugfix to pyopenssl vendoring
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -8,7 +8,7 @@ from urllib3.exceptions import (
     MaxRetryError,
     ReadTimeoutError,
     ResponseError,
-)
+    ProxyError)
 
 
 class TestRetry(object):
@@ -16,10 +16,10 @@ class TestRetry(object):
     def test_string(self):
         """ Retry string representation looks the way we expect """
         retry = Retry()
-        assert str(retry) == 'Retry(total=10, connect=None, read=None, redirect=None, status=None)'
+        assert str(retry) == 'Retry(total=10, connect=None, read=None, redirect=None, status=None, other=None)'
         for _ in range(3):
             retry = retry.increment(method='GET')
-        assert str(retry) == 'Retry(total=7, connect=None, read=None, redirect=None, status=None)'
+        assert str(retry) == 'Retry(total=7, connect=None, read=None, redirect=None, status=None, other=None)'
 
     def test_retry_both_specified(self):
         """Total can win if it's lower than the connect value"""
@@ -103,6 +103,14 @@ class TestRetry(object):
         with pytest.raises(MaxRetryError) as e:
             retry.increment(response=resp)
         assert str(e.value.reason) == ResponseError.SPECIFIC_ERROR.format(status_code=400)
+
+    def test_other_counter(self):
+        error = ProxyError()
+        retry = Retry(other=2)
+        retry = retry.increment(error=error)
+        retry = retry.increment(error=error)
+        with pytest.raises(MaxRetryError):
+            retry.increment(error=error)
 
     def test_backoff(self):
         """ Backoff is computed correctly """
--- a/urllib3/util/retry.py
+++ b/urllib3/util/retry.py
@@ -93,6 +93,12 @@ class Retry(object):
 
         Set to ``0`` to fail on the first retry of this type.
 
+    :param int other:
+        How many times to retry on issues that don't fall into other categories
+        (read, connect, redirect, status).
+
+        Set to ``0`` to fail on the first retry of this type.
+
     :param iterable method_whitelist:
         Set of uppercased HTTP method verbs that we should retry on.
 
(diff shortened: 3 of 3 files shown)


Lukasa:
Hrm, this is an interesting idea, but ultimately I don't know that this is the right approach. I think we generally want to try to keep a restricted scope of possible errors that can be retried and characterize them appropriately.

Does that seem reasonable to you?

tuukkamustonen:
Well, this doesn't remove anything, it just adds feature that you may ignore if you don't need it. So in that sense it should be harmless.

However, now that I'm taking a closer look [at the code](https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L626-L636), it seems that the amount of errors that are passed to `Retry.increment()` is very limited. I think only `SSLError` and `ProxyError` would hit [the else block](https://github.com/shazow/urllib3/blob/master/urllib3/util/retry.py#L369)? If so, it might make sense to split this `other` counter into `ssl` and `proxy` counters. But then, that's already quite a few counters, and if more exceptions are added over time, that would require more counters... so yeah, I don't know.

I know urllib3 only by surface. Does it make sense what I just wrote?

Lukasa:
> Well, this doesn't remove anything, it just adds feature that you may ignore if you don't need it.

Yeah, that's true, but that isn't a point in favour of merging it: it just means it's not a point in favour. 😉

In this instance I think we can probably suggest that we don't want another counter here. So long as we handle retries appropriately I think it'll be ok. In particular Proxy and SSL errors are not unambiguously things we can retry on: we should aim to get them to filter into the appropriate connection/read errors instead.

tuukkamustonen:
> Proxy and SSL errors are not unambiguously things we can retry on: we should aim to get them to filter into the appropriate connection/read errors instead.

Can you elaborate this? Do you mean mapping connection/read errors from proxy as connect/read errors from servers without proxy? That might make sense, yeah, if it's possible to distinguish them so.

But what about SSL errors? Aren't they radically different - no way they should be merged into connection/read errors?

Lukasa:
> Can you elaborate this? Do you mean mapping connection/read errors from proxy as connect/read errors from servers without proxy? That might make sense, yeah, if it's possible to distinguish them so.

Yes. 🙂

> But what about SSL errors? Aren't they radically different - no way they should be merged into connection/read errors?

Not really, no. Many SSL errors are symptomatic of bad connections, they just get treated as violations of the TLS framing layer. Retrying on these is not unreasonable: for example, there may be three origin servers behind a TCP load balancer where a subset of those servers is misconfigured, but the others aren't.

TLS certificate validation and downgrade attacks are not riskier if they are used as retry conditions, so I see no reason not to think of them as ordinary connection/read errors from the perspective of the retry logic.

tuukkamustonen:
> Retrying on these is not unreasonable: for example, there may be three origin servers behind a TCP load balancer where a subset of those servers is misconfigured, but the others aren't.

That's a good point.

> TLS certificate validation and downgrade attacks are not riskier if they are used as retry conditions, so I see no reason not to think of them as ordinary connection/read errors from the perspective of the retry logic.

Still, doesn't make sense to map them as read/connect errors, if that's what you suggested should happen, as they are not really "read timeout" or "connect timeout" type errors.

> Many SSL errors are symptomatic of bad connections, they just get treated as violations of the TLS framing layer.

There must be some reason to report them as SSL/TLS errors anyway, probably not worth trying to map them back - sounds complicated at least.

Lukasa:
> as they are not really "read timeout" or "connect timeout" type errors.

That's not the type of error in question. Note, for example, that `ProtocolError` is present in [this list](https://github.com/shazow/urllib3/blob/master/urllib3/util/retry.py#L279-L283).

> There must be some reason to report them as SSL/TLS errors anyway, probably not worth trying to map them back - sounds complicated at least.

We don't need to map them back: we just need to have the retry object think of them as the appropriate kind of exception. No need to change their type.

tuukkamustonen:
> We don't need to map them back: we just need to have the retry object think of them as the appropriate kind of exception. No need to change their type.

Yeah, I didn't mean that either (although it may have sounded like it). I just meant detecting which SSLError is actually a read/connect error and then handling like one sounds complicated.

Anyway, I think this discussion is beyond my understanding, and your time is better spent elsewhere. Thanks for taking a look at it, of course :)

tuukkamustonen:
@Lukasa Sorry for coming back to this, but I faced another problem: I am doing a request to log in via OAuth2 (for stupid reasons, I need to call the web flow programmatically) and it does multiple redirects. I need to follow those redirects but at the same time, I don't want to retry too many times. So I think I would need something like:

    Retry(total=1, redirect=10)

Which of course will not work. If I use:

    Retry(total=10)

That that could cause code to retry a read timeout 10 times. Doesn't sound good.

If I had the `other` counter, I could do:

    Retry(total=10, connect=1, read=1, status=1, other=1)

And that would allow 6-10 redirections.

Do you see the issue, would you have any suggestions?

Lukasa:
Yeah, this here is a good example of the problem with the current unification of redirects and retries. These shouldn't be handled by the same object, they're *completely* unrelated ideas. Really there should be two objects: a `Retry` object and a `Redirect` object. I think the correct change is to split these two entities out.

PR #1242, opened 2017-08-08
Goodbye Python 3.3. 👋

Lukasa:
Cryptography no longer supports it, so one of our fairly critical optional dependencies (PyOpenSSL) is no longer safe on Python 3.3. Given that it's out of support anyway, I think we should say goodbye to the old workhorse. Onward and upward!


SethMichaelLarson:
Why remove the 2.6 classifier if we still test against 2.6?

Lukasa:
Uh...because I thought we didn't, because I'm an idiot. 😅

SethMichaelLarson:
I mean it's more believably (imo) to be testing against 3.3 than 2.6 so it's understandable. 😉

PR #1244, opened 2017-08-09
Rewrite test_poolmanager.py to be pytest-style

alexwlchan:
Related: #1160.

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,5 +1,6 @@
 import socket
-import sys
+
+import pytest
 
 from urllib3.poolmanager import (
     PoolKey,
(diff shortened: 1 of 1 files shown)


test/test_poolmanager.py
```diff
@@ -13,30 +14,22 @@
...
-
 
-class TestPoolManager(unittest.TestCase):
+class TestPoolManager(object):
     def test_same_url(self):
         # Convince ourselves that normally we don't get the same object
         conn1 = connection_from_url('http://localhost:8081/foo')
         conn2 = connection_from_url('http://localhost:8081/bar')
-        self.addCleanup(conn1.close)
-        self.addCleanup(conn2.close)
```
Lukasa: Are we confident we don't need these?

alexwlchan: Yep. I ran the test before and after – no warnings either time about unclosed sockets. Additionally, I don’t see any new warnings appearing in CI.

Lukasa: Cool, ok then. :grin:

PR #1175, opened 2017-05-12
PyOpenSSL Load Chain Fix

theonewolf:
This pull request resolves #1060:

1. Created an intermediate CA
2. Created a client certificate off of that intermediate
3. Modified `dummyserver/server.py` to take certificates with `CERT_OPTIONAL`
    - Pre-existing test cases run unchanged
    - Test cases presenting validating certificates pass that along into the request handlers (Tornado)
4. Created two new test cases in `test/with_dummyserver/test_https.py`
    - One that tries a connection with client certificate alone
    - One that tries a connection with intermediate appended to client certificate

The last two commits fix the failing test case, and update the contributors list.

Issue #1060, opened 2016-11-30
No support for client certificate chains. urllib3 1.19.1


kevgliss:
Ran into an issue with client certificates where chain validation is failing.

Orginally reported:
https://github.com/kennethreitz/requests/issues/3732

Traceback:
```
Traceback (most recent call last):
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/contrib/pyopenssl.py", line 417, in wrap_socket
    cnx.do_handshake()
  File "/apps/lemur/lib/python3.5/site-packages/OpenSSL/SSL.py", line 1424, in do_handshake
    self._raise_ssl_error(self._ssl, result)
  File "/apps/lemur/lib/python3.5/site-packages/OpenSSL/SSL.py", line 1172, in _raise_ssl_error
    _raise_current_error()
  File "/apps/lemur/lib/python3.5/site-packages/OpenSSL/_util.py", line 48, in exception_from_error_queue
    raise exception_type(errors)
OpenSSL.SSL.Error: [('SSL routines', 'SSL3_READ_BYTES', 'tlsv1 alert unknown ca')]

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 594, in urlopen
    chunked=chunked)
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 350, in _make_request
    self._validate_conn(conn)
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 835, in _validate_conn
    conn.connect()
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/connection.py", line 323, in connect
    ssl_context=context)
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/util/ssl_.py", line 324, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/apps/lemur/lib/python3.5/site-packages/requests/packages/urllib3/contrib/pyopenssl.py", line 424, in wrap_socket
    raise ssl.SSLError('bad handshake: %r' % e)
ssl.SSLError: ("bad handshake: Error([('SSL routines', 'SSL3_READ_BYTES', 'tlsv1 alert unknown ca')],)",)

During handling of the above exception, another exception occurred:

[...]

PR #1246, opened 2017-08-15
PyOpenSSL client certificate load chain fix

rtdean:
Picking up the work from #1175, which addresses #1060.

I've rebased the existing PR#1175 against current master, and resolved the merge conflicts.  Additionally, I've fixed all the tests in #1175 so they pass on both mac OS (at least, on 10.12.x) and against the py26 environment.


rtdean:
The one failure, for py34 on macos, looks like a transient tornado test error.  We received the expected response on the first several retries, but then tornado started returning Internal Errors.  This error is against `test_client_no_intermediate`, which [I'm already ambivalent about](https://github.com/shazow/urllib3/pull/1175#issuecomment-321766463).

I'm not sure how to even begin writing tests around transient issues that may or may not pop up in tornado.  As I commented in #1175, I'm not sure the `test_client_no_intermediate` adds a ton of value aside from validating the test environment itself.  If it's purpose is to validate the test environment, then a more basic check for type of exception, as opposed to string matching, might be a better fit.  Alternatively, rerunning the rest will probably clear it.

dummyserver/server.py
```diff
@@ -30,6 +35,20 @@
 DEFAULT_CERTS = {
     'certfile': os.path.join(CERTS_PATH, 'server.crt'),
     'keyfile': os.path.join(CERTS_PATH, 'server.key'),
+    'cert_reqs': CERT_OPTIONAL,
+    'ca_certs': os.path.join(CERTS_PATH, 'cacert.pem'),
```
Lukasa: Why have we added these two lines?

rtdean: DEFAULT_CERTS is passed into tornado to init the SSL context.  We can't actually pass an SSL Context itself because that support wasn't added until 2.7.9, iirc.

The cert_reqs argument sets up the SSL Context to allow client certificate authentication, but not require it - you can connect via https without a certificate, but if you present a certificate, it needs to be validated.  ca_certs is used to contain the root CA for validation of the client certs - anything that can chain back to this root is considered A-OK.

I'll work up something for `CHANGES.rst` shortly.

Lukasa: Ah, sorry, I missed where this was used. Good call, seems reasonable to me.

test/with_dummyserver/test_https.py
```diff
@@ -67,6 +70,34 @@ def test_set_ssl_version_to_tlsv1(self):
...
+        )
+        https_pool = HTTPSConnectionPool(self.host, self.port,
+                                         cert_file=client_cert,
+                                         key_file=client_key)
+        try:
+            https_pool.request('GET', '/certificate', retries=False)
+        except SSLError as e:
+            self.assertTrue('alert unknown ca' in str(e) or
+                            'invalid certificate chain' in str(e) or
+                            'unknown Cert Authority' in str(e))
```
Lukasa: Nit: should we skip these two tests if `CERT_OPTIONAL` is `None`?

dummyserver/server.py
```diff
@@ -16,6 +16,11 @@
 import warnings
 from datetime import datetime
 
+try:
+    from ssl import CERT_OPTIONAL
+except:
+    CERT_OPTIONAL = None
```
Lukasa: Nit: this should probably catch only `ImportError`.

rtdean: ... :man_facepalming: I'll be honest, I didn't review more of the test code than what was necessary to fix the tests that were broken.  I'll go back and take a closer look at this.

rtdean: I can confirm that `ssl.CERT_OPTIONAL` is present in py26+.  I'm not sure why @theonewolf put the defensive coding around the import, but I'm fairly certain it doesn't need to be there.  In other tests dealing with SSL, I don't see any protections around `import ssl` statements, so I think this can safely be refactored back down to a simple from/import.

Lukasa: Cool, let's do that then and let the CI catch any issues. Assuming there aren't any, my note about skipping tests also naturally becomes irrelevant. :smile:

test/with_dummyserver/test_https.py
rtdean: `CERT_OPTIONAL` can't be None, so far as I can see.  Given that we don't have any tests around 'is the ssl module actually present on the system', these tests should be safe to always run.  I'll have another commit up momentarily refactoring the import.

dummyserver/server.py
rtdean: Alright, I'd like to ask for your thoughts, given that that this is your codebase.  `dummyserver` is currently free of imports of both `ssl` and `urllib3`.  the tornado webserver is completely initialized within dummynet, so we'll need to import one or the other here.

For the most part, tests in `test` just blindly `import ssl` - although there are a few tests that tiptoe around it, a bit.

There's also `urllib3.utils.ssl_` which has a function, `resolve_cert_reqs`, which would resolve the constant `CERT_OPTIONAL` to the correct numeric value.  `urllib3.utils.ssl_` appears to be written defensively to handle cases like a missing `ssl` module, but reading through the code I'm not 100% certain that it is safe for `CERT_*` resolution if the `ssl` module doesn't exist.

So, which do you prefer: the blind import that we use in more than half the cases in the `test` directory, or the possibly safer approach of bringing in `urllib3.utils.ssl_` into dummyserver... which would be the first dependency of dummyserver on urllib3.

rtdean: To be clear, I'm asking because if we do blindly `import ssl` into the dummyserver when running tests on an python install that doesn't have the `ssl` module, none of the tests will work, full stop... whereas they might work today, without the import.

On the other hand, keeping the dummyserver free of urllib code, so that we are only testing the urllib code as a client, not having any of it in the dummyserver, makes a certain amount of sense as well.

So which option is less-bad?

I must admit that I'm now curious how many of the tests would fail if run against a python without the `ssl` module...

Lukasa: I'm fine with the blind import. Realistically a Python without an `ssl` module is next to useless for HTTP, so I don't really mind.

test/with_dummyserver/test_https.py
mikeage: I just opened a PR (https://github.com/shazow/urllib3/pull/1251) and I had the travis build fail on a similar assert, but not this line:
```
    def test_client_no_intermediate(self):
        client_cert, client_key = (
            DEFAULT_CLIENT_NO_INTERMEDIATE_CERTS['certfile'],
            DEFAULT_CLIENT_NO_INTERMEDIATE_CERTS['keyfile']
        )
        https_pool = HTTPSConnectionPool(self.host, self.port,
                                         cert_file=client_cert,
                                         key_file=client_key)
        try:
            https_pool.request('GET', '/certificate', retries=False)
        except SSLError as e:
            self.assertTrue('alert unknown ca' in str(e) or
                            'invalid certificate chain' in str(e) or
>                           'unknown Cert Authority' in str(e))
E           AssertionError: False is not true
test/with_dummyserver/test_https.py:99: AssertionError
```
Could my failure be related to this commit?

Lukasa: Nope, your failure is just a transient one caused by Tornado misbehaving. =)

vingar:
Great! thanks a lot for this patch. Any idea when it'll be part of the next urllib3 release (1.23?) ?

Lukasa:
When such a release ships. Currently not expected for a few weeks.

theonewolf:
Thanks @rtdean!  I'm on a business trip in Bangalore, India; otherwise I would have normally worked on this.

@lukasa thanks for staying responsive throughout, and sorry I couldn't pull through sooner.

vpescado:
@Lukasa - Thank you for resolving this.  Any update on when this might ship?  It looks like there hasn't been a release in quite a bit.   Thanks!

haikuginger:
@vpescado I'm hoping to get a release out within the next week or two.

jnshey:
Thanks @haikuginger. Do you have any update on the new release? Thanks.

castrapel:
I'm also interested in the new release. Thank you!

SethMichaelLarson:
We're in the middle of a lead maintainer transition. A release will be discussed once the transition is completed. Thanks for your patience!

PR #1251, opened 2017-08-24
Add port to connection open print if it's non standard (not 80 or 443)

mikeage:
I do quite a lot of work with HTTP tunnels, and so the URL / IP I connect to is almost never enough to identify which endpoint I'm actually trying to reach. This commit adds the port, if it's not standard (80 for HTTP, 443 for HTTPS)

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -204,8 +204,9 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         Return a fresh :class:`HTTPConnection`.
         """
         self.num_connections += 1
-        log.debug("Starting new HTTP connection (%d): %s",
-                  self.num_connections, self.host)
+        log.debug("Starting new HTTP connection (%d): %s%s",
+                  self.num_connections, self.host,
+                  ":%s" % self.port if (self.port and self.port != 80) else "")
 
         conn = self.ConnectionCls(host=self.host, port=self.port,
                                   timeout=self.timeout.connect_timeout,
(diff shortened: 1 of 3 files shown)


mikeage:
Are you asking me or your fellow owners? ;)

Personally, I assumed it'd be best to minimize the changes for people who don't need this, especially considering HTTP implies 80 and HTTPS implies 443, and the current PR mimics the behavior of most browsers' address bars. On the other hand, it _is_ a debug level print, so maybe it doesn't hurt to be a bit verbose and specify everything explicitly. I'm sure I can twist myself into believing whatever I say last, so I'm not going to offer a recommendation!

Once we have a decision, I'll update CHANGES.rst.

Lukasa:
I was asking you. :wink:

My instinct is to be verbose here: the change will hit people who aren't expecting it one way or another, so I think we should just be complete and always offer that information.

mikeage:
OK, update on it's way. I noticed that sometimes `self.port` is None; I'll replace that with 80/443 as appropriate

mikeage:
@Lukasa , anything else you need from me?

urllib3/connectionpool.py
```diff
@@ -204,8 +204,9 @@ def _new_conn(self):
         Return a fresh :class:`HTTPConnection`.
         """
         self.num_connections += 1
-        log.debug("Starting new HTTP connection (%d): %s",
-                  self.num_connections, self.host)
+        log.debug("Starting new HTTP connection (%d): %s:%s",
+                  self.num_connections, self.host,
+                  self.port if self.port else "80")
```
haikuginger: Nit: here and below, you can use the idiom `self.port or "80"` to obtain the first truthy value.

mikeage:
Clever, thanks!

PR #1255, opened 2017-09-02
Use FQDN only for DNS and drop trailing dot for other operations

### LINKED ISSUE #1254: Match_hostname: hostname 'github.com.' doesn't match 'github.com' or 'www.github.com'
I'm getting hostname mismatch error when I try to access a domain with a trailing dot.

Example:

Version of urllib3 I'm using

```
(venv) ➜  ~ pip list --format=columns | grep urllib3
urllib3                      1.22    
```

```
(venv) ➜  ~ python
Python 2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
>>> import urllib3
>>> http = urllib3.PoolManager(cert_reqs='CERT_REQUIRED')
>>> r = http.request('GET', 'https://github.com/')
>>> print r.status
200
>>> r = http.request('GET', 'https://github.com./')
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/request.py", line 66, in request
    **urlopen_kw)
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/request.py", line 87, in request_encode_url
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/poolmanager.py", line 321, in urlopen
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/connectionpool.py", line 668, in urlopen
    **response_kw)
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/connectionpool.py", line 668, in urlopen
    **response_kw)
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/connectionpool.py", line 668, in urlopen
    **response_kw)
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/connectionpool.py", line 639, in urlopen
  File "/Users/karteek/venv/lib/python2.7/site-packages/urllib3/util/retry.py", line 388, in increment
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='github.com.', port=443): Max retries exceeded with url: / (Caused by SSLError(CertificateError("hostname 'github.com.' doesn't match either of 'github.com', 'www.github.com'",),))
```

Similar request using curl seems to go fine.

```
curl -v https://github.com./ 
*   Trying 192.30.255.112...
* Connected to github.com (192.30.255.112) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
* Server certificate: github.com
* Server certificate: DigiCert SHA2 Extended Validation Server CA
* Server certificate: DigiCert High Assurance EV Root CA
> GET / HTTP/1.1
> Host: github.com
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Thu, 31 Aug 2017 21:06:01 GMT
< Content-Type: text/html; charset=utf-8
< Transfer-Encoding: chunked
< Server: GitHub.com
< Status: 200 OK
< Cache-Control: no-cache
```

Also, going to similar URL https://github.com./robots.txt on browsers Firefox and Chrome doesn't show any SSL errors (They get redirected to https://github.com/robots.txt)

haikuginger:
Fixes #1254.

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -76,6 +76,23 @@ class SocketDummyServerTestCase(unittest.TestCase):
         if hasattr(cls, 'server_thread'):
             cls.server_thread.join(0.1)
 
+    def assert_header_received(
+        self,
+        received_headers,
+        header_name,
+        expected_value=None
+    ):
+        header_name = header_name.encode('ascii')
+        if expected_value is not None:
+            expected_value = expected_value.encode('ascii')
+        header_titles = []
+        for header in received_headers:
+            key, value = header.split(b': ')
+            header_titles.append(key)
+            if key == header_name and expected_value is not None:
+                self.assertEqual(value, expected_value)
+        self.assertIn(header_name, header_titles)
+
 
 class IPV4SocketDummyServerTestCase(SocketDummyServerTestCase):
     @classmethod
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1225,6 +1225,37 @@ class TestHeaders(SocketDummyServerTestCase):
         pool.request('GET', '/', headers=OrderedDict(expected_request_headers))
         self.assertEqual(expected_request_headers, actual_request_headers)
 
+    def test_request_host_header_ignores_fqdn_dot(self):
+
+        received_headers = []
+
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+
+            buf = b''
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+
+            for header in buf.split(b'\r\n')[1:]:
+                if header:
+                    received_headers.append(header)
+
+            sock.send((
+                u'HTTP/1.1 204 No Content\r\n'
+                u'Content-Length: 0\r\n'
+                u'\r\n').encode('ascii'))
+
+            sock.close()
+
+        self._start_server(socket_handler)
+
+        pool = HTTPConnectionPool(self.host + '.', self.port, retries=False)
+        self.addCleanup(pool.close)
+        pool.request('GET', '/')
+        self.assert_header_received(
+            received_headers, 'Host', '%s:%s' % (self.host, self.port)
+        )
+
     def test_response_headers_are_returned_in_the_original_order(self):
         # NOTE: Probability this test gives a false negative is 1/(K!)
         K = 16
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -124,6 +124,35 @@ class HTTPConnection(_HTTPConnection, object):
         # Superclass also sets self.source_address in Python 2.7+.
         _HTTPConnection.__init__(self, *args, **kw)
 
+    @property
+    def host(self):
+        """
+        Getter method to remove any trailing dots that indicate the hostname is an FQDN.
+
+        In general, SSL certificates don't include the trailing dot indicating a
+        fully-qualified domain name, and thus, they don't validate properly when
+        checked against a domain name that includes the dot. In addition, some
+        servers may not expect to receive the trailing dot when provided.
+
+        However, the hostname with trailing dot is critical to DNS resolution; doing a
+        lookup with the trailing dot will properly only resolve the appropriate FQDN,
+        whereas a lookup without a trailing dot will search the system's search domain
+        list. Thus, it's important to keep the original host around for use only in
+        those cases where it's appropriate (i.e., when doing DNS lookup to establish the
+        actual TCP connection across which we're going to send HTTP requests).
+        """
+        return self._dns_host.rstrip('.')
+
+    @host.setter
+    def host(self, value):
+        """
+        Setter for the `host` property.
+
+        We assume that only urllib3 uses the _dns_host attribute; httplib itself
+        only uses `host`, and it seems reasonable that other libraries follow suit.
+        """
+        self._dns_host = value
+
     def _new_conn(self):
         """ Establish a socket connection and set nodelay settings on it.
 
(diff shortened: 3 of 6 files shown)


haikuginger:
Looks like the new version of Tox doesn't like our config file. Looking into it; easy fix would be to pin to 2.7.0 for now.

Lukasa:
Cool, let's see how this goes.

appveyor.yml
```diff
@@ -57,7 +57,7 @@ install:
   # Upgrade to the latest version of pip to avoid it displaying warnings
   # about it being out of date.
   - pip install --disable-pip-version-check --user --upgrade pip wheel
-  - pip install tox virtualenv
+  - pip install tox==2.1.1 virtualenv
```
Lukasa: It would be nice if we could work out why tox is mad at us so we can remove these pins. Any idea?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1225,6 +1225,38 @@ def socket_handler(listener):
...
+            while not buf.endswith(b'\r\n\r\n'):
+                buf += sock.recv(65536)
+
+            headers_list = [header for header in buf.split(b'\r\n')[1:] if header]
+
+            for header in headers_list:
+                (key, value) = header.split(b': ')
+                if key.decode('ascii').startswith(u'Host'):
+                    received_host.append(value.decode('ascii'))
+
```
Lukasa: Might be nice to pull this header parsing logic out into a separate helper function. It's the kind of thing that might come up from time to time.

urllib3/connection.py
```diff
@@ -124,6 +124,32 @@ def __init__(self, *args, **kw):
...
+        """
+        return self._dns_host.rstrip('.')
+
+    @host.setter
+    def host(self, value):
+        """
+        Setter for the `host` property.
+        """
+        self._dns_host = value
+
```
Lukasa: Hrm, we might want to put a note here that points out that this assumes that only urllib3 code ever sets this property.

More generally, I wonder if we should just remove the setter and change the `__init__` to do the right thing. What are your thoughts on that?

sigmavirus24:
Tox looks to be broken on parsing `setenv` settings in `tox.ini` my suspicion would be in trying to parse https://github.com/shazow/urllib3/blob/5206734fda1f2365afc49a070efa46c9ba68372e/tox.ini#L38 since it's expecting that it can split on an `=`.

Lukasa:
Did that stop being valid syntax?

sigmavirus24:
I don't think so, I suspect it's a regression. They've had other regressions in the past involving references to other sections like this

test/with_dummyserver/test_socketlevel.py
haikuginger: Makes sense! Something like this?

```python
def assert_header_received(self, received_headers, header_name, expected_value=None):
    """
    Asserts that a header with the given name was received; if a value is passed to
    `expected_value`, asserts that the header has that value. 
    """
    pass
```

urllib3/connection.py
haikuginger: Makes sense to me.

I would tend towards not changing `__init__`; due to the way it's structured in httplib, we'd essentially have to copy out a chunk of stdlib code (not a huge chunk, sure, but a chunk) and make one small tweak; this is a relatively small change in comparison.

haikuginger:
[Looks like this is fixed now](https://github.com/tox-dev/tox/issues/595), so I'll remove the pins and check again.

haikuginger:
Okay, I think the single Appveyor fail is spurious, but I don't see a way to restart the build without pushing another commit, so that can wait until @Lukasa okays the test method I suggested. The GAE failure on Travis is pretty consistent though; @jonparrott, any thoughts on that as the resident GAE expert?

test/with_dummyserver/test_socketlevel.py
Lukasa: Yeah, sounds good to me!

urllib3/connection.py
Lukasa: Fair enough. :smile:

haikuginger:
Rebased and squashed from 62b9e1c.

haikuginger:
@Lukasa, it looks like Travis is currently experiencing issues with their macOS builds; if you want to give a runthrough and a final 👍, I can merge once those builds complete.

haikuginger:
Rebased from 88bcf98.

haikuginger:
Gentle ping to @Lukasa; please ignore if you're taking a break from OSS.

@sigmavirus24, can you provide a review in case @Lukasa is unavailable?

dummyserver/testcase.py
```diff
@@ -76,6 +76,23 @@ def tearDownClass(cls):
         if hasattr(cls, 'server_thread'):
             cls.server_thread.join(0.1)
 
+    def assert_header_received(
+            self,
+            received_headers,
+            header_name,
+            expected_value=None
+    ):
```
Lukasa: I'm not wild about this indentation. :wink: Mind matching it up with the rest of the project?

haikuginger: Sure! Another project I'm involved with hand linting set up to whine if the parameters line up with the first line of the function, and that's why I did it here, but I agree it looks skeevy.

karteek:
Ping @haikuginger. Please ignore if you are on break.

haikuginger:
@Lukasa, finally addressed that note.

Lukasa:
Looks like we need to add some pins to some of our dependencies for 2.6.

haikuginger:
Rebased from 36bf9cc to pick up test changes.

haikuginger:
@Lukasa, can you do a final run-through on this? I believe it should be set to merge once we have a green light from Travis (test issues were resolved elsewhere).

Issue #1256, opened 2017-09-02
Random test failures on some tests on master


reversefold:
I've been working on a patch and was trying to get a baseline for the test suites using tox on the master branch. I'm getting random hangs and test failures with the dummyserver tests. They will sometimes pass and sometimes fail. Here's an example of a failure:
```
________________________________________________________________________________________________ TestHTTPProxyManager.test_proxy_conn_fail _________________________________________________________________________________________________

self = <test.with_dummyserver.test_proxy_poolmanager.TestHTTPProxyManager testMethod=test_proxy_conn_fail>

    def test_proxy_conn_fail(self):
        host, port = get_unreachable_address()
        http = proxy_from_url('http://%s:%s/' % (host, port), retries=1, timeout=0.05)
        self.addCleanup(http.clear)                                                                                                                                                                                                                 self.assertRaises(MaxRetryError, http.request, 'GET',
                          '%s/' % self.https_url)
        self.assertRaises(MaxRetryError, http.request, 'GET',
                          '%s/' % self.http_url)

        try:
            http.request('GET', '%s/' % self.http_url)
            self.fail("Failed to raise retry error.")
        except MaxRetryError as e:
>           self.assertEqual(type(e.reason), ProxyError)
E           AssertionError: <class 'urllib3.exceptions.ConnectTimeoutError'> != <class 'urllib3.exceptions.ProxyError'>
                                                                                                                                                                                                                                            test/with_dummyserver/test_proxy_poolmanager.py:66: AssertionError                                                                                                                                                                          ----------------------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------------------
Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.VerifiedHTTPSConnection object at 0x111c7d610>, 'Connection to hmxysfyfmsecholvhzjufgbxfncwr
rtprlqjmcxktvhjdwzuydlfqyqaaplb timed out. (connect timeout=0.05)')': /
Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'ConnectTimeoutError(<urllib3.connection.VerifiedHTTPSConnection object at 0x111c7d610>, 'Connection to hmxysfyfmsecholvhzjufgbxfncwr
rtprlqjmcxktvhjdwzuydlfqyqaaplb timed out. (connect timeout=0.05)')': /
[...]

Issue #1252, opened 2017-08-27
PoolManager is not thread-safe


reversefold:
See, for reference this issue on the requests library:
https://github.com/requests/requests/issues/1871

In looking into the issue, which appears to happen when too many hosts are connected to through the same PoolManager in parallel, the issue appears to be with PoolManager's use of an LRU cache for the connection pool instances within it. A pool is pulled from the cache but within the time it takes for pool._get_conn to be called the PoolManager can potentially evict the ConnectionPool if too many other hosts are connected to in parallel.

In addition, since the pool also handles retries and redirects within urlopen, the pool needs to stay alive until the final connection is acquired from the pool, meaning we can't be sure we can release the pool from PoolManager until urlopen returns, essentially making PoolManager make requests serially only.

I only see a few options for fixing this.
1) Remove the LRU cache. This means either connection leaks or hard-limiting the number of hosts that a PoolManager can connect to. Not a very useful option.
2) Put the call(s) to pool.urlopen within a critical section. Basically a no-go as this means that concurrency is useless.
3) Implement ref-counting (a semaphore of some kind) to keep track of pools actually in use and don't allow them to be reaped if they have a reference. This seems complicated and a duplication of Python's own GC. Perhaps this could be implemented by using __del__ to close the pool and not explicitly closing the pool when it is evicted?
4) Refactor PoolManager and the connection pools such that, when using a PoolManager, retries are handled in the PoolManager and we can get up to the point of getting an actual connection out of the pool within a critical section then release the pool to be potentially evicted.

Lukasa:
Yes, this is a known issue. An abortive attempt to resolve it was made in #1237 but was eventually abandoned when the scope of the change came into the open. As I noted in #1232, the window of this concurrency problem is very small and in reality I think it can be fairly easily managed from outside urllib3, so it's not high up my priority list. However, if anyone else wanted to tackle this we'd happily accept patches to resolve it.

reversefold:
@Lukasa I think that the risk is a little higher than previously thought. Since HTTPConnectionPool.urlopen can end up recursing, the same pool can be used after network IO, potentially multiple times. This is a much larger window than from urlopen being invoked to the first _get_conn call as it involves actual network IO.

Lukasa:
The recursion is moderately risky, but in practice I'm still not convinced that the risk is that high. In the absence of data suggesting otherwise it's still unlikely to become a high priority for me.

reversefold:
[...]

PR #1257, opened 2017-09-03
Remove race conditions in PoolManager causing ClosedPoolErrors (#1252)

reversefold:
Fix for #1252. The implementation here is a little clunky for now as I was trying to change the existing code/logic as little as possible to avoid regressions.

List of changes:
* Introduces `test_retry_with_too_many_pools` which shows the `ClosedPoolError` happening in master
* Fixes an issue in the dummyserver's `retry_after` handler in Python 3 when using the default value for status
* Adds a `PersistentRetry` class which wraps `Retry` and allows for the same object to be used after `increment` is called
* Changes the retry logic in `HTTPConnectionPool` and `PoolManager` to loop instead of recurse
* Moves the base `HTTPConnectionPool.urlopen` functionality to `HTTPConnectionPool.urlopen_only` to allow for it to be called by `PoolManager` with no retry logic
* Refactors all retry logic in `HTTPConnectionPool.urlopen` into `URLOpenWithRetries` which takes a function to call for `urlopen_only` functionality.
  * Both `PoolManager` and `HTTPConnectionPool` end up calling `HTTPConnectionPool.urlopen_only` eventually
  * `HTTPConnectionPool`'s logic should be pretty much exactly the same as it was
  * `PoolManager` passes in a function for `urlopen_only` which acquires a pool from `self.pools` then calls `urlopen_only` on it. This means that if `URLOpenWithRetries` ends up retrying (currently will only happen if a `Retry-After` header is received) then `PoolManager` will create a new pool if the old one has been removed, avoiding the `ClosedPoolError` that can happen.
* `PoolManager` acquires `self.pools.lock` a second time now, before running its internal `urlopen_only` logic, to ensure that the lock stays acquired until the `HTTPConnectionPool` calls `_get_conn`. It passes the lock to `HTTPConnectionPool.urlopen_only` so that it can release the lock immediately once `_get_conn` is called.

Caveats/Potential improvements:
* Both URLOpenWithRetries and PoolManager still implement different redirect/retry logic. It would be nice to move all retry logic into URLOpenWithRetries.
  * Since `PoolManager` forces `redirect=False` when calling `HTTPConnectionPool.urlopen(_only)` there is only the `Retry-After` logic which can potentially cause a race condition on retry. Refactoring this further would reduce the complexity and remove the need for setting `redirect=False`.
* Passing the lock into `HTTPConnectionPool.urlopen_only` is a little hacky as currently implemented. It would be nice to have the connection acquiring logic also refactored out so that passing it through like this isn't needed and we can use the lock as a context manager.
* `PersistentRetry` may not be needed. This was implemented to ensure that the retries were maintained properly through the refactor as there are extra calls it may need to be passed through to maintain the context.
* Retry logic changed to loops makes the tracebacks slightly less informative with regards to retries but reduces the frame stack when retrying.
  * This should be able to be reverted if wanted. I originally made this change to make the logic a little easier to reason about and avoid potential issues when refactoring.
* `URLOpenWithRetries` is an odd name
* `URLOpenWithRetries` could possibly be changed to be a mixin, parent class, or wrapper instead of taking a function as a parameter to make its use more obvious
* `{[testenv]setenv}` change may not be needed but I ran into a lot of parsing errors with tox in some environments due to this.

I tested these changes on OSX (10.12.2) and on Ubuntu 16.04 with tox, as well as on travis-ci. There are random test failures, as mentioned in #1256, but the failures seem to correspond with those seen randomly on master in my tests. I have had each tox environment pass with no errors multiple times so I am fairly sure that there are no regressions.

https://travis-ci.org/reversefold/urllib3/builds/271452531 shows a fully passing run of this branch (except for gae, which appears to be broken).

---
This change is [Reviewable](https://reviewable.io/reviews/shazow/urllib3/1257)

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -243,5 +243,8 @@ In chronological order:
   * Fix ``util.selectors._fileobj_to_fd`` to accept ``long``.
   * Update appveyor tox setup to use the 64bit python.
 
+* Justin Patrin <papercrane@reversefold.com>
+  * Fix PoolManager and ConnectionPool to close race condition which can cause ClosedPoolError on retry
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -283,7 +283,7 @@ class TestingApp(RequestHandler):
         if datetime.now() - self.application.last_req < timedelta(seconds=1):
             status = request.params.get("status", "429 Too Many Requests")
             return Response(
-                    status=status.decode('utf-8'),
+                    status=status.decode('utf-8') if hasattr(status, 'decode') else status,
                     headers=[('Retry-After', '1')])
 
         self.application.last_req = datetime.now()
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -216,6 +216,24 @@ class TestPoolManager(HTTPDummyServerTestCase):
         r = http.request('GET', 'http://%s:%s/' % (self.host, self.port))
         self.assertEqual(r.status, 200)
 
+    def test_retry_with_too_many_pools(self):
+        http = PoolManager(num_pools=1, retries=5)
+        import threading
+        import time
+        threads = []
+        http.request('GET', '%s/retry_after' % (self.base_url,))
+        def invalidate_pool():
+            time.sleep(0.5)
+            http.request('GET', self.base_url_alt)
+        t = threading.Thread(target=invalidate_pool)
+        t.start()
+        threads.append(t)
+        try:
+            http.request('GET', '%s/retry_after' % (self.base_url,))
+        finally:
+            for t in threads:
+                t.join()
+
 
 @pytest.mark.skipif(
     not HAS_IPV6,
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -222,6 +222,7 @@ class TestPoolManager(HTTPDummyServerTestCase):
         import time
         threads = []
         http.request('GET', '%s/retry_after' % (self.base_url,))
+
         def invalidate_pool():
             time.sleep(0.5)
             http.request('GET', self.base_url_alt)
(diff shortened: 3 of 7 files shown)


Lukasa:
So this is generally an interesting direction, but I think it suffers from a bit of excessive complexity. 😁 The general goal seems to be achievable with slightly less overhead by invoking a careful refactor of urlopen to call a helper function that does no redirection, with redirection logic elsewhere. 

I think we should consider removing all retry logic from the ConnectionPool classes, which I think should make your life easier. Would it?

reversefold:
@Lukasa Removing the retry logic in the ConnectionPool classes would certainly reduce the complexity. As I said, I was trying to make this work without changing any existing functionality as I wasn't sure how much of this is intended to be externally exposed. Removing the retries from ConnectionPool seems like it could be a break in backwards compatibility.

If you're fine with moving that functionality to PoolManager I can make an attempt at that but it will likely take some test changes, making it a little harder to ensure that there are no regressions in top-level functionality.

Are there any places other than PoolManager which would need access to the retry logic.

reversefold:
Removed `PoolManager`'s redirect code as it isn't needed any more with the retry wrapper (with the addition of the `urljoin` call to support relative redirects). Now all retry/redirect code is in the wrapper.

Also renamed `URLOpenWithRetries` to `URLOpenRetryWrapper` and made a few other touchups.

The `urlopen` functions in `PoolManager` and `HTTPConnectionPool` are now fairly minimal as all of the logic now resides in `urlopen_only` and `URLOpenRetryWrapper`.

reversefold:
Ok, I've made the retry logic into a mixin and removed some more cruft that I'd ended up with during the initial work. This feels pretty solid to me. The only major issues I still have with it are the name ``urlopen_only`` and the passing of the lock into ``urlopen_only``. This still passes all tests without any test changes, so no tested logic should have regressed.

Lukasa:
Cool, so for my sake I think this needs to be broken into smaller PRs with consistent logical changes. 😁

reversefold:
I'll see what I can do.

reversefold:
@Lukasa I've opened 3 issues and PRs for the smaller parts of this PR. Once those are merged I'll work on the larger PRs which build upon them.

SethMichaelLarson:
I'm closing this PR due to staleness (see #1370) if you want this to get merged still I'll reopen following an update to this patch. Thanks!

PR #1259, opened 2017-09-06
Fix dummyserver decode errors when defaults are used with Python 3

reversefold:
Fixes issue https://github.com/shazow/urllib3/issues/1258
I searched for any usage of request.params.get() followed by a decode on the value.

A better version of a small part of PR https://github.com/shazow/urllib3/pull/1257 with some extra fixes.

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -142,8 +142,8 @@ class TestingApp(RequestHandler):
     def upload(self, request):
         "Confirm that the uploaded file conforms to specification"
         # FIXME: This is a huge broken mess
-        param = request.params.get('upload_param', 'myfile').decode('ascii')
-        filename = request.params.get('upload_filename', '').decode('utf-8')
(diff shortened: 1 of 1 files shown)


Lukasa:
Hey @jonparrott do you know what's going on with the GAE builds here?

reversefold:
@Lukasa @jonparrott After debugging the gae stuff I think I've figured out the problem. When I tested locally by installing the python env through gcloud it gave me `release: "1.9.57"` of `google_appengine`. If I download the latest the way that `_travis/install.sh` does, though, it gives me `release: "1.9.60"` which appears to have moved the `sandbox` module to `google.appengine.tools.devappserver2.python.runtime`.

I've opened a new PR for a fix: https://github.com/shazow/urllib3/pull/1265

reversefold:
Rebased on master so the gae test should work.

reversefold:
Rebased on master again.

PR #1261, opened 2017-09-06
Pass {posargs} to py.test so arguments can be passed through tox

reversefold:
From PR https://github.com/shazow/urllib3/pull/1257
Fixes issue https://github.com/shazow/urllib3/issues/1260

Allows for arguments to passed through tox to py.test so, for instance, individual tests can be run:
```
tox -e py36 -- -k 'test_poolmanager or test_connectionpool'
```


reversefold:
Rebased on master so the gae test should work.

Lukasa:
Thanks! :sparkles:

PR #1277, opened 2017-10-19
Update SNIMissingWarning text

nattofriends:
SNI is an abbreviation for Server Name Indication, not Subject Name Indication


Lukasa:
Thanks!

PR #1279, opened 2017-10-20
Prevent AttributeError to be raised when pool is already None

IceflowRE:
Calling close() multiple times will not result into an error now.

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -249,6 +249,23 @@ class TestConnectionPool(object):
         with pytest.raises(Empty):
             old_pool_queue.get(block=False)
 
+    def test_pool_close_twice(self):
+        pool = connection_from_url('http://google.com:80')
+
+        # Populate with some connections
+        conn1 = pool._get_conn()
+        conn2 = pool._get_conn()
+        pool._put_conn(conn1)
+        pool._put_conn(conn2)
+
+        pool.close()
+        assert pool.pool is None
+
+        try:
+            pool.close()
+        except AttributeError:
+            pytest.fail("Pool of the ConnectionPool is None and has no attribute get.")
+
     def test_pool_timeouts(self):
         with HTTPConnectionPool(host='localhost') as pool:
             conn = pool._new_conn()
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -411,6 +411,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         """
         Close all pooled connections and disable the pool.
         """
+        if self.pool is None:
+            return
         # Disable access to the pool
         old_pool, self.pool = self.pool, None
 


Lukasa:
Mind bringing this up to date with the current master?

IceflowRE:
Do i have to add a test?

haikuginger:
I'd want to see a test - obviously we were missing a flow where `self.pool` was `None`, but that'd be up to @Lukasa. Should be straightforward to create a pool, and then close it twice, asserting that it doesn't raise.

IceflowRE:
```python
        try: 
            pool.close() 
        except AttributeError: 
            pytest.fail("AttributeError raised.")
```
Would be this ok?
Here: https://github.com/shazow/urllib3/blob/master/test/test_connectionpool.py#L248

nateprewitt:
@IceflowRE, you'll probably want to make a new test to test this issue specifically. You can use the structure from the test you linked to for building up the new test.

Also using the pytest.raises context manager is probably preferable to a try/except block.

Lukasa:
Yup, adding a new test is the way to go. I also recommend adding a few *positive* assertions (that is, active assert rather than just let the test complete with no failing assertions).

IceflowRE:
Ok, iam not familiar with pytest itself.
But i think pytest.raises **must** raise an exception?

haikuginger:
I actually think the code block you have above should work; you'll just want it in its own test, and you'll also want to write a more explicit string about the test failure than "AttributeError raised."

test/test_connectionpool.py
```diff
@@ -247,6 +247,21 @@ def test_pool_close(self):
...
+
+        # Populate with some connections
+        conn1 = pool._get_conn()
+        conn2 = pool._get_conn()
+        pool._put_conn(conn1)
+        pool._put_conn(conn2)
+
+        try:
+            pool.close()
+            pool.close()
```
Lukasa: Presumably only the second call is expected to throw in the normal use case? In that case, let's remove the first call to `pool.close()` from the `try` block.

Can we also add some positive assertions? After the first call to `pool.close()` you can make some assertions about the `None`ness of the reference to the underlying pool.

IceflowRE: About both two points i wasnt sure.
1. Ok i will do
2. I let it out, becasuse its already tested inside the other test function and i want to prevent duplicated code, but if you want i will add this.

Lukasa: It's just not great to have tests that never explicitly assert anything.

IceflowRE:
Did it.

IceflowRE:
Something else has to do here?

haikuginger:
@IceflowRE, sorry, it looks like this PR got lost in the shuffle! Can you rebase it on master to resolve the flake8 issues? Once you've done so, this looks good to me.

IceflowRE:
Done.

haikuginger:
Thanks so much for the change, @IceflowRE! It's much appreciated!

PR #1283, opened 2017-11-20
Provide _original_response in AppEngine adapter

erlichmen:
_original_response is missing from the AppEngine adapter this breaks many flows like cookie parsing. This is an attempt to fix this issue.

### DIFF
--- a/urllib3/contrib/appengine.py
+++ b/urllib3/contrib/appengine.py
@@ -236,12 +236,21 @@ class AppEngineManager(RequestMethods):
             encodings.remove('chunked')
             urlfetch_resp.headers['transfer-encoding'] = ','.join(encodings)
 
-        return HTTPResponse(
+        original_response = HTTPResponse(
             # In order for decoding to work, we must present the content as
             # a file-like object.
             body=BytesIO(urlfetch_resp.content),
+            msg=urlfetch_resp.header_msg,
             headers=urlfetch_resp.headers,
             status=urlfetch_resp.status_code,
+            **response_kw
+        )
+
+        return HTTPResponse(
+            body=BytesIO(urlfetch_resp.content),
+            headers=urlfetch_resp.headers,
+            status=urlfetch_resp.status_code,
+            original_response=original_response,
             **response_kw
         )
 
--- a/urllib3/response.py
+++ b/urllib3/response.py
@@ -112,7 +112,7 @@ class HTTPResponse(io.IOBase):
 
     def __init__(self, body='', headers=None, status=0, version=0, reason=None,
                  strict=0, preload_content=True, decode_content=True,
-                 original_response=None, pool=None, connection=None,
+                 original_response=None, pool=None, connection=None, msg=None,
                  retries=None, enforce_content_length=False, request_method=None):
 
         if isinstance(headers, HTTPHeaderDict):
@@ -132,6 +132,7 @@ class HTTPResponse(io.IOBase):
         self._fp = None
         self._original_response = original_response
         self._fp_bytes_read = 0
+        self.msg = msg
 
         if body and isinstance(body, (basestring, binary_type)):
             self._body = body
@@ -190,6 +191,9 @@ class HTTPResponse(io.IOBase):
     @property
     def connection(self):
         return self._connection
+
+    def isclosed(self):
+        return is_fp_closed(self._fp)
 
     def tell(self):
         """
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -216,10 +216,12 @@ class TestResponse(object):
                 assert not resp.closed
                 assert not resp._fp.isclosed()
                 assert not is_fp_closed(resp._fp)
+                assert not resp.isclosed()
                 resp.read()
                 assert resp.closed
                 assert resp._fp.isclosed()
                 assert is_fp_closed(resp._fp)
+                assert resp.isclosed()
         finally:
             hlr.close()
 


Lukasa:
@jonparrott Can I get your eyes on this patch?

jonparrott:
This seems reasonable.

urllib3/response.py
```diff
@@ -132,6 +132,7 @@ def __init__(self, body='', headers=None, status=0, version=0, reason=None,
         self._fp = None
         self._original_response = original_response
         self._fp_bytes_read = 0
+        self._msg = msg
```
sigmavirus24: Is there a reason we need to use a property for this?

urllib3/response.py
```diff
@@ -191,6 +192,13 @@ def data(self):
     def connection(self):
         return self._connection
 
+    def isclosed(self):
+        return is_fp_closed(self._fp)
+
+    @property
+    def msg(self):
```
sigmavirus24: I'd prefer we add a docstring (or comment) that explains why we added this.

urllib3/response.py
erlichmen: msg has the headers in list from (headers has it in dict).
This is important for the Set-Cookie and  Set-Cookie2 headers that can are multiple-values-same-key headers that the headers by combining them into a single key lose the ability to parse them and set the value accordingly in the session header.

msg is also used in the _original_response context only by the stack which expects Response object of httplib2. A different solution might be to subclass this class and put msg only in the sub class and make _original_response instantiate that new sub class but this looks like a bit overrkill for the problem in hand.

urllib3/contrib/appengine.py
```diff
@@ -236,13 +236,21 @@ def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
...
             urlfetch_resp.headers['transfer-encoding'] = ','.join(encodings)
 
+        original_response = HTTPResponse(
+          # In order for decoding to work, we must present the content as
+          # a file-like object.
+          body=BytesIO(urlfetch_resp.content),
+          msg=urlfetch_resp.header_msg,
+          headers=urlfetch_resp.headers,
+          status=urlfetch_resp.status_code,
+          **response_kw)
```
Lukasa: The indentation here is off: can you indent another two spaces?

urllib3/contrib/appengine.py
```diff
@@ -236,13 +236,21 @@ def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
...
-            body=BytesIO(urlfetch_resp.content),
-            headers=urlfetch_resp.headers,
-            status=urlfetch_resp.status_code,
-            **response_kw
+          body=BytesIO(urlfetch_resp.content),
+          headers=urlfetch_resp.headers,
+          status=urlfetch_resp.status_code,
+          original_response=original_response,
+          **response_kw
         )
```
Lukasa: Same here.

urllib3/contrib/appengine.py
Lukasa: Also can you match the house style of having the closing parenthesis on its own line?

erlichmen:
@Lukasa Intent fixed.

urllib3/response.py
sigmavirus24: Sorry, I guess my question was not clear enough. Let me rephrase: Why can't we more simply do `self.msg = msg`?

erlichmen:
This is a read only property, I don’t see a reason to allow modifying it
after the ctor

sigmavirus24:
@erlichmen given the usecase it's being added to support, we don't need that boilerplate, in other words, since we don't anticipate people modifying it, why are we even bothering to go through with this. Let's keep things simple and if we find there's a convincing case for this boilerplate, we can change it.

erlichmen:
@sigmavirus24 done

urllib3/response.py
```diff
@@ -572,7 +576,6 @@ def read_chunked(self, amt=None, decode_content=None):
         """
         Similar to :meth:`HTTPResponse.read`, but with an additional
         parameter: ``decode_content``.
-
```
sigmavirus24: Please add this new-line back?

erlichmen:
@sigmavirus24 done (+ PR is now squashed)
@Lukasa take another look

sigmavirus24:
That all looks good. Now we can focus on test coverage. I think most of what we're missing is your new method for `isclosed`. Adding some tests around that should get things to pass.

trickidicki:
Hoping this can get resolved quickly as it's a blocking issue for me at the moment...

sigmavirus24:
@trickidicki while I appreciate that this is preventing you from completing whatever it is you're working on, this patch will not be resolved until the code coverage is fixed. It seems like that hasn't happened since my last comment. If you want it to move along, perhaps you should write some tests for this to make it pass coverage. Otherwise, please be patient. @erlichmen, @haikuginger and I all have personal and professional lives that come before this project and your deadlines.

erlichmen:
I add to an existing test assert over the new code, hopefully that eventually it will show up here.

sigmavirus24:
Closing and re-opening to give AppVeyor a kick

urllib3/contrib/appengine.py
```diff
@@ -236,12 +236,21 @@ def _urlfetch_response_to_http_response(self, urlfetch_resp, **response_kw):
...
             # In order for decoding to work, we must present the content as
             # a file-like object.
+            body=BytesIO(urlfetch_resp.content),
+            msg=urlfetch_resp.header_msg,
+            headers=urlfetch_resp.headers,
+            status=urlfetch_resp.status_code,
+            **response_kw
+        )
+
+        return HTTPResponse(
```
haikuginger: Minor question: why are we setting `msg` on `original_response`, but not on the response object we return?

erlichmen:
Only the code flow that needs original_response is using the msg field. I
don’t want to create a “new” API just because the internal code needs it

haikuginger:
@erlichmen, great, thanks! Can you rebase this on master? It seems to be having some issues with our coverage system.

erlichmen:
@haikuginger done

PR #1284, opened 2017-11-20
Don't try to run IPv6 tests if IPv6 is disabled at runtime

alex:


alex:
Hmmm, travis appears to be failing, but in precisely the way I expected this patch to resolve...

Lukasa:
Those tests use IPv6HTTPDummyProxyTestCase, which I think your patch missed.

alex:
@Lukasa good catch, fixed. (flake8 is still failing, should I fix those up, or can that go in it's own PR?)

Lukasa:
They can be fixed separately, thanks Alex!

Lukasa:
Oh, lol, no they can’t: they’re a required status

PR #1287, opened 2017-11-30
prevent sending IP address as SNI hostname

### LINKED ISSUE #1224: when connecting to https by IP address, client hello message has a server_name SNI block with literal ip address
I originally opened this in [requests](https://github.com/requests/requests/issues/4178), but was asked to raise it here.

When connecting to a machine over https using an IP address, the SSL client hello includes a server_name extension with the ip address listed as the server name. This is not allowed in the [SNI spec](https://tools.ietf.org/html/rfc3546#section-3.1), and it should not include the server_name extension. 

Reproduce:
1. Set up packet logging
2. ```python
    import urllib3
    http = urllib3.PoolManager()
    http.request('GET', 'https://8.8.8.8')
    ````
3. Inspect packet logs for client hello. It'll look a bit like this:
```
Frame 31: 571 bytes on wire (4568 bits), 571 bytes captured (4568 bits) on interface 0
Ethernet II, Src: BizlinkK_XX:XX:XX (9c:eb:e8:XX:XX:XX), Dst: JuniperN_XX:XX:XX (08:81:f4:XX:XX:XX)
Internet Protocol Version 4, Src: 192.168.89.18, Dst: 8.8.8.8
Transmission Control Protocol, Src Port: 1156, Dst Port: 443, Seq: 1, Ack: 1, Len: 517
Secure Sockets Layer
    TLSv1.2 Record Layer: Handshake Protocol: Client Hello
        Content Type: Handshake (22)
        Version: TLS 1.0 (0x0301)
        Length: 512
        Handshake Protocol: Client Hello
            Handshake Type: Client Hello (1)
            Length: 508
            Version: TLS 1.2 (0x0303)
            Random
            Session ID Length: 0
            Cipher Suites Length: 148
            Cipher Suites (74 suites)
            Compression Methods Length: 1
            Compression Methods (1 method)
            Extensions Length: 319
            Extension: server_name
                Type: server_name (0x0000)
                Length: 12
                Server Name Indication extension
                    Server Name list length: 10
                    Server Name Type: host_name (0)
                    Server Name length: 7
                    Server Name: 8.8.8.8
            Extension: ec_point_formats
            Extension: elliptic_curves
            Extension: SessionTicket TLS
            Extension: signature_algorithms
            Extension: Heartbeat
            Extension: Padding
```

The particular server I'm connecting to (azure windows 2016 Datacenter edition) immediately closes the connection when it sees a client hello like this.

nitzmahone:
* fixes #1224 (when the `ipaddress` module is available)

* Due to a Python SSL bug, IP addresses are sent as DNS hostnames in the SNI extension during TLS handshake (in violation of RFC6066). This causes some server implementations (eg, Microsoft http.sys) to abort the connection. This change uses the `ipaddress` module, if available, to prevent IP addresses from being sent as an SNI hostname to `wrap_socket()`.

### DIFF
--- a/urllib3/util/ssl_.py
+++ b/urllib3/util/ssl_.py
@@ -11,6 +11,7 @@ from ..exceptions import SSLError, InsecurePlatformWarning, SNIMissingWarning
 
 SSLContext = None
 HAS_SNI = False
+HAS_IPADDRESS = False
 IS_PYOPENSSL = False
 IS_SECURETRANSPORT = False
 
@@ -52,6 +53,12 @@ try:
 except ImportError:
     OP_NO_SSLv2, OP_NO_SSLv3 = 0x1000000, 0x2000000
     OP_NO_COMPRESSION = 0x20000
+
+try:
+    import ipaddress
+    HAS_IPADDRESS = True
+except ImportError:
+    pass
 
 # A secure default.
 # Sources for more information on TLS ciphers:
@@ -323,19 +330,42 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         # try to load OS default certs; works well on Windows (require Python3.4+)
         context.load_default_certs()
 
+    # Due to a bug in Python's handling of SNI extensions (sends IP as SNI hostname when connecting via IP address)
+    # we shouldn't send things that parse as IP addresses to wrap_socket as a hostname.
+    # If ipaddress is available, use it to filter those out.
+    server_hostname_is_ip = False
+
+    if HAS_IPADDRESS:
+        try:
+            # ipaddress doesn't accept byte strings
+            if isinstance(server_hostname, u''.__class__):
+                server_hostname_uni = server_hostname
+            else:
+                server_hostname_uni = server_hostname.decode('utf-8')
+
+            ipaddress.ip_address(server_hostname_uni)
+            server_hostname_is_ip = True
+        except ipaddress.AddressValueError:
+            pass
+
     if certfile:
         context.load_cert_chain(certfile, keyfile)
-    if HAS_SNI:  # Platform-specific: OpenSSL with enabled SNI
-        return context.wrap_socket(sock, server_hostname=server_hostname)
 
-    warnings.warn(
-        'An HTTPS request has been made, but the SNI (Server Name '
-        'Indication) extension to TLS is not available on this platform. '
-        'This may cause the server to present an incorrect TLS '
-        'certificate, which can cause validation failures. You can upgrade to '
-        'a newer version of Python to solve this. For more information, see '
-        'https://urllib3.readthedocs.io/en/latest/advanced-usage.html'
-        '#ssl-warnings',
-        SNIMissingWarning
-    )
-    return context.wrap_socket(sock)
+    wrap_socket_kwargs = {}
+
+    if HAS_SNI:  # Platform-specific: OpenSSL with enabled SNI
+        if not server_hostname_is_ip:
+            wrap_socket_kwargs['server_hostname'] = server_hostname
+    else:
+        warnings.warn(
+            'An HTTPS request has been made, but the SNI (Server Name '
+            'Indication) extension to TLS is not available on this platform. '
+            'This may cause the server to present an incorrect TLS '
+            'certificate, which can cause validation failures. You can upgrade to '
+            'a newer version of Python to solve this. For more information, see '
+            'https://urllib3.readthedocs.io/en/latest/advanced-usage.html'
+            '#ssl-warnings',
+            SNIMissingWarning
+        )
+
+    return context.wrap_socket(sock, **wrap_socket_kwargs)


nitzmahone:
cc @Lukasa

urllib3/util/ssl_.py
```diff
@@ -53,6 +54,12 @@ def _const_compare_digest_backport(a, b):
     OP_NO_SSLv2, OP_NO_SSLv3 = 0x1000000, 0x2000000
     OP_NO_COMPRESSION = 0x20000
 
+try:
+    import ipaddress
```
tiran: You can use the socket module to check if a string is an IP address:

```
import socket

def check_ipaddr(hostname):
    families = [socket.AF_INET]
    if hasattr(socket, 'AF_INET6'):
        families.append(socket.AF_INET6)
    for af in families:
        try:
            socket.inet_pton(af, hostname)
        except socket.error:
            pass
        else:
            return True
    return False
```

urllib3/util/ssl_.py
```diff
@@ -323,19 +330,42 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
...
-        'This may cause the server to present an incorrect TLS '
-        'certificate, which can cause validation failures. You can upgrade to '
-        'a newer version of Python to solve this. For more information, see '
-        'https://urllib3.readthedocs.io/en/latest/advanced-usage.html'
-        '#ssl-warnings',
-        SNIMissingWarning
-    )
-    return context.wrap_socket(sock)
+        if not server_hostname_is_ip:
+            wrap_socket_kwargs['server_hostname'] = server_hostname
```
tiran: You need to do hostname verification yourself if you do not supply server_hostname. Otherwise you do not check if the IP address is valid for the server's cert.

urllib3/util/ssl_.py
nitzmahone: Nice- that's clever. We'd actually been kicking around ways to do this in Ansible without taking the ipaddress dep- too bad it doesn't always work for IPv6 on Py2, but still...

tiran: It should always work when the platform support IPv6. Needs verification :)

urllib3/util/ssl_.py
```diff
@@ -323,19 +330,42 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
...
 
+    # Due to a bug in Python's handling of SNI extensions (sends IP as SNI hostname when connecting via IP address)
+    # we shouldn't send things that parse as IP addresses to wrap_socket as a hostname.
+    # If ipaddress is available, use it to filter those out.
+    server_hostname_is_ip = False
+
+    if HAS_IPADDRESS:
+        try:
+            # ipaddress doesn't accept byte strings
+            if isinstance(server_hostname, u''.__class__):
```
haikuginger: Please use `six.text_type` explicitly instead.

urllib3/util/ssl_.py
```diff
@@ -323,19 +330,42 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
...
+                server_hostname_uni = server_hostname
+            else:
+                server_hostname_uni = server_hostname.decode('utf-8')
```
haikuginger: Should we assume UTF-8? `locale.getpreferredencoding()` might be a safer bet.

urllib3/util/ssl_.py
haikuginger: Yeah; we would need a solution that 1. prevents the IP address from being sent pre-emptively as part of SNI but 2. does check the eventual received cert.

urllib3/util/ssl_.py
```diff
@@ -323,19 +330,42 @@ def ssl_wrap_socket(sock, keyfile=None, certfile=None, cert_reqs=None,
         # try to load OS default certs; works well on Windows (require Python3.4+)
         context.load_default_certs()
 
+    # Due to a bug in Python's handling of SNI extensions (sends IP as SNI hostname when connecting via IP address)
+    # we shouldn't send things that parse as IP addresses to wrap_socket as a hostname.
+    # If ipaddress is available, use it to filter those out.
+    server_hostname_is_ip = False
+
+    if HAS_IPADDRESS:
```
haikuginger: I'd rather see this in a separate util function.

urllib3/util/ssl_.py
SethMichaelLarson: If the `locale` route is taken we should probably make sure to pass `False` into the function to avoid a `setlocale()`? `setlocale()` isn't thread-safe.

nitzmahone:
Thanks for the review @haikuginger! I've been meaning to get back to this- while a "proper" fix in pyopenssl will be great, there will be a lot of broken stuff on old Python/pyopenssl out there for a long time to come... Some version of this fix in urllib3 would solve a lot of pain. I'll try to find a hole during the Ansible release cycle to finish it up.

mcassaniti:
Any further progress @nitzmahone? We hit #1224 at work with Ansible. It would be much quicker to patch urllib3 in our environment even if we had to carry our own patch then to build a full DNS solution to fix this issue. For now I've stripped out the SNI code from urllib3 on our Ansible controllers.

tiran:
@mcassaniti I have fixed the SNI bug in Python upstream. The next releases of Python 2.7, 3.6, and 3.7 will no longer send IP addresses in the SNI TLS extension. I'll talk to some people to get the fix into Fedora, Debian, RHEL, and Ubuntu earlier.

bread crumbs for @doko42 and @encukou

encukou:
I filed the bugs in Fedora, so we'll get to it soon.
I'm a bit concerned about this process, though: if there's a bugfix that needs to get to distributions ASAP, why is CPython not doing a release with it? That would reach further than pinging just the Fedora/Debian maintainers.

encukou:
(That may have sounded more hostile than I intended. Let me try again.)

Thank you for letting me know; I appreciate it! 
I'm just a bit concerned about less popular distros – it doesn't seem fair that they didn't get pinged. But they should get the fix soon in the next release.
IOW, I'm honored that I'm a “VIP”, but if anyone sees an easy alternative with no VIPs at all, I'd prefer it :)

mcassaniti:
Thank you for the upstream work @tiran, it is much appreciated. Like @nitzmahone I'm using Ansible. It is driven from an Alpine Linux container. I'm concerned about how long it may take the distro to get an updated Python 2.7 release into their stable branch (edge currently has 2.7.14-r2). For now I've cut out the SNI code from urllib3 since the only devices being connected to from this container are WinRM endpoints. I'd prefer a patch that I could incorporate as a stop-gap into my container, but what I have is working OK.

Again, thanks for the work upstream. My appreciation of the open source community as a whole continues to grow all the time.

SethMichaelLarson:
I'm closing this PR due to staleness (see #1370) if you want this to get merged still I'll reopen following an update to this patch. Thanks!

stintel:
FYI, ran into this problem myself, and it is fixed by e1b40a73e1f5b6140f46aaf8638e1a14b04d7cad which can be cleanly applied to v1.22.

PR #1289, opened 2017-12-02
Handle gzip responses with multiple members

isaacto:
Signed-off-by: Isaac To <isaacto@gmail.com>

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -1,4 +1,5 @@
 import socket
+import zlib
 
 from io import BytesIO, BufferedReader
 
@@ -96,7 +97,6 @@ class TestResponse(object):
         assert r.read() == b''
 
     def test_decode_deflate(self):
-        import zlib
         data = zlib.compress(b'foo')
 
         fp = BytesIO(data)
@@ -105,7 +105,6 @@ class TestResponse(object):
         assert r.data == b'foo'
 
     def test_decode_deflate_case_insensitve(self):
-        import zlib
         data = zlib.compress(b'foo')
 
         fp = BytesIO(data)
@@ -114,7 +113,6 @@ class TestResponse(object):
         assert r.data == b'foo'
 
     def test_chunked_decoding_deflate(self):
-        import zlib
         data = zlib.compress(b'foo')
 
         fp = BytesIO(data)
@@ -132,7 +130,6 @@ class TestResponse(object):
         assert r.read() == b''
 
     def test_chunked_decoding_deflate2(self):
-        import zlib
         compress = zlib.compressobj(6, zlib.DEFLATED, -zlib.MAX_WBITS)
         data = compress.compress(b'foo')
         data += compress.flush()
@@ -150,7 +147,6 @@ class TestResponse(object):
         assert r.read() == b''
 
     def test_chunked_decoding_gzip(self):
-        import zlib
         compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
         data = compress.compress(b'foo')
         data += compress.flush()
@@ -164,6 +160,53 @@ class TestResponse(object):
         assert r.read(2) == b'oo'
         assert r.read() == b''
         assert r.read() == b''
+
+    def test_decode_gzip_multi_member(self):
+        compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
+        data = compress.compress(b'foo')
+        data += compress.flush()
+        data = data * 3
+
+        fp = BytesIO(data)
+        r = HTTPResponse(fp, headers={'content-encoding': 'gzip'})
+
+        assert r.data == b'foofoofoo'
+
+    def test_decode_gzip_error(self):
+        fp = BytesIO(b'foo')
+        with pytest.raises(DecodeError):
+            HTTPResponse(fp, headers={'content-encoding': 'gzip'})
+
+    def test_decode_gzip_swallow_garbage(self):
+        # When data comes from multiple calls to read(), data after
+        # the first zlib error (here triggered by garbage) should be
+        # ignored.
+        compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
+        data = compress.compress(b'foo')
+        data += compress.flush()
+        data = data * 3 + b'foo'
+
+        fp = BytesIO(data)
+        r = HTTPResponse(
+            fp, headers={'content-encoding': 'gzip'}, preload_content=False)
+        ret = b''
+        for _ in range(100):
+            ret += r.read(1)
+            if r.closed:
+                break
+
+        assert ret == b'foofoofoo'
+
+    def test_chunked_decoding_gzip_swallow_garbage(self):
+        compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
+        data = compress.compress(b'foo')
+        data += compress.flush()
+        data = data * 3 + b'foo'
+
+        fp = BytesIO(data)
+        r = HTTPResponse(fp, headers={'content-encoding': 'gzip'})
+
+        assert r.data == b'foofoofoo'
 
     def test_body_blob(self):
         resp = HTTPResponse(b'foo')
@@ -292,7 +335,6 @@ class TestResponse(object):
             next(stream)
 
     def test_gzipped_streaming(self):
-        import zlib
         compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
         data = compress.compress(b'foo')
         data += compress.flush()
@@ -308,7 +350,6 @@ class TestResponse(object):
             next(stream)
 
     def test_gzipped_streaming_tell(self):
-        import zlib
         compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
         uncompressed_data = b'foo'
         data = compress.compress(uncompressed_data)
@@ -331,8 +372,6 @@ class TestResponse(object):
     def test_deflate_streaming_tell_intermediate_point(self):
         # Ensure that ``tell()`` returns the correct number of bytes when
         # part-way through streaming compressed content.
-        import zlib
-
         NUMBER_OF_READS = 10
 
         class MockCompressedDataReading(BytesIO):
@@ -382,7 +421,6 @@ class TestResponse(object):
         assert len(ZLIB_PAYLOAD) == end_of_stream
 
     def test_deflate_streaming(self):
-        import zlib
         data = zlib.compress(b'foo')
 
         fp = BytesIO(data)
@@ -396,7 +434,6 @@ class TestResponse(object):
             next(stream)
 
     def test_deflate2_streaming(self):
-        import zlib
         compress = zlib.compressobj(6, zlib.DEFLATED, -zlib.MAX_WBITS)
         data = compress.compress(b'foo')
         data += compress.flush()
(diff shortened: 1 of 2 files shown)


urllib3/response.py
```diff
@@ -63,7 +63,12 @@ def __getattr__(self, name):
     def decompress(self, data):
         if not data:
             return data
-        return self._obj.decompress(data)
+        ret = self._obj.decompress(data)
+        while self._obj.unused_data:
+            (self._obj, data) = (zlib.decompressobj(16 + zlib.MAX_WBITS),
+                                 self._obj.unused_data)
+            ret += self._obj.decompress(data)
```
Lukasa: So if there is junk at the end of a gzipped file for any reason, this change will break that. I wonder how much we care about it: should we aim to tolerate errors here?

isaacto: I suggest doing a simple test like this: Construct a file that has some junk at the end of the gzip file, put it to some web server, and see how the popular browsers behave for them.  If they report errors for these files, we can just ignore the issue and let it raise an exception.  In this case we are fixing the problem in these cases too: we should have reported an error but we haven't.

sigmavirus24: @isaacto great. Can you perform that test and report back?

isaacto: I tried firefox, which doesn't even handle a multi-member gzip-encoded HTML file.  I found that there is a bug report about a similar issue in Chrome, which didn't seem to have been resolved (see https://bugs.chromium.org/p/chromium/issues/detail?id=20884).

I tried the gzip program, and found that it can report trailing garbage in case of errors:

```
$ echo hello > f
$ gzip f
$ echo foo >> f.gz
$ gunzip f.gz

gzip: f.gz: decompression OK, trailing garbage ignored
```

That made me wonder a bit about whether we should add proper handling of gzip file in urllib3.  But since, unlike a regular browser, urllib3 automatically decompress downloaded files in its usual usage, I think the files should be properly handled.  (I discovered this problem when I tried downloading large files from Reuters, so we can expect that other users will sooner or later discover the same issue.)

If we really want to add trailing garbage tolerance to the code, we need quite a bit of machinery.  In particular, we need (1) to check whether an error from zlib is due to missing header, and (2) to check whether there is some content corrected decompressed.  If both answers are "yes", then we can tolerate the error.  To do so, we need to mark the decoder with a state which will ignore further content.

I wonder whether this complexity is desirable.

Lukasa: > But since, unlike a regular browser, urllib3 automatically decompress downloaded files in its usual usage

Well, this isn't quite true. urllib3 only decompresses files that are marked as having compression that is not part of the core representation (that is, with `Content-Encoding: gzip`), which should not be applied to files whose core representation is a compressed file.

Nonetheless, a few web servers do like doing that, so it can happen.

Given that it seems that browsers don't support multi-member gzip I think merely supporting that use-case is sufficient. I'm ok swallowing errors on garbage trailing data. It should be easy enough to do that: a single boolean flag is probably sufficient.

isaacto: See whether my new PR looks better.

test/test_response.py
```diff
@@ -165,6 +165,57 @@ def test_chunked_decoding_gzip(self):
...
+        data += compress.flush()
+        data = data * 3
+
+        fp = BytesIO(data)
+        r = HTTPResponse(fp, headers={'content-encoding': 'gzip'})
+
+        assert r.data == b'foofoofoo'
+
+    def test_decode_gzip_error(self):
+        import zlib
```
sigmavirus24: Do we need this line or the one that follows in this test?

test/test_response.py
```diff
@@ -165,6 +165,57 @@ def test_chunked_decoding_gzip(self):
         assert r.read() == b''
         assert r.read() == b''
 
+    def test_decode_gzip_multi_member(self):
+        import zlib
```
sigmavirus24: Let's import zlib at the module scope?

test/test_response.py
```diff
@@ -165,6 +165,57 @@ def test_chunked_decoding_gzip(self):
...
+                         preload_content=False)
+        ret = b''
+        for _ in range(100):
+            ret += r.read(1)
+            if r.closed:
+                break
+
+        assert ret == b'foofoofoo'
+
+    def test_chunked_decoding_gzip_swallow_garbage(self):
```
sigmavirus24: What do you mean by `chunked` here? I usually associate that with `Transfer-Encoding: chunked` so a docstring on this test would be helpful

test/test_response.py
isaacto: No problem.  This is imported many times in the module, do you want me to remove all the similar imports elsewhere in this commit?

test/test_response.py
isaacto: Ah I forgot to check whether this line is still useful after implementing it.  Will remove it.

test/test_response.py
isaacto: This is the same terminology in the deflate tests I think.  The difference is that here the decoding is done with preload_content=False, and after that read() with various lengths.  Here the main objective is to test the case where the read's continues after the first failure, which require such "chunked" decoding to be exercised.

test/test_response.py
haikuginger: @isaacto, yes please, unless doing so leads to issues - for example, if we're importing within methods to avoid circular dependencies. If that's the case, then the current arrangement is fine.

urllib3/response.py
```diff
@@ -56,14 +56,30 @@ class GzipDecoder(object):
...
-        return self._obj.decompress(data)
+        try:
+            ret = self._obj.decompress(data)
+            while self._obj.unused_data:
+                self._state = 'other'
+                (self._obj, data) = (zlib.decompressobj(16 + zlib.MAX_WBITS),
+                                     self._obj.unused_data)
+                ret += self._obj.decompress(data)
+            return ret
+        except Exception as ex:
```
haikuginger: Let's catch the specific class of exception we expect and not swallow everything. Always remember to keep try/except blocks as narrow as possible both in terms of lines-of-code covered and class-of-error covered. `try/except/else` is a useful construct to use.

urllib3/response.py
```diff
@@ -56,14 +56,30 @@ class GzipDecoder(object):
...
         self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
+        self._state = 'first'
 
     def __getattr__(self, name):
         return getattr(self._obj, name)
 
     def decompress(self, data):
+        ret = binary_type()
+        if self._state == 'error':
+            return ret
```
haikuginger: In the case of an error state, should we be returning nothing, or should we be raising an exception since there's no reasonable way to proceed?

urllib3/response.py
```diff
@@ -56,14 +56,30 @@ class GzipDecoder(object):
 
     def __init__(self):
         self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
+        self._state = 'first'
```
haikuginger: If we're going with an arbitrary set of states to move between, let's create an Enum higher up to list those states comprehensively, and then use the appropriate namespaced variables, rather than free-form strings.

urllib3/response.py
```diff
@@ -56,14 +56,30 @@ class GzipDecoder(object):
...
+        if self._state == 'error':
+            return ret
         if not data:
             return data
-        return self._obj.decompress(data)
+        try:
+            ret = self._obj.decompress(data)
+            while self._obj.unused_data:
+                self._state = 'other'
+                (self._obj, data) = (zlib.decompressobj(16 + zlib.MAX_WBITS),
```
haikuginger: There is no reason to shoehorn these items onto one line. Please separate them onto their own lines for better readability.

urllib3/response.py
```diff
@@ -56,14 +56,30 @@ class GzipDecoder(object):
...
 
     def decompress(self, data):
+        ret = binary_type()
+        if self._state == 'error':
+            return ret
         if not data:
             return data
-        return self._obj.decompress(data)
+        try:
+            ret = self._obj.decompress(data)
```
haikuginger: Can we formulate the loop in such a way that we don't have to "prime" it with this extra line?

urllib3/response.py
isaacto: I actually consider using string for these simple state changes more readable.  In languages like Java where we care about the last drop of performance we don't like such strings.  But in Python I tends to use plain strings for such three-state variables.

But since you like it the other way round, and since I don't really care, it is in the new commit.

urllib3/response.py
isaacto: It should be returning an empty string.  The whole point of all these handling is to allow members other than the first to trigger errors and we will ignore them, so as to support files that have garbage at the end.  If this is not required, all the error handling can be removed and we can let the zlib error to propagate out.  I've added some comments in the code to explain this.

urllib3/response.py
isaacto: Done.

urllib3/response.py
isaacto: Done.

urllib3/response.py
isaacto: Done.  But the exception catching is still essentially on the whole function.  They are for the two .decompress calls, either can cause the exception, and the handling is the same.  The other code has no potential to cause exceptions, so there is no harm doing it the current way.

haikuginger:
@isaacto, how does something like this look to you?

```python
    def decompress(self, data):
        ret = binary_type()
        if self._state == GzipDecoderState.SWALLOW_DATA or not data:
            # Return immediately if the decoder encountered an error or there's no data
            return ret
        while data:
            self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
            try:
                # Decompress the data we have and concatenate it to the return value
                ret += self._obj.decompress(data)
            except zlib.error:
                previous_state, self._state = self._state, GzipDecoderState.SWALLOW_DATA
                if ret and previous_state == GzipDecoderState.OTHER_MEMBERS:
                    # Swallow the error if we've pulled at least some data off the stream
                    return ret
                else:
                    # Re-raise the exception if we didn't get any data
                    raise
            else:
                # Get the next chunk of unused data
                data = self._obj.unused_data
                if data:
                    # If there's unused data on the decoder, set the state to indicate that
                    self._state = GzipDecoderState.OTHER_MEMBERS
        return ret
```

This narrows the scope of the try/except clause substantially and also removes the special case for the first decode attempt.

isaacto:
A single stream of data can come through multiple calls to decompress().

This means that the first self._obj assignment is wrong, it will clobber up the decompression in progress.

It also means that if previous_state == GzipDecoderState.OTHER_MEMBERS we need to return empty string no matter whether ret is empty.  It is perfectly legal if a previous call just completed in creating one member, and the next call generates an error without any new data.

But I get your ideas, and have made a new commit with it.

That said, I think we shouldn't be too serious with minimizing the scope of try-except.  Of course it is bad if we have large block of irrelevant code in an except block.  But if it is just a few statements of harmless statements I think it is fine.  It is one of the main benefit of exceptions over error code: make the code much easier to read.

test/test_response.py
haikuginger: @sigmavirus24, does that answer your question?

urllib3/response.py
```diff
@@ -52,18 +52,42 @@ def decompress(self, data):
...
-        if not data:
-            return data
-        return self._obj.decompress(data)
+        ret = binary_type()
+        if self._state == GzipDecoderState.SWALLOW_DATA or not data:
+            return ret
+        while True:
+            try:
+                ret += self._obj.decompress(data)
+            except zlib.error as ex:
```
haikuginger: Nit: We don't use `ex` anywhere, so there's no benefit to assigning a variable. `except zlib.error:` should be fine.

isaacto: Done.

test/test_response.py
sigmavirus24: 👍

PR #1292, opened 2017-12-11
Simplify tox configuration

jdufresne:
tox sets good defaults for the `basepython` value. Simply use that instead of respecifying it. Helps to slightly simplify the configuration by removing repetitive boilerplate.

For details, see tox documentation:

https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings

> tox provides good defaults for basepython setting, so the above ini-file can be further reduced by omitting the basepython setting.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,13 +2,6 @@
 envlist = flake8-py3, py26, py27, py34, py35, py36, pypy
 
 [testenv]
-basepython =
-    py26: {env:TOXPY26:python2.6}
-    py27: {env:TOXPY27:python2.7}
-    py34: {env:TOXPY34:python3.4}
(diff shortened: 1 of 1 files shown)


jdufresne:
Rebased.

haikuginger:
Thanks for the contribution, @jdufresne! I appreciate it!

itziakos:
The tox defaults for windows will always point to the 32 bit python. These changes in tox.ini result in running tests on 32 bit python in appveyor instead of using the 64 bit python that is in the path. (see https://ci.appveyor.com/project/shazow/urllib3/build/1.0.440/job/vyg6v7ybcqx71rym#L45)

PR #1293, opened 2017-12-11
Fix uses of deprecated assertEquals() to use assertEqual() instead

jdufresne:
When running tests with Python warnings enabled, fixes warnings of the form:

```
DeprecationWarning: Please use assertEqual instead.
```

For a full list of deprecated test functions, see:

https://docs.python.org/3/library/unittest.html#deprecated-aliases

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -1000,7 +1000,7 @@ class TestRetryPoolSize(HTTPDummyServerTestCase):
 
     def test_pool_size_retry(self):
         self.pool.urlopen('GET', '/not_found', preload_content=False)
-        self.assertEquals(self.pool.num_connections, 1)
+        assert self.pool.num_connections == 1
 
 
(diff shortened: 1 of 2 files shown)


sigmavirus24:
It would be better to rewrite this to use pytest style assertions since we're touching this.

haikuginger:
I agree with @sigmavirus24. Should be a relatively simple change that I'd have no issues accepting.

jdufresne:
I've updated the PR such that `assertEquals()` is now using plain old `assert` instead.

Is this what you had in mind? If you were asking for something else, please let me know.

haikuginger:
Perfect! Once you rebase on master, we should be set.

jdufresne:
Rebased.

haikuginger:
@Lukasa, I don't seem to have rebuild access on Appveyor yet - could you restart the failed build? Looks spurious.

Lukasa:
When appveyor tests fail you can’t do anything but close/open.

haikuginger:
Thanks for the contribution, @jdufresne!

PR #1294, opened 2017-12-11
Use pyenv to install all versions of Python for macOS; update Python3 versions

haikuginger:
Currently, our macOS tests for Python 2.7 expect a version of Python 2.7 to already be installed, and then install pip. However, for whatever reason, this currently appears to be failing. To fix this issue and avoid the same class of issue going forward (where different versions behave differently under testing), this change updates the test suite to explicitly install Python 2.7 (including pip) with pyenv, as we already do for other versions.

This change also updates the versions of Python 3 to test with macOS to their respective newest revisions.

### DIFF
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -22,24 +22,24 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
             pyenv global 2.6.9
             ;;
         py27)
-            curl -O https://bootstrap.pypa.io/get-pip.py
-            python get-pip.py --user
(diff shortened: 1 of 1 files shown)


haikuginger:
I'd also be interested in discussion on whether we should keep macOS builds on Travis set to be required. The delay seems substantial, and I've had two ephemeral failures just on this one PR, just today. That really hurts turnaround time.

Lukasa:
Yeah, so currently they do test some function that is not tested elsewhere. That suggests they should be kept running.

PR #1299, opened 2017-12-20
Don't send IP address as SNI.

jichao-zhang:
In RFC-6066 section-3, it was said that SNI is for host name not for IP
address. This code closes shazow/urllib3#1298 .

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -25,6 +25,7 @@ from urllib3.util.ssl_ import (
     resolve_ssl_version,
     ssl_wrap_socket,
     _const_compare_digest_backport,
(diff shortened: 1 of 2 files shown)


tiran:
This PR is a duplicate of #1287. As far as I remember urllib3 internals, both PRs have a severe security bug. They no longer verify that the IP address matches the certificate.

I just created a PR that will fix the issue in CPython, https://github.com/python/cpython/pull/4938

jichao-zhang:
@tiran Thank you very much. We will wait for new Python version which include your PR: python/cpython#4938.

PR #1304, opened 2017-12-31
Move RECENT_DATE to 2017-06-30

### LINKED ISSUE #1303: RECENT_DATE needs to be bumped
Test suite started failing, as RECENT_DATE is not recent enough:

```
[   62s]     def test_recent_date(self):
[   62s]         # This test is to make sure that the RECENT_DATE value
[   62s]         # doesn't get too far behind what the current date is.
[   62s]         # When this test fails update urllib3.connection.RECENT_DATE
[   62s]         # according to the rules defined in that file.
[   62s]         two_years = datetime.timedelta(days=365 * 2)
[   62s] >       assert RECENT_DATE > (datetime.datetime.today() - two_years).date()
[   62s] E       AssertionError: assert datetime.date(2016, 1, 1) > datetime.date(2016, 1, 1)
[   62s] E        +  where datetime.date(2016, 1, 1) = <built-in method date of datetime.datetime object at 0x7fb6899be198>()
[   62s] E        +    where <built-in method date of datetime.datetime object at 0x7fb6899be198> = (datetime.datetime(2017, 12, 31, 10, 16, 14, 472906) - datetime.timedelta(730)).date
[   62s] E        +      where datetime.datetime(2017, 12, 31, 10, 16, 14, 472906) = <built-in method today of type object at 0x7fb68e948d20>()
[   62s] E        +        where <built-in method today of type object at 0x7fb68e948d20> = <class 'datetime.datetime'>.today
[   62s] E        +          where <class 'datetime.datetime'> = datetime.datetime
```

DimStar77:
The test suite expects the current date to be no more than two years in the future
of RECENT_DATE, which just serves as a reference point.

Also clarify the comment about how to update RECENT_DATE

Fixes #1303

### DIFF
--- a/urllib3/connection.py
+++ b/urllib3/connection.py
@@ -56,10 +56,11 @@ port_by_scheme = {
     'https': 443,
 }
 
-# When updating RECENT_DATE, move it to
-# within two years of the current date, and no
-# earlier than 6 months ago.
-RECENT_DATE = datetime.date(2016, 1, 1)
+# When updating RECENT_DATE, move it to within two years of the current date,
+# and not less than 6 months ago.
+# Example: if Today is 2018-01-01, then RECENT_DATE should be any date on or
+# after 2016-01-01 (today - 2 years) AND before 2017-07-01 (today - 6 months)
+RECENT_DATE = datetime.date(2017, 6, 30)
 
 
 class DummyConnection(object):
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -246,5 +246,8 @@ In chronological order:
 * Akamai (through Jesse Shapiro) <jshapiro@akamai.com>
   * Ongoing maintenance
 
+* Dominique Leuenberger <dimstar@opensuse.org>
+  * Minor fixes in the test suite
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]


haikuginger:
@DimStar77, thanks for the contribution! This looks good to me; can you make sure that you're included in the `AUTHORS` file?

urllib3/connection.py
```diff
@@ -57,9 +57,9 @@ class ConnectionError(Exception):
 }
 
 # When updating RECENT_DATE, move it to
-# within two years of the current date, and no
-# earlier than 6 months ago.
-RECENT_DATE = datetime.date(2016, 1, 1)
+# within two years of the current date, and not
+# less than 6 months ago.
```
sigmavirus24: Should we add an example just for ultimate clarity, or a formula for generating this? For example

```
If today is 2018-01-01 then the RECENT_DATE shouldn't be any date after 2017-07-01. Roughly `datetime.date.today() - datetime.timedelta(days=6 * 31)`
```

DimStar77: > If today is 2018-01-01 then the RECENT_DATE shouldn't be any date after 2017-07-01. Roughly `datetime.date.today() - datetime.timedelta(days=6 * 31)`

That's not my understanding, but rather: ```if Today is 2018-01-01, then RECENT_DATE SHOULD be any date on or after 2016-01-01 (today - 2 years) AND before 2017-07-01 (today - 6 months)```

sigmavirus24: You're correct. I was more hoping to clarify the latter half because I still don't think it's entirely clear :) and what you have there looks actually perfect.

DimStar77: Ok, I'll update the comment in the file with this example - then I think we should be clear for next time around we have to change it in about 18 months.

PR #1308, opened 2018-01-09
Begin testing against Python 3.7

SethMichaelLarson:
Since Python 3.7 is starting to see alpha releases we should probably start testing against the `3.7-dev` stream in order to start catching issues.
It's marked as an `allow_failure` in Travis so that failures won't stop existing changes from being implemented.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,11 @@ matrix:
     - language: generic
       os: osx
       env: TOXENV=py36
+    - language: generic
+      os: osx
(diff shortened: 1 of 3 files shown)


haikuginger:
@SethMichaelLarson, do we want to test against 3.7 on macOS too?

SethMichaelLarson:
Probably, I'll add that as well.

SethMichaelLarson:
Restarting Travis because the last commit failed one job with a cryptic message.

PR #1309, opened 2018-01-10
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(name='urllib3',
                 'urllib3.contrib._securetransport', 'urllib3.util',
                 ],
       requires=[],
+      python_requires=">=2.6, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, <4",
       tests_require=[
           # These are a less-specific subset of dev-requirements.txt, for the
           # convenience of distro package maintainers.


haikuginger:
This seems good to me; let's also have the `<4` specifier suggested in that first link.

jdufresne:
Done. Thanks.

sigmavirus24:
It seems like Travis doesn't know we exist any longer: https://travis-ci.org/shazow/urllib3 (404)

SethMichaelLarson:
Travis is back in business! :)

sigmavirus24:
It used to be that I could restart individual builds on Travis but apparently that is no more? So I'm going to close this and re-open it to trigger a new build.

haikuginger:
Thanks for the contribution, @jdufresne! It's appreciated!

cj81499:
> This seems good to me; let's also have the `<4` specifier suggested in that first link.

I no longer see this suggestion in the link (content relocated to https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires), and its inclusion in this change caused me some headache today.

I use `poetry` to manage dependencies for https://github.com/cj81499/advent-of-code, and today I got a dependabot alert for urllib3 (CVE-2018-20060).

I tried a `poetry update`, but poetry did not update urllib. `poetry add urllib3@latest` revealed the problem.

```command
> poetry add urllib3@latest
Using version ^1.26.12 for urllib3

Updating dependencies
Resolving dependencies... (0.0s)

The current project's Python requirement (>=3.9) is not compatible with some of the required packages Python requirement:
  - urllib3 requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4, so it will not be satisfied for Python >=4

Because urllib3 (1.26.12) requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, <4
 and no versions of urllib3 match >1.26.12,<2.0.0, urllib3 is forbidden.
So, because aoc-cj depends on urllib3 (^1.26.12), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    
    For urllib3, a possible solution would be to set the `python` property to ">=3.9,<4"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

As a result of this change, I've been quietly stuck on `urllib3 v1.22` without even realizing it.

As suggested by poetry, I've changed _my_ python requirement from ">=3.9" to ">=3.9,<4" to be able to update, but I'm not sure there's a good reason to restrict the supported python versions to `<4`. Like I mentioned earlier, it seems this is no longer recommended by the Python docs.

In addition, since it didn't set a `python_requires`, a (future) Python 4 user might try to install urllib3 and get `v1.22`(!), which seems like pretty undesirable behavior to me.

sethmlarson:
@cj81499  Could you open a new separate issue with this info?

cj81499:
I'm about to log off for the night, but I'll make a note to do that tomorrow. Thanks for the quick response!

cj81499:
@sethmlarson #2761 😄

PR #1320, opened 2018-01-23
Add User-Agent header

### LINKED ISSUE #1296: No User-Agent header by default
It appears that, by default, urllib3 does not send a `User-Agent` header.

By a plain reading of [RFC-7231](https://tools.ietf.org/html/rfc7231#section-5.5.3), it appears that user agents SHOULD generally set such a header unless specifically overridden or suppressed:

> A user agent SHOULD send a User-Agent field in each request unless specifically configured not to do so.

I couldn't find prior discussion on this, so it does seem worth talking about. Should we send a User-Agent header by default, and if so, what should it be?

CC @sigmavirus24 in particular for an opinion on the intentions of the RFC. Are we a user agent as described in the document? Or, since we expect to be used for the network-level connectivity of a larger process or library, are we exempt from the requirement?

Madhu94:
Trying to close #1296 

Is there a better way to get the version of a module from within?

Thank you!

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -781,6 +781,7 @@ class TestProxyManager(SocketDummyServerTestCase):
                              b'Accept: */*',
                              b'',
                              b'',
+                             b'User-Agent: python-urllib3/dev'
                          ]))
 
     def test_headers(self):
--- a/urllib3/request.py
+++ b/urllib3/request.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 from .filepost import encode_multipart_formdata
 from .packages.six.moves.urllib.parse import urlencode
 
+import sys
 
 __all__ = ['RequestMethods']
 
@@ -40,6 +41,12 @@ class RequestMethods(object):
 
     def __init__(self, headers=None):
         self.headers = headers or {}
+        try: 
+            version = sys.modules['urllib3'].__version__
+            self.headers.update({'User-Agent': 'python-urllib3/%s' % version})
+        except KeyError:
+            pass
+        
 
     def urlopen(self, method, url, body=None, headers=None,
                 encode_multipart=True, multipart_boundary=None,


urllib3/request.py
```diff
@@ -40,6 +41,12 @@ class RequestMethods(object):
 
     def __init__(self, headers=None):
         self.headers = headers or {}
+        try: 
+            version = sys.modules['urllib3'].__version__
+            self.headers.update({'User-Agent': 'python-urllib3/%s' % version})
```
SethMichaelLarson: We would want to do a `setdefault()` here instead of `update()` so that if a user sets a `User-Agent` header ours doesn't take precedence. We should have a test case for this as well.

urllib3/request.py
```diff
@@ -40,6 +41,12 @@ class RequestMethods(object):
 
     def __init__(self, headers=None):
         self.headers = headers or {}
+        try: 
+            version = sys.modules['urllib3'].__version__
```
SethMichaelLarson: We should probably just import the `__version__` information in our header instead of inspecting `sys.modules`.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -781,6 +781,7 @@ def echo_socket_handler(listener):
                              b'Accept: */*',
                              b'',
                              b'',
+                             b'User-Agent: python-urllib3/dev'
```
SethMichaelLarson: This testcase will fail on any tagged version. Need to update this test to pass with any setting of `__version__`.

urllib3/request.py
haikuginger: We actually don't want to be here at all; the object-default headers don't get used if the user passes a set of headers in. A good place to set it might be in `extra_kw` a bit lower down.

test/with_dummyserver/test_socketlevel.py
haikuginger: Good catch. Yeah, we need to get the version dynamically - since we're working outside urllib3, we should be able to do `from urllib3 import __version__ as package_version` and use that variable.

urllib3/request.py
haikuginger: I'm not sure if that'll work; we might need to pull `__version__` into a `constants.py`, and import that from both `__init__.py` and `request.py` (or wherever else we decide to put this code).

urllib3/request.py
haikuginger: Actually, check out `make_headers` in `util/request.py` as a potential very good spot for this.

Madhu94:
Thanks for the feedback @haikuginger @SethMichaelLarson.

>  we might need to pull __version__ into a constants.py

This was what I was hoping for! I'll do this, add tests and push again.

haikuginger:
@Lukasa @shazow does that move of the version datum make sense to you?

SethMichaelLarson:
@haikuginger We'll have to change the logic to parse the `__version__` in the `setup.py` file as well.

haikuginger:
@SethMichaelLarson, good point, as well as the deploy script.

shazow:
The whole "one place to keep the version and use it everywhere else" thing in Python was in a pretty poor state back in my day (hence the hack in setup.py). I'm honestly not sure what the Best Practice is these days. Would be curious to do a survey of modern libraries to see what they do.

SethMichaelLarson:
@shazow I don't think much has changed beyond using an `__about__.py` file to store all dunder-metadata.

SethMichaelLarson:
Hey @Madhu94! Are you still willing to make these changes to fix this issue?

jonparrott:
This Pull Request has been inactive for several weeks while waiting for the author to respond, so I'm going to close it to keep things tidy. We are more than happy to consider this so please feel free to let us know if you want to re-open.

Thank you!

Issue #1323, opened 2018-02-01
RFC: what do you think of this branch adding async support to urllib3?


njsmith:
We (mostly @pquentin and I) [have been working on a proof of concept for adding pluggable async support to urllib3](https://github.com/python-trio/urllib3/issues/1), with the hope of eventually getting this into the upstream urllib3. It's reached the point where there's still lots of missing bits, but there's an end-to-end demo working and we don't see any major obstacles to getting everything else working, so I wanted to start getting feedback from the urllib3 maintainers about whether this looks like something you'd be interested in eventually merging, and what it would take to get there.

# Demo

So: hi! Check it out – a single py2.py3 wheel that keeps the classic synchronous API, and on python 3.6+ it can also run in async mode on both Trio and Twisted:

```
# Demo from commit 1ca67ee53e18f823d0cb in the python-trio/urllib3 bleach-spike branch
$ curl -O https://vorpus.org/~njs/tmp/async-urllib3-demo.zip
$ unzip async-urllib3-demo.zip
$ cd async-urllib3-demo
$ ls
sync-demo.py
async-demo.py
urllib3-2.0.dev0+bleach.spike.proof.of.concept.dont.use-py2.py3-none-any.whl

$ virtualenv -p python3.6 py36-venv
$ py36-venv/bin/pip install trio twisted[tls] urllib3-2.0.dev0+bleach.spike.proof.of.concept.dont.use-py2.py3-none-any.whl

$ py36-venv/bin/python sync-demo.py
--- urllib3 using synchronous sockets ---
URL: http://httpbin.org/uuid
Status: 200
Data: b'{\n  "uuid": "a2c28245-47b8-4a50-b64c-da09d27bf626"\n}\n'

$ py36-venv/bin/python async-demo.py
--- urllib3 using Trio ---
URL: http://httpbin.org/uuid
Status: 200
Data: b'{\n  "uuid": "dab50c1a-1b20-483f-903e-fe74494629f2"\n}\n'

--- urllib3 using Twisted ---
URL: http://httpbin.org/uuid
Status: 200
Data: b'{\n  "uuid": "72196b66-7caa-40dd-9c7c-af65bb4f7fb6"\n}\n'

$ virtualenv -p python2 py2-venv
$ py2-venv/bin/pip install urllib3-2.0.dev0+bleach.spike.proof.of.concept.dont.use-py2.py3-none-any.whl
$ py2-venv/bin/python sync-demo.py
[...]

PR #1335, opened 2018-03-03
Async support as a PR for the v2 branch

njsmith:
This is the branch discussed in #1323, proposed as a PR to the v2 branch. Discussion there seems to have slowed down, and my best guess at the best way forward is still that we should merge this branch into v2 and then continue iterating on it via PRs in the main repo, so that the regular maintainers can see what's going on as it happens. (Also maybe this will make @sigmavirus24 [happy](https://github.com/shazow/urllib3/issues/1323#issuecomment-365592730).) So my proposal is (a) merge this, (b) give me permissions to the repo (which I will only use for the v2 branch unless told otherwise), (c) profit? Totally open to other suggestions too.

The diff is difficult to review, because of file movements that git is not clever enough to handle. (`git blame` still works, but `git diff` shows full-file delete/re-adds.) Also, there's been a merge from master, which dragged in a lot of the random changes below (e.g. to `CHANGES.rst`).

Fortunately the merge from master is near the end of the branch, so we can make a diff of the branch without that, and it contains most of the semantic changes here. This diff is still not perfect b/c of the renames, but it's much more readable, and it [can be viewed by following this finely crafted link](https://github.com/njsmith/urllib3/compare/v2...be5165ff59354e93c0d22fbf496bfb51e561252f).

In general this is a work in progress that still requires significant iteration to finish (which is why it's only being proposed to merge into a branch, not into `master`), and in the end the criterion for "is it done?" is going to involve getting the test suite passing, testing against major downstream libraries, etc. Instead of stressing about doing a detailed review now, my suggestion would be to review the high-level architecture and some of the major pieces of new code. Probably the main bits are:

* The [`urllib3/_async` directory](https://github.com/njsmith/urllib3/tree/bleach-spike/urllib3/_async) is mostly similar to the corresponding code currently in urllib3, but with `async`/`await` annotations added, *except* that `connection.py` is new code based on h11 and has seen significant changes on top of @Lukasa's v2 work. This is the code that gets transformed into `urllib3/_sync` by [`setup.py`](https://github.com/njsmith/urllib3/blob/bleach-spike/setup.py).

* The [`urllib3/_backends` directory](https://github.com/njsmith/urllib3/tree/bleach-spike/urllib3/_backends) has the different I/O backends. They all implement the same API (a private one, so we can change it later if we want to), and choosing the right one allows urllib3 to run on different I/O frameworks.

* There are several files like [`urllib3/connectionpool.py`](https://github.com/njsmith/urllib3/blob/bleach-spike/urllib3/connectionpool.py) which are replaced by backwards-compatibility shims that just re-export symbols from the corresponding file in `urllib3/_sync/`.

* There are lots of `@pytest.mark.xfail` and `@pytest.mark.skip` decorators added to broken tests; the idea is that by getting the test suite passing *at all*, we can start writing PRs that incrementally remove these while keeping the test suite passing (so we at least know that once something is working, it stays working).

Closes: #1323


njsmith:
Oh hmm, those CI tests are going to blow up because we haven't either got tox quite set up yet or tweaked travis/appveyor to use the test running script yet.

njsmith:
Ping.

shazow:
@njsmith Hey Nathaniel, thank you so much for all your hard work on this. There is a lot to process, and we're in the middle of some maintainer-logistics administrative overhaul, so please give us another week or two before we can get urllib3 running in full gear again. :) Please do ping again if we don't get back to this in a couple of weeks. <3

pquentin:
First codecov report, I guess that's progress. 🙂

pquentin:
It has been clear for some time now that the future of our work here isn't going to replace the v2 branch or happen in this repository, so let's close this pull request!

I'm still planning to keep the differences to a minimum by sharing fixes between the two branches as long as possible, though.

PR #1345, opened 2018-03-25
Closed HTTPResponse.read_chunked() returns no chunks instead of exception

### LINKED ISSUE #1088: If the response is complete, stream/read_chunked should short-circuit return the empty string.
Originally seen in kennethreitz/requests#3807.

Basically, read_chunked should check whether `_fp` is `None`. If it is, it should immediately return the empty byte string rather than do any processing.

Put another way, this shouldn't blow up:

```python
import urllib3

http = urllib3.PoolManager()
resp = http.request('GET', 'http://http2bin.org/stream/100')
resp.read()
next(resp.read_chunked())
```

I recommend someone who wants to help out pick this issue up, it should be easily fixed and tested.

SethMichaelLarson:
Brings the behavior more in-line with `HTTPResponse.read()` and returns no chunks when the response object is already closed. Closes #1088.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -706,6 +706,16 @@ class TestConnectionPool(HTTPDummyServerTestCase):
         with pytest.raises(StopIteration):
             next(response.read_chunked())
 
+    def test_read_chunked_on_closed_response(self):
+        response = self.pool.request(
+            'GET',
+            '/chunked',
+            preload_content=False
+        )
+        response.close()
+        with pytest.raises(StopIteration):
+            next(response.read_chunked())
+
     def test_chunked_gzip(self):
         response = self.pool.request(
                 'GET',
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -5,6 +5,7 @@ import sys
 import unittest
 import time
 import warnings
+import pytest
 
 import mock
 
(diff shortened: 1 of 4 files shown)


test/with_dummyserver/test_connectionpool.py
```diff
@@ -695,6 +696,16 @@ def test_stream_keepalive(self):
...
         self.assertEqual(self.pool.num_requests, x)
 
+    def test_read_chunked_short_circuit(self):
+        response = self.pool.request(
+            'GET',
+            '/chunked',
+            preload_content=False
+        )
+        response.read()
+        with pytest.raises(StopIteration):
```
haikuginger: Wouldn't we want to validate that this _doesn't_ raise? Or would we want more of a test that explicitly closes and then tries to read?

SethMichaelLarson: I can add a test for closing as well. I lifted this test case from the original issue attached.

SethMichaelLarson: read_chunked() is a generator so raising StopIteration is the equivalent of an empty string read.

haikuginger: Gotcha. Yeah that's my bad; didn't think about it properly.

SethMichaelLarson: Oh no worries! :) I'll add the additional test case.

SethMichaelLarson:
If this looks good and no one has review comments I'll merge this this afternoon.

PR #1346, opened 2018-03-26
Remove Authorization header when redirecting cross-host

### LINKED ISSUE #1316: Auth header remains during redirects
Requests does it: https://github.com/request/request/pull/1184

SethMichaelLarson:
This change makes the default behavior of an HTTP redirect when the `Location` specifies a different host than the original request to remove the `Authentication` header if present. This behavior can be turned off by using `Retry(forward_auth_headers_across_hosts=True)` which defaults to `False`. Closes #1316.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,10 @@ dev (master)
 
 * Lazily load `uuid` to boost performance on imports (Pull #1270)
 
+* Stop forwarding the ``Authentication`` HTTP header on a cross-host
+  redirect unless disabled by setting ``forward_auth_headers_across_hosts``
+  to True on ``Retry`` objects. (Issue #1316)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,9 +18,9 @@ dev (master)
 
 * Lazily load `uuid` to boost performance on imports (Pull #1270)
 
-* Stop forwarding the ``Authentication`` HTTP header on a cross-host
-  redirect unless disabled by setting ``forward_auth_headers_across_hosts``
-  to True on ``Retry`` objects. (Issue #1316)
+* Allow providing a list of headers to strip from requests when redirecting
+  to a different host. Defaults to the ``Authorization`` header. Different
+  headers can be set via ``Retry.remove_headers_on_redirect``. (Issue #1316)
 
 * ... [Short description of non-trivial change.] (Issue #)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,10 @@ Changes
 dev (master)
 ------------
 
+* Allow providing a list of headers to strip from requests when redirecting
+  to a different host. Defaults to the ``Authorization`` header. Different
+  headers can be set via ``Retry.remove_headers_on_redirect``. (Issue #1316)
+
 * Fix ``util.selectors._fileobj_to_fd`` to accept ``long`` (Issue #1247).
 
 * Dropped Python 3.3 support. (Pull #1242)
@@ -17,10 +21,6 @@ dev (master)
 * Add the port to the connectionpool connect print (Pull #1251)
 
 * Lazily load `uuid` to boost performance on imports (Pull #1270)
-
-* Allow providing a list of headers to strip from requests when redirecting
-  to a different host. Defaults to the ``Authorization`` header. Different
-  headers can be set via ``Retry.remove_headers_on_redirect``. (Issue #1316)
 
 * ... [Short description of non-trivial change.] (Issue #)
 
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -312,8 +312,11 @@ class PoolManager(RequestMethods):
 
         kw['assert_same_host'] = False
         kw['redirect'] = False
+
         if 'headers' not in kw:
-            kw['headers'] = self.headers
+            kw['headers'] = headers = self.headers
+        else:
+            headers = kw['headers']
 
         if self.proxy is not None and u.scheme == "http":
             response = conn.urlopen(method, url, **kw)
@@ -335,6 +338,13 @@ class PoolManager(RequestMethods):
         if not isinstance(retries, Retry):
             retries = Retry.from_int(retries, redirect=redirect)
 
+        # Strip the Authentication header if redirecting to a new host
+        # and we don't want to forward that header across hosts.
+        if (not retries.forward_auth_headers_across_hosts
+                and 'Authentication' in headers
+                and not conn.is_same_host(redirect_location)):
+            headers.pop('Authentication')
+
         try:
             retries = retries.increment(method, url, response=response, _pool=conn)
         except MaxRetryError:
@@ -344,6 +354,7 @@ class PoolManager(RequestMethods):
 
         kw['retries'] = retries
         kw['redirect'] = redirect
+        kw['headers'] = headers
 
         log.info("Redirecting %s -> %s", url, redirect_location)
         return self.urlopen(method, redirect_location, **kw)
--- a/urllib3/poolmanager.py
+++ b/urllib3/poolmanager.py
@@ -338,12 +338,12 @@ class PoolManager(RequestMethods):
         if not isinstance(retries, Retry):
             retries = Retry.from_int(retries, redirect=redirect)
 
-        # Strip the Authentication header if redirecting to a new host
-        # and we don't want to forward that header across hosts.
-        if (not retries.forward_auth_headers_across_hosts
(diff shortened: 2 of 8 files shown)


SethMichaelLarson:
Hah, I should read MDN instead of assuming header names.

And I was definitely thinking about allowing user-defined header stripping. That sounds very useful. I'll add it using whatever logic is decided to be best.

Also quick question: Is PoolManager the only place for this sort of logic or is me adding this to ConnectionPool also necessary?

haikuginger:
A ConnectionPool should, IIRC, only ever connect to a single host, so it makes sense for this logic to be either in PoolManager or Redirect.

jonparrott:
This is a good start, but I would ask for a slightly more robust implementation (I'm happy to help with that as well).

1. `PoolManager` is the only thing that does cross-host requests, so the logic should live (mostly) there.
2. `Retry` configures redirect handling, so the configuration should live there.
3. Instead of stripping just one hard-coded header, I'd prefer if we had a `remove_headers_on_redirect` which can be set to a `Sequence`. By default, this will just be `{'Authorization'}`.

Probably useful to reference https://curl.haxx.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html as that's where this behavior comes from. It is not formally specified in any RFC.

SethMichaelLarson:
Still trying to shake transient macOS SSL tests failing.

shazow:
@SethMichaelLarson You might have permissions to do this directly on https://travis-ci.org/shazow/urllib3/ now that you're a collaborator, but I'm not entirely sure how TravisCI permissions work these days.

SethMichaelLarson:
@shazow GitHub still says I'm a Contributor, not Collaborator. :)

pquentin:
(While it adds noise, the good thing about closing/opening is that the old builds are kept, which is why it's recommended by https://github.com/python-trio/trio/issues/200.)

shazow:
@pquentin Interesting point, thanks!

@SethMichaelLarson Fixed.

Anyways, carry on as you will. :)

SethMichaelLarson:
Of course everything passes right after being given Collaborator. :)

urllib3/poolmanager.py
```diff
@@ -335,6 +338,13 @@ def urlopen(self, method, url, redirect=True, **kw):
         if not isinstance(retries, Retry):
             retries = Retry.from_int(retries, redirect=redirect)
 
+        # Strip headers marked as unsafe to forward to the redirected location.
+        if (retries.remove_headers_on_redirect
+                and not conn.is_same_host(redirect_location)):
+            for header in retries.remove_headers_on_redirect:
+                if header in headers:
```
jonparrott: you can just use `headers.pop(header, None)`

urllib3/poolmanager.py
```diff
@@ -312,8 +312,11 @@ def urlopen(self, method, url, redirect=True, **kw):
 
         kw['assert_same_host'] = False
         kw['redirect'] = False
+
         if 'headers' not in kw:
-            kw['headers'] = self.headers
+            kw['headers'] = headers = self.headers
```
jonparrott: since you set `kw['headers']` below, do you need to set it here?

Also, you should probably copy if using `self.headers` so you don't modify that dict.

jonparrott:
This is mostly looking good, I just have one major concern:

This is a *very* big behavioral change that's likely to break some users. Some users won't be able to directly control how PoolManager is constructed. Should we consider some sort of environment variable to change this back to the old behavior?

urllib3/poolmanager.py
SethMichaelLarson: Good catch, I've implemented this.

SethMichaelLarson:
Failing PyOpenSSL test on AppVeyor.

SethMichaelLarson:
I'm +0 on an environment variable to toggle this behavior, others will have a better history of experience with this type of feature. :)

shazow:
In my day, we erred on the side of breaking things for the sake of better security defaults. (Seems just about the most justifiable reason for breaking backwards compatibility.)

Do other projects do env vars? I feel like by the time you looked up how to fix it via env var, it would be the same amount of work to just fix the problem properly.

Would definitely worth putting this change at the top of the CHANGES list for the next release.

What do @haikuginger @Lukasa think?

CHANGES.rst
```diff
@@ -18,6 +18,10 @@ dev (master)
 
 * Lazily load `uuid` to boost performance on imports (Pull #1270)
 
+* Allow providing a list of headers to strip from requests when redirecting
```
shazow: Maybe worth adding another entry at the top of the CHANGES list that explicitly mentions the breaking change.

urllib3/util/retry.py
```diff
@@ -152,7 +156,8 @@ class Retry(object):
     def __init__(self, total=10, connect=None, read=None, redirect=None, status=None,
                  method_whitelist=DEFAULT_METHOD_WHITELIST, status_forcelist=None,
                  backoff_factor=0, raise_on_redirect=True, raise_on_status=True,
-                 history=None, respect_retry_after_header=True):
+                 history=None, respect_retry_after_header=True,
+                 remove_headers_on_redirect=None):
```
shazow: Might be nicer if this were a constant value, like `remove_headers_on_redirect=DEFAULT_REDIRECT_HEADERS_BLACKLIST` or somesuch.

CHANGES.rst
SethMichaelLarson: I can add this, I also want to make mention to that this only occurs when the hostname in the redirect is different than the original requests hostname.

urllib3/poolmanager.py
```diff
@@ -335,6 +336,12 @@ def urlopen(self, method, url, redirect=True, **kw):
         if not isinstance(retries, Retry):
             retries = Retry.from_int(retries, redirect=redirect)
 
+        # Strip headers marked as unsafe to forward to the redirected location.
+        if (retries.remove_headers_on_redirect
```
shazow: Optional nit: Would be equivalent to have

```python
        if not conn.is_same_host(redirect_location)):
            for header in retries.remove_headers_on_redirect:
                kw['headers'].pop(header, None)
```

SethMichaelLarson: Only reason I chose to do this is because of the `TODO: socket.gethostname()` within `ConnectionPool.is_same_host()` in order to short-circuit a potential future network call if not needed. :)

urllib3/util/retry.py
SethMichaelLarson: Sure thing, that would be an easy addition.

urllib3/poolmanager.py
shazow: Good point. In that case, a comment would be wise so a future contributor doesn't simplify it. :)

urllib3/util/retry.py
```diff
@@ -19,6 +19,10 @@
 
 log = logging.getLogger(__name__)
 
+
+DEFAULT_REDIRECT_HEADERS_BLACKLIST = ['Authorization']
```
shazow: Would it be better to put this under the `Retry` object next to `RETRY_AFTER_STATUS_CODES` etc as a frozenset too?

(Sorry I should have mentioned that before)

SethMichaelLarson: I was looking for globals and didn't find them! Didn't think to look at the bottom of the module. :)

jonparrott:
> In my day, we erred on the side of breaking things for the sake of better security defaults. (Seems just about the most justifiable reason for breaking backwards compatibility.)

That's reasonable, I just want us to treat it with the weight that it carries. :)

> Do other projects do env vars? I feel like by the time you looked up how to fix it via env var, it would be the same amount of work to just fix the problem properly.

The problem is that urllib3 is used a lot *indirectly*, so users may not be in control of what constructs the `PoolManager/Redirect`. However, making the default list a module-level constant gives us some recourse for users in that situation as they *can* monkey patch the list if absolutely necessary. 

> Definitely worth putting this change at the top of the CHANGES list for the next release.

Agreed.

SethMichaelLarson:
@jonparrott So once Travis macOS builders catch up this looks good to you in its current state?

urllib3/util/retry.py
jonparrott: Off, I don't like `None` meaning `default` here. I assume you did this to avoid making the default parameter a mutable value - in which case, you can make the constant a `frozenset`.

SethMichaelLarson: Is flake8 smart enough to know `frozenset` isn't a mutable value/doesn't report mutable default value for `frozenset`? I honestly don't know.

SethMichaelLarson: Oh, duh, it's used above this parameter as a default so of course it works! I'll make this change.

PR #1349, opened 2018-03-27
Don't pin minor version of Python release

SethMichaelLarson:
Noticed we were testing against 3.6.3 instead of 3.6.4 on macOS builders and figured we'd rather have the latest release version for 3.6.x.

Another thought is if we're able to cache pyenv on macOS? My local installation of 3.6.4 under `~/.pyenv/versions/3.6.4/...` is 716KiB, well under [Travis' suggested limit of a "few hundred Megabytes"](https://docs.travis-ci.com/user/caching#How-does-the-caching-work%3F).

### DIFF
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -18,28 +18,28 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
 
     case "${TOXENV}" in
         py26)
-            pyenv install 2.6.9
-            pyenv global 2.6.9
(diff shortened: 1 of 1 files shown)


SethMichaelLarson:
Unfortunately it looks like there's no major.minor consistency for all major.minor combinations for pyenv so I'll close this. Looks like this is only available for 2.7 :man_shrugging:

PR #1350, opened 2018-03-28
Don't use SNI when connecting to an IP address

### LINKED ISSUE #1224: when connecting to https by IP address, client hello message has a server_name SNI block with literal ip address

SethMichaelLarson:
When we detect that we're connecting to an IP address we shouldn't use SNI. I used the code provided by @tiran to detect IP addresses without requiring the `ipaddress` module. Closes #1224.

I'm not a TLS master so I'd love confirmation if this is the right avenue to go down for this.


jonparrott:
Looks good, but can we add another test for `ssl_wrap_socket` that asserts SNI isn't used when specifying an IP address?

SethMichaelLarson:
That's exactly why I left as WIP! :D Will add during lunch if I can wrangle GitHub on mobile.

jonparrott:
Gotcha, I missed the WIP. :)

SethMichaelLarson:
Oh no worries at all, review comments are welcome anytime! :) Also gotta add change entry

SethMichaelLarson:
One question I had was that if the TLS library is compiled without SNI support and the user uses an IP address they'll receive the SNIMissingWarning even though SNI wouldn't have been sent anyways. Is this okay or should the behavior be changed to only emit the warning if SNI would have been used?

jonparrott:
Yeah we should probably not warn about missing SNI if we're not using SNI.

SethMichaelLarson:
Looks like `inet_pton` was Linux-only in 2.7 and earlier. Might be able to wrangle using `inet_aton` as a fallback but it'll be for IPv4 only... I guess we continue to send SNI on those platforms where we don't have `inet_pton` and are given an IPv6 address?

SethMichaelLarson:
This should be close to a final product, review comments welcome @jonparrott.

test/test_ssl.py
```diff
@@ -0,0 +1,73 @@
...
+     (False, 'www.python.org', False),
+     (False, '0.0.0.0', False),
+     (True, 'www.google.com', True),
+     (True, None, False),
+     (False, None, False)]
+)
+def test_context_sni_with_ip_address(monkeypatch, has_sni, server_hostname, uses_sni):
+    monkeypatch.setattr(ssl_, 'HAS_SNI', has_sni)
+
+    sock = mock.Mock()
```
jonparrott: Can you use `create_autospec` instead of a bare mock throughout?

SethMichaelLarson: All done!

SethMichaelLarson:
Looks like problems with using `create_autospec` on socket objects?

jonparrott:
Oh yeah, yikes, @SethMichaelLarson you're welcome to not use autospec in this case (which is unfortunate, but I don't expect the socket interface to change).

jonparrott:
LGTM, please ping me if tests are flaky and I'll merge.

SethMichaelLarson:
Did some debugging and within `ssl.py` the `match_hostname()` function matches with `DNS` certificate entries before `commonName` and the `DNS` entry on our cert is of course `localhost`. Hmph.

SethMichaelLarson:
Hm, that's not good. All of our dummyserver tests are being skipped on Travis because it can't import SecureTransport.

SethMichaelLarson:
This just turned into a more important PR. :)

SethMichaelLarson:
Basically it looks like any platform that couldn't import SecureTransport would skip ALL the dummyserver tests. We didn't see this problem with PyOpenSSL because we could import that on every platform that we tested on. I moved around the skip logic so that this doesn't occur and instead we only skip the SecureTransport variants of the tests when SecureTransport isn't available.

tiran:
FYI, the issue has been solved in Python 3.6.5 and will be solved in the upcoming release 2.7.15. In Fedora, the fix has been ported to all current releases of Python, too.

SethMichaelLarson:
Thanks for the update @tiran! :) I'll have to see how this code performs on 3.6.5 and 2.7.15.

SethMichaelLarson:
Only failure now seems to be Windows+Python 2.7 and it's due to `is_ipaddress('127.0.0.1')` returning `False` so that function needs to be looked at. I'm not sure what is causing it to error but the `ipaddress` module is installed so it's using `ipaddress.ip_address()`. Maybe there's something else funny going on being masked by exceptions. I'll find out and then this will finally be ready to be reviewed. :)

SethMichaelLarson:
Got CI to pass on Windows so this is all ready for a final review @jonparrott . :)

SethMichaelLarson:
I'll do a rebase, thanks for the review! :)

SethMichaelLarson:
Rebase complete, that looks a lot better.

PR #1355, opened 2018-03-31
be tolerant of non-unix line endings for securetransport cert bundles

reaperhulk:
While pip is resolving this for their use case, we should probably just make it so `\r\n` linefeeds get rewritten. I'm not handling the `\r` classic mac case because I hope that world no longer exists?


SethMichaelLarson:
Thanks for this @reaperhulk! @Lukasa is definitely the one to review this change. :)

SethMichaelLarson:
@jonparrott This is ready for a force-merge unless you get AppVeyor up and running this week. :)

jonparrott:
Appveyor is running, just inconsistently.

Issue #229, opened 2013-08-17
AttributeError thrown in connectionpool.py


YaronK:
While using Requests' post function (requests.post(<some url>)),
This following exception was thrown:

> File "C:\Python27\lib\site-packages\requests\api.py", line 88, in post
> return request('post', url, data=data, *_kwargs)
> File "C:\Python27\lib\site-packages\requests\api.py", line 44, in request
> return session.request(method=method, url=url, *_kwargs)
> File "C:\Python27\lib\site-packages\requests\sessions.py", line 335, in request
> resp = self.send(prep, *_send_kwargs)
> File "C:\Python27\lib\site-packages\requests\sessions.py", line 438, in send
> r = adapter.send(request, *_kwargs)
> File "C:\Python27\lib\site-packages\requests\adapters.py", line 272, in send
> conn = self.get_connection(request.url, proxies)
> File "C:\Python27\lib\site-packages\requests\adapters.py", line 199, in get_connection
> conn = self.poolmanager.connection_from_url(url)
> File "C:\Python27\lib\site-packages\requests\packages\urllib3\poolmanager.py", line 123, in connection_from_url
> return self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
> File "C:\Python27\lib\site-packages\requests\packages\urllib3\poolmanager.py", line 109, in connection_from_host
> pool = self._new_pool(scheme, host, port)
> File "C:\Python27\lib\site-packages\requests\packages\urllib3\poolmanager.py", line 79, in _new_pool
> return pool_cls(host, port, **kwargs)
> File "C:\Python27\lib\site-packages\requests\packages\urllib3\connectionpool.py", line 186, in __init__
> self.pool = self.QueueCls(maxsize)
> File "C:\Python27\Lib\Queue.py", line 35, in **init**
> self.mutex = _threading.Lock()
> AttributeError: 'NoneType' object has no attribute 'Lock'

LifoQueue inherits from Queue, which uses _threading (imported in Queue.py).
Because the Queue module isn't explicitly imported, _threading does not exist.

connectionpool.py, imports:
from Queue import LifoQueue, Empty, Full
import Queue # Duct tape solution

PR #1356, opened 2018-03-31
Switch to deque LifoQueue

### LINKED ISSUE #1087: Custom LifoQueue
During some profiling, I found out one slow operation is `self.pool.get()` in `_get_conn()`.

In urllib3, in `connectionpool.ConnectionPool`, it uses `LifoQueue`:
https://github.com/shazow/urllib3/blob/master/urllib3/connectionpool.py#L65

Which uses a list instead of a much faster deque:
https://github.com/python/cpython/blob/master/Lib/queue.py#L237

My suggestion is one could use a custom LifoQueue that uses the underlying deque, as:
```python
class LifoQueue(Queue):
    def _get(self):
        return self.queue.pop()
```

SethMichaelLarson:
Switch to using our own LifoQueue that uses a `collections.deque` backend which is a lot faster than a list backend. Closes #1087.

### DIFF
--- a/urllib3/connectionpool.py
+++ b/urllib3/connectionpool.py
@@ -41,11 +41,7 @@ from .util.response import assert_header_parsing
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import get_host, Url
-
-
-if six.PY2:
-    # Queue is imported for side effects on MS Windows
-    import Queue as _unused_module_Queue  # noqa: F401
+from .util.queue import LifoQueue
 
 xrange = six.moves.xrange
 
@@ -62,7 +58,7 @@ class ConnectionPool(object):
     """
 
     scheme = None
-    QueueCls = queue.LifoQueue
+    QueueCls = LifoQueue
 
     def __init__(self, host, port=None):
         if not host:
--- a/urllib3/util/queue.py
+++ b/urllib3/util/queue.py
@@ -1,0 +1,21 @@
+import collections
+from ..packages import six
+from ..packages.six.moves import queue
+
+if six.PY2:
+    # Queue is imported for side effects on MS Windows. See issue #229.
+    import Queue as _unused_module_Queue  # noqa: F401
+
+
+class LifoQueue(queue.Queue):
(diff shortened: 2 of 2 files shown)


urllib3/connectionpool.py
```diff
@@ -41,11 +40,8 @@
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import get_host, Url
-
-
-if six.PY2:
-    # Queue is imported for side effects on MS Windows
-    import Queue as _unused_module_Queue  # noqa: F401
+from .util.queue import LifoQueue
+from .packages.six.moves import queue
```
sigmavirus24: Why did this move down here? The rest of the imports from `packages` live elsewhere?

SethMichaelLarson: No reason, I deleted it and upon realizing that `queue.Empty` was used I put it back but not in the right place. I'll fix it and amend the commit.

urllib3/util/queue.py
```diff
@@ -0,0 +1,21 @@
+import collections
+from ..packages import six
+from ..packages.six.moves import queue
+
+if six.PY2:
+    # Queue is imported for side effects on MS Windows
```
jonparrott: No kidding? Got a link that explains what these side-effects are and why they're useful?

SethMichaelLarson: Let me find the commit where that was added in the blame.

SethMichaelLarson: Found the reason and PR where this was added: #229 https://github.com/urllib3/urllib3/commit/18e6e46ec7da085269ad4c7fcf01b13a7c572484
I'll add a link to it so no one has to go mucking through the enormous history log like I just did again. :)

SethMichaelLarson:
Ready for re-review. :)

SethMichaelLarson:
Might need an admin force merge for this one until AppVeyor is back :)

jonparrott:
Hrm, is appveyor still not reporting? I'll need to look into that a bit more closely on Friday.

PR #1358, opened 2018-04-04
Remove selectors backport

njsmith:
The selectors backport that urllib3 is carrying around is a huge
amount of complicated code, its tests are flaky, and the way urllib3
uses epoll/kqueue actually makes them *less* efficient than
select/poll.

So this PR makes urllib3 use select/poll directly. Check out that
diffstat 👀.

Also, the new wait_for_socket primitive is the same one that we need
for the bleach-spike branch (#1323), so this also helps set us up for
that work.

For ease of reviewing, each commit in this PR is self-contained.

CC: @pquentin


SethMichaelLarson:
Sounds good to me! The selectors code was probably too much for the task it was trying to accomplish. I remember vaguely an idea to eventually share a selector among all connections in a connection pool but that obviously didn't occur.

Couple of things that are handled by the old code that aren't handled by this current implementation:
- Monkeypatching of `select` module by gevent, eventlet, and friends. People directly or indirectly do this monkeypatching and expect their code to just work.
- If a platform incorrectly advertises support for `select.poll` but then on use will fail with `ENOSYS`. This has happened on multiple machines apparently.

What are your thoughts on these issues?

urllib3/util/wait.py
```diff
@@ -1,40 +1,130 @@
...
+
+
+def null_wait_for_socket(*args, **kwargs):
+    raise RuntimeError("no select-equivalent available")
+
+
+if hasattr(select, "poll"):
+    wait_for_socket = poll_wait_for_socket
+    HAS_WAIT_FOR_SOCKET = True
+elif hasattr(select, "select"):  # Platform-specific: Windows.
```
SethMichaelLarson: We're testing on Windows so we should remove this coverage-skipper.

test/test_wait.py
```diff
@@ -0,0 +1,125 @@
...
+        finally:
+            # Stop delivering SIGALRM
+            signal.setitimer(signal.ITIMER_REAL, 0)
+        end = monotonic()
+        dur = end - start
+        assert 0.9 < dur < 3
+    finally:
+        signal.signal(signal.SIGALRM, old_handler)
+
+    assert interrupt_count[0] > 0
```
SethMichaelLarson: I'd like to see a test for a closed socket as well.

njsmith: I've added a test checking that after the peer closes a socket, it becomes readable – is that what you had in mind?

njsmith:
> Couple of things that are handled by the old code that aren't handled by this current implementation:
>
> * Monkeypatching of select module at import time by gevent, eventlet, and friends. People directly or indirectly do this monkeypatching (incorrectly) and expect their code to just work.
> * If a platform incorrectly advertises support for select.poll but then on use will fail with ENOSYS. This has happened on multiple machines apparently.

I'm not sure I fully understand the desired behavior in the monkeypatching use case, but looking at [this code here](https://github.com/urllib3/urllib3/blob/7bab7ae1d2b416aeda3ecac8e94d1636e972407f/urllib3/util/selectors.py#L554-L592), it sounds like the solution for both is just to try calling `poll` before assuming it works? I've added that in my latest push.

SethMichaelLarson:
Basically anything we check about the `select` module on import-time is subject to change post-import time as users will do things like this:

```python
import urllib3
from gevent import monkey
monkey.patch_all()
```

which will modify the `select` module which we previously made assumptions about. Our solution was to check once on first use and then assume things will stay the same after first use and that seemed to work for people doing this.

urllib3/util/wait.py
SethMichaelLarson: Yeah our `.coveragerc` file has a lot of magic in it. :)

test/test_wait.py
SethMichaelLarson: I was more thinking of closing a socket on our side and selecting on it and see what the side-effects are for both select and poll.

njsmith: Waiting for an fd that you've already closed is a bug: the fd could be reallocated. In the special case where that fd is wrapped in a Python `socket.socket` object and only access through the `socket.socket` interface, it will at least reliably catch this bug, because `socket.socket.close` sets the internal fd attribute to -1, and then any attempt to do anything with the socket raises an error, because -1 is not a valid fd value. Still, just to be sure, I added a test verifying this.

njsmith:
Addressed all review comments.

SethMichaelLarson:
Code looks good to me, seems like there are a few lint errors. Can you address those?

njsmith:
Lint is passing now.

jonparrott:
Is the 3.4 failure on MacOS just flakiness? If so, I'll force merge.

pquentin:
Maybe @wbond or @SethMichaelLarson know if this test is part of the macOS flaky tests? At least the failure does not appear to be related to the change.

wbond:
The 3.4 error is one of the intermittent errors I've seen with the securetransport.py implementation. It is throwing a permission denied error when trying to create a keychain to import the custom CA certs into, I believe. The strange part is that the permission error is intermittent.

I don't know anything about the test infrastructure for this project, but it would be nice to set the test `test_client_no_intermediate`, `test_client_certs_one_file` and `test_client_certs_two_files` as allowing to fail when `sys.platform == 'darwin'`.

Based on what I've seen, all of the weird CI errors are related to client cert support. There may be more tests related to client certs that could error. Most likely someone needs to go over the securetransport implementation of client certs and look for errors.

SethMichaelLarson:
Looks like we missed a coverage drop here. @njsmith could you open a separate PR adding a testcase for `float('inf')`.

njsmith:
Oh, I see. Sure, I'll do that.

njsmith:
See #1376

PR #1359, opened 2018-04-05
Allow SecureTransport to work on Python 2.6

wbond:


SethMichaelLarson:
There are some tests in `test_securetransport.py` that are failing due to this change, could you address those? :)
@Lukasa would be the authority here although on the face of it this change LGTM.

wbond:
Yes, the tests were assuming that Python 2.6 would always have warnings, so I've fixed those. However, there still seems to be a failing test related to custom keychains. It isn't clear to me what is going on there.

njsmith:
Will the next release of urllib3 support 2.6?

wbond:
This is intended to fix pip 9 with the system Python 2.6 on Sierra. If it isn’t something that you’ll want for a future release, that’s no issue as I can submit the patch directly to pip.

SethMichaelLarson:
I'm really not a TLS/SSL expert, I'm unsure whether that error message is what we should expect or not with that situation.

wbond:
Alright, it looks like the issue was just the wording of the SSL exception when the client cert chain was incomplete.

SethMichaelLarson:
I'm unsure if this is the correct SSLError though. The SSLError we saw for the test was:
```
SSLError: [Errno 1] _ssl.c:499: error:14089086:SSL routines:ssl3_get_client_certificate:certificate verify failed
```
Why are we getting this error instead of one of the other errors on macOS 2.6 compared to any other OS/Python version?

urllib3/contrib/securetransport.py
```diff
@@ -207,9 +211,17 @@ def _read_callback(connection_id, data_buffer, data_length_pointer):
...
+                    chunk_size = base_socket.recv_into(
+                        buffer_view[read_count:requested_length]
+                    )
+                else:
+                    remaining = requested_length - read_count
+                    # This is likely innefficient, but works on Python 2.6
+                    buffer = (ctypes.c_char * remaining).from_address(
+                        data_buffer + read_count
+                    )
+                    chunk_size = base_socket.recv_into(buffer, remaining)
```
Lukasa: If this works, surely we don't need the `buffer_view` at all?

wbond: I don't know about the efficiency of constructing a new ctypes object from an address vs using a memoryview. I presume a memoryview is more efficient, just because less Python code is executing, and that ctypes isn't the most efficient way to do FFI.

Lukasa: The reason I was using `memoryview` was to be able to use `recv_into`. In general, so long as `recv_into` is working (that is, we aren't writing data into an ephemeral buffer) it's probably worth avoiding the `memoryview` altogether for the sake of simplicity.

wbond: Okay, I update it to just use the memory-offset ctypes array.

SethMichaelLarson:
Can you add an entry to the changelog and also add yourself to the list of authors? :) Thank you for tackling this!

urllib3/contrib/securetransport.py
```diff
@@ -207,9 +211,17 @@ def _read_callback(connection_id, data_buffer, data_length_pointer):
...
-                chunk_size = base_socket.recv_into(
-                    buffer_view[read_count:requested_length]
-                )
+                if buffer_view is not None:
+                    chunk_size = base_socket.recv_into(
+                        buffer_view[read_count:requested_length]
+                    )
+                else:
+                    remaining = requested_length - read_count
+                    # This is likely innefficient, but works on Python 2.6
```
pquentin: nitpick: I think the correct spelling is "inefficient"?

SethMichaelLarson:
This looks pretty good to me, I'll wait for a green CI and I'll say this is probably ready to merge. :)

wbond:
I’ll need to dig into this a little more, but it would appear there is something going on here. It kind of feels like some issues I was having with testing the SecureTransport backend of oscrypto.tls on Travis. The fact that the previous run failed on 2.6 and 3.6, but then this latest commit is green makes me think that something is amiss. Especially since 3.6 appears to indicate it was an internal error in SecureTransport.

SethMichaelLarson:
FYI macOS Travis builders are flaky as all hell. :) Just pay attention to tests you are working on currently.

wbond:
I'm opening and closing to trigger Travis to run again, hopefully to see what failure 2.6 is intermittently seeing. My guess is that it is the internal error that 3.6 saw at https://travis-ci.org/urllib3/urllib3/jobs/365916852.

SethMichaelLarson:
Btw if just one or two macOS builds are failing feel free to ping me. I can usually restart builds from mobile when needed. :)

wbond:
I'm just trying to get a sense of what is going on. Unfortunately I ditched Travis for OS X build because they got slow and seemed to have random failures all the time that I couldn't reproduce. Circle CI seems much more reliable for Mac builds.

The other issues is that I tried, but was unable to reproduce the testing setup here on my Mac. I forget what was going on, unfortunately, and didn't have time to debug another crazy testing config.

I'd love to just trigger the securetransport tests repeatedly to see if anything consistent is going on. If it were simpler/easier to run tests this PR probably would have been fixed up sooner. Ah well, CI seems to be the bane of the Python projects I work on these days. :-\

wbond:
I've just been going through the build history on Travis and it seems the securetransport tests fail pretty regularly giving false positives.

wbond:
Periodically the contents of `data_buffer` param to `_write_callback` has junk in the TLS record payload for a handshake message. The TLS record type is correct, the TLS version field and the length field is correct, but the first byte of the payload is not a valid Handshake record type. See https://tools.ietf.org/html/rfc5246#section-7.4. It should be 0, 1, 2, 11, 12, 13, 14, 15, 16, or 20. But looking through the failures as I was adding debugging information, it seems to contain gibberish. Here are examples of the failures:

 - https://travis-ci.org/urllib3/urllib3/jobs/366756643 - handshake type of `r`, aka 114
 - https://travis-ci.org/urllib3/urllib3/jobs/366784891 - handshake type of `V`, aka 86
 - https://travis-ci.org/urllib3/urllib3/jobs/366790746 - handshake type of `\xb0`, aka 176

This occurs at the same time as an `EPIPE` error in the write callback. Probably the next step is to trace the full handshake back and forth, but my interest in fixing this is waning. It appears this bug has been present for a while, looking at the build history, and I don't have much free time these days.

SethMichaelLarson:
Thanks for putting in so much effort into this! Don't burn yourself out on fixing this along with the original issue, maybe we should open up a separate issue and dump all the information you've found and perhaps someone else can get to it at a later time? Enjoy your weekend! :)

wbond:
I think this PR is in ok shape - it doesn't fix the random memory corruption that was there before, but it should allow pip to actually work on Python 2.6 for the 9.x branch, which is currently broken due to the new infrastructure and TLS 1.2-only.

SethMichaelLarson:
Sounds good to me, could you open an issue with your findings for the memory corruption?
I would revert to c794f27ba5a70ba79d5b4516740cdce56e73b63b and then rebase on master and we can merge this! :)

SethMichaelLarson:
Went ahead and resolved the merge conflicts with master.

wbond:
Sorry, was away for a few days - thanks for taking care of cleaning this up!

SethMichaelLarson:
Of course! Thanks for contributing this patch :)

PR #1363, opened 2018-04-18
Allow SOCKSConnection to parse username and password from URL

Kwentar:
Current version does not using auth info from url, i.e. 'socks5://user:password@id:port' user and password is ignored even so they exist on url

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -249,7 +249,7 @@ In chronological order:
 * Dominique Leuenberger <dimstar@opensuse.org>
   * Minor fixes in the test suite
 
-* Aleksei Alekseev alekseev.yeskela@gmail.com
+* Aleksei Alekseev <alekseev.yeskela@gmail.com>
   * using auth info for socks proxy
 
 * [Your name or handle] <[email or website]>
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -249,5 +249,8 @@ In chronological order:
 * Dominique Leuenberger <dimstar@opensuse.org>
   * Minor fixes in the test suite
 
+* Aleksei Alekseev alekseev.yeskela@gmail.com
+  * using auth info for socks proxy
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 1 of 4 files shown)


SethMichaelLarson:
Hey @Kwentar, thanks for opening this issue! I'm no expert on SOCKS but this seems entirely reasonable to me. One thing we do need is a test within [`test/contrib/test_socks.py`](https://github.com/urllib3/urllib3/blob/master/test/contrib/test_socks.py) to get full test coverage of this change.

CONTRIBUTORS.txt
```diff
@@ -249,5 +249,8 @@ In chronological order:
 * Dominique Leuenberger <dimstar@opensuse.org>
   * Minor fixes in the test suite
 
+* Aleksei Alekseev alekseev.yeskela@gmail.com
```
SethMichaelLarson: Nit: Surround your email with `<>`

Kwentar:
@SethMichaelLarson I see two errors in CI, but I cannot interpretate it, can you help me, what should I do to fix it?

SethMichaelLarson:
Eh, flaky macOS tests are flaky. I restarted the build.

SethMichaelLarson:
Looks like CI is passing. This is ready for merge after you take a look, @jonparrott.

urllib3/contrib/socks.py
```diff
@@ -152,6 +152,10 @@ def __init__(self, proxy_url, username=None, password=None,
                  num_pools=10, headers=None, **connection_pool_kw):
         parsed = parse_url(proxy_url)
 
+        if username is None and password is None and parsed.auth is not None:
+            split = parsed.auth.split(':')
+            if len(split) == 2:
```
jonparrott: If it's not two parts is it acceptable for us to silently ignore this or should we raise a `ValueError`?

Kwentar: I guess (if this question for me) it is not neccesary, because in this case we will have name or password is None, so it will work like as before. In my opinion, correctness of auth info should be in parse_url, probably parse_url should return not auth but pair of username/password, I am not sure. Never mind, if we need ValueError for wrong auth data - it should be in parse_url, not here

jonparrott: Sounds good.

SethMichaelLarson:
:tada: Thanks for this @Kwentar! If you can consider sticking around and helping us with our SOCKS support as none of us maintainers have any experience (to my knowledge) using SOCKS in our daily lives. Your experience could prove valuable.

Kwentar:
@SethMichaelLarson Unfortunately, I am not expert of SOCKS, I just fixed the problem which I had (third party library don't give me access to username/password, only url), but if I see any other things which I can perform better - I will do it :)

PR #1375, opened 2018-04-20
Address #1365 by fixing failing test on darwin + 3.7

### LINKED ISSUE #1365: Fix CI failures for macOS and Python 3.7
Before we know it Python 3.7 will be upon us so we should work towards fixing some of these failures and moving `py37` out of the `allow_failures` list within Travis.

crwilcox:
In python 3.7 on darwin I was seeing the following error. This PR loosens the requirement to allow the str repr

> E           assert "hostname 'ba...t match 'foo'" == '("hostname \'...ch \'foo\'",)'
> E             - hostname 'bar' doesn't match 'foo'
> E             + ("hostname 'bar' doesn't match 'foo'",)

Closes #1365

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,4 @@ matrix:
       env: TOXENV=py37
   allow_failures:
   - python: pypy-5.4
-  - python: 3.7-dev
-  - env: TOXENV=py37
 sudo: false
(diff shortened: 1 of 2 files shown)


crwilcox:
Test failed on py34 and py35. Seems unrelated to this PR though.

PR #1376, opened 2018-04-22
Add test for wait_for_socket EINTR handling w/ infinite timeout

njsmith:
This covers an otherwise uncovered branch in _retry_on_intr.

As requested here:
  https://github.com/urllib3/urllib3/pull/1358#issuecomment-383324003

### DIFF
--- a/test/test_wait.py
+++ b/test/test_wait.py
@@ -1,9 +1,11 @@
 import signal
 import socket
+import threading
 try:
     from time import monotonic
 except ImportError:
     from time import time as monotonic
+import time
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


test/test_wait.py
```diff
@@ -72,8 +74,9 @@ def test_wait_for_socket(wfs, spair):
     assert wfs(a, read=True, timeout=0)
 
     # Waiting for a socket that's actually been closed is just a bug, and
-    # raises an appropriate exception:
-    with pytest.raises(ValueError):
+    # raises some kind of helpful exception (exact details depend on the
+    # platform).
+    with pytest.raises(Exception):
```
SethMichaelLarson: Could this `Exception` be made more specific so we're not masking other failures?

njsmith: Probably, but it seems to depend on both python version and OS, and the operation we're testing there is total nonsense that can't possibly work, doesn't add any coverage, and should never happen in correct code, so it doesn't seem worth trying to curate a list of which exact exceptions happen in which configurations?

njsmith: But what I've seen so far is `ValueError` and `socket.error`, which could become the initial list of you want.

SethMichaelLarson: That is true, you've convinced me.

SethMichaelLarson:
Hey @jonparrott, all our checks are passing but I'm still not able to merge this, is this something that needs to change in the org/repo settings?

jonparrott:
@SethMichaelLarson try now?

SethMichaelLarson:
Works for me! :)

PR #1378, opened 2018-04-24
Only build the master branch and PRs

SethMichaelLarson:


pquentin:
I like to test my pull requests on a local branch before creating a pull request. Will this use case continue to be supported with that change?

SethMichaelLarson:
Hm, I wasn't aware people were doing that against `urllib3/urllib3`. When I want to do that I usually use my forked copy of urllib3 because I don't have to wait for other people's builds to complete.

SethMichaelLarson:
Maybe I should close this and just create PRs from my forked copy instead of use urllib3/urllib3. :) I'll close this.

pquentin:
The local branch would be in my fork indeed but not on the master branch. I think you do this too? If I understand correctly, the only way to get a Travis build at this point is to create a pull request.

But maybe that's not what's going on here?

SethMichaelLarson:
That would be the case if I made this change, I wasn't aware people were relying on builds of branches instead of pull requests only. Don't want to break anyone's flow for my convenience. ;)

sigmavirus24:
Travis doesn't understand the collaboration model where folks push to a branch on a Repository and send a PR and don't want 2 separate runs. The best way forward is to fork, sadly

Issue #1372, opened 2018-04-20
Address flaky macOS tests


theacodes:
From @wbond at https://github.com/urllib3/urllib3/pull/1358#issuecomment-383175550

> The 3.4 error is one of the intermittent errors I've seen with the securetransport.py implementation. It is throwing a permission denied error when trying to create a keychain to import the client certs into, I believe. The strange part is that the permission error is intermittent.

> I don't know anything about the test infrastructure for this project, but it would be nice to set the test test_client_no_intermediate, test_client_certs_one_file and test_client_certs_two_files as allowing to fail when sys.platform == 'darwin'.

> Based on what I've seen, all of the weird CI errors are related to client cert support. There may be more tests related to client certs that could error. Most likely someone needs to go over the securetransport implementation of client certs and look for errors.

This seems reasonable for now until we can figure out a proper fix

crwilcox:
I am submitting a PR that will patch the tests to not fail for now. The underlying issue is still there. These two tests fail because mac reports 'Import/Export format unsupported' under the `test_client_intermediate` and `test_client_no_intermediate` tests.

crwilcox:
@SethMichaelLarson , did your fix take hold for this? I think this can be closed now.

SethMichaelLarson:
I think so! There's still one flaky macOS socket level test but it's unrelated to my issue. I'll close for now!

PR #1379, opened 2018-04-24
Use b16encode instead of b64encode

SethMichaelLarson:
`b64encode()` can return strings with `'/'` characters which would make `Security.SecKeychainCreate()` think it was creating a file within a directory that didn't exist within our temporary directory. See #1372. cc: @wbond @crwilcox

### DIFF
--- a/urllib3/contrib/_securetransport/low_level.py
+++ b/urllib3/contrib/_securetransport/low_level.py
@@ -186,8 +186,8 @@ def _temporary_keychain():
     # some random bytes to password-protect the keychain we're creating, so we
     # ask for 40 random bytes.
     random_bytes = os.urandom(40)
-    filename = base64.b64encode(random_bytes[:8]).decode('utf-8')
-    password = base64.b64encode(random_bytes[8:])  # Must be valid UTF-8
+    filename = base64.b16encode(random_bytes[:8]).decode('utf-8')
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Is there a compelling reason against using `altchars`? https://docs.python.org/3.5/library/base64.html#base64.b64encode

SethMichaelLarson:
There didn't appear to be a reason to use b64 over b16/32 so I went with b16 for simplicity. Should we still be using b64?

sigmavirus24:
I don't think there's a good reason for it. I just worry that someone is relying on this (for very poor reasons) and will be surprised with the change. I don't think it's a reason to go back to b64, I'm just anticipating the worst

jonparrott:
considering it's a temporary file that was already a random name, I think the risk here is relatively low.

sigmavirus24:
github3.py has made temporary files in the past that people try to find and use before they disappear. The things users will do boggles the mind

jonparrott:
> github3.py has made temporary files in the past that people try to find and use before they disappear. 

what

ofek:
I think the standard way to do this is https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode

"Encode bytes-like objects using the URL- and filesystem-safe alphabet, which substitutes `-` instead of `+` and `_` instead of `/` in the standard Base64 alphabet, and return the encoded bytes."

Issue #1272, opened 2017-10-02
Add functionality to get the final request URL after redirects, like the function urllib2.urlopen().geturl().


bosihu:

Lukasa:
You cannot: this functionality is not supported at this time. Patches to add it would be welcome. :smile:

crwilcox:
I was investigating this and have a possible solution, but it may make more sense to have this as a utility than to put it off of urlopen.  HttpResponse, which is returned by `urllib3.PoolManager().request` and `urllib3.PoolManager().urlopen`, doesn't seem to have the url that the data came from. However, HttpResponse does have retry history which, in the case there was a redirect, the url can be discovered.  A field containing the location could be added but I am not sure it makes sense to add an additional field to each HttpResponse object. This field would be needed to handle the case that the response was not from a redirect.

If you try 
`
r = http.request('GET', 'http://github.com/shazow/urllib3')
r.retries.history
`

You will get the following:
`
RequestHistory(method='GET', url='http://github.com/shazow/urllib3', error=None, status=301, redirect_location='https://github.com/shazow/urllib3')
RequestHistory(method='GET', url='https://github.com/shazow/urllib3', error=None, status=301, redirect_location='https://github.com/urllib3/urllib3')
`

This will be empty in the case of no redirects.

The following code seems to do what you want: 

`
def geturl(request_type, url):
    http = urllib3.PoolManager()
    response = http.request(request_type, url)
    if len(response.retries.history):
        return response.retries.history[-1].redirect_location
    else:
        return url
`

SethMichaelLarson:
Hey @crwilcox could you turn that into a PR with some tests? I'd gladly accept that change. :)

crwilcox:
@SethMichaelLarson sure. feel free to assign this to me and I can probably do it Friday :)

SethMichaelLarson:
If you've got time I'd still like to see a PR for this functionality even if it adds a field to response objects. It's genuinely useful functionality we're missing compared to httplib.

SethMichaelLarson:
Closed by #1382

PR #1382, opened 2018-04-27
Implements #1272 by adding a geturl method to HTTPResponse objects

crwilcox:


crwilcox:
FYI: Will be making tests for this.

crwilcox:
Thanks @SethMichaelLarson . I have updated `CHANGES.rst`.

One thing I realized went unsaid. I made the name `geturl` but could see changing it to be more pythonic by naming it `get_url`. I couldn't decide which was more important; maintaining the name from `urllib2` or conforming to convention.

PR #1389, opened 2018-05-27
Pin pyOpenSSL major version to 17.* for Python 2.6

SethMichaelLarson:
See: https://github.com/pyca/pyopenssl/blob/74de8a137d435d45c100b74cc971be556166a559/CHANGELOG.rst#1800-2018-05-16

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 from setuptools import setup
 
 import os
+import sys
 import re
 import codecs
 
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ version = VERSION
 
 # pyOpenSSL version 18.0.0 dropped support for Python 2.6
 if sys.version_info < (2, 7):
-    PYOPENSSL_VERSION += 'pyOpenSSL >= 0.14, < 18.0.0'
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -63,7 +70,7 @@
       test_suite='test',
       extras_require={
           'secure': [
-              'pyOpenSSL>=0.14',
+              PYOPENSSL_VERSION,
```
sigmavirus24: This may require a change in setup.cfg too for the wheel

sigmavirus24: Actually it does: https://github.com/urllib3/urllib3/blob/30e9855307aaa9f9f09651f2bbddeb7350eed666/setup.cfg#L14 Your logic here will need to become more complicated.

SethMichaelLarson: Dang, reasons we can't have nice things. :P Good catch!

SethMichaelLarson:
Would something to the effect of:
```
pyOpenSSL>=0.14, <18.0.0; python_version=="2.6" and extra == 'secure'
pyOpenSSL>=0.14; python_version=="2.7" and extra == 'secure'
```
be sufficient?

sigmavirus24:
Yes, that should do what you want... I think.

SethMichaelLarson:
Thanks for the review @sigmavirus24!

PR #1391, opened 2018-06-03
Retry.from_int() checks whether its argument is an int, not a Retry

marnanel:
This solves a recurring problem when two versions of urllib3 are
installed, and a Retry object created by one is passed to the from_int()
of another. The intended behaviour is to return that Retry object
unchanged, but at present we test

	if isintance(retries, Retry):

The test fails, because "retries" is a Retry object from a
different installation of urllib3. So we get a new Retry whose
"total" member is nonsensically another Retry object. This causes
a baffling-looking exception:

	TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'

See, for example, #3560.

The solution is to return "retries" if it's *not* an integer.

### DIFF
--- a/urllib3/util/retry.py
+++ b/urllib3/util/retry.py
@@ -202,7 +202,7 @@ class Retry(object):
         if retries is None:
             retries = default if default is not None else cls.DEFAULT
 
-        if isinstance(retries, Retry):
+        if not isinstance(retries, int):
             return retries
 
         redirect = bool(redirect) and None


SethMichaelLarson:
When would two versions of urllib3 be installed in one installation? That seems like a way to get into trouble regardless of what we do here.

sigmavirus24:
> When would two versions of urllib3 be installed in one installation? That seems like a way to get into trouble regardless of what we do here.

@SethMichaelLarson This happens when, for example, someone has a much older version of Requests installed that still vendors urllib3 *and* urllib3 installed. Instead of importing the `Retry` object from the vendored package (`requests.packages.urllib3`), they import it from `urllib3`. A similar problem happens with linux distribution-provided packages that have hacked apart requests (when it still vendored its dependencies) and while `urllib3.util.Retry` and `requests.package.urllib3.util.Retry` come from the same file on disk, Python does not consider them to be equivalent.

In other words, this is potentially easily fixed by users unless it's coming from some package doing the wrong thing that they don't control and which their linux distributor isn't keeping properly up-to-date with bug fixes for the package.

The tl;dr is that this could potentially help in a very small set of areas but the likelihood of it helping in the areas where it is most needed is very very slim.

pquentin:
@sigmavirus24 Thanks for the explanation! Does that mean you're in favor of the change? If it's not helpful I think it slightly degrades readability.

urllib3/util/retry.py
```diff
@@ -202,7 +202,7 @@ def from_int(cls, retries, redirect=True, default=None):
         if retries is None:
             retries = default if default is not None else cls.DEFAULT
```
SethMichaelLarson: Might be good to leave a comment here about why we're doing this?

sigmavirus24:
@pquentin I'm borderline against this change because it doesn't hold the people who are breaking our perfectly stable source code accountable. Further, this is a problem that has been happening for years and the likelihood that accepting this PR helps a user is pretty low in my estimation. The downstream redistributors are just too damned busy to care about this patch. They have bigger fish to fry. Furthermore, Requests doesn't vendor urllib3 anymore so the appropriate thing to do is import directly from urllib3 and if the packagers kept Requests and urllib3 up-to-date, this problem would disappear.

So yeah, I'm not in favor of this because, as you mention, it hurts readability and it fixes a problem that likely won't benefit the people who need it most.

SethMichaelLarson:
I'm agreeing with @sigmavirus24 here, it's a -0 from me.

theacodes:
I'm with Ian. Let's not set a precedent by doing this.

pquentin:
@sigmavirus24 I'd like to say that I appreciated your thoughtful explanations! 🌷

jhawkinson:
Sorry I'm a little late to this party, but I don't think that you guys should use isinstance() in the cases where you do.  This statement is largely because you are a low level library and you cannot control the imports for your users like you'd expect.  Here is an example

libs/a.py
```
class A(object):

    def is_instance(self, obj):
        print(A)
        print(obj)
        print(isinstance(obj, A))
```
libs/b.py
`from a import A`

c.py
```
import os
import sys

from libs.a import A

sys.path.insert(0, os.path.join(os.path.dirname(__file__), "libs"))
from b import A as B

C = B()
A().is_instance(C)

```

Which outputs....

python3 c.py
```
<class 'libs.a.A'>
<a.A object at 0x7f6eb692ec88>
False

```
I think I'm running into a similar problem in the issue linked above (but I haven't sorted through everything to confirm).   It seems like you could do something else for type checking perhaps like have a class variable which you can test across multiple instances or something?  idk.  personally in my code I test using inspect.getmro() (but that wouldn't catch cases of actual different versions on the class).

The problem I'm having is that I have two third party libraries that -- on their own -- work 100% fine.  However if I try to use them both in the same application they fail.  It seems unreasonable for all other vendors, or consumers of your lib to check with each other in how they use this code when the problem can be fixed here.

Sorry if I'm missing something, as mentioned in the issue report, I'm not very strong with isinstance() and the subtle namespace variances we're talking about here.

PR #1397, opened 2018-06-15
Support a servername parameter on HTTPSConnections

JackOfMostTrades:
The servername parameter allows callers to override the hostname used for SNI. If specified this hostname will also be used as the value for hostname verification instead of the actual hostname (unless assert_hostname is set).

This is the same option and semantics that [node.js provides](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) via the "servername" attribute, or that [golang provides](https://golang.org/pkg/crypto/tls/#Config) via the "ServerName" option.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ dev (master)
 
 * ... [Short description of non-trivial change.] (Issue #)
 
+* Add a server_hostname parameter to HTTPSConnection which allows for
+  overriding the SNI hostname sent in the handshake. (Pull #1397)
+
 
 1.23 (2018-06-05)
 -----------------
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -242,7 +242,7 @@ class HTTPSConnection(HTTPConnection):
 
     def __init__(self, host, port=None, key_file=None, cert_file=None,
                  strict=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                 ssl_context=None, **kw):
+                 ssl_context=None, server_hostname=None, **kw):
 
         HTTPConnection.__init__(self, host, port, strict=strict,
                                 timeout=timeout, **kw)
@@ -250,6 +250,7 @@ class HTTPSConnection(HTTPConnection):
         self.key_file = key_file
         self.cert_file = cert_file
         self.ssl_context = ssl_context
+        self.server_hostname = server_hostname
 
         # Required property for Google AppEngine 1.9.0 which otherwise causes
         # HTTPS requests to go out as HTTP. (See Issue #356)
@@ -270,6 +271,7 @@ class HTTPSConnection(HTTPConnection):
             keyfile=self.key_file,
             certfile=self.cert_file,
             ssl_context=self.ssl_context,
+            server_hostname=self.server_hostname
         )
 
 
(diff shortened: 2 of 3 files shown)


urllib3/connection.py
```diff
@@ -242,14 +242,15 @@ class HTTPSConnection(HTTPConnection):
 
     def __init__(self, host, port=None, key_file=None, cert_file=None,
                  strict=None, timeout=socket._GLOBAL_DEFAULT_TIMEOUT,
-                 ssl_context=None, **kw):
+                 ssl_context=None, servername=None, **kw):
```
theacodes: Can we use the same argument name as `ssl_wrap_socket`?

JackOfMostTrades: Yep, can do. Changed it to server_hostname everywhere.

theacodes:
Also we'll need to make sure coverage stays at 100%.

JackOfMostTrades:
Sure thing; I've updated the PR with a changelog update and added a test which asserts that server_hostname gets passed through to the wrapped socket and gets the code coverage back up to 100%.

JackOfMostTrades:
And by the way, the docs in this project for helping a new developer get tests running are fantastic. That's always been such a struggle with other projects and it was incredibly easy this time around. I appreciate the time spent writing that and keeping it up to date! :)

theacodes:
> And by the way, the docs in this project for helping a new developer get tests running are fantastic. That's always been such a struggle with other projects and it was incredibly easy this time around. I appreciate the time spent writing that and keeping it up to date! :)

That's great to hear!

SethMichaelLarson:
Thanks for submitting this patch. Can you help me understand the use-case for this parameter? I'm unsure of a situation where you'd want to override the hostname sent via SNI compared to the one you're connecting to.

JackOfMostTrades:
> Can you help me understand the use-case for this parameter?

Sure. This doesn't really apply on the public internet, but it's something I've run into quite a bit in internal infrastructure. The most common case where I've run into this where you're connecting by IP instead of hostname (but you still want a particular virtual host served back). There's a few reasons you might be connecting by IP:
* You're not using DNS for host resolution; e.g. Netflix uses [Eureka](https://github.com/Netflix/eureka) which has built-in load-balancing and failover logic.
* You're on an internal network and the DNS name would resolve to a public IP address which doesn't route correctly from the internal network.
* You're running integration tests, e.g. against localhost. The service might not actually get added to the DNS hostname until after the integration tests have run and verified that the virtualhost configuration is working as expected.

SethMichaelLarson:
Thanks for explaining it! Sounds good to me. :+1:

SethMichaelLarson:
Looks like a merge conflict that needs resolving for `urllib3/connection.py`. Could you address this and resubmit? :)

pquentin:
urllib3/connection.py moved to src/urllib3/connection.py, so that should not be too hard to fix. :) The good news is that there is a small chance that a rebase will also fix the test_client_no_intermediate error on Windows.

JackOfMostTrades:
Rebased. The windows test cleared up, although now a random macOS run of test_client_no_intermediate failed.

SethMichaelLarson:
Uh oh, @pquentin?

pquentin:
@SethMichaelLarson Yes, this confirms that the serial number length issue is only a macOS 10.13+ issue or at least was not the only problem to fix. That at least makes sense, but unfortunately that also means that the macOS 10.12 flakiness on CI is not solved yet. At least now on every failing build we can see 1/ the actual exception 2/ where it got thrown.

pquentin:
(However, the problem is unrelated to this pull request.)

SethMichaelLarson:
Merged! Thanks @JackOfMostTrades! :tada:

PR #1399, opened 2018-06-18
Handle EINTR from poll with zero timeout

### LINKED ISSUE #1396: Interrupted system call while profiling with plop
Using Python 2.7.12, requests 2.19.0, urllib3 1.23 and using [plop](https://github.com/bdarnell/plop) for profiling, I'm intermittently hitting this stack trace in long-running code:
```
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/requests/sessions.py", line 525, in get
    return self.request('GET', url, **kwargs)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/requests/sessions.py", line 512, in request
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/requests/sessions.py", line 622, in send
  File "/home/bmerry/work/sdp/git/katdal/katdal/chunkstore_s3.py", line 56, in send
    return super(_TimeoutHTTPAdapter, self).send(request, stream, timeout, *args, **kwargs)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/requests/adapters.py", line 445, in send
    timeout=timeout
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/connectionpool.py", line 588, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/connectionpool.py", line 239, in _get_conn
    if conn and is_connection_dropped(conn):
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/util/connection.py", line 23, in is_connection_dropped
    return wait_for_read(sock, timeout=0.0)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/util/wait.py", line 146, in wait_for_read
    return wait_for_socket(sock, read=True, timeout=timeout)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/util/wait.py", line 107, in poll_wait_for_socket
    return bool(_retry_on_intr(do_poll, timeout))
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/util/wait.py", line 47, in _retry_on_intr
    return fn(timeout)
  File "/home/bmerry/work/sdp/env/local/lib/python2.7/site-packages/urllib3/util/wait.py", line 105, in do_poll
    return poll_obj.poll(t)
select.error: (4, 'Interrupted system call')
Profiling timer expired
```

Looking at the implementation of `_retry_on_intr` for older Pythons, it has this special case:
```python
        if timeout is not None and timeout <= 0:
            return fn(timeout)
```
which in turn seems to apply to the call stack above (see is_connection_dropped, which passes a timeout of 0.0). So apparently there are cases where poll can fail with EINTR even with a zero timeout. FWIW, I'm running Ubuntu 16.04 and Linux 4.4.0-116-generic. I'll try commenting out that fast path and doing some testing overnight to confirm that that is the problem.

I don't yet had a minimal reproducible example, but I'll work on it (my first attempt of just banging on some URL in a loop hasn't worked). I wanted to file this before I forgot.

bmerry:
While common sense says that it should be impossible, it seems that
poll() can indeed fail with EINTR on Linux even when the timeout is
zero. There was logic in _retry_on_intr which bypassed the restart
handling when the timeout was zero, which led to EINTRs leaking out.

Fixed by setting the deadline to 0 when the timeout is 0. This avoids an
extra system call to get monotonic(). It assumes that monotonic() is
always non-negative; that isn't documented but seems a reasonable
assumption given that the reference point tends to be something like
boot time.

Closes #1396.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -259,5 +259,5 @@ In chronological order:
   * Improve contribution guide
   * Add ``HTTPResponse.geturl`` method to provide ``urllib2.urlopen().geturl()`` behavior
 
-* [Your name or handle] <[email or website]>
-  * [Brief summary of your changes]
+* Bruce Merry <http://www.brucemerry.org.za>
+  * Fix leaking exceptions when system calls are interrupted with zero timeout
--- a/urllib3/util/wait.py
+++ b/urllib3/util/wait.py
@@ -43,11 +43,10 @@ if sys.version_info >= (3, 5):
 else:
     # Old and broken Pythons.
     def _retry_on_intr(fn, timeout):
-        if timeout is not None and timeout <= 0:
-            return fn(timeout)
-
         if timeout is None:
             deadline = float("inf")
+        elif timeout <= 0:
+            deadline = 0.0
         else:
             deadline = monotonic() + timeout
 
@@ -117,7 +116,7 @@ def _have_working_poll():
     # from libraries like eventlet/greenlet.
     try:
         poll_obj = select.poll()
-        poll_obj.poll(0)
+        _retry_on_intr(poll_obj.poll, 0)
     except (AttributeError, OSError):
         return False
     else:
--- a/test/test_wait.py
+++ b/test/test_wait.py
@@ -142,6 +142,43 @@ def test_eintr(wfs, spair):
     reason="need setitimer() support"
 )
 @pytest.mark.parametrize("wfs", variants)
+def test_eintr_zero_timeout(wfs, spair):
+    a, b = spair
+    interrupt_count = [0]
+
+    def handler(sig, frame):
+        assert sig == signal.SIGALRM
+        interrupt_count[0] += 1
+
+    old_handler = signal.signal(signal.SIGALRM, handler)
+    try:
+        assert not wfs(a, read=True, timeout=0)
+        start = monotonic()
+        try:
+            # Start delivering SIGALRM 1000 times per second,
+            # to trigger race conditions such as
+            # https://github.com/urllib3/urllib3/issues/1396.
+            signal.setitimer(signal.ITIMER_REAL, 0.001, 0.001)
+            # Hammer the system call for a while to trigger the
+            # race.
+            for i in range(100000):
+                wfs(a, read=True, timeout=0)
+        finally:
+            # Stop delivering SIGALRM
+            signal.setitimer(signal.ITIMER_REAL, 0)
+        end = monotonic()
+        dur = end - start
+    finally:
+        signal.signal(signal.SIGALRM, old_handler)
+
+    assert interrupt_count[0] > 0
+
+
+@pytest.mark.skipif(
+    not hasattr(signal, "setitimer"),
+    reason="need setitimer() support"
+)
+@pytest.mark.parametrize("wfs", variants)
 def test_eintr_infinite_timeout(wfs, spair):
     a, b = spair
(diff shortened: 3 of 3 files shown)


CONTRIBUTORS.txt
```diff
@@ -259,5 +259,5 @@ In chronological order:
   * Improve contribution guide
   * Add ``HTTPResponse.geturl`` method to provide ``urllib2.urlopen().geturl()`` behavior
 
-* [Your name or handle] <[email or website]>
-  * [Brief summary of your changes]
```
pquentin: Note that you need to leave this here for future contributors. :)

bmerry: Doh! Fixed.

bmerry:
The current Travis failure doesn't look like anything related to the changes I made - is it expected?

urllib3/util/wait.py
```diff
@@ -117,7 +116,7 @@ def _have_working_poll():
     # from libraries like eventlet/greenlet.
     try:
         poll_obj = select.poll()
-        poll_obj.poll(0)
+        _retry_on_intr(poll_obj.poll, 0)
```
SethMichaelLarson: Nice catch here.

SethMichaelLarson:
Yeah it seems like GAE is broken and Python 3.7 is no longer dev so that needs to be fixed. Those failures aren't related.

SethMichaelLarson:
cc @njsmith

urllib3/util/wait.py
bmerry: I can't take credit - the unit test exposed it.

njsmith:
Huh, this can really happen? I guess I learned something new about Unix then!

> Fixed by setting the deadline to 0 when the timeout is 0. This avoids an
extra system call to get monotonic(). It assumes that monotonic() is
always non-negative; that isn't documented but seems a reasonable
assumption

The POSIX spec does actually require it to be positive: "the amount of time (in seconds and nanoseconds) since an unspecified point **in the past**" (my emphasis; [reference](http://pubs.opengroup.org/onlinepubs/9699919799/functions/clock_gettime.html)).

However, people don't always follow specs, and cleverness makes code harder to read. Plus trying to optimize out a call to `monotonic` here is pretty pointless; it's faster than almost anything else this function does. So please use negative infinity or `monotonic() - 1` or something instead of 0.

Otherwise looks good to me.

bmerry:
> Huh, this can really happen? I guess I learned something new about Unix then!

I was pretty surprised too.

> The POSIX spec does actually require it to be positive

Good to know. I was reading the Python docs since monotonic is more than just POSIX, but I guess we wouldn't be dealing with EINTR on other platforms.

The only reason I tried to be clever is that I assumed the original special case was trying to be a faster path. You're right about `monotonic` being fast: it seems to be faster than evaluating `float("-inf")` for example. I'll just take out special handling for zero timeouts entirely.

urllib3/util/wait.py
SethMichaelLarson: You can take credit for the unit test though. ;)

SethMichaelLarson:
Thanks @bmerry!

PR #1401, opened 2018-06-20
Remove unneeded dependency psutil

frenzymadness:
Hello.

I tried to remove `psutil` from RPM specfile in Fedora and it seems that it's unneeded dependency because everything works without it.

I cannot find any usage in the source code, only in git history, so it's probably just forgotten.

Will see if the CI tests pass.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,7 +6,6 @@ wheel==0.24.0
 tornado==4.2.1
 PySocks==1.5.6
 pkginfo>=1.0,!=1.3.0
-psutil==4.3.1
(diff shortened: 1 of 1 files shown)


frenzymadness:
One of sixteen jobs failed but the failure doesn't seem to be related to this change. Could anybody please restart that one job?

frenzymadness:
It's not needed for installing on Fedora but we are running tests during a build of RPM package so we had psutil as a build dependency for urllib3.

pquentin:
@SethMichaelLarson I believe this PR can be merged, what do you think?

PR #1412, opened 2018-07-10
Fix test_client_no_intermediate on macOS 10.13+

pquentin:
RFC 3280 and RFC 5280 specify that [serial numbers must not be longer
than 20 octets][0], but the serial number in client_no_intermediate.pem
was 38 bytes long. Using 20 bytes fixes the issue.

This test was passing on macOS 10.12. But since macOS 10.13, the serial
number are rejected when they're too long. macOS 10.12 was only warning
about the issue, see [SecCertificateParse][1]:

    /* RFC5280 4.1.2.2 limits serial number values to 20 octets.
       %%% At some point, this should be treated as a hard error.*/

But on macOS 10.13, [SecCertificateParse][2] now says:

    /* Note: RFC5280 4.1.2.2 limits serial number values to 20 octets.
       For now, we warn about larger values, but will still create the
       certificate with values up to 36 octets to avoid breaking some
       nonconforming certs with slightly longer serial numbers.
       We also explicitly allow serial numbers of 21 octets where the
       leading byte is 0x00 which is used to make a negative 20 octet
       value positive.
    */

It turns out that the [serial number in our certificate is 38 bytes
long][3], which is larger than the new 36 bytes limit. We want to use a
conforming serial number, which is why this new certificate uses 20
bytes.

The new certificate was generated using [trustme][4]. The long term plan
is to use trustme inside tests to generate all certificates, which will
allow to keep the generation of the certificate close to their usage,
and take advantage of future trustme improvements. However, trustme
does not support Python 2.6, so we'll need to drop Python 2.6 support in
urllib3 before switching to trustme.

test_client_intermediate also fails on macOS 10.13, but replacing the
certificate requires producing a certificate with an intermediate CA,
which is not possible yet.

Thanks to:

* @bartonjs for pointing out that the serial number length was the issue
* @yan12125 for showing log stream and the source of SecCertificateParse
* @njsmith for their help on using trustme

[0]: https://tools.ietf.org/html/rfc5280#section-4.1.2.2
[1]: https://opensource.apple.com/source/Security/Security-57740.60.18/OSX/sec/Security/SecCertificate.c.auto.html
[2]: https://opensource.apple.com/source/Security/Security-58286.41.2/OSX/sec/Security/SecCertificate.c.auto.html
[3]: https://stackoverflow.com/a/51194617
[4]: https://github.com/python-trio/trustme


pquentin:
(Note that I don't understand why the test was flaky CI, but this pull request can only improve the situation.)

yan12125:
Thanks for this! I can confirm this fixed client_no_intermediate on macOS 10.13.5.

> However, trustme does not support Python 2.6, so we'll need to drop Python 2.6 support in urllib3 before switching to trustme.

I think it's OK to switch without dropping Python 2.6 as it's not used by urllib3 client; it's for generating testing data only.

pquentin:
@yan12125 Sorry if I was not being clear, but I was talking about generating new certs dynamically during each test run. Python 2.6 is something that is holding urllib3 back anyway, and it was only kept for pip who recently removed support for it.

theacodes:
This is awesome, thank you @pquentin!

njsmith:
@pquentin yayy

Can you post the script you used to generate the new cert somewhere, even like as a comment here or something? Just to make sure it doesn't get lost. Also, I'm curious about how you convinced trustme to generate a cert with intermediates :-)

pquentin:
So while that did fix the import issue, we just had a test_client_no_intermediate failure on Python 2.6 with SecureTransport: "connection closed via error": https://travis-ci.org/urllib3/urllib3/jobs/402308747. *sigh*.

@njsmith sure, I have forgotten to do that, sorry. I only used code from the cheat sheet:

```
import trustme
ca = trustme.CA()
server_cert = ca.issue_server_cert(u"test-host.example.org")
server_cert.private_key_and_cert_chain_pem.write_to_path("server.pem")
```

(Maybe there's something better to use as the hostname here.)

And I split server.pem in two parts: client_no_intermediate.key and client_no_intermediate.pem, which you can see in this PR.

Note that I did not fix test_client_intermediate on macOS 10.13, only test_client_**no**_intermediate, which was the flaky test on Travis. I do intend to modify trustme to generate certificates with intermediate CAs, but have not got around to do that yet.

njsmith:
Ah, I had the impression that what the test was checking was a cert that required an intermediate cert, but where the intermediate cert was missing.

pquentin:
@njsmith How is that different from a cert that requires a CA cert, but where the CA is missing? (Sorry, I don't know much about SSL certificates.)

SethMichaelLarson:
Thanks @njsmith! Could we mend that certificate to bring the test back to its original code path in a PR?

pquentin:
Thanks @njsmith! That's really helpful.

@SethMichaelLarson I now plan to fix the certificate once I can add support for intermediate CAs in trustme, which is needed to fix test_client_intermediate (which also fails on macOS 10.13+).

PR #1414, opened 2018-07-14
Support using tornado-5 for tests

mgorny:
Here's a minimal change that makes the test support compatible with tornado-5. Apparently, the io_loop parameter to server classes has been [removed in 5.0.0](http://www.tornadoweb.org/en/stable/releases/v5.0.0.html) in favor of using a single event loop everywhere.

I've kept the `io_loop` parameter inside internal API to make this change as small as possible. I can remove the parameters and inline `tornado.ioloop.IOLoop.current()` everywhere if you prefer that.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,8 @@ coverage==3.7.1
 tox==2.1.1
 twine==1.5.0
 wheel==0.24.0
-tornado==4.2.1
+tornado==5.0.2; python_version>"2.6"
+tornado==4.2.1; python_version<="2.6"
 PySocks==1.5.6
 pkginfo>=1.0,!=1.3.0
 pytest-cov==2.5.1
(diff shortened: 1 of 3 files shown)


pquentin:
Thanks @mgorny! That's really useful. The failure is a [known issue we're still working on](https://github.com/urllib3/urllib3/pull/1412), and upgrading to Tornado 5 was one of the things I wanted to try. Since we should be able to remove support for Python 2.6 soon, it would be nice if Python 2.6/Tornado 4 was the only version that failed, but we'll have to run more builds to see if it's true.

I personally would not ask you to make more changes: they could be done in a separate PR. But of course that decision should be made by the urllib3 maintainers.

SethMichaelLarson:
Thanks @mgorny! :tada:

mgorny:
Thanks a lot for merging it!

PR #1418, opened 2018-08-06
Revert "Fix test_client_no_intermediate on macOS 10.13+ (#1412)"

pquentin:
This reverts commit 6ef3c06249b7144838b6701c6e9adfa76650ccea from PR https://github.com/urllib3/urllib3/pull/1412 (but keeps the assertion changes that makes it easier to see the traceback on failure). The test was supposed to test a missing intermediate CA, but after that change it was incorrectly testing a missing root CA instead. See https://github.com/urllib3/urllib3/pull/1412#issuecomment-404385457 from @njsmith for a nice explanation of the difference.

I now realize that we should fix the serial number length in test_client_intermediate first. Once that is done, it's easy to simply remove the intermediate certificate for test_client_no_intermediate, to make sure that we still test the same thing as before, but with compliant certificates. Unfortunately, I still don't know how to fix test_client_intermediate, and in the meantime the change from PR https://github.com/urllib3/urllib3/pull/1412 is incorrect. That change also fails to fix the test suite on macOS 10.13 since it only modifies test_client_no_intermediate while macOS 10.13 also rejects the certificates from test_client_intermediate.


SethMichaelLarson:
How are we planning on tracking that these tests no longer run for SecureTransport? I don't want to lose traceability once we put a skip decorator on these tests.

pquentin:
@SethMichaelLarson I opened https://github.com/urllib3/urllib3/issues/1419 to keep track of the problem, do you think that's enough?

SethMichaelLarson:
Sounds good to me, thank you. I'll merge this.

pquentin:
Thanks @SethMichaelLarson!

PR #1420, opened 2018-08-06
Require cert verification by default for the TestHTTPS pool

SethMichaelLarson:


SethMichaelLarson:
I actually don't think we want this on the server side, wouldn't that mean the server requires the client to provide a certificate?

pquentin:
The [user guide](https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification) notes "By default, urllib3 does not verify HTTPS requests." It makes sense to include a test for the default behavior, don't you think? We can probably rename test_simple to test_unverified for clarity?

(Another unrelated issue with self._pool is that it's modified by test_ssl_correct_system_time and test_ssl_wrong_system_time, so the order of tests matter, which is not a good idea.)

SethMichaelLarson:
Aren't there unverified request occurring throughout the test suite though? And that's a good catch for those system time tests.

pquentin:
> Aren't there unverified request occurring throughout the test suite though?

Hmm, I would expect most requests in the test suite to be HTTP, but this is an unverified HTTPS request, so that's a bit different, I think?

SethMichaelLarson:
I think your idea to keep as an unverified request and then change the name to test_unverified_request() or the like is the best idea.

SethMichaelLarson:
Closing this as it's no longer needed.

PR #1423, opened 2018-08-13
Fix test_client_intermediate and test_client_no_intermediate

pquentin:
See the commits for details.


pquentin:
Thanks. :)

(I think we'll continue to see failures on Windows, but this should hopefully fix the test on macOS.)

SethMichaelLarson:
:tada:

PR #1424, opened 2018-08-14
Testing with Python 3.6.5

oixhwotl:
Code changes to run sample code on Python 3.6.5

import urllib3
http = urllib3.PoolManager()
r = http.request('GET', 'https://httpbin.org/robots.txt')

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -251,6 +251,7 @@ class HTTPSConnection(HTTPConnection):
         self.cert_file = cert_file
         self.ssl_context = ssl_context
         self.server_hostname = server_hostname
+        self._dns_host = host
 
         # Required property for Google AppEngine 1.9.0 which otherwise causes
         # HTTPS requests to go out as HTTP. (See Issue #356)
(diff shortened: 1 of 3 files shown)


SethMichaelLarson:
Hello, thanks for submitting a PR. Could you give an explanation of what your change improves?

oixhwotl:
I will try to explain:
1. util/timeout.py: Handling Timeout object as value.
The control flows to "except TypeError:# Python 3", it's not because it is Python3 issue. It's because Timeout object is not handled properly. Now I think it's stupid to use hidden property inside the class method. It's better to call "value.read_timeout()" instead of accessing "value._read" directly.
2. connectionpool.py
It caused error on Python 3, and I can't understand the error message from the interpreter saying HTTPSConnecitonPool is not related to self. There is a simpler way in Python 3, so I add Python 3 syntax.
3. connection.py
The variable "_dns_host" is used to in line 170, but it is not assiged anywere. The better way is to resolve the issue inside HTTPConnection with "self._dns_host = kw.pop('host', None)".

Now I'm sorry to propose half-baked code to merge. I will just close this request, and you can add changes in a better way.

PR #1427, opened 2018-08-17
tox: report missed lines in terminal

pquentin:


pquentin:
Okay, now that there is a base commit with coverage info, codecov works better, even though the `__init__` file looks wrong: https://codecov.io/gh/urllib3/urllib3/pull/1427/src/src/urllib3/__init__.py

@SethMichaelLarson Are you able to require coverage in github? I thought https://github.com/urllib3/urllib3/blob/master/codecov.yml was enough, but looks like it isn't!

SethMichaelLarson:
Unfortunately I can't make it required, and yeah that `__init__.py` file definitely doesn't look right.

theacodes:
I can if we wanna do that.

pquentin:
Closing in favor of https://github.com/urllib3/urllib3/pull/1428. It think the wrong `__init__.py` was coming from the tests.

@theacodes We do want to require 100% coverage. That's what https://github.com/urllib3/urllib3/blob/master/codecov.yml is supposed to do, but maybe there's something to configure in GitHub?

Issue #883, opened 2016-05-31
Drop Python 2 support


haikuginger:
I'm wondering whether we should start a conversation with our biggest partners (pip, Requests, et. al.) around what the long-term plan for Python 2 ought to be.

The platform itself is planned to have no further support after 2020, so we should probably have a transition plan laid out well in advance so that we can start providing notification and documentation to our users.

Obviously, what that plan ought to be, or even whether we should start considering it yet, is up for discussion.
##

sigmavirus24:
Hey @haikuginger 

So I'm 100% on board with killing off Python 2 today, but we can't. We will likely have to continue supporting it until 2020 or past that date. urllib3 and requests are both (what is frequently called) "core infrastructure". This is because of pip's dependency on both. What this means to me is that until pip can stop supporting Python 2, we can not drop that support.

I don't think we can make a plan now, three years out, because there's three years of data to be collected around usage and that absolutely should factor into our decision making *then*.

If we want to make a plan to make a decision, that's fine by me. Deciding what our deprecation path now is imprudent. I don't think we'll be able to make a decision without that *data*.

shazow:
Chatted with @sigmavirus24 and folks IRL today, and it seems like a reasonable ambition to drop support for a Python in urllib3 v2.0 (added to the wishlist) which will be a big backwards-compat breaking set of changes.

Not sure when that will happen, though. Hopefully sometime before 2020.

haikuginger:
Sounds good - so we'll basically plan to release a 2.0 version with a brand-new codebase that drops Python 2.x support, but provide extended maintenance support for 1.x as long as pip needs it?

Just making sure I understand the situation; wish I could have made it out to PyCon, but there you are.

shazow:
For some definition of extended maintenance, yes. :P

haikuginger:
@sigmavirus24 @Lukasa, it looks like pip has dropped support for Python 2.6. How does the Requests team feel about doing so as well in coordination with urllib3?

nateprewitt:
@haikuginger the current decision was to drop Python 2.6 with Requests 3.0 and we have a PR doing so merged into that branch. I think some of the changes slated for Requests 3.0 were dependent on urllib3 2.0 though.

[...]

PR #1429, opened 2018-08-23
Drop support for EOL Python 2.6

hugovk:
For #883.

Requests and pip have already dropped Python 2.6.

This PR removes support for it, removes redundant code, and also does some modernising to use newer Python features.


src/urllib3/connection.py
```diff
@@ -108,21 +104,10 @@ def __init__(self, *args, **kw):
         if six.PY3:  # Python 3
             kw.pop('strict', None)
 
-        # Pre-set source_address in case we have an older Python like 2.6.
-        self.source_address = kw.get('source_address')
```
SethMichaelLarson: I think we should keep this attribute on the `HTTPConnection` object, might be being used by others.

src/urllib3/util/ssl_.py
```diff
@@ -115,10 +114,7 @@ def inet_pton(_, host):
 except ImportError:
     import sys
 
-    class SSLContext(object):  # Platform-specific: Python 2 & 3.1
-        supports_set_ciphers = ((2, 7) <= sys.version_info < (3,) or
-                                (3, 2) <= sys.version_info)
-
+    class SSLContext(object):  # Platform-specific: Python 2
```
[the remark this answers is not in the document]
hugovk: Reverted.

src/urllib3/connection.py
hugovk: Kept.

SethMichaelLarson:
Could you comment on how you found the occurrences for these optimizations? Was it a manual effort or is there a tool for the unittest improvements?

hugovk:
Sure, for the set/dict/list literals, I used the code inspection (Code > Inspect Code...) in PyCharm (they have a lightweight [Community edition](https://www.jetbrains.com/pycharm/download/) and also free licences for the [full version for open source projects](https://www.jetbrains.com/opensource/)).

For the unit tests, I used the script at https://github.com/hugovk/upgrade-unittest which is just some sed commands, not a full AST parser, so you need to check the changes. There's also the https://github.com/jparise/flake8-assertive plugin, which I've just run and found some others which I've updated.

SethMichaelLarson:
I'm approving these changes, just got to let CI pass now. :)

SethMichaelLarson:
Thanks for doing this! :tada:

nateprewitt:
As a quick note, we dropped support for 2.6 but haven’t merged the removal changes into master since that would be a breaking change for Requests 2.X.

We can probably go ahead and drop it, but we’ll need to make an announcement that it breaks now.

hugovk:
You're welcome!

pquentin:
Thank you again @hugovk, that was a quick and high-quality PR. :)

Would you be interested in working on [urllib3 async support](https://github.com/urllib3/urllib3/pull/1335)?

hugovk:
Thanks, happy to help out, but I'm not yet really familiar with async.

SethMichaelLarson:
Also @hugovk since you did the bulk of this work could you add a changelog entry for dropping Python 2.6 and add yourself to contributors list if you're not already there.

hugovk:
@SethMichaelLarson Done! Please see https://github.com/urllib3/urllib3/pull/1431.

pquentin:
@hugovk There are many ways to help even if you're not familiar with async yet! Removing Python 2.6 support was one of those things. :)

Another helpful thing to do is to improve the setup.py script: https://github.com/python-trio/trio/issues/567#issuecomment-408583539. There are many things to do but they are not documented yet.

Feel free to open an issue in https://github.com/python-trio/urllib3/ or send me an email at quentin.pradet@gmail.com if anything is unclear.

But no pressure!

PR #1440, opened 2018-09-14
Remove unnecessary use of six.binary_type and six.b

jdufresne:
The `bytes` type is available on all support Pythons. On Python 2 it is an alias of `str` (same as six). Likewise, the `bytes` literal is also available. Reduce unnecessary compatibility shims and by using modern Python idioms. Makes the code more forward compatible with Python 3.


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -50,7 +50,7 @@ def _test_body(self, data):
 
         self.assertIn(b'Transfer-Encoding: chunked', header.split(b'\r\n'))
         if data:
-            bdata = data if isinstance(data, six.binary_type) else data.encode('utf-8')
+            bdata = data if isinstance(data, bytes) else data.encode('utf-8')
```
SethMichaelLarson: Now that this is removed we can remove the import of `six` within this file.

jdufresne:
Yup! Thanks for catching. I've removed the unused import.

PR #1442, opened 2018-09-17
Handle multiple Content-Encodings

### LINKED ISSUE #1441: Content-Encoding with multiple encodings not supported
It is possible for a server to respond with a Content-Encoding header like:
`Content-Encoding: deflate, deflate`

From RFC7231:
> If one or more encodings have been applied to a representation, the
> sender that applied the encodings MUST generate a Content-Encoding
> header field that lists the content codings in the order in which
> they were applied.

This can be tested with:

```python
def test_multiple_content_encodings():
    data = zlib.compress(zlib.compress(b'foo'))
    fp = BytesIO(data)
    r = HTTPResponse(fp, headers={'Content-Encoding': 'deflate, deflate'})
    assert r.data == b'foo'
````

Would it make sense to do something like:
````python
if len(self.headers.get('content-encoding').split(',') > 1:
    self._decoder = MultiDecoder(self.headers.get('content-encoding'))
````
where MultiDecoder just iterates through the list applying standard decoders?

jbramleycl:
Add in ability to handle multiple content encodings (e.g. Content-Encoding: deflate, deflate).

This resolves #1441.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 dev (master)
 ------------
 
+* Add ability to handle multiple Content-Encodings (Issue #1441 and Pull #1442)
+
 * Skip DNS names that can't be idna-decoded when using pyOpenSSL (Issue #1405).
 
 * Add a server_hostname parameter to HTTPSConnection which allows for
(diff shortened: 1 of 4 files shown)


SethMichaelLarson:
@theacodes Can we make the `codecov/project` status required for the `master` branch? [Codecov is showing 100% on this PR](https://codecov.io/gh/urllib3/urllib3/pull/1442/tree/src/urllib3) for the record before I merge this.

SethMichaelLarson:
Thanks @jbramleycl! :tada:

PR #1448, opened 2018-10-04
Remove leftover Python 2.6 variable

pquentin:
set_ciphers is always supported in Python 2.7+, so this is no longer needed.

(I noticed this because the v2 branch dropped supports_set_ciphers when dropping Python 2.6 support.)


SethMichaelLarson:
Thanks! :tada:

Issue #1455, opened 2018-10-16
problems testing with pluggy 0.8.0


caboteria:
Pluggy 0.8.0 was released today and it is causing my tox runs to fail.  I get:

```
$ tox -e py36
GLOB sdist-make: /home/tcabot/Code/urllib3/setup.py
py36 inst-nodeps: /home/tcabot/Code/urllib3/.tox/dist/urllib3-dev.zip
py36 installed: /home/tcabot/Code/urllib3/.tox/py36/lib/python3.6/distutils/__init__.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses,  import imp,asn1crypto==0.24.0,astroid==2.0.4,atomicwrites==1.2.1,attrs==18.2.0,certifi==2018.10.15,cffi==1.11.5,chardet==3.0.4,coverage==4.5.1,cryptography==2.3.1,gcp-devrel-py-tools==0.0.15,idna==2.7,ipaddress==1.0.22,isort==4.3.4,lazy-object-proxy==1.3.1,mccabe==0.6.1,mock==2.0.0,more-itertools==4.3.0,packaging==18.0,pbr==4.3.0,pkginfo==1.4.2,pluggy==0.8.0,py==1.7.0,pycparser==2.19,pylint==2.1.1,pyOpenSSL==18.0.0,pyparsing==2.2.2,PySocks==1.6.8,pytest==3.6.4,pytest-timeout==1.3.1,requests==2.19.1,requests-toolbelt==0.8.0,retrying==1.3.3,six==1.11.0,tornado==5.0.2,tox==2.9.1,tqdm==4.27.0,twine==1.11.0,typed-ast==1.1.0,urllib3===dev,virtualenv==16.0.0,wrapt==1.10.11
py36 runtests: PYTHONHASHSEED='4079776103'
py36 runtests: commands[0] | pip --version
/home/tcabot/Code/urllib3/.tox/py36/lib/python3.6/distutils/__init__.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
pip 18.1 from /home/tcabot/Code/urllib3/.tox/py36/lib/python3.6/site-packages/pip (python 3.6)
py36 runtests: commands[1] | python --version
Python 3.6.6
py36 runtests: commands[2] | python -c import struct; print(struct.calcsize('P') * 8)
64
[...]

PR #1457, opened 2018-10-16
Resolve pytest pluggy version conflict

caboteria:
This resolves issue #1455 .

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -11,3 +11,4 @@ pytest==3.8.2
 # https://github.com/GoogleCloudPlatform/python-repo-tools/issues/23
 pylint<2.0;python_version<="2.7"
 gcp-devrel-py-tools==0.0.15
+pluggy==0.8.0
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,7 +7,7 @@ tornado==5.0.2
 PySocks==1.6.8
 pkginfo==1.4.2
 pytest-timeout==1.3.1
-pytest==3.6.4
+pytest==3.8.2
 # https://github.com/GoogleCloudPlatform/python-repo-tools/issues/23
 pylint<2.0;python_version<="2.7"
 gcp-devrel-py-tools==0.0.15


dev-requirements.txt
```diff
@@ -7,7 +7,8 @@ tornado==5.0.2
 PySocks==1.6.8
 pkginfo==1.4.2
 pytest-timeout==1.3.1
-pytest==3.6.4
+pytest==3.8.2
 # https://github.com/GoogleCloudPlatform/python-repo-tools/issues/23
 pylint<2.0;python_version<="2.7"
 gcp-devrel-py-tools==0.0.15
+pluggy==0.8.0
```
SethMichaelLarson:
```suggestion
```

SethMichaelLarson: Can we get away with not pinning pluggy here?

caboteria: We can get away with not pinning pluggy, but keep in mind that a new version of pluggy is what broke the build yesterday.  I saw the other requirements that were pinned to specific versions and inferred that this project likes to have stable dependencies but the build will succeed without it.

caboteria: The commit that upgrades pytest (and fixes the build) is 41a665a and the one that pins pluggy (which is just future-proofing) is 2f21ef8 so if you *don't* want to pin pluggy you can merge 41a665a but if you *do* you can merge 2f21ef8.

pquentin: Arguably that was a bug in pytest 3.6.4, which should have pinned pluggy, just like requests pins urllib3. I think we either want to go the full pip-compile + dependabot route, or just list the dependencies we use, which means we'll get occasional breakage like this. Of course, the maintainers decide. :)

caboteria: Pytest 3.6.4 did require a range for pluggy (>=0.5,<0.8) but some other module that was loaded sooner caused pluggy 0.8.0 to load so when pytest asked for <0.8 things got crashy.  The pain of "DLL hell" is still real.

pquentin: Oh, right. tox 2.9.1 requires `pluggy>=0.3.0,<1.0` and is placed before pytest, which is why pluggy 0.8.0 was installed. Really, the only real way to fix this is to use pip-compile, and possibly dependabot to help with the constant updates.

SethMichaelLarson: I think merging this as-is is fine for now. We'll have to revisit our whole reproducible build strategy collectively to solve this problem completely :)

PR #1463, opened 2018-10-23
Restore context.set_ciphers() to create_urllib3_context()

### LINKED ISSUE #1462: Can no longer create a context with specific ciphers
The recent change to remove support for Python 2.6 has made it so that the "ciphers" parameter of the create_urllib3_context function in ssl_.py no longer has any effect.  The parameter is unused for the rest of the function.

This was working properly in Python 3.6, so I don't know why it was tagged as being Python 2.6 specific.

SethMichaelLarson:
Fixes #1462.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Changes
 dev (master)
 ------------
 
+* Restored functionality of `ciphers` parameter for `create_urllib3_context()`. (Issue #1462)
+
 * ... [Short description of non-trivial change.] (Issue #)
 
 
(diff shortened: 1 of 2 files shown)


SethMichaelLarson:
Looks like gdal has gcc as a dependency now (it didn't in previous installs) and is failing because gcc is already installed on the system and it can't overwrite it?

test/test_ssl.py
```diff
@@ -70,3 +70,22 @@ def test_sni_missing_warning_with_ip_addresses(monkeypatch, has_sni, server_host
...
+    context = mock.create_autospec(ssl_.SSLContext)
+    context.set_ciphers = mock.Mock()
+    context.options = 0
+    monkeypatch.setattr(ssl_, "SSLContext", lambda *_, **__: context)
+
+    assert ssl_.create_urllib3_context(ciphers=ciphers) is context
+
+    assert context.set_ciphers.call_count == 1
+    assert context.set_ciphers.call_args == mock.call(expected_ciphers)
+
```
pquentin: It looks like CI is failing because of this empty line.

Issue #1465, opened 2018-10-25
Any interest in an UnixHTTPConnectionPool?


mattrobenolt:
I'm happy to create a PR, but it seems like a pretty common use case to want to establish an HTTP connection over a unix socket.

I see this abstraction happening more commonly through `requests`, but was a little surprised that this doesn't just exist as a connection within urllib3.

The code I've thrown together (not yet battle tested) is pretty trivial to do when you get rid of the `requests` bits, and it's boiled down to just:

```python
class UnixHTTPConnection(HTTPConnection):
    def __init__(self, host, **kwargs):
        # We're using the `host` as the socket path, but
        # urllib3 uses this host as the Host header by default.
        # If we send along the socket path as a Host header, this is
        # never what you want and would typically be malformed value.
        # So we fake this by sending along `localhost` by default as
        # other libraries do.
        self.socket_path = host
        super(UnixHTTPConnection, self).__init__(host='localhost', **kwargs)

    def _new_conn(self):
        sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
        if self.timeout is not socket._GLOBAL_DEFAULT_TIMEOUT:
            sock.settimeout(self.timeout)
        sock.connect(self.socket_path)
        return sock


class UnixHTTPConnectionPool(HTTPConnectionPool):
    ConnectionCls = UnixHTTPConnection
```

If this is something you'd be interested in pulling into `urllib3` as a core connection pool, I think this would be great to have, and I'll convert this into a PR. If not, would be curious if this seems outside the scope of urllib3 or some other reason.

The sources of this off the top of my head are within docker-py, which again, uses this sorta abstraction except a bit heavy handed since it also needs the `requests` adapter https://github.com/docker/docker-py/blob/master/docker/transport/unixconn.py. Then requests-unixsocket project https://github.com/msabramo/requests-unixsocket.

And again, nothing explicitly for those of us that like just directly using urllib3.

I plan on adding this into Sentry very soon, and will cut over to using this for some of our critical internal services to get it battle tested.

I also noticed that this is sorta being used within test suite: https://github.com/urllib3/urllib3/blob/0f85e05af9ef2ded671a7b47506dfd24b32decf0/test/test_connectionpool.py#L35-L45 I'm not entirely sure of the context here, but it's definitely just used for some mock class. Seems relevant though. :)

mattrobenolt:
That's fine with me. Like I said, was just curious if there was interest in having support for this in core. :)

[...]

PR #1469, opened 2018-11-01
WIP Unix domain socket support

mattrobenolt:
Addressing #1465

From my perspective, this seems pretty feature complete. We've also been using our version of this in production for a bit now without any issues. (https://github.com/getsentry/sentry/blob/master/src/sentry/net/http.py#L156)

I've made a couple design decisions based on the discussion in the other ticket:

* Added a `UnixHTTPPoolManager` which provides a `http+unix` scheme and a URI quoted file path so a URI like `http+unix://%2Ftmp%2Fserver.sock/hello` works. This also subclasses normal `PoolManager` so technically this one could just be swapped in to get full http/http/http+unix support.

* Two different `ConnectionPool` classes. One takes the URI quoted version of the path (`UnixHostHTTPConnectionPool`) and one that takes `socket_path` directly (`UnixHTTPConnectionPool`). The first one is to support the API needed for the `PoolManager` and is simply responsible for unquoting the path and passing it along.

I have started modifying the test suite to accommodate a Unix domain socket dummyserver, but haven't added any real tests yet.

I'm ready for feedback on APIs here and implementation, and will spend time figuring out tests after we are comfortable. I've tested stuff manually so far, so it's working in current state.

Thanks everyone! ❤️

### DIFF
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -20,6 +20,7 @@ from datetime import datetime
 from urllib3.exceptions import HTTPWarning
 
 from tornado.platform.auto import set_close_exec
+from tornado.netutil import bind_unix_socket
 import tornado.httpserver
 import tornado.ioloop
 import tornado.web
@@ -237,8 +238,12 @@ def run_tornado_app(app, io_loop, certs, scheme, host):
     else:
         http_server = tornado.httpserver.HTTPServer(app)
 
-    sockets = bind_sockets(None, address=host)
-    port = sockets[0].getsockname()[1]
+    if host[:1] == '/':
+        sockets = [bind_unix_socket(host)]
+        port = None
+    else:
+        sockets = bind_sockets(None, address=host)
+        port = sockets[0].getsockname()[1]
     http_server.add_sockets(sockets)
     return http_server, port
 
@@ -267,7 +272,8 @@ def get_unreachable_address():
 if __name__ == '__main__':
     # For debugging dummyserver itself - python -m dummyserver.server
     from .testcase import TestingApp
-    host = '127.0.0.1'
+    # host = '127.0.0.1'
+    host = '/tmp/dummyserver.sock'
 
     io_loop = tornado.ioloop.IOLoop.current()
     app = tornado.web.Application([(r".*", TestingApp)])
@@ -275,4 +281,7 @@ if __name__ == '__main__':
                                    'http', host)
     server_thread = run_loop_in_thread(io_loop)
 
-    print("Listening on http://{host}:{port}".format(host=host, port=port))
+    if port:
+        print("Listening on http://{host}:{port}".format(host=host, port=port))
+    else:
+        print("Listening on unix@{host}".format(host=host))
--- a/src/urllib3/contrib/unix.py
+++ b/src/urllib3/contrib/unix.py
@@ -1,0 +1,107 @@
+from __future__ import absolute_import
+import logging
+import socket
+
+from ..connection import HTTPConnection
+from ..connectionpool import HTTPConnectionPool
+from ..packages.six.moves.urllib.parse import unquote_plus
+from ..poolmanager import PoolManager
+from ..util.connection import _set_socket_options
+
+
+log = logging.getLogger(__name__)
+
+
+class UnixHTTPConnection(HTTPConnection):
+    """
+    An HTTP connection via a Unix domain socket.
+    """
+    default_socket_options = []
+
+    def __init__(self, socket_path, **kwargs):
+        self.socket_path = socket_path
+        # host needs to be sent along as a fake value in order
+        # to be used as the HTTP Host header when one isn't supplied.
+        # Since this is a unix socket, there's no sensible default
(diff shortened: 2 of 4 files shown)


pquentin:
@mattrobenolt There's a flake8 failure left, and I guess you need to make sure Windows does not see this feature!

mattrobenolt:
Yeah, I also don’t have tests. Hence the WIP. Was hoping to get early feedback before investing more time and people disagreeing with the direction. If all is good, I’m happy to polish things up.

mattrobenolt:
For sure, just wanted to actually get the code out there since the direction was theoretical without an implementation. Now I have an implementation.

theacodes:
The general idea and approach is good with me. :)

SethMichaelLarson:
The implementation looks good to me too, just waiting for some test cases. :) And I know this is a module we can probably cover 100% so maybe we need to be more granular when it comes to exempting things within `urllib3/contrib`?

mattrobenolt:
Cool, I'll get things wrapped up here.

With regards to Windows, obviously we can skip tests for Windows since it won't work, but do we have opinions about just letting the entire class just explode if someone tries to use it on Windows? Or should we be nicer and provide a stub class that says something like, "Unix sockets don't work on Windows". Or just let the module itself ImportError if trying to import on Windows.

I also don't have access to a Windows machine, so will be flying a bit blind there.

SethMichaelLarson:
Going from our `urllib3.contrib.socks` it looks like we give a warning if we can't import things we need (in this case `AF_UNIX`)? Is this path something we want to continue?

sethmlarson:
@mattrobenolt did you still want to land this patch?

dummyserver/server.py
```diff
@@ -267,12 +272,16 @@ def get_unreachable_address():
 if __name__ == '__main__':
     # For debugging dummyserver itself - python -m dummyserver.server
     from .testcase import TestingApp
-    host = '127.0.0.1'
+    # host = '127.0.0.1'
```
sethmlarson: Let's revert this change.

dummyserver/server.py
```diff
@@ -237,8 +238,12 @@ def run_tornado_app(app, io_loop, certs, scheme, host):
     else:
         http_server = tornado.httpserver.HTTPServer(app)
 
-    sockets = bind_sockets(None, address=host)
-    port = sockets[0].getsockname()[1]
+    if host[:1] == '/':
```
sethmlarson: [:1] is the same as [0] so let's use the latter for better readability

dummyserver/server.py
```diff
@@ -267,12 +272,16 @@ def get_unreachable_address():
...
     app = tornado.web.Application([(r".*", TestingApp)])
     server, port = run_tornado_app(app, io_loop, None,
                                    'http', host)
     server_thread = run_loop_in_thread(io_loop)
 
-    print("Listening on http://{host}:{port}".format(host=host, port=port))
+    if port:
+        print("Listening on http://{host}:{port}".format(host=host, port=port))
+    else:
+        print("Listening on unix@{host}".format(host=host))
```
sethmlarson: Let's change this scheme for this message to be the same as what we're using as our scheme below (ie `http+unix`)

test/with_dummyserver/test_unix.py
```diff
@@ -0,0 +1,20 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        cls._stop_server()
+        try:
+            os.remove(cls.host)
+        except OSError:
+            pass
+
+
+class TestUnixHTTPConnectionPool(UnixHTTPDummyServerTestCase):
```
sethmlarson: Can we somehow reuse the tests already written for our HTTPConnectionPool here? Would be great to have them generic enough to reuse.

dummyserver/server.py
sethmlarson: I agree they don't have the same behavior in general, but isn't the type of the `host` parameter in Python 3 a `str` not `bytes` here and empty strings invalid hosts anyways?

sethmlarson:
Take your time, thanks for tackling this for us!

src/urllib3/contrib/unix.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    A thread-safe connection pool for one Unix socket, addressed by URI.
+    :param host:
+        URI quoted path to be used for this HTTP Connection (e.g. "%2Fvar%2Frun%2Fserver.sock").
+    """
+    scheme = 'http+unix'
+
+    # port exists for API compatibility, but is ignored
+    def __init__(self, host, port=None, **conn_kw):
+        socket_path = unquote_plus(host)
```
sethmlarson: I don't think `unquote_plus` is what we want here, `+` is allowed in the authority section of URLs and we don't want to decode a `+` to a space character.

pquentin:
Closing as this pull request is now stale (and needs changes).

@mattrobenolt Feel free to reopen if you resume this work.

If someone else is interested, please take this work and open a new pull request. Thanks!

Issue #1470, opened 2018-11-02
contrib/_appengine_environ.py KeyError SERVER_SOFTWARE


wittfabian:
https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/_appengine_environ.py

In the development environment, the variable SERVER_SOFTWARE is sometimes not set, so the query should read as follows:

`os.environ.get('SERVER_SOFTWARE', '').startswith('Development')`

```
def is_local_appengine():
    return ('APPENGINE_RUNTIME' in os.environ and
            os.environ.get('SERVER_SOFTWARE', '').startswith('Development'))


def is_prod_appengine():
    return ('APPENGINE_RUNTIME' in os.environ and
            not os.environ.get('SERVER_SOFTWARE', '').startswith('Development') and
            not is_prod_appengine_mvms())
```

SethMichaelLarson:
I would accept a PR that makes this change. cc: @theacodes

theacodes:
Yeah, fine with me. I'd be curious to know when the local server doesn't set the `SERVER_SOFTWARE` and what effect that has on detection.

wittfabian:
The Travis CI build could not complete ...

wittfabian:
Any innovations here? @pquentin @theacodes

zevdg:
I just ran into this bug, (simplified reproduction of how I ran into it https://repl.it/repls/ThinSpatialNumerator) and was about to submit a fix much like the two that have already been proposed.

There is some new, and possibly relevant information.  This function

```python
def is_prod_appengine_mvms():
    return os.environ.get('GAE_VM', False) == 'true'
```
seems to always returns False now.

App Engine stopped setting this variable and [removed it from their PHP docs](https://github.com/GoogleCloudPlatform/getting-started-php/pull/200/files) (I couldn't find any references to it in the python docs).

It looks like the reasons the original CL never made it in seem to be because they didn't work quite right on flex, but since the flex check isn't working anymore anyway, and no one seems to be complaining about it, I suggest we move forward with the fix now.

PR #1471, opened 2018-11-03
Update _appengine_environ.py

wittfabian:
Fixed Bug https://github.com/urllib3/urllib3/issues/1470


src/urllib3/contrib/_appengine_environ.py
```diff
@@ -17,12 +17,12 @@ def is_appengine_sandbox():
...
 def is_local_appengine():
     return ('APPENGINE_RUNTIME' in os.environ and
-            'Development/' in os.environ['SERVER_SOFTWARE'])
+            os.environ.get('SERVER_SOFTWARE', '').startswith('Development'))
 
 
 def is_prod_appengine():
     return ('APPENGINE_RUNTIME' in os.environ and
-            'Google App Engine/' in os.environ['SERVER_SOFTWARE'] and
+            not os.environ.get('SERVER_SOFTWARE', '').startswith('Development') and
```
SethMichaelLarson: This is a different behavior than the one @theacodes had originally so I'll have her review it. :)

wittfabian:
An other possibility would be

https://github.com/potatolondon/djangae/blob/master/djangae/environment.py#L50

src/urllib3/contrib/_appengine_environ.py
theacodes: I would say let's just `not is_local_appengine()`?

pquentin:
The tests failures don't seem related to the changes, so maybe retrying will fix the failures?

wittfabian:
Who can restart the tests?

pquentin:
Members/owners can do it. You can also do it by closing/reopening the pull request!

pquentin:
Maybe the issue is that is_prod_appengine is now returning True in situations where it returned False in the past?

SethMichaelLarson:
That's totally it, thanks for the sanity check @pquentin :)

wittfabian:
Maybe APPENGINE_RUNTIME is the solution.

src/urllib3/contrib/_appengine_environ.py
```diff
@@ -6,24 +6,20 @@
...
-            'Development/' in os.environ['SERVER_SOFTWARE'])
+    return ('SERVER_SOFTWARE' not in os.environ or
+            os.environ['SERVER_SOFTWARE'].startswith('Development'))
 
 
 def is_prod_appengine():
-    return ('APPENGINE_RUNTIME' in os.environ and
-            'Google App Engine/' in os.environ['SERVER_SOFTWARE'] and
-            not is_prod_appengine_mvms())
+    return not is_local_appengine()
```
SethMichaelLarson: This now returns `True` if we're not in an AppEngine environment where it previously didn't. Maybe we should retain the `('APPENGINE_RUNTIME in os.environ' and ...` logic to this function?

src/urllib3/contrib/_appengine_environ.py
```diff
@@ -6,24 +6,20 @@
...
 
 
 def is_appengine_sandbox():
     return is_appengine() and not is_prod_appengine_mvms()
 
 
 def is_local_appengine():
-    return ('APPENGINE_RUNTIME' in os.environ and
-            'Development/' in os.environ['SERVER_SOFTWARE'])
+    return ('SERVER_SOFTWARE' not in os.environ or
```
SethMichaelLarson: Let's keep the `'APPENGINE_RUNTIME'` check here as well.

SethMichaelLarson:
[Looks like a syntax error in latest commit](https://ci.appveyor.com/project/urllib3/urllib3/builds/20563904/job/gtgqnd1d3l9tslmh#L78)

SethMichaelLarson:
I'll wait for @theacodes for final review of this. Thanks! :)

src/urllib3/contrib/_appengine_environ.py
```diff
@@ -6,24 +6,21 @@
...
 
 
 def is_appengine_sandbox():
     return is_appengine() and not is_prod_appengine_mvms()
 
 
 def is_local_appengine():
-    return ('APPENGINE_RUNTIME' in os.environ and
-            'Development/' in os.environ['SERVER_SOFTWARE'])
+    return is_appengine() and \
```
theacodes: This isn't going to work right. This will return `True` for Flex as `SERVER_SOFTWARE` won't be set but `APPENGINE_RUNTIME` will be. We either need to *require* `SERVER_SOFTWARE` as part of the local check or find a more reliable way to determine whether or not you're in the local environment.

wittfabian: Why `is_local_appengine` needs to know weather it is an flex environment or not?
For this distinction we use `is_prod_appengine_mvms`.

wittfabian: The alternative is to add `not is_prod_appengine_mvms` to `is_prod_appengine` and `is_local_appengine`

theacodes: Basically each of these checks needs to be definitive:

`is_local_appengine` should return True if and only if you're on the local app engine dev server.

`is_prod_appenegine` should return Tue if and only if you're in the production app engine (standard or flex).

and so on. Does that make sense?

wittfabian: What about the import check?

PR #1472, opened 2018-11-04
Use unittest.mock if available

opoplawski:
Mock is in python 3.3+, use that if possible.  Make mock package conditional on python < 3.3.  Specify setuptools version requirement for use of environment markers.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,10 @@ sys.path.insert(0, root_path)
 # Mock some expensive/platform-specific modules so build will work.
 # (https://read-the-docs.readthedocs.io/en/latest/faq.html#\
 #  i-get-import-errors-on-libraries-that-depend-on-c-modules)
-import mock
+try:
+    import unittest.mock as mock
+except ImportError:
+    import mock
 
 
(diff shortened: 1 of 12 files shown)


SethMichaelLarson:
Hey @opoplawski, thanks for opening this change! For the sake of consistency (and because we'll have to do this work again in the future anyways) I'd like to make this change to use only `unittest.mock` when we move to Python 3.x only. What do you think?

opoplawski:
Well, it's currently a pain keeping the separate mock package around in distributions so it would be nice to be able to drop it now if possible.  But I can easily make the change just in the urllib3 downstream package.  (I'm dealing with packaging urllib3 for Fedora EPEL)

sigmavirus24:
Please don't do this. The `mock` package contains backports of new features and bug-fixes to older versions of Python.

The second you start doing this you run into the same issues that unittest has always had:

* restraining ourselves to methods only present on the oldest version
* writing shim methods to fix the wrong behaviour on older versions
* etc

It's easier for us, if we can rely on one source of truth `mock` rather than `mincompatibility(unittest.mock)` + whatever shims we add on top.

PR #1478, opened 2018-11-23
Uninstall oclint if it's installed

SethMichaelLarson:
Doing this because I saw on the Python 3.8 build on master we were having this issue again. I think it's because the builds get cached on success so we need to check if it's installed every time because caches go away on PRs and new branches.


SethMichaelLarson:
Ah, I missed that message, Travis must have realized that package was causing a lot of problems. :)

PR #1479, opened 2018-11-23
Allow missing SERVER_SOFTWARE in appengine checks

pquentin:
Closes https://github.com/urllib3/urllib3/issues/1470

I've taken inspiration from https://github.com/urllib3/urllib3/issues/1470, https://github.com/urllib3/urllib3/pull/1471 and https://cloud.google.com/appengine/docs/standard/python/tools/using-local-server#detecting_application_runtime_environment. The tests pass, but does it fix @wittfabian's original issue? I don't know.

### DIFF
--- a/src/urllib3/contrib/_appengine_environ.py
+++ b/src/urllib3/contrib/_appengine_environ.py
@@ -17,12 +17,12 @@ def is_appengine_sandbox():
 
 def is_local_appengine():
     return ('APPENGINE_RUNTIME' in os.environ and
-            'Development/' in os.environ['SERVER_SOFTWARE'])
+            os.getenv('SERVER_SOFTWARE', '').startswith('Development'))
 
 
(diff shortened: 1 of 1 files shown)


wittfabian:
This corresponds to the original solution from #1470.

pquentin:
@wittfabian It does! But does it help with your problem?

wittfabian:
No, because if SERVER_SOFTWARE is not defined, both methods return False.

See: https://github.com/potatolondon/djangae/blob/master/djangae/environment.py#L50

pquentin:
Okay, I understand now. So we want to support that, we need to make sure that having `is_local_appengine` return True does not break anything else. Indeed, at least in tests, contrary to djangae, we not only want to differentiate local appengine from production appengine, but we also want to detect "not running in appengine at all".

PR #1480, opened 2018-11-26
Fix inject/extract calls in tests

SethMichaelLarson:
Looks like we're injecting and extracting the wrong contrib module in some of our `test/contrib/` tests.


SethMichaelLarson:
Transient error on AppVeyor

PR #1484, opened 2018-11-28
Create a more efficient HTTPResponse.__iter__() method

### LINKED ISSUE #1483: Maybe it should be explicitly documented that iterating over an HTTPResponse is bad
HTTPResponse inherits from io.IOBase, meaning it's iterable. By python's documentation this means it iterates over lines in the response.

The problem is that it calls IOBase's readline, which in turn **calls the object's read method with size of 1, until it gets a newline.**

While discovering this when sending a huge payload and figuring out why its slow, this might not be obvious to people using the library and by accident calling:
for chunk in response:
  ...
instead of
for chunk in response.stream(sane_size):
  ...
Maybe iterating over the response should be disallowed, or documented that it's bad to do so.

SethMichaelLarson:
Per #1483 if you iterate over the response object you get single-byte reads over and over again. This implementation maintains the same functionality of splitting by line but doesn't use single-byte reads. Closes #1483.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,11 +4,17 @@ Changes
 dev (master)
 ------------
 
+* Implemented a more efficient ``HTTPResponse.__iter__()`` method (Issue #1483)
+
+* ... [Short description of non-trivial change.] (Issue #)
+
+
+1.24.1 (2018-11-02)
+-------------------
+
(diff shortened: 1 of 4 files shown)


SethMichaelLarson:
Sorry @tzickel I didn't see your original comment before merging. I'll explore using io.BufferedIOBase here.

tzickel:
np, let's keep the issue open till then ?

PR #1485, opened 2018-11-28
travis: use python.org MacPython builds

pquentin:
It's about ten minutes faster than using homebrew, but means we need to drop 3.8-dev on macOS and only run test_ca_dir_verified on recent MacPython builds (currently 2.7, 3,6 and 3.7).

Inspiration comes from https://github.com/python-trio/trio/blob/master/ci/travis.sh#L7-L17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,9 +58,6 @@ matrix:
     - language: generic
       os: osx
       env: TOXENV=py37
-    - language: generic
-      os: osx
-      env: TOXENV=py38
   allow_failures:
   - python: pypy-5.4
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ script:
 before_install:
   - env
   - openssl version
+  - python -c "import ssl; print(ssl.OPENSSL_VERSION)"
 after_success:
   - ./_travis/upload_coverage.sh
 cache:
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,4 +1,5 @@
 import warnings
+import platform
 import sys
 import errno
 import functools
@@ -135,6 +136,17 @@ def fails_on_travis_gce(test):
     def wrapper(*args, **kwargs):
         if os.environ.get("TRAVIS_INFRA") == "gce":
             pytest.xfail("%s is expected to fail on Travis GCE builds" % test.__name__)
+        return test(*args, **kwargs)
+    return wrapper
+
+
+def requires_bundled_OpenSSL_on_mac(test):
+    """Skips this test unless your MacPython bundles OpenSSL"""
+
+    @functools.wraps(test)
+    def wrapper(*args, **kwargs):
+        if platform.system() == 'Darwin' and sys.version_info[0:2] < (3, 6):
+            pytest.skip("{name} requires MacPython >= 3.6".format(name=test.__name__))
         return test(*args, **kwargs)
     return wrapper
 
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -145,8 +145,10 @@ def requires_bundled_OpenSSL_on_mac(test):
 
     @functools.wraps(test)
     def wrapper(*args, **kwargs):
-        if platform.system() == 'Darwin' and sys.version_info[0:2] < (3, 6):
-            pytest.skip("{name} requires MacPython >= 3.6".format(name=test.__name__))
+        major, minor = sys.version_info[0:2]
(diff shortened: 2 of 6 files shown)


pquentin:
AppVeyor + Python 3.7 failed on `TestSocketClosing.test_connection_cleanup_on_read_timeout`.

test/__init__.py
```diff
@@ -139,6 +140,17 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def requires_bundled_OpenSSL_on_mac(test):
+    """Skips this test unless your MacPython bundles OpenSSL"""
+
+    @functools.wraps(test)
+    def wrapper(*args, **kwargs):
+        if platform.system() == 'Darwin' and sys.version_info[0:2] < (3, 6):
```
SethMichaelLarson: Should we allow Python 2.7 as well per your comment?

.travis.yml
```diff
@@ -58,9 +58,6 @@ matrix:
     - language: generic
```
SethMichaelLarson: Within the `before_install` section can we get a `python -c "import ssl; print(ssl.OPENSSL_VERSION)"`

SethMichaelLarson:
Also why can we only run with ca_dir on macOS Python builds with bundled OpenSSL?

pquentin:
macOS only provides OpenSSL/LibreSSL for compatibility reasons, but they're really out of date and should no longer be used. For this reason, the macOS Python 3.6+ (https://github.com/python/cpython/commit/dde4f63a54a75e75cdd08a40ea27e08353317e56#diff-460ef58e151f754f20075b855d145976) and Python 2.7.15+ (https://github.com/python/cpython/commit/ee8e4b61d6181f5bfadf6ee3d229f0ebda8a6ae9) from python.org now come with a more recent OpenSSL version bundled. Here are the OpenSSL versions in macOS for all the Python versions we support:

- 3.4.4 - OpenSSL 0.9.8zh 14 Jan 2016
- 3.5.4 - OpenSSL 0.9.8zh 14 Jan 2016
- 2.7.15 - OpenSSL 1.0.2o  27 Mar 2018
- 3.6.7 - OpenSSL 1.0.2p  14 Aug 2018
- 3.7.1 - OpenSSL 1.1.0i  14 Aug 2018

This means that between OpenSSL 0.9.8zh and OpenSSL 1.0.20, something happened that fixed test_ca_dir_verified. That test calls SSL_CTX_load_verify_locations with a custom directory, and even though the 0.9.8zh docs say it's supported, the verification fails with OpenSSL 0.9.8zh and works with OpenSSL 1.0.2o.

Maybe `requires_bundled_OpenSSL_on_mac` should be clarified?

test/__init__.py
pquentin: Thanks! Fixed.

.travis.yml
pquentin: In the `before_install`, this would show the default Python version, so I added it in _travis/install.sh instead. What do you think?

pquentin:
(By the way, the total build time for all versions decreases from 30 minutes to 10 minutes with this PR.)

pquentin:
TestSocketClosing.test_release_conn_param_is_respected_after_timeout_retry failed on macOS Python 3.5 and TestSocketClosing.test_connection_cleanup_on_read_timeout failed on macOS Python 3.6.

test/__init__.py
```diff
@@ -139,6 +140,19 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def requires_bundled_OpenSSL_on_mac(test):
```
SethMichaelLarson: Could we change the name of this decorator and make it an xfail instead of a skip? Also let's be clearer on where this fails, so far we've seen that it fails on macOS with OpenSSL < 1.0.0? Let's test for that rather than a specific Python version.

SethMichaelLarson:
Another issue I see with this is that we're not using OpenSSL 1.1 except on Python 3.7. Is that the first Python to expose functionality for 1.1?

test/__init__.py
pquentin: Thanks, fixed. It's not fun to parse the version robustly in a string like "OpenSSL 0.9.8zh 14 Jan 2016" and I guess it's better to be too specific than too broad so I now simply match that 0.9.8zh string.

.travis.yml
SethMichaelLarson: I think it's exactly what I actually wanted :) Thanks.

PR #1487, opened 2018-11-29
Implement RFC 3986 URL parsing

### LINKED ISSUE #466: Url: handle parsing of "oddball schemes" (data, magnet etc)
In `requests`, there has been a commit added, which skips urllib3's `parse_url` in certain cases: https://github.com/kennethreitz/requests/commit/b149be5d864cc7f65ac22113db3d0e4d2ed2b49e

It's meant to not cause e.g. a `LocationParseError`, but then also accidentally skips e.g. `localhost:3128`.

While that issue is specific to `requests`, I think it would be nice if `url_parse` would handle these.
##

### LINKED ISSUE #859: Consider replacing urlparse with something that can handle the WHATWG's 'generous' definition of a URL.
For basically its entire lifetime, urllib3 has used urlparse and friends from the standard library to handle the task of working with URLs.

However, it has recently been pointed out that the WHATWG maintains a URL specification that its proponents claim is merely a codification of common browser practice. That URL specification contains a parsing section that is _dramatically_ more generous than urlparse and friends are: for example, it allows the scheme to be separated from the authority portion of the URL by a colon followed by _at least_ two slashes (allowing for the possibility of having an unbounded amount of slashes: @bagder's testing has found that Firefox will handle a string like `'http:' + ('/' * 8000) + 'example.com'` without complaint).

Given this remarkably expansive definition of the allowed format of a URL, and given that this apparently _really does happen_ on the web, and given also that the position of the WHATWG is that this generosity of parsing is a net good, we should at least consider adopting that parser.

It may be sensible to for us to write a standalone library (as part of the hyper project, perhaps) that is then vendored into urllib3/requests as appropriate. Regardless, we should discuss whether we want to do this or whether we think we should be less lenient.
##

### LINKED ISSUE #952: urllib3 does not urlencode redirect targets on redirect.

### LINKED ISSUE #1096: Improve support for path-only URLs.
Encountered while doing a substantial backend rewrite for v2.

Currently, `parse_url` doesn't do well with non-urlencoded URLs that are intended to have no host portion, as frequently used by HTTP Location headers:

```python
>>> from urllib3.util.url import parse_url
>>> parse_url('/redirect?target=http://localhost:61020/')
Url(scheme='/redirect?target=http', auth=None, host='localhost', port=61020, path='/', query=None, fragment=None)
```

As you can see here, the `://` in the URL confuses urllib3's URL handling logic, which means it ends up incorrectly assuming that the scheme is `/redirect?target=http`. This is causing problems in the v2 backend, which uses `parse_url` to attempt to isolate out the request URL to put into the HTTP/1.1 request.

Frustratingly, urlparse gets this right:

```python
>>> from urllib.parse import urlparse
>>> urlparse('/redirect?target=http://localhost:61020/')
ParseResult(scheme='', netloc='', path='/redirect', params='', query='target=http://localhost:61020/', fragment='')
```
It would be good if `parse_url` could spot a URL that is path-only and handle it sensibly.

This is part of a larger problem, which is that `parse_url` and the `Url` class cannot be used to incrementally build up a URL target. This is required to emulate browser logic for redirects if that's a thing that we want to do: it needs to be possible to handle all URLs that could be returned in Location headers, including absolute, scheme-relative, absolute-path, and path-relative, but right now urllib3 can't really do that.

The larger issue doesn't block v2, but this does. It's probably something that @sigmavirus24 could give guidance on if someone wants to pick it up.

SethMichaelLarson:
This PR adds the `rfc3986` module into `urllib3.packages` and replaces our current URL parser with the one from `rfc3986`. This brings our URL parser in-line with Python standard library parsing which all use RFC 3986.

As far as I can see the only backwards compatibility we lose here is we can no longer parse `http://@`.  I'm hoping breaking this use case is acceptable. :) I've added a few URLs from various sources that other URL parsers had trouble with.

Hoping I can get a review on your thoughts @sigmavirus24?

Reference for one of the example URLs (maybe more here?): https://www.blackhat.com/docs/us-17/thursday/us-17-Tsai-A-New-Era-Of-SSRF-Exploiting-URL-Parser-In-Trending-Programming-Languages.pdf

Closes #466
Closes #859
Closes #952
Closes #1096


sigmavirus24:
I think there are some open bugs that should be fixed before we move over to this wholesale.

SethMichaelLarson:
I can work on some of those bugs now, is there a list of bugs you'd like squashed before you're comfortable with this merge?

sigmavirus24:
https://github.com/python-hyper/rfc3986/issues/32 is the main one I think

sigmavirus24:
This is probably a nice to have as well: https://github.com/python-hyper/rfc3986/issues/36

SethMichaelLarson:
I'll take a look at these issues, thanks!

sigmavirus24:
And if we ever use parsed URIs as dictionary keys, we'll likely need: https://github.com/python-hyper/rfc3986/pull/35 (based on the description)

SethMichaelLarson:
Thanks for reviewing that change @sigmavirus24.

Pending python-hyper/rfc3986#37 being merged if we assume that python-hyper/rfc3986#32 is a non-issue until @kennethreitz gets back to us is there anything else that needs to be done for a new release of rfc3986 being tagged?

sigmavirus24:
I think we need release notes and a version bump. Once merged I can run the automation for it.

sigmavirus24:
Once updated to 1.2.0 I'm :+1: on this

src/urllib3/util/ssl_.py
```diff
@@ -44,12 +44,11 @@ def _const_compare_digest_backport(a, b):
...
 # and IPv6 addresses for use in is_ipaddress()
 _IP_ADDRESS_REGEX = re.compile(
-    r'^(?:%s|%s|%s|%s|%s)$' % (
+    r'^(?:%s|%s|%s|%s)$' % (
         abnf_regexp.IPv4_RE,
         abnf_regexp.IPv6_RE,
         abnf_regexp.IPv6_ADDRZ_RE,
-        abnf_regexp.IPv_FUTURE_RE,
-        abnf_regexp.IP_LITERAL_RE
+        abnf_regexp.IPv_FUTURE_RE
```
sigmavirus24: I wonder if rfc3986 could just provide a regular expression for use here. :woman_shrugging:

SethMichaelLarson: Probably, something like `misc.IP_ADDRESS_MATCHER`?

sigmavirus24: I don't know that rfc3986 will ever need to use the matcher so I'd prefer not to add more time compiling a regexp we're not going to use. We could just have them all in one RegExp though in abnf_regexp

SethMichaelLarson: That's true, I can create that issue to track the idea.

SethMichaelLarson:
Current issue is specifically Windows Python 2.7 having issues with `socks5://` URLs sending the incorrect IP address. Looks like somehow getting `0.0.0.1` after deserializing `socket.inet_ntoa()` which implies it's reading `b'\x00\x00\x00\x01'` from the socket? Need to figure out why that's occurring and why only on Windows+Python 2.7, all other tests are running fine.

SethMichaelLarson:
After reading the `PySocks` source it looks like there's a usage of `socket.inet_pton` which is not available for Windows on Python 2.7. Maybe we need to add a conditional requirement of `win_inet_pton` when Python < 3.4 for the `socks` extra?
Also according to our documentation IPv6 isn't supported by PySocks in some situations but I might want to revisit that.

SethMichaelLarson:
Looks like the unicode addresses with `win_inet_pton` was the problem. Going to open an issue on that repo as `socket.inet_pton` supports unicode addresses on non-Windows Python 2.7. I'll clean up all the little commits I've made from the PR and add some test cases for ensuring that each piece of the URL is not unicode.

src/urllib3/contrib/socks.py
```diff
@@ -75,6 +75,7 @@ def _new_conn(self):
         if self.socket_options:
             extra_kw['socket_options'] = self.socket_options
 
+        print(self.host, self.port, self._socks_options)
```
sigmavirus24: Left over print debugging?

test/contrib/test_socks.py
```diff
@@ -80,7 +80,8 @@ def _address_from_socket(sock):
         return socket.inet_ntop(socket.AF_INET6, ipv6_addr)
     elif addr_type == b'\x03':
         addr_len = ord(sock.recv(1))
-        return _read_exactly(sock, addr_len)
+        data = _read_exactly(sock, addr_len)
```
sigmavirus24: should we revert this unrelated change?

SethMichaelLarson: Yeah I'm going to revert all of these commits completely when I clean up this branch :)

SethMichaelLarson:
Ready for a rereview @sigmavirus24, thanks for catching those artifacts.

SethMichaelLarson:
Woo! :tada:

shazow:
Yay this is great!! :)

theacodes:
> Yay this is great!! :)

For real. Thanks for doing this, @SethMichaelLarson and thank you for reviewing, @sigmavirus24!

sigmavirus24:
I did almost no hard work whatsoever. :clap: @SethMichaelLarson

trustedsec:
There appears to be normalization of URL's occurring that removes intended functionality through GET/POST requests. 

Built a scanner that looks for a directory traversal issue to help to troubleshoot missing patches and it strips out the ../ in the request.

Even with allow_redirects=False the request still gets stripped.

Example:

$ python3
Python 3.7.4 (default, Dec 13 2019, 01:02:18)
[GCC 7.3.1 20180712 (Red Hat 7.3.1-6)] on linux
>>> import requests
>>> req = requests.get('https://x.x.x.x/vpn/../vpns/etc/smb.conf', allow_redirects=False)
>>> req.content
b'<html><head><META HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8"><script type="text/javascript" src="/vpn/resources.js"></script><script type="text/javascript" language="javascript">var Resources = new ResourceManager("/logon/themes/Default/resources/{lang}", "REDIRECTION_BODY");</script></head><body><span id="This object may be found "></span><a href="/vpn/tmindex.html"><span id="here"></span></a><span id="Trailing phrase after here"></span><script type="text/javascript" language="javascript">Resources.Load();</script></body></html>'
>>> req.status_code
302
>>> req.url
'https://x.x.x.x/vpns/etc/smb.conf'
>>> 

Notice the request for ../ is stripped/removed.

jbwdevries:
FYI: This broke an integration with PLC software for us, because it changes the URLs send out (via `requests`).

This is the changed behaviour:
```sh
python3 -m venv venv

venv/bin/pip install urllib3==1.23
venv/bin/python -c 'from urllib3.util import parse_url ; print(parse_url("http://foo.tld/?a=[]"))'
# http://foo.tld/?a=[]

venv/bin/pip install urllib3==1.26.11
venv/bin/python -c 'from urllib3.util import parse_url ; print(parse_url("http://foo.tld/?a=[]"))'
# http://foo.tld/?a=%5B%5D
```

Currently trying to find a workaround.

PR #1489, opened 2018-12-01
Add support for password-protected client keyfiles

### LINKED ISSUE #1275: Make sure PyOpenSSLContext.load_cert_chain sets password as byte string
I ran into an issue creating a TransportAdapter using an SSL Context with a passpharse due to the fact that pyopenssl expects a byte string and not unicode (see https://github.com/pyca/pyopenssl/issues/701).

Apparently, they have a reason for not casting it, so it would be nice if `urllib3` did the type check and casting instead. I'd be happy to submit a PR but just wanted to run the idea by you guys first. This could be something as simple as:

```
    def load_cert_chain(self, certfile, keyfile=None, password=None):
        self._ctx.use_certificate_file(certfile)
        if password is not None:
            self._ctx.set_passwd_cb(lambda max_length, prompt_twice, userdata: six.binary_type(password))
        self._ctx.use_privatekey_file(keyfile or certfile)
```

Or it could have a bit more robust type checking.

sethmlarson:
The standard library supports passing text or binary passwords to `SSLContext` so we should also support that for `PyOpenSSLContext`/`SecureTransportContext` objects. This PR adds support for that and tests for all situations where this functionality is supported.

Currently the only way we support this is via creating your own `SSLContext` and calling `load_cert_chain()` yourself. Adding the `key_password` parameter to `HTTPSConnectionPool` allows this to be defined from the `PoolManager` level without having to create your own `SSLContext`.

Another problem that I encountered first-hand is that if OpenSSL encounters an encrypted keyfile but the `password` parameter is not given:

> [If the password argument is not specified and a password is required, OpenSSL’s built-in password prompting mechanism will be used to interactively prompt the user for a password.](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain)

which basically means if you try using an encrypted keyfile without a password your program just straight up blocks which breaks a lot of things in non-interactive contexts. (PyCharm locks up irreversibly, pytest appears to be blocking indefinitely due to swallowing stdout).  I was wondering if there's a way for us to stop this bad behavior on our side and raise an exception instead? Besides detecting an encrypted keyfile within `ssl_wrap_socket` I don't know what else we can do here. :(

Closes #1275.


SethMichaelLarson:
The commit I just pushed implements simple logic to detect an encrypted key file and raise a helpful error message before OpenSSL has a chance to block indefinitely. If there's other ideas or we decide against landing that commit I can roll it back.

sigmavirus24:
cc @alex

alex:
Quick skim looks fine to me.

sethmlarson:
Thanks all!

src/urllib3/util/ssl_.py
```diff
@@ -356,5 +367,15 @@ def is_ipaddress(hostname):
     if six.PY3 and isinstance(hostname, bytes):
         # IDN A-label bytes are ASCII compatible.
         hostname = hostname.decode('ascii')
-
     return _IP_ADDRESS_REGEX.match(hostname) is not None
+
+
+def _is_key_file_encrypted(key_file):
+    """Detects if a key file is encrypted or not."""
+    with open(key_file, 'r') as f:
```
jco-odoo, line 375: @sethmlarson There would not be a way to put this in a try...except or something?  We want to avoid files or temporary files and we put a special HTTPAdapter to pass the contents immediately instead of the files, but then it tracebacks at this point.  Ideally, there would be support for passing the data immediately instead of opening files.

sethmlarson: @jco-odoo Can you open a separate issue for this?

jco-odoo: https://github.com/urllib3/urllib3/issues/2508

PR #1495, opened 2018-12-07
Upgrade coverage to 4.5.2

sethmlarson:
Turns out per nedbat/coveragepy#714 that phantom lines are generated when running on Python 3.8, that's probably the root of our issues? If you check the report on codecov for this PR it's at 100% now. :)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -12,9 +12,10 @@ omit =
 
 [paths]
 source =
-   src/urllib3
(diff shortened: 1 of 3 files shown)


pquentin:
Ah, yes, of course! I've seen this in https://github.com/python-trio/trio/pull/784 too but for some reason did not make the connection here. Nice catch.

PR #1496, opened 2018-12-07
Add support for TLS 1.3 to all HTTPSConnection implementations

### LINKED ISSUE #1269: Connection fails via https when certificate contains IPv6 address in SAN
The URL is `https://[2001:db8::17]/`, the server is running with a certificate that includes both `DNS:2001:db8::17` and `IP:2001:db8::17` as `alt_names`. Accessing that URL with e.g. `curl` works just fine, when using urllib3 I am seeing this error instead:

```
A problem was encountered with the certificate that prevented urllib3 from finding the SubjectAlternativeName field. This can affect certificate validation. The error was Codepoint U+003A at position 5 of u'2001:db8::17' not allowed
```

### LINKED ISSUE #1368: Missing License text on securetransport.py
One thing that should happen is adding the license text from https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/_securetransport/bindings.py#L12-L30 to https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py. `securetransport.py` seems to have chunks taken verbatim from https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py, along with sections inspired by. No doubt there more differences than similarities, but the exact same variable, function and parameters names, combined with identical control flow in areas would imply to me that at least portions of the code started their life from the oscrypto project.

Some examples:

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L90
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L83

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L94
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L181

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L144-L149
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L218-L223

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L209
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L238

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L246-L251
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L269-L274

 - https://github.com/wbond/oscrypto/blob/master/oscrypto/_osx/tls.py#L264-L265
 - https://github.com/urllib3/urllib3/blob/master/urllib3/contrib/securetransport.py#L291-L292

 -

### LINKED ISSUE #1386: Remove TLS13-* suites from DEFAULT_CIPHERS
OpenSSL 1.1.1 changed the name of the TLS 1.3 cipher suites and added a new API to configure them. ``SSL_CTX_set_cipher_list()`` no longer affects TLS 1.3 suites. Therefore the ``TLS13`` suites can be removed from ``DEFAULT_CIPHERS``. The suites are now configured with ``SSL_CTX_set_ciphersuites()``. Neither CPython nor PyOpenSSL expose the feature yet.

https://github.com/openssl/openssl/pull/5392

### LINKED ISSUE #1451: Remove/fix all deprecated features in the ssl module
Python 3.6.x/3.7.x deprecated a lot of features in the ssl module. We should upgrade to using the latest ones that aren't deprecated and provide a proper fallback.

### LINKED ISSUE #1500: Add support for TLS 1.3  for pyOpenSSL when cryptography 2.5 is released
Leaving this for myself in the future when cryptography 2.5 is released as a part of the work in #1496.

### LINKED ISSUE #1537: OpenSSL.SSL.Error raised with invalid client cert and cryptography 2.5
In requests/requests#4961, I stumbled onto an issue where it appears that in the right scenario, OpenSSL errors are raised from `urllib.connectionpool.HTTPConnectionPool._make_request`. It happens with these deps

```
asn1crypto==0.24.0
certifi==2018.11.29
cryptography==2.5
ddt==1.2.0
idna==2.8
pyOpenSSL==19.0.0
requests==2.21.0
requests-unixsocket==0.1.5
trustme==0.5.0
urllib3==1.24.1
```

[among others](https://github.com/cherrypy/cheroot/issues/173#issuecomment-460411259) on multiple Pythons and operating systems including Python 3.7.2 on macOS or Python 3.7 on Linux.

Downgrading to pyOpenSSL 18 has no effect, but downgrading to cryptography 2.4 resolves the issue.

I'm still unsure how cryptography 2.5 is relevant to the issue and I've been yet unable to replicate the issue in an isolated environment. One way you can currently replicate the issue is by checking out cherrypy/cheroot@e616c2542d and run `tox -- -k test_tls_client_auth --pdb` (to get a pdb prompt when the exception occurs).

Unfortunately, because an exception was already raised, the `cl_pem` and other state are already cleared out. Let me see if I can't distill that test into a more isolated test that replicates the issue.

sethmlarson:
TLS 1.3 support came in OpenSSL 1.1.1 and Python 3.7 and we should start testing against it.
TLS 1.3 will be available under the following conditions:
- You're using Python 3.7+ compiled against OpenSSL 1.1.1+ OR
- You're using SecureTransport on macOS 10.13+ with TLS 1.3 enabled OR
- You're using pyOpenSSL with cryptography 2.5+

Most of our users will be using TLS 1.3 through pyOpenSSL.

If TLS 1.3 is available and you don't hard-code a protocol version it'll be used by default.

Closes #1368
Closes #1386 
Closes #1269 
Closes #1451 
Closes #1500 
Closes #1537


src/urllib3/contrib/securetransport.py
```diff
@@ -124,7 +124,7 @@
 # Basically this is simple: for PROTOCOL_SSLv23 we turn it into a low of
 # TLSv1 and a high of TLSv1.2. For everything else, we pin to that version.
```
pquentin: I think the high is now TLSv1.3

SethMichaelLarson: I'll update this. Also is `kTLSProtocol13` not available? I'm not sure when it was added to SecureTransport but I'm getting Illegal Parameter errors for using it. :(

pquentin: https://developer.apple.com/documentation/security/sslprotocol/ktlsprotocol13?language=objc mentions macOS 10.13+, which is what Travis uses. That's all I know, sorry!

SethMichaelLarson:
TODO: I need to have a better marker for whether TLS versions are available in our unit tests.

SethMichaelLarson:
I rolled the IPv6 feature into this PR to avoid a monster of a merge conflict if that's fine with reviewers. If we want to split that into it's own PR I can do so.

sethmlarson:
The ordering is now as follows:
- ECDHE / DHE (Perfect Forward Security over Static Key Parameters)
- AES-GCM (AES commonly implemented in hardware)
- ChaCha20 (Alternative to AES-GCM, good where AES hardware acceleration isn't available)
- ECDHE ECDSA (ECDSA is uncommon but if available is preferable to ECDHE RSA)
- ECDHE RSA
- DHE RSA (More computationally expensive than ECDHE)
- AES-256 (AES-256 and AES-128 similar security-wise)
- AES-128
- SHA-384 (SHA ordering is already done by OpenSSL)
- SHA-256
- SHA-1 / SHA
- No DSS because it's uncommon and insecure
- No MD5 because it's insecure
- No unencrypted or unauthenticated ciphers for obvious reasons

src/urllib3/util/ssl_.py
```diff
@@ -83,30 +93,34 @@ def inet_pton(_, host):
...
 # - prefer ECDHE over DHE for better performance,
 # - prefer any AES-GCM and ChaCha20 over any AES-CBC for better performance and
 #   security,
 # - prefer AES-GCM over ChaCha20 because hardware-accelerated AES is common,
-# - disable NULL authentication, MD5 MACs and DSS for security reasons.
+# - disable NULL authentication, MD5 MACs, DSS, and other
+#   insecure ciphers for security reasons.
+# - NOTE: TLS 1.3 cipher suites are managed through a different interface
+#   not exposed by CPython (yet!) and are enabled by default if they're available.
 DEFAULT_CIPHERS = ':'.join([
```
shazow: I wish there was a more... formal way of composing this list. It doesn't exactly match up with hynek's post (per the notes), and it's somewhat different from what other platforms do (quickly looked at golang/chrome/firefox) but the spirit seems similar enough.

Somewhat of a frustrating situation.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -626,7 +626,7 @@ def socket_handler(listener):
             # First request should fail.
             response = pool.urlopen('GET', '/', retries=0,
                                     preload_content=False,
-                                    timeout=Timeout(connect=1, read=0.001))
+                                    timeout=Timeout(connect=1, read=0.1))
```
shazow: :cry:

src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -562,26 +564,22 @@ class SecurityConst(object):
     TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 = 0xC030
     TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256 = 0xC02B
     TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 = 0xC02F
-    TLS_DHE_DSS_WITH_AES_256_GCM_SHA384 = 0x00A3
+    TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 = 0xCCA9
```
shazow: How are these updated? (How does someone validate that these constants were pulled in correctly?)

sethmlarson: I saw this comment and I'll address the rest of your comments later but since this chunk is a lot of work to review: https://developer.apple.com/documentation/security/1550981-ssl_cipher_suite_values?language=objc

shazow: Confirmed. A comment on the source of the constants would be handy. :)

dummyserver/testcase.py
```diff
@@ -115,13 +115,16 @@ class HTTPDummyServerTestCase(unittest.TestCase):
     scheme = 'http'
     host = 'localhost'
     host_alt = '127.0.0.1'  # Some tests need two hosts
-    certs = DEFAULT_CERTS
+
+    @classmethod
```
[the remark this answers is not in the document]
sethmlarson: Could be mistaken because I've never tried it but properties aren't available to the class only the instance?  We need it for `_start_server()` which is also a `classmethod`.

src/urllib3/util/ssl_.py
sethmlarson: I agree it's a confusing configuration method. 

What parts of Hynek's post don't match? I ask because I tried to base this list off of that post while adding more explicit ordering that he offered to favor ECDSA over RSA. He mentioned it in a comment if I recall that it could be done it'd just double the size of the list.

From what I've seen of Firefox, Golang, and Chrome:
- Golang has ChaCha20 above AES-GCM (Could be wrong here but it's what it looks like to me?) which I don't think is great because of how common AES hardware acceleration is available in CPUs.
- Firefox (at least on my machine) offers AES-CBC in front of ChaCha20 where AES-CBC isn't an AEAD cipher. Also offers 3DES which is much closer to being insecure than AES or ChaCha.
- Chrome is best of the three, offers 3DES for compatibility reasons I'm sure.

The only thing we're losing from these changes is dropping static curve/DH parameters and DSS which I think are both safe to drop, DSS especially because it seems like we intended to not have it available (per the comment in our cipher section) but it was not disabled and so therefore was available. (Whoopsie!)

src/urllib3/contrib/_securetransport/bindings.py
sethmlarson: Agreed, I'll add this!

dummyserver/testcase.py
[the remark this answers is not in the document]
shazow: Just curious why the change was necessary, I assumed the old `certs` would be available to the class too.

src/urllib3/util/ssl_.py
shazow: I might be reading the blog post wrong, but here's the comparison:

```
| hynek        | ours                  |
+--------------+-----------------------+
 ECDH+AESGCM     ECDHE+ECDSA+AESGCM
 ECDH+CHACHA20   ECDHE+ECDSA+CHACHA20
 DH+AESGCM       ECDHE+AESGCM
 ECDH+AES256     ECDHE+CHACHA20
 DH+AES256       DHE+AESGCM
 ECDH+AES128     DHE+CHACHA20
 DH+AES          ECDHE+ECDSA+AES256
 RSA+AESGCM      ECDHE+ECDSA+AES128
 RSA+AES         ECDHE+AES256
 !aNULL          ECDHE+AES128
 !MD5            DHE+AES256
 !DSS            DHE+AES
                 RSA+AESGCM
                 RSA+AES
                 !aNULL
                 !eNULL
                 !MD5
                 !DSS
```

I mean it looks reasonable based on what I understand--I don't see anything nefarious. :) I just wish we could use a list verbatim from somewhere for the sake of not having to justify decisions.

sethmlarson: I changed the list to:
```python
DEFAULT_CIPHERS = ':'.join([
    'ECDH+AESGCM',
    'ECDH+CHACHA20',
    'DH+AESGCM',
    'DH+CHACHA20',
    'ECDH+AES',
    'DH+AES',
    'RSA+AESGCM',
    'RSA+AES',
    '!aNULL',
    '!eNULL',
    '!MD5',
    '!DSS',
    '!aDSS',
])
```

and I'm still getting `TLS_DH_DSS_WITH_AES_128_CBC_SHA256` in the list of offered ciphers despite `!DSS` and `!aDSS` in there and I don't understand that in the slightest.

sethmlarson: Okay I've changed the list to be almost identical to Hynek's list. The only changes I made are:

- Force ECDHE and DHE in front of  ECDH and DH so we get perfect forward secrecy on servers that honor cipher ordering. (OpenSSL doesn't guarantee ordering of ephemeral key exchanges before static)
- Not delineate between AES256 and AES128 because security-wise they're the same and OpenSSL already does this for us.

| Hynek         | Ours           |
|---------------|----------------|
| ECDH+AESGCM   | ECDHE+AESGCM   |
| ECDH+CHACHA20 | ECDHE+CHACHA20 |
| DH+AESGCM     | DHE+AESGCM     |
| ECDH+AES256   | DHE+CHACHA20   |
| DH+AES256     | ECDH+AESGCM    |
| ECDH+AES128   | DH+AESGCM      |
| DH+AES        | ECDH+AES       |
| RSA+AESGCM    | DH+AES         |
| RSA+AES       | RSA+AESGCM     |
| !aNULL        | RSA+AES        |
| !MD5          | !aNULL         |
| !DSS          | !eNULL         |
|               | !MD5           |
|               | !DSS           |

Want to get your approval on this list @shazow so I can reimplement the cipher list for SecureTransport.

dummyserver/testcase.py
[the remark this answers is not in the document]
sethmlarson: I used a function here because we're making a copy of `DEFAULT_CERTS` and changing the `ssl_version` key per-test for the TLS version tests.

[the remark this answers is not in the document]
shazow: So you're saying before we'd have to do something like...

```python
tlsv1_certs = DEFAULT_CERTS.copy()
tlsv1_certs['ssl_version'] = ssl.PROTOCOL_TLSv1

class TestHTTPS_TLSv1(TestHTTPS_TLSVersion):
    tls_protocol_name = 'TLSv1'
    certs = tlsv1_certs
```

Instead of...

```python
class TestHTTPS_TLSv1(TestHTTPS_TLSVersion):
    tls_protocol_name = 'TLSv1'

     @classmethod
    def certs(cls):
        certs = DEFAULT_CERTS.copy()
        certs['ssl_version'] = ssl.PROTOCOL_TLSv1
        return certs
```

?

Not obvious to me what the advantage of the latter one is, but I don't have strong feelings here. :)

src/urllib3/util/ssl_.py
shazow: @sethmlarson I don't have the subject matter expertise to make an informed judgement on that. :)

As before, I don't see anything nefarious. Just asking questions so we can better document our thought process for context, in case anyone else wants to audit this later.

sethmlarson:
After seeing the [latest build on master fail](https://travis-ci.org/urllib3/urllib3/builds/496129211) for the same reason as this PR makes me feel better about this issue. Will have to investigate on master, might be due to the new cryptography version?

sethmlarson:
cc @theacodes @shazow I think this PR is ready for another round of review. The failures on the requests downstream tests are failing both on our master and on requests master.

sethmlarson:
Unfortunately I can't squash+merge myself due to CI failures. macOS Python is now failing to build cryptography wheels which I'll investigate outside of this PR. I'm ready for this to be merged anytime.

Thanks reviewers! :)

theacodes:
Just to confirm, do you want me to use black magic to force merge?

sethmlarson:
@theacodes Yes please! :sparkles:

webknjaz:
Thanks for the fix!

@theacodes @shazow @sethmlarson It looks like the last release was ~5 months ago. Is there any chance to get this released anytime soon?

sethmlarson:
@webknjaz The next release is pending #1531 and python-hyper/rfc3986#50.

webknjaz:
@sethmlarson thanks for the hint!

Issue #1501, opened 2018-12-19
App Engine Warning: TypeError: expected httplib.Message, got <type 'instance'>


speedplane:
While running on App Engine, I received the following error message in the logs when making HTTP requests: `TypeError: expected httplib.Message, got <type 'instance'>`. I believe this is because App Engine uses a different HTTP lib with it's own specially named instances. These instances *should* be compatible with `httplib.Message`, so the rest of the function can proceed, but the initial check will not match.

I'll put together a PR that is able to skip the `isinstance` check in `assert_header_parsing` when we're on app engine.

The complete call stack of this error is below:
```
File "project\libs\elasticsearch_six\connection\http_urllib3.py", line 172, in perform_request
    response = self.pool.urlopen(method, url, body, retries=Retry(False), headers=request_headers, **kw)
  File "project\libs\urllib3\connectionpool.py", line 601, in urlopen
    chunked=chunked)
  File "project\libs\urllib3\connectionpool.py", line 396, in _make_request
    assert_header_parsing(httplib_response.msg)
  File "project\libs\urllib3\util\response.py", line 56, in assert_header_parsing
    type(headers)))
TypeError: expected httplib.Message, got <type 'instance'>.
```

However, the error does not cause an exception, it is caught and is printed as a wanring in `_make_request`.

speedplane:
Provided a quick fix in PR #1502.

sethmlarson:
We haven't had this reported to us again in some time, going to close this as potentially a transient issue with App Engine? Without more information it's tough to know the root cause. (Plus App Engine isn't getting special attention in v2.x+)

PR #1502, opened 2018-12-19
App Engine: Don't raise an exception if the header instance type is incorrect

speedplane:
App Engine uses its own httplib, which are (generally) compatible with standard httplib, but with different names. Addresses #1501.

### DIFF
--- a/src/urllib3/util/response.py
+++ b/src/urllib3/util/response.py
@@ -4,6 +4,7 @@ from ..packages.six.moves import http_client as httplib
 from ..exceptions import HeaderParsingError
 from ..contrib import _appengine_environ
 
+
 def is_fp_closed(obj):
     """
     Checks whether a given file-like object is closed.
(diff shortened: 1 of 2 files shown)


src/urllib3/util/response.py
```diff
@@ -51,7 +52,8 @@ def assert_header_parsing(headers):
 
     # This will fail silently if we pass in the wrong kind of parameter.
     # To make debugging easier add an explicit check.
-    if not isinstance(headers, httplib.HTTPMessage):
+    if not _appengine_environ.is_appengine() and \
```
SethMichaelLarson: Can we improve our test suite somehow to catch issues like this?

theacodes:
@speedplane this PR still has outstanding comments and a merge conflict. If you'd still like it merged, please let us know! I'm going to close it for now, though.

Issue #1503, opened 2018-12-19
App Engine Error with HTTPS


speedplane:
I believe the change in PR #1430 broke urllib3 on Google App Engine, which doesn't have `_tunnel_host` in its implementation of httplib. Not sure why this wasn't picked up by tests, but the fix is easy, simply adding `if getattr(self, '_tunnel_host', None)` part should go back in.

Below is a stacktrace of the error. Note that this is pure urllib3, no requests are being used.

```
Traceback (most recent call last):
  ...
  File "/project/updater/search_backend.py", line 1353, in setup
    "/_alias/%s"%self.index_name)
  File "/project/libs/elasticsearch_five/transport.py", line 312, in perform_request
    status, headers, data = connection.perform_request(method, url, params, body, ignore=ignore, timeout=timeout)
  File "/project/libs/elasticsearch_five/connection/http_urllib3.py", line 114, in perform_request
    response = self.pool.urlopen(method, url, body, retries=False, headers=self.headers, **kw)
  File "/project/libs/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/project/libs/urllib3/connectionpool.py", line 343, in _make_request
    self._validate_conn(conn)
  File "/project/libs/urllib3/connectionpool.py", line 839, in _validate_conn
    conn.connect()
  File "/project/libs/urllib3/connection.py", line 304, in connect
    if self._tunnel_host:
AttributeError: 'VerifiedHTTPSConnection' object has no attribute '_tunnel_host'
```

speedplane:
See PR #1504 for the related PR.

PR #1430, opened 2018-08-24
Remove more Python 2.6 code

pquentin:
This also fixes an installation error of gcp-devrel-py-tools==0.0.15, see https://github.com/GoogleCloudPlatform/python-repo-tools/issues/23 for details. I can separate that in two PRs if needed.

PR #1504, opened 2018-12-19
Fix App Engine Tunnel Host

speedplane:
App Engine does not have _tunnel_host set in its httplib implementation. Make sure we check for that. #1503. Note this reverts part of the change in #1430.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -171,7 +171,7 @@ class HTTPConnection(_HTTPConnection, object):
 
     def _prepare_conn(self, conn):
         self.sock = conn
-        if self._tunnel_host:
+        if getattr(self, '_tunnel_host', None):
             # TODO: Fix tunnel so it doesn't depend on self.sock state.
             self._tunnel()
             # Mark this connection as not reusable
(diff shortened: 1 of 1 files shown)


src/urllib3/connection.py
```diff
@@ -301,7 +301,8 @@ def connect(self):
         conn = self._new_conn()
         hostname = self.host
 
-        if self._tunnel_host:
+        # Google App Engine's different HTTP lib may not define_tunnel_host.
```
pquentin: I think you need a space after "define": "define _tunnel_host".

src/urllib3/connection.py
```diff
@@ -171,7 +171,7 @@ def _new_conn(self):
 
     def _prepare_conn(self, conn):
         self.sock = conn
-        if self._tunnel_host:
+        if getattr(self, '_tunnel_host', None):
```
SethMichaelLarson: This looks fine to me, my only question is how do our tests not catch this?

theacodes:
@speedplane this PR still has outstanding comments. If you'd still like it merged, please let us know! I'm going to close it for now, though.

PR #1505, opened 2018-12-20
Add Tidelift and remove BountySource

sethmlarson:
- BountySource hasn't had much activity so we're removing it (I'll remove labels from the issues as well).
- Combined the section on Grants and Sponsorships.
- Added GOVCERT.LU to the list of financial contributors! :tada:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,6 +77,13 @@ urllib3 happily accepts contributions. Please see our
 for some tips on getting started.
 
 
+Security Disclosures
+--------------------
+
+To report a security vulnerability, please use the
+`Tidelift security contact <https://tidelift.com/security>`_.
+Tidelift will coordinate the fix and disclosure with maintainers.
+
 Maintainers
 -----------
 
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -97,8 +104,24 @@ Maintainers
 Sponsorship
 -----------
 
+.. |tideliftlogo| image:: https://nedbatchelder.com/pix/Tidelift_Logos_RGB_Tidelift_Shorthand_On-White_small.png
+   :width: 75
+   :alt: Tidelift
+
```
nedbat: I noticed your home page had a Tidelift display similar to mine, and wondered how you did it, since it took me a few tries to get it right.  I guess you used my markup! :) 
Now I know to keep that image file in-place for the future.
BTW: on GitHub, the logo itself just links to the logo file.  I added an href to the image with this: https://github.com/nedbat/coveragepy/commit/7fb7e9dac6f4fcb470368336ce99eb16b71e42c0

sethmlarson: The easiest solution! :) I think we're planning on moving some of our Tidelift banners around a bit so we'll see what we end up with. Hopefully Tidelift can come up with some of their own banner implementations.

PR #1507, opened 2018-12-21
Use cert_reqs=CERT_REQUIRED by default

### LINKED ISSUE #1454: Shouldn't urllib3 verify SSL/TLS certificates by default ?
The README says:

> ... urllib3 brings many critical features that are missing from the Python standard libraries: 
> ...
> Client-side SSL/TLS verification.
> ...

and only in https://urllib3.readthedocs.io/en/latest/user-guide.html#certificate-verification it's explained that 

> By default, urllib3 does not verify HTTPS requests.

Because since [pep-0476](https://www.python.org/dev/peps/pep-0476) (ie python >= 2.7.9 or > 3.4 ) standard libraries verifies SSL certificates by default, nowadays projects switching from urllib2 or httplib to urllib3 without explicitly enabling certificate validation would loose the feature of certificate validation that was provided by standard library.

Of course, this would be an incompatible breaking change, like it was for python at the time, but is it really still a sane default to not verify certificates nowadays ?

sethmlarson:
Initial implementation of requiring and validating certificates when using HTTPS by default.
I'm sure I'll have to rewrite some tests, will work on that after I see all the failures. :)

The `InsecureRequestWarning` will still be emitted when explicitly using `CERT_NONE` or `CERT_OPTIONAL`.

Closes #1454.


shazow:
Yay excited about this one. :)

Should we bolt on more docs to cert failure exceptions so people know what to do?

sigmavirus24:
> Should we bolt on more docs to cert failure exceptions so people know what to do?

Only if those docs are actually helpful

sethmlarson:
Definitely! I'll take a look at the docs as well tonight.

docs/user-guide.rst
```diff
@@ -202,8 +202,13 @@ recommended to set the ``Content-Type`` header::
 Certificate verification
 ------------------------
 
+ .. note:: *New in version 1.25*
+
+    HTTPS connections are now verified by default
+    (``cert_reqs = 'CERT_REQUIRED'``) regardless of whether
+    CA certs are specified.
+
 It is highly recommended to always use SSL certificate verification.
```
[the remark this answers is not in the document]
theacodes: Let's reword it, maybe "While you can disable certification verification, it is highly recommend to leave it on."

docs/user-guide.rst
```diff
@@ -202,8 +202,13 @@ recommended to set the ``Content-Type`` header::
...
-**By default, urllib3 does not verify HTTPS requests**.
 
 In order to enable verification you will need a set of root certificates. The easiest
```
theacodes: Is this still true?

sethmlarson: I don't know if this was ever completely true? You could set `cert_reqs` to `REQUIRED` and we'd load the default verification locations.

So now by default urllib3 tries to load the default certificate locations via [`SSLContext.load_default_certs()`](https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.load_default_certs) which apparently works well on Windows and on non-Windows uses [`SSLContext.set_default_verify_locations()`](https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.set_default_verify_paths) which is also reliable in my experience.

The only platforms that will have to use certifi or their own bundle are platforms that don't have `SSLContext.load_default_certs()` or don't define `SSLContext` at all because our implementation of `SSLContext` doesn't define `load_default_certs()` (Should our implementation try to implement this?).

As for `contrib/`: `PyOpenSSLContext` defines a functioning `load_default_certs()` and `SecureTransportContext` defines the method but does absolutely nothing with it so maybe it will fail cert verification unless SecureTransport loads certs by default? Will have to look at docs.

To sum it up, it's a mixed bag of whether we load system certs if no certificate locations are given. :)

docs/user-guide.rst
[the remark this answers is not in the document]
sethmlarson: Done!

docs/user-guide.rst
```diff
@@ -202,8 +202,13 @@ recommended to set the ``Content-Type`` header::
 Certificate verification
 ------------------------
 
+ .. note:: *New in version 1.25*
+
+    HTTPS connections are now verified by default
```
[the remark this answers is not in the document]
sethmlarson: Done!

sethmlarson:
Thanks for the comments, I've updated the documentation in the areas mentioned. What do we think about the wordings now?

docs/user-guide.rst
theacodes: Then we should probs remove this, yeah?

sethmlarson: When you say this you mean the whole section on certifi? I'm not sure what your comment is directed at. If we remove that subsection then the section as a whole is pretty bare and the last part mentions system certificates as well. Maybe this whole section needs a rewrite?

docs/user-guide.rst
```diff
@@ -202,11 +202,15 @@ recommended to set the ``Content-Type`` header::
...
-**By default, urllib3 does not verify HTTPS requests**.
+ .. note:: *New in version 1.25*
 
-In order to enable verification you will need a set of root certificates. The easiest
-and most reliable method is to use the `certifi <https://certifi.io/>`_ package which provides Mozilla's root certificate bundle::
+    HTTPS connections are now verified by default (``cert_reqs = 'CERT_REQUIRED'``).
+
+While you can disable certification verification, it is highly recommend to leave it on.
+
+For certificate verification to work properly you will need a set of root certificates. Unless otherwise
```
shazow: > For certificate verification to work properly you will need a set of root certificates. Unless otherwise specified urllib3 will try to load the default system certificate stores.

Aren't these two sentences contradictory? The first one sounds like you have to do something additional for urllib3 to work at all, then the second sentence sounds like it does something sensible by default and you don't need to do anything.

IMO remove the first sentence, or rephrase.

sethmlarson: Sounds good to me :)

shazow:
:custard:

sethmlarson:
We can change the documentation in a separate PR, would like to merge this change into the TLS 1.3 PR so we can finally land that monster.

PR #1509, opened 2018-12-23
Ensure types are maintained on URL components

sethmlarson:
Noticed this while testing the requests downstream that we weren't maintaining the string type in all components from the original URL. Attempting to fix that here.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -176,9 +176,9 @@ class TestUtil(object):
                                               host='localhost',
                                               path='/')),
 
-        # Unicode type
+        # Unicode type (Python 2.x)
         (u'http://foo:bar@localhost/', Url(u'http', auth=u'foo:bar', host=u'localhost', path=u'/')),
-        (b'http://foo:bar@localhost/', Url(b'http', auth=b'foo:bar', host=b'localhost', path=b'/')),
+        ('http://foo:bar@localhost/', Url('http', auth='foo:bar', host='localhost', path='/')),
     ]
 
     non_round_tripping_parse_url_host_map = [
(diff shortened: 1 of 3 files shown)


test/test_util.py
```diff
@@ -174,7 +174,17 @@ def test_parse_url_normalization(self, url, expected_normalized_url):
...
                                               host='localhost',
-                                              path='/'))
+                                              path='/')),
+
+        # Unicode type (Python 2.x)
+        (u'http://foo:bar@localhost/', Url(u'http',
+                                           auth=u'foo:bar',
+                                           host=u'localhost',
+                                           path=u'/')),
+        ('http://foo:bar@localhost/', Url('http',
```
sigmavirus24: Should these be bytes? E.g.,

```
b'http://foo:bar@localhost/'
```

sethmlarson: I need to do a full investigation on Python 2 and 3 to see what values used to be acceptable.

Currently on Python 3.x you can't use bytes gives a type error. On Python 2.x it looks like both unicode and str are acceptable. :)

sethmlarson:
But it looks like our socks tests are having trouble with the now unicode host being returned on Python 2.x :/ More investigation required here.

sethmlarson:
So what I can see in urllib3 1.24.1 is the following:

| Input Type  | Python 2.x | Python 3.x |
|---------|------------|------------|
| bytes   | str        | TypeError  |
| str     | str        | str        |
| unicode | unicode    | N/A        |

Now I need to figure out the regression on Windows win_inet_pton being affected.

sethmlarson:
So this is an issue with the win_inet_pton package which I've got a PR open for and I'll try to merge today and get a new PyPI release. Then we can potentially merge this! :)

sethmlarson:
Thanks!

PR #1511, opened 2018-12-29
Remove Authorization header regardless of case when redirecting to cross-site

### LINKED ISSUE #1510: authorization header be forwarded to cross-site when redirecting
#1346 fixes only "Authorization" header. "authorization" header isn't supported.

RFC7230 section 3.2 "Header Fields" (https://tools.ietf.org/html/rfc7230#section-3.2) says the following.

> Each header field consists of a case-insensitive field name followed
   by a colon (":"), optional leading whitespace, the field value, and
   optional trailing whitespace.

kyoshidajp:
Fixes #1510.

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -343,9 +343,10 @@ class PoolManager(RequestMethods):
         # conn.is_same_host() which may use socket.gethostbyname() in the future.
         if (retries.remove_headers_on_redirect
                 and not conn.is_same_host(redirect_location)):
-            for k in six.viewkeys(kw['headers'].copy()):
-                if k.lower() in retries.remove_headers_on_redirect:
-                    del kw['headers'][k]
+            headers = list(six.iterkeys(kw['headers']))
+            for header in headers:
+                if header.lower() in retries.remove_headers_on_redirect:
+                    kw['headers'].pop(header, None)
 
         try:
             retries = retries.increment(method, url, response=response, _pool=conn)
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -342,8 +342,12 @@ class PoolManager(RequestMethods):
         # conn.is_same_host() which may use socket.gethostbyname() in the future.
         if (retries.remove_headers_on_redirect
                 and not conn.is_same_host(redirect_location)):
-            for header in retries.remove_headers_on_redirect:
-                kw['headers'].pop(header, None)
+            headers = kw['headers']
+            newheaders = {}
+            for k, v in iter(headers.items()):
+                if not k.lower() in retries.remove_headers_on_redirect:
+                    newheaders[k] = v
+            kw['headers'] = newheaders
 
         try:
             retries = retries.increment(method, url, response=response, _pool=conn)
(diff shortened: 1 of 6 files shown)


src/urllib3/util/retry.py
```diff
@@ -179,7 +179,7 @@ def __init__(self, total=10, connect=None, read=None, redirect=None, status=None
         self.raise_on_status = raise_on_status
         self.history = history or tuple()
         self.respect_retry_after_header = respect_retry_after_header
-        self.remove_headers_on_redirect = remove_headers_on_redirect
+        self.remove_headers_on_redirect = [h.lower() for h in remove_headers_on_redirect]
```
sethmlarson: I think we should change this structure into a frozenset to protect users from adding headers not through the constructor if we're going to enforce lowercase-only.

src/urllib3/poolmanager.py
```diff
@@ -342,8 +342,12 @@ def urlopen(self, method, url, redirect=True, **kw):
         # conn.is_same_host() which may use socket.gethostbyname() in the future.
         if (retries.remove_headers_on_redirect
                 and not conn.is_same_host(redirect_location)):
-            for header in retries.remove_headers_on_redirect:
```
sethmlarson: Let's keep some of the original structure of this logic by instead looping through all the header names with `six.viewkeys()` (so we don't have to make a copy) and then seeing if that header name is within the `remove_headers_on_redirect` frozenset.

test/with_dummyserver/test_poolmanager.py
```diff
@@ -123,6 +123,17 @@ def test_redirect_cross_host_remove_headers(self):
...
+        r = http.request('GET', '%s/redirect' % self.base_url,
+                         fields={'target': '%s/headers' % self.base_url_alt},
+                         headers={'authorization': 'foo'})
+
+        self.assertEqual(r.status, 200)
+
+        data = json.loads(r.data.decode('utf-8'))
+
+        self.assertNotIn('authorization', data)
+        self.assertNotIn('Authorization', data)
```
sethmlarson: Let's also add a new test for an additional header (eg `X-API-Secret`) and seeing that that header is removed.

sethmlarson:
Also don't worry about force-pushing, we squash all commits. :)

test/with_dummyserver/test_poolmanager.py
sethmlarson: `test_redirect_cross_host_set_removed_headers`

src/urllib3/util/retry.py
kyoshidajp: You mean we should remove `remove_headers_on_redirect` parameter from the constructor?

src/urllib3/poolmanager.py
kyoshidajp: I changed by 4357014. But made a copy by `.copy()` due to getting `RuntimeError: dictionary changed size during iteration` if without `.copy()`.

src/urllib3/util/retry.py
sethmlarson: No, I mean converting the iterable we receive from remove_headers_on_redirect into a frozenset. Something like `self.remove_headers_on_redirect = frozenset([h.lower() for h in remove_headers_on_redirect])`

kyoshidajp:
Thanks a lot!

kyoshidajp:
@sethmlarson Could you review again?

sethmlarson:
Thanks for this!

PR #1512, opened 2018-12-30
Clear Cookie header when redirecting to cross-site

kyoshidajp:
Cookie header with confidential data should be cleared when redirecting to cross-site, not only Authorization header.


sethmlarson:
We don't claim to handle cookies, that's the responsibility of our users and this change would probably break those users.

PR #1521, opened 2019-01-11
Inject into urllib3.util.SSLContext

sethmlarson:
Previously we weren't injecting into `urllib3.util.SSLContext` when we inject pyOpenSSL or SecureTransport and we should do that.


sethmlarson:
Cycling because our AppVeyor tests didn't fire, probably due to GitHub outage

PR #1524, opened 2019-01-11
Skip broken pipe errors

robermorales:
In the case the server legitimally close a connection but we are still
sending data (e.g. sending a POST request), the old code prevents the
response to be retrieved.

RFC2616 refers different scenarios where a server can legitimately close the connection.

https://tools.ietf.org/html/rfc2616#section-8.1.2 (8.1.2.1)
https://tools.ietf.org/html/rfc2616#section-10.4.14

With this change, we avoid a broken pipe in the request write process,
to end the read process.

I tested curl behavior when presented to this exact scenario and it is also ignoring the broken pipes and continue parsing the legitimate response.

This behavior often get reached when using urllib3 through requests, so the related issues are from it:

Related issues:

https://github.com/requests/requests/issues/4778
https://github.com/requests/requests/issues/2422


src/urllib3/connectionpool.py
```diff
@@ -351,7 +351,15 @@ def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
         if chunked:
             conn.request_chunked(method, url, **httplib_request_kw)
         else:
-            conn.request(method, url, **httplib_request_kw)
+            try:
```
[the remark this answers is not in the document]
robermorales: Cool idea.

src/urllib3/connectionpool.py
```diff
@@ -351,7 +351,15 @@ def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
...
+                conn.request(method, url, **httplib_request_kw)
+            except BrokenPipeError:  # Python 3
+                pass
+            except IOError as e:  # Python 2
```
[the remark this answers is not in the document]
robermorales: Thanks!

src/urllib3/connectionpool.py
```diff
@@ -351,7 +351,15 @@ def _make_request(self, conn, method, url, timeout=_Default, chunked=False,
         if chunked:
             conn.request_chunked(method, url, **httplib_request_kw)
         else:
-            conn.request(method, url, **httplib_request_kw)
```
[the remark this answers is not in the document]
robermorales: Done :)

robermorales:
I made the changes you suggested. I cannot add the test right now, but will try add it and commit to the branch soon.

Thanks,

test/with_dummyserver/test_connectionpool.py
```diff
@@ -1055,5 +1055,13 @@ def test_pool_size_redirect(self):
         assert self.pool.num_connections == 1
 
 
+class TestBrokenPipeIgnore(HTTPDummyServerTestCase):
+    def setUp(self):
+        self.pool = HTTPConnectionPool(self.host, self.port)
+
+    def test_broken_pipe_ignore(self):
```
sethmlarson: Let's add this testcase under `TestHTTPConnectionPool` so it'll be re-used in other places automatically.

sethmlarson: And can we also add a testcase nearby for a chunked request as well?

robermorales: Cool ideas!

test/with_dummyserver/test_connectionpool.py
```diff
@@ -772,6 +772,13 @@ def test_mixed_case_hostname(self):
         response = pool.request('GET', "http://LoCaLhOsT:%d/" % self.port)
         self.assertEqual(response.status, 200)
 
+    def test_broken_pipe_ignore(self):
```
sethmlarson: Let's split this testcase into two testcases one with the default `self.pool.urlopen()` which defaults to unchunked and one that actually specifies `chunked=True` but for the chunked testcase we need to skip on Google App Engine because it doesn't support chunked requests in URLFetch.  You might have to create a new decorator for `notAppEngine` within `test/__init__.py` to do this.

sethmlarson: You could maybe try skipping if `urllib3.contrib.appengine.is_appengine()` is `True`?

robermorales: I tried to push the change but you already pushed it haha. Thanks!!!

robermorales:
> So I ran this code with the old urllib3 without the exception handling and it passed fine, are we sure we're triggering the pipe error with that testcase?

@sethmlarson I tried to reproduce a test case that fails with the old implementation and passes with the new one. I coudn't mimic a broken pipe in the test application, but I found a test that passes only with the change. What do you think that could be a good way to mimic the behaviour that we are covering? I am open to implement a new test that better covers it.

sethmlarson:
I'd definitely like to trigger a pipe error as least, maybe we need a lower-level interface than what Tornado is giving us to call a `socket.close()` before reading the body? Try looking in `test/with_dummyserver/test_socketlevel.py`, maybe we move this testcase to that file instead?

robermorales:
@sethmlarson cool idea! I will return with some implementation. Thanks for all your help!

sethmlarson:
Of course! :) Thank you for all your effort tackling this issue!

bbbco:
Any updates on this?

sethmlarson:
Nope, the original author appears to be MIA.

robermorales:
Sorry for the delay, I will return to the development of this soon. Thanks!

EmptyRabbit:
when could this pr merge...... I blocked by this cause...

sethmlarson:
I'll be taking a look at this again after the holidays.

sethmlarson:
@hodbn An interesting article about the error you're running into: http://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/

src/urllib3/connectionpool.py
```diff
@@ -386,10 +387,23 @@ def _make_request(
...
+        # We are swallowing BrokenPipeError (errno.EPIPE) since the server is
+        # legitimately able to close the connection after sending a valid response.
+        # With this behaviour, the received response is still readable.
+        except BrokenPipeError:  # Python 3
+            pass
+        except IOError as e:
+            # https://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/
+            if e.errno == errno.EPROTOTYPE:  # macOS
+                pass
+            elif e.errno != errno.EPIPE:  # Python 2
```
sethmlarson, line 405: We can simplify this a bit to `if e.errno not in (errno.EPROTOTYPE, errno.EPIPE): raise`

pquentin: Yeah, I did it like this to prepare for Python 2 removal, but happy to merge those as long as we still say "Python 2" and "macOS" around those lines

sethmlarson: That's fair too, it took me a moment to verify that we were doing the same thing for the `BrokenPipeError` and the `e.errno==EPIPE/EPROTOTYPE` case so wanted to prevent that from happening to others.

pquentin: Oh, I understand now. BrokenPipeError is actually not only documented as EPIPE but also ESHUTDOWN, and I failed at replacing it with errno checks, but didn't try too hard.

pquentin:
Current status: now Windows fails, I'm not sure why.

sethmlarson:
Sorry this stalled out, that's my fault. Trying to get it landed for v1.26.

Pulled this down locally and verified the current set of changes works nicely on Linux.
For Windows I added `10053` which is "Connection Aborted Error" which can happen for multiple reasons it seems (antivirus, firewall, and your peer closing the connection) so hopefully that'll fix up the issue we were seeing on Windows.

sethmlarson:
I'm not happy to skip the test on Windows but I'm unsure if there's anything we can do? Windows seems to be erroring out consistently on `sock.recv()` once the socket is closed.

hodbn (dismissed):

pquentin (dismissed):
Thanks!

We'll have to merge from master again to make sure we don't have two "notWindows" decorators

sethmlarson:
Thanks @pquentin and @hodbn for the reviews and @robermorales for putting together this change. This is a bigger fix than you think it is, many users are going to see real improvements to their network and API performance. :tada:

bessbd:
Hi @sethmlarson ,

When do you think this fix would be included in an upcoming release?
Is there something I could do to speed that process up? Should I open a PR that backports this change to any branch?

Thank you,
Donat

sethmlarson:
This change won't be backported since it's not a bug / security fix. Not much to speed things up, we're wrapping up the last things that need to be done for 1.26.

The 1.26 release is special since it's the last minor before v2 so we're trying to wrap it up a lot more "completely" compared to a typical minor release.

PR #1527, opened 2019-01-24
Handle httplib implementations that don't set _tunnel_host

jpevarnek:
urllib3 removed support for Python versions before 2.7 in a previous
pull request, unfortunately, this broke environments that use a
different implementation of httplib than the one in Python's standard
library as well. This includes Google App Engine's urlfetch-based
implementation
(https://github.com/GoogleCloudPlatform/python-compat-runtime/blob/master/appengine-compat/exported_appengine_sdk/google/appengine/dist27/gae_override/httplib.py#L362)

This pull request is nearly identical to urllib3/urllib3#1504 which was
closed due to its author not responding to comments.

Fixes urllib3/urllib3#1503

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -171,7 +171,8 @@ class HTTPConnection(_HTTPConnection, object):
 
     def _prepare_conn(self, conn):
         self.sock = conn
-        if self._tunnel_host:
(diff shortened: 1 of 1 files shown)


sethmlarson:
Thanks for this change! :tada:

PR #1531, opened 2019-01-27
Use rfc3986.validator.Validator for parse_url

### LINKED ISSUE #1529: URLs without // don't parse correctly
URLs like: `mailto:johndoe@example.com` don't parse correctly due to our logic on when to add a scheme delimeter to a URL.

sethmlarson:
Add a set of tests to make sure URLs fail when they've got invalid characters in specific components. Closes #1529


sethmlarson:
Looks like validator is upset with IPv6 addresses in our tests. I'll add some to the URL tests and investigate why we're breaking.

src/urllib3/util/url.py
```diff
@@ -175,6 +176,15 @@ def parse_url(url):
     except (ValueError, RFC3986Exception):
         raise LocationParseError(url)
 
+    # Run validator on the internal URIReference within the ParseResult
+    validator = Validator()
+    try:
+        validator.check_validity_of(
+            *validator.COMPONENT_NAMES
```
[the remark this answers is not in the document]
sigmavirus24: `validate_all_the_components_<emoji>` :trollface:

sigmavirus24:
@sethmlarson the failures look related to Zone Identifiers (RFC 6874). rfc3986 added support for those in https://github.com/python-hyper/rfc3986/issues/2 but I wonder if I overlooked support for them in the Validator work or if our IPv6 validation is a bit too strict.

Let me know if you need more information than that.

sethmlarson:
Thanks for this info, I can look more tonight. I figured out the issue was with zoning and messaged you about it on Keybase. Is that an avenue I can contact you or should I stick to email?

src/urllib3/util/url.py
```diff
@@ -14,12 +15,12 @@
 NORMALIZABLE_SCHEMES = ('http', 'https', None)
 
 # Regex for detecting URLs with schemes. RFC 3986 Section 3.1
-SCHEME_REGEX = re.compile(r"^[a-zA-Z][a-zA-Z0-9+\-.]*://")
```
sethmlarson: @sigmavirus24 What do you think of taking this `.` out of this scheme regex? I did this because I don't think we support any scheme that has this `.` here but we support a lot of URLs where the authority section looks like a scheme. (`www.google.com` is a valid "scheme").  Should we get even more strict and only support schemes that start with `http`? I'm not sure.

sigmavirus24: So, I don't understand the purpose the `.` is serving. We can't, however, limit ourselves to what we think of as normal schemes because we (fortunately, or not) support `http+unix://`

sigmavirus24:
> I figured out the issue was with zoning and messaged you about it on  Keybase. Is that an avenue I can contact you or should I stick to email?

I don't really look at or check Keybase much.

src/urllib3/util/url.py
sethmlarson: The `.` is part of the scheme spec but I couldn't find an example scheme that contained a period in it.  The three schemes I know of that we support are `http`, `https`, and `http+unix` case insensitive.

sethmlarson:
@sigmavirus24 I think I may have found the source of the issue? Our unit tests use Zone ID format from RFC 4007 which says to use a literal percent character instead of a percent-encoded percent character (`%` instead of `%25`) which causes host validation to fail.  Per RFC 3986 % isn't valid within the host component.

Should we drop support for this now that RFC 6874 calls out that RFC 4007 breaks URI syntax rules? If that's the case I can fix all the unit tests that use this syntax.

One alternative is to allow Zone IDs to be parsed with RFC 4007 with just `%` if the next two bytes aren't `25` (In either `rfc3986` or urllib3)?

sigmavirus24:
> One alternative is to allow Zone IDs to be parsed with RFC 4007 with just % if the next two bytes aren't 25 (In either rfc3986 or urllib3)?

I think going from 

`[::1%eth0]` to `[::1%25eth0]` and normalizing 4007 syntax to 6874 makes sense to me for rfc3986 to do. But I think that 3986 probably needs an update to accommodate 6874 zone Ids.

sethmlarson:
Sounds good to me, I'll create an issue and PR for that.

sethmlarson:
Pushing rfc3986 master to this PR for testing purposes, will need another release of rfc3986 before we can merge this PR.

test/test_util.py
```diff
@@ -220,6 +221,10 @@ def test_parse_url(self, url, expected_url):
 
     @pytest.mark.parametrize('url, expected_url', parse_url_host_map)
     def test_unparse_url(self, url, expected_url):
+
+        if '/../' in url:
```
sigmavirus24: This is odd

sethmlarson:
@sigmavirus24 Is it acceptable to normalize a path of `/..` to `/`? That's what our current behavior is and it seems strange to me.

sigmavirus24:
99% certain the folks who wrote the RFC3986 section on Resolution of Partial references intended it to act like `cd ..`

sigmavirus24:
Nope. Mis-remembering that

        C.  if the input buffer begins with a prefix of "/../" or "/..",
           where ".." is a complete path segment, then replace that
           prefix with "/" in the input buffer and remove the last
           segment and its preceding "/" (if any) from the output
           buffer; otherwise,

Source: https://tools.ietf.org/html/rfc3986#section-5.2.4

sethmlarson:
So `/..` -> `/` is expected behavior due to the "(if any)"?

sigmavirus24:
No, reading further, I think `/..` -> `` the `if any` means if it looks like `/foo/` and you have `/..` then you it becomes `/foo` and if you have `/foo` and `/..` would become `/` if I remember correctly. There's examples of input and output for the "remove_dot_segments" routine they describe. I don't recall if I use those to test rfc3986 though

sethmlarson:
Sure, so there needs to be an update to rfc3986 here? I can open another PR if you'd like.

sethmlarson:
I'll note here that `urljoin` normalizes `/..` to `/` (ie `urljoin('http://google.com', '/..') -> 'http://google.com/'`)

sigmavirus24:
So `http://google.com` and `http://google.com/` are equivalent. There's no where else to go in that case.

> Also both http://google.com/foo/ and http://google.com/foo joined with /.. goes to http://google.com/ which are in line with what we currently do?

I think you're saying that resolving `http://google.com/foo/` and `http://google.com/foo` with `/..` in rfc3986 both work as expected. Is that correct?

sethmlarson:
Sorry I'm finding discussing this issue difficult in the text-based format. :) This is what I'm seeing, I don't know if it's expected or not but the `/..` cases look like they're working. The `/../` cases are curious to me.

```python
import rfc3986

uri = rfc3986.URIReference.from_string('/..')
print(uri.resolve_with('http://google.com/'))
# URIReference(scheme='http', authority='google.com', path='/', query=None, fragment=None)

uri = rfc3986.URIReference.from_string('/..')
print(uri.resolve_with('http://google.com'))
# URIReference(scheme='http', authority='google.com', path='/', query=None, fragment=None)

uri = rfc3986.URIReference.from_string('/..')
print(uri.resolve_with('http://google.com/foo/'))
# URIReference(scheme='http', authority='google.com', path='/', query=None, fragment=None)

uri = rfc3986.URIReference.from_string('/..')
print(uri.resolve_with('http://google.com/foo'))
# URIReference(scheme='http', authority='google.com', path='/', query=None, fragment=None)

uri = rfc3986.URIReference.from_string('/../')
print(uri.resolve_with('http://google.com/foo/'))
# URIReference(scheme='http', authority='google.com', path='', query=None, fragment=None)

uri = rfc3986.URIReference.from_string('/../')
print(uri.resolve_with('http://google.com/foo'))
# URIReference(scheme='http', authority='google.com', path='', query=None, fragment=None)
```

sigmavirus24:
Ah, i see what you mean. The algorithm might be slightly off in that we're not appending the final `/` in those two last cases. I don't know why, I thought we followed the pseudo-code closely, but I could be wrong. That said, semantically speaking, I think this is correct

sethmlarson:
I found some RFC 3986 battery tests I can run against the library and see what I find.

sethmlarson:
@sigmavirus24 Another thing I'm seeing are that the requests integration tests are failing, previously it looks like we handled international hosts in our URL parser. I'm not sure if that was even the intention but the previous parser was much simpler.

sigmavirus24:
Pretty sure that's why we need: 

- https://github.com/python-hyper/rfc3986/issues/23
- https://github.com/python-hyper/rfc3986/issues/31

I think Requests already relies on `idna` but I'm not sure that we do here? I think we can `parse` a URIReference but normalization/validation is where we fall down right now.

I think we definitely intended to handle IRIs, I just forgot we handled them when helping us over onto rfc3986.

sethmlarson:
I guess I'll add those issues to my backlog as well.

sethmlarson:
@sigmavirus24 What are your thoughts on adding the following changes to this branches `urllib3.packages.rfc3986` into `rfc3986`? I figured I try things out here before moving them to that repo.

sigmavirus24:
`is_iri` is not something I'd prefer. Instead let's have separate types `IRI` versus `URI`. And a `IRI` can be converted to `URI` via something like an `encode` method.

sigmavirus24:
That is to say "Yes! Let's add it to rfc3986, with these caveats"

sethmlarson:
Sounds good to me, I'll make those updates and once we get (hopefully) one more release I can update this PR.

sethmlarson:
I've opened python-hyper/rfc3986#50, once that is merged we can close this PR out

sethmlarson:
IRI support has landed in rfc3986 v1.3.0 so now this PR can continue! :tada:

sethmlarson:
Woo!!! :D

PR #1534, opened 2019-01-29
Update requires-dist and documentation for Brotli

sethmlarson:
One more! :)


sigmavirus24:
We should (in a separate PR) also move much of our setuptools config to setup.cfg so that we dont' need tow orry about duplicates and updating different requirements, etc. For an example see [rush](/sigmavirus24/rush/tree/master/setup.cfg).

sethmlarson:
I agree the current solution isn't great. Do all setuptools versions support setup.cfg though?

northtree:
@sethmlarson Could I know when we will have the new release version?

PR #1551, opened 2019-03-18
Remove unreachable line

pquentin:
rfc3986 only returns Unicode strings, so `to_input_type` only needs to
encode the result back to UTF-8: it never needs to convert from bytes to
unicode.

This restores 100% coverage.

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -196,8 +196,6 @@ def parse_url(url):
     def to_input_type(x):
         if x is None:
             return None
-        elif is_string and isinstance(x, six.binary_type):
-            return x.decode('utf-8')
         elif not is_string and not isinstance(x, six.binary_type):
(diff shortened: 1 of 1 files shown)


pquentin:
The AppVeyor build failed for an unrelated reason. Retrying.

pquentin:
This time we failed to download a file from python.org. :(

sethmlarson:
Gotta love that build reproducibility.

pquentin:
The error is now `E   TypeError: object of type 'MarkDecorator' has no len()` from the downstream requests test suite. Retrying...

sethmlarson:
I already restarted the build, probably won't work I'm afraid. Those downstream requests tests are rather noisy.

pquentin:
Rebased on top of master.

pquentin:
Thank you!

PR #1567, opened 2019-04-19
Update urllib3.contrib.socks documentation

### LINKED ISSUE #1565: Update our SOCKS documentation
- `socks4a://` is useless, all DNS queries are resolved client-side for SOCKS4.
- `socks5://` should have a warning about DNS queries being resolved client-side and that `socks5h://` is recommended for secure use (Should have been the default on our end).
- PySocks supports IPv6 connections to the SOCKS proxy.
- Need to check all our other assumptions in those docs.

sethmlarson:
Closes #1565.

### DIFF
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -1,25 +1,38 @@
 # -*- coding: utf-8 -*-
 """
 This module contains provisional support for SOCKS proxies from within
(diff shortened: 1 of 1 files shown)


sethmlarson:
Thanks for the review @haikuginger, it looks like we're failing with the same reason as master. Will have to investigate that failure on Windows/Python 3.7. In the meantime @theacodes or @shazow can you force a Squash merge here?

shazow:
:+1:

Issue #1392, opened 2018-06-05
New release process?


theacodes:
Having just gone through the release process the first time myself I found that I wasn't a huge fan of it.

I propose we do the following for releases, but I can be convinced otherwise:

1. Create a new branch, `release-MAJOR.MINOR.PATCH`.
1. Add a single commit to that branch that updates the version in `urllib3/__init__.py` and `CHANGES.rst`.
1. Send a PR to merge that branch into `master`. Someone *must* review.
1. Merge the PR and delete the `HEAD` branch.
1. Tag the *squashed merge commit* with `MAJOR.MINOR.PATCH`.
1. Let Travis upload the package to PyPI when built with tags.

This would mean *removing* the `release` branch.

/cc @urllib3/maintainers for #opinions

theacodes:
Oh I forgot to justify this, here's some benefits:

1. Requires a set of eyeballs *and* CI to pass before a release is cut.
2. Allows non-maintainers to initiate a release. Maintainers can approve the release.

SethMichaelLarson:
What if we did away with release branches entirely and just stuck with master? Like you said, any branch can be a release branch as long as on of us reviews it and it passes CI. Is tagging not sufficient?

theacodes:
That's exactly what I'm proposing. I updated my comment to be a little more explicit on that. The `release-MAJOR.MINOR.PATCH` branch is just short-lived branch to initiate the Pull Request.

SethMichaelLarson:
Sorry I misunderstood from the formatting of the release branches! I'm 100% on board having our release process be automated with this model. Eager to hear from others.

shazow:
The thought behind the master/release dichotomy was for people who wanted to pull the library straight from git (as opposed to pypi), "master" would always point to the latest cutting edge work (which might be broken sometimes as things happen, but that's what you'd base your new patches onto) and "release" would always point to the last release (considered to be more "stable", safe to use in production).

By removing the release branch, the user would need to rely on fetching the latest tag by semantic sorting, which is probably fine?

> Oh I forgot to justify this, here's some benefits:
>
> 1. Requires a set of eyeballs *and* CI to pass before a release is cut.
> 2. Allows non-maintainers to initiate a release. Maintainers can approve the release.

Is this different from making a PR into the `release` branch?

I am generally :+1: for automation.

My only concern would be leaving publishing credentials on Travis, and it would mean that release commits would no longer be GPG-signed by the maintainer?

[...]

PR #1568, opened 2019-04-21
Release 1.25

sethmlarson:
Testing our new release system! I figure we should wait for Monday to release this as it's quite a large one. cc @shazow @theacodes 

Requests already has a PR for unpinning: https://github.com/kennethreitz/requests/pull/5063

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,34 +1,32 @@
 Changes
 =======
 
-dev (master)
-------------
+1.25 (2019-04-22)
+-----------------
 
-* Implemented a more efficient ``HTTPResponse.__iter__()`` method. (Issue #1483)
+* Require and validate certificates by default when using HTTPS (Pull #1507)
 
 * Upgraded ``urllib3.utils.parse_url()`` to be RFC 3986 compliant. (Pull #1487)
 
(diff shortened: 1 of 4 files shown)


CHANGES.rst
```diff
@@ -1,34 +1,32 @@
 Changes
 =======
 
-dev (master)
```
shazow: Is the dev section going away in master? Or was this meant to be merged into release?

sethmlarson: Per #1392 we won't be tagging releases on the release branch anymore, will be doing so on the master branch. Whenever we do additional development on master we'll have to add the `dev (master)` section back to whichever changelog entry comes next.

shazow: Ah, didn't realize we'd be getting rid of `__version__ = 'dev'` in master. Historically, it has been very useful to know if the install version is directly from the repo vs a tagged release. Will the immediate next release also bump the `__version__` to an unreleased version too?

Don't think I mentioned that in #1392 but I won't stand in the way. :)

theacodes:
We can send a follow-up PR right after this to restore `__version__` to `dev` and restore the `dev` entry in `CHANGELOG`.

PR #1586, opened 2019-04-25
Percent-encode invalid characters with request target

### LINKED ISSUE #1585: parse_url no longer accepts [] in the URL
With urllib3 1.25.1:

```
>>> parse_url('https://example.com/somevar[]')
  File "/home/testrunner/testvenv/lib/python2.7/site-packages/urllib3/util/url.py", line 212, in parse_url
    six.raise_from(LocationParseError(url), None)
  File "/home/testrunner/testvenv/lib/python2.7/site-packages/urllib3/packages/six.py", line 718, in raise_from
    raise value
urllib3.exceptions.LocationParseError: Failed to parse: https://example.com/somevar[]
```

Using 1.24.2:

```
>>> parse_url('https://example.com/somevar[]')
Url(scheme='https', auth=None, host='example.com', port=None, path='/somevar[]', query=None, fragment=None)
```

```
>>> parse_url('https://example.com/?somevar["whatever"]')
Url(scheme='https', auth=None, host='example.com', port=None, path='/', query='somevar["whatever"]', fragment=None)
```

This is used by various bits of software (ElasticSearch for instance):

```
>>> parse_url('http://example.com/api/search/?grouping=datasets&q=dontcare&projection=["variables.nonexistingfield"]')
Url(scheme='http', auth=None, host='example.com', port=None, path='/api/search/', query='grouping=datasets&q=dontcare&projection=["variables.nonexistingfield"]', fragment=None)
```

sethmlarson:
Closes #1585. cc @theacodes @sigmavirus24

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -6,7 +6,7 @@ from ..exceptions import LocationParseError
 from ..packages import six, rfc3986
 from ..packages.rfc3986.exceptions import RFC3986Exception, ValidationError
 from ..packages.rfc3986.validators import Validator
-from ..packages.rfc3986 import abnf_regexp, normalizers, compat
+from ..packages.rfc3986 import abnf_regexp, normalizers, compat, misc
 
 
 url_attrs = ['scheme', 'auth', 'host', 'port', 'path', 'query', 'fragment']
@@ -153,7 +153,7 @@ def _encode_invalid_chars(component, allowed_chars, encoding='utf-8'):
     percent_encodings = len(normalizers.PERCENT_MATCHER.findall(
                             compat.to_str(component, encoding)))
 
-    uri_bytes = compat.to_bytes(component, encoding)
+    uri_bytes = component.encode('utf-8', 'surrogatepass')
     is_percent_encoded = percent_encodings == uri_bytes.count(b'%')
 
     encoded_component = bytearray()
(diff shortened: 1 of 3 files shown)


sethmlarson:
Yeah I need to add some tests so we don't regress on this behavior.

sethmlarson:
@theacodes This is ready for review I believe.

sethmlarson:
I thought about putting the `_encode_invalid_chars` calls within the conditional for normalizable schemes but `http+unix` isn't impacted by changing the path or query so I think we're safe to unconditionally do that operation?

sethmlarson:
So it's going to sound strange but I can't think of any scenarios that would get us an exception out of the `validator.check_validity_of()` branch? We already percent-encode all characters that aren't valid within path, query, and fragment and the authority (userinfo, host, and port) are handled in the `uri_ref.normalize()`. Might be a good job for hypothesis to find some good test cases if there even are any?

src/urllib3/util/url.py
```diff
@@ -252,7 +251,7 @@ def idna_encode(name):
             *validator.COMPONENT_NAMES
         ).validate(uri_ref)
     except ValidationError:
-        six.raise_from(LocationParseError(url), None)
+        return six.raise_from(LocationParseError(url), None)
```
sigmavirus24: Is the `return` here actually necessary?

sethmlarson: Not at all, however IDEs don't recognize that `six.raise_from()` won't allow execution to propagate. Should I remove it? Don't mind either way.

pquentin:
I was wondering, is it expected that util/url.py is not at 100% coverage?

sethmlarson:
Yeah, I need to figure out a scenario and investigate the failure on 2.7.

sethmlarson:
@pquentin And by "yeah" I mean "no", my Minnesotan is leaking through. :)

PR #1587, opened 2019-04-28
Implement max_retry_wait_length for Retry objects

### LINKED ISSUE #1338: Retry.get_retry_after does't respect time-out

isidentical:
Retry object gets an extra parameter at initalization for max_retry_wait_length
and when it calls get_retry_after with a response that contains Retry-After header
it checks if the value of that greater than max_retry_wait_length. If it
is, then a ExceedingWaitTime exception will be raised.

Fixes #1338

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ dev (master)
 ------------
 
 * Change ``is_ipaddress`` to not detect IPvFuture addresses. (Pull #1583)
+* Implement ``max_retry_wait_length`` for Retry objects. (Pull #1587)
 
 
 1.25.1 (2019-04-24)
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -13,6 +13,7 @@ from ..exceptions import (
     ReadTimeoutError,
     ResponseError,
     InvalidHeader,
+    ExceedingWaitTime,
 )
 from ..packages import six
 
@@ -144,6 +145,10 @@ class Retry(object):
         Sequence of headers to remove from the request when a response
         indicating a redirect is returned before firing off the redirected
         request.
+
+    :param int max_retry_wait_length:
+        :raises urllib3.exceptions.ExceedingWaitTime: if response attempts to give
+        longer wait length than specified.
     """
 
     DEFAULT_METHOD_WHITELIST = frozenset([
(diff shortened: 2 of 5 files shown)


pquentin:
@sethmlarson For what it's worth, the Python 3.5 macOS build is failing on the master branch and this pull request because of https://github.com/tox-dev/tox/issues/809#issuecomment-443436586. Indeed, [lazy-object-proxy now has a setup_requires dependency](https://github.com/lazy-object-proxy/ionelmc/python-lazy-object-proxy/commit/fdc7797e9947dfe512a21173c2276324e32a9a87) which is installed by easy_install. But this uses the Travis macOS OpenSSL which only supports TLS 1.1 while PyPI wants TLS 1.2.

sethmlarson:
Cycling to get an automatic rebase for CI

isidentical:
any update @sethmlarson

CHANGES.rst
```diff
@@ -24,6 +24,7 @@ dev (master)
 -------------------
 
 * Change ``is_ipaddress`` to not detect IPvFuture addresses. (Pull #1583)
+* Implement ``max_retry_wait_length`` for Retry objects. (Pull #1587)
```
sethmlarson: This might be my bad, but can you put this entry in the proper location.

src/urllib3/util/retry.py
```diff
@@ -270,7 +277,15 @@ def get_retry_after(self, response):
         if retry_after is None:
             return None
 
-        return self.parse_retry_after(retry_after)
+        retry_after = self.parse_retry_after(retry_after)
+
+        if self.max_retry_wait_length != -1 and retry_after > self.max_retry_wait_length:
+            raise ExceedingWaitTime(
```
sethmlarson: Instead of raising an exception what are your thoughts on limiting the amount of time that would be slept in the `sleep_for_retry` method? I guess I don't understand the use-case for raising here.

src/urllib3/util/retry.py
```diff
@@ -173,6 +179,7 @@ def __init__(
         history=None,
         respect_retry_after_header=True,
         remove_headers_on_redirect=DEFAULT_REDIRECT_HEADERS_BLACKLIST,
+        max_retry_wait_length=-1,
```
sethmlarson: Let's have this default to `None` instead of an integer value.

Issue #1553, opened 2019-03-18
CRLF injection vulnerability


hannob:
At https://bugs.python.org/issue36276 there's an issue in Python's urllib that an attacker controlling the request parameter can inject headers by injecting CR/LF chars.

A commenter mentions that the same bug is present in urllib3:
https://bugs.python.org/issue36276#msg337837

So reporting it here to make sure it gets attention.

haikuginger:
I confirmed that I was not able to reproduce as of the current master. Will investigate to see what the bisection point was.

haikuginger:
Verified as fixed in #1487, but that change has not been released. I recommend we release 1.24.2 with that change as soon as possible, @sethmlarson @theacodes.

theacodes:
I will try to cut a release soon, unless Seth wants to beat me to it. Thank
you, Jess.

sethmlarson:
Without #1531 or IRI support in rfc3986 releasing master in it's current state will break backwards compatibility with international URLs.

haikuginger:
I've done some checking and `requests` appears to have handling so indirect consumers via requests should not be vulnerable through us, if that's a useful datum to have.

vstinner:
CVE-2016-5699, CVE-2019-9740 and CVE-2019-9947 have been assigned to HTTP Header Injection vulnerabilities in Python: see https://bugs.python.org/issue30458 for more information.

Plazmaz:
I was able to reproduce this with `requests.get` using urllib, haven't tested with urllib3 though.

carnil:
CVE-2019-11236 was assigned for this issue.

rschultheis:
👋  Hi!  I've noticed that CVE-2019-11236 has been assigned to the CRLF injection issue described here.  It seems that the library has been patched in GitHub, but no new release has been made to pypi.  Will a new release containing the fix be made to pypi soon?    Based on @theacodes  [comment](https://github.com/urllib3/urllib3/issues/1553#issuecomment-474032699) it seems like a release was going to be made, but I also see her status has her perhaps unavailable.  Is someone else perhaps able to cut a new release into pypi?

sethmlarson:
@rschultheis [Per my comment](https://github.com/urllib3/urllib3/issues/1553#issuecomment-474046652) we cannot cut a release until IRI support is fixed in rfc3986.

rschultheis:
Thank you @sethmlarson ,  I will watch that PR 🙇

[...]

PR #1591, opened 2019-04-30
prevent CVE-2019-9740 in 1.24.x

ryanpetrello:
adapted from https://github.com/python/cpython/pull/12755

related: https://github.com/urllib3/urllib3/issues/1553

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 from collections import namedtuple
+import re
 
 from ..exceptions import LocationParseError
 
@@ -9,6 +10,8 @@ url_attrs = ['scheme', 'auth', 'host', 'port', 'path', 'query', 'fragment']
 # We only want to normalize urls with an HTTP(S) scheme.
 # urllib3 infers URLs without a scheme (None) to be http.
 NORMALIZABLE_SCHEMES = ('http', 'https', None)
+
+_contains_disallowed_url_pchar_re = re.compile('[\x00-\x20\x7f]')
 
 
 class Url(namedtuple('Url', url_attrs)):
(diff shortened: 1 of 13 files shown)


sethmlarson:
I missed a few things on the fast-forward it looks like.

ryanpetrello:
@sethmlarson just rebased

sethmlarson:
Can you update the dotted_fqdn test the same way as it's update in master? ie skip it on GCE

ryanpetrello:
@sethmlarson would it make sense to just cherry-pick https://github.com/urllib3/urllib3/commit/e3aebd4e79779ad4eef902e81ebbb6ec54335dfc ?

ryanpetrello:
@sethmlarson do you know what's up w/ the `lint` failures?  Maybe a newer version of pycodestyle, or some new rules got applied in `master` to work around it?

How would you like me to proceed with that?

sethmlarson:
Might have to apply this commit to the branch as well: d0255e01b616f02b45eb207459937a9666e13af6

ryanpetrello:
@sethmlarson looks like Travis is still unhappy :/

src/urllib3/util/url.py
```diff
@@ -155,6 +158,11 @@ def parse_url(url):
         # Empty
         return Url()
 
+    # Prevent CVE-2019-9740.
+    # adapted from https://github.com/python/cpython/pull/12755
+    if _contains_disallowed_url_pchar_re.search(url):
+        raise LocationParseError("URL can't contain control characters. {}".format(url))
```
gpshead: I recommend using a repr of the URL rather than the raw url containing the offending text in the error message.  `{!r}` iirc.

ryanpetrello:
Thanks @gpshead, addressed.

ryanpetrello:
@sethmlarson I think I've got the correct things ported from `master` to make Travis happy now.  It came down to a few new decorators for skipping tests in certain environments, and some actual flake8 updates.

ryanpetrello:
image

pypy on 27 is failing for me, but it's also failing on recent commits to `master`, and it looks like it's non-voting in Travis.

ryanpetrello:
@sethmlarson 

Thanks for the quick turn around on this 😄.

At the risk of being a nag, what are the odds this will wind up in an official release on PyPI soon-ish?

PR #1592, opened 2019-05-01
Release 1.24.3

sethmlarson:


ryanpetrello:
🎉

sethmlarson:
@ryanpetrello So due to the downstream tests for requests failing we might need to change course a bit, maybe we should instead percent-encode all invalid bytes in a URL. Bring it more in line with what happens in 1.25.x to mitigate this issue.

ryanpetrello:
@sethmlarson something like this, perhaps?

```python
diff --git a/src/urllib3/util/url.py b/src/urllib3/util/url.py
index e8e1bd7..44e2031 100644
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -3,6 +3,7 @@ from collections import namedtuple
 import re

 from ..exceptions import LocationParseError
+from ..packages.six.moves.urllib.parse import quote


 url_attrs = ['scheme', 'auth', 'host', 'port', 'path', 'query', 'fragment']
@@ -160,8 +161,7 @@ def parse_url(url):

     # Prevent CVE-2019-9740.
     # adapted from https://github.com/python/cpython/pull/12755
-    if _contains_disallowed_url_pchar_re.search(url):
-        raise LocationParseError("URL can't contain control characters. {!r}".format(url))
+    url = _contains_disallowed_url_pchar_re.sub(lambda match: quote(str(match.group())), url)

     scheme = None
     auth = None
```

ryanpetrello:
@sethmlarson does this need a rebase to pass?

ryanpetrello:
@sethmlarson also, if it would be helpful, I could throw this branch into my integration test suite for https://github.com/ansible/awx and see if anything troubling falls out (I'll go kick that off soon).

sethmlarson:
That would be awesome. :) If you'd like to be a champion of maintaining downstream tests for ansible we'd gladly accept a contribution for that to 1.24-series and master branch.

sethmlarson:
And also I don't think it needs a rebase, Travis PR builds automatically build against the hypothetical merge commit.

PR #1593, opened 2019-05-01
avoid CVE-2019-9740 in 1.24.x by percent-encoding invalid path characters

ryanpetrello:
this is to avoid breaking changes in downstream libraries like requests

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -3,6 +3,7 @@ from collections import namedtuple
 import re
 
 from ..exceptions import LocationParseError
+from ..packages.six.moves.urllib.parse import quote
 
 
 url_attrs = ['scheme', 'auth', 'host', 'port', 'path', 'query', 'fragment']
@@ -160,8 +161,7 @@ def parse_url(url):
 
     # Prevent CVE-2019-9740.
     # adapted from https://github.com/python/cpython/pull/12755
-    if _contains_disallowed_url_pchar_re.search(url):
-        raise LocationParseError("URL can't contain control characters. {!r}".format(url))
+    url = _contains_disallowed_url_pchar_re.sub(lambda match: quote(match.group()), url)
 
     scheme = None
     auth = None
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -200,10 +200,26 @@ class TestUtil(object):
         with pytest.raises(ValueError):
             parse_url('[::1')
 
-    def test_parse_url_contains_control_characters(self):
+    @pytest.mark.parametrize('url, expected_url', [
+        (
+            'http://localhost/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:',
+            Url('http', host='localhost', port=None,
+                path='/%20HTTP/1.1%0D%0AHEADER:%20INJECTED%0D%0AIgnore:')
+        ),
+        (
+            u'http://localhost/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:',
+            Url('http', host='localhost', port=None,
+                path='/%20HTTP/1.1%0D%0AHEADER:%20INJECTED%0D%0AIgnore:')
+        ),
+        (
+            'http://localhost/ ?q=\r\n',
+            Url('http', host='localhost', path='/%20', query='q=%0D%0A')
+        ),
+    ])
+    def test_parse_url_contains_control_characters(self, url, expected_url):
         # see CVE-2019-9740
-        url = parse_url('http://localhost:8000/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:')
-        assert url.path == '/%20HTTP/1.1%0D%0AHEADER:%20INJECTED%0D%0AIgnore:'
+        url = parse_url(url)
+        assert url == expected_url
 
     def test_Url_str(self):
         U = Url('http', host='google.com')
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -202,8 +202,8 @@ class TestUtil(object):
 
     def test_parse_url_contains_control_characters(self):
         # see CVE-2019-9740
-        with pytest.raises(LocationParseError):
-            parse_url('http://localhost:8000/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:')
+        url = parse_url('http://localhost:8000/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:')
+        assert url.path == '/%20HTTP/1.1%0D%0AHEADER:%20INJECTED%0D%0AIgnore:'
 
     def test_Url_str(self):
         U = Url('http', host='google.com')


ryanpetrello:
@gpshead you have any thoughts on this approach, since you're the author of the original fix in cpython?

src/urllib3/util/url.py
```diff
@@ -160,8 +161,7 @@ def parse_url(url):
 
     # Prevent CVE-2019-9740.
     # adapted from https://github.com/python/cpython/pull/12755
-    if _contains_disallowed_url_pchar_re.search(url):
-        raise LocationParseError("URL can't contain control characters. {!r}".format(url))
+    url = _contains_disallowed_url_pchar_re.sub(lambda match: quote(str(match.group())), url)
```
gpshead: the call to str() on match.group() seems suspicious; shouldn't it already be a str?  when won't it be and why is calling str() useful rather than checking for PY2 unicode and explicitly encoding?

test/test_util.py
```diff
@@ -202,8 +202,8 @@ def test_parse_url_invalid_IPv6(self):
 
     def test_parse_url_contains_control_characters(self):
         # see CVE-2019-9740
-        with pytest.raises(LocationParseError):
-            parse_url('http://localhost:8000/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:')
+        url = parse_url('http://localhost:8000/ HTTP/1.1\r\nHEADER: INJECTED\r\nIgnore:')
```
gpshead: Is this the API people _want_?  It feels odd to me that something called parse_url would perform quoting on a subset of characters that don't fall into its parsing model.

per the existing docs at https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.url.parse_url

It doesn't seem like quoting of a subset of characters would be expected rather than saying "this isn't a valid URL".

ryanpetrello: > Is this the API people want?

Yea, that's a fair question.

https://github.com/urllib3/urllib3/pull/1591 is the original approach I took, but it looks like it doesn't play nicely with requests' tests downstream:

https://github.com/urllib3/urllib3/pull/1592#issuecomment-488408927
https://travis-ci.org/urllib3/urllib3/jobs/527011269#L512

Maybe there's a better way to bridge this gap between resolving this CVE and addressing @sethmlarson's comment at https://github.com/urllib3/urllib3/pull/1592#issuecomment-488408927 ?

src/urllib3/util/url.py
ryanpetrello: You're right, this was cruft from when I had a typo and was debugging.

ryanpetrello:
@sethmlarson:

```shell
[CVE-2019-9740-quote:●] ~/dev/urllib3 TRAVIS_BUILD_DIR=`pwd` DOWNSTREAM=requests ./_travis/run.sh
============================================================================== test session starts ==============================================================================
platform linux -- Python 3.6.5, pytest-4.0.2, py-1.8.0, pluggy-0.9.0
rootdir: /home/ryanpetrello/dev/urllib3/requests, inifile: pytest.ini
plugins: mock-1.10.4, httpbin-0.3.0
collected 544 items

tests/test_help.py ...                                                                                                                                                    [  0%]
tests/test_hooks.py ...                                                                                                                                                   [  1%]
tests/test_lowlevel.py ...............                                                                                                                                    [  3%]
tests/test_packages.py ...                                                                                                                                                [  4%]
tests/test_requests.py .........................127.0.0.1 - - [02/May/2019 09:44:50] "GET /get HTTP/1.1" 200 134
.................127.0.0.1 - - [02/May/2019 09:44:50] "GET /get HTTP/1.1" 200 218
..127.0.0.1 - - [02/May/2019 09:44:50] "GET /user-agent HTTP/1.1" 200 29
............................................127.0.0.1 - - [02/May/2019 09:44:51] "GET /status/404 HTTP/1.1" 404 0
.127.0.0.1 - - [02/May/2019 09:44:51] "GET /status/500 HTTP/1.1" 500 0
..............pytest-httpbin server hit an exception serving request: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:858)
attempting to ignore so the rest of the tests can run
........................................... [ 31%]
.............................................127.0.0.1 - - [02/May/2019 09:44:52] "GET /stream/4 HTTP/1.1" 200 496
.x.......................................................................................                                    [ 55%]
tests/test_structures.py ....................                                                                                                                             [ 59%]
tests/test_testserver.py ......s....                                                                                                                                      [ 61%]
tests/test_utils.py ..s.................................................................................................................................................. [ 88%]
.............................................ssssssssss.....                                                                                                              [100%]

============================================================== 531 passed, 12 skipped, 1 xfailed in 63.59 seconds ===============================================================
```

test/test_util.py
sethmlarson: Thanks @ryanpetrello for putting this together and @gpshead for reviewing the PR.

@gpshead Being a client on the web is a game of interoperability. If we change our policy towards URLs to be strict it'll just strand users and force them to use other solutions besides urllib3 / requests. There are tons of URLs in the wild that contain spaces and aren't trying to inject headers.

The previous `parse_url` was basically a delimiter splitter and just knew that requests would handle the percent-encoding and invalid character problem. But we have users too that depend on us providing a secure solution so we decided we'd percent encode to protect our users and maintain compatibility with upstream.

sethmlarson:
@ryanpetrello I pushed a commit to the branch to add a few more test cases. Take a look and tell me what you think. :) (May not work 100% or have a typo since I used the GitHub web editor)

ryanpetrello:
Just saw that, and makes sense to me.  Let's see how Travis-CI feels about it.

ryanpetrello:
@sethmlarson fixed and re-uploaded

sethmlarson:
If you could run again against requests downstream that'd be great :)

ryanpetrello:
@sethmlarson test results:

```shell
~/dev/urllib3 tox -- -k control_characters\
...
  flake8-py3: commands succeeded
  py27: commands succeeded
ERROR:  py34: InterpreterNotFound: python3.4
ERROR:  py35: InterpreterNotFound: python3.5
  py36: commands succeeded
ERROR:  py37: InterpreterNotFound: python3.7
  pypy: commands succeeded
```

```shell
~/dev/urllib3 TRAVIS_BUILD_DIR=`pwd` DOWNSTREAM=requests ./_travis/run.sh
531 passed, 12 skipped, 1 xfailed in 68.66 seconds
```

gpshead:
> a game of interoperability. If we change our policy towards URLs to be strict it'll just strand users and force them to use other solutions besides urllib3 / requests. There are tons of URLs in the wild that contain spaces and aren't trying to inject headers.

sethmlarson:
No you're fine I can merge as is. :)

Kronopt:
Am I correct in assuming this PR fix [CVE-2019-11236](https://nvd.nist.gov/vuln/detail/CVE-2019-11236) as well?
This PR is mentioned as a fix for CVE-2019-11236 in #1553, but it is only described as a fix for [CVE-2019-9740](https://nvd.nist.gov/vuln/detail/CVE-2019-9740).

sethmlarson:
Yes this patch also fixes that CVE.

PR #1601, opened 2019-05-07
Update connectionpool.py

sebastiaanspeck:
Make an API_TOKEN replaced by a REDACTED string to overcome that a api_token is in the logging


sethmlarson:
Thank you for opening this PR! This seems very specific a singular use-case rather than for a broader audience. For instance what if another user passed their API token via a different query key?

Perhaps you can achieve what you want via installing a custom logging handler for the `urllib3.connectionpool` logger?

PR #1604, opened 2019-05-11
Drop Python 3.4

pquentin:
It reached its end-of-life on 2019-03-18.

As far as I can tell, remaining references are either still valid (eg. ssl.create_default_context exists since Python 3.4) or in contrib packages like rfc3986 or scripts like _appveyor/install.ps1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,6 @@ matrix:
     # Unit tests
     - python: 2.7
       env: NOX_SESSION=test-2.7
-    - python: 3.4
-      env: NOX_SESSION=test-3.4
     - python: 3.5
       env: NOX_SESSION=test-3.5
     - python: 3.6
@@ -69,9 +67,6 @@ matrix:
     - language: generic
       os: osx
       env: NOX_SESSION=test-2.7
-    - language: generic
-      os: osx
-      env: NOX_SESSION=test-3.4
     - language: generic
       os: osx
       env: NOX_SESSION=test-3.5
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -26,7 +26,6 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
     # Mac OS setup.
     case "${NOX_SESSION}" in
         test-2.7) MACPYTHON=2.7.15 ;;
-        test-3.4) MACPYTHON=3.4.4 ;;
         test-3.5) MACPYTHON=3.5.4 ;;
         test-3.6) MACPYTHON=3.6.7 ;;
         test-3.7) MACPYTHON=3.7.1 ;;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,11 +12,6 @@ environment:
       PYTHON_ARCH: "64"
       NOX_SESSION: "test-2.7"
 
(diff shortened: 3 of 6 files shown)


theacodes:
This is really up to pip. Has pip dropped 3.4 support? We won't until they
do.

pquentin:
Ah, right. pip deprecated 3.4 but did not remove it yet. They forgot to remove it in 19.1. See https://github.com/pypa/pip/issues/6106

I also need to figure out why the coverage is decreasing.

pquentin:
There is a section in wait.py that is covered by Python 3.4 and Python 2.7. After this pull request it is only covered by Python 2.7, but codecov does not receive the 2.7 coverage info. This is why the coverage drops.

Maybe https://github.com/urllib3/urllib3/pull/1596 only fixed the coverage the issue for Python 3?

sethmlarson:
I think there's an issue in our `.coveragerc` file.

pquentin:
All the Python 2.7 coverage uploads are failing in Travis, I think we need to fix that first

pquentin:
Closing as this PR now conflicts and pip won't remove Python 3.4 support until pip 19.2, in order to wait for six months between the deprecation and the removal: https://github.com/pypa/pip/issues/6106#issuecomment-453651209. (I said above that pip should have removed Python 3.4 in 19.1, but I had missed that comment, sorry.)

pquentin:
pip finally dropped 3.4 support in 19.2 thanks to @hugovk in https://github.com/pypa/pip/pull/6685, so we can do this in urllib3 now.

sethmlarson:
Woo!!!

pquentin:
Rebased, but 3.7 fails on Appveyor. Closing/reopening.

pquentin:
The Python 3.7 Windows failure is reproducible. I compared the [last good 3.7 build](https://ci.appveyor.com/project/urllib3/urllib3/builds/26274262/job/9678kar0ro9ohn1j) with the [first bad 3.7 build](https://ci.appveyor.com/project/urllib3/urllib3/builds/26524209/job/vshduc3gpd898dqp), and noticed that Python was updated from 3.7.3 to 3.7.4, and OpenSSL from 1.1.0j to 1.1.1c as part of https://bugs.python.org/issue34631.

Here's my understanding of the failures:

 *  TestConnectionPool.test_source_address_error appears to fail because the [InvalidUrl is now being raised in the standard library](https://github.com/python/cpython/blob/3.7/Lib/http/client.py#L1114-L1118) and I guess we no longer catch the correct exception? But why only on Windows?
 * TestHTTPS.test_client_no_intermediate and TestHTTPS_TLSv1_3.test_client_no_intermediate seem to fail because the OpenSSL exception is being converted into a SocketError in https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py#L345-L346, maybe because the execution path changed?
 * TestHTTPS_TLSv1_3.test_set_ssl_version_to_tls_version fails because there are no certs defined in TestHTTPS_TLSv1_3, so self.certs["ssl_version"] can't work

I'm not planning to continue working on this, sorry!

sethmlarson:
Thanks for your investigation @pquentin. No problem. :)

PR #1607, opened 2019-05-14
Improve implementation of respect_retry_after_header

Eronarn:
`respect_retry_after_header` is accepted as an argument to `Retry`, but it isn't propagated on `Retry.new()`. This means that after the first retry is incremented, this setting will be lost.

Additionally, this setting isn't properly applied in cases where the `retry_after` header is present, but the intent is _not_ to respect it. This edge case can occur if you explicitly allow retries for the same codes as `RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])`, but _don't_ respect the value in the header, and instead fall back to your `Retry`'s exponential backoff settings. In our case, we're dealing with an API that returns a large static value in this header even though we can actually try again much sooner because rate limits are governed by a token bucket.

I've attached a patch that addresses these two cases and adds a test case for one of them. However, I can't get the test suite to run locally (at all), I'm afraid.

### DIFF
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -210,6 +210,7 @@ class Retry(object):
             raise_on_status=self.raise_on_status,
             history=self.history,
             remove_headers_on_redirect=self.remove_headers_on_redirect,
+            respect_retry_after_header=self.respect_retry_after_header
         )
         params.update(kw)
         return type(self)(**params)
@@ -294,7 +295,7 @@ class Retry(object):
         this method will return immediately.
         """
 
-        if response:
+        if self.respect_retry_after_header and response:
             slept = self.sleep_for_retry(response)
             if slept:
                 return
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -210,7 +210,7 @@ class Retry(object):
             raise_on_status=self.raise_on_status,
             history=self.history,
             remove_headers_on_redirect=self.remove_headers_on_redirect,
-            respect_retry_after_header=self.respect_retry_after_header
+            respect_retry_after_header=self.respect_retry_after_header,
         )
         params.update(kw)
         return type(self)(**params)
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -288,34 +288,26 @@ class TestRetry(object):
         retry = Retry()
         assert retry.parse_retry_after(value) == expected
 
-    @pytest.mark.parametrize('respect_retry_after_header', [
-        True,
-        False
-    ])
-    def test_respect_retry_after_header_propagated(self,
-                                                   respect_retry_after_header):
+    @pytest.mark.parametrize("respect_retry_after_header", [True, False])
+    def test_respect_retry_after_header_propagated(self, respect_retry_after_header):
 
         retry = Retry(respect_retry_after_header=respect_retry_after_header)
         new_retry = retry.new()
-        assert new_retry.respect_retry_after_header \
-            == respect_retry_after_header
+        assert new_retry.respect_retry_after_header == respect_retry_after_header
 
     @pytest.mark.parametrize(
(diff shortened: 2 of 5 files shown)


Eronarn:
I took a glance at the test failures, looks spurious?

pquentin:
Right, they're not related to your changes and happen in other pull requests too.

sethmlarson:
Cycling to get an automatic rebase for CI

Eronarn:
Looks like it passed this time. Is this mergeable? Do I need any additional tests, contributor agreement, signature in blood, etc.? :)

src/urllib3/util/retry.py
```diff
@@ -192,7 +192,8 @@ def new(self, **kw):
             raise_on_redirect=self.raise_on_redirect,
             raise_on_status=self.raise_on_status,
             history=self.history,
-            remove_headers_on_redirect=self.remove_headers_on_redirect
+            remove_headers_on_redirect=self.remove_headers_on_redirect,
+            respect_retry_after_header=self.respect_retry_after_header
```
sethmlarson: Nice catch here!

src/urllib3/util/retry.py
```diff
@@ -274,7 +275,7 @@ def sleep(self, response=None):
         this method will return immediately.
         """
 
-        if response:
+        if self.respect_retry_after_header and response:
```
sethmlarson: Could we add a new unit test that hits this condition?

Eronarn:
@sethmlarson I did this:

- rebased with current master (post-black)
- moved the unit tests for this header from `util.py` to `retry.py` to centralize them
- added an additional unit test
- applied black to the modified files

What's the appropriate way for a changelog bump? Does this become a new tagged release by itself, or is there a "on master but not released" place to put changelogs?

test/test_retry.py
```diff
@@ -271,3 +274,47 @@ def test_retry_set_remove_headers_on_redirect(self):
...
 
         assert list(retry.remove_headers_on_redirect) == ["x-api-secret"]
+
+    @pytest.mark.parametrize("value", ["-1", "+1", "1.0", six.u("\xb2")])  # \xb2 = ^2
+    def test_parse_retry_after_invalid(self, value):
+        retry = Retry()
+        with pytest.raises(InvalidHeader):
+            retry.parse_retry_after(value)
+
+    @pytest.mark.parametrize(
```
sethmlarson: Can we add a test case for an http-date formatted Retry-After value? May require mocking time.time() to get a consistent result.

Eronarn: Added four cases here (whether the time is in the future or not X whether we're respecting the header or not).

Eronarn:
@sethmlarson added the above, how's that look?

test/test_retry.py
```diff
@@ -271,3 +276,68 @@ def test_retry_set_remove_headers_on_redirect(self):
...
+
+        # Date header syntax can specify an absolute date; compare this to the
+        # time in the parametrized inputs above.
+        current_time = mock.MagicMock(
+            return_value=time.mktime(
+                datetime.datetime(year=2019, month=6, day=3, hour=11).timetuple()
+            )
+        )
+
+        with mock.patch("time.sleep") as sleep_mock, mock.patch(
```
sethmlarson: TIL about this construction! Thanks for showing me this. :)

sethmlarson:
This PR is only failing for Python 3.8 due to reasons not related to this PR. Once I resolve those issues I'll cycle CI and merge. Thank you so much for this! :) If you've got more time and want to help out more there are plenty of issues marked "[Contributor Friendly](https://github.com/urllib3/urllib3/issues?q=is%3Aissue+is%3Aopen+label%3A%22Contributor+Friendly+%E2%99%A5%22)" that need eyes on them.

sethmlarson:
Thanks for this! :D

PR #1577, opened 2019-04-23
Update setup.cfg for pytest 4.4 support

felixonmars:
```
Failed: [pytest] section in setup.cfg files is no longer supported, change to [tool:pytest] instead.
```

PR #1611, opened 2019-05-18
Add [tool:pytest] in setup.cfg for pytest

RatanShreshtha:


sethmlarson:
Thanks for this. Can you also [implement the comment](https://github.com/urllib3/urllib3/pull/1577#issuecomment-485802165) for a very similar PR?

RatanShreshtha:
yeah sure

RatanShreshtha:
I tried to change pytest to latest version but CI was failing it seems to work with current version of pytest.

sethmlarson:
Is there any way we could upgrade to the latest? Might have to change some of our parameter tests.

pquentin:
@sethmlarson Is there any reason why we want to couple the setup.cfg fix with a pytest version upgrade? Those two issues seem orthogonal to me

sethmlarson:
Because they're both to do with using pytest 4.x if I understand correctly?

pquentin:
You're right that this is about pytest 4.x. But `tool:pytest` is useful starting from pytest 4.0, and the failing tests issue starts with pytest 4.2! :)

I suspect that pytest 4.2 introduced a bug where `unittest.TestCase` tests with `setUpClass` and inheritance don't work well anymore. But I'd prefer migrating to pure pytest instead of figuring out if this is really a bug or not, because:

 1. I prefer `assert a == 2` to `self.assertEquals(a, 2)`!
 2. [We would like to use pytest-trio in our async branch](https://github.com/python-trio/urllib3/pull/92), but pytest-trio does not support `unittest.TestCase`.

However, the switch is going to change a lot of test code. The idea would be to open a PR here, and then merge the changes into our branch.

Do you think switching to pure pytest is acceptable?

sethmlarson:
I'd love to switch to pure pytest. If you've got the time to put in that work I'd definitely review it. :)

RatanShreshtha:
I can work on pure pytest migration.

pquentin:
@RatanShreshtha Awesome! The first step would be to replace all `self.assert*` calls with Python assertions. And only do that. We have a lot of those:
```

$ rg --no-filename 'self.assert[A-Z]' | tr -s ' ' | cut -d '(' -f1 | sort | uniq -c
 305  self.assertEqual
   1  self.assertFail
   6  self.assertFalse
   2  self.assertGreaterEqual
  19  self.assertIn
  11  self.assertIsInstance
  12  self.assertIsNone
   2  self.assertIsNotNone
   3  self.assertLess
   2  self.assertNotEqual
   6  self.assertNotIn
  59  self.assertRaises
  17  self.assertTrue
   6  with self.assertRaises
```

You can look at the [unittest documentation](https://docs.python.org/3/library/unittest.html#assert-methods) and [pytest assert documentation](https://docs.pytest.org/en/latest/assert.html) to see the changes that need to be made.

For example: 

 * `self.assertEqual(a, b)` → `assert a == b`
 * `self.assertEqual(a, b, c)` → `assert a == b, c`
 * `self.assertIsNone(a)` → `assert a is None`

While you can simply replace `self.assertRaises(ReadTimeoutError, pool.request, 'GET', '/')` with `pytest.raises(ReadTimeoutError, pool.request, 'GET', '/')`, I believe the context manager form is more readable:

```
pytest.raises(ReadTimeoutError):
    pool.request('GET', '/')
```

I think there is a tool that replaces those automatically but I can't find it right now. Maybe those regexes help: http://www.davidchudzicki.com/posts/replace-ugly-assert-methods/ (on Linux you need to replace `-i ".bak"` to just `-i`.)

RatanShreshtha:
Thanks @pquentin  for the information its is really helpful 👍

RatanShreshtha:
I found the tool to change unittest assertions to pytest ones https://pypi.org/project/unittest2pytest/

pquentin:
That's the one! Hopefully it still works.

PR #1614, opened 2019-05-22
WIP: Migrate to pytest

RatanShreshtha:
This PR is for migrating to pytest from uniittest, it is still a work in progress.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,8 +6,8 @@ tornado==5.1.1
 PySocks==1.6.8
 pkginfo==1.4.2
 pytest-timeout==1.3.1
-pytest==3.8.2
-pluggy==0.8.0
+pytest==4.0.0
+pluggy==0.11.0
 
 # https://github.com/ionelmc/python-lazy-object-proxy/issues/30
 lazy-object-proxy==1.4.0
--- a/test/appengine/test_gae_manager.py
+++ b/test/appengine/test_gae_manager.py
@@ -12,7 +12,7 @@ from test.with_dummyserver import test_connectionpool
 # This class is used so we can re-use the tests from the connection pool.
 # It proxies all requests to the manager.
 class MockPool(object):
-    def __init__(self, host, port, manager, scheme='http'):
+    def __init__(self, host, port, manager, scheme="http"):
         self.host = host
         self.port = port
         self.manager = manager
@@ -28,18 +28,15 @@ class MockPool(object):
 
     def _absolute_url(self, path):
         return urllib3.util.url.Url(
-            scheme=self.scheme,
-            host=self.host,
-            port=self.port,
-            path=path).url
+            scheme=self.scheme, host=self.host, port=self.port, path=path
+        ).url
 
 
 # Note that this doesn't run in the sandbox, it only runs with the URLFetch
 # API stub enabled. There's no need to enable the sandbox as we know for a fact
 # that URLFetch is used by the connection manager.
-@pytest.mark.usefixtures('testbed')
+@pytest.mark.usefixtures("testbed")
 class TestGAEConnectionManager(test_connectionpool.TestConnectionPool):
-
     def setUp(self):
         self.manager = appengine.AppEngineManager()
         self.pool = MockPool(self.host, self.port, self.manager)
@@ -49,25 +46,27 @@ class TestGAEConnectionManager(test_connectionpool.TestConnectionPool):
     def test_exceptions(self):
         # DeadlineExceededError -> TimeoutError
         with pytest.raises(urllib3.exceptions.TimeoutError):
-            self.pool.request('GET', '/sleep?seconds=0.005', timeout=0.001)
+            self.pool.request("GET", "/sleep?seconds=0.005", timeout=0.001)
 
         # InvalidURLError -> ProtocolError
         with pytest.raises(urllib3.exceptions.ProtocolError):
-            self.manager.request('GET', 'ftp://invalid/url')
+            self.manager.request("GET", "ftp://invalid/url")
 
         # DownloadError -> ProtocolError
         with pytest.raises(urllib3.exceptions.ProtocolError):
-            self.manager.request('GET', 'http://0.0.0.0')
+            self.manager.request("GET", "http://0.0.0.0")
 
         # ResponseTooLargeError -> AppEnginePlatformError
         with pytest.raises(appengine.AppEnginePlatformError):
-            self.pool.request('GET', '/nbytes?length=33554433')  # One byte over 32 megabtyes.
+            self.pool.request(
+                "GET", "/nbytes?length=33554433"
+            )  # One byte over 32 megabtyes.
 
         # URLFetch reports the request too large error as a InvalidURLError,
         # which maps to a AppEnginePlatformError.
-        body = b'1' * 10485761  # One byte over 10 megabytes.
+        body = b"1" * 10485761  # One byte over 10 megabytes.
         with pytest.raises(appengine.AppEnginePlatformError):
-            self.manager.request('POST', '/', body=body)
+            self.manager.request("POST", "/", body=body)
 
     # Re-used tests below this line.
     # Subsumed tests
@@ -104,25 +103,22 @@ class TestGAEConnectionManager(test_connectionpool.TestConnectionPool):
     test_dns_error = None
 
 
-@pytest.mark.usefixtures('testbed')
-class TestGAEConnectionManagerWithSSL(
-        dummyserver.testcase.HTTPSDummyServerTestCase):
-
+@pytest.mark.usefixtures("testbed")
+class TestGAEConnectionManagerWithSSL(dummyserver.testcase.HTTPSDummyServerTestCase):
     def setUp(self):
         self.manager = appengine.AppEngineManager()
-        self.pool = MockPool(self.host, self.port, self.manager, 'https')
+        self.pool = MockPool(self.host, self.port, self.manager, "https")
 
     def test_exceptions(self):
         # SSLCertificateError -> SSLError
         # SSLError is raised with dummyserver because URLFetch doesn't allow
         # self-signed certs.
         with pytest.raises(urllib3.exceptions.SSLError):
-            self.pool.request('GET', '/')
+            self.pool.request("GET", "/")
 
 
-@pytest.mark.usefixtures('testbed')
+@pytest.mark.usefixtures("testbed")
 class TestGAERetry(test_connectionpool.TestRetry):
-
     def setUp(self):
         self.manager = appengine.AppEngineManager()
         self.pool = MockPool(self.host, self.port, self.manager)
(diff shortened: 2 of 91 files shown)


RatanShreshtha:
@sethmlarson please take a look at this PR.

pquentin:
Thank for your hard work!

However, I suggested to only remove the `self.assert*` calls and nothing else for now because 1/ it's the bulk of the change so it's already tedious to review 2/ if the tests fail, we know an assertion has not been converted correctly. But this pull request unfortunately contains other changes such as the removed `self.addCleanup` calls. And right now the tests don't pass, and it's hard to know why. :(

Now I'm not sure if it's easier to fix the issues or remove the other changes. Anyway, two things that we will eventually have to fix:

 * you can't use `super()` in Python 2.7, you need to use `super(ChildClass, self)`
 * you can't just remove the `addCleanup` calls, because we need to close the connections. One option is teardown_method, and the other one is to use context managers

RatanShreshtha:
@pquentin  I have force pushed a new commit with pytest style assertions and tests are passing on my machine

pquentin:
@RatanShreshtha Fantastic, thanks! You only have a bunch of lint errors to fix: https://travis-ci.org/urllib3/urllib3/jobs/536104158

For example, 

```
        with pytest.raises(
            ConnectTimeoutError):
            pm.request('GET', 'http://example.com',
            timeout=0.001, retries=False)
```

should become

```
        with pytest.raises(ConnectTimeoutError):
            pm.request('GET', 'http://example.com',
                       timeout=0.001, retries=False)
```

or even in this case:

```
        with pytest.raises(ConnectTimeoutError):
            pm.request('GET', 'http://example.com', timeout=0.001, retries=False)
```

since the line is not too long (the current limit is 99 characters).

You can use `nox -s lint` to check for lint errors locally.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -115,36 +119,38 @@ def test_timeout(self):
...
 
         wait_for_socket(ready_event)
         conn = pool._new_conn()
         # FIXME: This assert flakes sometimes. Not sure why.
-        self.assertRaises(ReadTimeoutError,
-                          pool._make_request,
-                          conn, 'GET', '/',
-                          timeout=SHORT_TIMEOUT)
+        with pytest.raises(ReadTimeoutError):
+            pool._make_request(conn, 'GET', '/',
```
pquentin: You don't need this extra newline here

test/with_dummyserver/test_connectionpool.py
```diff
@@ -156,24 +162,26 @@ def test_connect_timeout(self):
...
         big_timeout = Timeout(read=LONG_TIMEOUT, connect=LONG_TIMEOUT)
         pool = HTTPConnectionPool(host, port, timeout=big_timeout, retries=False)
         self.addCleanup(pool.close)
         conn = pool._get_conn()
-        self.assertRaises(ConnectTimeoutError,
-                          pool._make_request,
-                          conn, 'GET', url,
-                          timeout=timeout)
+        with pytest.raises(ConnectTimeoutError):
+            pool._make_request(conn, 'GET', url,
```
pquentin: You don't need this extra newline here

test/with_dummyserver/test_connectionpool.py
```diff
@@ -313,7 +327,7 @@ def test_nagle(self):
         self.addCleanup(conn.close)
         pool._make_request(conn, 'GET', '/')
         tcp_nodelay_setting = conn.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY)
-        self.assertTrue(tcp_nodelay_setting)
+        assert tcp_nodelay_setting
```
pquentin: `assert tcp_nodelay_setting is True` would be better: we don't want to check if the value is [truthy](https://docs.python.org/3/library/stdtypes.html#truth-value-testing), we want to check that it is `True`, and it's not already obvious from the name that `tcp_nodelay_setting` is a boolean.

test/with_dummyserver/test_https.py
```diff
@@ -439,9 +438,9 @@ def test_assert_invalid_fingerprint(self):
                                         'AA:AA:AA:AA:AA:AA:AA:AA:AA'
 
         def _test_request(pool):
-            with self.assertRaises(MaxRetryError) as cm:
+            with pytest.raises(MaxRetryError) as cm:
                 pool.request('GET', '/', retries=0)
-            self.assertIsInstance(cm.exception.reason, SSLError)
+                assert isinstance(cm.exception.reason, SSLError)
```
pquentin: This assert can never run, because the `pool.request()` call raises. You need to remove those two spaces to execute this assert *after* the with block. And yes, `cm` will still be defined at this point. I agree that it's confusing! But [this is how it's intended to work](https://docs.pytest.org/en/latest/assert.html#assertraises).

test/with_dummyserver/test_https.py
```diff
@@ -463,9 +462,9 @@ def test_verify_none_and_bad_fingerprint(self):
 
         https_pool.assert_fingerprint = 'AA:AA:AA:AA:AA:AAAA:AA:AAAA:AA:' \
                                         'AA:AA:AA:AA:AA:AA:AA:AA:AA'
-        with self.assertRaises(MaxRetryError) as cm:
+        with pytest.raises(MaxRetryError) as cm:
             https_pool.request('GET', '/', retries=0)
-        self.assertIsInstance(cm.exception.reason, SSLError)
+            assert isinstance(cm.exception.reason, SSLError)
```
pquentin: You also have two extra spaces here

test/with_dummyserver/test_https.py
```diff
@@ -547,19 +547,23 @@ def new_pool(timeout, cert_reqs='CERT_REQUIRED'):
 
         https_pool = new_pool(Timeout(connect=0.001))
         conn = https_pool._new_conn()
-        self.assertRaises(ConnectTimeoutError, https_pool.request, 'GET', '/')
-        self.assertRaises(ConnectTimeoutError, https_pool._make_request, conn,
-                          'GET', '/')
+        with pytest.raises(ConnectTimeoutError):
+            https_pool.request('GET', '/')
+        with pytest.raises(ConnectTimeoutError):
+            https_pool._make_request(conn,
```
pquentin: You don't need this extra newline

test/with_dummyserver/test_https.py
```diff
@@ -547,19 +547,23 @@ def new_pool(timeout, cert_reqs='CERT_REQUIRED'):
...
+                                     'GET', '/')
 
         https_pool = new_pool(Timeout(connect=5))
-        self.assertRaises(ConnectTimeoutError, https_pool.request, 'GET', '/',
-                          timeout=Timeout(connect=0.001))
+        with pytest.raises(ConnectTimeoutError):
+            https_pool.request('GET', '/',
```
pquentin: You don't need this extra newline

test/with_dummyserver/test_https.py
```diff
@@ -547,19 +547,23 @@ def new_pool(timeout, cert_reqs='CERT_REQUIRED'):
...
+                               timeout=Timeout(connect=0.001))
 
         t = Timeout(total=None)
         https_pool = new_pool(t)
         conn = https_pool._new_conn()
-        self.assertRaises(ConnectTimeoutError, https_pool.request, 'GET', '/',
-                          timeout=Timeout(total=None, connect=0.001))
+        with pytest.raises(ConnectTimeoutError):
+            https_pool.request('GET', '/',
```
pquentin: You don't need this extra newline

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -46,34 +46,36 @@ def test_nagle_proxy(self):
...
     def test_proxy_conn_fail(self):
         host, port = get_unreachable_address()
         http = proxy_from_url('http://%s:%s/' % (host, port), retries=1, timeout=0.05)
         self.addCleanup(http.clear)
-        self.assertRaises(MaxRetryError, http.request, 'GET',
-                          '%s/' % self.https_url)
-        self.assertRaises(MaxRetryError, http.request, 'GET',
-                          '%s/' % self.http_url)
+        with pytest.raises(MaxRetryError):
+            http.request('GET',
```
pquentin: You don't need this extra newline

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -46,34 +46,36 @@ def test_nagle_proxy(self):
...
+                         '%s/' % self.https_url)
+        with pytest.raises(MaxRetryError):
+            http.request('GET',
```
pquentin: You don't need this extra newline

test/appengine/test_gae_manager.py
```diff
@@ -48,43 +48,38 @@ def setUp(self):
 
     def test_exceptions(self):
         # DeadlineExceededError -> TimeoutError
-        self.assertRaises(
-            urllib3.exceptions.TimeoutError,
-            self.pool.request,
-            'GET',
-            '/sleep?seconds=0.005',
-            timeout=0.001)
+        with pytest.raises(
```
pquentin: You have a number of `pytest.raises` calls that include a newline, but you never need one. Usually the call below is not needed either. So you could simply write:

```python3
with pytest.raises(urllib3.exceptions.TimeoutError):
    self.pool.request('GET', '/sleep?seconds=0.005', timeout=0.001)
```

Use `grep 'pytest.raises($' test/*/*.py` or `rg 'pytest.raises\($'` to find them, they're mostly in test_gae_manager.py, but there's also one in test_socks.py.

test/with_dummyserver/test_https.py
```diff
@@ -302,25 +300,26 @@ def test_verified_without_ca_certs(self):
...
         pool.ConnectionCls = None
         self.addCleanup(pool.close)
-        self.assertRaises(SSLError, pool._new_conn)
-        with self.assertRaises(MaxRetryError) as cm:
+        with pytest.raises(SSLError):
+            pool._new_conn()
+        with pytest.raises(MaxRetryError) as cm:
             pool.request('GET', '/', retries=0)
-        self.assertIsInstance(cm.exception.reason, SSLError)
+            assert isinstance(cm.exception.reason, SSLError)
```
pquentin: You also have two extra spaces here.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1058,9 +1069,9 @@ def socket_handler(listener):
         pool = HTTPSConnectionPool(self.host, self.port)
         self.addCleanup(pool.close)
 
-        with self.assertRaises(MaxRetryError) as cm:
+        with pytest.raises(MaxRetryError) as cm:
             pool.request('GET', '/', retries=0)
-        self.assertIsInstance(cm.exception.reason, SSLError)
+            assert isinstance(cm.exception.reason, SSLError)
```
pquentin: You also have two extra spaces here

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1133,11 +1145,12 @@ def request():
             finally:
                 pool.close()
 
-        with self.assertRaises(MaxRetryError) as cm:
+        with pytest.raises(MaxRetryError) as cm:
             request()
-        self.assertIsInstance(cm.exception.reason, SSLError)
+            assert isinstance(cm.exception.reason, SSLError)
```
pquentin: You also have two extra spaces here

RatanShreshtha: It is giving error if it is not in the block of with.

pquentin: Ah! Nice catch. That will teach me to read the documentation I link to. :) The syntax with pytest has changed. You need to use this form:

```python3
with pytest.raises(MaxRetryError) as excinfo:
    pool.request('GET', '/', retries=0)
assert isinstance(excinfo.value.reason, SSLError)
```

(excinfo.value instead of cm.exception)

Sorry about that!

test/with_dummyserver/test_connectionpool.py
RatanShreshtha: `tcp_nodelay_setting` is not a boolean here, as shown by result of pytest 
```
tcp_nodelay_setting = conn.sock.getsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY)
>       assert tcp_nodelay_setting is True
E       AssertionError: assert 4 is True
```

pquentin:
Oops, you messed up your force push somehow, there are many commits now.

test/with_dummyserver/test_connectionpool.py
pquentin: Sorry! Please forget about my comment, then!

dev-requirements.txt
```diff
@@ -6,8 +6,8 @@ tornado==5.1.1
 PySocks==1.6.8
 pkginfo==1.4.2
 pytest-timeout==1.3.1
-pytest==3.8.2
-pluggy==0.8.0
+pytest==4.0.0
```
sethmlarson: Can we unpin this to the most recent version with this current PR?

RatanShreshtha: Not yet there is still some work pending
1. Deleting the unittest.TestCase base class everywhere
2. Replacing all def setUp with def setup_method
3. Delete all other references to unittest (imports, if __name__ == "__main__": unittest.main())
4. declare *TestCase classes as test classes
5. call super().setup_class() manually

pquentin: However we can remove the pluggy dependency as urllib3 does not use directly and the latest version now works. @sethmlarson Maybe that's what you were referring to?

sethmlarson: I'm just wondering what work still needs to be done for us to be using the latest pytest. :) This is a good start down that road though. Is this PR ready for a larger review or?

RatanShreshtha: There are few formatting issues also

pquentin: @sethmlarson Thanks for asking! Yeah, I don't think you need to look at this until my comments are addressed. There are not many of them left!

RatanShreshtha:
I have merged from master and formatted the code using black

PR #1620, opened 2019-05-27
Use Brotli instead of brotlipy by default

felixonmars:
brotlipy is stuck at brotli 0.6 and upstream is inactive. Let's switch
to the official binding by default which is up-to-date.

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -279,7 +279,7 @@ Brotli Encoding
 
 Brotli is a compression algorithm created by Google with better compression
 than gzip and deflate and is supported by urllib3 if the
-`brotlipy <https://github.com/python-hyper/brotlipy>`_ package is installed.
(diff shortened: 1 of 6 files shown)


sethmlarson:
Since brotlipy was actually downloaded more before we made this decision I'd actually prefer we upgrade brotlipy with the desired support. :) Would you be up to making that PR on python-hyper/brotlipy?

felixonmars:
Current PRs in brotlipy wasn't reviewed and the repo has been silent for at least half a year. The most recent release is also two years old. I made a PR before to add shared brotli support with was merged, but never made it into a release.

So... I can work on some updates but not sure if it's going to help or worth the effort :(

sethmlarson:
I believe I have merge rights on that repository. I can talk to Cory to see if I can get publish rights as well. Would you mind making the changes to that project so I can review them there?

sethmlarson:
Going to close this as we'll be sticking with brotlipy for now.

PR #1621, opened 2019-05-27
fix wrapped socket shutdown() signature

badeadan:


badeadan:
> Can we add a simple unit test for this case so we don't regress this behavior?

I'm trying to add a test function in contrib/test_securetransport.py but no matter what I write in that file (including a top level "assert False") it doesn't fail when running: "make test". Any suggestion?

sethmlarson:
Oh that file is out of date! Try installing nox via `python -m pip install nox` and running `nox -e test`. We should upgrade our contributor docs.

badeadan:
> Oh that file is out of date! Try installing nox via `python -m pip install nox` and running `nox -e test`. We should upgrade our contributor docs.

Any hints how to install "Security"? It's needed by test_securetransport.py :

test/contrib/test_securetransport.py:22: Could not import SecureTransport: ImportError('The library Security could not be found',)

sethmlarson:
Should be available on macOS? What version are you running?

badeadan:
Ubuntu 18.04.2 LTS in Docker kuralabs/python3-dev:latest

sethmlarson:
That would explain it, that functionality is only available on macOS. If you don't have a system available get the test case working for PyOpenSSL and then duplicate the test within `test_securetransport.py` and it should work. :)

badeadan:
I noticed you are using 'black' to format the code. Do you use any command line options?

sethmlarson:
Try `nox -e blacken` when you have nox installed.

badeadan:
Will open a new pull request with unit test and additional signature fixes.

PR #1624, opened 2019-05-29
WIP: remove self.addCleanup for pytest migration

RatanShreshtha:
I have opened this PR to get an early feedback on few things like https://github.com/urllib3/urllib3/blob/a1f6d1080d1a5236915c10fb6cbb673c945ba4a9/test/with_dummyserver/test_proxy_poolmanager.py#L34-L48 
What should be the right way to rewrite the test ? 

And 

In https://github.com/urllib3/urllib3/blob/a1f6d1080d1a5236915c10fb6cbb673c945ba4a9/test/with_dummyserver/test_https.py#L526-L569.

I have noticed that first https_pool is created but it is not used is it intentional ?

### DIFF
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -230,13 +230,12 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
 
         self._start_server(request_handler)
         proxy_url = "socks5://%s:%s" % (self.host, self.port)
-        pm = socks.SOCKSProxyManager(proxy_url)
-        self.addCleanup(pm.clear)
-        response = pm.request("GET", "http://16.17.18.19")
+        with socks.SOCKSProxyManager(proxy_url) as pm:
+            response = pm.request("GET", "http://16.17.18.19")
 
-        assert response.status == 200
-        assert response.data == b""
-        assert response.headers["Server"] == "SocksTestServer"
+            assert response.status == 200
+            assert response.data == b""
+            assert response.headers["Server"] == "SocksTestServer"
 
     def test_local_dns(self):
         def request_handler(listener):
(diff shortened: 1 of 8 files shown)


pquentin:
I would rewrite test_nagle_proxy like this:

```
def test_nagle_proxy(self):
        """ Test that proxy connections do not have TCP_NODELAY turned on """
        with proxy_from_url(self.proxy_url) as http:
            hc2 = http.connection_from_host(self.http_host, self.http_port)
            try:
                conn = hc2._get_conn()
                hc2._make_request(conn, "GET", "/")
                tcp_nodelay_setting = conn.sock.getsockopt(
                    socket.IPPROTO_TCP, socket.TCP_NODELAY
                )
                assert tcp_nodelay_setting == 0, (
                    "Expected TCP_NODELAY for proxies to be set "
                    "to zero, instead was %s" % tcp_nodelay_setting
                )
            finally:
                conn.close()
                # alternatively, use hc2._put_conn(conn) to put the connection
                # back into the pool so that its get closed when we exit from
                # the proxy_from_url context
```

(Also, `proxy_from_url` has changed over the time but no longer provides any value. So we could remove it and use `with ProxyManager(self.proxy_url) as http` instead.)

> I have noticed that first https_pool is created but it is not used is it intentional ?

I don't know what was the intention behind https://github.com/urllib3/urllib3/commit/b4e5a8f491eb45ef9bfac5b24b118ad022babf89 but I don't think it's useful.

RatanShreshtha:
thanks @pquentin your version of test looks much more readable than the version I thought, maybe @sethmlarson has more idea about https://github.com/urllib3/urllib3/commit/b4e5a8f491eb45ef9bfac5b24b118ad022babf89.

sethmlarson:
I think you're free to remove the duplicate pool.

RatanShreshtha:
I have removed all the `self.addCleanup` calls and the tests seems to pass everywhere except on python 3.8, I don't know why it is so ?

RatanShreshtha:
All the tests are failing for python3.8 in travis for all the pull requests, but I am able to run the tests for python3.8 in my local machine.

RatanShreshtha:
what is the next thing to do for this PR?

sethmlarson:
All that's left is for me to remember to merge this! ;)

sethmlarson:
Thanks for this @RatanShreshtha!

PR #1626, opened 2019-05-30
match WrapSocket with socket.socket signature

badeadan:
reopened https://github.com/urllib3/urllib3/pull/1621 after adding unit test and extra signature fixes

----

Update WrapSocket methods in pyopenssl and securetransport
to match socket.socket. See:

https://docs.python.org/2.7/library/socket.html
https://docs.python.org/3.7/library/socket.html


sethmlarson:
Why do you need this change? Wondering because it's been so long since all of these interfaces have been implemented.

badeadan:
I don't need it that much. I discovered the issue while integrating with a piece of code that was calling socket.shutdown(SHUT_RDWR) and failed because of the missing parameter in WrapSocket. I thought it would be nice to update signatures to match wrapped socket.

In case you decide to accept this PR then I need some help understanding why  CodeCov fails so I can fix it.

felixfontein:
I've encountered the problem that `urllib3.contrib.pyopenssl.WrappedSocket` does not support the `how` argument for `shutdown()` (when calling `socket.shutdown(SHUT_RDWR)`). Unfortunately, just adjusting the signature does **not** fix this. The implementation calls `self.connection.shutdown()`, which also shuts down the connection for reading, which is **not** what the caller wants when calling `socket.shutdown(SHUT_RDWR)`. I have absolutely no idea how that can be properly implemented, though...

(I've encountered this when using the Docker SDK for Python (which under the hoods uses requests, which in turn uses urllib3) to send something to the stdin of a container: I need to close the connection for writing to indicate that stdin is EOF, and for non-SSL sockets, this can be achieved with `socket.shutdown(SHUT_RDWR)`.)

felixfontein:
Ok, after some more playing around, I *think* that `shutdown()` also works, but after calling it, I'll get a `OpenSSL.SSL.ZeroReturnError` exception when calling `socket.recv(4096)` and EOF actually happened. When I treat that exception as "EOF", it seems to work.

pquentin:
Merged from master into your branch to fix the conflicts caused by the isort pull request.

sethmlarson:
This PR needs substantial changes to be up-to-date with `main` and hasn't moved much since being opened. I think I'm going to close this one for now.

PR #1629, opened 2019-05-31
Remove the Makefile as it's not maintained

sethmlarson:
Just confuses new users. We need to improve our onboarding docs.


sethmlarson:
Wish that GitHub had a "rebase and push" button I could push here. :)

Issue #1625, opened 2019-05-30
Add time unit (seconds) to Timeout docs


bgrant:
Neither the [docs](https://urllib3.readthedocs.io/en/latest/user-guide.html#using-timeouts) nor the [docstrings](https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/timeout.py#L20) say what unit timeouts should be provided in.  Looking at the code, it appears to be "seconds", but it would be nice if the docs said this explicitly.

sigmavirus24:
/joking

But how else will we convince people to read our beautiful beautiful source code?

sethmlarson:
@bgrant Thanks for opening this issue! I agree that this would be a good addition. Do you think this is something you could accomplish?

bgrant:
Sure- happy to.

PR #1630, opened 2019-05-31
Add time units to Timeout docs

bgrant:
In both docstrings and in the user guide, specify that timeouts are specified in seconds.

Fixes https://github.com/urllib3/urllib3/issues/1625


sethmlarson:
Thanks for this! :)

PR #1633, opened 2019-06-04
Allow Python 3.8-dev to fail

sethmlarson:


sethmlarson:
All our CI builds have been broken for a few days, I can only imagine it being due to the amount of issues that CPython 3.8 pre-beta is having.

Issue #1634, opened 2019-06-06
Enable TLS 1.3 post-handshake authentication


tiran:
urllib3 should enable post-handshake authentication by default:

* https://docs.python.org/3/library/ssl.html#ssl.SSLContext.post_handshake_auth
* https://tools.ietf.org/html/rfc8446#section-4.6.2

PHA is well defined for HTTPS and necessary to support path-specific TLS client cert authentication with TLS 1.3. Without PHA, a server cannot request client cert authentication based on HTTP request parameters like method or path.

PHA is disabled by default, because some protocols cannot deal with PHA yet. In my understanding it is always safe to enable PHA for a HTTP/1.1 client. There are some theoretical issue with HTTP/2.0, but that does not affect urllib3.

PR #1635, opened 2019-06-06
Enable TLS 1.3 post-handshake authentication

tiran:
Fixes: https://github.com/urllib3/urllib3/issues/1634
Signed-off-by: Christian Heimes <christian@python.org>

### DIFF
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -149,3 +149,18 @@ def test_wrap_socket_default_loads_default_certs(monkeypatch):
     ssl_.ssl_wrap_socket(sock)
 
     context.load_default_certs.assert_called_with()
+
+
+@pytest.mark.parametrize(
+    ["pha", "expected_pha"], [(None, None), (False, True), (True, True)]
+)
+def test_create_urllib3_context_pha(monkeypatch, pha, expected_pha):
+    context = mock.create_autospec(ssl_.SSLContext)
+    context.set_ciphers = mock.Mock()
+    context.options = 0
+    context.post_handshake_auth = pha
+    monkeypatch.setattr(ssl_, "SSLContext", lambda *_, **__: context)
+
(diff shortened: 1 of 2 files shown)


sethmlarson:
Thanks for this! Could you add a changelog entry for this change and a simple test that shows that this value is either being set to `True` or is not an attribute for the `SSLContext`? I'm not sure if we can easily test the functionality within our current test suite.

tiran:
There is currently a bug in either OpenSSL or Python that causes a side effect when PHA is enabled: https://bugs.python.org/issue37428

Let me first sort out the bug, then we can land this PR.

stratakis:
https://bugs.python.org/issue37428 has been resolved on the python side.

tiran:
@sethmlarson I have rebased the PR and added an additional test case for PHA. Let's see of AppVeyor passes now.

pquentin:
AppVeyor fails because of an unrelated issue (https://github.com/urllib3/urllib3/issues/1664), and Travis fails because you now need to run `nox -s blacken`. So this change looks good!

sethmlarson:
Thanks for this @tiran! I'm going to merge it since the falling AppVeyor isn't related.

test/test_ssl.py
```diff
@@ -149,3 +149,18 @@ def test_wrap_socket_default_loads_default_certs(monkeypatch):
     ssl_.ssl_wrap_socket(sock)
 
     context.load_default_certs.assert_called_with()
+
+
+@pytest.mark.parametrize(
+    ["pha", "expected_pha"], [(None, None), (False, True), (True, True)]
```
pquentin: @tiran We were wondering with @jab in https://github.com/python-trio/urllib3/pull/127: any reason why setting PHA to False sets it back to True? In other words, why `(False, True)` and not `(False, False)`?

PR #1637, opened 2019-06-20
Remove spurious TypeError from exception chain also for BaseException.

### LINKED ISSUE #1636: Spurious TypeError if BaseException occurs during getresponse()
This issue is a repeat of #861, but for exceptions that derive from `BaseException` and not `Exception`.

Specifically, if an `Exception` gets raised from `conn.getresponse()` in this line of code: https://github.com/urllib3/urllib3/blob/728d9244665ef5b03103cb74d7b409ebe4f23b43/src/urllib3/connectionpool.py#L411

then `urllib3` calls `raise from None` to remove the spurious `TypeError` from the exception chain: https://github.com/urllib3/urllib3/blob/728d9244665ef5b03103cb74d7b409ebe4f23b43/src/urllib3/connectionpool.py#L405-L415

However, if a `BaseException` like `KeyboardInterrupt` or `SystemExit` is raised during the call to `conn.getresponse()`, then the `TypeError` _isn't_ removed. This is because the try-except only catches and re-raises `Exception` and not `BaseException`. As mentioned in the original #861, seeing the spurious `TypeError` makes it look like there is bug e.g. in the error handling that occurs while the program exits. The `BaseException` can occur e.g. if a request is taking a long time and the application is terminated from another process.

The fix is to change `Exception` to `BaseException` in this line of code: https://github.com/urllib3/urllib3/blob/728d9244665ef5b03103cb74d7b409ebe4f23b43/src/urllib3/connectionpool.py#L412

I would file a PR, but I'm not sure I'd have time to craft a test. I note, though, that the original PR #861 didn't have a test.

cjerdonek:
Fixes #1636: https://github.com/urllib3/urllib3/issues/1636

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -278,5 +278,9 @@ In chronological order:
 * James Meickle <https://permadeath.com/>
   * Improve handling of Retry-After header
 
+* Chris Jerdonek <chris.jerdonek@gmail.com>
+  * Remove a spurious TypeError from the exception chain inside
+    HTTPConnectionPool._make_request(), also for BaseExceptions.
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -409,9 +409,10 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 # Python 3
                 try:
                     httplib_response = conn.getresponse()
-                except Exception as e:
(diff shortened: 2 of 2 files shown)


sethmlarson:
What exception in particular were you hitting this case with? `KeyboardInterrupt`?

cjerdonek:
In my case it was `SystemExit`. It was a long-running request, and the process was timed out / terminated by an outside process.

cjerdonek:
Thanks!

sethmlarson:
Thank you very much @cjerdonek! If you have more time there are plenty of other issues available to be taken up. :)

PR #1645, opened 2019-07-03
Update RECENT_DATE to 2019-1-1

sethmlarson:
First PR from the new laptop! :)

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -54,11 +54,9 @@ log = logging.getLogger(__name__)
 
 port_by_scheme = {"http": 80, "https": 443}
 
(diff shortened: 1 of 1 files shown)


theacodes:
Could we just... idk.. generate this with `utcnow()` and `timedelta`?

pquentin:
Sure, but we need to do it on a system where we know that the time is correct! So this could be done as part of the release process.

sethmlarson:
If this value was generated right before a release (for example on one of Travis's builders) you'd run into problems with users who bundle.  They wouldn't get a `RECENT_DATE`.

sethmlarson:
If we started using pre-commit we could create a custom one just for this value. :)

PR #1647, opened 2019-07-03
Percent-encode invalid characters within auth section

sethmlarson:
This issue has been cropping up all over both our and requests issue tracker. Mostly proxy users that have their email address as their "username" and therefore have an `@` symbol within the auth section. Within this PR I'm also going to be removing rfc3986 as a vendored dependency and instead use a simpler yet still strict and effective solution.


sethmlarson:
Going to merge this because codecov is giving me zero information about which lines are actually missing. Seems to be an issue that we don't have a coverage report for master.

PR #1649, opened 2019-07-05
Stop using unittest

pquentin:
This pull request builds on the work of @RatanShreshtha, specifically pull requests #1614 and #1624 that contain the bulk of the work.

This allows us to switch to the latest pytest version compatible with Python 2.7: https://docs.pytest.org/en/latest/py27-py34-deprecation.html. One benefit is that warnings are now grouped by source, which makes them easier to read (and hopefully to fix).

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,9 +5,8 @@ wheel==0.30.0
 tornado==5.1.1
 PySocks==1.6.8
 pkginfo==1.4.2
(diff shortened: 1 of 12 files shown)


sethmlarson:
Thanks @pquentin! :)

pquentin:
Thank you for the quick review! I'm glad that we got this mini project done.

PR #1663, opened 2019-08-13
Add host to insecure connection warning.

Steltek:
Sometimes, you're not the one who wrote the code, or urllib3 is being used within existing software, and it will give off the 'Unverified HTTPS request' warning, leaving the end user confused and unable to trace why they are getting it. (Example: Home Assistant)

This PR adds the hostname to the warning, making it more obvious to end users where the source of the request lies.


sethmlarson:
This is great but you're probably going to get a CI failure due to linting. Fix that issue then I think this is a great addition.

PR #1669, opened 2019-08-25
Fix for Python 4

hugovk:
We don't yet know if 3.10 or 4.0 will follow Python 3.9, but it will probably happen in [2020 (when Python 3.9 reaches beta)](https://www.python.org/dev/peps/pep-0596/#schedule).

There's several places in the codebase like:

```python
if six.PY3:
    pass  # Python 3+ stuff

if not six.PY3:
    pass  # Python 2 stuff
```

But when run on Python 4, it will skip the Python 3+ stuff and run the Python 2 stuff!

This PR flips it around:

```python
if not six.PY2:
    pass  # Python 3+ stuff

if six.PY2:
    pass  # Python 2 stuff
```

Found using:
```bash
pip install -U flake8-2020
flake8 --select YTT
```

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -99,7 +99,7 @@ class HTTPConnection(_HTTPConnection, object):
     is_verified = False
 
     def __init__(self, *args, **kw):
-        if six.PY3:
+        if not six.PY2:
             kw.pop("strict", None)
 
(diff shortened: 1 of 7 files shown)


pquentin:
Thank you @hugovk! You're an unsung hero.

PR #1671, opened 2019-08-28
Minimal CI fixes

### LINKED ISSUE #1664: Windows Python 3.7.4 tests fail in AppVeyor CI
Quoting myself from https://github.com/urllib3/urllib3/pull/1604#issuecomment-519009281:

> The Python 3.7 Windows failure is reproducible. I compared the [last good 3.7 build](https://ci.appveyor.com/project/urllib3/urllib3/builds/26274262/job/9678kar0ro9ohn1j) with the [first bad 3.7 build](https://ci.appveyor.com/project/urllib3/urllib3/builds/26524209/job/vshduc3gpd898dqp), and noticed that Python was updated from 3.7.3 to 3.7.4, and OpenSSL from 1.1.0j to 1.1.1c as part of https://bugs.python.org/issue34631.
> 
> Here's my understanding of the failures:
> 
>  *  TestConnectionPool.test_source_address_error appears to fail because [InvalidUrl is now being raised in the standard library](https://github.com/python/cpython/blob/3.7/Lib/http/client.py#L1114-L1118) and I guess we no longer catch the correct exception? But why only on Windows?
>  * TestHTTPS.test_client_no_intermediate and TestHTTPS_TLSv1_3.test_client_no_intermediate seem to fail because the OpenSSL exception is being converted into a SocketError in https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/pyopenssl.py#L345-L346, maybe because the execution path changed?
>  * TestHTTPS_TLSv1_3.test_set_ssl_version_to_tls_version fails because there are no certs defined in TestHTTPS_TLSv1_3, so self.certs["ssl_version"] can't work

pquentin:
This does not fix flaky tests but should help with the tests that are consistently failing right now.

Closes #1664

### DIFF
--- a/test/contrib/test_securetransport.py
+++ b/test/contrib/test_securetransport.py
@@ -29,12 +29,15 @@ def teardown_module():
         pass
 
 
+# While SecureTransport supports TLS 1.3, we can't test for it because
+# Security.SSLSetProtocolVersionMin and Security.SSLSetProtocolVersionMax don't accept
+# kTLSProtocol13
+# See https://forums.developer.apple.com/thread/110369 which includes the relevant code
 from ..with_dummyserver.test_https import (  # noqa: F401
     TestHTTPS,
     TestHTTPS_TLSv1,
     TestHTTPS_TLSv1_1,
     TestHTTPS_TLSv1_2,
-    TestHTTPS_TLSv1_3,
 )
 from ..with_dummyserver.test_socketlevel import (  # noqa: F401
     TestSNI,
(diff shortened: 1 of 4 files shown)


test/contrib/test_securetransport.py
```diff
@@ -29,12 +29,15 @@ def teardown_module():
         pass
 
 
+# While SecureTransport supports TLS 1.3, we can't test for it because
```
pquentin: I now know that this comment is wrong: SecureTransport does not support TLS 1.3 at all

sethmlarson: I'll give it an update and merge it once CI passes on all code changes. :) Thanks @pquentin!

PR #1672, opened 2019-08-28
pass assert_hostname to socket if specified for verified https connections

eljohnson92:
Currently, in VerifiedHTTPSConnection, if a user specifies a value for assert_hostname, that value is used when validating the server cert, but not for SNI. If no value is specified for server_hostname but one is specified for assert_hostname, urllib3 should prefer that value to host for SNI, as that's almost certainly what the caller intended (i.e., if the caller expects a hostname to be present on the certificate, then presumably they'd want to tell the server to send a certificate with that hostname). The proposed patch implements that behavior.


src/urllib3/connection.py
```diff
@@ -312,9 +312,7 @@ def connect(self):
             # Override the host with the one we're requesting data from.
             hostname = self._tunnel_host
 
-        server_hostname = hostname
-        if self.server_hostname is not None:
-            server_hostname = self.server_hostname
+        server_hostname = self.server_hostname or self.assert_hostname or hostname
```
pquentin: Is there any reason for putting `self.assert_hostname` before `hostname`?

eljohnson92: we are putting `self.assert_hostname` over `hostname` so if a user wants to assert a different hostname is present on the server's certificate then the hostname they used to resolve DNS, then urllib3 will ask for a certificate that's valid for that custom hostname

eljohnson92: @pquentin have you had a chance to look at this any more?

pquentin: Thanks for pinging me!

The thing here is that `assert_hostname` is intended to be a boolean for *disabling* hostname verification (see https://github.com/urllib3/urllib3/pull/194/files). It turns out that due to how Python works and how the code evolved you can assign it to an actual hostname, but that's not the intended or documented behavior and you should use server_hostname instead, which the standard name for this: https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.server_hostname

Thank you for the pull request and for your patience, but I'm going to close this one. Would you like to open another pull request to improve our reference documentation here?

pquentin:
Closing per https://github.com/urllib3/urllib3/pull/1672#discussion_r369388989

PR #1673, opened 2019-08-29
Percent-encode targets and URLs in ConnectionPool

sethmlarson:
This will fix our issues with the source address error tests and Python 3.7.4 exploding when we include ` `, `\r`, or `\n` in the request target.

cc @pquentin


pquentin:
Thanks! Do you know why the new test is failing?

PR #1676, opened 2019-09-01
Add force_retry_callback to enable custom retry handling

### LINKED ISSUE #1675: Custom retry handling support
Hi, thanks for implementing and maintaining **urllib3**. 

Retry mechanism based on _status_forcelist_ and _method_whitelist_ perfectly works. But in some rare cases, when APIs are not well-constructed, it is required to add custom retry handling. 

* Consider following response format:

`
"status_code": 400,
"body": {
  "error": " Service temporarily unavailable."
}
`

Generally, you don't want to retry on bad requests(400). An ideal response code for this request might be 503. But since it is not restricted, it might be some other HTTP formats. So my proposal is to add a generic solution for retry mechanism which enables  **urllib3** users to integrate their custom logic.

I found it useful to integrate callback function  into [Retry](https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/retry.py#L29), which injects response as a parameter to callback function when [urlopen()](https://github.com/urllib3/urllib3/blob/master/src/urllib3/connectionpool.py#L488) is triggered on [HTTPConnectionPool](https://github.com/urllib3/urllib3/blob/master/src/urllib3/connectionpool.py#L97). Thus, retry logic can be implemented by parsing this response. 

* Example callback function: 
```python
    def callback(res):
            expected_error_msg = "Service temporarily unavailable"
            response_json = res.json()
            error_msg = response_json["error"]
            return res.status_code == 400 and error_msg == expected_error_msg
```

sonmezonur:
Resolves #1675

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -803,7 +803,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         # Check if we should retry the HTTP response.
         has_retry_after = bool(response.getheader("Retry-After"))
-        if retries.is_retry(method, response.status, has_retry_after):
+        if retries.is_retry(method, response.status, response, has_retry_after):
             try:
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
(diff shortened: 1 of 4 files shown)


sethmlarson:
Unfortunately this would be a breaking change for users that implement the old Retry interface. urllib3 isn't accepting breaking changes currently, sorry!

PR #1679, opened 2019-09-18
Add support for HTTPS connections to proxies.

jalopezsilva:
Currently there's no way to validate the identify of the proxy you are connecting to.  Proxies supporting HTTPS endpoints are becoming more common and extending support for them is necessary.

This commit adds the ability to establish an HTTPS connection to a proxy. 

When an HTTPS proxy is provided, instead of doing an HTTP CONNECT, we'll forward any requests directly to the proxy and ultimately to the destination, similar to how HTTP connections through a proxy are handled.

Another implementation alternative would have been to do TLS to the proxy, later an HTTP CONNECT and within the tunnel, another TLS connection to the destination. This is more complex and requires us to use the ``SSLContext.wrap_bio`` methods (i.e TLS within TLS).  I opted for this simpler approach as a start but I'm happy to extend if needed.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -112,6 +112,9 @@ class HTTPConnection(_HTTPConnection, object):
         #: provided, we use the default options.
         self.socket_options = kw.pop("socket_options", self.default_socket_options)
 
+        # Protocol used to talk to the proxy.
+        self.proxy_scheme = kw.pop("proxy_scheme", None)
+
         _HTTPConnection.__init__(self, *args, **kw)
 
     @property
@@ -174,10 +177,13 @@ class HTTPConnection(_HTTPConnection, object):
 
         return conn
 
+    def _is_using_tunnel(self):
+        # Google App Engine's httplib does not define _tunnel_host
+        return getattr(self, "_tunnel_host", None)
+
     def _prepare_conn(self, conn):
         self.sock = conn
-        # Google App Engine's httplib does not define _tunnel_host
-        if getattr(self, "_tunnel_host", None):
+        if self._is_using_tunnel():
             # TODO: Fix tunnel so it doesn't depend on self.sock state.
             self._tunnel()
             # Mark this connection as not reusable
@@ -314,9 +320,9 @@ class VerifiedHTTPSConnection(HTTPSConnection):
         conn = self._new_conn()
         hostname = self.host
 
-        # Google App Engine's httplib does not define _tunnel_host
-        if getattr(self, "_tunnel_host", None):
+        if self._is_using_tunnel():
             self.sock = conn
+
             # Calls self._set_hostport(), so self.host is
             # self._tunnel_host below.
             self._tunnel()
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -111,10 +111,6 @@ class HTTPConnection(_HTTPConnection, object):
         #: The socket options provided by the user. If no options are
         #: provided, we use the default options.
         self.socket_options = kw.pop("socket_options", self.default_socket_options)
-
-        # Protocol used to talk to the proxy.
-        self.proxy_scheme = kw.pop("proxy_scheme", None)
-
         _HTTPConnection.__init__(self, *args, **kw)
 
     @property
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -632,7 +632,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # Merge the proxy headers. Only done when not using HTTP CONNECT. We
         # have to copy the headers dict so we can safely change it without those
         # changes being reflected in anyone else's copy.
-        if self.scheme == "http" or self.proxy and self.proxy.scheme == "https":
+        if self.scheme == "http" or (self.proxy and self.proxy.scheme == "https"):
             headers = headers.copy()
             headers.update(self.proxy_headers)
 
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -212,8 +212,6 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # We cannot know if the user has added default socket options, so we cannot replace the
             # list.
             self.conn_kw.setdefault("socket_options", [])
-            # Capture the proxy scheme to properly establish the connection.
-            self.conn_kw["proxy_scheme"] = self.proxy.scheme
 
     def _new_conn(self):
         """
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -212,6 +212,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # We cannot know if the user has added default socket options, so we cannot replace the
             # list.
             self.conn_kw.setdefault("socket_options", [])
+            # Capture the proxy scheme to properly establish the connection.
+            self.conn_kw["proxy_scheme"] = self.proxy.scheme
 
     def _new_conn(self):
         """
@@ -941,10 +943,15 @@ class HTTPSConnectionPool(HTTPConnectionPool):
 
     def _prepare_proxy(self, conn):
         """
-        Establish tunnel connection early, because otherwise httplib
-        would improperly set Host: header to proxy's IP:port.
+        Establishes a tunnel connection through HTTP CONNECT.
+
+        Tunnel connection is established early because otherwise httplib would
+        improperly set Host: header to proxy's IP:port.
         """
-        conn.set_tunnel(self._proxy_host, self.port, self.proxy_headers)
+
+        if self.proxy.scheme != "https":
+            conn.set_tunnel(self._proxy_host, self.port, self.proxy_headers)
+
         conn.connect()
 
     def _new_conn(self):
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -631,10 +631,10 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # [1] <https://github.com/urllib3/urllib3/issues/651>
         release_this_conn = release_conn
 
-        # Merge the proxy headers. Only do this in HTTP. We have to copy the
-        # headers dict so we can safely change it without those changes being
-        # reflected in anyone else's copy.
-        if self.scheme == "http":
+        # Merge the proxy headers. Only done when not using HTTP CONNECT. We
+        # have to copy the headers dict so we can safely change it without those
+        # changes being reflected in anyone else's copy.
+        if self.scheme == "http" or self.proxy and self.proxy.scheme == 'https':
             headers = headers.copy()
             headers.update(self.proxy_headers)
 
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -634,7 +634,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # Merge the proxy headers. Only done when not using HTTP CONNECT. We
         # have to copy the headers dict so we can safely change it without those
         # changes being reflected in anyone else's copy.
-        if self.scheme == "http" or self.proxy and self.proxy.scheme == 'https':
+        if self.scheme == "http" or self.proxy and self.proxy.scheme == "https":
             headers = headers.copy()
             headers.update(self.proxy_headers)
 
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -6,7 +6,12 @@ import logging
 from ._collections import RecentlyUsedContainer
 from .connectionpool import HTTPConnectionPool, HTTPSConnectionPool
 from .connectionpool import port_by_scheme
-from .exceptions import LocationValueError, MaxRetryError, ProxySchemeUnknown
+from .exceptions import (
+    LocationValueError,
+    MaxRetryError,
+    ProxySchemeUnknown,
+    ProxySchemeUnsupported,
+)
 from .packages import six
 from .packages.six.moves.urllib.parse import urljoin
 from .request import RequestMethods
@@ -395,6 +400,11 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
+    :param allow_https_proxy_to_see_traffic:
+        Allows forwarding of HTTPS requests to HTTPS proxies. The proxy will
+        have visibility of all the traffic sent. ONLY USE IF YOU KNOW WHAT
+        YOU'RE DOING.
+
     Example:
         >>> proxy = urllib3.ProxyManager('http://localhost:3128/')
         >>> r1 = proxy.request('GET', 'http://google.com/')
@@ -414,6 +424,7 @@ class ProxyManager(PoolManager):
         num_pools=10,
         headers=None,
         proxy_headers=None,
+        allow_https_proxy_to_see_traffic=False,
         **connection_pool_kw
     ):
 
@@ -436,6 +447,8 @@ class ProxyManager(PoolManager):
 
         connection_pool_kw["_proxy"] = self.proxy
         connection_pool_kw["_proxy_headers"] = self.proxy_headers
+
+        self.allow_insecure_proxy = allow_https_proxy_to_see_traffic
 
         super(ProxyManager, self).__init__(num_pools, headers, **connection_pool_kw)
 
@@ -464,9 +477,18 @@ class ProxyManager(PoolManager):
             headers_.update(headers)
         return headers_
 
+    def _validate_proxy_scheme_url_selection(self, url_scheme):
+        if (
+            url_scheme == "https"
+            and self.proxy.scheme == "https"
+            and not self.allow_insecure_proxy
+        ):
+            raise ProxySchemeUnsupported("TLS within TLS is not yet supported.")
+
     def urlopen(self, method, url, redirect=True, **kw):
         "Same as HTTP(S)ConnectionPool.urlopen, ``url`` must be absolute."
         u = parse_url(url)
+        self._validate_proxy_scheme_url_selection(u.scheme)
 
         if u.scheme == "http" or self.proxy.scheme == "https":
             # For connections using HTTP CONNECT, httplib sets the necessary
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -319,7 +319,7 @@ class PoolManager(RequestMethods):
                     base_pool_kwargs[key] = value
         return base_pool_kwargs
 
-    def _proxy_requires_complete_url(self, parsed_url):
+    def _proxy_requires_url_absolute_form(self, parsed_url):
         """
         Indicates if the proxy requires the complete destination URL in the
         request.
@@ -349,7 +349,7 @@ class PoolManager(RequestMethods):
         if "headers" not in kw:
             kw["headers"] = self.headers.copy()
 
-        if self._proxy_requires_complete_url(u):
+        if self._proxy_requires_url_absolute_form(u):
             response = conn.urlopen(method, url, **kw)
         else:
             response = conn.urlopen(method, u.request_uri, **kw)
@@ -408,11 +408,6 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
-    :param _enable_https_proxies:
-        Setting this to true will enable contacting HTTPS proxies if one is
-        provided through the configuration. Defaults to False, will be switched
-        to True in a future release.
-
     :param _allow_https_proxy_to_see_traffic:
         Allows forwarding of HTTPS requests to HTTPS proxies. The proxy will
         have visibility of all the traffic sent. ONLY USE IF YOU KNOW WHAT
@@ -438,7 +433,6 @@ class ProxyManager(PoolManager):
         num_pools=10,
         headers=None,
         proxy_headers=None,
-        _enable_https_proxies=False,
         _allow_https_proxy_to_see_traffic=False,
         **connection_pool_kw
     ):
@@ -453,16 +447,6 @@ class ProxyManager(PoolManager):
 
         if proxy.scheme not in ("http", "https"):
             raise ProxySchemeUnknown(proxy.scheme)
-
-        if proxy.scheme == "https" and not _enable_https_proxies:
-            proxy = proxy._replace(scheme="http")
-            warnings.warn(
-                "Your proxy configuration specified an HTTPS scheme for the proxy. "
-                "urllib3 previoustly didn't support contacting a proxy through HTTPS."
-                "If you want to use HTTPS to contact the proxy, provide the "
-                "_enable_https_proxies flag. Otherwise set the proxy scheme to HTTP.",
-                InvalidProxyConfigurationWarning,
-            )
 
         if not proxy.port:
             port = port_by_scheme.get(proxy.scheme, 80)
@@ -509,7 +493,18 @@ class ProxyManager(PoolManager):
             and self.proxy.scheme == "https"
             and not self.allow_insecure_proxy
         ):
-            raise ProxySchemeUnsupported("TLS within TLS is not yet supported.")
+            warnings.warn(
+                "Your proxy configuration specified an HTTPS scheme for the proxy. "
+                "Are you sure you want to use HTTPS to contact the proxy? "
+                "This most likely indicates an error in your configuration."
+                "If you are sure you want use HTTPS to contact the proxy, enable "
+                "the _allow_https_proxy_to_see_traffic.",
+                InvalidProxyConfigurationWarning,
+            )
+
+            raise ProxySchemeUnsupported(
+                "Contacting HTTPS destinations through HTTPS proxies is not supported."
+            )
 
     def urlopen(self, method, url, redirect=True, **kw):
         "Same as HTTP(S)ConnectionPool.urlopen, ``url`` must be absolute."
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -2,11 +2,13 @@ from __future__ import absolute_import
 import collections
 import functools
 import logging
+import warnings
 
 from ._collections import RecentlyUsedContainer
 from .connectionpool import HTTPConnectionPool, HTTPSConnectionPool
 from .connectionpool import port_by_scheme
 from .exceptions import (
+    HTTPWarning,
     LocationValueError,
     MaxRetryError,
     ProxySchemeUnknown,
@@ -20,6 +22,12 @@ from .util.retry import Retry
 
 
 __all__ = ["PoolManager", "ProxyManager", "proxy_from_url"]
+
+
+class InvalidProxyConfigurationWarning(HTTPWarning):
+    """Raised when a user has an HTTPS proxy without enabling HTTPS proxies."""
+
+    pass
 
 
 log = logging.getLogger(__name__)
@@ -400,6 +408,11 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
+    :param _enable_https_proxies:
+        Setting this to true will enable contacting HTTPS proxies if one is
+        provided through the configuration. Defaults to False, will be switched
+        to True in a future release.
+
     :param _allow_https_proxy_to_see_traffic:
         Allows forwarding of HTTPS requests to HTTPS proxies. The proxy will
         have visibility of all the traffic sent. ONLY USE IF YOU KNOW WHAT
@@ -425,6 +438,7 @@ class ProxyManager(PoolManager):
         num_pools=10,
         headers=None,
         proxy_headers=None,
+        _enable_https_proxies=False,
         _allow_https_proxy_to_see_traffic=False,
         **connection_pool_kw
     ):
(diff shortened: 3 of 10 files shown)


jalopezsilva:
Any feedback? Happy to adjust as needed. 

Facebook needs to support HTTPS proxies internally and we would love to get this patch through!

gregprosser:
What's blocking this getting merged into the codebase?  I have several downstream projects that are using urllib3 that would benefit from having this support?

YogeshMhatre:
+1 - I have a few critical projects waiting on this feature. It would be *incredibly* useful if this is merged

src/urllib3/connectionpool.py
```diff
@@ -938,10 +940,15 @@ def _prepare_conn(self, conn):
...
-        would improperly set Host: header to proxy's IP:port.
+        Establishes a tunnel connection through HTTP CONNECT.
+
+        Tunnel connection is established early because otherwise httplib would
+        improperly set Host: header to proxy's IP:port.
         """
-        conn.set_tunnel(self._proxy_host, self.port, self.proxy_headers)
+
+        if self.proxy.scheme != "https":
+            conn.set_tunnel(self._proxy_host, self.port, self.proxy_headers)
```
aausch: You're losing the `proxy_headers` on the `conn.connect()`

jalopezsilva: Thanks @aausch, great catch.

I've added another commit in which we are now adding the proxy headers when using an HTTPS proxy. Instead of sending them with the HTTP CONNECT, we'll be including them in the requests forwarded to the proxy similar to what's done for HTTP requests.

aausch: @jalopezsilva merging the headers is incorrect behavior. Unless I'm missing changes elsewhere, it looks to me like there are multiple problems with the commit as is:

1. No tunnel is actually being established. As far as I can tell, the `CONNECT` message is never being sent. When the `CONNECT` message does go out, it's expected to contain the authentication portion, at the very least, for the proxy headers. Though it probably will need all the headers since that's the last message that will be sent to the proxy without the inner connection's encryption layer?
2. The headers should not be merged. Proxy headers only should be sent to the proxy, on the outer-tunnel connection, and not merged then sent on a mixed connection or inner connection (depending on how you think of the two SSL encryption layers)
3. The inner connection is not getting wrapped in an SSL wrapper at any point; this will prevent any attempts to connect to HTTPS sites through the proxy. Also, this is a non trivial exercise, since urllib, and the underlying python C wrappers around the OpenSSL, don't provide native wrappers. And the OpenSSL wrapper that's passed all the way up doesn't support wrapping encrypted connections/sockets - it only supports OS native sockets. See for example the conversation here: https://bugs.python.org/issue29394

aausch: see https://github.com/python/cpython/blob/3.8/Lib/http/client.py#L883 ->

1. I'm under the impression httplib doesn't appear to properly support encrypted tunnels as of Python 3.8 (and still only connects to proxies via `HTTP 1.0` instead of 1.1!)
2. I'm fairly confident (unless I'm missing something in one of the libraries?) that without an explicit call to `_tunnel`, you won't end up establishing a proxy connection

aausch: Again, maybe I'm missing something and this code just works? But what I expected to see isn't this CR

---

For `http` proxies, the existing code works - it can just reuse the connection into the proxy:
1. get a new connection, to the proxy
2. send the transparent connect/ tunnelling message request
3. wrap the proxy connection in SSL encryption aimed at the `https` site we want to reach (which bottoms out in calls to OpenSSL or w/e on the underlying system, and properly upgrades the connection/socket/file descriptor)
4. shoot encrypted messages to the proxy and they'll get bounced through

---

This won't work for https proxies; what has to happen:
1. get a new (http) connection to the proxy
2. escalate to/ wrap w the proxy's `https` certificates
3. sent a `CONNECT` message/open a tunnel
4. create an in-memory equivalent of a buffered connection/socket/fd, plugging the output into the proxy's https encrypted connection/socket/fd
5. wrap the in-memory buffered connection/socket/fd with the end sites' `https` certificate
6. use the in memory buffered connection/socket/fd for any further communication

---

Currently, I don't think step 3 is happening anywhere. I would expect that steps 4-6 should really be happening somewhere inside of the OpenSSL library, or, alternatively, somewhere in the CPython implementation, for reasons of CPU performance and to avoid memory pressure issues, though the https://bugs.python.org/issue29394 thread seems to suggests a Python level alternative is available? At any rate, I don't see those steps addressed in this CR, but maybe I'm missing a change in a library upgrade or elsewhere in middleware

jalopezsilva: @aausch,  Thanks for the review. Your understanding of the PR is correct, I am intentionally not establishing the tunnel through HTTP CONNECT. I am instead establishing a TLS connection to the proxy and forwarding requests to the proxy within that connection. The proxy should establish the TLS connection with the destination and forward the requests upstream. I tried to be clear that was my intent in my first comment of the PR (apologies if it wasn't clear enough!)

The implementation that I chose avoids the client having to do TLS within TLS. The connection to the destination is secure: there's a TLS connection between client<->proxy and proxy<->server. The only disadvantage is that the client won't have a direct TLS session with the destination. (Which prevents mutual authentication with the destination) There's also an implicit trust set on the proxy by the client. Given that we're explicitly trusting the proxy by establishing a TLS session to it I thought this was agreeable. Support in forward proxies for this feature is relatively common (although we can do a quick audit and verify).

The other implementation alternative would be to perform TLS to the proxy, HTTP CONNECT to the destination and TLS to the destination within the tunnel. This lines up with what you outlined in the comment above. As you said, implementing this is not trivial. It is indeed possible within Python if you use the ``SSLContext.wrap_bio`` methods. The tricky part of implementing it through the ``wrap_bio`` methods is that you have to manage the IO within the socket and the memoryBIO yourself. You can see an example of how this looks like on the following ``CPython`` test case: https://github.com/python/cpython/blob/master/Lib/test/test_ssl.py#L2150-L2235.  I have a tentative implementation of an ``SSLTransport`` object which takes a socket, wraps it and performs the IO within itself using the ``wrap_bio`` APIs. It would work like an ``SSLSocket`` except it wouldn't take ownership of the underlying native socket. 

I think we have three options for moving forward:

1. **Leave this PR as is and merge it** - The PR will add initial support for HTTPS proxies. We can add documentation indicating what the behavior is and the limitations.

2. **Add TLS within TLS support to urllib3** - I can add the ``SSLTransport`` object I mentioned before to urllib3/contrib/ssl.py and only use it when we need to do TLS within TLS. It will probably take us a couple of iterations to get it 100% right but we can give it a shot. 

3. **Add TLS within TLS within CPython** - Similar to #2 but I try to upstream the ``SSLTransport`` object to CPython. This will probably take the longest time as we would have to wait for Python 3.9 or even beyond.

I'm happy to move forward with any of the options. Out of the three, #3 is probably my least favorite given that it would take a long time to get this feature out. Ultimately I think we need an easy way to do TLS within TLS in Python. If we go ahead with #2, I could try doing #3 once we have a stable implementation within urllib3.

aausch: >The implementation that I chose avoids the client having to do TLS within TLS. The connection to the destination is secure: there's a TLS connection between client<->proxy and proxy<->server. 

yes, I saw that this is your intent. But the code as is does not achieve this, as far as I can tell.  'tunneling' means 
1. 'tell the proxy to create the second connection' ie. what you call `proxy <-> server` in your comment, and also
2. tell the proxy to blindly forward messages along it to a specific server. 

the specific protocol for it, is to send a special version of the http connect message - something like  `CONNECT <ultimate destination server>` - to the proxy. Without that message, the proxy doesn't know where to forward messages to/what connection to open on our behalf.  as far as I can tell, this pull request doesn't, anywhere, tell the proxy to open the second connection and to which server to open it to.

>  The only disadvantage is that the client won't have a direct TLS session with the destination.

That is a significant disadvantage. Would you be ok if your proxy provider can read, for example, your authentication information for logging in to your bank? or your login information for Netflix?  Your keys for making a bitcoin exchange? Assuming you can deliver on this, you're guaranteeing a MITM will read all of that information, and do whatever they want with it. 

I'm not sure if there is a legitimate use case you're looking to address here (maybe developers who are using a mock proxy during code testing?), but, if and assuming there is one, I would:

1. guard this specific use case a flag (you'd want the end user to explicitly set a flag "allow proxy to read all my information" to true or otherwise indicate they're interested in trusting the proxy with their info, so maybe add an argument into the function and check it)
2. spit out a large Warning message in bright, bold, red text, across every possible readable output anytime anything is sent across this connection

>  Given that we're explicitly trusting the proxy by establishing a TLS session to it I thought this was agreeable.

This is a **significant** error or misinterpretation of what TLS is, and what it does. TLS **does not imply any kind of trust**; it just protects you from having your ISP, or a government, or law enforcement, or your friendly hacker on a Starbucks wifi connection, know that you're using a proxy. 

Also, per my comment above and more generally, even if there was a method for trusting proxies in general, I would highly recommend against trusting proxies with most of the things that you would use a proxy for.

> Support in forward proxies for this feature is relatively common (although we can do a quick audit and verify).

If by "this feature" you mean anything other than tunnelling, then I'm really not sure which feature you're talking about - please provide a reference or more description of it.

---

> 1. **Leave this PR as is and merge it** 

As far as I can tell, this PR as is does not actually provide any additional value, and should not yet be merged. It does allow opening of HTTPS connections to proxies, which is an improvement, but then does not go on to create a tunnel or, through any other protocol, tell the proxy it should pass messages on to a specific target server - so no messages are then sent past the proxy to the ultimate target server. Also, it introduces two significant bugs, caused by the merging of the headers:
- as written, all the headers are only being sent into the proxy on the initial connect into it, and won't be passed through to the ultimate target server; also, i believe the headers are no longer correctly supporting authentication into the proxy, so if you're dealing with a proxy that expects you to log in to it, the connect will fail
- i'm not sure if the proxy will know how to handle the merged headers at all. think about it this way: what if the second server you're connecting to, is also a proxy, which is expecting certain proxy headers? how will the first proxy know which proxy headers it should interpret, and which proxy headers it should send along to the second server?

The other two options are, I think, the only options which will allow proxying into an endserver via the HTTPS protocol, and one of them should eventually be implemented (also assuming the bugs listed above are fixed). I also think there's a possible intermediate option:

1.5: **Fix this PR so that it at least works correctly for the  `TLS <> proxy`, `proxy <> HTTP site` use case, then merge it**

I believe this PR is pretty close, as is, to supporting forwards to HTTP sites only, over a TLS proxy connection. To get there, just make sure that after the connection is tls wrapped, you send the correct command to establish a tunnel to the end server(including the proxy headers and correctly assembled proxy authentication information). 

As a bonus, it would be also awesome to switch to the `HTTP 1.1 ` version of the tunnelling request, since that allows for persistent connections through the proxy and I suspect would be much more efficient than reopening the connection every time

aausch: Oh, and, if going with option `1.5`, probably check if the end site protocol is tls via a tls proxy, and if it is, spit out a warning or error warning the user that it's not yet supported, but it will be supported soon

jalopezsilva: > As far as I can tell, this PR as is does not actually provide any additional value, and should not yet be merged. It does allow opening of HTTPS connections to proxies, which is an improvement, but then does not go on to create a tunnel or, through any other protocol, tell the proxy it should pass messages on to a specific target server - so no messages are then sent past the proxy to the ultimate target server.

@aausch, I'm a little surprised by this comment and I want to make sure we are in the same page. 

When proxies are presented with a absolute URI in a request, ``GET http://example.com``, the proxy will forward that request for you. This is the expected behavior when providing an absolute URI to the proxy. (Documented [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html)).  The proxy will establish a TCP connection and forward the request to the destination.

urllib3 uses this when requesting HTTP resources through an HTTP proxy. Notice how in ``HTTPConnectionPool``, ``_prepare_proxy`` doesn't establish any tunnel. The request is simply forwarded with the absolute URI. (See lines 327-330, ``poolmanager.py``). There's no CONNECT request made by urllib3 to the proxy. When fetching HTTPS resources through an HTTP proxy we do send the CONNECT as we need to establish the TLS session to the destination. (``HTTPSConnectonPool``, ``_prepare_proxy`` does this). Here's an excerpt of a tcpdump verifying that: https://pastebin.com/ZNyBWH5B 

This PR re-uses the forwarding functionality to fetch resources from an HTTPS proxy. I agree that fetching HTTPS resources from an HTTPS proxy is insecure and I'll be adding CONNECT support for that, but fetching HTTP resources through an HTTPS proxy works as intended in this PR. See the integration tests that I added with (test/with_dummyserver/test_proxy_poolmanager.py) which verify that.  I have also added this local patch to integration tests in my company as well. 

> * as written, all the headers are only being sent into the proxy on the initial connect into it, and won't be passed through to the ultimate target server; also, i believe the headers are no longer correctly supporting authentication into the proxy, so if you're dealing with a proxy that expects you to log in to it, the connect will fail
> * i'm not sure if the proxy will know how to handle the merged headers at all. think about it this way: what if the second server you're connecting to, is also a proxy, which is expecting certain proxy headers? how will the first proxy know which proxy headers it should interpret, and which proxy headers it should send along to the second server?

The header behavior is the same that urllib3 uses when fetching HTTP resources with an HTTP proxy. In that case headers are also merged. (See src/urllib3/connectionpool.py, line 637). This works relatively well. The proxy will remove [per-hop headers](https://tools.ietf.org/html/rfc2616#section-13.5.1) to prevent the server from seeing headers it shouldn't see and forward the request.

The case which isn't supported is a second proxy which might expect a second set of headers. This currently doesn't work on the HTTP proxy / HTTP resource case. I think we can continue merging headers on the HTTPS proxy / HTTP resource case as well.


## Moving forward
Let me know if this makes sense. 

I'll throw an exception when users attempt to fetch HTTPS resources through an HTTPS proxy as that's insecure as we agreed until I add the TLS within TLS support.. I can also add a parameter ``allow_insecure_proxy_to_see_my_traffic`` if we want to keep that behavior configurable by the client.

aausch: > When proxies are presented with an absolute URI in a request, GET http://example.com, the proxy will forward that request for you. This is the expected behavior when providing an absolute URI to the proxy. (Documented here). The proxy will establish a TCP connection and forward the request to the destination.

Yeah, you're right! I forgot about this usage pattern - I typically think of it as more of a bridge, than a proxy, and I forget that historically many proxies were nothing but bridges. Without a tunnel, the client has zero control over what happens (and fully exposes the communication to the proxy),  and most of the internet has moved past the point where bridges work correctly, at all, for most of what everyone does online.

>  I don't see the value of doing an HTTP CONNECT to fetch a HTTP resource

Without a tunnel, you only have what I think of as a bridged connection past the proxy. In general, that severely limits what you can do when communicating with the downstream server (there are lots of ways to hack more value out of it, but they're one offs and hacky) - eg., you can't hop over multiple proxies, open sockets/websockets, perform correct https/tls, play most modern video games, and most other web communication doesn't work correctly, etc... - i.e., you're supporting at most 20% of the kind of traffic we see on the internet, but probably closer to 10%. 

[aside: I am surprised, honestly, to be reminded that this kind of traffic is something people still use for anything, beyond highly specialized use cases (eg., when behind a corporate wall). It's also runs the risk of being horrendously inefficient! A tunnel, esp. over HTTP 1.1, has stronger guarantees to be a persistent, reusable, connection, which significantly speeds everything along.]

---

> This PR re-uses the forwarding functionality to fetch resources from an HTTPS proxy. I agree that fetching HTTPS resources from an HTTPS proxy is insecure and I'll be adding CONNECT support for that, but fetching HTTP resources through an HTTPS proxy works as intended in this PR. 

I agree that this is definitely something that needs to be supported and clearly useful/of value. I am surprised to find out that urllib3 does not yet support using tunnelling for anything but HTTPs traffic - but that is out of scope here and probably that should be a separate feature request (eg. dynamically figure out if tunnelling is available, default to a tunnel, and only if failing that). 

---

> Let me know if this makes sense. I don't see the value of doing an HTTP CONNECT to fetch a HTTP resource, unless you're concerned that the proxy defined as an "https proxy" in the client configuration proxy only supports CONNECT requests? Modern proxy servers support both though, CONNECT and forwarding requests.

I agree with almost everything. Definitely, HTTP CONNECT should be the default for all web communication, and basic forwarding should be just a fall back, but that work is out of scope I think for this PR and it makes sense for me to be a part of a different pull request.

> I'll throw an exception when users attempt to fetch HTTPS resources through an HTTPS proxy as that's insecure as we agreed until I add the TLS within TLS support.. I can also add a parameter allow_insecure_proxy_to_see_my_traffic (default False of course) if we want to keep that behavior configurable by the client.

If it's guarded by a False default flag, it's ok with me. Looking at the code, it would even make sense to me to just not support TLS, without tunnelling in place. It's a very niche, insecure, corner case, and wasn't supported historically by urllib3, probably due to the security concerns? I now wonder if there are (non-shady) proxies which support this, even.

pquentin:
@jalopezsilva @aausch Thank you for the discussion above, this was very useful to me as I'm not familiar with the way proxies work. I'm glad that you two now agree.

@jalopezsilva I'm now trying to understand the bigger picture here. I've spend most of my day trying to understand what's going on in. As far as I can tell, the way urllib3 currently supports proxies is:

 1. for HTTP proxies and HTTP resources, we send requests with the absolute URL cf. [RFC 7230, section 5.4](https://tools.ietf.org/html/rfc7230#section-5.4)
 2. for HTTP/HTTPS proxies and HTTPS resources, we send a HTTP CONNECT in plaintext, upgrade the connection to the original server to TLS, and send requests as usual

(You've said above that you would like to add CONNECT support for fetching HTTPS resources from an HTTPS proxy, but I think that already exists.)

Anyway, now we're adding a third mode: for HTTPS proxies we connect to the proxy using TLS from the start, then we request the absolute URL just like in the HTTP proxy mode above.

Compared to the existing HTTP CONNECT support, we never talk in plain HTTP to the proxy, so an attacker listening our connection to the proxy does not know what origin server we are requesting. However, compared to HTTP CONNECT we need to trust the proxy, because it's now able to see all our communication to the origin server, which was not possible before.

The best of both worlds would be obtained with TLS-in-TLS, but that's harder to implement so we're leaving this for future work.

Is this correct?

@jalopezsilva @gregprosser @YogeshMhatre Can you share more about your use case? Is this about fetching HTTP resources from HTTPS proxies? Understanding what kind of resources you're accessing through what kind of proxy would be helpful to understand where you're coming from.

src/urllib3/connection.py
```diff
@@ -109,6 +109,9 @@ def __init__(self, *args, **kw):
         #: provided, we use the default options.
         self.socket_options = kw.pop("socket_options", self.default_socket_options)
 
+        # Protocol used to talk to the proxy.
+        self.proxy_scheme = kw.pop("proxy_scheme", None)
```
pquentin: Where is `self.proxy_scheme` used?

aausch:
> (You've said above that you would like to add CONNECT support for fetching HTTPS resources from an HTTPS proxy, but I think that already exists.)

Pretty sure this doesn't currently work correctly

pquentin:
> > (You've said above that you would like to add CONNECT support for fetching HTTPS resources from an HTTPS proxy, but I think that already exists.)
> 
> Pretty sure this doesn't currently work correctly

Yeah, I think urllib3 just ignores the fact that the proxy supports HTTPS and just treats is as a HTTP proxy. But it still "works". Doing things correctly would require TLS in TLS.

src/urllib3/connectionpool.py
jalopezsilva: In total agreement @aausch, moving to HTTP CONNECT for everything instead of "bridging" is much more flexible but you're totally right that we probably should tackle this as a separate PR. Some people might be relying on the HTTP resource / HTTP proxy "bridging" behavior. 

I am now checking it we are attempting to fetch HTTPS resources through an HTTPS proxy and throwing an exception as this is not supported and insecure. I added a flag ``allow_insecure_proxy_to_see_my_traffic``, defaulted to False, and can be changed if a client wants the behavior.

Once we close this PR, I'll follow up with the TLS in TLS support to that we can close it out.

src/urllib3/connection.py
jalopezsilva: Good catch, it's been removed. Left over from initial code.

jalopezsilva:
@pquentin Thanks for your initial review. I'm adding more context here. 

Our use case is fetching HTTP / HTTPS resources through an HTTPS proxy. We run a forward proxy in our production environment through which employee and micro-services traffic goes through. We want to ensure this traffic is encrypted in all cases and we also want to easily identify clients through mutual authentication. (i.e clients will present a certificate to the proxy identifying themselves. Traffic not properly authenticated will be rejected)

Right now urllib3 ignores the proxy scheme. If you provide a proxy with an HTTPS scheme ``https://proxy``, the scheme will be silently ignored and urllib3 will attempt to contact the proxy through plaintext. HTTPS proxies are becoming more common ([libcurl](https://curl.haxx.se/libcurl/c/CURLOPT_PROXYTYPE.html) now supports them) and as key library in the python ecosystem, urllib3 should also support them.

This PR adds support for HTTPS proxies in two ways:
* HTTP resources through an HTTPS proxy by "bridging" or forwarding.
* HTTPS resources through an HTTPS proxy by "bridging" or forwarding. This is very insecure though, and after our conversation above with Alex, I'm throwing an exception if anyone attempts to use this unless a proper flag has been given.

Proper fetching of HTTPS resources through an HTTPS proxy requires TLS to the proxy, HTTP CONNECT, and TLS to the destination. This is the behavior libcurl provides. I have ideas of how to add this (discussion above about ``SSLContext.wrap_bio`` methods) but that should go in a separate PR I believe,

Hopefully this all makes sense, let me know if you have any further questions or comments.

pquentin:
@jalopezsilva Thank you for the context, that makes a lot of sense. The client authentication argument is particularly convincing! It's an elegant way to trace the requests of both employees and other micro-services, I might steal that idea at $DAYJOB one day. :)

I opened https://github.com/urllib3/urllib3/pull/1733 to fix the lint failure, sorry about that.

Here's what needs to happen from my point of view before we can merge this PR:

 * Improve our documentation. The "Proxies" section of the "Advanced usage" page is currently way too low on details: we need to document how to use proxies with 1. explicit plaintext HTTP 2. HTTP CONNECT + TLS to the end resource and 3. explicit HTTPS
 * Rename `allow_https_proxy_to_see_traffic` to `_allow_https_proxy_to_see_traffic` and mention that it could be removed at any time -- I want to be able to drop this option when we get TLS within TLS support
 * I need to test this locally, but I don't know yet how to configure mitmproxy or Charles Proxy (or anything else that runs on macOS) to just bridge/forward explicit HTTP/HTTPS requests while still connecting to the proxy using TLS
 * Discuss with @urllib3/maintainers to see how to handle this breaking change

@urllib3/maintainers Okay, so why is this a breaking change?

urllib3 only uses proxies through HTTP right now. We have two modes:
 * When requesting an HTTP resource, we ask the proxy to make the request on our behalf.
 * When requesting an HTTPS resource, we use HTTP CONNECT and set up a TLS connection to the requested resource, using the proxy as a dumb pipe.

While it's currently possible to specify an HTTPS proxy for both HTTP and HTTPS traffic, we still connect to the proxy via HTTP, not HTTPS. After this pull request, for HTTPS proxies we'll:

 1. connect to the proxy with HTTPS, but there's no guarantee that the HTTPS connection will succeed even if the HTTP connection succeeded.
 2. refuse to proceed for HTTPS requests without the escape hatch, but that means we could start refusing requests that were going through until now. (Another option here is to continue to request HTTPS traffic through HTTP, even if it's bad for security.)

However, those are absolutely the right things to do, and even if most users are using HTTP proxies, we might have users in the wild who think they're connecting to their proxy using TLS while they're not, or users who want things to work this way.

How do we want to handle this?

aausch:
> I need to test this locally, but I don't know yet how to configure mitmproxy or Charles Proxy (or anything else that runs on macOS) to just bridge/forward explicit HTTP/HTTPS requests while still connecting to the proxy using TLS

You can use something like `nginx` to terminate the https connection, and have the proxy then just do `http` -> `http` behind it

pquentin:
@jalopezsilva Thank you for the new commit!

I just realized there's a good way around the compatibility issue: we should provide a new HTTPSProxyManager for HTTPS proxies, and keep ProxyManager for HTTP proxies. And possibly add a warning when ProxyManager is used with a https proxy URL. What do you think?

jalopezsilva:
Having a ```HTTPSProxyManager`` is a good idea, although I wonder if it will make adoption of HTTPS proxies more difficult for libraries using urllib3. Most of the libraries I've seen (requests, botocore) use the ``proxy_from_url`` method to get the corresponding ``ProxyManager``. Are you thinking of also updating ``proxy_from_url`` to return a ``HTTPSProxyManager`` when provided with an HTTPS proxy URL? That would make it easier for libraries to start using HTTPS proxies. We could probably add a flag (``enable_https_proxies``) on ``proxy_from_url`` to enable or disable HTTPS proxies. The flag would be disabled initially but perhaps down the road we can make it enabled by default.

Another option would be to add that ``enable_https_proxies`` flag directly on ``ProxyManager`` and use that as a gating mechanism. Disabled initially, enabled by default in the future hopefully! 

Either way works for me! Let me know your thoughts @pquentin and I'll make the needed changes. Also let me know if you need help setting up a forward proxy to test this out, I can try to set up a docker image or something along those lines.

sethmlarson:
I'm just popping in to say:

- A breaking change that makes HTTPS proxy connections actually use HTTPS instead of silently using HTTP is a welcome change. :) People shouldn't not know that their connections aren't happening securely.
- I'll give this PR a review soon, busy time in my life. Thanks to @jalopezsilva and all participating :tada:

src/urllib3/connection.py
```diff
@@ -171,10 +170,13 @@ def _new_conn(self):
 
         return conn
 
+    def _is_using_tunnel(self):
```
sethmlarson: This is a good change here.

jalopezsilva:
Sorry for the delay in updating here, busy times in Novemeber! I have updated the PR with two additional commits:

- Commit # 1 : Addresses the documentation and CONTRIBUTORS.txt comments that @pquentin had. 

- Commit # 2 : The aim of this commit is to minimize some of the concerns that we have with introducing a breaking change. It's possible that clients right now have proxy configurations with an HTTPS proxy scheme which are being contacted through HTTP. If we merge my changes as is, urllib3 will stop working for such clients. To minimize the impact, I'm disabling the HTTPS proxy behavior unless a flag is provided (``_enable_https_proxies``). I'm printing a warning when I detect there's an invalid configuration and I'm keeping the old behavior. In future releases, we can remove this flag and have HTTPS proxies enabled by default.

Hope the second commit addresses the concerns with the breaking change! We can keep commit # 2 or discard it, whatever works.

Looking forwarding to merging this soon! Thanks!

jalopezsilva:
Happy New Year @pquentin, @sethmlarson! Any chance we could review this? I believe my last commit addresses the remaining concerns with breaking changes. 

Would love to have this merged soon!

pquentin:
Thanks for the friendly tone @jalopezsilva, I'm sure the situation is very frustrating to you, and I would like to apologize for delay here. The pull request looks good to me and the general idea has been approved by @sethmlarson -- I'm now going to coordinate with him to understand what the next steps are.

src/urllib3/poolmanager.py
```diff
@@ -306,6 +319,18 @@ def _merge_pool_kwargs(self, override):
                     base_pool_kwargs[key] = value
         return base_pool_kwargs
 
+    def _proxy_requires_complete_url(self, parsed_url):
```
sethmlarson: Let's use HTTP terminology here with `_proxy_requires_url_absolute_form`?

src/urllib3/poolmanager.py
```diff
@@ -383,6 +408,17 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
+    :param _enable_https_proxies:
```
sethmlarson: I'm not a fan of adding flags, especially if this is a new feature. Can we have this feature be additive and then do what we're already doing in this PR and error out in an invalid scenario (HTTPS+HTTPS)?

Logical next-steps are to figure out if we can also support HTTPS+HTTPS before we release this so there's less confusion.

jalopezsilva: Sounds good, I'll error out then. I preferred to set a warning as I was concerned that multiple clients with an incorrect configuration (i.e HTTPS + HTTPS) would be affected. But then again, it's an incorrect configuration so they might as well fix it now. 

That way, when we add the HTTPS + HTTPS support clients should have their configuration up to date.

I'll make the changes and update here.

sethmlarson: That sounds great, thank you for this work! It'll make many more users connections secure. :)

jalopezsilva:
@sethmlarson Whenever you have a chance, give this another go. Should be good to go!

I'll create another PR for the HTTPS / HTTPS support as we had agreed.

jalopezsilva:
Hey @sethmlarson, I'm starting to add the TLS within TLS support over in #1806. It would be super useful if we can merge this in so I can continue work on that PR. I think we're good to go here, let me know if we need to make any further changes.

jalopezsilva:
Hey @sethmlarson @pquentin, another ping here as I haven't heard from you guys in a while. Are we still planning on merging these changes?

Facebook has been using them internally for the past 4 months without any issues. I would prefer to merge them as I want to avoid having to fork FB's internal version of urllib3. That said, I'm starting to get concerned on how long this is taking and whether we'll be able to merge them or not.  

Do let me know so I know how to proceed forward. Thanks!

src/urllib3/poolmanager.py
```diff
@@ -452,14 +487,35 @@ def _set_proxy_headers(self, url, headers=None):
...
+            and not self.allow_insecure_proxy
+        ):
+            warnings.warn(
+                "Your proxy configuration specified an HTTPS scheme for the proxy. "
+                "Are you sure you want to use HTTPS to contact the proxy? "
+                "This most likely indicates an error in your configuration."
+                "If you are sure you want use HTTPS to contact the proxy, enable "
+                "the _allow_https_proxy_to_see_traffic.",
+                InvalidProxyConfigurationWarning,
+            )
```
pquentin: I'm trying to understand: what is the purpose of this warning if we're raising an exception next about the same issue?

pquentin:
Merging, even with the typo, we can always fix it later, eg. by removing "the". Thanks @jalopezsilva for your work and your patience. Tested again, it works well, and the code looks good to me.

sethmlarson:
Sorry for fumbling the review of this PR hard, I had tested the patch locally and it seemed to work. Thanks for tackling this. :)

jalopezsilva:
Thanks for the prompt response guys! I'll fix the typo and follow up in #1806.

PR #1686, opened 2019-09-20
Drop Python 3.4

pquentin:
It reached its end-of-life on 2019-03-18 and pip dropped support too.

(It's a follow-up to #1604)


pquentin:
@sethmlarson Thank you for the review! Is there anything left to do here?

sethmlarson:
Nope, just to merge! Thank you for doing this @pquentin :)

PR #1689, opened 2019-09-23
travis: install MacPython only once

pquentin:
The Travis macOS base image is documented to include Python: we can use
it to run nox instead of installing Python 3 explicitly, which should
result in faster builds.


pquentin:
The coverage improves because upload_coverage.sh now succeeds to install coverage in the macOS Python 2.7 build. The real fix is to just use `python` there, but I was planning to do that in another PR.

pquentin:
Thank you @sethmlarson for the quick reviews: they're really helpful!

sethmlarson:
No problem! :) Thank you for making all these changes.

PR #1691, opened 2019-09-24
Add hostname to insecure connection warning.

Steltek:
Sometimes, you're not the one who wrote the code, or urllib3 is being used within existing software, and it will give off the 'Unverified HTTPS request' warning, leaving the end user confused and unable to trace why they are getting it. (Example: Home Assistant components calling out to various hosts)

This PR adds the hostname to the warning, making it more obvious to end users where the source of the request lies.

(Very bad coder here so this is my second attempt. I doubt the code quality is up to standards, but it's what "black" gave me, and it does pass the Travis CI and AppVeyor builds on my end now.)

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -996,12 +996,10 @@ class HTTPSConnectionPool(HTTPConnectionPool):
         if not conn.is_verified:
             warnings.warn(
                 (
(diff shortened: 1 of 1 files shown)


pquentin:
Thanks @Steltek and @sethmlarson!

PR #1700, opened 2019-09-27
Merge strings that fit on a single line

pquentin:
The previous state was caused by black reformatting the code.


pquentin:
Yeah, that would be nice. Rebased on top of master, let's see if the build passes.

pquentin:
Thank *you*!

PR #1703, opened 2019-10-02
Remove TLS 1.3 support in SecureTransport

### LINKED ISSUE #1674: SecureTransport module does not actually support TLS 1.3
I believe that we can't make get TLS 1.3 support from SecureTransport, even if the documentation suggests otherwise. I made various requests and looked at the Wireshark traces:

 * Firefox and Safari do use TLS 1.3
 * urllib3 uses TLS 1.2 at best, and actually fails when TLS 1.3 is requested by modifying the code, exactly like https://forums.developer.apple.com/thread/110369
 * curl (compiled with SecureTransport support) uses TLS 1.2 at best, but never fails

I made sure to:

* use `sudo defaults write /Library/Preferences/com.apple.networkd tcp_conect_enable_tls13 1`,
* look at the ClientHello message
* check that TLS 1.3 messages were real, even if they're pretending they're TLS 1.2 per the spec.

Additionally, we never tested SecureTansport TLS 1.3 support in CI because we were using Python 3.7.1 which does not support TLS 1.3. Using Python 3.7.4, `test_tls_protocol_name_of_socket` started to fail with SecureTransport: we were requesting TLS 1.3, but got back TLS 1.2. (which is why I removed TLS 1.3 from SecureTransport tests in https://github.com/urllib3/urllib3/pull/1671.)

Now that SecureTransport is deprecated, we won't get any help from Apple (again, see https://forums.developer.apple.com/thread/110369 which recommends using the new library instead). I think we should remove the TLS 1.3 code from the SecureTransport module and document that we don't support TLS 1.3 with SecureTransport.

pquentin:
It's not actually supported by the OS. In other words, instead of trying
TLS 1.3 and being forced to fallback on TLS 1.2, we just use TLS 1.2 by
default.

Closes #1674.

### DIFF
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -29,9 +29,6 @@ if [[ "$(uname -s)" == 'Darwin' ]]; then
 
     install_mac_python $MACPYTHON
 
-    # Enable TLS 1.3 on macOS
-    sudo defaults write /Library/Preferences/com.apple.networkd tcp_connect_enable_tls13 1
-
     # Install Nox
     python3 -m pip install nox
 
--- a/src/urllib3/contrib/_securetransport/bindings.py
+++ b/src/urllib3/contrib/_securetransport/bindings.py
@@ -415,6 +415,7 @@ class SecurityConst(object):
     kTLSProtocol1 = 4
     kTLSProtocol11 = 7
     kTLSProtocol12 = 8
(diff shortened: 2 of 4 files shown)


sethmlarson:
Thanks @pquentin! :tada:

pquentin:
Thank you for the review!

PR #1704, opened 2019-10-03
Stop relying on SERVER_SOFTWARE being set

### LINKED ISSUE #1470: contrib/_appengine_environ.py KeyError SERVER_SOFTWARE

zevdg:
fixes #1470 
Also, mark functions that rely on GAE_VM as broken.

### DIFF
--- a/src/urllib3/contrib/_appengine_environ.py
+++ b/src/urllib3/contrib/_appengine_environ.py
@@ -6,7 +6,7 @@ import os
 
 
 def is_appengine():
-    return 'APPENGINE_RUNTIME' in os.environ
+    return "APPENGINE_RUNTIME" in os.environ
 
 
 def is_appengine_sandbox():
@@ -15,14 +15,15 @@ def is_appengine_sandbox():
 
 
 def is_local_appengine():
-    return (is_appengine() and
-            os.environ.get('SERVER_SOFTWARE','').startswith('Development/'))
+    return is_appengine() and os.environ.get("SERVER_SOFTWARE", "").startswith(
+        "Development/"
+    )
 
 
 def is_prod_appengine():
-    return (
-        is_appengine() and
-        os.environ.get('SERVER_SOFTWARE','').startswith('Google App Engine/'))
+    return is_appengine() and os.environ.get("SERVER_SOFTWARE", "").startswith(
+        "Google App Engine/"
+    )
 
 
 def is_prod_appengine_mvms():
--- a/src/urllib3/contrib/_appengine_environ.py
+++ b/src/urllib3/contrib/_appengine_environ.py
@@ -6,27 +6,25 @@ import os
 
 
 def is_appengine():
-    return is_local_appengine() or is_prod_appengine() or is_prod_appengine_mvms()
+    return 'APPENGINE_RUNTIME' in os.environ
 
 
 def is_appengine_sandbox():
-    return is_appengine() and not is_prod_appengine_mvms()
+    """Deprecated. Use is_appengine instead."""
+    return is_appengine()
 
 
 def is_local_appengine():
-    return (
-        "APPENGINE_RUNTIME" in os.environ
-        and "Development/" in os.environ["SERVER_SOFTWARE"]
-    )
+    return (is_appengine() and
+            os.environ.get('SERVER_SOFTWARE','').startswith('Development/'))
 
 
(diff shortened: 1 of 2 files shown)


src/urllib3/contrib/_appengine_environ.py
```diff
@@ -6,27 +6,26 @@
...
-    return (
-        "APPENGINE_RUNTIME" in os.environ
-        and "Google App Engine/" in os.environ["SERVER_SOFTWARE"]
-        and not is_prod_appengine_mvms()
+    return is_appengine() and os.environ.get("SERVER_SOFTWARE", "").startswith(
+        "Google App Engine/"
     )
 
 
 def is_prod_appengine_mvms():
```
theacodes: Is managed VMs actually **decommissioned**? We shouldn't remove this unless we have *no* users on it.

src/urllib3/contrib/_appengine_environ.py
```diff
@@ -6,27 +6,26 @@
 
 
 def is_appengine():
-    return is_local_appengine() or is_prod_appengine() or is_prod_appengine_mvms()
+    return "APPENGINE_RUNTIME" in os.environ
 
 
 def is_appengine_sandbox():
-    return is_appengine() and not is_prod_appengine_mvms()
+    """Deprecated. Use is_appengine instead."""
```
theacodes: This changes the foundational behavior of this method. It should detect if the app is running under the legacy 2.7 runtime in App Engine standard. Making it detect *any* app engine runtime is likely to cause user code to fail.

src/urllib3/contrib/_appengine_environ.py
zevdg: Yes.  Managed VMs are totally decommissioned.  There are no users still on it.

src/urllib3/contrib/_appengine_environ.py
zevdg: I guess I don't really know what the "foundational behavior" of this method is supposed to be.  I couldn't find any documentation for this function, so I only had the old code to go on which I interpreted as "all appengine runtimes except managed VMs are sandboxed."

Given that managed VMs are decommissioned, it seemed that now the situation is that "all appengine runtimes are sandboxed."  Am I missing something?

theacodes: So "sandboxed" here is a bit weird. You'll have to kinda go back to when this was written - which was before gVisor.

When urllib3 first added rudimentary support for App Engine, there was App Engine (with no qualifier) in a heavily (ptrace) sandboxed environment and there was App Engine managed VMs which had a much less severe KVM sandbox. Urllib3 can run as usual in the KVM sandbox, but can't run as usual in the ptrace sandbox. So this function is used to determine "can I really use sockets and ssl the in a normal way?" which breaks down to this table these days:

Standard legacy 2.7: *no*
Standard 3.x: *yes*
Flexible 2.x or 3.x: *yes*

So this function should return `True` only in the legacy 2.7 runtime *or* when using the legacy 2.7 runtime with `devappserver`. User code uses this function to determine whether or not to use the `AppEngineManager` (and therefore `urlfetch`) instead of the normal `PoolManager` (and therefore `sockets`).

zevdg: That makes sense.  I've changed the function to return True for the python27 runtime and added some documentation to explain the behavior.

theacodes:
@sethmlarson this should be ready to merge at your will. :)

zevdg:
@sethmlarson Friendly ping on this.  Is there anything left to do before this can be merged?

sethmlarson:
Thanks for this patch @zevdg! :tada:

Issue #1714, opened 2019-10-28
Retries during chunked transfers fail due to loss of chunked flag


lmvlmv:
Possibly related to [https://github.com/urllib3/urllib3/issues/1352](https://github.com/urllib3/urllib3/issues/1352)

Using urllib via requests but this doesn't seem to be the cause. [Recently, in requests](https://github.com/psf/requests/pull/5128), a fix was implemented to use the innate chunking capability in urllib. 

If a custom Retry class is implemented, a chunked transfer is taking place and a retry-able condition occurs during the transfer the chunked flag is dropped when urlopen() is re-called during the retry.

In this case a generator was passed as the data argument and results in the non-chunking version of the code attempting to read it like raw data: TypeError: 'generator' object has no attribute 'len'

lmvlmv:
Prototype fix here [https://github.com/lmvlmv/urllib3/tree/chunkfix](https://github.com/lmvlmv/urllib3/tree/chunkfix)

Having trouble getting nox to pass tests in my environment before I submit as a PR...

PR #1715, opened 2019-10-28
Preserve chunked flag when re-calling urlopen() in retry loop

lmvlmv:


lmvlmv:
Attmpts to fix [https://github.com/urllib3/urllib3/issues/1714](https://github.com/urllib3/urllib3/issues/1714)

lmvlmv:
NP. Thanks for that. Couldn't get the test env working locally. Since the change to requests more people will be visiting this function call so good to get it fixed first. Not sure when the next requests release will be...

sethmlarson:
Thanks @lmvlmv and @pquentin! :tada:

PR #1717, opened 2019-10-29
Unify timeouts in tests

### LINKED ISSUE #1706: Unify timeout values used in tests
We have lots of tests that use timeout values, eg. to check that a given operation timeouts as expected or completes a given amount of work before a certain amount of time. Unfortunately, in CI those tests tend to fail some of the time, because say 1% of the time everything is super slow.

In response to this, we went from 0.001 to 0.01 and sometimes 0.1 in the places that failed the most often (#1554 and #1496 are recent examples). But there are still some tests that fail because they were not updated. Since there is no test that is inherently slow in the test suite that I'm aware of, I think it would be nice to share a few timeout values (ideally two) in all tests, so that we can:

 1. bump them all at once if needed
 2. use smaller values in local environments, and larger values in CI

pquentin:
Please review #1712 first, as I included it in this pull request to avoid conflicts in the future.

Closes #1706

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -28,6 +28,16 @@ VALID_SOURCE_ADDRESSES = [(("::1", 0), True), (("127.0.0.1", 0), False)]
 # RFC 3849: 2001:db8::/32 is for documentation only.
 INVALID_SOURCE_ADDRESSES = [("192.0.2.255", 0), ("2001:db8::1", 0)]
 
+# We use timeouts in three different ways in our tests
+#
+# 1. To make sure that the operation timeouts, we can use a short timeout.
+# 2. To make sure that the test does not hang even if the operation should succeed, we
+#    want to use a long timeout, even more so on CI where tests can be really slow
+# 3. To test our timeout logic by using two different values, eg. by using different
+#    values at the pool level and at the request level.
+SHORT_TIMEOUT = 0.001
+LONG_TIMEOUT = 0.1 if os.environ.get("CI") else 0.01
+
 
 def clear_warnings(cls=HTTPWarning):
     new_filters = []
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -36,7 +36,7 @@ INVALID_SOURCE_ADDRESSES = [("192.0.2.255", 0), ("2001:db8::1", 0)]
 # 3. To test our timeout logic by using two different values, eg. by using different
 #    values at the pool level and at the request level.
 SHORT_TIMEOUT = 0.001
-LONG_TIMEOUT = 0.1 if os.environ.get("CI") else 0.01
+LONG_TIMEOUT = 0.5 if os.environ.get("CI") else 0.01
 
 
 def clear_warnings(cls=HTTPWarning):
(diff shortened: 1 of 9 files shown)


test/__init__.py
```diff
@@ -28,6 +28,16 @@
...
 INVALID_SOURCE_ADDRESSES = [("192.0.2.255", 0), ("2001:db8::1", 0)]
 
+# We use timeouts in three different ways in our tests
+#
+# 1. To make sure that the operation timeouts, we can use a short timeout.
+# 2. To make sure that the test does not hang even if the operation should succeed, we
+#    want to use a long timeout, even more so on CI where tests can be really slow
+# 3. To test our timeout logic by using two different values, eg. by using different
+#    values at the pool level and at the request level.
+SHORT_TIMEOUT = 0.001
```
sethmlarson: Looks like Travis (which is faster than AppVeyor usually) timed out on this `0.001` value. Maybe we need to have two separate values for the `SHORT_TIMEOUT` as well?

test/appengine/test_gae_manager.py
```diff
@@ -46,7 +47,7 @@ def setup_method(self, method):
     def test_exceptions(self):
         # DeadlineExceededError -> TimeoutError
         with pytest.raises(urllib3.exceptions.TimeoutError):
-            self.pool.request("GET", "/sleep?seconds=0.005", timeout=0.001)
+            self.pool.request("GET", "/sleep?seconds=0.005", timeout=SHORT_TIMEOUT)
```
sethmlarson: If we implement the above comment, can we have the `seconds=X` call be dynamic? Maybe 2 * SHORT_TIMEOUT?

test/__init__.py
pquentin: Oh, `test_ssl_read_timeout` failed, I had not noticed that, sorry. I think the better fix is to switch to `LONG_TIMEOUT` here. Here's my reasoning.

In this test, we:
 1. read the headers -- this should not timeout, so using `LONG_TIMEOUT` makes sense
 2. read the body -- this *should* timeout, so using `SHORT_TIMEOUT` makes sense

However, both timeouts are read timeouts and are controlled by a single value, so we have to compromise. Switching to `LONG_TIMEOUT` is only a compromise for this test, while bumping `SHORT_TIMEOUT` would be a compromise for all tests.

What do you think?

test/appengine/test_gae_manager.py
pquentin: Ah yes, great catch! I think I'll continue to use 5 * SHORT_TIMEOUT

pquentin:
Rebased on top of master and added two commits for your first two remarks.

pquentin:
I broke the lint, sorry. Will fix as soon as I can.

pquentin:
Closing/reopening to see if I just got lucky or not.

pquentin:
The AppVeyor failure is due to my test from #1715, I'll fix it later. The Travis failure is due to 100ms not being enough, I bumped LONG_TIMEOUT to 500ms.

pquentin:
Closing/reopening because it was a good idea the last time. :)

pquentin:
One last try.

test/__init__.py
sethmlarson: +1 LONG_TIMEOUT works for that test.

PR #1718, opened 2019-10-29
Retry failed dummyserver tests once

### LINKED ISSUE #1595: Add a pytest retry solution to all dummyserver tests

pquentin:
Closes #1595

### DIFF
--- a/test/with_dummyserver/test_chunked_transfer.py
+++ b/test/with_dummyserver/test_chunked_transfer.py
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 
+import pytest
+
 from urllib3 import HTTPConnectionPool
 from dummyserver.testcase import SocketDummyServerTestCase
+
+# Retry failed tests
+pytestmark = pytest.mark.flaky
 
 
 class TestChunkedTransfer(SocketDummyServerTestCase):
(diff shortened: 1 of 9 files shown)


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
 
+import pytest
+
 from urllib3 import HTTPConnectionPool
 from dummyserver.testcase import SocketDummyServerTestCase
 
+# Retry failed tests
+pytestmark = pytest.mark.flaky
```
sethmlarson: I've never seen this construction before, where is it documented? Must it be named `pytestmark`?

pquentin: Yes: http://doc.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules

I agree that it's quite obscure, which is why I added the comment. But it's way more convenient than marking every class or every test. Another option is do this for *all* tests, not only dummyserver tests: we could simply use a command-line option then.

sethmlarson: Today I learned! This looks good ready to go then. :)

PR #1720, opened 2019-10-30
Restore comment next to its intended line

pquentin:
I don't know what those errors mean, but I can tell that initially `For OSX` was next to `errnor.EHOSTUNREACH`: see #335. @shazow was right five years ago, because we're now using four lines anyway, and doing it from the start would have preserved the correct position. :)


shazow:
Loool my prediction was off by almost 2 years though, I'll work on that for next time.

Issue #1705, opened 2019-10-11
Generate trustme certificates when running the test suite


pquentin:
Since #1423 we're using certificates generated by [trustme](https://github.com/python-trio/trustme/) for the intermediate certificates tests. However, those certificates are still stored as opaque blobs, as all other test certificates.

It would be nice to always generate certificates on the fly when running the tests: we would stop storing blobs, we would benefit from trustme fixes, and the useful parts of our certificates would be fully specified in our test code. This is how trustme is intended to be used, is what [httpx does](https://github.com/encode/httpx/blob/fc3df514e81de8ff14d9ef43ee53caab408a50ee/tests/conftest.py#L153-L195) and will hopefully avoid issues like https://github.com/urllib3/urllib3/pull/1412 in the future.

A downside here is that while trustme currently supports Python 2.7, we don't know when it will drop support. However, having contributed to the code base for #1423, I can tell that it's a small library that is easy to work with, so adding it as a test dependency does not worry me too much. Worst case: we just use the last working version to generate blobs again.

I would consider this issue closed when all test certificates are ported by trustme (barring any feature that trustme does not support).

pquentin:
Our test certificates currently don't work on macOS 10.15, which has stricter rules about TLS certificates: https://support.apple.com/en-us/HT210176.

However, the certificates generated by trustme comply with those rules (see https://github.com/python-trio/trustme/issues/106). So closing this issue will also allow us to run our test suite with macOS 10.15. This nicely highlights the "we would benefit from trustme fixes" benefit mentioned in my previous message.

sethmlarson:
@pquentin That's great to hear. I didn't realize our certs wouldn't work on macOS 10.15. :)

pquentin:
Well I only realized it because the tests were failing on my machine. :)

pquentin:
I am tempted to close this issue, see https://github.com/urllib3/urllib3/pull/1831 for details.

sethmlarson:
Let's keep it open, I have someone who might want to take a stab at a new issue. :)

eduzen:
Hi, yes, it's me. I will take a look.

pquentin:
@eduzen The best way to learn what this is about is to look at the merged pull requests above.

I guess one nice self-contained task is to see how hard it is compute fingerprints from an in-memory trustme certificate. It does not have to be included in the code right away.

Another option is to try reducing the number of references to the remaining dummyserver/certificates certs.

[...]

PR #1721, opened 2019-10-30
Remove unused certs

pquentin:
Preliminary work towards #1705


pquentin:
Right, but switching completely to trustme won't be that straightforward, I don't know when the transition will be complete.

Thank you for the review!

PR #1722, opened 2019-10-30
Generate certificate directory at runtime

pquentin:
Instead of storing it in git.

Done as part of #1705


pquentin:
I just found out about http://doc.pytest.org/en/latest/tmpdir.html#the-tmpdir-fixture, so I pushed a new commit here

pquentin:
This pull request includes commits from #1724, so please review/merge #1724 first so that we have a record of it in the history.

pquentin:
This is also ready for review. :) But again, #1724 should be merged first.

sethmlarson:
Is there a conflict on this PR? Do we need these changes still?

pquentin:
This was merged as part of #1725. Thanks! Closing.

PR #1724, opened 2019-10-30
Fix our decorators to be compatible with pytest fixtures

pquentin:
This should help with the mess in #1722


sethmlarson:
Looks like a merge conflict on this branch.

PR #1608, opened 2019-05-16
Load system CA certificates by default on empty HTTPSConnection

sethmlarson:
Closes #1603.

PR #1726, opened 2019-10-31
Get coverage back to 100%

pquentin:
This means we should probably refactor the connection classes, but that would change the public API and I think it should be discussed in another pull request.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -251,40 +251,6 @@ class HTTPSConnection(HTTPConnection):
         # HTTPS requests to go out as HTTP. (See Issue #356)
         self._protocol = "https"
 
-    def connect(self):
-        conn = self._new_conn()
-        self._prepare_conn(conn)
-
-        # Wrap socket using verification with the root certs in
-        # trusted_root_certs
-        default_ssl_context = False
-        if self.ssl_context is None:
-            default_ssl_context = True
-            self.ssl_context = create_urllib3_context(
-                ssl_version=resolve_ssl_version(self.ssl_version),
-                cert_reqs=resolve_cert_reqs(self.cert_reqs),
-            )
-
-        # Try to load OS default certs if none are given.
-        # Works well on Windows (requires Python3.4+)
-        context = self.ssl_context
-        if (
-            not self.ca_certs
-            and not self.ca_cert_dir
-            and default_ssl_context
-            and hasattr(context, "load_default_certs")
-        ):
-            context.load_default_certs()
-
-        self.sock = ssl_wrap_socket(
-            sock=conn,
-            keyfile=self.key_file,
-            certfile=self.cert_file,
-            key_password=self.key_password,
-            ssl_context=self.ssl_context,
-            server_hostname=self.server_hostname,
-        )
-
 
 class VerifiedHTTPSConnection(HTTPSConnection):
     """
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -322,9 +322,6 @@ def _idna_encode(name):
 
 def _encode_target(target):
     """Percent-encodes a request target so that there are no invalid characters"""
-    if not target.startswith("/"):
-        return target
-
     path, query = TARGET_RE.match(target).groups()
     target = _encode_invalid_chars(path, PATH_CHARS)
     query = _encode_invalid_chars(query, QUERY_CHARS)
--- a/test/appengine/test_gae_manager.py
+++ b/test/appengine/test_gae_manager.py
@@ -168,3 +168,11 @@ class TestGAERetryAfter(test_connectionpool.TestRetryAfter):
         # Disable urlfetch which doesn't respect Retry-After header.
         self.manager = appengine.AppEngineManager(urlfetch_retries=False)
         self.pool = MockPool(self.host, self.port, self.manager)
+
+
+def test_gae_environ():
+    assert not appengine.is_appengine()
+    assert not appengine.is_appengine_sandbox()
+    assert not appengine.is_local_appengine()
+    assert not appengine.is_prod_appengine()
+    assert not appengine.is_prod_appengine_mvms()


src/urllib3/util/url.py
```diff
@@ -322,17 +322,14 @@ def _idna_encode(name):
...
-
     path, query, fragment = TARGET_RE.match(target).groups()
     target = _encode_invalid_chars(path, PATH_CHARS)
     query = _encode_invalid_chars(query, QUERY_CHARS)
     fragment = _encode_invalid_chars(fragment, FRAGMENT_CHARS)
     if query is not None:
         target += "?" + query
     if fragment is not None:
-        target += "#" + target
+        target += "#" + fragment
```
sethmlarson: Nice bug fix!

test/with_dummyserver/test_connectionpool.py
```diff
@@ -715,6 +715,11 @@ def test_percent_encode_invalid_target_chars(self):
             r = pool.request("GET", "/echo_params?q=\r&k=\n \n")
             assert r.data == b"[('k', '\\n \\n'), ('q', '\\r')]"
 
+    def test_percent_encode_fragment(self):
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            r = pool.request("GET", "/echo_params?q=1#fragment")
+            assert r.data == b"[('q', '1#fragment')]"
```
sethmlarson: Shouldn't this be `q=1` without the fragment? Maybe the `/echo_params` endpoint isn't the best choice for this unit test.

src/urllib3/connection.py
```diff
@@ -251,40 +251,6 @@ def __init__(
         # HTTPS requests to go out as HTTP. (See Issue #356)
         self._protocol = "https"
 
-    def connect(self):
```
sethmlarson: We probably shouldn't remove this, what happens if a user has an unverified HTTPS connection, is this class used?

pquentin: No, it's not used, unverified connections go through VerifiedHTTLS connection

sethmlarson: Oh, huh. I wonder what the impact of nuking the class and rebinding the name would do to downstream. :man_shrugging:

test/with_dummyserver/test_connectionpool.py
pquentin: Should we send the fragment identifier, though?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Identifying_resources_on_the_Web#Fragment says "the part after the #, also known as fragment identifier, is never sent to the server with the request."

As will all URL questions, this appears to be quite a complicated subject, because it looks like there are special cases with redirects, for example. If only we had an URL expert available... :) cc @sigmavirus24

pquentin: @sethmlarson I'm now using `/echo_uri` here.

src/urllib3/connection.py
pquentin: @sethmlarson I opened https://github.com/urllib3/urllib3/issues/1730 to unblock this pull request and talk about our options here, including the upstream impact.

src/urllib3/util/url.py
pquentin: Yeah, only 7 uncovered lines, and one bug! I should have seen it when reviewing the PR, but didn't :)

pquentin:
This is ready for another round of review!

src/urllib3/util/url.py
sethmlarson: I'm pretty sure we shouldn't be sending this in our request tbh. Request target is path+query.

src/urllib3/connection.py
sethmlarson: Could we try running vcrpy's test suite on this branch to ensure removing this function doesn't break anything?

pquentin: I can confirm that the vcrpy test suite continue to work with this change, both for requests and urllib3

src/urllib3/util/url.py
pquentin: Resolving as this was fixed in #1732.

test/with_dummyserver/test_connectionpool.py
pquentin: Resolving as this was fixed in #1732.

src/urllib3/util/url.py
pquentin: Resolving as this was fixed in #1732.

pquentin:
Travis fails because of #1733, I'll rebase this PR when it gets merged, that will fix the issue. AppVeyor fails because my test for #1715 is not fully reliable. I plan to fix this soon too.

pquentin:
test_preserve_chunked_on_retry is a test that I recently added and it failed *again*. It was retried due to our flaky integration, but we got two different exceptions, both unexpected. The test is definitely not reliable on Windows, and I decided to change tactics to test this code path: an explicit Retry-After header. Opened https://github.com/urllib3/urllib3/pull/1743 for this.

sethmlarson:
I'd like to release 1.25.7 before making large changes, will return to it after the release has been made. :)

pquentin:
Works for me 👍

pquentin:
@sethmlarson Is there anything I can do here? I understand that it's frightening, but I'm only removing code that was already broken anyway.

pquentin:
Thanks!

src/urllib3/connection.py
zogzog: > No, it's not used, unverified connections go through VerifiedHTTPSConnection.

You mean, it was not covered by the existing test base and you thought it was a good case for its removal ?

sethmlarson: @zogzog Sorry for causing an issue. I've edited your above comment as I read it as sarcastic which isn't constructive. Hope you agree with my edit.

zogzog: > I don't think anybody is using this class: connect() referenced variables that don't exist, so could not work. 

Weird statement. My code definitely go through this path.
Indeed I had to add

    cert_reqs = 'CERT_NONE'
    ca_certs = None
    ca_cert_dir = None

6 months ago because someone removed them or something, and urllib3 doesn't have coverage for this indeed. This not the correct way to kill a feature though.

I think it came with https://github.com/urllib3/urllib3/pull/1608/commits/6c6196717eaa29ffb2269037762f958cb4f6ed79 

> An additional wrinkle here is the bottom of the file:
> 
> ```
> if ssl:
>     # Make a copy for testing.
>     UnverifiedHTTPSConnection = HTTPSConnection
>     HTTPSConnection = VerifiedHTTPSConnection
> ```
> 
> So one option is to merge VerifiedHTTPSConnection into HTTPSConnection, and add `UnverifiedHTTPSConnection = HTTPSConnection` at the end of the file.

People should really understand that, even though it is generally considered a bad thing and should be discouraged, there is a need for:

* unverified https connections,
* that don't spit warnings on the console.

I have to maintain something very complicated like this

```python

# HTTPS verfication/un-warning

# HTTPS warning disabled for mercure requests
# We don't want to verify the Mercure certificate because
# it is too complicated to maintain.

# We cannot disable this warning using standard catch_warning
# tricks because they are not thread-safe,
# hence we go to greater lengths to *not emit* the warning.

# We monkey-patch HTTPSConnectionPool._validate_conn to make it
# recognize a new special case, and provide a Session object
# which triggers the special case.


def _validate_conn(self, conn):
    """
    Called right before a request is made, after the socket is created.
    """
    # Force connect early to allow us to validate the connection.
    if not getattr(conn, 'sock', None):  # AppEngine might not have  `.sock`
        conn.connect()

    # PATCH
    if isinstance(conn, SilentUnverifiedHTTPSConnection):
        return
    # /PATCH

    if not conn.is_verified:
        warn((
            'Unverified HTTPS request is being made. '
            'Adding certificate verification is strongly advised. See: '
            'https://urllib3.readthedocs.io/en/latest/advanced-usage.html'
            '#ssl-warnings'),
            requests.packages.urllib3.exceptions.InsecureRequestWarning)


requests.packages.urllib3.connectionpool.HTTPSConnectionPool._validate_conn = _validate_conn


class SilentUnverifiedHTTPSConnection(
        requests.packages.urllib3.connection.UnverifiedHTTPSConnection):
    # the only significant information here is a connection of a
    # specific type, sufficient for an isinstance check

    cert_reqs = 'CERT_NONE'
    ca_certs = None
    ca_cert_dir = None


class NoHTTPSVerifyAdapter(requests.sessions.HTTPAdapter):

    def cert_verify(self, cn, *_args):
        # we override this (class) attribute on the fly
        # (cn is actually a connection pool)
        cn.ConnectionCls = SilentUnverifiedHTTPSConnection


class NoHTTPSVerifySession(requests.Session):

    def __init__(self):
        super(NoHTTPSVerifySession, self).__init__()
        # we redo the adapters mapping with our custom
        # silence-inducing adapters
        self.adapters = OrderedDict()
        self.mount('https://', NoHTTPSVerifyAdapter())
        self.mount('http://', NoHTTPSVerifyAdapter())

```

I would LOVE getting rid of this.

pquentin: Aurélien, je comprends ta colère, je sais ce que c'est de tomber sur un commit quelque part sur GitHub qui casse ton code. Je te présente d'ailleurs mes excuses. Maintenant que ça c'est dit, on apprécierait des interventions plus courtoises !

> Indeed I had to add [...] 6 months ago because someone removed them or something

We would have appreciated a heads-up at this point! This would have avoided the whole issue.

I'd be happy to accept a pull request that restores a working and covered connect().

> People should really understand that, even though it is generally considered a bad thing and should be discouraged, there is a need for unverified https connections that don't spit warnings on the console.

Can you use our fingerprinting feature if the certificate does not change too often?

Is silencing the warning an option?

By the way, it looks like the TLS certificate of https://pythonian.fr/ is currently expired.

zogzog: Sorry for venting my frustration, I didn't wait long enough to let it wind down before writing here. Now a constructive interaction there is all we need :)

6 months ago I just didn't have the time to do a root cause analysis of our issues. Priorities, etc. That's life. 

The whole point about unverified ssl (in our case, I guess it is a real issue in the corporate world in many places) is that we can't afford to break our tools if the IT department forgets to update a certificate or even uses a self-signed thing on some pre-production instance.

I have nothing against unverified https connection warning _by default_ but having a "don't warn, I know what I am doing" flag would be appreciated.

The tool has a command line for casual queries and we don't want the end-users to be pestered about the "unsafe" connection. We are basically powerless against it.

So yes, silencing the warning should be an option imho.

pquentin: I mean, why don't you just silence the warning? Isn't this working for you?

```python3
import urllib3

urllib3.disable_warnings(urllib3.exceptions.SecurityWarning)
http = urllib3.PoolManager(cert_reqs="CERT_NONE")
r = http.request("GET", "https://pythonian.fr/webfiles/css.css")
print(r.status)
```

pquentin: Also, for what it's worth, here are the thoughts of Seth on verifying HTTPS: https://sethmlarson.dev/blog/2019-11-26/designing-for-real-world-https. I happen to fully agree.

sethmlarson: There's also this method (using certificate pinning as mentioned above) that doesn't require you to disable security warnings. :)

```python
http = urllib3.PoolManager(
    cert_reqs="CERT_NONE",
    assert_fingerprint="e677022f45ad8ecfc7a66dbb8d57e9c7b01a52c9f880ecb104af6ee67ee9f18a"
)
resp = http.request("GET", "https://pythonian.fr/webfiles/css.css")
```

Just remember this PoolManager will only work for the one website. Might be worth even using a ConnectionPool instead.

zogzog: > I mean, why don't you just silence the warning? Isn't this working for you?
> 
> ```python
> import urllib3
> 
> urllib3.disable_warnings(urllib3.exceptions.SecurityWarning)
> http = urllib3.PoolManager(cert_reqs="CERT_NONE")
> r = http.request("GET", "https://pythonian.fr/webfiles/css.css")
> print(r.status)
> ```

Looks like it globally disables warnings, right ? We don't want that !

zogzog: > Also, for what it's worth, here are the thoughts of Seth on verifying HTTPS: https://sethmlarson.dev/blog/2019-11-26/designing-for-real-world-https. I happen to fully agree.

I know all of this. Yet this does not seem to address all the real world issues. We are walking a fine line there between the need to operate securely by default and the need to make things work in the face of a failing security setup. Operational constraints make all the propositions there, though well-meant, not workable in practice.

zogzog: > There's also this method (using certificate pinning as mentioned above) that doesn't require you to disable security warnings. :)
> 
> ```python
> http = urllib3.PoolManager(
>     cert_reqs="CERT_NONE",
>     assert_fingerprint="e677022f45ad8ecfc7a66dbb8d57e9c7b01a52c9f880ecb104af6ee67ee9f18a"
> )
> resp = http.request("GET", "https://pythonian.fr/webfiles/css.css")
> ```
> 
> Just remember this PoolManager will only work for the one website. Might be worth even using a ConnectionPool instead.

This fingerprint, which is a piece of data related to the _current_ possibly broken certificate would cease to operate when the certificate is renewed right ? Also I would have to either:
* hardcode it and then the software would need to be re-released if the upstream certificate changed
* made a config item, whose actual value would have to be transmitted to end users in the case of change

Am I missing something ? Either way this is not practical enough.

PR #1734, opened 2019-11-05
Preserve chunked parameter

pquentin:
As promised in #1715 

cc @lmvlmv @sethmlarson 

The lint fails because of #1733.


pquentin:
Rebased on top of master to get the lint fix.

pquentin:
This is a typical case where the codecov report makes no sense, this change does not affect line coverage at all!

pquentin:
Okay the codecov report now looks good.

PR #1745, opened 2019-11-11
Release 1.25.7

sethmlarson:


pquentin:
Yay, a release!

I think there are a few missing changelog entries:

 * Preserve chunked parameter on retries (Pull #1715, Pull #1734)
 * Allow unset SERVER_SOFTWARE in App Engine (Pull #1704, Issue #1470)
 * Remove TLS 1.3 support in SecureTransport (Pull #1703)

Not sure about the wording, but they are user-visible changes. Well, the SecureTransport change is not really visible, but we advertised TLS 1.3 support for all TLS backends in the past, and it's no longer there.

sethmlarson:
You're right, I'll add all those to the changelog! Thanks :)

sethmlarson:
@pquentin When you get a second to gloss over my addition that'd be great.

sethmlarson:
:tada: Will be uploading the release shortly.

PR #1747, opened 2019-11-15
Update dummyserver to work with newer Tornado

felixonmars:
The tornado.web.asynchronous decorator was deprecated in Tornado 5, and
removed in Tornado 6. This commit rewrites the dummyserver to work on
newer Tornado.

### DIFF
--- a/dummyserver/proxy.py
+++ b/dummyserver/proxy.py
@@ -28,6 +28,7 @@
 import sys
 import socket
 
+import tornado.gen
 import tornado.httpserver
 import tornado.ioloop
 import tornado.iostream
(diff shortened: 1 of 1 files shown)


pquentin:
I confirmed that this works with Tornado 6 (as we only test with Tornado 5 in CI), and even works with the changes from #1679. Thank you @felixonmars, this is a great contribution!

As more and more dependencies drop Python 2.7, I wonder if we should use the newer stuff for Python 3.x and the latest supported release for Python 2.7?

felixonmars:
That would be the best, perhaps with conditional dependencies?

pquentin:
@sethmlarson What do you think?

PR #1750, opened 2019-11-17
Add default user agent header

DonaCthulhuote:
For #1296 

No tests fail that didn't already fail when I ran them locally (which isn't all that hot, I know; not sure what's going on there)

Should I try to add this behaviour to the docs as well, or it okay being implicit?

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -205,8 +205,6 @@ class HTTPConnection(_HTTPConnection, object):
         headers = HTTPHeaderDict(headers if headers is not None else {})
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
-        elif headers["User-Agent"] is None:
-            del headers["User-Agent"]
         super(HTTPConnection, self).request(method, url, body=body, headers=headers)
 
     def request_chunked(self, method, url, body=None, headers=None):
@@ -222,8 +220,6 @@ class HTTPConnection(_HTTPConnection, object):
         )
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
-        elif headers["User-Agent"] is None:
-            del headers["User-Agent"]
         for header, value in headers.items():
             self.putheader(header, value)
         if "transfer-encoding" not in headers:
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -47,7 +47,7 @@ from .util.ssl_ import (
 )
 
 
-from .util import connection
+from .util import connection, SUPPRESS_USER_AGENT
 
 from ._collections import HTTPHeaderDict
 from ._version import __version__
@@ -205,6 +205,8 @@ class HTTPConnection(_HTTPConnection, object):
         headers = HTTPHeaderDict(headers if headers is not None else {})
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
+        elif headers["user-agent"] == SUPPRESS_USER_AGENT:
+            del headers["user-agent"]
         super(HTTPConnection, self).request(method, url, body=body, headers=headers)
 
     def request_chunked(self, method, url, body=None, headers=None):
@@ -220,6 +222,8 @@ class HTTPConnection(_HTTPConnection, object):
         )
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
+        elif headers["user-agent"] == SUPPRESS_USER_AGENT:
+            del headers["user-agent"]
         for header, value in headers.items():
             self.putheader(header, value)
         if "transfer-encoding" not in headers:
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -50,6 +50,7 @@ from .util.ssl_ import (
 from .util import connection
 
 from ._collections import HTTPHeaderDict
+from ._version import __version__
 
 log = logging.getLogger(__name__)
 
@@ -200,14 +201,11 @@ class HTTPConnection(_HTTPConnection, object):
 
         return _HTTPConnection.putrequest(self, method, url, *args, **kwargs)
 
-    def request(self, method, url, **httplib_request_kw):
-        headers = httplib_request_kw.pop("headers", {})
-        headers = HTTPHeaderDict(headers)
+    def request(self, method, url, body=None, headers=None):
+        headers = HTTPHeaderDict(headers if headers is not None else {})
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
-        super(HTTPConnection, self).request(
-            method, url, headers=headers, **httplib_request_kw
-        )
+        super(HTTPConnection, self).request(method, url, body=body, headers=headers)
 
     def request_chunked(self, method, url, body=None, headers=None):
         """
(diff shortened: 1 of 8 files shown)


DonaCthulhuote:
Well, that clearly isn't good.  Let me look into that tomorrow

DonaCthulhuote:
Hopefully that one does it; lacking access to a non-windows machine at the moment, which is my guess for why I'm seeing tens of inconsistent failures locally.

src/urllib3/__init__.py
```diff
@@ -20,9 +20,9 @@
 import logging
 from logging import NullHandler
 
-__author__ = "Andrey Petrov (andrey.petrov@shazow.net)"
-__license__ = "MIT"
-__version__ = "1.25.7"
+__author__ = constants.__author__
```
sethmlarson: Let's revert the addition of `constants.py` and instead create a function that generates the `User-Agent` header value. Within that function you can import `__version__` without causing a cyclic import. :)

src/urllib3/request.py
```diff
@@ -67,6 +71,11 @@ def request(self, method, url, fields=None, headers=None, **urlopen_kw):
         :meth:`request_encode_url`, :meth:`request_encode_body`,
         or even the lowest level :meth:`urlopen`.
         """
+        if headers is None:
```
sethmlarson: Instead of adding the value upstream we should add it within `HTTPConnection` so we aren't modifying any upstream behaviors.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1401,6 +1403,41 @@ def test_headers_are_sent_with_the_original_case(self):
...
+            "User-Agent": _DEFAULT_USER_AGENT,
+        }
+        expected_headers.update(headers)
+
+        with HTTPConnectionPool(self.host, self.port, retries=False) as pool:
+            pool.request("GET", "/", headers=HTTPHeaderDict(headers))
+            assert expected_headers == parsed_headers
+
+    def test_ua_header_can_be_overwridden(self):
+        headers = {"uSeR-AgENt": "Allstate, like Geico"}
```
sethmlarson: Let's remove the reference to All State and Geico.

We should also have a test case for the User-Agent being set at the `ConnectionPool` level.

DonaCthulhuote:
Whelp, I'll figure that one out tomorrow.

DonaCthulhuote:
I think the mysteries of the app engine is a story for another day.

pquentin:
@DonaCthulhuote We do follow your progress with attention! And are happy to help out if you're stuck.

DonaCthulhuote:
Oh, just realized what that comment sounded like; shouldn't write things other people can see when I'm tired.  I simply thought it was funny that the context for my statements was disappearing.

And not stuck (yet); just discovering that writing code for multiple python versions is a skill in and of itself.

DonaCthulhuote:
I am once again certain all the tests will pass.  Prove me wrong, Travis!

DonaCthulhuote:
I'm certain the tests will all pass now.  Prove me wrong, travis!

pquentin:
Just a quick note to mention that the remaining failures here are not in your code! So if you think this is ready for review, then it is ready indeed :)

DonaCthulhuote:
Have at it, if you will.

pquentin:
@DonaCthulhuote Sorry I'm going to ask one more thing first -- can you please rebase your commit on top of master? We only merge pull requests that pass CI, and the only way to do that here is to rebase.

DonaCthulhuote:
I'm hoping those should still be green; my local for some reason decided it'd have trouble running tests after rebasing. If travis discovers something, I'll take a deeper look tonight.

pquentin:
@DonaCthulhuote we need to review this, sorry that we haven't be able to do it yet. You no longer need to rebase your commit on top of master since the tests pass and there are no conflicts!

src/urllib3/connection.py
```diff
@@ -406,6 +417,12 @@ def _match_hostname(cert, asserted_hostname):
         raise
 
 
+def _get_default_user_agent():
+    from . import __version__
```
PleasantMachine9, line 421: Isn't it better to import this at the top of the module? Does python optimize repeated imports like this somehow?

pquentin: I'm not sure about performance, but it would be better to import this at the top, right. Unfortunately we failed to review this promptly and this now needs to be rebased on top of master to fix the conflicts and CI.

src/urllib3/__init__.py
hodbn: Done in the original work.

test/with_dummyserver/test_socketlevel.py
hodbn: Done in the original work.

src/urllib3/request.py
hodbn: Done in the original work.

src/urllib3/connection.py
hodbn: Done.

hodbn:
I've rebased the work @DonaCthulhuote did on top of master.
Although I think a default user agent is the correct behavior, I think we should let users a way to restore old, no user agent behavior. I've added a new constant named ```urllib3.util.SUPPRESS_USER_AGENT``` to allow this.

I still need to figure out where and how to document the new user agent choosing logic (default and suppressing).

Regarding [@user234683's comment](https://github.com/urllib3/urllib3/issues/1296#issuecomment-466732006), I'm also leaning towards removing the version from the user agent. @sethmlarson @pquentin your thoughts?

DonaCthulhuote:
I'd use None for the suppress headers, if it's being added only to be removed, but I don't have a good justification other than that it feels cleaner in my head.

Thanks for picking this up; I got run over by work last week, and hadn't had a chance to take another look at this.

hodbn:
@DonaCthulhuote You’re right, it is cleaner and it was my first attempt. I’ve reverted it because it didn’t work for ```urllib3.make_request```. 

The same problem will probably rise for higher-level APIs (i.e. requests) that will expose user agent as an optional argument.

What do you think?

DonaCthulhuote:
I think I clearly missed that case, and that you've clearly put more thought into the problem then I have :) .  I'll defer to you and the maintainers on the matter.

pquentin:
I don't have a strong opinion here. requests sends `python-requests/2.23.0`, but I'm happy with just `urllib3` too.

@hodbn What's the problem with `make_request`?

hodbn:
My bad :upside_down_face: not ```make_request``` but ```make_headers```:

https://github.com/urllib3/urllib3/blob/4901ec7715fe3b8589e82807283d4ccb96eb7236/src/urllib3/util/request.py#L22-L29
Callers will not be able to suppress the default behavior. I guess there might be other callers that will have the same problem.
I'm not that happy with the constant I've introduced, but I couldn't find a cleaner way to get around this :grimacing:.

jalopezsilva:
@pquentin just FYI, I ran into a bug with this PR while merging my TLS in TLS changes on top of 1.25.10.

The bug is not really in urllib3. It seems botocore is encoding all request values as bytes by accident in python3. (https://github.com/boto/botocore/blob/develop/botocore/endpoint.py#L123). Now that we're accessing the request user agent before sending the request: _collection.py, ``get_item`` will complain that there are byte arrays in the container.

An easy repro of the error:

```
>>> value = ["key", "value".encode("utf-8")]
>>> ", ".join(value[1:])
```

I'm creating a PR on their side at the moment, will update with it here.

jalopezsilva:
PR is https://github.com/boto/botocore/pull/2123/

Additionally, we could ensure sure all items in ``_collection.py.HTTPHeaderDict`` are strings in botocore but I'll leave the decision up to you guys :)

pquentin:
Thanks for bringing that up and for the boto3 pull request!

I agree that this is a bug and that it should be fixed in urllib3 too.

Another option is to fix this PR to accept bytes and strings in the user agent header by converting to bytes before comparing the values, using something like `header.encode("ascii", "backslashreplace")`. I'd need to look at the code more to see what would be best.

sethmlarson:
@jalopezsilva Wanted to echo @pquentin's thanks for tracking down this issue. I agree that we should maintain the ability to handle headers as bytes gracefully.

PR #1752, opened 2019-11-21
Update RECENT_DATE dynamically

akhait:
We've stumbled upon this in [Fedora](https://src.fedoraproject.org/rpms/python-urllib3/pull-request/7) and thought this patch could be sent to upstream.
No tests fail and didn't fail already (run locally).

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -54,9 +54,9 @@ log = logging.getLogger(__name__)
 
 port_by_scheme = {"http": 80, "https": 443}
 
-# When it comes time to update this value as a part of regular maintenance
-# (ie test_recent_date is failing) update it to ~6 months before the current date.
-RECENT_DATE = datetime.date(2019, 1, 1)
+# This value will be updated dynamically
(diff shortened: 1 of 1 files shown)


pquentin:
@akhait Thank you for providing your patch upstream!

The reason we can't do this in urllib3 is that RECENT_DATE is used to detect systems where the clock is giving nonsensical results, such as 1970-01-01. I think updating the value dynamically makes sense for Fedora because you know the time is correct when building the package, but our code can also be vendored directly in other Python packages, and in those cases we don't want a dynamic value.

We agree that it's annoying, and there could be a better way. A few ideas have been mentioned in https://github.com/urllib3/urllib3/pull/1645. Maybe we should update RECENT_DATE as part of the release process, the same way we bump the version number? I don't know.

sethmlarson:
Closing and reopening to test commit statuses for Codecov

pquentin:
Unfortunately we can't merge this pull request as is, so I'll close it. We're still happy to consider other options here, unfortunately this one does not fit our constraints.

I'd like to thank you again for upstreaming your patch!

Issue #1753, opened 2019-11-22
urllib3 installation requires setuptools >= 36.2.5


Johnlisk:
when I install urllib3 by setup.py,I got this:
```
[root@localhost urllib3]# python2 setup.py install
/usr/lib64/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'project_urls'
  warnings.warn(msg)
/usr/lib64/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'python_requires'
  warnings.warn(msg)
error in urllib3 setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```
I use python==2.7.5 setuptools==41.6.0

pquentin:
Hello @Johnlisk! I'm sorry that you had trouble installing urllib3. Thank you for 1/ reporting your problem 2/ offering a fix! This is really appreciated.

However, the current syntax is valid and means that the `ipaddress` packages should only be installed with Python 2.7 for users that use `pip install urllib3[secure]`.

`python_version=='2.7'` is an [environment marker](https://www.python.org/dev/peps/pep-0508/), and such markers are accepted in extras_requires since setuptools v36.2.5 (see https://github.com/pypa/setuptools/issues/1087 for details).

Here are a few questions to better understand your problem:

 * Are you sure you're using the latest setuptools? What is the output of `python2 -c 'import setuptools; print setuptools.__version__'`?
 * Is this the full message you get? Only mentions on distutils, no mentions of setuptools?
 * Are you on RHEL ? What version?

Johnlisk:
that's the output, I jsut run `pip install setuptools --upgrade` to upgrade my setuptools to 41.6.0 before I run `python setup.py install`,
```
[root@localhost ~]# python2 -c 'import setuptools'
[root@localhost ~]# python2
Python 2.7.5 (default, Aug  7 2019, 00:51:29)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import setuptools
>>> print setuptools.__version__
0.9.8
>>>
```
the messages in the issue is messages all I got,nothing for setuptools
I'm on CentOS Linux release 7.6.1810 (Core)
pip version 19.1.1

pquentin:
[...]

PR #1754, opened 2019-11-22
Update setup.py

Johnlisk:
bug fixed:issue #1753 extras_require error in setup.py


pquentin:
@Johnlisk Hello, and thank you for your contribution! Unfortunately this fix is not valid, as explained in the issue that you helpfully opened. I'll close this PR, and we may open a new one when we better understand the issue you're facing.

Thank you!

PR #1759, opened 2019-11-25
Fix established connections are replaced by None values

jaimefrites:
When pool is in blocking mode, there is no
connections in pool, and getting connection
from pool is timed out.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -261,17 +261,22 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 )
             pass  # Oh well, we'll create a new connection then
 
-        # If this is a persistent connection, check if it got disconnected
-        if conn and is_connection_dropped(conn):
-            log.debug("Resetting dropped connection: %s", self.host)
-            conn.close()
-            if getattr(conn, "auto_open", 1) == 0:
-                # This is a proxied connection that has been mutated by
-                # httplib._tunnel() and cannot be reused (since it would
-                # attempt to bypass the proxy)
-                conn = None
+        try:
+            # If this is a persistent connection, check if it got disconnected
+            if conn and is_connection_dropped(conn):
+                log.debug("Resetting dropped connection: %s", self.host)
+                conn.close()
+                if getattr(conn, "auto_open", 1) == 0:
+                    # This is a proxied connection that has been mutated by
+                    # httplib._tunnel() and cannot be reused (since it would
+                    # attempt to bypass the proxy)
+                    conn = None
 
-        return conn or self._new_conn()
+            return conn or self._new_conn()
+        except Exception:
+            # Return connection to the pool if something went wrong
+            self._put_conn(conn)
+            raise
 
     def _put_conn(self, conn):
         """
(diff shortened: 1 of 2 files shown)


pquentin:
Hello @jaimefrites and thank you for your contribution!

Can you share more about the exact scenario this helps with? This would help using a more precise exception.

Also, it looks like your test passes even without the `self._put_conn(conn)` call. Do you know why?

Thanks!

jaimefrites:
_Here is a more verbose description of the scenario that I’m trying to handle._

There is a pool of size N (block=True)
There are so many requests to the pool, so some requests don’t get a connection in pool_timeout seconds
None value is put to the pool when request can’t get a connection
If N requests can’t get connections, then the pool will be filled with None values completely
Established connections are thrown out, because None values occupied there places in the pool

This degrades application performance dramatically, because application has to establish new connections continually instead of reuse old connections, though there is no good reason to do that.

_About put_conn on exception in get_conn_
                                                                            
I have to put connection back to pool in  HTTPConnectionPool._get_conn method because of the way I’m trying to fix the issue – my changes to urlopen method assume that connection shouldn’t be put to pool when _get_conn method raises exception. But _get_conn method do extra things besides getting connection from pool. If an exception raised after a connection was got from a pool, then urlopen never put the connection to the pool (in other words pool size decreased by 1).

I have added another one test to cover this problem.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -1115,3 +1116,49 @@ def test_pool_size_redirect(self):
...
+                    time.sleep(0)
+                with pytest.raises(EmptyPoolError):
+                    consume_connection(pool)
+            finally:
+                block_send_event.set()
+                consume_connection_thread.join()
+
+            assert pool.pool.get(True, SHORT_TIMEOUT) is not None
+
+    def test_return_conn_to_pool_when_get_conn_failed(self):
```
sethmlarson: Maybe I'm not understanding this test, why would we want to return a connection to the pool if it raises an exception in the constructor?

src/urllib3/connectionpool.py
```diff
@@ -261,17 +261,22 @@ def _get_conn(self, timeout=None):
...
+                log.debug("Resetting dropped connection: %s", self.host)
+                conn.close()
+                if getattr(conn, "auto_open", 1) == 0:
+                    # This is a proxied connection that has been mutated by
+                    # httplib._tunnel() and cannot be reused (since it would
+                    # attempt to bypass the proxy)
+                    conn = None
+
+            return conn or self._new_conn()
+        except Exception:
```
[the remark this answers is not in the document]
jaimefrites: We are expecting any exception which self.ConnectionCls can raise.

test/with_dummyserver/test_connectionpool.py
jaimefrites: If we don't return the connection to the pool, we will get EmptyPoolError on every next call of self._get_conn.

jaimefrites:
https://github.com/urllib3/urllib3/pull/1866 is better approach to solve the issue

Issue #1755, opened 2019-11-22
3 test failures in git HEAD: TestHTTPProxyManager.test_cross_host_redirect, TestHTTPProxyManager.test_cross_protocol_redirect, TestIPv6HTTPProxyManager.test_basic_ipv6_proxy


mgorny:
I suspect it doesn't like something about my system but I really don't know what it is it doesn't like.

```
$ nox
nox > Running session test-2.7
nox > Creating virtual environment (virtualenv) using python2.7 in .nox/test-2-7
nox > pip install -r dev-requirements.txt
nox > pip install .[socks,secure,brotli]
nox > pip --version
pip 19.3.1 from /tmp/urllib3/.nox/test-2-7/lib/python2.7/site-packages/pip (python 2.7)
nox > python --version
Python 2.7.17
nox > python -c import struct; print(struct.calcsize('P') * 8)
64
nox > python -m OpenSSL.debug
pyOpenSSL: 19.1.0
cryptography: 2.8
cffi: 1.13.2
cryptography's compiled against OpenSSL: OpenSSL 1.1.1d  10 Sep 2019
cryptography's linked OpenSSL: OpenSSL 1.1.1d  10 Sep 2019
Pythons's OpenSSL: OpenSSL 1.1.1d  10 Sep 2019
Python executable: /tmp/urllib3/.nox/test-2-7/bin/python
Python version: 2.7.17 (default, Nov 11 2019, 18:48:18) 
[GCC 9.2.0]
Platform: linux2
sys.path: ['', '/tmp/urllib3/.nox/test-2-7/lib64/python27.zip', '/tmp/urllib3/.nox/test-2-7/lib64/python2.7', '/tmp/urllib3/.nox/test-2-7/lib64/python2.7/plat-linux2', '/tmp/urllib3/.nox/test-2-7/lib64/python2.7/lib-tk', '/tmp/urllib3/.nox/test-2-7/lib64/python2.7/lib-old', '/tmp/urllib3/.nox/test-2-7/lib64/python2.7/lib-dynload', '/usr/lib64/python2.7', '/usr/lib/python2.7', '/usr/lib64/python2.7/lib-tk', '/tmp/urllib3/.nox/test-2-7/lib/python2.7/site-packages']
nox > coverage run --parallel-mode -m pytest -r a --tb=native --no-success-flaky-report test/
=============================================================== test session starts ===============================================================
[...]

PR #1762, opened 2019-11-30
Skip IPv6 test class when IPv6 is not available

pquentin:
Should hopefully fix the first failure mentioned in https://github.com/urllib3/urllib3/issues/1755 by @mgorny.


pquentin:
Thanks!

PR #1763, opened 2019-11-30
Remove unneeded test timeout

pquentin:
Should hopefully fix the last two failures mentioned in #1755 by @mgorny.

It's not fun when a test suite hangs, but it never happened to me when working on urllib3, while having false positives due to timeouts happened to me a *lot*. So I think we should optimize the latter case, and revisit later if we start to see hanged tests.

### DIFF
--- a/test/contrib/test_securetransport.py
+++ b/test/contrib/test_securetransport.py
@@ -10,6 +10,8 @@ try:
 except ImportError:
     pass
 
+pytestmark = pytest.mark.skip()
+
 
 def setup_module():
     try:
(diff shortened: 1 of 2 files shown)


test/contrib/test_securetransport.py
```diff
@@ -10,6 +10,8 @@
 except ImportError:
     pass
 
+pytestmark = pytest.mark.skip()
```
sethmlarson: ?

pquentin: No. :) It doesn't work on macOS 10.15, so I use to run tests locally, and I forgot to remove it. Thanks!

pquentin:
@sethmlarson Thank you for the review. ❤️

PR #1774, opened 2020-01-06
Drop Python 3.4

pquentin:
Python 3.4 reached its end-of-life on 2019-03-18 and represents <4% of our daily downloads.

This is https://github.com/urllib3/urllib3/pull/1686 rebased on top of master. It got reverted because we were waiting for a good release to happen.

This should also fix the CI error noticed in https://github.com/urllib3/urllib3/pull/1750.


pquentin:
Thank you for the review!

PR #1776, opened 2020-01-07
Add explicit support for Python 3.9

sethmlarson:
Start testing on 3.9, add testing on Python 3.8 where there is finally support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,10 @@ matrix:
       env: NOX_SESSION=test-3.6
     - python: 3.7
       env: NOX_SESSION=test-3.7
-    - python: 3.8-dev
+    - python: 3.8
       env: NOX_SESSION=test-3.8
+    - python: 3.9-dev
+      env: NOX_SESSION=test-3.9
     - python: pypy2.7-6.0
       env: NOX_SESSION=test-pypy
     - python: pypy3.5-6.0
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -46,8 +46,10 @@ matrix:
       env: NOX_SESSION=test-3.6
     - python: 3.7
       env: NOX_SESSION=test-3.7
-    - python: 3.8-dev
+    - python: 3.8
       env: NOX_SESSION=test-3.8
+    - python: 3.9-dev
```
pquentin: There's no 3.9-dev on Travis, but there's a "python-nightly" for both 16.04 and 18.04. Maybe that's 3.9-dev?

appveyor.yml
```diff
@@ -27,6 +27,11 @@ environment:
       PYTHON_ARCH: "64"
       NOX_SESSION: "test-3.7"
 
+    - PYTHON: "C:\\Python38-x64"
+      PYTHON_VERSION: "3.8.x"
```
pquentin: And this one does not work yet because Tornado 6 does not support the proactor event loop which is the default on Windows + 3.8 so we need to configure the event loop differently here.

pquentin: Opened https://github.com/urllib3/urllib3/pull/1777 for this one!

PR #1777, opened 2020-01-07
Fix tests on Python 3.8 + Windows

pquentin:
I wrote the same fix for hip in https://github.com/python-trio/hip/commit/7a31c47707da98ac0ab69f9f45863b13fb0f54b2


pquentin:
Thank you for the review! 👍

PR #1779, opened 2020-01-10
Generate NO_SAN certs using trustme

pquentin:
Generating server certs using trustme requires using those certs in the tests themselves. The simplest option here is to use a fixture.

I've started with @agronholm's work in the asynctest branch of hip: https://github.com/python-trio/hip/commit/e35a4bfa7b4b6fc3a41ab35448bc515ad2216be7, then simplified it because we don't care about async here, and then added support for communicating back more things that only the URL. So not much remains of his work, but this has the advantage of being close to the way hip is going to do this, which will allow us to continue sharing fixes between the two projects.

The idea is to eventually switch all HTTPS tests to this new way of configuring TLS certs in Tornado, so that we remove all blobs from git and get the SecureTransport tests to run on macOS 10.15+.


pquentin:
I've split this pull request into two commits because it was doing two different things, hopefully that should make the review easier.

PR #1784, opened 2020-01-15
Generate IPv6 certificates using trustme

pquentin:
Yet another part of https://github.com/urllib3/urllib3/issues/1705

Please consider reviewing and including those changes commit by commit, as there's a refactoring that's not related to the IPv6 changes.


pquentin:
You mean skipping the test from the fixture if the system does not support IPv6? Sure, that's possible, even though each IPv6 fixture is specific to a test right now. I'll add a commit to do that.

Actually I'm starting to wonder if fixtures are the right solution here. They're far away from the tested code, and are going to add a lot of noise for classes that have many tests. I'm considering converting them into class methods that would live alongside the tests, not in another file. I'll evaluate that when working on the main HTTPS test class.

pquentin:
@sethmlarson Is that what you meant regarding the HAS_IPv6 checks?

sethmlarson:
Yeah that looks great!

PR #1791, opened 2020-01-24
Generate client certs dynamically using trustme

pquentin:
Another part of #1705 

Not using a fixture and pytest is slightly more painful as it requires more work for Python 2 compatibility and I can't use a context manager, but it allows to keep the certificate generation code near the tests code.


pquentin:
Thank you for the review! I continue to appreciate the work you're putting here.

Issue #1790, opened 2020-01-24
Issue with Parsing URIs - Breaks Security Tools when testing for Path Traversal 


mazen160:
Hi all,

I have noticed an odd behaviour in requests module, which uses urllib3. I inspected the root cause via regression testing and I found that the root cause of the issue is a change that was introduced in urlib3 

```python
import requests
requests.get("http://127.0.0.1/../../../../doing/certain/check")
```
Which should typically send to the following path: `/../../../doing/certain/check`. Instead, it's requesting `/doing/certain/check` and modifying the specified input.

This breaks a lot of security tools that are performing fuzzing for path traversal vulnerabilities in web-applications.

The only solution for current tools is to revert to `urlib3==1.24.3`.

This may be related: https://github.com/urllib3/urllib3/commit/0aa3e24fcd75f1bb59ab159e9f8adb44055b2271#diff-26a37c674beb46c0ae8f77d76946e054

Can we work on fixing this issue?

sethmlarson:
Yeah I agree we should keep the previous behavior. My thoughts are that we should only do path normalization on redirects. Are you able to potentially provide a PR to fix this issue?

mazen160:
Hi @sethmlarson 

Thanks for your response. 

I have spent more time today morning to find the root cause. The normalization happens at `urlib3/util/url.py`. It was introduced in this commit https://github.com/urllib3/urllib3/commit/5b047b645f5f93900d5e2fc31230848c25eb1f5f#diff-c289049c6f3c9d7397378a8be7e700c6. 

I have made a PR that can be found here: https://github.com/urllib3/urllib3/pull/1792

Unfortunately making a specific use-case where to normalize URI when doing a redirect sounds difficult. I'm not familiar with urlib3 codebase, but this PR should be a fix for the base behaviour. 

Thanks again!

pquentin:
I'm sorry that urllib3 broke your workflow, and I would also like to apologize for [ruining your weekend](https://twitter.com/mazen160/status/1220738021626208256).

I'm not an Information Security Specialist like you are and am not the lead maintainer, but I'd like to step back for a moment. It seems to me that the fact that we were able to fix this vulnerability for all our users is actually a good thing! And I'm sure you'll appreciate that lowering our security in the name of security is quite ironic.

Are there other ways to check for this vulnerability? Maybe using a lower-level tool like [http.client](https://docs.python.org/3/library/http.client.html) would be more appropriate here?

[...]

PR #1792, opened 2020-01-25
Fix - Issue with Parsing URIs - Breaks Security Tools when testing for Path Traversal  

mazen160:
This PR fixes https://github.com/urllib3/urllib3/issues/1790

The code introduced in https://github.com/urllib3/urllib3/commit/5b047b645f5f93900d5e2fc31230848c25eb1f5f#diff-c289049c6f3c9d7397378a8be7e700c6 brought this unexpected functionality. The PR removes the responsible function and its uses.


sethmlarson:
In the mentioned issue we've decided that `ConnectionPool` not normalizing the request target is enough to resolve this issue.

PR #1793, opened 2020-01-26
Generate client password-protected cert with trustme 

pquentin:
Part of #1705 

This also includes a commit to move some cert generation code in the class that uses it, see https://github.com/urllib3/urllib3/pull/1791 for the rationale.


pquentin:
@sethmlarson Thanks for the review! Are you saying that we could store all the code that generate the certificates in say dummyserver/testcase.py?

PR #1796, opened 2020-02-03
Expose peer certificate

jschlyter:
In order to do more advanced fingerprinting on certificates, it would be most useful if the peer certificate could be exposed in the connection and in responses.

If possible, I'd like to expose the entire peer certificate chain for things like DANE/TLSA validation, but I'm not sure where do find this information.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -294,6 +294,7 @@ class VerifiedHTTPSConnection(HTTPSConnection):
         self.assert_fingerprint = assert_fingerprint
         self.ca_certs = ca_certs and os.path.expanduser(ca_certs)
         self.ca_cert_dir = ca_cert_dir and os.path.expanduser(ca_cert_dir)
+        self.peer_cert = None
 
     def connect(self):
         # Add certificate verification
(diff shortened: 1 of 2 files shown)


jschlyter:
Perhaps @sethmlarson can provide some guidance on where to find the entire peer certificate chain?

pquentin:
We can't get the entire peer certificate chain because it's not possible in the standard library, see https://bugs.python.org/issue18233. And we probably don't want to allow this in the PyOpenSSL and SecureTransport contrib backends if it's not possible with the standard library?

jschlyter:
> We can't get the entire peer certificate chain because it's not possible in the standard library, see https://bugs.python.org/issue18233. And we probably don't want to allow this in the PyOpenSSL and SecureTransport contrib backends if it's not possible with the standard library?

I agree that certificate chain support be added later once/if support by the standard library.

sethmlarson:
This might be a good use-case for sub-classing `HTTPConnection` and adding this additional step to `connect()`.  The library allows subclassing and using your custom class in a pool via `ConnectionPool.ConnectionCls`.

sethmlarson:
Going to close this PR per above comment.

jschlyter:
Perhaps now possible as described in https://github.com/urllib3/urllib3/issues/2438

PR #1800, opened 2020-02-13
Raise ValueError if method contains control characters

sethmlarson:
Protects against [BPO-39603](https://bugs.python.org/issue39603) which was reported to CPython and Requests. The risk of this being exploited is minimal as attackers will rarely have direct control of the request method, but in that case header injection is possible in `http.client`, `urllib3` and `requests`.

This change makes that not possible, unfortunately there's not really an exception that fits the bill for this error, but hopefully this situation is rare enough that we don't have to care.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -59,7 +59,7 @@ port_by_scheme = {"http": 80, "https": 443}
 # (ie test_recent_date is failing) update it to ~6 months before the current date.
 RECENT_DATE = datetime.date(2019, 1, 1)
 
-_CONTAINS_CONTROL_CHAR_RE = re.compile("[\x00-\x20\x7f]")
+_CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 
 
 class DummyConnection(object):
@@ -191,7 +191,7 @@ class HTTPConnection(_HTTPConnection, object):
         match = _CONTAINS_CONTROL_CHAR_RE.search(method)
         if match:
             raise ValueError(
-                "Method cannot contain control characters. %r (found at least %r)"
+                "Method cannot contain non-token characters %r (found at least %r)"
                 % (method, match.group())
             )
 
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+import re
 import datetime
 import logging
 import os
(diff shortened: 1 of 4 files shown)


test/with_dummyserver/test_poolmanager.py
```diff
@@ -312,6 +312,12 @@ def test_http_with_ca_cert_dir(self):
             r = http.request("GET", "http://%s:%s/" % (self.host, self.port))
             assert r.status == 200
 
+    @pytest.mark.parametrize("char", [" ", "\r", "\n", "\x00"])
+    def test_invalid_method_not_allowed(self, char):
+        with pytest.raises(ValueError):
+            with PoolManager() as http:
+                http.request("GET" + char, "/")
+
```
pquentin: For what it's worth, I think I would have left out this test and considered that the connection pool test is enough

src/urllib3/connection.py
```diff
@@ -58,6 +59,8 @@ class ConnectionError(Exception):
 # (ie test_recent_date is failing) update it to ~6 months before the current date.
 RECENT_DATE = datetime.date(2019, 1, 1)
 
+_CONTAINS_CONTROL_CHAR_RE = re.compile("[\x00-\x20\x7f]")
```
pquentin: RFC 2730 mentions that the method is a token, and that a token is a number of VCHAR (US-ASCII visible characters):

```
     method         = token
     token          = 1*tchar

     tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                    / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                    / DIGIT / ALPHA
                    ; any VCHAR, except delimiters
```

[h11 defines token](https://github.com/python-hyper/h11/blob/68e32db/h11/_abnf.py) as `token = r"[-!#$%&'*+.^_`|~0-9a-zA-Z]+"`, so maybe we can steal that regex? It would have prevent the attack mentioned in the bug.

test/with_dummyserver/test_poolmanager.py
sethmlarson: That's true, I'll do that. :)

src/urllib3/connection.py
sethmlarson: Good idea, I'll copy h11.

sethmlarson:
Yeah it's because `putrequest()` is now a member on `HTTPConnection` but the documentation in the Python source code is apparently not Sphinx-compatible so it's throwing an error. Tried to figure out how to make autosphinx skip a single member but to no avail yet. :/

pquentin:
The remaining failure is a strange macOS Python 2.7 bug that also happens on master. It claims that it does not find the coverage program. Investigating...

pquentin:
https://github.com/urllib3/urllib3/pull/1802 should fix the issue we're seeing here

sethmlarson:
Cycling the build, thanks for the fix in #1802!

PR #1801, opened 2020-02-17
Remove deprecated and unused test arguments from setup.py

jdufresne:
Since setuptools v41.5.0 (27 Oct 2019), the 'test' command and its
arguments are formally deprecated and should not be used. Tests are
still easy to run using nox, as before.


sethmlarson:
Thanks!

PR #1804, opened 2020-02-19
Allow the CA certificate data to be passed as a string.

jeamland:
There are some circumstances where we may want to pass the PEM-encoded certificates directly to the SSLContext rather than passing a file or directory. This just plums `SSLContext.load_verify_locations()`'s `cadata` parameter through to places where it can be reached.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -282,5 +282,8 @@ In chronological order:
   * Remove a spurious TypeError from the exception chain inside
     HTTPConnectionPool._make_request(), also for BaseExceptions.
 
+* Benno Rice <benno@jeamland.net>
+  * Allow cadata parameter to be passed to underlying ``SSLContext.load_verify_locations()``.
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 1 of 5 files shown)


sigmavirus24:
Doesn't urllib3 allow users to pass their own SSLContext?

jeamland:
It does but that's not what requests appears to do. Do you feel that it should?

sigmavirus24:
I think you can define your own `Adapter` class in Requests that will allow you to pass down to urllib3 a custom `SSLContext`. That would give you the functionality you want as far as I understand, especially since it sounds like you control the code calling Requests.

jeamland:
Actually at the moment you can't even import requests as it calls `certifi.where()` unconditionally. This is part of an effort to remove that in a way that makes requests "just work".

sigmavirus24:
> Actually at the moment you can't even import requests as it calls `certifi.where()` unconditionally. This is part of an effort to remove that in a way that makes requests "just work".

I understand that `PyOxidizer` doesn't have a file-system like thing but does that mean `__file__` or `os.path.join` are broken? That was not made remotely clear on certifi

jeamland:
PyOxidizer does not set `__file__` at all. `os.path.join` would work for accessing "normal" files but using it to access resources by pivoting off the filesystem path of a module will never work.

sigmavirus24:
This should make it easier to use cadata from requests though. I'll let one of the _real_ maintainers here approve/provide feedback but otherwise this looks good to me

Issue #1799, opened 2020-02-08
Rename VerifiedHTTPSConnection to HTTPSConnection


sethmlarson:
The current `HTTPSConnection` object is broken and doesn't serve much purpose. We should use the current `VerifiedHTTPSConnection` object and rebind the name to preserve backwards compatibility.

kobayashi:
Hello folks,

Do I have chance to work for this?

sethmlarson:
@kobayashi Yes please! Let me know if you need any assistance. :)

kobayashi:
Thx @sethmlarson , I made the first PR to fix this.

PR #1805, opened 2020-02-24
Rename VerifiedHTTPSConnection

kobayashi:
Fix: #1799 
Renamed `VerifiedHTTPSConnection` to `HTTPSConnection`.  `VerifiedHTTPSConnection` is backported.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -237,7 +237,12 @@ class HTTPConnection(_HTTPConnection, object):
 class HTTPSConnection(HTTPConnection):
     default_port = port_by_scheme["https"]
 
+    cert_reqs = None
+    ca_certs = None
+    ca_cert_dir = None
+    ca_cert_data = None
     ssl_version = None
+    assert_fingerprint = None
 
     def __init__(
         self,
@@ -264,20 +269,6 @@ class HTTPSConnection(HTTPConnection):
         # Required property for Google AppEngine 1.9.0 which otherwise causes
         # HTTPS requests to go out as HTTP. (See Issue #356)
         self._protocol = "https"
-
-
-class VerifiedHTTPSConnection(HTTPSConnection):
-    """
-    Based on httplib.HTTPSConnection but wraps the socket with
-    SSL certification.
-    """
-
-    cert_reqs = None
-    ca_certs = None
-    ca_cert_dir = None
-    ca_cert_data = None
-    ssl_version = None
-    assert_fingerprint = None
 
     def set_cert(
         self,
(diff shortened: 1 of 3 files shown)


src/urllib3/connection.py
```diff
@@ -410,6 +401,10 @@ def connect(self):
         )
 
 
+class VerifiedHTTPSConnection(HTTPSConnection):
+    pass
+
```
dopstar: I think that on the lines 423-428:
```python
if ssl:
    # Make a copy for testing.
    UnverifiedHTTPSConnection = HTTPSConnection
    HTTPSConnection = VerifiedHTTPSConnection
else:
    HTTPSConnection = DummyConnection
```

We should consider  the following:
- The `HTTPSConnection` class should not be shadowed by reassignment so that it is exported as is (im not sure about the use case of the `DummyConnection` in the else statement)
- if we are keeping `UnverifiedHTTPSConnection = HTTPSConnection` then we should also have `VerifiedHTTPSConnection = HTTPSConnection` and then remove `VerifiedHTTPSConnection` class definition code.
  -  Alternatively, a `UnverifiedHTTPSConnection` class might need to be defined just like how `VerifiedHTTPSConnection` is defined.

kobayashi: Thanks for your review! Overlooked these lines. Could we delete the if statement? I like the third option you mentioned, so defining `UnverifiedHTTPSConnection` like `VerifiedHTTPSConnection` is simpler IMO. 
I am not sure the purpose of `DummyConnection`.
Let me know any thoughts.

kobayashi: `DummyConnection` seems to be used only for here.
https://github.com/urllib3/urllib3/blob/75d271954b5d662d0567e95b054475e208619dff/src/urllib3/connectionpool.py#L962-L965

sethmlarson: I'm thinking we do something like:

```
if ssl:
    HTTPSConnection = DummyConnection

VerifiedHTTPSConnection = HTTPSConnection
```

so we still get the nice error if the user is running Python without an `ssl` module and always get the name rebound to `VerifiedHTTPSConnection`. :)

Since the `UnverifiedHTTPSConnection` class was totally borked for the past ~6 months and we only received one issue on it I think we're good to remove it in the next release. :tada: Just need to document in the changelog that there's only one HTTPSConnection object now. :)

kobayashi: Thanks @sethmlarson, Let me check the if statement. We assign `DummyConnection` to `HTTPSConnection` if not ssl, right? So the if statement is like this?

```
if not ssl:
    HTTPSConnection = DummyConnection

VerifiedHTTPSConnection = HTTPSConnection
```

sethmlarson: You're absolutely right haha, nice catch. :)

kobayashi: Thank you! modified the statement and added to CHANGES.rst

pquentin:
Thanks! I think we're going to need a `  # NOQA` comment to silence the F811 flake8 warnings: https://travis-ci.org/urllib3/urllib3/jobs/656104926

sethmlarson:
I've made that change, thanks @pquentin for identifying. Will merge as soon as CI passes. :tada:

sethmlarson:
Thanks so much @kobayashi for this change! :rocket:

PR #1806, opened 2020-02-25
Add support for TLS in TLS for HTTPS proxies.

jalopezsilva:
This PR aims to add support for HTTPS proxies by adding a class to easily provide TLS in TLS support.  TLS within TLS is not supported easily within the ssl library. The SSLSocket actually takes over the existing socket (https://github.com/python/cpython/blob/master/Lib/ssl.py#L999-L1006)
instead of wrapping it entirely. The only way to support to TLS within TLS is with the ``SSLContext.wrap_bio`` methods.

The first commit introduces a class called SSLTransport which wraps a socket in TLS using the provided SSL context.  I'm currently in the process of adding unit tests and integration tests but it would be useful if we can merge #1679 for that.

Once I can fully integrate this within ``urlib3`` I'll perform some production testing on my own to validate the changes work as expected.

### DIFF
--- a/src/urllib3/contrib/ssl.py
+++ b/src/urllib3/contrib/ssl.py
@@ -1,0 +1,197 @@
+import ssl
+import socket
+import io
+
+SSL_BLOCKSIZE = 16384
+
+
+class SSLTransport:
+    """
+    The SSLTransport wraps an existing socket and establishes an SSL connection.
+
+    Contrary to Python's implementation of SSLSocket, it allows you to chain
+    multiple TLS connections together. It's particularly useful if you need to
+    implement TLS within TLS.
+
+    The class supports most of the socket API operations.
+    """
+
+    def __init__(
+        self, socket, ssl_context, suppress_ragged_eofs=True, server_hostname=None
+    ):
+        """
+        Create an SSLTransport around socket using the provided ssl_context.
+        """
+        self.incoming = ssl.MemoryBIO()
+        self.outgoing = ssl.MemoryBIO()
+
+        self.suppress_ragged_eofs = suppress_ragged_eofs
+        self.socket = socket
+
+        self.sslobj = ssl_context.wrap_bio(
+            self.incoming, self.outgoing, server_hostname=server_hostname
+        )
+
+        # Perform initial handshake.
+        self._ssl_io_loop(self.sslobj.do_handshake)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *_):
+        self.close()
+
+    def fileno(self):
+        return self.socket.fileno()
+
+    def read(self, len=1024, buffer=None):
+        return self._wrap_ssl_read(len, buffer)
+
+    def recv(self, len=1024, flags=0):
+        if flags != 0:
+            raise ValueError("non-zero flags not allowed in calls to recv")
+        return self._wrap_ssl_read(len)
+
+    def recv_into(self, buffer, nbytes=None, flags=0):
+        if flags != 0:
+            raise ValueError("non-zero flags not allowed in calls to recv_into")
+        if buffer and (nbytes is None):
+            nbytes = len(buffer)
+        elif nbytes is None:
+            nbytes = 1024
+        return self.read(nbytes, buffer)
+
+    def sendall(self, data, flags=0):
+        if flags != 0:
+            raise ValueError("non-zero flags not allowed in calls to sendall")
+        count = 0
+        with memoryview(data) as view, view.cast("B") as byte_view:
+            amount = len(byte_view)
+            while count < amount:
+                v = self.send(byte_view[count:])
+                count += v
+
+    def send(self, data, flags=0):
+        if flags != 0:
+            raise ValueError("non-zero flags not allowed in calls to send")
+        response = self._ssl_io_loop(self.sslobj.write, data)
+        return response
+
+    def makefile(
+        self, mode="r", buffering=None, encoding=None, errors=None, newline=None
+    ):
+        """
+        Python's httpclient uses makefile and buffered io when reading HTTP
+        messages and we need to support it.
+
+        This is unfortunately a copy and paste of socket.py makefile with small
+        changes to point to the socket directly.
+        """
+        if not set(mode) <= {"r", "w", "b"}:
+            raise ValueError("invalid mode %r (only r, w, b allowed)" % (mode,))
+
+        writing = "w" in mode
+        reading = "r" in mode or not writing
+        assert reading or writing
+        binary = "b" in mode
+        rawmode = ""
+        if reading:
+            rawmode += "r"
+        if writing:
+            rawmode += "w"
+        raw = socket.SocketIO(self, rawmode)
+        self.socket._io_refs += 1
+        if buffering is None:
+            buffering = -1
+        if buffering < 0:
+            buffering = io.DEFAULT_BUFFER_SIZE
+        if buffering == 0:
+            if not binary:
+                raise ValueError("unbuffered streams must be binary")
+            return raw
+        if reading and writing:
+            buffer = io.BufferedRWPair(raw, raw, buffering)
+        elif reading:
+            buffer = io.BufferedReader(raw, buffering)
+        else:
+            assert writing
+            buffer = io.BufferedWriter(raw, buffering)
+        if binary:
+            return buffer
+        text = io.TextIOWrapper(buffer, encoding, errors, newline)
+        text.mode = mode
+        return text
+
+    def unwrap(self):
+        self._ssl_io_loop(self.sslobj.unwrap)
+
+    def close(self):
+        self.socket.close()
+
+    def getpeercert(self, binary_form=False):
+        return self.sslobj.getpeercert(binary_form)
+
+    def version(self):
+        return self.sslobj.version()
+
+    def cipher(self):
+        return self.sslobj.cipher()
+
+    def selected_alpn_protocol(self):
+        return self.sslobj.selected_alpn_protocol()
+
+    def selected_npn_protocol(self):
+        return self.sslobj.selected_npn_protocol()
+
+    def shared_ciphers(self):
+        return self.sslobj.shared_ciphers()
+
+    def compression(self):
+        return self.sslobj.compression()
+
+    def settimeout(self, value):
+        self.socket.settimeout(value)
+
+    def gettimeout(self):
+        return self.socket.gettimeout()
+
+    def _decref_socketios(self):
+        self.socket._decref_socketios()
+
+    def _wrap_ssl_read(self, len, buffer=None):
+        response = None
+        try:
+            response = self._ssl_io_loop(self.sslobj.read, len, buffer)
+        except ssl.SSLError as e:
+            if e.errno == ssl.SSL_ERROR_EOF and self.suppress_ragged_eofs:
+                response = 0  # eof, return 0.
+            else:
+                raise
+        return response
+
+    def _ssl_io_loop(self, func, *args):
+        """ Performs an I/O loop between incoming/outgoing and the socket."""
+        should_loop = True
+
+        while should_loop:
+            errno = None
+            try:
+                ret = func(*args)
+            except ssl.SSLError as e:
+                if e.errno not in (ssl.SSL_ERROR_WANT_READ, ssl.SSL_ERROR_WANT_WRITE):
+                    # WANT_READ, and WANT_WRITE are expected, others are not.
+                    raise e
+                errno = e.errno
+
+            buf = self.outgoing.read()
+            self.socket.sendall(buf)
+
+            if errno is None:
+                should_loop = False
+            elif errno == ssl.SSL_ERROR_WANT_READ:
+                buf = self.socket.recv(SSL_BLOCKSIZE)
+                if buf:
+                    self.incoming.write(buf)
+                else:
+                    self.incoming.write_eof()
+        return ret
--- a/test/contrib/test_ssltransport.py
+++ b/test/contrib/test_ssltransport.py
@@ -1,0 +1,470 @@
+from dummyserver.testcase import SocketDummyServerTestCase, consume_socket
+from dummyserver.server import (
+    DEFAULT_CERTS,
+    DEFAULT_CA,
+)
+
+from urllib3.contrib.ssl import SSLTransport
+
+import select
+import pytest
+import socket
+import ssl
+import sys
+import platform
+
+
+# consume_socket can iterate forever, we add timeouts to prevent halting.
+PER_TEST_TIMEOUT = 60
+
+
+def server_client_ssl_contexts():
+    if hasattr(ssl, "PROTOCOL_TLS_SERVER"):
+        server_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+    else:
+        # python 2.7 and 3.5 workaround.
+        # PROTOCOL_TLS_SERVER was added in 3.6
+        server_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+    server_context.load_cert_chain(DEFAULT_CERTS["certfile"], DEFAULT_CERTS["keyfile"])
+
+    if hasattr(ssl, "PROTOCOL_TLS_CLIENT"):
+        client_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    else:
+        # python 2.7 and 3.5 workaround.
+        # PROTOCOL_TLS_SERVER was added in 3.6
+        client_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
+        client_context.verify_mode = ssl.CERT_REQUIRED
+        client_context.check_hostname = True
+
+    client_context.load_verify_locations(DEFAULT_CA)
+    return server_context, client_context
+
+
+def sample_request(binary=True):
+    request = (
+        b"GET http://www.testing.com/ HTTP/1.1\r\n"
+        b"Host: www.testing.com\r\n"
+        b"User-Agent: awesome-test\r\n"
+        b"\r\n"
+    )
+    return request if binary else request.decode("utf-8")
+
+
+def validate_request(provided_request, binary=True):
+    assert provided_request is not None
+    expected_request = sample_request(binary)
+    assert provided_request == expected_request
+
+
+def sample_response(binary=True):
+    response = b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n"
+    return response if binary else response.decode("utf-8")
+
+
+def validate_response(provided_response, binary=True):
+    assert provided_response is not None
+    expected_response = sample_response(binary)
+    assert provided_response == expected_response
+
+
+def validate_peercert(ssl_socket):
+
+    binary_cert = ssl_socket.getpeercert(binary_form=True)
+    assert type(binary_cert) == bytes
+    assert len(binary_cert) > 0
+
+    cert = ssl_socket.getpeercert()
+    assert type(cert) == dict
+    assert "serialNumber" in cert
+    assert cert["serialNumber"] != ""
+
+
+@pytest.mark.skipif(sys.version_info < (3, 5), reason="requires python3.5 or higher")
+class SingleTLSLayerTestCase(SocketDummyServerTestCase):
+    """
+    Uses the SocketDummyServer to validate a single TLS layer can be
+    established through the SSLTransport.
+    """
+
+    @classmethod
+    def setup_class(cls):
+        cls.server_context, cls.client_context = server_client_ssl_contexts()
+
+    def start_dummy_server(self, handler=None):
+        def socket_handler(listener):
+            sock = listener.accept()[0]
+            with self.server_context.wrap_socket(sock, server_side=True) as ssock:
+                request = consume_socket(ssock)
+                validate_request(request)
+                ssock.send(sample_response())
+
+        chosen_handler = handler if handler else socket_handler
+        self._start_server(chosen_handler)
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_start_closed_socket(self):
+        """ Errors generated from an unconnected socket should bubble up."""
+        sock = socket.socket(socket.AF_INET)
+        context = ssl.create_default_context()
+        sock.close()
+        with pytest.raises(OSError):
+            SSLTransport(sock, context)
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_close_after_handshake(self):
+        """ Socket errors should be bubbled up """
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            ssock.close()
+            with pytest.raises(OSError):
+                ssock.send(b"blaaargh")
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_wrap_existing_socket(self):
+        """ Validates a single TLS layer can be established.  """
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            assert ssock.version() is not None
+            ssock.send(sample_request())
+            response = consume_socket(ssock)
+            validate_response(response)
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_unbuffered_text_makefile(self):
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            with pytest.raises(ValueError):
+                ssock.makefile("r", buffering=0)
+            ssock.send(sample_request())
+            response = consume_socket(ssock)
+            validate_response(response)
+
+    @pytest.mark.skip(reason="Disabled as it's currently flaky.")
+    def test_unwrap_existing_socket(self):
+        """
+        Validates we can break up the TLS layer
+        Request is sent over TLS. response received over regular TCP.
+
+        Currently disabled as its flaky on occassion.
+        """
+
+        def shutdown_handler(listener):
+            sock = listener.accept()[0]
+            ssl_sock = self.server_context.wrap_socket(sock, server_side=True)
+            request = consume_socket(ssl_sock)
+            validate_request(request)
+            sock = ssl_sock.unwrap()
+            sock.send(sample_response())
+
+        self.start_dummy_server(shutdown_handler)
+        sock = socket.create_connection((self.host, self.port))
+        ssock = SSLTransport(sock, self.client_context, server_hostname="localhost")
+        ssock.send(sample_request())
+        ssock.unwrap()
+        response = consume_socket(sock)
+        validate_response(response)
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_ssl_object_attributes(self):
+        """ Ensures common ssl attributes are exposed """
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            cipher = ssock.cipher()
+            assert type(cipher) == tuple
+
+            # No chosen protocol through ALPN or NPN.
+            assert ssock.selected_alpn_protocol() is None
+            assert ssock.selected_npn_protocol() is None
+
+            shared_ciphers = ssock.shared_ciphers()
+            assert type(shared_ciphers) == list
+            assert len(shared_ciphers) > 0
+
+            assert ssock.compression() is None
+
+            validate_peercert(ssock)
+
+            ssock.send(sample_request())
+            response = consume_socket(ssock)
+            validate_response(response)
+
+    @pytest.mark.timeout(PER_TEST_TIMEOUT)
+    def test_socket_object_attributes(self):
+        """ Ensures common socket attributes are exposed """
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            assert ssock.fileno() is not None
+            test_timeout = 10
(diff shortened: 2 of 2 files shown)


pquentin:
Now that #1679 is merged, I'd love to move this forward. So... this appears to make sense, and is surprisingly compact, but I would love to get sample code or even an integration with urllib3 to be able to play with it.

jalopezsilva:
Alright @pquentin, I've updated the PR with an initial integration. A  couple of things to keep in mind:

* **No support for less than Python3** - The ``wrap_bio`` methods used to perform the TLS in TLS connection were only added in Python 3. I gated the feature following the same conventions I saw in the code base. 
* **No support for SSL-less environments** - Similar to the previous comment, if we don't have support for SSL, there's no way we'll be able to do TLS in TLS. I gated the feature following the code base conventions.
* **Kept the allow_insecure_proxy option** - For now, I kept the flag we added in the last PR that toggles the behavior to use the absolute URI for HTTPS proxes with HTTPS connections. The option would only be used if provided by the user and would be helpful in cases where we can't do TLS in TLS (less than Python3).  We also found this useful for inspecting outgoing traffic for our internal services. (i.e to verify they're not sending prohibited data externally) I understand if you want to remove it though, we can chat about this more. 

There are a couple of things I want to do before merging this but I wanted to send it early to get some feedback. Items I still have to address: 

* More tests on the ``SSLTransport`` component. 
* Improve the integration. I have mixed feelings about the ``connection_requires_http_tunnel`` method. I might change how that currently works. 
* Production testing. There are two teams internally that need this patch. I'm planning on updating our local versions and running it on production for a while to make sure it works as expected.

Play with it and let me know if you have any feedback!

pquentin:
@jalopezsilva Thanks! Those assumptions make sense to me. I believe @sethmlarson has stronger feelings about the `allow_insecure_proxy` option, but that's something we can figure out later.

I restarted a failing PyPy run (sorry, our PyPy tests are quite flaky), and now the codecov checks work as expected since as you noted coverage is not at 100% yet. For what it's worth, https://codecov.io/gh/urllib3/urllib3/pull/1806/tree/src/urllib3 is the best way to see see covered lines.

I'll try to play with it, and will also try to find a way to review it. Please tell us when this runs in production. :)

src/urllib3/connection.py
```diff
@@ -402,6 +413,56 @@ def connect(self):
             or self.assert_fingerprint is not None
         )
 
+    def _connection_requires_tls_in_tls(self):
+        """
+        Indicates if the current connection requires two TLS connections, one to
+        the proxy and one to the destination server.
+        """
+        # Not supported in < py3 or ssl-less environments.
+        if not six.PY3 or not ssl:
```
sethmlarson, line 422: If we're asked by the user for TLS+TLS proxying we shouldn't quietly not provide it. This should be a proxy/connection error.

src/urllib3/contrib/ssl.py
```diff
@@ -0,0 +1,173 @@
...
+import socket
+import io
+
+
+class SSLTransportError(Exception):
+    "Will wrap any SSL errors received during socket IO."
+    pass
+
+
+class SSLTransport:
```
sethmlarson, line 11: How will this class work with pyOpenSSL and SecureTransport SSLContext objects?

src/urllib3/poolmanager.py
```diff
@@ -76,6 +71,18 @@ class InvalidProxyConfigurationWarning(HTTPWarning):
 #: The namedtuple class used to construct keys for the connection pool.
 #: All custom key schemes should include the fields in this key at a minimum.
 PoolKey = collections.namedtuple("PoolKey", _key_fields)
+_proxy_config_fields = ("ssl_context", "allow_insecure_proxy", "destination_scheme")
+
+
+class ProxyConfig:
```
sethmlarson, line 77: I'm not loving having an additional proxy config object here. The `allow_insecure_proxy` option shouldn't (?) be needed, if a user makes a request with `scheme=https, proxy_scheme=http` then they're opting in to an insecure proxy. Might not be "obvious" to some users but we can document this?

I don't think we lose a lot by having a singular `SSLContext` object shared between Proxy TLS and Connection TLS as any non-local HTTPS proxies are either corporate proxies that are providing their own bundle and MITM connections anyways or are services that have domain names and valid certificates that'd work with certifi, etc.
Am I thinking about this correctly?

src/urllib3/poolmanager.py
```diff
@@ -322,14 +330,26 @@ def _merge_pool_kwargs(self, override):
...
-        return parsed_url.scheme == "http" or self.proxy.scheme == "https"
+        return not connection_requires_http_tunnel(self.proxy, self.proxy_config)
+
+    def _validate_proxy_scheme_url_selection(self, url_scheme):
+        if (
+            not PY3
+            and url_scheme == "https"
+            and self.proxy is not None
+            and self.proxy.scheme == "https"
+            and not self.proxy_config.allow_insecure_proxy
```
sethmlarson, line 347: I'm not feeling great about this set of operations here. Maybe a plaintext explanation of what is being checked would make it easier to verify we're doing the right thing.

src/urllib3/poolmanager.py
```diff
@@ -408,6 +430,10 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
+    :param proxy_ssl_context:
```
sethmlarson, line 433: See above comment, might not need this.

test/test_ssl.py
```diff
@@ -63,7 +63,7 @@ def test_context_sni_with_ip_address(monkeypatch, has_sni, server_hostname, uses
     if uses_sni:
         context.wrap_socket.assert_called_with(sock, server_hostname=server_hostname)
     else:
-        context.wrap_socket.assert_called_with(sock)
+        context.wrap_socket.assert_called_with(sock, server_hostname=None)
```
sethmlarson, line 66: I think this will hurt Python versions that have `ssl` but don't have SNI support?
Might be good to verify this.

test/test_util.py
```diff
@@ -769,7 +775,9 @@ def test_ssl_wrap_socket_with_no_sni_warns(self):
                     sock=socket,
                     server_hostname="www.google.com",
                 )
-            mock_context.wrap_socket.assert_called_once_with(socket)
+            mock_context.wrap_socket.assert_called_once_with(
```
sethmlarson, line 778: See above SNI comment

test/test_util.py
```diff
@@ -826,3 +834,29 @@ def test_ip_family_ipv6_disabled(self):
     def test_assert_header_parsing_throws_typeerror_with_non_headers(self, headers):
         with pytest.raises(TypeError):
             assert_header_parsing(headers)
+
+    def test_connection_requires_http_tunnel_no_proxy(self):
+        assert not connection_requires_http_tunnel(None, None)
```
sethmlarson, line 839: Use keyword arguments for these tests so we know what `None` is referencing.

src/urllib3/poolmanager.py
```diff
@@ -408,6 +430,10 @@ class ProxyManager(PoolManager):
...
+        The proxy SSL context is used to establish the TLS connection to the
+        proxy when using HTTPS proxies.
+
     :param _allow_https_proxy_to_see_traffic:
```
sethmlarson, line 437: Can we remove this parameter now?

src/urllib3/connection.py
jalopezsilva: Agreed! I'll throw an error in this case.

src/urllib3/contrib/ssl.py
jalopezsilva: This is a good point and something I wasn't considering. I'll double-check if it's compatible.

test/test_ssl.py
jalopezsilva: The default parameter for ``server_hostname`` is actually None.  It made the code I needed to add simpler to add this as a parameter but I'll double-check it doesn't impact Python versions without SNI.

src/urllib3/poolmanager.py
jalopezsilva: We actually do need a second SSLContext object for our internal use case :(. I'll explain more in the bottom.

src/urllib3/poolmanager.py
jalopezsilva: see below.

jalopezsilva:
Hey @sethmlarson, thanks for the quick review. I'll address most of your comments. I wanted to highlight my response to two of them here.

* **Need for a proxy SSLContext** - We actually need this! We're moving all of our internal clients to use mutual authentication for our outgoing proxy. Instead of relying on a ``Proxy-Authorization`` header, we'll use a client certificate to authenticate with the proxy. In order to avoid the client certificate being sent to the destination server, we need two different SSLContexts, one for the proxy and one for the destination.  What I'm doing in the code right now is use the proxy SSLContext if provided, if not, use the same one that we'll use for the destination server. (i.e we create a SSLContext based on ``ca_certs`` provided by the user). Should I pass itdirectly instead of using the ``ProxyConfig`` object? I'm flexible here, I thought the ``ProxyConfig`` object made it easier to manage. 

* **Keeping the _allow_https_proxy_to_see_traffic option** - I kept the option as an alternative to the TLS within TLS for ``HTTPS`` destinations with an ``HTTPS`` proxy.  When the option is set, TLS within TLS is skipped. This is a nice to have for us, not a need to have. It makes it easier to inspect the outgoing traffic when running a corporate proxy. (i.e no need to MitM) The client would have to opt-in by setting this option to true. That said, I understand not everyone is running in a trusted corporate environment. It could be dangerous if users don't understand the implication of the option. I would prefer to keep it and make sure it's properly documented but I'm flexible and we can remove it if needed.  Let me know.

test/test_ssl.py
sethmlarson: Thank you! Will leave this conversation open until that's verified.

sethmlarson:
@jalopezsilva Thanks for all you're doing for this feature, I'm really happy with the work so far. It's an amazing improvement to the library and I'm really excited to be able to ship it to our users. :sparkles:

- You're totally spot-on about the client certificates, keep `proxy_ssl_context`.
- Having the `ProxyConfig` object be mutable worries me. Is there a way we can track the `destination_scheme` differently such that the config object can be immutable?
- `_allow_https_proxy_to_see_traffic` correct me if I'm not seeing this correctly, but this is essentially saying "I'm running an HTTPS proxy but I want to use forwarding proxy mode even if I request an HTTPS URL from my proxy? If we're keeping this feature let's promote it to public and then document it in laymans terms with a warning and technically as well w/ proxy mode and "who can see what".

Also feel free to call me out if I'm totally missing something here, this is a lot to think about this early in the morning. :)

src/urllib3/poolmanager.py
sethmlarson: Per below it'd be great to make this object a tuple for immutability.

jalopezsilva:
Quick update for you @pquentin, @sethmlarson, I'm testing this with a couple of teams in production.

We ran into an interesting bug that I thought would share. Do you know why requests by default will attempt to use ``pyopenssl`` if available?  https://github.com/psf/requests/blob/master/requests/__init__.py#L95-L100

``PyOpenSSLContext`` doesn't support the ``wrap_bio`` methods needed for TLS in TLS and ends up generating en error. I can check if we're using ``pyopenssl`` and throw an error in this case but I was curious why we still default to ``pyopenssl``. 

I suspect we did this before python had SNI support on the ssl module. ( Less than python 3.2 ) but I'm surprised we don't have a check for ssl.HAS_SNI in ``requests`` nowdays to toggle pyopenssl or not. I can also bring this up on the ``requests`` repo, just figured I would ask here first.

sethmlarson:
@jalopezsilva You're discovering some of the difficult parts of our current TLS situation with urllib3+requests. :) [Others and I have voiced concern](https://github.com/psf/requests/issues/5267) over having pyOpenSSL be the preferred TLS implementation when available, especially nowadays with SNI on almost all supported platforms.

I'd like to remove the unconditional use of pyOpenSSL, that'd definitely be something that would move us forwards. That way on the few platforms without SNI installing `requests[security]` wouldn't break, and then in the future the install of pyOpenSSL and `requests[security]` extra can be removed/blanked altogether.

As for this feature as it currently stands without BIO support on pyOpenSSL we can't provide TLS+TLS for HTTPS proxies. If we're going to implement it now we'd have to error out and explain to users why they're experiencing the error w/ potential fixes which may include uninstalling pyOpenSSL if Requests still unconditionally patches. Maybe we're fine with that given the low occurrence of TLS+TLS w/ `requests[security]`?

Also are we still unsure about SecureTransport, is there something we can do there?

Note as a part of 2.x I'd really like to drop our pyOpenSSL TLS implementation as well.

sethmlarson:
I opened a PR on Requests to drop the unconditional monkey-patching: https://github.com/psf/requests/pull/5443

pquentin:
Cool, that requests PyOpenSSL change is now released!

@jalopezsilva Can we do anything to help?

jalopezsilva:
Sorry for the lack of updates here! I got sidetracked at work with other priorities as well with everything going on in the world. Production testing has been mostly good, we found some bugs and got some feature requests that teams had. I'll put up a brief summary here:

* **Bug:** Using PyOpenSSL as a default on [requests](https://github.com/psf/requests/pull/5443)  was the # 1 bug we found. I had added a workaround in my internal patch but some teams were still affected. I'm glad the fix has finally rolled out. I'll integrate it internally and see if it works as expected.

* **Feature:** Using environment variables instead of passing certs + keys. This was a common request teams had. Quite often they were not using ``urllib3`` directly, but rather other library that used ``urlib3`` underneath. Most of the time, other libraries don't have an easy way to pass extra parameters to ``urllib3`` so even though ``urllib3`` supported TLS in TLS, teams couldn't configure it. I was hoping to add environment variables such as ``PROXY_TLS_CERT`` and ``PROXY_TLS_KEY`` and use them to generate the proxy SSLContext. What do you think?

* **Feature** More than a ``urllib3`` feature this is a ``ssl`` python module feature. PyOpenSSL supports creating SSLContext with certificates/key from memory. ``ssl`` doesn't. We had some teams excited to use this feature but they couldn't because they needed the in-memory support. While investigating this, I confirmed that PyOpenSSL supports ``wrap_bio`` methods which we can potentially use here as well. This is something that we can do as a follow up once we release this initial support. 

* **Bug:** There was a tricky bug involving SNI with a particular server that I was troubleshooting when I got side tracked. We found a server that wasn't taking the SNI hint when we used TLS in TLS. A regular TLS connection seemed to work as expected. I have to go back and confirm the problem wasn't in our code, or CPython. My hope is that this was a problem with the server implementation but I'll confirm once I have more data.

And that's about it. I'm currently closing up some other projects and my plan is to get back at this on July so we can close it out. I'll post here if I need help with anything!

pquentin:
Thanks for the update!

> I was hoping to add environment variables such as `PROXY_TLS_CERT` and `PROXY_TLS_KEY` and use them to generate the proxy SSLContext. What do you think?

@sigmavirus24 I've seen you mention that environment variables were generally a bad idea for urllib3/requests (https://github.com/psf/requests/issues/3070#issuecomment-319228483), do you think you can expand in order to explain when it's a bad idea and when it's not? It's possible to configure urllib3 with requests adapters, but it's more cumbersome.

> PyOpenSSL supports creating SSLContext with certificates/key from memory. `ssl` doesn't.

This would be really nice if `ssl` supported that, yeah. It would make writing tests using [trustme](https://github.com/python-trio/trustme/) easier, for example.

jalopezsilva:
>  I've seen you mention that environment variables were generally a bad idea for urllib3/requests (psf/requests#3070 (comment)), do you think you can expand in order to explain when it's a bad idea and when it's not? It's possible to configure urllib3 with requests adapters, but it's more cumbersome.

If I have to guess this is related to the ``HTTP_PROXY`` and ``HTTPS_PROXY`` environment variables. 

These environment variables can be used to configure if a proxy should be used or not when contacting a destination. The problem arises when these environment variables modify all of your requests to use a proxy. One of the problems that I've seen in the past is when users use urllib3/requests to contact internal (not requiring a proxy) and external destinations (requiring a proxy). If you unintentionally set the ``HTTP_PROXY`` environment variables, all of your traffic will attempt to use the proxy which is incredibly annoying and difficult to debug.

The environment variables I was thinking (``PROXY_SSL_CERT``, and ``PROXY_SSL_KEY``) would only allow you to configure the cert and key provided to the proxy if and only if you have already specified you want to talk to a proxy using ``HTTPS``. Hope that makes sense.

I'm prepping the changes with the environment variables included. We can discuss during code review too if needed.

src/urllib3/contrib/ssl.py
jalopezsilva: Update here:

This doesn't support pyOpenSSL or SecureTransport SSLContext. SecureTransport doesn't provide the ``wrap_bio`` methods we need to create the second TLS layer. pyOpenSSL does but I would need to implement it. I hope that, for now, adding support for just ``ssl`` is sufficient.

jalopezsilva:
All right folks, it's time.

I have done my best to test these changes on production and address the bugs that we have identified. I'm updating the PR with the latest changes. For now, I have sent a single commit which adds the ``SSLTransport`` class and corresponding unit tests. Let's start the review with this for now. 

I hope to update tomorrow with the second commit (which has the urllib3 integration).

src/urllib3/contrib/ssl.py
sethmlarson: This is fine, it's good that we got Requests using Python's TLS implementation by default now.

src/urllib3/contrib/ssl.py
```diff
@@ -0,0 +1,199 @@
+import ssl
+import socket
+import io
+
+
+class SSLTransportError(Exception):
```
sethmlarson, line 6: Why do we have a separate Exception type over reraising the `ssl.SSLError` within `_ssl_io_loop`?

src/urllib3/contrib/ssl.py
```diff
@@ -0,0 +1,199 @@
...
+            self.incoming, self.outgoing, server_hostname=server_hostname
+        )
+
+        # Perform initial handshake.
+        self._ssl_io_loop(self.sslobj.do_handshake)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
```
sethmlarson, line 44: nit: can change args to `*_`

src/urllib3/contrib/ssl.py
```diff
@@ -0,0 +1,199 @@
...
+        text.mode = mode
+        return text
+
+    def unwrap(self):
+        self._ssl_io_loop(self.sslobj.unwrap)
+
+    def close(self):
+        self.socket.close()
+
+    def getpeercert(self):
```
sethmlarson, line 134: Need an extra parameter here `binary_form=False` so we can both check certificate hashes and certificate SAN

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,393 @@
...
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            assert ssock.cipher() is not None
+            assert ssock.selected_alpn_protocol() is None
+            assert ssock.selected_npn_protocol() is None
+            assert ssock.shared_ciphers() is not None
+            assert ssock.compression() is None
+            assert ssock.getpeercert() is not None
```
sethmlarson, line 110: Will have to add test cases for `binary_form=True` and `False`, would be nice to change the `is not None` cases to type checks.

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,393 @@
...
+        self.start_dummy_server()
+
+        sock = socket.create_connection((self.host, self.port))
+        with SSLTransport(
+            sock, self.client_context, server_hostname="localhost"
+        ) as ssock:
+            assert ssock.fileno() is not None
+            test_timeout = 10
+            ssock.settimeout(test_timeout)
+            assert ssock.gettimeout() == test_timeout
```
sethmlarson, line 132: Can we also verify `sock.gettimeout()` was impacted here too

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,393 @@
...
+            ssock.send(
+                b"GET http://www.testing.com/ HTTP/1.1\r\n"
+                b"Host: www.testing.com\r\n"
+                b"User-Agent: awesome-test\r\n"
+                b"\r\n"
+            )
+            response = consume_socket(ssock)
+            assert response is not None
```
sethmlarson, line 140: Same as above

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,393 @@
...
+            assert ssock.getpeercert() is not None
+
+            ssock.send(
+                b"GET http://www.testing.com/ HTTP/1.1\r\n"
+                b"Host: www.testing.com\r\n"
+                b"User-Agent: awesome-test\r\n"
+                b"\r\n"
+            )
+            response = consume_socket(ssock)
+            assert response is not None
```
sethmlarson, line 119: Are the results of `consume_socket()` semi-consistent? If so we can check what we're getting out the other end to make sure it's not garbled/still encrypted.

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,393 @@
...
+        )
+        with self.client_context.wrap_socket(
+            sock, server_hostname="localhost"
+        ) as proxy_sock:
+            with pytest.raises(Exception) as e:
+                SSLTransport(
+                    proxy_sock, self.client_context, server_hostname="veryverywrong"
+                )
+            # Accommodate different python3 versions
+            assert e.type in [SSLTransportError, ssl.CertificateError]
```
sethmlarson, line 297: Why are we receiving two different error types here? Can we change this if we end up removing `SSLTransportError`?

src/urllib3/contrib/ssl.py
```diff
@@ -0,0 +1,199 @@
...
+                    raise SSLTransportError(e)
+                errno = e.errno
+
+            buf = self.outgoing.read()
+            self.socket.sendall(buf)
+
+            if errno is None:
+                should_loop = False
+            elif errno == ssl.SSL_ERROR_WANT_READ:
+                buf = self.socket.recv(4096)
```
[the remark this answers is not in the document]
jalopezsilva: Sounds good, I change dit to be 16KiB.

src/urllib3/contrib/ssl.py
jalopezsilva: No good reason to be honest, it's been removed.

test/contrib/test_ssltransport.py
jalopezsilva: I added ``binary_form`` as well as extended the checks done.

test/contrib/test_ssltransport.py
jalopezsilva: Not sure what you mean here Seth, can you clarify?

The timeout is applied but never takes effect. I would also prefer to not add a test that validates the timeout kicks in as they can be flakey. (e.g set timeout to 2 seconds, sleep for 3, verify exception is raised)

sethmlarson:
@jalopezsilva Did you want to get this PR merged with `SSLTransport` and work on the other aspects of TLS-in-TLS in a separate PR? Or would you rather get it all done here. Either way it'd be good to keep the `SSLTransport` work in a separate place to the proxy/config routing.

test/contrib/test_ssltransport.py
jalopezsilva: Good idea, I added validations on every response, request.

test/contrib/test_ssltransport.py
sethmlarson: Wanted to add a verification that the underlying `SSLTransport.socket.gettimeout()` is equal to the value set within `SSLTransport.settimeout()`. Tbh it's not that important since it can be clearly verified.

test/contrib/test_ssltransport.py
jalopezsilva: Prior to versions 3.6, SSL.CertificateError was actually a child of ValueError instead of SSLError. After 3.7 they changed it to be SSLError thankfully.

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+def validate_request(provided_request, binary=True):
+    assert provided_request is not None
+    expected_request = sample_request(binary)
+    assert provided_request == expected_request
+
+
+def sample_response(binary=True):
+    response = b"HTTP/1.1 200 OK\r\n" b"Content-Length: 0\r\n" b"\r\n"
```
sethmlarson, line 51: This is three separate `bytes` instances joining together, could we change this to `b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n"

jalopezsilva:
@sethmlarson Totally up to you guys! Whatever works best. We can open another PR if it's easier. 

I expect we'll have two more commits:

- Next commit: Adds the integration with ``urllib3`` main components. ``connection.py``, ``poolmanager.py``, etc.
- Last commit: Documentation about the feature, examples, etc.

sethmlarson:
Let's focus on the `SSLTransport` in this one and open another PR for future work :) Thanks for all this, pinging @pquentin for a review.

test/contrib/test_ssltransport.py
sethmlarson: Huh, I should have known that oddity :) Could this assert be `assert e.type == ssl.CertificateError`?

pquentin:
> pinging @pquentin for a review

I won't be able to look at this until Monday, sorry! Feel free to merge this part now

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,402 @@
...
+import socket
+import ssl
+import sys
+
+
+def server_client_ssl_contexts():
+    if hasattr(ssl, "PROTOCOL_TLS_SERVER"):
+        server_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+    else:
+        server_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
```
pquentin, line 20: Can we mention that this is only for Python 2.7 and 3.5? We'll remove them in a few months, and even if coverage will show the missing lines, it will help with the grep work.

test/contrib/test_ssltransport.py
```diff
@@ -0,0 +1,402 @@
...
+    server_context.load_cert_chain(DEFAULT_CERTS["certfile"], DEFAULT_CERTS["keyfile"])
+
+    if hasattr(ssl, "PROTOCOL_TLS_CLIENT"):
+        client_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    else:
+        client_context = ssl.SSLContext(ssl.PROTOCOL_TLS)
```
pquentin, line 26: Ditto

test/contrib/test_ssltransport.py
jalopezsilva: good idea.

test/contrib/test_ssltransport.py
jalopezsilva: ah, I added it anyway :)

jalopezsilva:
Everyone, thanks for the review and the comments. I have addressed the final nits and comments. Let me know if I missed anything. 

Couple of comments:

* Current failures - Current failures are on python 3.10 and pypy and they are unrelated. 

* Windows - Great catch @pquentin. There is a problem with the ``makefile`` tests and Windows environments. I have added a ``pytest.skipif`` option for now. I'll have to set up a Windows environment to test what's going on with ``makefile`` but I'll circle back through another PR to enable it. (I'm hoping we can merge this one as is)

* More coverage - I added a couple of more tests to increase coverage. The ``unwrap`` test works but is currently a tad flakey. For now I added another ``pytest.skip`` while I investigate this further.

I have created the next PR with the commit that performs the integration (https://github.com/urllib3/urllib3/pull/1923) Until we merge this one, the ``SSLTransport`` commit will show up there but I'll remove it once this is merged.

Let me know if we need to change anything else here. Thank you!

pquentin:
@jalopezsilva Thanks for the fixes! I'd be happy to merge this as-is, but I won't be able to until you rebase on top of https://github.com/urllib3/urllib3/pull/1922. Can you please do it? Thanks!

Also, a small note about process: I know there are two schools of though here, but I usually prefer when review comments are addressed through new commits, that way it's easy to see what changed as a reviewer. We squash commits when merging, so those follow-up commits won't show in the final history anyway.

jalopezsilva:
Thanks for the feedback @pquentin! I'll make sure to make follow up commits when addressing feedback during review. I'm normally used to the rebase and re-submit model, but I can see how subsequent commits make it easier to review changes.

I've rebased my changes past #1922. Hope that works. 

I also tried to get a clean run on the checks but I keep on getting certain flaky failures. Is ``test/with_dummyserver/test_socketlevel.py`` known to have flaky tests? Everything runs correctly locally. If you think it's related to my changes I can dig deeper but I thought I would ask.

pquentin:
The socket level tests are definitely flaky and not related to your changes, sorry about that!

I actually worked on flaky urllib3 tests quite a lot but it's a difficult problem to solve completely. To merge pull requests, we have a superpower that you don't have: as members of the GitHub org, we can restart individual Travis jobs.

pquentin:
I don't understand why the Python 3.10 build wasn't allowed to fail, and I also don't get why it turns out I cant' restart an individual job. Closing/reopening to see if that helps.

pquentin:
Oh wait, no, the build is failing but there's small text after that saying it's allowed to fail. I can't restart the failing PyPy job though. @sethmlarson Can you?

pquentin:
Okay, signing out of Travis and signing back in fixed the issue.

pquentin:
But now there's an AppVeyor failure, and I can't retry tests there. This is embarassing :) Closing/reopening.

sethmlarson:
Swooping in to save the day with a forced merge! :superhero:  Take that, CI scum! :boom::supervillain:

pquentin:
@jalopezsilva Now that our CI is less flaky, we can see that this pull request introduced four uncovered lines: https://codecov.io/gh/urllib3/urllib3/src/master/src/urllib3/contrib/ssl.py

Should we exclude them from coverage, should they be removed, or can you add tests for those?

jalopezsilva:
@pquentin I can add additional tests for lines # 60, 61 # 111.

I have a test for Line # 126 but it is currently disabled as it's flakey. Can you exclude # 126 for now until I get that test working? 

I'll follow up on a separate PR for the missing tests.

jalopezsilva:
PR for the missing tests over in #1961

PR #1812, opened 2020-03-12
Ensure load_verify_locations raises SSLError for all backends

pquentin:
This also adds TestSSL to the classes tested in SecureTransport and PyOpenSSL, since:

1. TestSSL was the most natural place for this test.
2. The test only makes sense when run against all SSL backends.

This is my take on #1517 from @pilou-: I've backported it to master, and made a few tweaks. In particular, we can't test the exact message, it's different with every backend.

Closes #1517

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -450,9 +450,12 @@ class PyOpenSSLContext(object):
             cafile = cafile.encode("utf-8")
         if capath is not None:
             capath = capath.encode("utf-8")
-        self._ctx.load_verify_locations(cafile, capath)
-        if cadata is not None:
-            self._ctx.load_verify_locations(BytesIO(cadata))
+        try:
+            self._ctx.load_verify_locations(cafile, capath)
+            if cadata is not None:
+                self._ctx.load_verify_locations(BytesIO(cadata))
+        except OpenSSL.SSL.Error as e:
+            raise ssl.SSLError("unable to load trusted certificates: %r" % e)
 
     def load_cert_chain(self, certfile, keyfile=None, password=None):
         self._ctx.use_certificate_chain_file(certfile)
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1395,7 +1395,6 @@ class TestSSL(SocketDummyServerTestCase):
         """
         with pytest.raises(SSLError) as exc:
             ssl_wrap_socket(None, ca_certs=os.devnull)
-        assert exc.type == SSLError
 
 
 class TestErrorWrapping(SocketDummyServerTestCase):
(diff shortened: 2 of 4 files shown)


pquentin:
@pilou- I can't assign you, but would love a review too! Thanks

src/urllib3/contrib/pyopenssl.py
```diff
@@ -450,9 +450,12 @@ def load_verify_locations(self, cafile=None, capath=None, cadata=None):
...
             capath = capath.encode("utf-8")
-        self._ctx.load_verify_locations(cafile, capath)
-        if cadata is not None:
-            self._ctx.load_verify_locations(BytesIO(cadata))
+        try:
+            self._ctx.load_verify_locations(cafile, capath)
+            if cadata is not None:
+                self._ctx.load_verify_locations(BytesIO(cadata))
+        except OpenSSL.SSL.Error as e:
+            raise ssl.SSLError("unable to load trusted certificates: %r" % e)
```
sethmlarson: Do we want a `six.raise_from(..., None)` here to remove the "error raised while handling another exception" messaging?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1387,6 +1389,14 @@ def socket_handler(listener):
                         pool.request("GET", "/", timeout=SHORT_TIMEOUT)
                     context.load_default_certs.assert_not_called()
 
+    def test_load_verify_locations_exception(self):
+        """
+        Ensure that load_verify_locations raises SSLError for all backends
+        """
+        with pytest.raises(SSLError) as exc:
+            ssl_wrap_socket(None, ca_certs=os.devnull)
+        assert exc.type == SSLError
```
sethmlarson: Isn't this assert redundant with the `pytest.raises()`?

src/urllib3/contrib/pyopenssl.py
pquentin: I don't know:

 * it's useful to see the original exception, eg. was it an issue with cafile/capath or cadata?
 * `six.raise_from(ssl.SSLError("unable to load trusted certificates"), e)` will loose the original exception for Python 2 users
 * All other raises in this file follow this form

I'd be happy to include your change if you still think it's a good idea, I genuinely don't know!

test/with_dummyserver/test_socketlevel.py
pquentin: Nice catch, thanks! Fixed.

src/urllib3/contrib/pyopenssl.py
pquentin: Oh, the point is that we don't need information about `e` twice?

sethmlarson: Ah it's no problem if this is the standard for this file!

pquentin:
Thanks @sethmlarson for the lint fix and the approval. Thank you @pilou- for your contribution!

PR #1813, opened 2020-03-12
ci: remove now-unneeded virtualenv workaround

pquentin:
See https://github.com/urllib3/urllib3/pull/1802 for details.


pquentin:
Case in point: One macOS job never started, so I'll have to cycle this pull request because I can't stop or restart this jobs.

PR #1817, opened 2020-03-18
Hang inside call to PoolManager request

### LINKED ISSUE #1808: urllib3 hanging inside call to PoolManager request
I have a repeatable issue with urllib3 hanging inside PoolManager request for an unusual edge case.  I have repeated this issue on two separate machines.

I am using Windows 10, 64-bit (fully patched) and Python 3.8.2, 64-bit.  The urllib3 version is 1.25.8 in a fresh venv environment.

The edge case is:
  1. Second call to urllib3 PoolManager request hangs inside request call and never returns
  2. Using streaming to read and write response to file
  3. url has a redirect (http status 301)

The original code used concurrent.futures, but the issue is the same without it using a simple loop.

I have put together a small extract from the code I have been running for about one year.  It was only when I recently started with a new use case that co-incidentally started using urls with redirects that I found this problem where each of concurrent.futures workers in my Python application would hang.  Then I removed concurrent features.

After stepping through the urllib3 source code, I could not figure out the root cause.  

I ask if others can reproduce the issue and then help find find the root cause.

# Code:
```
import logging
from pathlib import Path
import shutil
import urllib3

_logger = logging.getLogger(__name__)
_BUFFER_SIZE = 1024 * 1024   # buffer for downloading remote resource

url_headers = urllib3.make_headers(keep_alive=True, accept_encoding=True)
url_retries = urllib3.Retry(total=2, backoff_factor=1, status_forcelist=[500, 502, 503, 504])
url_client = urllib3.PoolManager(timeout=urllib3.Timeout(total=15.0), retries=url_retries, block=True,
                                 headers=url_headers)


def stream_response(url: str, filepath: Path, id: int):
    # Must call release_conn() after file copied but opening/writing exception is possible
    rsp = None
    try:
        _logger.debug(f'> {id:4d} GET: {url}')
        rsp = url_client.request('GET', url, preload_content=False)
        _logger.debug(f'> {id:4d} resp code: {rsp.status}')
        if rsp.status == 200:
            _logger.debug(f'> {id:4d} writing: "{filepath.name}"')
            with filepath.open('wb', buffering=_BUFFER_SIZE) as rfp:
                shutil.copyfileobj(rsp, rfp, length=_BUFFER_SIZE)
    except Exception as ex:
        _logger.exception(f'> {id:4d} Error')
    finally:
        if rsp:
            rsp.release_conn()

    return rsp.status


logging.basicConfig(level=logging.DEBUG,
            handlers=[logging.FileHandler('output.log', mode='w'), logging.StreamHandler()])
logging.captureWarnings(True)

_logger.info('Started')
items = [
    ['https://fwd.aws/v3mz8', Path('d1.pdf')],
    ['https://fwd.aws/88JKk', Path('d2.pdf')],
    ['https://fwd.aws/D9yDG', Path('d3.pdf')],
    ['https://fwd.aws/NmXza', Path('d4.pdf')],
    ['https://fwd.aws/NmXza', Path('d5.pdf')]
]

for id, it in enumerate(items):
    stream_response(it[0], it[1], id)
```

# Output:
```
D:\Temp\Python issue>
D:\Temp\Python issue>.env\Scripts\activate

(.env) D:\Temp\Python issue>
(.env) D:\Temp\Python issue>python urllib3_locking.py
INFO:__main__:Started
DEBUG:__main__:>    0 GET: https://fwd.aws/v3mz8
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): fwd.aws:443
DEBUG:urllib3.connectionpool:https://fwd.aws:443 "GET /v3mz8 HTTP/1.1" 301 0
DEBUG:urllib3.util.retry:Incremented Retry for (url='https://fwd.aws/v3mz8'): Retry(total=2, connect=None, read=None, redirect=None, status=None)
INFO:urllib3.poolmanager:Redirecting https://fwd.aws/v3mz8 -> https://aws-quickstart.s3.amazonaws.com/quickstart-hashicorp-consul/doc/hashicorp-consul-on-the-aws-cloud.pdf
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): aws-quickstart.s3.amazonaws.com:443
DEBUG:urllib3.connectionpool:https://aws-quickstart.s3.amazonaws.com:443 "GET /quickstart-hashicorp-consul/doc/hashicorp-consul-on-the-aws-cloud.pdf HTTP/1.1" 200 1610783
DEBUG:__main__:>    0 resp code: 200
DEBUG:__main__:>    0 writing: "d1.pdf"
DEBUG:__main__:>    1 GET: https://fwd.aws/88JKk

Process finished with exit code -1
```

I have to manually terminate the application (CPU usage for process is zero) and the output never shows the response http status code.

[Python issue - 200307.zip](https://github.com/urllib3/urllib3/files/4301755/Python.issue.-.200307.zip)

tycarac:
Patch for urllib3 hanging inside call to PoolManager request

Closes #1808


src/urllib3/poolmanager.py
```diff
@@ -380,17 +383,35 @@ def urlopen(self, method, url, redirect=True, **kw):
...
+            ):
+                pass
+
         try:
             retries = retries.increment(method, url, response=response, _pool=conn)
         except MaxRetryError:
             if retries.raise_on_redirect:
+                # Drain and release the connection for this response, since
+                # we're not returning it to be released manually.
+                drain_and_release_conn(response)
```
pquentin: Great catch, thanks! I hadn't noticed this part of the code

test/with_dummyserver/test_poolmanager.py
```diff
@@ -22,6 +22,27 @@ def setup_class(cls):
...
+                    fields={"target": "%s/?page=%d" % (self.base_url, number)},
+                    preload_content=preload
+                )
+
+        with PoolManager(block=True, maxsize=3) as http:
+            for j in range(1, 5):
+                r1 = build_request(number=j, preload=True)
+                r2 = build_request(number=j, preload=False)
+
+                data = bytes("Dummy server! page %d" % j, "utf-8")
```
pquentin: Writing `b"Dummy server! page %d" % j` here instead should fix the Python 2.7 builds

src/urllib3/poolmanager.py
```diff
@@ -380,17 +383,35 @@ def urlopen(self, method, url, redirect=True, **kw):
...
 
+        def drain_and_release_conn(response):
+            try:
+                # discard any remaining response body, the connection will be
+                # released back to the pool once the entire response is read
+                response.read()
+            except (
+                HTTPError,
+                SocketError,
+                BaseSSLError,
```
pquentin: Can you please explain why we're catching less exceptions here compared to the existing `drain_and_release_explain`? https://github.com/urllib3/urllib3/blob/8c7a43b4a4ca0c8d36d55f132daa2a43d06fe3c4/src/urllib3/connectionpool.py#L763-L776

(It's probably perfectly fine, I just haven't thought about that yet.)

tycarac: Looking at exceptions.py, the exceptions TimeoutError, ProtocolError, and SSLError are subclassed from HTTPError and thus caught by HTTPError.  Since no individual action is required, using HTTPError should suffice.
I could find no code for HTTPException.  There are imports such as
`from .packages.six.moves.http_client import HTTPException  # noqa: F401`
but I could not find the code.  Does it exist?

tycarac: My coding style is not to have silent failures as they can cause insidious bugs and undesired side-effects.  Instead of "pass", would it be good practice to log a debug statement here?

tycarac:
@pquentin 
Good feedback is hard to come by, so I appreciate it.  Also, a code base has history, context, practices and styles of its own.
Are you OK with a unit test on Dummy Server?  Initially I was experimenting with binding two listening ports with Tornado server, and then I found Dummy Server.  The two ports were for the redirect handling but I did not get as far to create a response for the second (redirect) binding.

src/urllib3/poolmanager.py
pquentin: > I could find no code for HTTPException. There are imports such as
> `from .packages.six.moves.http_client import HTTPException # noqa: F401`
> but I could not find the code. Does it exist?

six is a package we use to help with Python 2/3 compatibility. `six.moves` allows to use the same name for standard library modules that were renamed. See https://six.readthedocs.io/#module-six.moves for details. And indeed, httplib was renamed to http.client in Python 3. And this is where `HTTPException` is defined: https://docs.python.org/3/library/http.client.html#http.client.HTTPException. (Unfortunately, urllib3 is deeply entangled with http.client/httplib, which is one reason why https://github.com/python-trio/hip exists.)

Why `.packages`, you may ask? Because we vendor all urllib3 dependencies, six is currently included in the source: https://github.com/urllib3/urllib3/blob/master/src/urllib3/packages/six.py

Now, you're right that if we include `HTTPException`, we don't need other classes, but that was already the case in the existing `drain_and_release_conn` function. Unless we have another reason to have two different variations of that function, we should move it to a common place where it can be called. https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

This is why I was suggesting moving that function to src/urllib3/response.py. Does that make sense?

> My coding style is not to have silent failures as they can cause insidious bugs and undesired side-effects. Instead of "pass", would it be good practice to log a debug statement here?

This makes sense in general, but in that case it's not a failure, we expect to see those exceptions in the happy path. (Maybe we'll want to refine the list in the future because some exceptions should not be silently ignored, but that's an another issue.)

pquentin:
>  Are you OK with a unit test on Dummy Server? Initially I was experimenting with binding two listening ports with Tornado server, and then I found Dummy Server. The two ports were for the redirect handling but I did not get as far to create a response for the second (redirect) binding.

I'm not sure I understand the question, since the dummy server uses Tornado. We could try to modify the port and the host between redirects to in order to get multiple connection pools just like in your reproducing example in #1808, but I'm not sure we need this. I think we could just have one redirect, and then make sure we have 1/ only one connection pool 2/ no connection left open in that pool. I'll be able to experiment more on Monday, I'm unlikely to have enough time until then.

pquentin:
The only CI failure is about lint. You need to install black and run it on `test/with_dummyserver/test_poolmanager.py`. (I'd suggest configuring it in your editor too.) You can then run `nox -s lint` to see if the problem is fixed.

tycarac:
Regarding Tornado server.
So blur of me.  I remember what I did.  I started with two ports as I thought I would need to endpoints to test redirection.  Then I realized I that "localhost" and "127.0.0.1" map to two different pools and thus I did not need to use two ports.  Please disregard.

src/urllib3/poolmanager.py
tycarac: @pquentin 
I understand HTTPException and will add it to the list.  As for logging, I agree with your reasoning.

I was looking at the code base, I and I see in response.py
```
def release_conn(self):
    if not self._pool or not self._connection:
        return

    self._pool._put_conn(self._connection)
    self._connection = None
```
The function **drain_and_release_conn()** acts on the response object and not the connection.  Thus putting the function in response.py is good (DRY), but I think the function name should change.  Being in the HTTPReponse object, I would suggest a name **drain()** so the call would be:
`repoonse.draing()`

Your thoughts?

tycarac:
@pquentin 
I figured out a bit more about git and that forked git repos do not update.  Anyway, I made some changes and submitted a new pull request.  I am curious to know what you think.

pquentin:
Thanks! This is looking much better. I'm very happy with the progress here.

I think we can improve the tests by 1/ simplifying the existing PoolManager one 2/ testing the `retries.raise_on_redirect` case. Unfortunately I won't be able to figure out the details until tomorrow morning at best. Sorry about that.

I don't know what to think about the drain() name right now. I think I would have just left the function out of the Response class. What I know is that we should not have to comment "this releases the connection!" every time we call this function, it's better to encode this in the name somehow.

pquentin:
Here is how I would rewrite test_redirect_preload:

```python3
def test_redirect_without_preload_releases_connection(self):
    with PoolManager(block=True, maxsize=2) as http:
        r = http.request(
            "GET", "%s/redirect" % self.base_url, preload_content=False
        )
        assert r._pool.num_requests == 2
        assert r._pool.num_connections == 1
        assert len(http.pools) == 1
```

It's carefully designed to be as concise as possible but still:

1. make sure we don't leak any connections
2. avoid hanging without the fix.

We're checking that with a redirect and `preload_content=False`, we did make two requests, but only used one connection because the first one was released (objective 1). We use `maxsize=2` so that the test does not hang if we do use two connections (objective 2). And just to be sure, we check that we only used one connection pool, as we are at the PoolManager level (objective 1).

(By the way, since it's not the simplest case, I would not place it first in the TestPoolManager class.)

Does that make sense?

(The next step is to figure out how to test the raise_on_redirect case.)

src/urllib3/response.py
```diff
@@ -29,6 +30,17 @@
 log = logging.getLogger(__name__)
 
 
+def drain_conn(response):
```
pquentin: (For what it's worth, I now believe having `drain_conn` inside the response class is perfectly fine, I just needed a better name, and we now have that. But keeping the function outside of the class is fine too.)

tycarac:
@pquentin 
The Travis CI build failed with error in the integration testing:
```
FAIL: test_stress_test_token_bucket (tests.functional.retries.test_bucket.TestTokenBucketThreading)
  File "/home/travis/build/urllib3/urllib3/botocore/tests/functional/retries/test_bucket.py", line 109, in test_stress_test_token_bucket
    self.assertTrue(not any(x < (0.3 * mean) for x in distribution))
AssertionError: False is not true
```
Looks to be inside the AWS Python Boto library.  Would this be a test platform failure?  Can the Travis CI build be re-triggered?

tycarac:
@pquentin 
The code I have been using to fetch web content to local files is:
```
try:
    rsp = url_client.request('GET', url)
    _logger.debug(f'> {id:4d} resp code:  {rsp.status}')
    if rsp.status == 200:
        _logger.debug(f'> {id:4d} write:      "{filepath.name}"')
        with filepath.open('wb', buffering=_BUFFER_SIZE) as rfp:
            shutil.copyfileobj(rsp, rfp, length=_BUFFER_SIZE)
except urllib3.exceptions.HTTPError as ex:
    _logger.exception(f'> {id:4d} HTTP error')
except Exception as ex:
    _logger.exception('Unexpected')
    raise ex
finally:
    if rsp:
        rsp.release_conn()
```
Looking at this code, I am wondering if function **release_conn** should internally call **drain_conn**.  Initially, when I looked at the code for **release_conn**, I saw it was catching socket errors, eg timeout, which implied to me either a full response was received or it was not.  However, with the pattern to stream content, if I have it right, I am wondering if I did not get http status code 200, then is there a possibility of outstanding data on the socket that is not being read but released back to the pool.

So, should **release_conn** internally call **drain_conn**?

pquentin:
What do you think of the test mentioned above? Does it make sense? If yes, I suggest including it in your code.

>  The Travis CI build failed with error in the integration testing:

I don't know why Travis ran this integration test. @sethmlarson I thought this would only apply to release commits?

> So, should **release_conn** internally call **drain_conn**?

Oh, that's a really good question. I just looked at this, and there's no possibility of outstanding data, because before reusing the connection we call `_get_conn`, and it closes the socket if needed.

But that means we do have two very similar ways to drain connections. One option is to replace the call to drain_conn() with a call to release_conn(). For now let's keep things like they are, and I'll try to think about it.

tycarac:
@pquentin, @sethmlarson 
I had a look at **_get_conn** and other functions in connectionpool.py.  I may be blur but I do not see the mechanism to clear any residual response data.  I was thinking how I could test this.

I had a look at [rfc7231 Redirect](https://tools.ietf.org/html/rfc7231#section-6.4) and most redirect status codes do not state a HTTP body to be prohibited or optional (except 307, 308).  It seems silly and very unlikely to have a HTTP body response for a HTTP redirect, but it is allowed.  But this is my first time delving into standards and there is much context and practice I do not have.

I am still of the mind to call **drain_conn** from inside **release_conn** as a precaution.  

Your thoughts?

pquentin:
> If this mechanism closes the connection, then would we loose advantages of pooling with the overhead of opening connections?

Oh, you're right, we don't want to close connections when we can avoid it. Thanks, I had totally missed that. Your understanding of all this is impressing!

> I am still of the mind to call **drain_conn** from inside **release_conn** as a precaution.

Can you please open an issue so that we can discuss about this separately? It would be a much broader change than the one we're discussing here, and I'd like to get this code merged.

The change currently looks really good, we just need to find a way to test the `if retries.raise_on_redirect:` case. (Unfortunately our coverage measurement is broken right now so we need to be extra careful here.) I don't think we can test this without explicitly checking that drain_conn() was called since we raise an exception on the first redirect and I don't think we can figure out if the connection was released or not here.

pquentin:
Okay, here is the final bit:

```diff
diff --git a/test/with_dummyserver/test_poolmanager.py b/test/with_dummyserver/test_poolmanager.py
index 3f130a94..1160a2b6 100644
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -115,6 +115,7 @@ class TestPoolManager(HTTPDummyServerTestCase):
                         % (self.base_url, self.base_url)
                     },
                     retries=1,
+                    preload_content=False,
                 )
 
             with pytest.raises(MaxRetryError):
@@ -126,8 +127,15 @@ class TestPoolManager(HTTPDummyServerTestCase):
                         % (self.base_url, self.base_url)
                     },
                     retries=Retry(total=None, redirect=1),
+                    preload_content=False,
                 )
 
+            # Even with preload_content=False and raise on redirects, we reused the same
+            # connection
+            assert len(http.pools) == 1
+            pool = http.connection_from_host(self.host, self.port)
+            assert pool.num_connections == 1
+
     def test_redirect_cross_host_remove_headers(self):
         with PoolManager() as http:
             r = http.request(

```

This changes allows to test the `retries_on_redirect` case. With this included, I'll finally be able to merge this! I can't push to your branch, so you'll have to do it. Thanks

pquentin:
@tycarac We'll also need a rebase from master to make the new coverage checks happy. I'm happy to do this myself in another pull request. I'll make sure to credit you for this change, of course.

tycarac:
@pquentin 
Sorry for the delay.  I will make another pull request shortly.  I have been unwell but and keen to continue.  There are two issues that can be done later: a) possible issue with draining redirect responses and b) the test on redirect.
If the pull request needs tweaking, please do so to get the release out.

pquentin:
> Sorry for the delay. I will make another pull request shortly.

To be clear, *you* don't need to make another pull request, you can change your existing branch, by merging master, on rebasing on top of master. If I was to do this myself, then another pull request would have been needed, as I can't change your branch. That being said, feel free to open another pull request if it's easier.

> I have been unwell but and keen to continue.

Oh, sorry to hear that. Don't forget that health comes first! (And that many other things come before open source work.)

> There are two issues that can be done later: a) possible issue with draining redirect responses and b) the test on redirect.

Yeah, I think we can open an issue to figure out a) later. b) however is ready to be included, as mentioned in https://github.com/urllib3/urllib3/pull/1817#issuecomment-605806613.

> If the pull request needs tweaking, please do so to get the release out.

No worries, your pull request is not blocking a release. Code waits for releases, but releases don't wait for code. :)

tycarac:
@pquentin 
Made the pull request correctly (I think).  GitHub build checks now in progress.  More firsts in learning to apply a git diff :-)

PR #1819, opened 2020-03-18
Ensure test host resolves to IPv6 in tests

### LINKED ISSUE #1818: TestSNI:test_hostname_in_first_request_packet hangs
My system supports IPv6 and resolves `localhost` to `127.0.0.1`.
When I run the test suite on my system, `TestSNI:test_hostname_in_first_request_packet` hangs.
This also fail about 80 more tests with `MaxRetryError`.

### Expected Result
1. The test should pass.
2. The test should never hang.

### Actual Result
The test uses `_start_server` to create a `SocketServerThread` that binds on an IPv6 address by default (if possible):
```python
 70 HAS_IPV6_AND_DNS = _has_ipv6("localhost")
...
 90     USE_IPV6 = HAS_IPV6_AND_DNS
...
101         if self.USE_IPV6:
102             sock = socket.socket(socket.AF_INET6)
```
but the client connection might still resolve to IPv4 address and use it:
```python
 61     for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
 62         af, socktype, proto, canonname, sa = res
 63         sock = None
 64         try:
 65             sock = socket.socket(af, socktype, proto)
...
103     if HAS_IPV6:
104         family = socket.AF_UNSPEC
```
The test waits until an event is set, which is never, because the server socket never gets accepted.

### Reproduction Steps
```
$ cat noxfile-single.patch
30a31,32
>         "-o", "log_cli=true",
>         "-k", "test_hostname_in_first_request_packet",
$ patch noxfile.py noxfile-single.patch
patching file noxfile.py
$ nox --sessions test-2.7
nox > Running session test-2.7
nox > Creating virtual environment (virtualenv) using python2.7 in .nox/test-2-7
nox > pip install -r dev-requirements.txt
nox > pip install .[socks,secure,brotli]
nox > pip --version
pip 20.0.2 from /home/user/dev/urllib3/.nox/test-2-7/lib/python2.7/site-packages/pip (python 2.7)
nox > python --version
Python 2.7.17
nox > python -c import struct; print(struct.calcsize('P') * 8)
64
nox > python -m OpenSSL.debug
pyOpenSSL: 19.1.0
cryptography: 2.8
cffi: 1.14.0
cryptography's compiled against OpenSSL: OpenSSL 1.1.1d  10 Sep 2019
cryptography's linked OpenSSL: OpenSSL 1.1.1d  10 Sep 2019
Pythons's OpenSSL: OpenSSL 1.1.1  11 Sep 2018
Python executable: /home/user/dev/urllib3/.nox/test-2-7/bin/python
Python version: 2.7.17 (default, Nov  7 2019, 10:07:09)
[GCC 7.4.0]
Platform: linux2
sys.path: ['', '/usr/lib/python2.7', '/usr/lib/python2.7/plat-x86_64-linux-gnu', '/usr/lib/python2.7/lib-tk', '/usr/lib/python2.7/lib-old', '/usr/lib/python2.7/lib-dynload', '/home/user/dev/urllib3/.nox/test-2-7/lib/python2.7/site-packages']
nox > coverage run --parallel-mode -m pytest -o log_cli=true -k test_hostname_in_first_request_packet -r a --tb=native --no-success-flaky-report test/
==================================== test session starts ====================================
platform linux2 -- Python 2.7.17, pytest-4.6.6, py-1.8.1, pluggy-0.13.1
rootdir: /home/user/dev/urllib3, inifile: setup.cfg
plugins: flaky-3.6.1, timeout-1.3.3
collected 1342 items / 1339 deselected / 3 selected

test/contrib/test_pyopenssl.py::TestSNI::test_hostname_in_first_request_packet
[hangs]
```

### System Information
```
$ git rev-parse --short HEAD
eee53a69
$ cat /etc/hosts
127.0.0.1       localhost
127.0.1.1       ubuntu

# The following lines are desirable for IPv6 capable hosts
::1     ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

### Fix
I can add a timeout to the test, but that's a solution to the root problem.
I suggest that `ServerSocketThread` treat a system where `localhost` does not resolve to `::1` as having insane IPv6 configuration and use the current IPv4 fallback mechanism.

hodbn:
Closes #1818

### DIFF
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -40,6 +40,20 @@ DEFAULT_CA = os.path.join(CERTS_PATH, "cacert.pem")
 DEFAULT_CA_KEY = os.path.join(CERTS_PATH, "cacert.key")
 
 
+def _sane_ipv6_dns():
+    """ Returns True if the system resolves localhost to an IPv6 address by default. """
+    sane_ipv6_dns = False
+    try:
+        for res in socket.getaddrinfo(host, None, socket.AF_UNSPEC):
+            af, _, _, _, _ = res
+            if af == socket.AF_INET6:
+                sane_ipv6_dns = True
+    except socket.gaierror:
+        pass
+
+    return sane_ipv6_dns
+
+
 def _has_ipv6(host):
     """ Returns True if the system can bind an IPv6 address. """
     sock = None
@@ -54,7 +68,7 @@ def _has_ipv6(host):
         try:
             sock = socket.socket(socket.AF_INET6)
             sock.bind((host, 0))
-            has_ipv6 = True
+            has_ipv6 = _sane_ipv6_dns()
         except Exception:
             pass
 
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -40,18 +40,18 @@ DEFAULT_CA = os.path.join(CERTS_PATH, "cacert.pem")
 DEFAULT_CA_KEY = os.path.join(CERTS_PATH, "cacert.key")
 
 
-def _sane_ipv6_dns():
-    """ Returns True if the system resolves localhost to an IPv6 address by default. """
-    sane_ipv6_dns = False
+def _resolves_to_ipv6(host):
+    """ Returns True if the system resolves host to an IPv6 address by default. """
+    resolves_to_ipv6 = False
     try:
         for res in socket.getaddrinfo(host, None, socket.AF_UNSPEC):
             af, _, _, _, _ = res
             if af == socket.AF_INET6:
-                sane_ipv6_dns = True
+                resolves_to_ipv6 = True
     except socket.gaierror:
         pass
 
-    return sane_ipv6_dns
+    return resolves_to_ipv6
 
 
 def _has_ipv6(host):
@@ -68,7 +68,7 @@ def _has_ipv6(host):
         try:
             sock = socket.socket(socket.AF_INET6)
             sock.bind((host, 0))
-            has_ipv6 = _sane_ipv6_dns()
+            has_ipv6 = _resolves_to_ipv6("localhost")
         except Exception:
             pass
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -284,7 +284,7 @@ In chronological order:
 
 * Jorge Lopez Silva <https://github.com/jalopezsilva>
   * Added support for forwarding requests through HTTPS proxies.
-  
+
 * Benno Rice <benno@jeamland.net>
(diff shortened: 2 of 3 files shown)


dummyserver/server.py
```diff
@@ -40,6 +40,20 @@
 DEFAULT_CA_KEY = os.path.join(CERTS_PATH, "cacert.key")
 
 
+def _sane_ipv6_dns():
+    """ Returns True if the system resolves localhost to an IPv6 address by default. """
+    sane_ipv6_dns = False
+    try:
+        for res in socket.getaddrinfo(host, None, socket.AF_UNSPEC):
```
pquentin: Don't we want to pass `host` to this function? And then replace "localhost" with "host" in the docstring?

dummyserver/server.py
```diff
@@ -40,6 +40,20 @@
 DEFAULT_CA_KEY = os.path.join(CERTS_PATH, "cacert.key")
 
 
+def _sane_ipv6_dns():
```
pquentin: Can we please avoid using `sane` here? I know we're using the word "sanity" elsewhere in the project, but I'd like to avoid using it more. `_resolves_to_ipv6` avoids this issue and is also more descriptive.

dummyserver/server.py
hodbn: You are right :) done.

dummyserver/server.py
hodbn: Done :)

pquentin:
I can confirm that this changes fixes the test when localhost does not resolve to `::1`. But, how did you end up without `::1 localhost` in /etc/hosts on Ubuntu?

This does not appear to be a default setting:

```
$ docker run ubuntu cat /etc/hosts
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
Digest: sha256:04d48df82c938587820d7b6006f5071dbbffceb7ca01d2814f81857c631d44df
Status: Downloaded newer image for ubuntu:latest
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
172.17.0.2      26c71c3e7ed3
```

hodbn:
I have no idea, but I can see someone referencing the same at https://askubuntu.com/questions/880270/what-are-the-default-settings-for-etc-hosts-and-etc-hostname.

pquentin:
Thanks @hodbn for your contribution! Merged in masted.

PR #1820, opened 2020-03-18
Skip a test when using unsupported dns resolvers

### LINKED ISSUE #1809: Dotted FQDN tests may fail or succeed depending on system DNS configuration
Related to #1475, #1648.

I've recently spent some time debugging test failures reported by Gentoo users, and came to the conclusion that resolving `localhost.` may work or not depending on the system resolver configuration. I've been able to make it work reliably by listing `localhost.` in `/etc/hosts` explicitly or by using `nss_myhostname`. It also may work when external DNS is used and works, though I have no clue why glibc makes it work like that. On the other hand, if the system can't connect to the DNS resolver, it fails.

Since this test is unreliable, could you please disable (skip) it by default and enable only on known-good travis infra? I would supply a patch but I don't really know what values `TRAVIS_INFRA` can have.

hodbn:
Closes #1809

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -39,6 +39,24 @@ SHORT_TIMEOUT = 0.001
 LONG_TIMEOUT = 0.5 if os.environ.get("CI") else 0.01
 
 
+def _can_resolve(host):
+    """ Returns True if the system can resolve host to an address. """
+    can_resolve = False
+
+    try:
+        socket.getaddrinfo(host, None, socket.AF_UNSPEC)
+        can_resolve = True
+    except socket.gaierror:
+        pass
+
+    return can_resolve
+
+
+# Some systems might not resolve "localhost." correctly. We treat such
+# resolvers as unsupported. See https://github.com/urllib3/urllib3/issues/1809
+HAS_SUPPORTED_RESOLVER = _can_resolve("localhost.")
+
+
 def clear_warnings(cls=HTTPWarning):
     new_filters = []
     for f in warnings.filters:
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -41,15 +41,11 @@ LONG_TIMEOUT = 0.5 if os.environ.get("CI") else 0.01
 
 def _can_resolve(host):
     """ Returns True if the system can resolve host to an address. """
-    can_resolve = False
-
     try:
         socket.getaddrinfo(host, None, socket.AF_UNSPEC)
-        can_resolve = True
+        return True
     except socket.gaierror:
-        pass
-
-    return can_resolve
+        return False
 
 
 # Some systems might not resolve "localhost." correctly. We treat such
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -48,9 +48,10 @@ def _can_resolve(host):
         return False
 
 
-# Some systems might not resolve "localhost." correctly. We treat such
-# resolvers as unsupported. See https://github.com/urllib3/urllib3/issues/1809
-HAS_SUPPORTED_RESOLVER = _can_resolve("localhost.")
+# Some systems might not resolve "localhost." correctly.
+# See https://github.com/urllib3/urllib3/issues/1809 and
+# https://github.com/urllib3/urllib3/pull/1475#issuecomment-440788064.
+RESOLVES_LOCALHOST_FQDN = _can_resolve("localhost.")
 
 
 def clear_warnings(cls=HTTPWarning):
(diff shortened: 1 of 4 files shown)


pquentin:
@sethmlarson @mgorny @hodbn I'd like to test this. How can I get a system where resolving `localhost.` fails?

mgorny:
I suppose it's sufficient to remove `dns` (or `resolve`, or...) from your `/etc/nsswitch.conf` and just leave `files` in `hosts:`, and make sure you don't have `localhost.` in `/etc/hosts`.

hodbn:
Thanks, I can now see that there are more tests that depend on resolving "localhost.".
I'm pushing a commit that skips more relevant tests, can you please review it?

If you would like to leave everything as is you can use:
```c
#define _GNU_SOURCE

 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 #include <dlfcn.h>
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netdb.h>

 int getaddrinfo(const char *node, const char *service,
                 const struct addrinfo *hints, struct addrinfo **res) {
     int (*orig)(const char *, const char *, const struct addrinfo *,
                 struct addrinfo **);
     char *block_node = getenv("BLOCK_NODE");

     if (block_node && !strcmp(block_node, node)) {
         printf("\033[0;31m");
         printf("Blocking resolve of '%s'\n", node);
         printf("\033[0m");
         return EAI_AGAIN;
     }
     orig = dlsym(RTLD_NEXT, "getaddrinfo");
     return (*orig)(node, service, hints, res);
 }
```
run with:
```shell
$ gcc -Wall -fPIC -shared -o resolve-blocker.so resolve-blocker.c -ldl
$ LD_PRELOAD=./resolve-blocker.so BLOCK_NODE="localhost." python
Python 2.7.17 (default, Nov  7 2019, 10:07:09)
[GCC 7.4.0] on linux2
>>> import socket
>>> socket.getaddrinfo("localhost", None)
[(10, 1, 6, '', ('::1', 0, 0, 0)), (10, 2, 17, '', ('::1', 0, 0, 0)), (10, 3, 0, '', ('::1', 0, 0, 0)), (2, 1, 6, '', ('127.0.0.1', 0)), (2, 2, 17, '', ('127.0.0.1', 0)), (2, 3, 0, '', ('127.0.0.1', 0))]
>>> socket.getaddrinfo("localhost.", None)
Blocking resolve of 'localhost.'
socket.gaierror: [Errno -3] Temporary failure in name resolution
```

test/__init__.py
```diff
@@ -39,6 +39,24 @@
...
+    """ Returns True if the system can resolve host to an address. """
+    can_resolve = False
+
+    try:
+        socket.getaddrinfo(host, None, socket.AF_UNSPEC)
+        can_resolve = True
+    except socket.gaierror:
+        pass
+
+    return can_resolve
```
mgorny: Wouldn't the following be simpler?

```suggestion
    try:
        socket.getaddrinfo(host, None, socket.AF_UNSPEC)
        return True
    except socket.gaierror:
        return False
```

mgorny:
Besides that one comment, looks good and seems to work fine on my system.

pquentin:
@mgorny Thanks! Are those tests failing without this pull request on your system?

mgorny:
Prior to it, `test_dotted_fqdn` and `test_request_host_header_ignores_fqdn_dot` fail for me. Plus `test_basic_ipv6_proxy` that fails prior and post, but it's irrelevant to the topic at hand.

pquentin:
@mgorny That's really helpful. One last question hopefully, is `test_basic_ipv6_proxy` skipped on master now that I've merged #1819? Thanks!

mgorny:
Yes: `1135 passed, 207 skipped, 138 warnings` ;-).

test/__init__.py
```diff
@@ -39,6 +39,20 @@
...
+    try:
+        socket.getaddrinfo(host, None, socket.AF_UNSPEC)
+        return True
+    except socket.gaierror:
+        return False
+
+
+# Some systems might not resolve "localhost." correctly. We treat such
+# resolvers as unsupported. See https://github.com/urllib3/urllib3/issues/1809
+HAS_SUPPORTED_RESOLVER = _can_resolve("localhost.")
```
pquentin: I don't understand what a supported resolver is. Can we rename this to `RESOLVES_LOCALHOST_FQDN` or `RESOLVES_DOTTED_LOCALHOST`? And change the skipif reason to `"Can't resolve localhost."`.

Also, since 1/ we don't want to repeat ourselves and 2/ we want to be consistent with the skips below, we should write a function like `notBrotlipy` but adapted for this use case. Not sure about the name, maybe `resolvesLocalhostFQDN` or `resolvesDottedLocalhost`? This would allow us to just decorate the failing tests, like we do for all the other cases.

test/with_dummyserver/test_https.py
```diff
@@ -134,6 +135,9 @@ def test_simple(self):
             assert r.status == 200, r.data
 
     @fails_on_travis_gce
```
pquentin: Is this still needed with this change?

test/__init__.py
hodbn: Done, with `RESOLVES_LOCALHOST_FQDN` and `resolvesLocalhostFQDN`.

test/with_dummyserver/test_https.py
hodbn: You are right :)

PR #1821, opened 2020-03-19
Stop skipping SecureTransport tests

pquentin:
They're failing on macOS 10.15, which means I can't run them locally,
which is why I skip them. Sorry about that.

### DIFF
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -819,6 +819,11 @@ class SecureTransportContext(object):
         if capath is not None:
             raise ValueError("SecureTransport does not support cert directories")
 
+        # Raise if cafile does not exist.
+        if cafile is not None:
+            with open(cafile):
+                pass
+
         self._trust_bundle = cafile or cadata
 
     def load_cert_chain(self, certfile, keyfile=None, password=None):
--- a/test/contrib/test_securetransport.py
+++ b/test/contrib/test_securetransport.py
@@ -10,8 +10,6 @@ try:
 except ImportError:
     pass
 
-pytestmark = pytest.mark.skip()
-
 
 def setup_module():
     try:
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1394,7 +1394,7 @@ class TestSSL(SocketDummyServerTestCase):
         Ensure that load_verify_locations raises SSLError for all backends
         """
         with pytest.raises(SSLError):
-            ssl_wrap_socket(None, ca_certs=os.devnull)
+            ssl_wrap_socket(None, ca_certs="/tmp/fake-file")
 
 
 class TestErrorWrapping(SocketDummyServerTestCase):
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -55,6 +55,7 @@ from test import (
     SHORT_TIMEOUT,
     LONG_TIMEOUT,
     notPyPy2,
+    notSecureTransport,
 )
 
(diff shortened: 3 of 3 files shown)


sethmlarson:
No problem! This is why visual indicators for test run differences are useful :P

pquentin:
Sigh, and obviously the `ca_certs=os.devnull` hack introduced in #1812 works fine for the standard library and PyOpenSSL but breaks on macOS. I plan to try to fix that for a few days, but if I don't find enough time or otherwise fail, I'll revert #1812 and merge this one.

> This is why visual indicators for test run differences are useful :P

Do you mean the codecov coverage? Yeah, that would help, but not in this case. The problem here is that SecureTransport is a "contrib" module and is not part of coverage computations.

(I'm considering dropping codecov and instead make sure each Python 3 run is at 100%, by enforcing it with `coverage` and ignoring a few currently covered lines. It would be way more reliable.)

sethmlarson:
I mean more of a "here's what tests ran and for which config and how that differs from the master branch". We don't get that with Codecov or Travis or any CI unfortunately.

sethmlarson:
Looks like there is an issue that was introduced in #1812, could you take a look @pquentin?

pquentin:
Just got the following unrelated error in `test/contrib/test_securetransport.py::TestSSL::test_ssl_read_timeout` at https://travis-ci.org/github/urllib3/urllib3/jobs/665716556.

```
  File "/Users/travis/build/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 425, in _error_catcher
    yield
  File "/Users/travis/build/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 503, in read
    data = self._fp.read() if not fp_closed else b""
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/http/client.py", line 467, in read
    s = self._safe_read(self.length)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/http/client.py", line 610, in _safe_read
    raise IncompleteRead(data, amt-len(data))
http.client.IncompleteRead: IncompleteRead(3 bytes read, 7 more expected)
  File "/Users/travis/build/urllib3/urllib3/test/with_dummyserver/test_socketlevel.py", line 1209, in test_ssl_read_timeout
    response.read()
  File "/Users/travis/build/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 503, in read
    data = self._fp.read() if not fp_closed else b""
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py", line 131, in __exit__
  File "/Users/travis/build/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 443, in _error_catcher
    raise ProtocolError("Connection broken: %r" % e, e)
urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(3 bytes read, 7 more expected)', IncompleteRead(3 bytes read, 7 more expected))
```

Cycling the pull request.

pquentin:
Closing/reopening did not restart the builds. Trying again...

pquentin:
Okay, so the 3.8 failure appears to be real: there's another one at https://travis-ci.org/github/urllib3/urllib3/jobs/665746917

pquentin:
So the failing test never worked on Python 3.8 + SecureTransport, since the commit who added it to the SecureTransport tests is the same commit that disabled SecureTransport tests. And at the time, I was using Python 3.7 for development.

It could be interesting to figure out the issue, but for now let's disable the test: it's more important to have working SecureTransport tests than to figure out why this new one is failing.

Please take another look.

Issue #1822, opened 2020-03-19
 InvalidHeader: Invalid Retry-After header: 1m0s with github.com


ssbarnea:
Apparently retry logic inside urllib3 chokes when receiving 429 replies from github itself.
```
self = Retry(total=2, connect=None, read=None, redirect=None, status=None)
retry_after = '1m0s'

    def parse_retry_after(self, retry_after):
        # Whitespace: https://tools.ietf.org/html/rfc7230#section-3.2.4
        if re.match(r"^\s*[0-9]+\s*$", retry_after):
            seconds = int(retry_after)
        else:
            retry_date_tuple = email.utils.parsedate(retry_after)
            if retry_date_tuple is None:
>               raise InvalidHeader("Invalid Retry-After header: %s" % retry_after)
E               InvalidHeader: Invalid Retry-After header: 1m0s
```

Even more interesting, the `parsedate` is now no longer part of the stdlib, https://github.com/urllib3/urllib3/blob/728d9244665ef5b03103cb74d7b409ebe4f23b43/src/urllib3/util/retry.py#L254

pquentin:
Thanks for opening an issue and a pull request! I have a few questions about the issue.

Why would email.utils.parsedate not be part of the standard library? I see it there: https://docs.python.org/3/library/email.utils.html#email.utils.parsedate

1m0s is indeed not a compliant Retry-After header: https://tools.ietf.org/html/rfc7231#section-7.1.3. Do you have a simple way to reproduce the GitHub issue? Did you try submitting a bug to them? They should just send "60" in that case.

Regarding your pull request, urllib3 does not depend on external libraries. We can vendor small libraries, but there needs to be a real use case for this.

ssbarnea:
I was sure I got an error about missing parsedate on newer version of python but apparently is was not removed https://github.com/python/cpython/blob/master/Lib/email/utils.py -- maybe it was caused by some shadowing from another library or relative path imports. I am glad to use it as adding extra dependency for this is a very bad idea.

I did not raise a bug with github yet, even if my hopes about them doing anything about it are very low. Here is the place where I found the issue https://review.rdoproject.org/zuul/build/5c23a6536eff46778c91beb75598c319 and the code generating it can be seen at https://review.rdoproject.org/r/#/c/24100/9/test/test_verify.py

I was still unable to reproduce this error locally, I guess i will have to do much more DDOS requests to trigger the 429 response from github.

[...]

PR #1823, opened 2020-03-19
Fix parse_retry_after exception

ssbarnea:
Fixes: #1822

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ setup(
         "urllib3.util",
     ],
     package_dir={"": "src"},
-    requires=[],
+    requires=["dateparser"],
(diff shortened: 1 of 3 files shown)


sethmlarson:
Given the comment by @pquentin in #1822 I don't think this is needed.

ssbarnea:
@sethmlarson Please reopen it so we can test it. Microsoft/GitHub did not reply at all and I am trying to find an acceptable workaround.

sethmlarson:
The issue is not in urllib3, if you can find a work-around that works for you locally you can keep that fix in your project. You should be able to run the test suite without our CI.

PR #1827, opened 2020-03-23
Fix #1003: Add Exception for Name Resolution

### LINKED ISSUE #1003: DNS lookup error exception

kobayashi:
This is fix of #1003 
We may need more discuss to add the exception?

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
+import socket
 from .packages.six.moves.http_client import IncompleteRead as httplib_IncompleteRead
 
 # Base Exceptions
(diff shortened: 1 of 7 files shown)


pquentin:
Not sure test_source_address_error fails on macOS and Windows, do you have an idea why?

src/urllib3/exceptions.py
```diff
@@ -258,3 +258,11 @@ def __init__(self, defects, unparsed_data):
 class UnrewindableBodyError(HTTPError):
     "urllib3 encountered an error when trying to rewind a body"
     pass
+
+
+class NameResolutionError(HTTPError):
```
sethmlarson: Could we multiclass this error as `HTTPError, socket.gaierror` so that code paths catching DNS errors aren't broken?

src/urllib3/util/connection.py
```diff
@@ -58,8 +58,12 @@ def create_connection(
     # The original create_connection function always returns all records.
     family = allowed_gai_family()
 
-    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
-        af, socktype, proto, canonname, sa = res
+    try:
+        res = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+    except socket.gaierror as e:
+        raise e
```
sigmavirus24, line 64: Is this really what we want?

sethmlarson:
I think this PR has stalled out. Closing for now.

PR #1828, opened 2020-03-23
Test that ConnectionPool preserves path dot segments

### LINKED ISSUE #1790: Issue with Parsing URIs - Breaks Security Tools when testing for Path Traversal

hodbn:
Closes #1790

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -67,7 +67,9 @@ class ConnectionPool(object):
     :class:`.HTTPConnectionPool` and :class:`.HTTPSConnectionPool`.
 
     .. note::
-       When used directly, does not normalize URLs.
+       ConnectionPool.urlopen() does not normalize or percent-encode target URIs
+       which is useful if your target server doesn't support percent-encoded
+       target URIs.
     """
 
     scheme = None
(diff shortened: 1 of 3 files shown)


src/urllib3/connectionpool.py
```diff
@@ -65,6 +65,9 @@ class ConnectionPool(object):
     """
     Base class for all connection pools, such as
     :class:`.HTTPConnectionPool` and :class:`.HTTPSConnectionPool`.
+
+    .. note::
```
sethmlarson: Can we change the message to "ConnectionPool.urlopen() does not normalize or percent-encode target URIs which is useful if your target server doesn't support percent-encoded target URIs". What do you think?

hodbn: Seems good :)

PR #1829, opened 2020-03-28
WIP: 555 content length 0

### LINKED ISSUE #555: Content-length header not set on PUT/POST with empty body
Lots of discussion over at Requests: kennethreitz/requests/issues/223 including a PR to work around the issue: kennethreitz/requests/pull/957

Basically, nginx is returning 411 responses when using urllib3 to make PUT or POST requests with an empty body because the content-length header is only set when the body isn't empty.

We're encountering this issue with the elasticsearch-py library, which is relying on urllib3 to set this header appropriately (see elasticsearch/elasticsearch-py@20aa35d791bbcfca555ef059ea1c619019a93ff7)

I've had a poke around the code but can't find the appropriate place to apply a fix for this.
##

eduzen:
Fixes #555 

Hi, this is my first contribution to urllib3. I'm a little bit lost. I read the suggestions in the thread and the suggestion to update `request.py`. But I'm not sure if this is the right approach. I'm open to suggestions and obviously how to test it.

Thanks!

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -294,5 +294,8 @@ In chronological order:
 * Himanshu Garg <garg_himanshu@outlook.com>
   * DOC & LICENSE Update
 
+* Eduardo Enriquez <eduardo.a.enriquez@gmail.com>
+  * Added content-length default value
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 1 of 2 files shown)


pquentin:
@eduzen Yay, we love first contributions! I've asked @hodbn to look at your pull request. It may take some time: even if it's a small change, there's a lot of context here, and we need to make sure that the change is valid and tested.

Are you familiar enough with git to be able to rebase your pull request on top of the master branch? You have a conflict in CONTRIBUTORS.txt.

eduzen:
Hi @pquentin it seems that the issues is not longer valid! thanks for your quick response. I will close it now.

PR #1831, opened 2020-03-30
Regenerate dummyserver/certs with trustme

pquentin:
It would have been better to generate them in tests/conftest.py, but it's quicker to do it that way and since those new certs are compatible with macOS 10.15 they will allow me to debug a SecureTransport issue.


pquentin:
Ideally, yes. However:

- preparing this pull request took like 15 minutes since we only have two certificates left
- removing those certs from git is likely to take one or two full days of work as I need to reorganize a numbet of tests and figure out how to generate the fingerprints (and the hash without using c_rehash)
- this pull request fixes my immediate issue (SecureTransport tests now run on my machine)

As a result, it won't be a priority anymore, and I'm unlikely to work on it anytime soon, if at all. I'd be happy to review pull requests, though.

PR #1832, opened 2020-03-30
Restore coverage measurements

pquentin:
The idea is to generate coverage.xml explicitly, and ask codecov to upload this file. This reduces possible errors significantly. For example, it avoids a case where codecov uploaded a coverage.py binary!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /dist
 /build
 /docs/_build
+coverage.xml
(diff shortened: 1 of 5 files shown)


pquentin:
Coverage is 100% on codecov, not sure why we don't see a codecov comment here.

https://codecov.io/gh/urllib3/urllib3/tree/183da1878a87633b27535b2afaebd5d483d5d95c

PR #1554, opened 2019-03-19
Miscellaneous test improvements

pquentin:
* Stop excluding various Python 3 branches
* Bump timeouts to help with CI flakiness
* Ensure test_create_connection_timeout can be run alone

PR #1833, opened 2020-03-30
Improve tests

hodbn:
TODO:
- [x] add a missing test
- [X] better exclusion rule for coverage
- [X] annotate Python version-specific code
- [X] annotate platform-specific code
- [X] deterministic and faster `get_unreachable_address`
- [X] deprecate dead appengine code
- [ ] add a context to `find_unused_port` to avoid reuse of the 'free' port
- [ ] investigate failing https tests (on my machine due to short timeout)

Coverage is now 100% on my machine :)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -24,6 +24,6 @@ exclude_lines =
     import
     raise
     .* # Platform-specific.*
-    .*:.* # Python \d.*
+    .*:.* # Python ${PY_OTHER_MAJOR}.*
     .* # Abstract
     .* # Defensive:
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -7,8 +7,6 @@ from __future__ import print_function
 
 import logging
 import os
-import random
-import string
 import sys
 import threading
 import socket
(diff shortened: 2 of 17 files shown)


hodbn:
Hi,

Thanks for the feedback, I will close this PR and split it to several smaller ones.

Regarding the `# Python 2/3` annotation, in some cases it improves coverage checks.
For example, this line was excluded from coverage because of the `# Platform-specific` annotation:
https://github.com/urllib3/urllib3/blob/c559dce1315c1c5b8c48dd227d70796118d7a044/src/urllib3/connectionpool.py#L447
Now, coverage running on python2.7 will include this line:
https://github.com/urllib3/urllib3/blob/bb39f50d0452d44f1dc5ad8fd4db5cf13a14d0d4/src/urllib3/connectionpool.py#L439

pquentin:
While PY_OTHER_MAJOR is an interesting hack, if we stick with codecov, we probably want to remove the `# Python ...` comments altogether, don't we? Like this: https://github.com/urllib3/urllib3/pull/1554/commits/c560e25e96469d923bf3ca6dc5afc91b7b5400d9

Maybe I'm not understanding the remark correctly?

Closing, but we can continue discussing this here.

hodbn:
Yeah, we should consider removing the comments altogether, or removing the exclusion rule from `.coveragerc` and leaving the comments where needed for readability.

PR #1830, opened 2020-03-28
Fix retry to account for errors inside a ProxyError

hodbn:
Closes #1744

PR #1837, opened 2020-03-31
Add context to find_unused_port

hodbn:
The helper function `find_unused_port` might return a port that will immediately be reused by the system.
I wrapped the helper in a `contextmanager` so the port is allocated to `tempsock` while inside the context.

This PR affects #1830, I will update it/this PR once one of them is accepted.

### DIFF
--- a/test/port_helpers.py
+++ b/test/port_helpers.py
@@ -2,6 +2,7 @@
 # library test suite.
 
 import socket
+from contextlib import contextmanager
 
 
 # Don't use "localhost", since resolving it uses the DNS under recent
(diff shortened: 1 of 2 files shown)


pquentin:
The function still says: `The temporary socket is then closed and deleted, and the ephemeral port is returned.`

Obviously the tests passed fine, but how does this work? Why is it not an issue to not close the socket before starting the test?

hodbn:
The `yield` is just after the `bind_port` call and before the `close()`, so the block inside the `with` is executed while the socket is still registered and the port is reserved by the OS. I've fixed the docs :)

PR #1839, opened 2020-04-01
Switch macOS CI to GitHub Actions

pquentin:
The macOS tests are the current bottleneck, which is why I'm moving them first. I've simply copy-pasted my work from https://github.com/python-trio/hip/blob/master/.github/workflows/ci.yml. GitHub Actions is nice because it gives up to 20x parallelism and is better integrated with GitHub, however the UI is barely usable when the tests are still running. Good news: a codecov token is no longer required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  macOS:
+    runs-on: macos-latest
+    
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+
+    steps:
+      - name: Checkout Repository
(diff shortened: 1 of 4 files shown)


pquentin:
I guess I should have opened this one from a branch in the main repo, as GitHub Actions can't be activated on the first pull request.

You can see the tests running here: https://github.com/pquentin/urllib3/actions/runs/68378346

sethmlarson:
Also I enabled Actions, they were previously disabled. Maybe that's why things didn't show up here?

pquentin:
Hm, did not work yet. Let's try merging this, we can always revert.

pquentin:
It worked! https://github.com/urllib3/urllib3/actions/runs/68722383 We'll need to update our badges list. Not sure if Travis can go fully away as I don't know how to run PyPy or 3.9-dev in Actions.

PR #1843, opened 2020-04-07
Revert "Add context to find_unused_port (#1837)"

pquentin:
This reverts commit a9508d8d8ac58ed95bf06d8970382664af4e91e3.

For some reason, this commit was making the tests really, really slow on macOS. For example, `test_proxy_connect_retry` went from nearly instant to 25 seconds on my computer. As a result, macOS tests in CI went from ~3 minutes to ~8 minutes. I thought this was because of GitHub Actions, but in fact the above commit was the last where tests ran on macOS. GitHub Actions is fine.

Sorry about that, @hodbn, but practicality beats purity here.

### DIFF
--- a/test/port_helpers.py
+++ b/test/port_helpers.py
@@ -2,7 +2,6 @@
 # library test suite.
 
 import socket
-from contextlib import contextmanager
 
 
 # Don't use "localhost", since resolving it uses the DNS under recent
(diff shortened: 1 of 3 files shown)


pquentin:
The build failure is fixed in https://github.com/urllib3/urllib3/pull/1844

PR #1844, opened 2020-04-07
Fix documentation build by skipping Sphinx 3.0.0

pquentin:
Sphinx 3 currently fails with an mysterious ValueError: `wrapper loop
when unwrapping <MockModule id='139897808864312'>`

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
 -r ../dev-requirements.txt
-sphinx<3
+sphinx!=3.0.0
 alabaster
 requests>=2,<2.16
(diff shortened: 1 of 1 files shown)


sethmlarson:
Give @hodbn's comment a try but if it doesn't work we can pin to 2.x. We don't gain a lot from 3.x from their changelog.

pquentin:
This is a bug in Sphinx that will [be fixed in 3.0.1](https://github.com/sphinx-doc/sphinx/pull/7431) so I just skip 3.0.0 for now.

pquentin:
I fixed the macOS flaky test in https://github.com/urllib3/urllib3/pull/1846, please consider merging this as is: the docs CI passed.

PR #1845, opened 2020-04-07
URI parsing: disallow backslash in the authority part

hodbn:
The backslash character is not allowed in the authority part of a URI by [rfc3986](https://tools.ietf.org/html/rfc3986#appendix-A). Including it might even lead to [dangerous cases](https://bugs.xdavidhu.me/google/2020/03/08/the-unexpected-google-wide-domain-check-bypass/).

As far as I can understand, the [example regex in rfc3986](https://tools.ietf.org/html/rfc3986#appendix-B) allows the character, but we probably still want to disallow it.

Pretty much the same as https://github.com/python-hyper/rfc3986/pull/64.


sigmavirus24:
This looks okay to me.

The failing Travis build seems related to https://github.com/urllib3/urllib3/pull/1844

Issue #1850, opened 2020-04-16
Why am I seeing InvalidProxyConfigurationWarning when using an HTTPS proxy?


sethmlarson:
urllib3 up to v1.25.x doesn't support HTTPS proxies. When connecting to an HTTPS URL, urllib3 contacts the proxy via HTTP instead of HTTPS **even if your proxy URL specifies HTTPS**. In urllib3 v1.26.x we're planning on supporting HTTPS proxies properly and are giving an early warning to users to switch their proxy URLs from HTTPS to HTTP to not encounter issues when upgrading later.

```python
import urllib3

# HTTPS proxy, should change!
http = urllib3.ProxyManager("https://1.2.3.4")
http.request("GET", "https://example.com")  # Warning would be raised here.

# Switch to this, will maintain current behavior when connecting to HTTPS URLs.
http = urllib3.ProxyManager("http://1.2.3.4")
http.request("GET", "https://example.com")  # Warning won't be raised, same behavior as above.
```

Your proxy may be configured externally like in a `HTTPS_PROXY` environment variable or via `requests.Session(proxy_url=...)` or configured by your OS. Look in all these places before raising an issue or commenting here.

**Please don't leave a comment unless you have something to add, this will email many people.** If you've received this warning please leave a :+1: reaction so we can gauge how widespread this issue is.

If you do believe there is an issue please open one separately and we will get to it as soon as possible. Thank you. :)

sigmavirus24:
For what it may be worth, you'll have a _lot_ of Requests users running into this potentially

sethmlarson:
@sigmavirus24 Yeah Nate is aware of all of this, I'm hoping that including a link in the warning itself will steer users to the correct landing place. :)

sigmavirus24:
Based on past warnings like this with a link, Requests users will just comment here complaining rather than following your extremely clear instructions. It may be worthwhile to consider locking comments here just to avoid the noise

sethmlarson:
That is definitely an option if things get out of hand. Will monitor the issue closely.

sethmlarson:
This issue has served its purpose now that 1.26 supports HTTPS proxies, thanks all for the :+1:s!

PR #1851, opened 2020-04-16
Add InvalidProxyConfigurationWarning for HTTPS proxies

sethmlarson:
I pre-allocated #1850 for users that are receiving the issue. I have a feeling we may need to lock the conversation if it attracts too much traffic but it's better than leaving users without an official means of solving the issue.

cc @jalopezsilva

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -470,6 +470,7 @@ class ProxyManager(PoolManager):
                 "Read this issue for more info: "
                 "https://github.com/urllib3/urllib3/issues/1850",
                 InvalidProxyConfigurationWarning,
+                stacklevel=3,
             )
 
     def urlopen(self, method, url, redirect=True, **kw):
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 import collections
 import functools
 import logging
+import warnings
 
 from ._collections import RecentlyUsedContainer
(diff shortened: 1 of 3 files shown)


joelverhagen:
Oh no, what have I done?

sethmlarson:
@joelverhagen My apologies, I meant @jalopezsilva :) But I'm sure you're up to no good! ;)

src/urllib3/poolmanager.py
```diff
@@ -459,9 +461,21 @@ def _set_proxy_headers(self, url, headers=None):
             headers_.update(headers)
         return headers_
 
+    def _validate_proxy_scheme_url_selection(self, url_scheme):
+        if url_scheme == "https" and self.proxy.scheme == "https":
+            warnings.warn(
```
pquentin, line 466: Is there a sensible stacklevel to use here?

sethmlarson: Probably! I'll add one.

sethmlarson:
I like your wording more @pquentin, was struggling to come up with something late last night. :)

src/urllib3/poolmanager.py
sethmlarson: Settled with 3 so the traceback lands on wherever `urlopen()` is called. There's not really a great way to do [multi-entrypoint warning stacklevel](https://twitter.com/sethmlarson/status/1245074746645057542). :(

pquentin: Yeah, that tweet is how I found out about stacklevel. In my case I'm using `urllib3.ProxyManager` and stacklevel=5 ends up in my code

PR #1855, opened 2020-04-16
Fix PyPI long description

### LINKED ISSUE #1854: Raw HTML in README.rst causing problems with uploading dists
See: https://travis-ci.org/github/urllib3/urllib3/builds/675807537

For now I've manually uploaded the dists.

We should strip this section from our `long_description` field and maybe run `twine check` within our CI to make sure we don't run into this issue on release time in the future.

pquentin:
The first commit runs "twine check". I reused @sethmlarson's code from hip, but simplified it since `twine check` returns a non-zero error code on failure. You can see the check working as expected here: https://github.com/pquentin/urllib3/runs/593112721?check_suite_focus=true

The second commits fixes the twine warnings by:

 1. removing the raw directive from the long description, and
 2. specifying the content type.

Parsing README.rst is quite brittle, what do you think? It works now (see the resulting PKG-INFO), but who knows how it could break in the future.

Closes #1854

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,21 @@ name: CI
 on: [push, pull_request]
 
 jobs:
+  package:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v1
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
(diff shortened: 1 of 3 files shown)


pquentin:
I hate flaky tests.

sethmlarson:
Codecov action acting up, maybe they should write it in Python instead of NodeJS ;)

PR #1852, opened 2020-04-16
Prepare v1.25.9

sethmlarson:
Blocked on #1851 but after that is merged and we add a release date to this PR we should be good to go :tada:

PR #1856, opened 2020-04-17
Check README.rst and CHANGES.rst are valid

pquentin:
This is a follow-up of #1852, where @sethmlarson asked: 

>  I agree the RST parsing is brittle, at least there is now CI checking for valid RST?

The twine check was only superficial. Now we do check for validity. :)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,5 @@
 .. raw:: html
+
    <p align="center">
       <a href="https://github.com/urllib3/urllib3">
          <img src="./docs/images/banner.svg" width="60%" alt="urllib3" />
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -74,13 +75,13 @@ Tidelift will coordinate the fix and disclosure with maintainers.
...
 -----------
 
-- `@sethmlarson <https://github.com/sethmlarson>`_ (Seth M. Larson)
-- `@pquentin <https://github.com/pquentin>`_ (Quentin Pradet)
-- `@theacodes <https://github.com/theacodes>`_ (Thea Flowers)
-- `@haikuginger <https://github.com/haikuginger>`_ (Jess Shapiro)
-- `@lukasa <https://github.com/lukasa>`_ (Cory Benfield)
-- `@sigmavirus24 <https://github.com/sigmavirus24>`_ (Ian Stapleton Cordasco)
-- `@shazow <https://github.com/shazow>`_ (Andrey Petrov)
+- `@sethmlarson <https://github.com/sethmlarson>`__ (Seth M. Larson)
```
sethmlarson, line 78: I've never seen this construction before, hm

pquentin: Yes, it did. Like the double colon, it's a reST subtlety, see https://stackoverflow.com/questions/27420317/restructured-text-rst-http-links-underscore-vs-use

pquentin:
No, twine checks if [readme_renderer](https://github.com/pypa/readme_renderer) (the PyPI renderer) does not complain when rendering the reST. It also uses docutil, but the checks appear to be more lenient. Not sure about the precise differences.

I'll try to see if I can check the generated reST, but it probably requires parsing PKG-INFO.

sethmlarson:
Yeah let's not do that, I think this combined with the twine check are enough to prevent mistakes in the future. Thanks much! :D

PR #1857, opened 2020-04-18
Add CODEOWNERS restricting access to files related to deployment

sethmlarson:
Basically want to reduce the current surface of accounts that'd be required to either:
- Make a malicious release via our automated deploy
- Dump the `PYPI_PASSWORD` from Travis by merging into master

The first is thwarted by making `setup.py` and `_version.py` protected files. That way even if an attacker commits changes and tags a commit on `master` the upload would fail due to there being a duplicate version.

The second is thwarted by disallowing changes to scripts that have access to `PYPI_PASSWORD` on Travis to require admin approval.

In total this reduces our attack surface from requiring access to 2 committer accounts or 1 admin account to only the 1 admin account scenario, and in that case we're pretty hosed anyways. :)

### DIFF
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -14,7 +14,7 @@ from .util.request import make_headers
 from .util.url import get_host
 from .util.timeout import Timeout
 from .util.retry import Retry
-from ._version import __version__ as __version__
+from ._version import __version__
 
 
 # Set default logging handler to avoid "No handler found" warnings.
@@ -23,6 +23,7 @@ from logging import NullHandler
 
 __author__ = "Andrey Petrov (andrey.petrov@shazow.net)"
 __license__ = "MIT"
+__version__ = __version__
 
 __all__ = (
     "HTTPConnectionPool",
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -14,6 +14,7 @@ from .util.request import make_headers
 from .util.url import get_host
 from .util.timeout import Timeout
 from .util.retry import Retry
+from ._version import __version__ as __version__
 
 
 # Set default logging handler to avoid "No handler found" warnings.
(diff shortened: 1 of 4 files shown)


src/urllib3/__init__.py
```diff
@@ -22,7 +23,6 @@
 
 __author__ = "Andrey Petrov (andrey.petrov@shazow.net)"
 __license__ = "MIT"
-__version__ = "1.25.8"
```
pquentin, line 25: We have a lint failure here. Could it be fixed  using `__version__ = __version__`? It would have the benefit of being explicit too.

sethmlarson:
I saw that... let me cycle? That's real strange.

src/urllib3/__init__.py
sethmlarson: Will change to that!

pquentin:
The second build worked fine, but the third one gave another strange error (coverage run error code -14). Any reason not to push your fix?

sethmlarson:
@pquentin Was on mobile but still wanted a restarted build :) Will push now.

sethmlarson:
Now we're getting funky coverage issues, the Google Brotli builds on Travis both passed and uploaded coverage data though, might be related to the PR I just merged on Brotli?

sethmlarson:
That may have been it because now our coverage and CI are happy.

pquentin:
Yeah it's definitely what happened, but I don't understand why. Anyway, it's good that this got sorted out.

Now, I tried reading about CODEOWNERS, is it expected that this pull request does not add such a file? The comments are enough to activate the protection?

sethmlarson:
@pquentin Haha, our `.gitignore` has a clause (`.*`) that was preventing adding `.github/CODEOWNERS` automatically. Thanks for catching that!

pquentin:
Good, I see the lock on https://github.com/urllib3/urllib3/blob/master/setup.py as expected. I guess you now need to enable "Require review from Code Owners" in the settings: https://help.github.com/en/github/administering-a-repository/enabling-required-reviews-for-pull-requests

sethmlarson:
I've already done so, many thanks for the reviews! 🌟

PR #1858, opened 2020-04-19
Avoid hasattr in hot loop in brotli decoder

auscompgeek:
`hasattr` has a nasty performance penalty when it returns false: in the background it is actually a try/except AttributeError.  Since we know the method we want won't change, by instead caching the bound method object we can amortise the cost across an entire encoded body.

This removes the intermediate `decompress` method altogether, instead giving the underlying method the `decompress` name.


auscompgeek:
I haven't had a chance to actually benchmark this yet. I might get an opportunity to do so later this week, but no guarantees.

PR #1860, opened 2020-04-19
doco: Fix small typos

timgates42:
Fixes a handful of typos found.

PR #1861, opened 2020-04-20
Add a spelling check to Travis CI checks

timgates42:
This is the spelling check I run in my fork. After a discussion in #1860 I thought I might prepare a PR to open a wider discussion about whether a spelling CI check might be worth including in the primary urllib3 repository. I am keen to hear peoples thoughts and I'm happy to hear "thanks but no thanks this is not for us".

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+import subprocess
 
 import nox
 
@@ -103,3 +104,11 @@ def docs(session):
     if os.path.exists("_build"):
         shutil.rmtree("_build")
     session.run("sphinx-build", "-W", ".", "_build/html")
+
+
+@nox.session(python=["3.8"])
+def spelling(session):
+    subprocess.call(["sudo", "apt-get", "update"])
+    subprocess.call(["sudo", "apt-get", "install", "-y", "aspell", "aspell-en"])
+    session.install("spelling")
+    session.run("python", "-m", "spelling")
--- a/noxfile.py
+++ b/noxfile.py
@@ -108,7 +108,5 @@ def docs(session):
 
 @nox.session(python=["3.8"])
 def spelling(session):
-    subprocess.call(["sudo", "apt-get", "update"])
-    subprocess.call(["sudo", "apt-get", "install", "-y", "aspell", "aspell-en"])
     session.install("spelling")
     session.run("python", "-m", "spelling")
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,6 +1,5 @@
 import os
 import shutil
-import subprocess
 
 import nox
 
@@ -86,13 +85,15 @@ def blacken(session):
 
 @nox.session
 def lint(session):
-    session.install("flake8", "black")
+    session.install("flake8", "black", "spelling")
     session.run("flake8", "--version")
     session.run("black", "--version")
+    session.run("spelling", "--version")
     session.run(
         "black", "--check", "src", "dummyserver", "test", "noxfile.py", "setup.py"
     )
     session.run("flake8", "setup.py", "docs", "dummyserver", "src", "test")
+    session.run("spelling")
 
 
 @nox.session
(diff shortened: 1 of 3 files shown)


timgates42:
I'd imagine at a minimum I should update the documentation to include some information about the spelling CI checks but this really is only worthwhile if a spelling check is considered a good thing.

pquentin:
That sounds fine to me, what are your thoughts @sethmlarson?

@timgates42 Ideally this is something we'd want to put in GitHub Actions CI (https://github.com/urllib3/urllib3/blob/master/.github/workflows/ci.yml), but we can do that as a second step.

pquentin:
(I'm going to approve to see if the recent CODEOWNERS change works well.)

noxfile.py
```diff
@@ -103,3 +104,11 @@ def docs(session):
     if os.path.exists("_build"):
         shutil.rmtree("_build")
     session.run("sphinx-build", "-W", ".", "_build/html")
+
+
+@nox.session(python=["3.8"])
+def spelling(session):
+    subprocess.call(["sudo", "apt-get", "update"])
```
sigmavirus24, line 111: If someone tries to run this nox session on a non Debian Derived Linux or macOS or Windows, this will not work. Is there another way to get this functionality?

timgates42: Thanks for pointing this out @sigmavirus24 - I’ll do some research and see what options
there are.

timgates42:
Thanks @sethmlarson I've been doing some research and I think I can achieve this. I can also keep running this in my fork for a while and submit PRs for any spelling fixes found for a while to gain some trust. Once I've got something going in the lint task via pip I'll report back.

timgates42:
Thanks @sethmlarson for the great suggestion, it is looking a lot nicer now. I'll keep monitoring the false positives in my fork and let you know how I go.

pquentin:
So the current solution is to install aspell directly from your spelling package if the CI is detected: https://github.com/resplendent-dev/spelling/commit/0ee76f5713e607e72671e6ff9290d4b8a2d5ba76. I'm not sure what to think of this. I guess it allows centralizing installation procedures, but I'm not comfortable with using sudo in such a hidden place... I think installing aspell explicitly in our Travis code is better, and spelling should probably fail if aspell isn't installed. @sigmavirus24 @sethmlarson Thoughts?

Another issue is that we don't want to put this in the lint command because this is only an experiment and we don't want to have to install aspell to run the lint command.

timgates42:
Thanks @pquentin that’s a good point. I shouldn’t make it so implicit. I’ll add an explicit install step.

sigmavirus24:
> I'm not comfortable with using sudo in such a hidden place

I agree, it seems misleading to include this in the project that way, especially since it supports "CI" but not all CI (not azure pipelines, wouldn't work on an alpine container for testing, wouldn't work on a macOS VM, etc.) and I'm not sure that the added time of running those installation commands from the running process looks good for the spelling checker.

sethmlarson:
I'm -1 on having OS package managers being involved.

timgates42:
Thanks all - I'll close my PR for now and come back if I can remove the use of package managers entirely

sethmlarson:
Thanks for trying to get us a good solution here @timgates42, it's nice to have an automated solution find these issues rather than getting nickled and dimed by typo issues over time.

pquentin:
@timgates42 Let me try to clarify our position here. There's a tension between making contributions easy and fixing typos. Fixing typos is a nice addition, but after all your tool only found four of them, so we don't have a huge problem here.

However, making contributions easy is a harder problem, and we don't want to introduce any artificial barriers. So the bar to add a new check is quite high: all contributors should be able to execute that check locally, ideally without any other dependencies than Python. Given what I understand now about the problem, I'm not sure this is a good use of your time, as embedding aspell or hunspell in a Python package that works on Windows, macOS and Linux is probably quite difficult.

timgates42:
Thanks @pquentin that makes perfect sense, thanks all for your suggestions they have given me some good ideas even if not applicable for urllib3.

PR #1864, opened 2020-04-26
WIP: Add fixture for CA

eduzen:
Hi, I'm trying to help with #1705 I'm open to all suggestions!

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -300,5 +300,8 @@ In chronological order:
 * Chris Olufson <tycarac@gmail.com>
   * Fix for connection not being released on HTTP redirect and response not preloaded
 
+* Eduardo Enriquez <eduardo.a.enriquez@gmail.com>
+  * Test improvements with trustme
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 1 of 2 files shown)


eduzen:
@pquentin  I was trying to understand the problem. Since the certificates are hardcoded in the repo, the fingerprints are too. So if we switch them to trustme.CA we need to drop these fingerprints. I'm not familiar with fingerprints can we somehow extract these fingerprints from the trustme ones?

pquentin:
trustme uses [cryptography](https://cryptography.io/en/latest/) under the hood, which [gives access to fingerprints](https://cryptography.io/en/latest/x509/reference/?highlight=fingerprint#cryptography.x509.Certificate.fingerprint). We either want to add support for getting fingerprints in trustme, or we'll need to load certificates from the PEM bytes.

@njsmith Would you accept a pull request that exposes fingerprints as part of trustme? Or is out of scope? We can also give access to the underlying cryptography object, or offer an easy way to get it from the PEM bytes?

njsmith:
No particular objections, though urllib3 already has its own code to compute a fingerprint from a cert, right? Would it be simpler to just use that?

pquentin:
Hm, yes, we do have `assert_fingerprint`.

@eduzen Since you're in charge here :) How would you like to proceed?

eduzen:
Ok I agree. I will use assert_fingerprint for this change. And maybe next try to open a pull for trustme. Thank you @pquentin !

pquentin:
Closing for now as this work appears to have stalled. Feel free to reopen!

PR #1866, opened 2020-04-26
Do not put a None in a connection pool if it was empty

### LINKED ISSUE #1758: Established connections are replaced by None values
Hello!

When pool is in blocking mode, there is no connections in pool, and getting connection from pool is timed out.

https://github.com/urllib3/urllib3/pull/1759

hodbn:
Trying to resolve #1758 in a different approach from #1759 .
I think that there are two different issues.

First, ```HTTPConnection.urlopen``` should not except a ```queue.Empty``` error since these might only be raised in tests:
https://github.com/urllib3/urllib3/blob/898a16d09e4a6d9dbe10134a49b89eedfe8dae7f/test/test_connectionpool.py#L282
In an unpatched flow, ```HTTPConnection.urlopen``` might only raise an ```EmptyPoolError``` inside ```HTTPConnection._get_conn```:
https://github.com/urllib3/urllib3/blob/898a16d09e4a6d9dbe10134a49b89eedfe8dae7f/src/urllib3/connectionpool.py#L261-L267

Secondly, ```HTTPConnection.urlopen``` should not put back a ```None``` in the pool in case of an ```EmptyPoolError```. If it does, then the situation @jaimefrites described might occur.

AFAICT the test I removed in ```test_pool_size``` is not relevant since in real-life scenario the pool is never changed.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -698,9 +698,6 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # Everything went great!
             clean_exit = True
 
-        except queue.Empty:
-            # Timed out by queue.
-            raise EmptyPoolError(self, "No pool connections are available.")
 
         except (
             TimeoutError,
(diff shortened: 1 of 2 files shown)


pquentin:
I find it difficult to wrap my head around this issue... @jaimefrites what do you think of this version? Does that fix your issue too?

jaimefrites:
@hodbn @pquentin  Yes, this solves the issue and cleaner. Thank you

PR #1867, opened 2020-04-28
Feature/support env var sslkeylogfile

bastiaanb:
Adds support for creating a session key logfile by specifying environment variable `SSLKEYLOGFILE`. 
Depends on Python 3.8+, see https://bugs.python.org/issue34271

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import
 import errno
 import warnings
 import hmac
+import os
 import sys
 
 from binascii import hexlify, unhexlify
@@ -293,6 +294,12 @@ def create_urllib3_context(
         # We do our own verification, including fingerprints and alternative
         # hostnames. So disable it here
         context.check_hostname = False
+
+    # Enable logging of TLS session keys via defacto standard environment variable
+    # 'SSLKEYLOGFILE', if the feature is available (Python 3.8+).
+    if hasattr(context, "keylog_filename"):
+        context.keylog_filename = os.environ.get("SSLKEYLOGFILE")
+
     return context
 
 
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -703,6 +703,10 @@ class TestHTTPS(HTTPSDummyServerTestCase):
         reason="requires OpenSSL 1.1.1+",
     )
     @pytest.mark.skipif(sys.version_info < (3, 8), reason="requires python 3.8+")
+    @pytest.mark.xfail(
+        sys.platform == "win32",
+        reason="does not work reliably in Appveyor test enviroment for not yet known reasons"
+    )
     def test_sslkeylogfile(self, tmpdir, monkeypatch):
         keylog_file = tmpdir.join("keylogfile.txt")
         monkeypatch.setenv("SSLKEYLOGFILE", str(keylog_file))
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -698,6 +698,27 @@ class TestHTTPS(HTTPSDummyServerTestCase):
             finally:
                 conn.close()
 
+    def test_sslkeylogfile(self, tmpdir):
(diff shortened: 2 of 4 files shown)


test/with_dummyserver/test_https.py
```diff
@@ -698,6 +698,27 @@ def test_tls_protocol_name_of_socket(self):
             finally:
                 conn.close()
 
+    def test_sslkeylogfile(self, tmpdir):
+        if not hasattr(ssl.SSLContext, "keylog_filename"):
+            pytest.skip("SSLContext does not support keylog_filename")
+        try:
+            keylog_file = tmpdir.join("keylogfile")
+            os.environ["SSLKEYLOGFILE"] = str(keylog_file)
```
pquentin, line 706: Can we use monkeypatch.setenv from pytest instead?

src/urllib3/util/ssl_.py
```diff
@@ -293,6 +294,12 @@ def create_urllib3_context(
         # We do our own verification, including fingerprints and alternative
         # hostnames. So disable it here
         context.check_hostname = False
+
+    # Enable logging of TLS session keys via defacto standard environment variable
+    # 'SSLKEYLOGFILE', if the feature is available (Python 3.8+).
+    if hasattr(context, "keylog_filename"):
```
sethmlarson, line 300: Can we also add a check that the environment variable is "truthy" before assigning? In case users do `SSLKEYLOGFILE=`.

test/with_dummyserver/test_https.py
bastiaanb: ah, yes, monkeypatch is much nicer. Done.

src/urllib3/util/ssl_.py
bastiaanb: I don't think checking would add any value: the ssl code has to do input validation / graceful failure handling anyway, e.g. in case of non-existent or unwritable paths. It already ignores an empty value.

bastiaanb:
> This looks good! Any idea why this is failing on Windows?

No, I don't work on Windows, so I was hoping somebody else knows.

bastiaanb:
> This is a great feature to have and many users will be happy about.
> 
> Can we also add documentation about what the var does and maybe point to a guide on how to configure it within Wireshark for example?

I will add docs tomorrow.

bastiaanb:
Hi, 
I've been able to successfully run the tests on a Windows 10 x86_64 machine with Python 3.8.2. So I cannot reproduce the issue. 
Can someone from the project please look into this?

bastiaanb:
How can we move this forward? 
Would it be an option to declare this a "your mileage may vary" feature for Windows?
Then I'll document that and disable the checks on Windows. 
That way at least we have the feature on other platforms.

sethmlarson:
I think I'm fine with merging this and mark as xfail on Windows. Not our fault that Windows is not working properly. Thanks for chasing this down.

pquentin:
However, CPython tests this feature on Windows (except for debug builds) and documents that it works on all platforms. Something weird is going on here, and it would be nice to understand it

bastiaanb:
@pquentin of course it is good to understand what is going on. What do you suggest as a path forward? 
I cannot reproduce the problem on a local Windows 10 machine (test succeeds).
Who can and will look at the AppVeyor environment? 
Releasing as is will at least deliver the feature to most platform, and may attract a Windows user to investigate if it does not work reliably on Windows....

pquentin:
The remaining failures are not your fault, our build is broken: https://travis-ci.org/github/urllib3/urllib3/jobs/692095744

bastiaanb:
@pquentin Thanks for clearing that up, I was getting a bit frustated with making the build pass ;-) 
So the PR is ready for merging then?

pquentin:
No, it's blocked on https://github.com/urllib3/urllib3/pull/1882, sorry

pquentin:
Closing/reopening to get the fix from master

pquentin:
Looks like this now passes on AppVeyor! Can you please try removing the xfail?

pquentin:
(If it's not reliable on Appveyor we want use skip, not xfail)

bastiaanb:
Thanks!

Issue #260, opened 2013-10-16
More granular retry support


kevinburke:
I'm not sure what the interface for this would look like, or even at what level this would be implemented, but essentially, more function-level control over retries would be awesome. This is roughly the behavior we build in around a `requests.request`:
- We want to retry pretty much everything on idempotent requests that's not a 2xx level response - for GET, PUT, DELETE methods, because these are safe to retry.
- We want to retry **all** connection errors and connection timeouts - these indicate the request never made it to the other machine. We usually use an exponential backoff here, in the case of network failure
- In general, we don't want to retry failed POST requests, whether they were timeouts, closed connections, or 5xx errors - consider a customer getting billed multiple times, or duplicate accounts being created, or something else bad.

Again, this might not be appropriate to build in at the library level (maybe we need better primitives for this internally) but adding all this logic around our http requests leads to pretty gnarly code. 

Some kind of  `Retries` object would be nice, especially because it mirrors the `Timeout` interface and there are a lot of things you could specify if you wanted to run wild with it, say the following:
- `connection_retries`, the number of times to retry a connection error
- what i'd like to call `read_retries` or the number of times to retry in a situation where a connection **was** made to the server. a timeout, a closed connection, or an unacceptable HTTP error code, like 500
- `retry_codes`, a list of integers representing HTTP status codes, could also have named constants for `NON_200`, `5XX_ERROR`, etc, would default to 500 range
- `method_whitelist`, a list of HTTP methods to retry, defaults to HEAD, GET, PUT, DELETE, OPTIONS, TRACE, per the [spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)
- `backoff_factor`, some kind of multiplier to control how fast backoff occurs. Defaults to 0. Algorithm would be something like `backoff_factor * (2 ** (retry_attempt_number))`, so 1, 2, 4, 8...

PR #1868, opened 2020-05-01
set minimal six version to fix urllib3.exceptions.LocationParseError

doomedraven:
Hello i just installed ubuntu 20.04 and saw this issue, to fix it it requires `six>=1.12.0`

original error
```
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 380, in prepare_url
    scheme, auth, host, port, path, query, fragment = parse_url(url)
  File "/usr/lib/python3/dist-packages/urllib3/util/url.py", line 392, in parse_url
    return six.raise_from(LocationParseError(source_url), None)
  File "<string>", line 3, in raise_from
urllib3.exceptions.LocationParseError: Failed to parse: http://127.0.0.18000/
```

debugging i found that it caused by
https://github.com/urllib3/urllib3/blob/263da24d4b5c33392772e638b12d02e5b07abf9f/src/urllib3/util/url.py#L217
```
module 'six' has no attribute 'ensure_text' Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/urllib3/util/url.py", line 385, in parse_url
    path = _encode_invalid_chars(path, PATH_CHARS)
  File "/usr/lib/python3/dist-packages/urllib3/util/url.py", line 217, in _encode_invalid_chars
    component = six.ensure_text(component)
AttributeError: module 'six' has no attribute 'ensure_text
```

as you can see https://github.com/benjaminp/six/blob/master/CHANGES
```
1.12.0
------

- Issue #259, pull request #260: `six.add_metaclass` now preserves
  `__qualname__` from the original class.

- Pull request #204: Add `six.ensure_binary`, `six.ensure_text`, and
  `six.ensure_str`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,7 @@ requires-dist =
     ipaddress; python_version=="2.7" and extra == 'secure'
     PySocks>=1.5.6,<2.0,!=1.5.7; extra == 'socks'
     brotlipy>=0.6.0; extra == 'brotli'
+    six>=1.12.0
 
 [tool:pytest]
 xfail_strict = true


pquentin:
six is vendored in urllib3, see https://github.com/urllib3/urllib3/blob/master/src/urllib3/packages/six.py. Note that the vendored version includes `ensure_text`.

I think Ubuntu decided to unvendor the dependency, but included an outdated version. I believe this is an Ubuntu/Debian bug?

doomedraven:
ah ok thanks, yes looks like

Issue #872, opened 2016-05-23
PoolManager.request() with invalid scheme produces KeyError


haikuginger:
```
>>> pm = urllib3.PoolManager
>>> pm.request('GET', 'FAKESCHEME://github.com')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/urllib3/request.py", line 69, in request
    **urlopen_kw)
  File "/usr/local/lib/python3.5/site-packages/urllib3/request.py", line 90, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "/usr/local/lib/python3.5/site-packages/urllib3/poolmanager.py", line 155, in urlopen
    conn = self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
  File "/usr/local/lib/python3.5/site-packages/urllib3/poolmanager.py", line 128, in connection_from_host
    pool = self._new_pool(scheme, host, port)
  File "/usr/local/lib/python3.5/site-packages/urllib3/poolmanager.py", line 87, in _new_pool
    pool_cls = self.pool_classes_by_scheme[scheme]
KeyError: 'FAKESCHEME'
```

We should be fetching by that scheme more resiliently, and raising a more specific error message - possibly a `LocationValueError`?
##

shazow:
I'm open to proposals but I feel the error and stack trace are fairly self-explanatory.

haikuginger:
Fair enough! And we don't have any record of complaints, so it's obviously not affecting too many people.

I think I'll leave this open with the Contributor Friendly tag in case someone wants to take a crack at it.

jayvdb:
Could at least be a unit test, to ensure that this exception does become a different exception in the future...?

haikuginger:
@jayvdb - since we require 100% unit test success, the test addition and code change would need to come in at the same time. 🙂

sigmavirus24:
> since we require 100% unit test success, the test addition and code change would need to come in at the same time. 🙂

Huh? New tests can be added whenever without needing to change code. New tests will not affect the 100% coverage bit or 100% unit test success. @jayvdb wants to ensure the KeyError remains a KeyError, that's the test he's proposing (as much as I don't think we need to actually test this or handle it).

haikuginger:
> @jayvdb wants to ensure the KeyError remains a KeyError, that's the test he's proposing

[...]

PR #1872, opened 2020-05-05
Raise a meaningful error for an unknown scheme

### LINKED ISSUE #1871: KeyError can be triggered by a server
I noticed a behavior in urllib3 where a server can trigger an unexpected KeyError exception.

This happens if a server causes a redirect to a protocol that is not http or https (can also be a bogus protocol).

To reproduce you need a server that creates a simple redirect, e.g. with a python CGI something like this:
```
#!/usr/bin/python3
print("location: a://b")
print("")
```

Now do a request with urllib3's poolmanager, e.g.:
```
import urllib3
pool = urllib3.PoolManager()
try:
    r = pool.request("GET", 'http://localhost/)
except (ConnectionRefusedError, ConnectionResetError,
        urllib3.exceptions.HTTPError):
    pass
```

A similar issue has already been reported in issue #872. However it was closed without a fix and the discussion there indicates that this was due to the fact that one expects programmers not to provide invalid input.
This scenario here is different as the bug can be triggered by a server, which is not necessarily under control or trusted by the application. Causing a KeyError is a very unexpected thing for an HTTP library, so I can imagine that there are situations where this can be used to crash applications with an unexpected / unhandled exception.

hodbn:
Closes #1871 (and #872).

I agree that #872 doesn't justify a fix on its own, but testing for it since it is fixed now.

Added ```URLSchemeUnknown``` for a more meaningful error and changed ```ProxySchemeUnknown``` to inherit ```URLSchemeUnknown``` instead of ```ValueError```.

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -153,6 +153,16 @@ class LocationParseError(LocationValueError):
         self.location = location
 
 
+class URLSchemeUnknown(LocationValueError):
+    "Raised when a URL input has an unsupported scheme."
+
+    def __init__(self, scheme):
+        message = "Not supported URL scheme %s" % scheme
+        super(URLSchemeUnknown, self).__init__(message)
+
+        self.scheme = scheme
+
+
 class ResponseError(HTTPError):
     "Used as a container for an error reason supplied in a MaxRetryError."
     GENERIC_ERROR = "too many error responses"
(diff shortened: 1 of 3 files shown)


sethmlarson:
Can you verify that changing this exception works with Requests? Otherwise we'd have to subclass KeyError as well.

I agree that a server being able to cause a KeyError off a redirect is not something users would expect.

hodbn:
Tested on requests v2.23.0.
Requests looks up one of its adapters using the scheme first and raises an `InvalidSchema` if none was found.

PR #1873, opened 2020-05-11
Fixes errors thrown when the variable line is an empty string

boy-hack:
I ran into an error when using requests and ended up locating in response.py
```
  File "/venv/lib/python3.7/site-packages/urllib3/response.py", line 688, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''

  File "/venv/lib/python3.7/site-packages/urllib3/response.py", line 425, in _error_catcher
    yield
  File "/venv/lib/python3.7/site-packages/urllib3/response.py", line 755, in read_chunked
    self._update_chunk_length()
  File "/venv/lib/python3.7/site-packages/urllib3/response.py", line 692, in _update_chunk_length
    raise httplib.IncompleteRead(line)
http.client.IncompleteRead: IncompleteRead(0 bytes read)

```
Fixes errors thrown when the variable` line` is an empty string

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -692,6 +692,9 @@ class HTTPResponse(io.IOBase):
             return
         line = self._fp.fp.readline()
         line = line.split(b";", 1)[0]
+        if not line:
+            self.chunk_left = 0
+            return
         try:
             self.chunk_left = int(line, 16)
         except ValueError:


sigmavirus24:
This isn't correct as this isn't a bug. We want to raise an `IncompleteRead`

pquentin:
I opened https://github.com/nedbat/coveragepy/issues/988 for the nightly failure

pquentin:
And, yes, @sigmavirus24 is right, if we can't read the length of the chunk, it makes sense to raise IncompleteRead as the response is incorrect. There's even a comment that says why we're doing this.

Sorry, but the bug is probably server-side here.

boy-hack:
Although it is generated on the server side, the server side is not under my control. My fix ensures that the information is read even if there is an error on the server side.

sigmavirus24:
> My fix ensures that the information is read even if there is an error on the server side

I don't think I understand what you mean. If there's no chunk length (or _any kind of data_) that's been sent there's no possible way to read information sent by the server. There's physically _no data_ returned in that line so I can't understand how you could possibly believe that information is being read here.

PR #1888, opened 2020-06-09
Fix 1516

### LINKED ISSUE #1516: Getting exception in _update_chunk_length
Hi, 
While requesting a particular URL, I came across this error
```python
File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 601, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''

  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 360, in _error_catcher
    yield
  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 666, in read_chunked
    self._update_chunk_length()
  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 605, in _update_chunk_length
    raise httplib.IncompleteRead(line)
http.client.IncompleteRead: IncompleteRead(0 bytes read)

  aceback (most recent call last):
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/models.py", line 750, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 490, in stream
    for line in self.read_chunked(amt, decode_content=decode_content):
  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 694, in read_chunked
  File "/usr/lib/python3.5/contextlib.py", line 77, in __exit__
  File "/root/shivam/python3_env/lib/python3.5/site-packages/urllib3/response.py", line 378, in _error_catcher
    raise ProtocolError('Connection broken: %r' % e, e)
urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(0 bytes read)', IncompleteRead(0 bytes read))

  File "parse_dumped_data.py", line 87, in <module>
    Parse(f, entity)
  File "parse_dumped_data.py", line 17, in __init__
    self.parse_records()
  File "parse_dumped_data.py", line 32, in parse_records
    data_fields = self.get_data(record.get('data'))
  File "parse_dumped_data.py", line 50, in get_data
    data['image_url'] = self.get_image_url(data.get('image'), _id)
  File "parse_dumped_data.py", line 64, in get_image_url
    resp = requests.get(url)
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/api.py", line 75, in get
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/api.py", line 60, in request
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/sessions.py", line 533, in request
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/sessions.py", line 686, in send
    r.content
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/models.py", line 828, in content
    self._content = b''.join(self.iter_content(CONTENT_CHUNK_SIZE)) or b''
  File "/root/shivam/python3_env/lib/python3.5/site-packages/requests/models.py", line 753, in generate
    raise ChunkedEncodingError(e)
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(0 bytes read)', IncompleteRead(0 bytes read))
```

This was reported in requests module [here](https://github.com/requests/requests/issues/4248)

I fixed it as - 
```python
def _update_chunk_length(self):
    # First, we'll figure out length of a chunk and then
    # we'll try to read it from socket.
        return
    line = self._fp.fp.readline()
    line = line.split(b';', 1)[0]
    try:
        if len(line) == 0:
            self.chunk_left = 0
        else:
            self.chunk_left = int(line, 16)
    except ValueError:
        # Invalid chunked protocol response, abort.
        self.close()
        raise httplib.IncompleteRead(line)
```
or a one liner as
```python
    def _update_chunk_length(self):
        # First, we'll figure out length of a chunk and then
        # we'll try to read it from socket.
            return
        line = self._fp.fp.readline()
        line = line.split(b';', 1)[0]
        line = (len(line)>0 and line or "0")     # added this line
        try:
            self.chunk_left = int(line, 16)
        except ValueError:
            # Invalid chunked protocol response, abort.
            self.close()
            raise httplib.IncompleteRead(line)
```
Is it worth giving a PR regarding this??

hodbn:
Closes #1516.

* Add ```InvalidChunkLength``` error to indicate an error chunk with an invalid length
* Replace ```TestResponse::test_invalid_chunks``` with:
   - ```TestResponse::test_incomplete_chunk```
   - ```TestResponse::test_invalid_chunk_length```
   - ```TestResponse::test_buggy_incomplete_read```

That third test tests a platform specific flow. I think it should be tested in order to prevent confusion between ```IncompleteRead``` and ```httplib_IncompleteRead``` in the future.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -19,11 +19,12 @@ from .exceptions import (
     ReadTimeoutError,
     ResponseNotChunked,
     IncompleteRead,
+    InvalidChunkLength,
     InvalidHeader,
     HTTPError,
 )
 from .packages.six import string_types as basestring, PY3
-from .packages.six.moves import http_client as httplib
+from .packages.six.moves import http_client as httplib  # noqa: F401
 from .connection import HTTPException, BaseSSLError
 from .util.response import is_fp_closed, is_response_to_head
 
@@ -697,7 +698,7 @@ class HTTPResponse(io.IOBase):
         except ValueError:
             # Invalid chunked protocol response, abort.
             self.close()
-            raise httplib.IncompleteRead(line)
+            raise InvalidChunkLength(self, line)
 
     def _handle_chunk(self, amt):
         returned_chunk = None
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -16,6 +16,8 @@ from urllib3.exceptions import (
     ResponseNotChunked,
     ProtocolError,
     InvalidHeader,
+    httplib_IncompleteRead,
+    InvalidChunkLength,
 )
 from urllib3.packages.six.moves import http_client as httplib
 from urllib3.util.retry import Retry, RequestHistory
@@ -758,9 +760,9 @@ class TestResponse(object):
         with pytest.raises(ResponseNotChunked):
             next(r)
 
-    def test_invalid_chunks(self):
+    def test_incomplete_chunk(self):
         stream = [b"foooo", b"bbbbaaaaar"]
-        fp = MockChunkedInvalidEncoding(stream)
+        fp = MockChunkedIncompleteRead(stream)
         r = httplib.HTTPResponse(MockSock)
         r.fp = fp
         r.chunked = True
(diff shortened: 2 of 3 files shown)


test/test_response.py
```diff
@@ -758,19 +761,58 @@ def test_read_not_chunked_response_as_chunks(self):
         with pytest.raises(ResponseNotChunked):
             next(r)
 
-    def test_invalid_chunks(self):
+    def test_buggy_incomplete_read(self):
+        # Simulate buggy versions of Python
```
pquentin, line 765: This was fixed in Python 2.7.4 and Python 3 versions we no longer support, can we mention this here? Will be useful when we remove Python 2.7 support (or Python <2.7.9 support)

src/urllib3/response.py
```diff
@@ -19,11 +19,12 @@
     ReadTimeoutError,
     ResponseNotChunked,
     IncompleteRead,
+    InvalidChunkLength,
     InvalidHeader,
     HTTPError,
 )
 from .packages.six import string_types as basestring, PY3
-from .packages.six.moves import http_client as httplib
+from .packages.six.moves import http_client as httplib  # noqa: F401
```
pquentin, line 27: Why do we need the import? I think you removed the only use of `httplib` here

hodbn: There is a use of ```urllib3.response.httplib``` in two places in the tests:
https://github.com/urllib3/urllib3/blob/86fb90fdafee0077f4b90cbb2b9c3be99d4a1991/test/test_connectionpool.py#L13
https://github.com/urllib3/urllib3/blob/86fb90fdafee0077f4b90cbb2b9c3be99d4a1991/test/with_dummyserver/test_socketlevel.py#L12

I guess there is no reasonable excuse to export the import from ```test/response.py```, do you think moving it to ```test/__init__.py``` a more appropriate solution?

test/test_response.py
hodbn: Added a comment in the test

src/urllib3/response.py
pquentin: Maybe we should just use the six import in tests? In any case, we should stop importing httplib3 from urllib3.response, yeah.

pquentin: Would you like to do this here or in a another pull request?

PR #1894, opened 2020-06-25
Add HTTP/1.1 ALPN support

### LINKED ISSUE #1892: urllib3 should send http/1.1 ALPN extension by default
Some servers start to require ALPN extension and shut down HTTPS connections forecefully when a client does not indicate HTTP protocol version in ALPN extension. The fix is a trivial one-liner. Daniel told me that curl has been setting ALPN header for years without any negative side-effects.

See https://bugs.python.org/issue40968 for CPython bug
See https://github.com/python/cpython/pull/20959 for CPython PR

hodbn:
Added [ALPN support](https://tools.ietf.org/html/rfc7301), resolves #1892.

Sending ALPN protocols (```["http/1.1"]```) by default, if backend support is available.

TODO:
- [x] Better testing of ALPN support.
- [x] ```TestHTTPS_ALPN``` introduced too much tests.
- [x] Add SecureTransport support.

### Support:
Requires ```Python>=2.7.9``` (for ```SSLContext```) and backend support:
* ssl: needs ```OpenSSL>=1.0.2```
* PyOpenSSL: needs ```OpenSSL>=1.0.2, PyOpenSSL>=0.15, cryptography>=0.5```
* SecureTransport: needs ```macOS >= 10.12```

### DIFF
--- a/src/urllib3/contrib/_securetransport/bindings.py
+++ b/src/urllib3/contrib/_securetransport/bindings.py
@@ -276,6 +276,13 @@ try:
     Security.SSLSetProtocolVersionMax.argtypes = [SSLContextRef, SSLProtocol]
     Security.SSLSetProtocolVersionMax.restype = OSStatus
 
+    try:
+        Security.SSLSetALPNProtocols.argtypes = [SSLContextRef, CFArrayRef]
+        Security.SSLSetALPNProtocols.restype = OSStatus
+    except AttributeError:
+        # Supported only in 10.12+
+        pass
+
     Security.SecCopyErrorMessageString.argtypes = [OSStatus, c_void_p]
     Security.SecCopyErrorMessageString.restype = CFStringRef
 
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -56,6 +56,7 @@ import ctypes
 import errno
 import os.path
 import shutil
+import six
 import socket
 import ssl
 import threading
@@ -68,6 +69,7 @@ from ._securetransport.low_level import (
     _cert_array_from_pem,
     _temporary_keychain,
     _load_client_cert_chain,
+    _create_cfstring_array,
 )
 
 try:  # Platform-specific: Python 2
@@ -374,6 +376,19 @@ class WrappedSocket(object):
         )
         _assert_no_error(result)
 
+    def _set_alpn_protocols(self, protocols):
+        """
+        Sets up the ALPN protocols on the context.
+        """
+        if not protocols:
+            return
+        protos_arr = _create_cfstring_array(protocols)
+        try:
+            result = Security.SSLSetALPNProtocols(self.context, protos_arr)
+            _assert_no_error(result)
+        finally:
+            CoreFoundation.CFRelease(protos_arr)
+
     def _custom_validate(self, verify, trust_bundle):
         """
         Called when we have set custom validation. We do this in two cases:
@@ -441,6 +456,7 @@ class WrappedSocket(object):
         client_cert,
         client_key,
         client_key_passphrase,
+        alpn_protos,
     ):
         """
         Actually performs the TLS handshake. This is run automatically by
@@ -480,6 +496,9 @@ class WrappedSocket(object):
 
         # Setup the ciphers.
         self._set_ciphers()
+
+        # Setup the ALPN protocols.
+        self._set_alpn_protocols(alpn_protos)
 
         # Set the minimum and maximum TLS versions.
         result = Security.SSLSetProtocolVersionMin(self.context, min_version)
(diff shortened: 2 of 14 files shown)


sethmlarson:
I wonder if it makes sense for us to unconditionally add `http/1.1` to ALPN when available similar to how httplib will be doing so now and how curl has done so for forever. That way we can skip on all the conditional / `HAS_ALPN` detection logic and instead detect based on whether `SSLContext` has `set_alpn_protocols()` and catch `NotImplementedError`.

pquentin:
+1 to avoid adding yet another knob if we can avoid it

pquentin:
Good job so far, thanks!

hodbn:
I've removed API to customize ALPN protocols and most of ```HAS_ALPN``` references. Instead, I added a utility ```has_alpn``` function in order to allow precise testing.

What do you think?

src/urllib3/util/ssl_.py
```diff
@@ -397,6 +404,19 @@ def ssl_wrap_socket(
     return context.wrap_socket(sock)
 
 
+def has_alpn(ctx_cls=None):
```
sethmlarson, line 407: I'm not sure if this function is needed by users since ALPN both isn't configurable and is also "opportunistic" ie not the end of the world if it actually gets set or not. If this function is only used within the test suite let's remove it from `urllib3.util` and add it to test utilities.

src/urllib3/contrib/securetransport.py
```diff
@@ -374,6 +376,19 @@ def _set_ciphers(self):
         )
         _assert_no_error(result)
 
+    def _set_alpn_protocols(self, protocols):
```
sethmlarson, line 379: Would be great to confirm this works via capturing a TLS handshake on a supported platform, have we tried that yet?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -102,7 +102,7 @@ def socket_handler(listener):
             sock.close()
 
         self._start_server(socket_handler)
-        with HTTPConnectionPool(self.host, self.port) as pool:
+        with HTTPSConnectionPool(self.host, self.port) as pool:
```
sethmlarson, line 105: Huh, good catch here. Maybe we should try `self.buf.count(self.host.encode()) >= 2` in the verification too? MIght be interesting to see the results.

src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -276,6 +276,13 @@
     Security.SSLSetProtocolVersionMax.argtypes = [SSLContextRef, SSLProtocol]
     Security.SSLSetProtocolVersionMax.restype = OSStatus
 
+    try:
+        Security.SSLSetALPNProtocols.argtypes = [SSLContextRef, CFArrayRef]
```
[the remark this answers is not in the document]
hodbn: The link you posted shows the Swift view of the docs, but In the [ObjectiveC view](https://developer.apple.com/documentation/security/2867327-sslsetalpnprotocols?language=objc) you can see that both arguments are ```Ref```s.
According to [Apple's docs](https://developer.apple.com/documentation/swift/imported_c_and_objective-c_apis/working_with_core_foundation_types) Swift compiler removes the ```Ref``` suffix from CF types since it already manages references by itself. Cool 😃

src/urllib3/contrib/securetransport.py
```diff
@@ -441,6 +456,7 @@ def handshake(
         client_cert,
         client_key,
         client_key_passphrase,
+        alpn_protos,
```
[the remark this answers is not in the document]
hodbn: Done.

src/urllib3/util/ssl_.py
hodbn: Done, moved to ```test/__init__.py```.

src/urllib3/contrib/securetransport.py
hodbn: There is a test in ```TestALPN::test_alpn_protocol_in_first_request_packet```, is that what you meant?

test/with_dummyserver/test_socketlevel.py
hodbn: Shouldn't it be one in the HTTPS case too? It only worked before because the request was plaintext and the host header ws captured, but now it doesn't.

sethmlarson: Doh, right. Disregard me :)

src/urllib3/contrib/securetransport.py
sethmlarson: I meant it'd be great to have an indication this all works on a live macOS 10.12+ system by verifying manually w/ Wireshark. I don't have a macOS machine so I can't help here but we can do that outside this PR though :)

PR #1903, opened 2020-07-13
Fix unnecessary SNI warning with SecureTransport

### LINKED ISSUE #1902: Unnecessary SNI warning with SecureTransport
SNI is always enabled with the SecureTransport backend.
If a platform lacks SNI support and otherwise SNI would have been used (a valid ```server_hostname``` is given) - we issue an SNI warning.

With the SecureTransport backend, if one wraps a socket with no ```server_hostname```, an SNI warning is still being issued. This is inconsistent with the other backends behavior.

To reproduce, run on MacOS with and without SecureTransport injection:
```python
import socket
import ssl
import threading
# comment these 2 lines
from urllib3.contrib import securetransport
securetransport.inject_into_urllib3()
from urllib3 import util
import trustme

ca = trustme.CA()
cert = ca.issue_cert(u"localhost")
ca.cert_pem.write_to_path("/tmp/ca.pem")
cert.private_key_pem.write_to_path("/tmp/cert.key")
cert.cert_chain_pems[0].write_to_path("/tmp/cert.pem")
server_up = threading.Event()
c = socket.socket()
l = socket.socket()


def client():
    c.connect(('127.0.0.1', l.getsockname()[1]))
    ssl_sock = util.ssl_wrap_socket(c, ca_certs="/tmp/ca.pem")
    ssl_sock.close()


def server():
    l.bind(('127.0.0.1', 0))
    l.listen(1)
    server_up.set()
    s = l.accept()[0]
    ssl_sock = ssl.wrap_socket(s, server_side=True, certfile="/tmp/cert.pem", keyfile="/tmp/cert.key")
    ssl_sock.close()


client_thread = threading.Thread(target=client)
server_thread = threading.Thread(target=server)
server_thread.start()
assert server_up.wait(5)
client_thread.start()
server_thread.join()
client_thread.join()
```
An SNI warning is made only when using the SecureTransport backend, because of [these lines](https://github.com/urllib3/urllib3/blob/master/src/urllib3/util/ssl_.py#L380,L386).
We pass ```server_hostname``` to SecureTransport regardless of SNI hostname compliance (i.e. even if it is an IP address) since SecureTransport uses ```server_hostname``` to validate the certificate's CN.
A side effect of this commit, is issuing an SNI warning if ```server_hostname is None```.

I'm currently working on a better way to test SSL backends that will cover this scenario and a small refactor to the SNI logic in ```ssl_wrap_socket```.

hodbn:
Closes #1902.

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -375,14 +375,13 @@ def ssl_wrap_socket(
 
     # If we detect server_hostname is an IP address then the SNI
     # extension should not be used according to RFC3546 Section 3.1
-    # We shouldn't warn the user if SNI isn't available but we would
-    # not be using SNI anyways due to IP address for server_hostname.
-    if (
-        server_hostname is not None and not is_ipaddress(server_hostname)
-    ) or IS_SECURETRANSPORT:
-        if HAS_SNI and server_hostname is not None:
-            return context.wrap_socket(sock, server_hostname=server_hostname)
-
+    use_sni_hostname = server_hostname and not is_ipaddress(server_hostname)
+    # SecureTransport uses server_hostname in certificate verification.
+    send_sni = (use_sni_hostname and HAS_SNI) or (
+        IS_SECURETRANSPORT and server_hostname
+    )
+    # Do not warn the user if server_hostname is an invalid SNI hostname.
+    if not HAS_SNI and use_sni_hostname:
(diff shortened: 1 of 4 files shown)


sethmlarson:
Will give this a review today :)

pquentin:
@sethmlarson Is there anything making reviewing this difficult?

Issue #1904, opened 2020-07-16
macOS Big Sur dynamic linker cache causes contrib/securetransport to raise an ImportError


sheagcraig:
On macOS Big Sur, Apple has added a "dynamic linker cache". When ctypes tries to find the Security and CoreFoundation frameworks, an ImportError is raised, really because `ctypes.utils.find_library` can't see that the path exists. However, if you go ahead and load it anyway, everything is happy.

Please see: https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11-beta-release-notes
Specifically:
New in macOS Big Sur 11 beta, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

I have a PR for this ~incoming~ #1905 .

sethmlarson:
Closed in #1905

PR #1905, opened 2020-07-16
Add hardcoded paths to macOS frameworks for Big Sur compatibility.

sheagcraig:
ctypes.util.find_library is checking for the presence of the framework
files, which will fail due to macOS Big Sur's "dynamic linker cache".
Since these files really can't be anywhere else, we just hardcode the
path and load them. (see #1904 )

See: https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11-beta-release-notes
(relevent excerpt provided below)

This update takes the approach of leaving the existing behavior as is
for macOS 10.8 (lowest supported version) through 10.15.

Allegedly Apple is going to number Big Sur as macOS 11, although right
now the beta returns a version of "10.16", so this code handles both as
being greater than or equal to "10.16".

| New in macOS Big Sur 11 beta, the system ships with a built-in dynamic
linker cache of all system-provided libraries. As part of this change,
copies of dynamic libraries are no longer present on the filesystem.
Code that attempts to check for dynamic library presence by looking for
a file at a path or enumerating a directory will fail. Instead, check
for library presence by attempting to dlopen() the path, which will
correctly check for the library in the cache. (62986286)


src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -47,16 +47,6 @@
...
-if not security_path:
-    raise ImportError("The library Security could not be found")
-
-
-core_foundation_path = find_library("CoreFoundation")
-if not core_foundation_path:
-    raise ImportError("The library CoreFoundation could not be found")
-
-
 version = platform.mac_ver()[0]
```
pquentin, line 50: Can you please raise an `ImportError` if we're not on macOS at all? This is breaking the other tests, version `version` is the empty string, and `int("")` fails.

sheagcraig: Gotcha, yeah, no problem.

sheagcraig:
I do a lot with Saltstack's macOS working group, and I have a project for making requests Sessions that use the macOS keychain for mutual TLS and cert verification; there have been a lot of projects that use either PyObjC or ctypes directly to get at the macOS frameworks, and the failures are all pretty much due to checking for the existence of the files (which aren't there).

I have a Big Sur 10.16 beta2 machine on my desk that I tested this change out on.

I wanted to get this out there; obviously there's some time.

If anyone wants to direct me in any way, the version check and importing code could probably be cleaned up a bit. I mostly wanted to at least get a working POC out there.

sethmlarson:
Just chiming in to thank you for raising this issue to our attention and for providing a fix. We'll be sure to backport to 1.25-series and get a release out there before Big Sur is released.

sheagcraig:
👍🏻

I hate leaving it with a kind of speculative version check... May be 10.16, may be 11... wait until September!

pquentin:
According to [this answer by an Apple engineer](https://developer.apple.com/forums/thread/651512), it will be 11. But it's good if it works on the beta too, so we can leave 10.16 until the release.

sethmlarson:
@sheagcraig As long as the last released version we know of is 10.15 I'm fine with leaving the version check to be less restrictive. Sometimes with these things you have to cover all your bases.  Another way we could go about this is only doing the version lookup after a failed `find_library()` call, I'm unsure what the overhead for that is.

sheagcraig:
Cool @pquentin. And yes @sethmlarson, that sounds like a plan.

I'll see if I can cut out some of the repetition. I wasn't sure initially if the linter would be cranky with me for having the version check earlier, so I'll give it a little polish.

sheagcraig:
I have a contact at Apple that I'll ask about this; maybe we could use the hardcoded paths back further due to SIP making it very unlikely that those paths are anything but what we expect.

sheagcraig:
Well try this on for size. So I looked through `ctypes.util.find_library`, which is really running `macholib.dyld.dyld_find()`. It follows a search path, performing an `os.path.isfile()` looking for a path that is actually a file. This obviously won't work on Big Sur.

But the paths have been the same for all of the versions supported by the existing code (10.8+), and shouldn't be anywhere else.

So I just cleaned out a lot of the version conditionals and went straight to trying to load the libraries. If it fails to load for some reason, we catch the `OSError`, add a more friendly message, and reraise.

I also added a check to make sure we're even on macOS before going any further through the module.

I'm super green on urllib3 proper; I've poked around in the secure transport code a fair bit... But I guess I don't know why this was causing tests to fail on non-macOS, as the new code should have had the same behavior as the old? Regardless, this last couple of commits sidesteps that issue.

sheagcraig:
If this looks good, I'll yank that new test file too, since it basically does nothing now.

src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -47,26 +46,27 @@
 from ctypes import CDLL, POINTER, CFUNCTYPE
 
 
-security_path = find_library("Security")
```
sethmlarson, line 50: I'm concerned about changing this code path for macOS 10.15 or lower. How does something like this look?

```python
from urllib3.packages.six import raise_from
version_info = tuple(map(int, version.split(".")))

def load_cdll(name, macos10_16_path):
    """Loads a CDLL with a fallback path for macOS Big Sur
    10.16/11+ due to the dynamic linker cache.
    """
    try:
        # Big Sur is technically 11 but we use 10.16 due
        # to the Big Sur beta being labeled as 10.16.
        if version_info >= (10, 16):
            path = macos10_16_path
        else:
            path = find_library(name)
        if not path:
            raise OSError  # Caught and reraised as 'ImportError'
        return CDLL(path, use_errno=True)
    except OSError:
        raise_from(ImportError("The library %s failed to load" % name), None)

Security = load_cdll("Security", "<path for Security>")
CoreFoundation = load_cdll("CoreFoundation", "<path for CF>")
```

sheagcraig: That works too. I can rework to do it that way. It's certainly safer. I love trimming and cleaning code out, but even though the paths _should_ be correct all the way back, I think it can't hurt to err on the safe side.

sheagcraig: Do you want to do it the other way around? Try `find_library` and if that comes up with `None`, use the second argument to `load_cdll`?

Kind of like `dict.get`

sethmlarson: That works for me too!

sheagcraig:
Hah; well I just put your suggested changes in. It's fine like that!

test/contrib/test_securetransport_big_sur.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+import mock
```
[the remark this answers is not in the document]
sheagcraig: WIth the changes to retain the pre-Big Sur behavior, do it still make sense to take the test out?

src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -45,28 +44,44 @@
...
-core_foundation_path = find_library("CoreFoundation")
-if not core_foundation_path:
-    raise ImportError("The library CoreFoundation could not be found")
+from urllib3.packages.six import raise_from
 
 
 version = platform.mac_ver()[0]
 version_info = tuple(map(int, version.split(".")))
-if version_info < (10, 8):
+if platform.system() != "Darwin" or version_info < (10, 8):
```
pquentin, line 54: This checks need to be done before looking at `version_info`, because right now computing `version_info` raises an exception on Linux and Windows.

sheagcraig: Fixed.

Was this failing before? There was _no_ check for whether it was macOS before.

sethmlarson: Importerror was raised if find_library() didn't return.

pquentin: Right. Can you please raise ImportError instead of OSError? This is the expected exception indeed

sethmlarson:
I went ahead and made some changes myself on the branch:
- Fixed the imports so `ImportError` is raised on non-macOS as before.
- Deleted the test script, it's not necessary since CI doesn't support macOS 10.16 yet anyways so we can't test the alternate `load_cdll()` code path either way. Knowing it works on your system is enough for me to be confident it'll work for most others since the functionality is fairly simple. :)

pquentin:
@sethmlarson can't merge myself until you dismiss your review. Thanks for finishing this!

sethmlarson:
Thanks for the review @pquentin :)

pquentin:
Big Sur will be both 10.16 and 11.0, depending on context: https://eclecticlight.co/2020/07/21/big-sur-is-both-10-16-and-11-0-its-official/

PR #1907, opened 2020-07-16
Backport fixes to 1.25

sethmlarson:
Backported the following commits:
- (Add hardcoded paths for macOS Big Sur) https://github.com/urllib3/urllib3/commit/6bd33f64fb1c3107c56b23d479561322406ecfc9
- (Collapse Chunked Framing) https://github.com/urllib3/urllib3/commit/42ec81df905713045e69e2add4bb5787d89eab2f
- (SSLKEYLOGFILE) https://github.com/urllib3/urllib3/commit/fd1ad730bae89992e9591e84494713fc1cde3ad4
- (SSLKEYLOGFILE) https://github.com/urllib3/urllib3/commit/b8733538ef3a258ee76329c3900acb9456fae98e
- (Don't insert None into ConnectionPool) https://github.com/urllib3/urllib3/commit/46fc29d5108553a58cc14d28e73443e864be76c2
- (Change TARPIT_HOST) https://github.com/urllib3/urllib3/commit/898a16d09e4a6d9dbe10134a49b89eedfe8dae7f
- (Remove hasattr call in BrotliDecoder) https://github.com/urllib3/urllib3/commit/8fadde1f8a2add75b1de135a9374b5fa24a7c0d9

Commits that I didn't backport (because IMO they're features)
- (ALPN HTTP/1.1) https://github.com/urllib3/urllib3/commit/688584d692f3d77f5d6beca6bb97faef15f0977b
- (Default User-Agent) https://github.com/urllib3/urllib3/commit/a5a45dc36f3821e97bb9d6f2b4cd438a3f518af3
- (Unknown Scheme Error) https://github.com/urllib3/urllib3/commit/3ba4755960f54a21a6a1e492aa0d1df8b88aaba6
- (Invalid Chunk Length Error) https://github.com/urllib3/urllib3/commit/13953e8b2f5b85ff6a9c9b198cb0cde0c02ada28
- (Retry.other counter) https://github.com/urllib3/urllib3/commit/caad948e02b24c9516da453b7411df13af357449

I backported the `SSLKEYLOGFILE` environment variable commit because it doesn't change any functionality, it's just a change that's been long overdue and I'd like to get out there. Let me know if one of the commits I left out you think should also be backported.

Should be rebase merged once complete.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -300,5 +300,8 @@ In chronological order:
 * Chris Olufson <tycarac@gmail.com>
   * Fix for connection not being released on HTTP redirect and response not preloaded
 
+* [Bastiaan Bakker] <https://github.com/bastiaanb>
+  * Support for logging session keys via environment variable ``SSLKEYLOGFILE`` (Python 3.8+)
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/_travis/install.sh
+++ b/_travis/install.sh
@@ -10,7 +10,8 @@ set -exo pipefail
 if ! python3 -m pip --version; then
     curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
     sudo python3 get-pip.py
-    sudo python3 -m pip install nox
+    # https://github.com/theacodes/nox/issues/328
+    sudo python3 -m pip install nox==2019.11.9
 else
     # We're not in "dual Python" mode, so we can just install Nox normally.
     python3 -m pip install nox
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -72,7 +72,7 @@ Setting ``preload_content`` to ``False`` means that urllib3 will stream the
 response content. :meth:`~response.HTTPResponse.stream` lets you iterate over
 chunks of the response content.
 
-.. note:: When using ``preload_content=False``, you should call 
+.. note:: When using ``preload_content=False``, you should call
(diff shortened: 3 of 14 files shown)


pquentin:
I believe you'll need the nox fix from master to fix the Travis build. The lint is failing, too.

Honestly, the situation isn't satisfying. My understanding is that we only have 1.25.x because we don't want to merge the TLS proxy changes until TLS-in-TLS lands. Why don't we put the TLS proxy changes into a `next` branch, revert them from master, and release the features that are ready?

sethmlarson:
Agreed this is an unfortunate situation. I'd also have to move all the commits listed above too, don't really want to do that either. :/
I'll fix up the lint error and tests.

sethmlarson:
(Just an FYI, I didn't explicitly request a review from shazow, GitHub auto-assigns Code Owners on "required" reviews but it's optional for admins like shazow and I)

hodbn:
Regarding the Big Sur cache patch, [cpython already has a fix](https://github.com/python/cpython/pull/21564/commits/3940c867b6394d8950bfff725133e8a8241081b0).
Maybe we can somehow backport this to avoid the current changes?

Everything else looks great :smile:

sethmlarson:
The relevant change would be https://github.com/python/cpython/pull/21564/files#diff-13a79420cb44022a7d8cb3cfd490eba7R125 which I'm not sure I think would be better than what we already have given we're only loading two libraries not arbitrary libraries.

sethmlarson:
Thanks all for the reviews! :heart:

PR #1909, opened 2020-07-21
Fix handling http.client.RemoteDisconnected

keuko:
Handle RemoteDisconnected exception from http.client
and retry request.

RemoteDisconnected is raised when the attempt to read the
response results in no data read from the connection, indicating
that the remote end has closed the connection.

As server did not receive the request and closed connection,
so it should be safe to retry.

https://bugs.python.org/issue3566
https://hg.python.org/cpython/rev/eba80326ba53

### DIFF
--- a/src/urllib3/contrib/_securetransport/bindings.py
+++ b/src/urllib3/contrib/_securetransport/bindings.py
@@ -45,17 +45,11 @@ from ctypes import (
     c_bool,
 )
 from ctypes import CDLL, POINTER, CFUNCTYPE
+from urllib3.packages.six import raise_from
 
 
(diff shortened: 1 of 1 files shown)


src/urllib3/exceptions.py
```diff
@@ -14,6 +14,11 @@ class HTTPWarning(Warning):
     pass
 
 
+class DisconnectError(Exception):
```
sigmavirus24, line 17: I understand this isn't strictly an HTTP Error but I think we have everything sub-class that so that users can catch any urllib3 exception handily and this breaks that assumption severely.

PR #1911, opened 2020-07-23
Handle RemoteDisconnected exception from http.client and retry request.

keuko:
RemoteDisconnected is raised when the attempt to read the
response results in no data read from the connection, indicating
that the remote end has closed the connection.

As server did not receive the request and closed connection,
so it should be safe to retry.

https://bugs.python.org/issue3566
https://hg.python.org/cpython/rev/eba80326ba53


src/urllib3/exceptions.py
```diff
@@ -61,6 +61,16 @@ class ProtocolError(HTTPError):
     pass
 
 
+class RemoteDisconnectedError(ProtocolError):
+    "Raised by HTTPConnection.getresponse when the attempt to read the"
+    "response results in no data read from the connection, indicating that"
+    "the remote end has closed the connection."
```
sigmavirus24, line 67: Why not use a docstring?

keuko: Fixed, hope it is now OK.

keuko: > Why not use a docstring?

Done

gtherond:
Hey folks, is there anyone from urllib3 able to comment or validate on this issue? It seems to be pretty common and weirdly not implemented even if a proposal from 5 years ago is available.

src/urllib3/connectionpool.py
```diff
@@ -7,6 +7,11 @@
 from socket import error as SocketError, timeout as SocketTimeout
 import socket
 
+try:
+    from http.client import RemoteDisconnected
+except ImportError:
+    from httplib import BadStatusLine as RemoteDisconnected
```
hodbn, line 13: This line will catch other ```BadStatusLine``` exceptions that are not remote disconnection and might resend the request when inappropriate.

keuko: @hodbn in https://hg.python.org/cpython/rev/eba80326ba53 they noted this for exception exception:: RemoteDisconnected  ->  "  Previously, :exc:`BadStatusLine`\ ``('')``  
 was raised.  "

hodbn: Only until [that commit](https://github.com/python/cpython/commit/ee78ba2c819b0cd7671a575e584d8fe1b7adb3e4) 😞

keuko: Ok, If I understand correctly, I have to distinguish if BadStatusLine("No status line received - the server has closed the connection") or BadStatusLine(something) for py2. Do you have some recommendation how I should do it ?

hodbn: I can't think of a way that does not involve handling all three version ranges differently and peeking inside ```BadStatusLine``` for the newly introduced message 😞

sethmlarson: Quick question: since the old message contained the actual line could we see if `BadStatusLine.args[0] == ""`?

hodbn: Yeah, that should be ok with ```< 2.7.16```.

sethmlarson: It seems silly but could we do `BadStatusLine.args[0] in ("", "No status line received - the server has closed the connection")`? The worst that'd happen is we'd retry on a "legit" bad status error because some host decided to add that as its actual HTTP status.

keuko: > It seems silly but could we do `BadStatusLine.args[0] in ("", "No status line received - the server has closed the connection")`? The worst that'd happen is we'd retry on a "legit" bad status error because some host decided to add that as its actual HTTP status.

Hmm, but it looks like everytime when BadStatusLine is raised urllib should retry request. And my patch is doing this ...   https://github.com/urllib3/urllib3/blob/f0c43419e45db17a3c5f287be80aa0d14e2f58c6/test/with_dummyserver/test_socketlevel.py#L635

keuko:
Looks like commit [3671b807e83e28971cfe2ae75de7a9fb2742cee3](https://github.com/urllib3/urllib3/commit/3671b807e83e28971cfe2ae75de7a9fb2742cee3) is somehow breaking tests for ssl in travis :(

pquentin:
@keuko The currently failing tests don't appear to be SSL-related and are known to be flaky, this isn't related to https://github.com/urllib3/urllib3/commit/3671b807e83e28971cfe2ae75de7a9fb2742cee3. Sorry about the state of our test suite. Please ignore those failures for now

sethmlarson:
I pulled these changes locally but I think my Python 2.7 version is too recent to even test this. Will try to get to it :)

sethmlarson:
So I tried running the test case provided without the fix on Python 2.7.9, 2.7.15 and 2.7.16 and couldn't get the test to fail, are we sure this test case reproduces the issue or am I doing something wrong here?

keuko:
Hi,

Well, it is very intermittent issue , maybe i provided bad test , i will
check today afternoon when will be in work :)

Thank you for your review for now.

Michal Arbet ( kevko )

Dne ne 30. 8. 2020 17:12 uživatel Seth Michael Larson <
notifications@github.com> napsal:

keuko:
Hi Seth,

I found a problem , I will fix and update my pull request.

Very thanks for feedback :)

Michal Arbet ( kevko )

Dne po 31. 8. 2020 2:15 uživatel Michal Arbet <arbet.michal@gmail.com>
napsal:

keuko:
> @keuko I tried running the test case provided without the fix on Python 2.7.9, 2.7.15 and 2.7.16 and couldn't get the test to fail, are we sure this test case reproduces the issue or am I doing something wrong here?

Hi Seth,

I've already force-pushed edited unit test, check it with current code and with this patch.

keuko:
One more thing @sethmlarson, please read my original bug report in cpython https://bugs.python.org/issue41345 << .

I used apache with default keepalive timeout to 5 sec and sleep 5 sec in script.
On apache's server side I had simple script which just wrote received data to file to BE SURE that POST request never arrived to server when exception was thrown.

Sometimes it took long time to hit the bug, so be patient.
I've used iptables command below to decrease time needed to replicate the bug : 

iptables -A INPUT -p tcp --sport 80 -m statistic --mode random --probability 0.5 -j DROP

Michal Arbet ( kevko )

keuko:
Hmm, don't know why it is failing on CI , but locally I am passing :( 

(urllib3-python3) michalarbet@pixla:~/work/git/urllib3$ coverage run --parallel-mode -m pytest -r a --tb=native --no-success-flaky-report test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_server_closed_connection_right_before_request
=================================================================================================================================== test session starts ===================================================================================================================================
platform linux -- Python 3.8.2, pytest-4.6.9, py-1.9.0, pluggy-0.13.1
rootdir: /home/michalarbet/work/git/urllib3, inifile: setup.cfg
plugins: timeout-1.3.4, flaky-3.6.1
collected 1 item                                                                                                                                                                                                                                                                          

test/with_dummyserver/test_socketlevel.py .                                                                                                                                                                                                                                         [100%]

==================================================================================================================================== warnings summary =====================================================================================================================================
test/with_dummyserver/test_socketlevel.py::TestSocketClosing::test_server_closed_connection_right_before_request
  /home/michalarbet/work/git/urllib3/dummyserver/server.py:118: NoIPv6Warning: No IPv6 support. Falling back to IPv4.
    warnings.warn("No IPv6 support. Falling back to IPv4.", NoIPv6Warning)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================================================================== 1 passed, 1 warnings in 0.07 seconds ===========================================================================================================================

Dobatymo:
Do I understand this PR correctly that the `RemoteDisconnected` exception is proposed to be treated as a connection error which can always retried safely regardless of the HTTP method?

If so, that is a very bad idea. It might fix your usecase, however other servers can close the connection after receiving the request successfully and before sending a response, which triggers the `RemoteDisconnected` error as well.

keuko:
> Do I understand this PR correctly that the `RemoteDisconnected` exception is proposed to be treated as a connection error which can always retried safely regardless of the HTTP method?
>  Well, yes

> If so, that is a very bad idea. It might fix your usecase, however other servers can close the connection after receiving the request successfully and before sending a response, which triggers the `RemoteDisconnected` error as well.

No, it will not trigger RemoteDisconnected, or let's say this situation can't happen in real world.

Why ? Because there is no reason server will not send response after receive data. If exist some reason when client will not receive response... it is connection error - Timeout Error, not RemoteDisconnected.

RemoteDisconnected was implemented for this specific case , it is wery well documented in Cpython.
Please check discussion where RemoteDisconnected was introduced in Cpython - https://hg.python.org/cpython/rev/eba80326ba53.

And also code in cpython.

This bug is very annoying not only for me, google it and you will see this is long-term bug/problem

Dobatymo:
I am not sure what you are talking about

```python
import threading, sys

sys.path.insert(0, "src")
from dummyserver.server import SocketServerThread

from urllib3 import HTTPConnectionPool

THE_DATA = b"asdqwe"

def socket_handler(listener):
	sock = listener.accept()[0]

	buf = b""
	while not buf.endswith(THE_DATA):
		buf += sock.recv(65536)

	print(buf)

	sock.close()

ready_event = threading.Event()
server_thread = SocketServerThread(
	socket_handler=socket_handler, ready_event=ready_event, host="localhost"
)
server_thread.start()
ready_event.wait(5)
if not ready_event.is_set():
	raise Exception("most likely failed to start server")

with HTTPConnectionPool(server_thread.host, server_thread.port) as pool:
	r = pool.request("POST", "/", retries=0, body=THE_DATA)
```

will raise a `RemoteDisconnect` error and clearly print all the posted data.

keuko:
> I am not sure what you are talking about
> 
> ```python
> import threading, sys
> 
> sys.path.insert(0, "src")
> from dummyserver.server import SocketServerThread
> 
> from urllib3 import HTTPConnectionPool
> 
> THE_DATA = b"asdqwe"
> 
> def socket_handler(listener):
> 	sock = listener.accept()[0]
> 
> 	buf = b""
> 	while not buf.endswith(THE_DATA):
> 		buf += sock.recv(65536)
> 
> 	print(buf)
> 
> 	sock.close()
> 
> ready_event = threading.Event()
> server_thread = SocketServerThread(
> 	socket_handler=socket_handler, ready_event=ready_event, host="localhost"
> )
> server_thread.start()
> ready_event.wait(5)
> if not ready_event.is_set():
> 	raise Exception("most likely failed to start server")
> 
> with HTTPConnectionPool(server_thread.host, server_thread.port) as pool:
> 	r = pool.request("POST", "/", retries=0, body=THE_DATA)
> ```
> 
> will raise a `RemoteDisconnect` error and clearly print all the posted data. And it's happening in the real world. See my comment here: [jjjake/internetarchive#341 (comment)](https://github.com/jjjake/internetarchive/issues/341#issuecomment-689891649)
> 
> Of course the server SHOULD reply something, but there can be many reasons why it's not replying. In any case it's not safe to retry non-idempotent methods like this. There is a reason the retry logic in you first BPO link was not implemented.
> 
> Or am I missing something?


Ok, if I got your point .. it means that RemoteDisconnected can be catched ..but moreover it has to be tested if it is non-idempotent metod or NOT ..and raise Error which will not be retried ..but in other case has to be retried ..

As RFC say : 

A client, server, or proxy MAY close the transport connection at any time. For example, a client might have started to send a new request at the same time that the server has decided to close the "idle" connection. From the server's point of view, the connection is being closed while it was idle, but from the client's point of view, a request is in progress.

This means that clients, servers, and proxies MUST be able to recover from asynchronous close events. **Client software SHOULD reopen the transport connection and retransmit the aborted sequence of requests without user interaction so long as the request sequence is idempotent (see section 9.1.2). Non-idempotent methods or sequences MUST NOT be automatically retried, although user agents MAY offer a human operator the choice of retrying the request(s). Confirmation by user-agent software with semantic understanding of the application MAY substitute for user confirmation.** The automatic retry SHOULD NOT be repeated if the second sequence of requests fails.

keuko:
In current situation both types of request is failing , non-idempotent and  also idempotent.

Dobatymo:
Yes it should be fine to retry safe requests (OPTIONS, GET, HEAD). This is a low level library, so it should be using safe defaults, imo even idempotent but not safe requests (PUT, DELETE) should not be retried automatically.

For `safe` and `idempotent` see https://developer.mozilla.org/en-US/docs/Glossary/safe and https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

I think it's a good idea to reraise it as a proper exception (and not just as some argument in ProtocolError) so it can be handled easier from the user/application.

sethmlarson:
It sounds like there are some questions as to whether this is needed. Just an FYI I'd like to land this in 1.26 but I'm not going to mark it as a blocker for 1.26.0.

keuko:
> It sounds like there are some questions as to whether this is needed. Just an FYI I'd like to land this in 1.26 but I'm not going to mark it as a blocker for 1.26.0.

Now I am not sure if my patch is working perfectly as @Dobatymo noted above , please test yourself ..  @Dobatymo has a good point wit POST and SocketServerThread ... Please discuss it in urrlib ..maybe it can't be fixed :(

sethmlarson:
I think given the state of this PR and that we'd have to now rebase this on `main` instead of targetting 1.26 I'm going to close this. I'm still unsure if this is really needed, especially in the context of Python 3.x only for urllib3 v2.0. Would appreciate anyone looking into that further.

Thanks all for the work and weighing in on this PR!

PR #1912, opened 2020-07-23
Document parameters to HTTPSConnection, add advanced usage docs on SNI connections to an IP

### LINKED ISSUE #1884: Add additional documentation for HTTPSConnection parameters
I'm not sure if this is intentional or not, but the reference doc for connection objects ( https://urllib3.readthedocs.io/en/latest/reference/index.html#urllib3.connection.VerifiedHTTPSConnection ) seems a little bit spare on details about the parameters accepted.

In particular, I was looking at using `server_hostname`, and I didn't find it anywhere in the reference doc.
I did eventually find it (I assume with the same meaning) documented on [this utility function](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.ssl_wrap_socket).

Coming to these docs having very rarely needed to work directly with urllib3 (it's usually the transport used by some higher-level library I'm using, like `requests`), it seemed like the docs for the connection object were missing bits. If I were to open a PR adding a line to the docstring for each parameter, would it be welcome? Or are these params meant to be covered elsewhere in the docs?

sirosen:
This adds a small note in the docs that some of the params for HTTPSConnection pass through to ssl_wrap_socket and leaves it at that.

However, more directly useful for those trying to handle SNI by hand, as part of the advanced-usage narrative docs this adds a new section. It describes how to use SNI and connect to a host by IP. There might be other usages like this which are interesting, but choosing the address for a connection is a good, concrete example.

resolves #1884

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -172,6 +172,45 @@ verified with that bundle will succeed. It's recommended to use a separate
 :class:`~poolmanager.PoolManager` to make requests to URLs that do not need
 the custom certificate.
 
+.. _sni_custom:
+
+Custom SNI Hostname
+-------------------
+
+If you want to create a connection to a host over HTTPS which uses SNI, there
+are two places where the hostname is expected. It must be included in the Host
+header sent, so that the server will know which host is being requested. The
+hostname should also match the certificate served by the server, which is
+checked by urllib3.
+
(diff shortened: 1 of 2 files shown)


pquentin:
Thanks! Can you please run black on src/urllib3/connection.py? This is currently failing the build.

sirosen:
Whoops! `black`ed and amended my commit.

PR #1922, opened 2020-08-14
Allow Python nightly build to fail on Travis

sethmlarson:
It's looking like Python 3.10 will be an interesting release for tooling due to the 3 digit major version.


pquentin:
Apparently it's a CPython problem, see https://github.com/pypa/wheel/issues/354 and https://github.com/python/cpython/pull/20333. I subscribed to those issues to be able to restore the build as soon as possible

PR #1923, opened 2020-08-18
Integrate TLS-in-TLS support into urllib3

### LINKED ISSUE #1662: urllib3 does not allow to use HTTPS proxy
Hi, 
When I try to use HTTPS proxy, than urllib3 tries to create HTTPS tunnel by calling CONNECT. But my HTTPS proxy does not support it and resets connection after receiving it.

If I block running `self._prepare_proxy(conn)` function than all works fine, but I am not sure, that my fix does not broke somthing.
[changes](/SeyfSV/urllib3/commit/ec7c6390d92cdee36ece3b6f54ca0d0d88bf7ced)

jalopezsilva:
Integration PR after https://github.com/urllib3/urllib3/pull/1806 is closed.

The PR finally integrates the ``SSLTransport`` class to support TLS-in-TLS tunnels through proxies.  This will enable urllib3 to use HTTPS proxies with HTTPS destinations. The previous supported mode, forwarding the absolute URI method, is kept under a flag for HTTPS destinations through HTTPS proxies.  I added two new parameters for ProxyManagers ``proxy_ssl_context`` which can be used to pass an ``SSLContext`` to be used with the proxy and ``use_forwarding_for_https`` which enables the forwarding absolute URI mode. 

I decided to keep the forwarding mode as it has been incredibly useful within corporate environments when used with trusted proxies. (e.g we can audit no undesired information is sent through requests). This mode is by default disabled and I'm hoping that with enough documentation users won't use it unnecessarily. 

I also added three environment variables which can be used to configure the default proxy ssl context. ``PROXY_SSLCERT``, ``PROXY_SSLKEY``, ``PROXY_SSLPASSWD``. These environment variables will be super useful to configure the proxy ssl context when urllib3 is used within other libraries (e.g botocore, requests, etc)

Tests have been added. I have not added documentation yet, but I can do so easily. 

Closes #1662


test/__init__.py
```diff
@@ -167,7 +182,7 @@ def wrapper(*args, **kwargs):
 
 
 def notOpenSSL098(test):
-    """Skips this test for Python 3.5 macOS python.org distribution"""
+    """Skips this test for Python 3.5 , macOS python.org distribution"""
```
jalopezsilva, line 185: typo, I'll fix this.

jalopezsilva:
All right rebased after the closure of #1806, this should be good to review! I'll add subsequent commits when addressing feedback to make it easier.

src/urllib3/poolmanager.py
```diff
@@ -415,11 +443,13 @@ class ProxyManager(PoolManager):
         HTTPS/CONNECT case they are sent only once. Could be used for proxy
         authentication.
 
-    :param _allow_https_proxy_to_see_traffic:
+    :param proxy_ssl_context:
+        The proxy SSL context is used to establish the TLS connection to the
+        proxy when using HTTPS proxies.
+
+    :param _use_forwarding_for_https:
```
sethmlarson, line 450: This parameter should be public. I'd recommend documentation be something like:
```
(Defaults to False) If set to True will forward requests to the HTTPS proxy to be made on behalf of the client instead of
creating a TLS tunnel via the CONNECT method. **Enabling this flag means that
request and response headers and content will be visible from the HTTPS proxy** whereas tunneling
keeps request and response headers and content private.
IP address, target hostname, SNI, and port are always visible to an HTTPS proxy even when this flag is disabled.
```

src/urllib3/poolmanager.py
```diff
@@ -326,14 +328,38 @@ def _merge_pool_kwargs(self, override):
...
+
+        if self.proxy.scheme != "https":
+            return
+
+        if not PY3 and not self.proxy_config.use_forwarding_for_https:
+            raise ProxySchemeUnsupported(
+                "Contacting HTTPS destinations through HTTPS proxies is not supported on py2."
+            )
+
+        if (
```
sethmlarson, line 357: I wonder if checking the flags is too broad, we should probably check the actual `ssl_context` object that it's one from the stdlib/`urllib3.util.ssl_`? Technically users could be passing in an SSLContext manually even if pyOpenSSL/SecureTransport modes are injected by other programs.

src/urllib3/poolmanager.py
```diff
@@ -494,40 +528,29 @@ def _set_proxy_headers(self, url, headers=None):
...
-            # HTTPS to the proxy, we'll definitely need to set 'Host' at the
-            # very least.
+            # headers on the CONNECT to the proxy. If we're not using CONNECT,
+            # we'll definitely need to set 'Host' at the very least.
             headers = kw.get("headers", self.headers)
             kw["headers"] = self._set_proxy_headers(url, headers)
 
         return super(ProxyManager, self).urlopen(method, url, redirect=redirect, **kw)
 
+    def _set_destination_scheme(self, destination_scheme):
```
sethmlarson, line 545: Can't this change between requests? I don't think `connection_pool_kw` should be modified outside the constructor. Maybe need to find another way to get the scheme down to the ConnectionPool?

src/urllib3/util/proxy.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+
+def connection_requires_http_tunnel(
+    proxy_url=None, proxy_config=None, destination_scheme=None
+):
+    """
+    Returns True if the connection requires an HTTP CONNECT through the proxy.
+
+    :param destination_url:
```
sethmlarson, line 16: These parameters aren't the same as the signature

src/urllib3/util/proxy.py
```diff
@@ -0,0 +1,79 @@
...
+        and not ca_cert_dir
+        and not ca_cert_data
+        and hasattr(ssl_context, "load_default_certs")
+    ):
+        ssl_context.load_default_certs()
+
+    proxy_cert, proxy_key, proxy_pass = client_certificate_and_key_from_env()
+
+    if proxy_cert:
+        ssl_context.load_cert_chain(proxy_cert, keyfile=proxy_key, password=proxy_pass)
```
sethmlarson, line 65: If `proxy_pass` isn't given we shouldn't call with the `password` parameter. The parameter isn't supported on some Python 2.7 if I recall

src/urllib3/util/proxy.py
```diff
@@ -0,0 +1,79 @@
...
+
+    return ssl_context
+
+
+def client_certificate_and_key_from_env():
+    """
+    Attempts to retrieve a client certificate and key from the environment
+    variables to use with the proxy.
+    """
+    proxy_cert = os.environ.get("PROXY_SSLCERT")
```
sethmlarson, line 75: So I'm kinda in -1 territory when it comes to environment variables. Ideally our users would make these sorts of things configurable with their own interfaces. We don't support setting connection TLS config w/ environment variables either.

test/test_ssl.py
```diff
@@ -63,7 +63,7 @@ def test_context_sni_with_ip_address(monkeypatch, has_sni, server_hostname, uses
     if uses_sni:
         context.wrap_socket.assert_called_with(sock, server_hostname=server_hostname)
     else:
-        context.wrap_socket.assert_called_with(sock)
+        context.wrap_socket.assert_called_with(sock, server_hostname=None)
```
sethmlarson, line 66: We shouldn't be adding this parameter if `server_hostname` isn't given to not break on platforms without SNI

src/urllib3/poolmanager.py
jalopezsilva: Perfect, I like it!

src/urllib3/util/proxy.py
jalopezsilva: I agree with you in principle. Ideally all libraries should provide interfaces that allow clients to pass through a proxy ssl context and any necessary parameters to urllib3. I already have prepared upstream patches to do this for botocore and requests. I'll send those out once we release the changes in urllib3. 

That said though, there's a ton of libraries out there that would need clients to prepare PRs or changes to allow these parameters to go through. It seems easier to support those through environment variables for the time being.

Are you concerned about these being accidentally set? Could prefixing them by ```URLLIB3_*``` address that concern?

src/urllib3/poolmanager.py
jalopezsilva: Good idea, I'll move validation of pyopenssl and secure transport to util.ssl_.

sethmlarson:
Another thing I thought of while working through the docs overhaul:

I think that `SSLTransport` makes sense to move to `urllib3.util` since it's not a "third-party" module.
We'll also have to add documentation of these features but that can be its own PR :)

src/urllib3/poolmanager.py
jalopezsilva: I was able to improve how the destination scheme was used within ``poolmanager.py`` and ``connectionpool.py``. Let me know what you think!

jalopezsilva:
PR has been updated with two additional commits.

Commit #2 addresses the feedback received above. All of the items were addressed, let me know if I missed anything.

Commit #3 moved SSLTransport from contrib to util as requested by @sethmlarson. 

It seems there are some merge failures, I'll rebase once you've had a chance to review the two additional commits.

src/urllib3/poolmanager.py
```diff
@@ -326,14 +325,31 @@ def _merge_pool_kwargs(self, override):
...
+        """
+        if self.proxy is None or url_scheme != "https":
+            return
+
+        if self.proxy.scheme != "https":
+            return
+
+        if not PY3 and not self.proxy_config.use_forwarding_for_https:
+            raise ProxySchemeUnsupported(
+                "Contacting HTTPS destinations through HTTPS proxies is not supported on py2."
```
sethmlarson, line 351: Nit: let's say "Python 2" instead of "py2"

src/urllib3/util/proxy.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def connection_requires_http_tunnel(
+    proxy_url=None, proxy_config=None, destination_scheme=None
+):
+    """
+    Returns True if the connection requires an HTTP CONNECT through the proxy.
+
+    :param proxy_url:
+        :URL URL of the proxy.
```
sethmlarson, line 17: Can we use the `:param <type> <name>:` format?

src/urllib3/util/proxy.py
sethmlarson: Knowing when to use environment variables versus parameters in the program versus OS defaults is a big can of worms as we know from Requests, so I want to mark it out of scope for this PR and release. I can definitely see the value, I don't want to release it as a part of 1.26 though.

I appreciate that you have PRs for boto and requests though, that will be very useful! :)

src/urllib3/util/ssl_.py
```diff
@@ -426,3 +432,18 @@ def _is_key_file_encrypted(key_file):
...
 
     return False
+
+
+def _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname=None):
+    if tls_in_tls and SSLTransport:
+        validate_ssl_context_for_tls_in_tls(context)
+        return SSLTransport(sock, context, server_hostname)
+
+    wrapped_sock = None
```
sethmlarson, line 442: `wrapped_sock` isn't required, can `return from the `context.wrap_socket()`

src/urllib3/util/ssl_.py
```diff
@@ -426,3 +432,18 @@ def _is_key_file_encrypted(key_file):
                 return True
 
     return False
+
+
+def _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname=None):
```
sethmlarson, line 437: nit: let's mirror the parameter name with `ssl_context`

src/urllib3/util/ssl_.py
```diff
@@ -338,6 +342,8 @@ def ssl_wrap_socket(
     :param ca_cert_data:
         Optional string containing CA certificates in PEM format suitable for
         passing as the cadata parameter to SSLContext.load_verify_locations()
+    :param tls_in_tls:
+        Use SSLTransport of attempting to wrap the existing socket.
```
sethmlarson, line 346: "Use `SSLTransport` to wrap the existing socket"

src/urllib3/util/ssltransport.py
```diff
@@ -2,9 +2,26 @@
...
+def validate_ssl_context_for_tls_in_tls(ssl_context):
+    """
+    Raises a RuntimeError if the provided ssl_context can't be used for TLS in
+    TLS.
+
+    The only requirement is that the ssl_context provides the 'wrap_bio'
+    methods.
+    """
+
+    if not hasattr(ssl_context, "wrap_bio"):
```
sethmlarson, line 19: Maybe we should raise different error messages depending on the situation:
- If on Python 2.x: "TLS in TLS requires SSLContext.wrap_bio() which isn't supported on Python 2"
- If on Python 3.x: "TLS in TLS requires SSLContext.wrap_bio() which isn't available on non-native SSLContext"

Thoughts on this?

src/urllib3/util/ssltransport.py
```diff
@@ -2,9 +2,26 @@
...
 import io
 
+from urllib3.exceptions import ProxySchemeUnsupported
+
 SSL_BLOCKSIZE = 16384
 
 
+def validate_ssl_context_for_tls_in_tls(ssl_context):
+    """
+    Raises a RuntimeError if the provided ssl_context can't be used for TLS in
```
sethmlarson, line 12: Should be `ProxySchemeUnsupported` in the docs here, also should this be a private function? Maybe even add as a class method as `SSLTransport._validate_ssl_context()` so we don't have to do the optional import within `urllib3.util.ssl_`?

src/urllib3/connection.py
```diff
@@ -421,6 +434,43 @@ def connect(self):
...
+        return self.tls_in_tls_required
+
+    def _connect_tls_proxy(self, hostname, conn):
+        """
+        Establish a TLS connection to the proxy using the provided SSL context.
+        """
+        proxy_config = self.proxy_config
+        ssl_context = proxy_config.ssl_context
+        if not ssl_context:
+            ssl_context = generate_proxy_ssl_context(
```
sethmlarson, line 454: Should we support configuring `proxy_ca_certs`, `proxy_ssl_version`, etc that live on `ProxyConfig` instead of using `ca_certs` here? Allows for verifying proxy and destination server separately.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -60,21 +68,54 @@ def test_basic_proxy(self):
...
+            r = https.request("GET", "%s/" % self.http_url)
+            assert r.status == 200
+
+    @onlyPy2
+    def test_https_proxy_not_supported(self):
         with proxy_from_url(self.https_proxy_url, ca_certs=DEFAULT_CA) as https:
             r = https.request("GET", "%s/" % self.http_url)
             assert r.status == 200
 
             with pytest.raises(ProxySchemeUnsupported):
```
sethmlarson, line 86: Let's verify the message on the error once, or more if we decide to change the error message mentioned above

sethmlarson (changes requested):
Some more comments! :)

src/urllib3/util/ssltransport.py
jalopezsilva: It's a good idea, I like it and it will make the failure easier to understand. I'll make the corresponding changes.

src/urllib3/connection.py
jalopezsilva: What I was thinking about is if the user needs to provide custom client certificates or a custom CA to validate the proxy or a custom SSL versoin they should pass the ``proxy_ssl_context`` already prepared.

If not, we'll create one using the default settings we always use. 

That seemed easier and simple. If we introduce ``proxy_ca_certs`` in addition to ``proxy_ssl_context`` I wonder if people will get confused.

Let me know what you think!

jalopezsilva:
Great work on improving the CI! It's working much reliably now and it's also faster? (or is it just my impression?). In any case, I've updated the PR with two additional commits: One performs a merger from current ``urllib3/master`` and the second adrresses the last set of comments by @sethmlarson.

Let me know if we need more changes. I'll follow up on another PR for adding the missing tests on ``SSLTransport``.

sethmlarson:
@jalopezsilva That's all thanks to @pquentin and @hodbn! There's been so much work put into our CI and it's really paying off :clap:

src/urllib3/connectionpool.py
```diff
@@ -313,7 +319,7 @@ def _validate_conn(self, conn):
         """
         pass
 
-    def _prepare_proxy(self, conn):
+    def _prepare_proxy(self, conn, destination_scheme):
```
sethmlarson, line 322: Comment that must have gotten lost when I switched from mobile to PC:

Had a thought on being able to detect whether a `conn` is connected to HTTPS then we wouldn't need `destination_scheme` I believe? I saw many libraries defining `_prepare_proxy()` on GitHub, even though it's a "private" method we'd be breaking those users if we modified the signature like this so would be best if we could avoid doing so.

Initial thoughts: checking `hasattr(conn, "ssl_context")`, `isinstance(conn, HTTPSConnection)`? Since we don't support any of this unless we're using `HTTPSConnection` that might be enough?

sethmlarson (changes requested):
<removed because not needed>

src/urllib3/util/ssltransport.py
```diff
@@ -16,6 +19,28 @@ class SSLTransport:
     The class supports most of the socket API operations.
     """
 
+    @classmethod
+    def validate_ssl_context_for_tls_in_tls(cls, ssl_context):
```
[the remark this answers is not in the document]
jalopezsilva: sorry, totally forgot about making it private, all done!

src/urllib3/connectionpool.py
jalopezsilva: Good point. After reviewing about this for a while, we don't really need to modify the ``prepare_proxy`` method signature, just call it when we really need to do the HTTP CONNECT. I made some simplifications, let me know what you think!

jalopezsilva:
All right, another commit for the remaining comments.

@sethmlarson, happy to add another test case for ``AppEngineManager`` but I'm confused on how it would work. Isn't ``AppEngineManager`` a replacement for a PoolManager? Can you even specify proxies through it? Let me know if I'm missing something obvious, I went through https://github.com/urllib3/urllib3/blob/master/src/urllib3/contrib/appengine.py as I assumed that's what you were referring to.

pquentin:
@jalopezsilva On an unrelated note, should we credit your employer for sponsoring your time on this amazing feature? Or was it done on your time?

jalopezsilva:
@pquentin It's been a little bit of both. FB has been pretty supportive of me upstreaming the changes but I have also put personal time to make sure it goes through! No need to do any special announcements/credits, FB and I will just be happy to know the changes were merged in 😃

sethmlarson:
@sethmlarson Disregard the AppEngine comment, for some reason I thought it implemented the HTTPConnection interface. Doh.

src/urllib3/poolmanager.py
```diff
@@ -326,14 +325,32 @@ def _merge_pool_kwargs(self, override):
...
+        if self.proxy is None or url_scheme != "https":
+            return
+
+        if self.proxy.scheme != "https":
+            return
+
+        if not PY3 and not self.proxy_config.use_forwarding_for_https:
+            raise ProxySchemeUnsupported(
+                "Contacting HTTPS destinations through HTTPS proxies is not "
+                "supported on Python 2."
```
sethmlarson, line 352: Maybe should add `"through HTTPS proxies 'via CONNECT tunnels' is not supported in Python 2"`. Maybe we can mention `use_forwarding_for_https` option? Just worried users will blindly tick the box but hopefully most users trust their proxies... right? :shrug: 

also nit: remove the period

src/urllib3/connection.py
sethmlarson: I think not supporting all options for now is fine. We can always add more later, I suspect it'd be useful to specify client certificate/key instead of having to load it into an SSLContext yourself. Might be something to think about for documentation purposes! :)

I guess I'm then worried about below where all TLS options are used for both the proxy TLS handshake and the destination handshake. Before these values only were meant for the destination handshake.

Maybe some make sense but I think that client certificate doesn't make sense.

Here's my suggestion:
- If we're provided a `proxy_ssl_context` we don't use any values from the `HTTPSConnection`, assume everything is loaded as the user wants.
- If we're not given a `proxy_ssl_context` we can pass in `ca_certs`, `ca_cert_dir`, and `ca_cert_data` but not `keyfile`, `certfile`, and `key_password`. Can document how to achieve this if we're not adding `proxy_certfile`, etc quite yet.

src/urllib3/util/proxy.py
```diff
@@ -0,0 +1,60 @@
...
+        and proxy_config
+        and proxy_config.use_forwarding_for_https
+    ):
+        return False
+
+    # Otherwise always use a tunnel.
+    return True
+
+
+def generate_proxy_ssl_context(
```
sethmlarson, line 41: nit: let's use the same verb/prefix `create_` as with `create_urllib3_context`

src/urllib3/connection.py
```diff
@@ -428,6 +441,43 @@ def connect(self):
             or self.assert_fingerprint is not None
         )
 
+    def set_tls_in_tls_required(self):
```
sethmlarson, line 444: Can these two methods be a variable on the `HTTPSConnection` instance instead? `self.requires_tls_in_tls`?

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -63,21 +71,58 @@ def test_basic_proxy(self):
...
+
+    @onlySecureTransport
+    @onlyPy3
+    def test_https_proxy_securetransport_not_supported(self):
         with proxy_from_url(self.https_proxy_url, ca_certs=DEFAULT_CA) as https:
             r = https.request("GET", "%s/" % self.http_url)
             assert r.status == 200
 
             with pytest.raises(ProxySchemeUnsupported):
                 https.request("GET", "%s/" % self.https_url)
```
sethmlarson, line 114: can we add the `assert` from pyOpenSSL in this test as well

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -328,19 +374,34 @@ def test_https_headers(self):
...
             returned_headers = json.loads(r.data.decode())
             assert returned_headers.get("Foo") is None
             assert returned_headers.get("Baz") == "quux"
+            assert returned_headers.get("Hickory") is None
+            assert returned_headers.get("Host") == "%s:%s" % (
+                self.https_host,
+                self.https_port,
+            )
+
+    def test_https_headers_insecure(self):
```
sethmlarson, line 389: Can we rename this test to `https_headers_forwarding_for_https`?

sethmlarson (changes requested):
Phew, I think this is my final review! Thanks for the long haul on this, I know there are a couple extras to fill in after this gets merged. Feel free to use the issue tracker for tracking these too :)

mirkokg:
Weird, I created a simple code which requests https url with redirect.
It works few times and then suddenly it just returns the error.
Same URL and same proxy works perfectly in CURL.

```
import urllib3

proxy = urllib3.ProxyManager('https://user:password@proxy.com')
resp = proxy.request('GET', 'https://domain.com', redirect=False)
print(resp.headers.get('Location'))

```

```
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 700, in urlopen
    self._prepare_proxy(conn)
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 968, in _prepare_proxy
    conn.connect()
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connection.py", line 356, in connect
    self._tunnel()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/http/client.py", line 904, in _tunnel
    raise OSError("Tunnel connection failed: %d %s" % (code,
OSError: Tunnel connection failed: 407 Proxy Authentication Required

  File "t2.py", line 5, in <module>
    resp = proxy.request('GET', 'https://domain.com/', redirect=False)
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/request.py", line 75, in request
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/request.py", line 97, in request_encode_url
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/poolmanager.py", line 535, in urlopen
    return super(ProxyManager, self).urlopen(method, url, redirect=redirect, **kw)
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/poolmanager.py", line 378, in urlopen
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 787, in urlopen
    return self.urlopen(
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 787, in urlopen
    return self.urlopen(
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 787, in urlopen
    return self.urlopen(
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/connectionpool.py", line 759, in urlopen
    retries = retries.increment(
  File "/Users/mirko/Desktop/urllib3-tls_in_tls_integration/src/urllib3/util/retry.py", line 474, in increment
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='domain.com', port=443): Max retries exceeded with url: /Content/onevip-hls/Live/Channel(Sitel)/index.m3u8 (Caused by ProxyError('Cannot connect to proxy.', OSError('Tunnel connection failed: 407 Proxy Authentication Required')))
```

sethmlarson:
Thanks for reporting this. 

Is the proxy and target URL and proxy auth something that can be reproduced (or somehow can we reproduce the issue locally, we can discuss privately if needed)

If not could you capture your packet data with Wireshark and use SSLKEYLOGFILE? This would help us greatly!

mirkokg:
Email sent 👍

src/urllib3/poolmanager.py
jalopezsilva: Decided not to add the mention of ``use_forwarding_for_https``. If users need the feature they should research and understand it instead of blindly enabling it.

src/urllib3/connection.py
jalopezsilva: Yes, good idea!

jalopezsilva:
Added two more commits addressing comments in the previous review. We should be good to go I hope, let me know if there's anything more.

I'm going to get started in the documentation with a separate PR.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -63,21 +72,75 @@ def test_basic_proxy(self):
...
             r = https.request("GET", "%s/" % self.http_url)
             assert r.status == 200
 
-            with pytest.raises(ProxySchemeUnsupported):
+            with pytest.raises(ProxySchemeUnsupported) as excinfo:
                 https.request("GET", "%s/" % self.https_url)
 
+            assert "isn't available on non-native SSLContext" in str(excinfo.value)
+
+    def test_https_proxy_insecure(self):
```
sethmlarson, line 134: This testcase name might need a refresh?

sethmlarson (approved):
This looks good to me! :tada: I added one tiny comment but beyond that I'm happy with this. Would like final comments from other reviewers then we can ship this.

test/with_dummyserver/test_proxy_poolmanager.py
jalopezsilva: Oops, I'll update and grep around to make sure I'm not missing more.

pquentin (approved):
This looks great, thank you! It continues to complicate the code (I found the whole `*_required` logic difficult to follow) but it's definitely worth it! Thanks a lot.

sethmlarson (approved):
Woo!

PR #1924, opened 2020-08-19
Include mypy checks in noxfile and port initial pyi files

savarin:
The PR introduces mypy checks to the noxfile so that pre-specified files are type checked.

In addition, the .pyi stubs for `packages/ssl_match_hostname` are ported from [typeshed](https://github.com/python/typeshed/tree/master/third_party/2and3/requests/packages/urllib3). This is due to the fact that these files lie on the outermost part of the dependency tree. Additional .pyi stubs will be ported in due course.


src/urllib3/packages/ssl_match_hostname/__init__.py
```diff
@@ -10,10 +10,10 @@
 except ImportError:
     try:
         # Backport of the function from a pypi module
-        from backports.ssl_match_hostname import CertificateError, match_hostname
+        from backports.ssl_match_hostname import CertificateError, match_hostname  # type: ignore
```
savarin, line 13: Ignores type in conditional imports, as discussed in https://github.com/python/mypy/issues/1297.

noxfile.py
```diff
@@ -93,6 +102,27 @@ def lint(session):
     )
     session.run("flake8", "setup.py", "docs", "dummyserver", "src", "test")
 
+    # TODO: When all files are typed we can change this to .run("mypy", "--strict", "src/urllib3/")
+    session.log("mypy --strict src/urllib3/")
+    for typed_file in TYPED_FILES:
+        if not os.path.isfile(typed_file):
+            session.error(f"The file {typed_file!r} couldn't be found")
+        popen = subprocess.Popen(
+            f"mypy --strict --follow-imports=skip {typed_file}",
```
savarin, line 111: The default mypy behavior is to follow imports, this flag disables it as per https://mypy.readthedocs.io/en/latest/running_mypy.html#following-imports.

noxfile.py
```diff
@@ -93,6 +102,27 @@ def lint(session):
     )
     session.run("flake8", "setup.py", "docs", "dummyserver", "src", "test")
 
+    # TODO: When all files are typed we can change this to .run("mypy", "--strict", "src/urllib3/")
+    session.log("mypy --strict src/urllib3/")
+    for typed_file in TYPED_FILES:
+        if not os.path.isfile(typed_file):
+            session.error("The file {} couldn't be found".format(typed_file))
+        popen = subprocess.Popen(
+            "mypy --strict --follow-imports=skip {}".format(typed_file),
```
pquentin, line 111: I realized you just copied what Eland did but why not just use `session.run("mypy", "--strict", "--follow-imports=skip", typed_file)`? cc @sethmlarson 

It seems to work: https://github.com/urllib3/urllib3/commit/1926f5e5279cb680097d706df7e262a0228d587f

Doing that actually finds an issue: `src/urllib3/packages/ssl_match_hostname/_implementation.pyi:3: error: Function is missing a type annotation`

This fixes it: https://github.com/urllib3/urllib3/commit/33c4c13e4d3a5ce77b449b825d8f7214df2e2806

pquentin (commented):
Great start, thanks!

When this get in, I'd like a pull request to add this to our Github continuous integration checks.

Another possible side quest is to try to make `mypy src/urllib3` pass (that is, without strict mode). It'll be needed eventually and it reports way less errors than strict mode.

noxfile.py
sethmlarson: TIL about `--follow-imports=skip`, I'll modify Eland to use that option instead.

sethmlarson: Remembered the reason why not to use it! If you use `--follow-imports=skip` then when you're using something from outside the file then mypy complains that that thing isn't typed in some cases. I'm thinking about switching to something like this on Eland though: 

```python
    errors = []
    popen = subprocess.Popen(
        "mypy --strict eland/",
        env=session.env,
        shell=True,
        stdout=subprocess.PIPE,
        stderr=subprocess.STDOUT,
    )

    mypy_output = ""
    while popen.poll() is None:
        mypy_output += popen.stdout.read(8192).decode()
    mypy_output += popen.stdout.read().decode()

    for line in mypy_output.split("\n"):
        filepath = line.partition(":")[0]
        if filepath in TYPED_FILES:
            errors.append(line)
    if errors:
        session.error("\n" + "\n".join(sorted(set(errors))))
```

Both faster and simpler :)

savarin: @pquentin Thank you for https://github.com/urllib3/urllib3/commit/33c4c13e4d3a5ce77b449b825d8f7214df2e2806, I was pondering yesterday how do get the types here. May I ask how you got the types? I admit with my previous codebase I was doing `print(type(...))`.

@sethmlarson Great I've incorporated that. Just a quick flag that the `filepath` that are flagged are in fact the stub files, so I've made changes accordingly and renamed `TYPED_FILES` into `STUB_FILES`.

savarin:
> Great start, thanks!
> 
> When this get in, I'd like a pull request to add this to our Github continuous integration checks.
> 
> Another possible side quest is to try to make `mypy src/urllib3` pass (that is, without strict mode). It'll be needed eventually and it reports way less errors than strict mode.

@pquentin Appreciate the kind words! I'm afraid it's the first time I'm using Github CI, would be great if you could share what these changes look like. I'm also a little puzzled why the tests are failing, will experiment with a toy PR in a different branch.

noxfile.py
pquentin: > @pquentin Thank you for [33c4c13](https://github.com/urllib3/urllib3/commit/33c4c13e4d3a5ce77b449b825d8f7214df2e2806), I was pondering yesterday how do get the types here. May I ask how you got the types? I admit with my previous codebase I was doing `print(type(...))`.

`match_hostname` is a function that was added to the Python standard library with Python 3.5 and Python 2.7.9. Since urllib3 supported earlier versions of Python 3 at the time and still supports Python <2.7.9 to this day, we just copied its code to urllib3. Knowing that, I simply copied the type hints from https://github.com/python/typeshed/blob/master/stdlib/2and3/ssl.pyi. I included that link in the code, but I should probably have been more explicit about it.

Does that make sense?

However that fix isn't needed since it only applies when mypy is run in the context of nox, and using `subprocess` to run mypy fixes the issue.

@sethmlarson By the way, using `subprocess.run(text=True)` is probably even simpler as "wait for completion and provide all output to a string" will be already provided.

pquentin:
> I'm afraid it's the first time I'm using Github CI, would be great if you could share high level what these changes look like.

Actually I just realized that you're just modifying `nox -s lint`, and we're already calling that in our Travis CI configuration.

> I'm also a little puzzled why the tests are failing, will experiment with a toy PR in a different branch.

Not your fault, our tests are currently quite flaky...

PR #1927, opened 2020-08-21
Fix test_connection_cleanup_on_read_timeout flakiness

pquentin:
When calling `urlopen()`, even with `preload_content=False`, we read the
status line and headers, which can result in a read timeout. It happened
multiple times in CI since the read timeout is set to `SHORT_TIMEOUT`.

The fix is to simply allow those timeouts in the `urlopen()` call by
moving it inside the `pytest.raises()` context manager (and dropping
`preload_content=False`). This won't affect the test results as they're
about the connection pool: as long as we are able to connect, we're
good.

However, unlike `read()`, `urlopen()` can be retried on read timeouts,
which leads to the timeout exception being wrapped in `MaxRetryError` so
I had to take that into account.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -708,10 +708,12 @@ class TestSocketClosing(SocketDummyServerTestCase):
             poolsize = pool.pool.qsize()
             timeout = Timeout(connect=LONG_TIMEOUT, read=SHORT_TIMEOUT)
             try:
-                with pytest.raises(ReadTimeoutError):
-                    pool.urlopen(
-                        "GET", "/", retries=Retry(total=False), timeout=timeout,
-                    )
+                with pytest.raises((ReadTimeoutError, MaxRetryError)) as e:
+                    pool.urlopen("GET", "/", retries=0, timeout=timeout)
+                # Reading the status line can fail with a ReadTimeoutError inside a
+                # MaxRetryError
+                if isinstance(e, MaxRetryError):
+                    assert isinstance(e.value.reason.original_error, ReadTimeoutError)
                 assert poolsize == pool.pool.qsize()
             finally:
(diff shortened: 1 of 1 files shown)


pquentin:
Please take another look

test/with_dummyserver/test_socketlevel.py
```diff
@@ -707,12 +707,13 @@ def socket_handler(listener):
         with HTTPConnectionPool(self.host, self.port) as pool:
             poolsize = pool.pool.qsize()
             timeout = Timeout(connect=LONG_TIMEOUT, read=SHORT_TIMEOUT)
-            response = pool.urlopen(
-                "GET", "/", retries=0, preload_content=False, timeout=timeout
-            )
             try:
-                with pytest.raises(ReadTimeoutError):
-                    response.read()
+                with pytest.raises((ReadTimeoutError, MaxRetryError)) as e:
```
sethmlarson, line 711: It's strange to me that we can raise a `MaxRetryError` when `retries=0`.

sethmlarson (dismissed):
One comment but looks fine, merge when ready!

test/with_dummyserver/test_socketlevel.py
pquentin: `retries=False` is equivalent to `retries=Retry(total=False)`. My first commit was doing that, but I was getting connection timeouts, however I want to retry those at least once, which happens with `retries=0`.

sethmlarson: Shouldn't this only raise MaxRetryError then? Maybe I'm misunderstanding how our retry exhaustion behaves

pquentin: You're right, thanks! Fixed.

pquentin:
This still needs work. Closing for now.

Issue #1897, opened 2020-06-26
Start shipping our own type stubs


sethmlarson:
Typeshed houses very rudimentary type stubs for urllib3 but they're hidden within `requests.packages.urllib3` so Mypy can't take advantage of them when using urllib3 on its own. We could also probably improve on what is already in typeshed by adding better types than `Any` everywhere.

judej:
Seth, are you looking to add type info in the code or add .pyi files with the py files or contribute a more complete set at typeshed? I'd like to help with this

savarin:
Hi @sethmlarson! I'm at Recurse Center and got introduced to his by @shazow. Could I clarify, as per the title, what you have in mind is to have the `.pyi` files alongside `.py` files? Would the first step simply involve porting the one in [requests](https://github.com/python/typeshed/tree/master/third_party/2and3/requests/packages/urllib3), and then expanding on the `Any`?

I've actually been a little stuck getting the stubs hidden in [requests](https://github.com/python/typeshed/tree/master/third_party/2and3/requests/packages/urllib3) to work with `mypy`, even after importing `requests`.

For additional context, my previous role involved making py2 code compatible with py3 and adding type comments. The thought was there's less friction not having to review separate `.pyi` files when refactoring.

sethmlarson:
@savarin Glad you're interested in helping us here.  I was thinking starting with the type stubs that typeshed has (pyi) and making the `Any` more specific.

judej:
@sethmlarson, would you prefer that we modify the stubs in typeshed or would you rather have the .pyi files in this repo checked in alongside the .py files? OR, would you rather add the type info into the .py files?

sethmlarson:
I'd like the pyi files shipped with urllib3.

savarin:
Cool, sounds like a good starting point. Two questions, stemming from my lack of familiarity with stubs and/or `typeshed`.

1. How do you run a type check with the existing `typeshed` [stubs](https://github.com/python/typeshed/tree/master/third_party/2and3/requests/packages/urllib3) of `urllib3` since it's in `requests`? Detailed run below, importing `requests` doesn't seem to help.

```
$ cat run.py

import requests
import urllib3

def pool():
    return urllib3.PoolManager()

$ mypy run.py
run.py:3: error: Skipping analyzing 'urllib3': found module but no type hints or library stubs
run.py:3: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
[...]

PR #1928, opened 2020-08-21
Require mypy installation and add to contributors

savarin:
Addresses https://github.com/urllib3/urllib3/issues/1897, more specifically https://github.com/urllib3/urllib3/pull/1924#issuecomment-678282065.

Includes `mypy` as a requirement for the dev environment. Add to list of contributors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ jobs:
     - python: nightly
   include:
     # Lint & documentation.
-    - python: 3.6
+    - python: 3.8
       env: NOX_SESSION=lint
-    - python: 3.6
+    - python: 3.8
       env: NOX_SESSION=docs
 
     # Unit tests
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -303,5 +303,8 @@ In chronological order:
 * [Bastiaan Bakker] <https://github.com/bastiaanb>
   * Support for logging session keys via environment variable ``SSLKEYLOGFILE`` (Python 3.8+)
 
+* [Ezzeri Esa] <https://github.com/savarin>
+  * Ports and extends on types from typeshed
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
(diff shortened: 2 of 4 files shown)


pquentin (approved):
Thanks!

This time everything went fine, even codecov: https://codecov.io/gh/urllib3/urllib3/tree/f66cf3d20961220a72a8634e732a9285b3c7b720

I don't understand why it's red on GitHub, but this can be safely merged.

savarin:
Closing and reopening to rerun test suite.

sethmlarson:
Shouldn't we be adding `mypy` to be installed by `nox` within the `lint` task?

pquentin:
Oh, you mean `session.install("mypy")` can work even if we're using Popen() to run mypy outside of the nox venv?

I just realized that `nox -s lint` is broken anyway: we don't look at returncode, and thus don't detect failures like "mypy isn't installed". And we should be using `["mypy", "--strict", "src/urllib3"]`, not "mypy --strict src/urllib3". Maybe there are other problems.

pquentin:
I've added a few improvements.

 * Using a single string wasn't actually broken as we were using `shell=True`, but avoiding that is always good idea if possbile
 * It's useful to read the return code to make sure that mypy itself ran succesfully, but it's slightly tricky because we need to accept return code 1 which is "I found errors!"
 * I've switched to subprocess.run() which eliminates the need to do stdout decoding
 * I fixed a path error
 * I logged that mypy is being run, otherwise the errors appear to come fom flake8

sethmlarson:
@pquentin It looks like the `text` parameter was only added to `subprocess.run()` in Python 3.7, maybe we need to bump our Python version running the Travis `lint` job? I'm going to bump that and make sure it passes.

savarin:
> I've added a few improvements.
> 
> * Using a single string wasn't actually broken as we were using `shell=True`, but avoiding that is always good idea if possbile
> * It's useful to read the return code to make sure that mypy itself ran succesfully, but it's slightly tricky because we need to accept return code 1 which is "I found errors!"
> * I've switched to subprocess.run() which eliminates the need to do stdout decoding
> * I fixed a path error
> * I logged that mypy is being run, otherwise the errors appear to come fom flake8

@pquentin Nice! Much appreciate the additions and comments!

PR #1931, opened 2020-08-23
Include additional files to be typed checks, up to exceptions.py

savarin:
Addresses https://github.com/urllib3/urllib3/issues/1897, as per discussion in https://github.com/urllib3/urllib3/issues/1897#issuecomment-678806248.

This PR includes additional files to be typed checked, up to `exceptions.py`. Note that both the `.py` and `.pyi` files are included in the checks.

In particular, the `.pyi` file from `typeshed` for `exceptions.py` are ported over and expanded to satisfy type checks.


src/urllib3/exceptions.pyi
```diff
@@ -0,0 +1,50 @@
+from typing import Any, Optional, Union, Tuple
+
+class HTTPError(Exception): ...
+class HTTPWarning(Warning): ...
+
+class PoolError(HTTPError):
+    pool: Any
+    def __init__(self, pool: Any, message: str) -> None: ...
```
savarin, line 8: The `pool` here is from `connectionpool.py`, but importing here would create a circular dependency.

src/urllib3/exceptions.pyi
```diff
@@ -0,0 +1,50 @@
...
+
+ConnectionError: Any
+
+class MaxRetryError(RequestError):
+    reason: Any
+    def __init__(self, pool: Any, url: str, reason=Optional[str]) -> None: ...  # type: ignore
+
+class HostChangedError(RequestError):
+    retries: Any
+    def __init__(self, pool: Any, url: str, retries=int) -> None: ...  # type: ignore
```
savarin, line 29: I'm a little puzzled why I keep getting a mypy error here despite adding more types, so included an ignore.

```
src/urllib3/exceptions.pyi:25: error: Function is missing a type annotation for one or more arguments
src/urllib3/exceptions.pyi:29: error: Function is missing a type annotation for one or more arguments.
```

sethmlarson: Should be `retries: int` instead of =int :)

noxfile.py
```diff
@@ -8,8 +8,16 @@
 # Whenever type-hints are completed on a file it should be added here so that
 # this file will continue to be checked by mypy. Errors from other files are
 # ignored.
-STUB_FILES = {
+TYPED_FILES = {
```
sethmlarson, line 11: It's getting a little verbose to store both `.py` and `.pyi`, thoughts on adding a `.replace(".pyi", ".py")` to the end of the `filepath` logic to normalize everything to `.py`?

src/urllib3/exceptions.pyi
sethmlarson: If you're hitting circular imports you can sometimes use this construction:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from urllib3.connectionpool import ConnectionPool

class PoolError:
    pool: "ConnectionPool"
```

src/urllib3/exceptions.pyi
```diff
@@ -0,0 +1,50 @@
...
+class SSLError(HTTPError): ...
+class ProxyError(HTTPError): ...
+class DecodeError(HTTPError): ...
+class ProtocolError(HTTPError): ...
+
+ConnectionError: Any
+
+class MaxRetryError(RequestError):
+    reason: Any
+    def __init__(self, pool: Any, url: str, reason=Optional[str]) -> None: ...  # type: ignore
```
sethmlarson, line 25: Should be `:Optional[...]` instead of `=Optional[...]`, remove the `# type: ignore ` after verifying this works

src/urllib3/exceptions.pyi
```diff
@@ -0,0 +1,50 @@
...
+class TimeoutStateError(HTTPError): ...
+class TimeoutError(HTTPError): ...
+class ReadTimeoutError(TimeoutError, RequestError): ...
+class ConnectTimeoutError(TimeoutError): ...
+class EmptyPoolError(PoolError): ...
+class ClosedPoolError(PoolError): ...
+class LocationValueError(ValueError, HTTPError): ...
+
+class LocationParseError(LocationValueError):
+    location: Any
```
sethmlarson, line 40: I think we know `location` is a `str` from the constructor? Can change from `Any` to `str`

src/urllib3/exceptions.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-from .packages.six.moves.http_client import IncompleteRead as httplib_IncompleteRead
```
sethmlarson, line 2: I'm guessing this `# type: ignore` is temporary? Should we start by typing `urllib3.packages.six` per the dependency tree?

noxfile.py
savarin: Nice, very clean.

src/urllib3/exceptions.pyi
savarin: LOL I guess this is where I should have taken a break :D

savarin:
> Thanks for this, here are some comments for you :)
> 
> I'm wondering about the empty `.pyi` files, should we follow the dependency tree to commit type stubs for all the dependencies of a node before that node itself to not run into issues?

Thank you for the comments! I think that resolves all the generic `Any` and type ignores.

I admit I took the easy way out with `six.pyi`, but what the type stubs would look like there complete baffles me. Let me take another quick look, presume it'll just be for `http_client` for the time being?

savarin:
@sethmlarson OK I took a stab at `packages.six.pyi`, it doesn't appear we need one for `packages.__init__.pyi`.

savarin:
@sethmlarson My initial plan was to ping to see if you had any additional thoughts, but instead I added an additional downstream file to `TYPED_FILES` to test out `packages.six.pyi`. Since there errors were raised, this shows that the stubs in `packages.six.pyi` is not having the desired effect. I'll take another look tomorrow.

```
"src/urllib3/request.py:4: error: Cannot find implementation or library stub for module named 'urllib3.packages.six.moves.urllib.parse'"
```

savarin:
Merged as https://github.com/urllib3/urllib3/pull/1937 instead.

PR #1932, opened 2020-08-23
Use tz-aware functions in Retry.parse_retry_after()

### LINKED ISSUE #1896: parse_retry_after() in utils/retry.py computes seconds incorrectly when Retry-After value is a http-date and local time is not in the GMT timezone
For the given Retry-After value `Thu, 25 Jun 2020 21:30:12 GMT`, with the local time being `Thu, 25 Jun 2020 17:28:12 EST`, parse_retry_after() will return `14564.671981096268`.  This value is incorrect, and should be `~120`.

parse_retry_after() doesn't correctly handle the time difference between time zones.

sethmlarson:
Switches `parsedate` and `mktime` to the tz-aware equivalents in `email.utils`. Also adds Freezegun as a dev dependency for complete time-mocking.

Closes #1896

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,6 +7,7 @@ PySocks==1.7.1
 win-inet-pton==1.1.0
 pytest==4.6.9
 pytest-timeout==1.3.4
+pytest-freezegun==0.3.0.post1
 flaky==3.6.1
 trustme==0.5.3
 cryptography==2.8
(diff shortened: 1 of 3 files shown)


pquentin (approved):
Thanks! I like the symmetry between parsedate_tz and mktime_tz. :) And freezegun is a great find.

sethmlarson:
Pinned to `0.3.0.post1` because apparently there's no Python 2-compatible dist for that version. Now CI should work, going to merge after the install on Python 2 works.

pquentin:
(For what it's worth: my intent was to click on approve, not comment)

Issue #1930, opened 2020-08-22
Fix handling of bytes for User-Agent header


sethmlarson:
See:
- https://github.com/urllib3/urllib3/pull/1750#issuecomment-670785178
- https://github.com/boto/botocore/pull/2123

This is a blocker for 1.26.0

PR #1933, opened 2020-08-25
Fixed bytes header handling.

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/1930

It is my first contribution to this repo :heart_eyes: 
Did I miss something?

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -18,7 +18,7 @@ except ImportError:  # Platform-specific: No threads available
 
 from collections import OrderedDict
 from .exceptions import InvalidHeader
-from .packages.six import iterkeys, itervalues, PY3
+from .packages.six import ensure_str, iterkeys, itervalues, PY3
 
 
 __all__ = ["RecentlyUsedContainer", "HTTPHeaderDict"]
@@ -154,7 +154,7 @@ class HTTPHeaderDict(MutableMapping):
 
     def __getitem__(self, key):
         val = self._container[key.lower()]
-        return ", ".join(val[1:])
+        return ", ".join([ensure_str(v, "ascii") for v in val[1:]])
 
     def __delitem__(self, key):
         del self._container[key.lower()]
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -853,6 +853,14 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             assert no_ua_headers["User-Agent"] == SUPPRESS_USER_AGENT
             assert pool_headers.get("User-Agent") == custom_ua
 
+    def test_bytes_header(self):
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            headers = {"User-Agent": b"test header"}
+            r = pool.request("GET", "/headers", headers=headers)
+            request_headers = json.loads(r.data.decode("utf8"))
+            assert "User-Agent" in request_headers
+            assert request_headers["User-Agent"] == "test header"
+
 
 class TestRetry(HTTPDummyServerTestCase):
     def test_max_retry(self):


src/urllib3/_collections.py
```diff
@@ -220,6 +220,8 @@ def add(self, key, val):
         >>> headers['foo']
         'bar, baz'
         """
+        if PY3 and isinstance(val, bytes):
+            val = val.decode("utf-8")
```
sigmavirus24, line 224: I don't know that we unconditionally want to force things to utf8 as that may not always be safe to do for data validity/consistency

hramezani: Thanks for review. I changed it to `val = val.decode()`.

pquentin: Note that this strictly equivalent since in Python 3 the default encoding is 'utf-8': https://docs.python.org/3/library/stdtypes.html#bytes.decode. (By the way, we don't want to use `if PY3`, see https://github.com/urllib3/urllib3/pull/1669)

Honestly, I don't know what the real fix should be, but hey @sethmlarson is often optimistic when adding "Contributor Friendly" labels. :D

pquentin: It would helpful to first have a way to reproduce the problem, as it would help us to understand it better. Using boto3 would be fine, but it shouldn't require AWS credentials. @hramezani Do you think you can try to work on a such a reproducer script?

sethmlarson: What are our thoughts on adding a `ensure_str(x, "ascii")` to each header value within `HTTPHeaderDict.__getitem__()`?
The `__getitem__()` call would previously explode if users were adding bytes anyways so we know we're not "breaking" anyone and it seems like handing a `bytes` instance to httplib hasn't caused problems in the past.
Going that route preserves the ability to have `bytes` header types in `HTTPHeaderDict` which we've been unknowingly supporting all this time.

pquentin: `ensure_str` encodes strings to bytes? It's the only safe operation, yes, and `ascii` is the encoding used by httplib anyway, so that would make a lot of sense indeed. 👍 on that

hramezani: Thanks @sethmlarson for your suggestion. I changed my PR based on that. Don't know I got it right or not.

> Do you think you can try to work on a such a reproducer script?

@pquentin
Here is a simple script from botocore GitHub page:
```python
>>> import botocore.session
>>> session = botocore.session.get_session()
>>> client = session.create_client('ec2')
>>> print(client.describe_instances())
```
I reproduced the bug with the above script in my machine. Here are the steps:

- Create a virtualenv
- Install botocore `pip install botocore`
- Remove urllib3 `pip uninstall urllib3`
- Install urllib3 from my local repository `pip install -e <urllib3 path on my machine>`
- Run the script - Here I got the error `botocore.exceptions.HTTPClientError: An HTTP Client raised an unhandled exception: sequence item 0: expected str instance, bytes found`
- Apply the patch
- Run the script - Was successful

sigmavirus24:
I mean I guess this looks fine. I've completely forgotten at this point what's *correct* and so if we force everything to be ascii text (which I don't think is necessarily correct but :man_shrugging: ) that probably works for 99% of users. Part of me thinks the best way to handle all headers is to accept bytes and use bytes but I don't think we've done that and I think `httplib`/`http.client` gets real ornery when we pass bytes there but not elsewhere so, this is probably right.

pquentin:
> I think `httplib`/`http.client` gets real ornery when we pass bytes there but not elsewhere

@sigmavirus24 What are you referring to with "there" and "elsewhere"?

The best way to handle headers is definitely to use bytes and receive bytes, and this is what h11 does: https://h11.readthedocs.io/en/latest/api.html#headers-format. But we can't change our API, so our best bet is to convert to bytes internally.

http.client already encode strings to ASCII bytes in Python 3: https://github.com/python/cpython/blob/8ca8a2e8fb068863c1138f07e3098478ef8be12e/Lib/http/client.py#L1226-L1252, so doing that here should be safe.

I'm just curious: why are we doing that at in `__getitem__` and not at insertion time?

pquentin (approved):

sethmlarson:
My thinking is it's the smallest change that allows us to inspect HTTP headers before handing off to httplib (because apparently we weren't doing this before the User-Agent change) while not changing any code paths that were taken within httplib. If botocore has been doing this it's almost guaranteed other users are too.

If we were to force everything to bytes on ingest (which I agree is what we "should" be doing, maybe can be a breaking change in v2?) then `HTTPHeaderDict.get_all()` would not hand back the same types that were inserted. This is what the output looks like with this change:

```python
from urllib3._collections import HTTPHeaderDict

headers = HTTPHeaderDict()

headers.add("user-agent", b"value")
headers.add("user-agent", "value")
headers.add("user-agent", u"value")

print(headers.get_all("user-agent"))
print(headers.get("user-agent"))
print(headers["user-agent"])
```

Output in Python 2:
```
['value', 'value', u'value']
value, value, value
value, value, value
```

Output in Python 3:

```
[b'value', 'value', 'value']
value, value, value
value, value, value
```

If we were to encode to bytes on ingest the `.get_all()` output would be different.

sigmavirus24:
> @sigmavirus24 What are you referring to with "there" and "elsewhere"?

So `__getitem__` is only normalizing values. This doesn't protect against

```py
headers.add(b"user-agent", "value")
headers.add(b"user-agent", "other-text")
headers.add("user-agent", "yet-another-text')
```

Where httplib will get the key as bytes and something else as text. Mixing bytes and text always ends up confusing httplib. You can use one or the other but not both (at least on Py3) if my memory is serving me well.

sethmlarson:
Thanks @hramezani!

pquentin:
Thanks @sethmlarson and @sigmavirus24 for the detailed explanations! 👍

hramezani:
Thank you all for your explanations and for helping me!

pquentin:
@hramezani Would you be interested in working with us on other issues? https://github.com/urllib3/urllib3/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Proposed+Solution+Accepted%22 would be a good start!

hramezani:
@pquentin Sure! That would be great!
I will start working on them and will ping you if need help.
Thanks again.

PR #1935, opened 2020-08-25
Assume UTC timezone if none specified in Retry-After header

### LINKED ISSUE #1934: Incorrect handling of Retry-After (2.7)
`email.utils.parsedate_tz` behaves differently on python2.7/3 if the input string has no timezone. On [python3](https://docs.python.org/3/library/email.utils.html#email.utils.parsedate_tz), the last item of the return value is `0` where on [python2.7](https://docs.python.org/2/library/email.utils.html#email.utils.parsedate_tz) it is `None`. 

Since `email.utils.mktime_tz` assumes a ```None``` timezone to be the local timezone, the resulting date might be wrong when calculating `Retry-After` timeout of a value without a timezone.

To reproduce run the following test on a python2.7 machine with timezone != UTC:
`test/test_retry.py::TestRetry::test_respect_retry_after_header_sleep[Mon Jun  3 11:30:12 2019-True-1812]`

(Luckily, `freezegun` doesn't hide the issue since it freezes only the time using `tz_offset` but it doesn't patch the timezone.)

hodbn:
Closes #1934.


src/urllib3/util/retry.py
```diff
@@ -269,6 +269,12 @@ def parse_retry_after(self, retry_after):
             retry_date_tuple = email.utils.parsedate_tz(retry_after)
             if retry_date_tuple is None:
                 raise InvalidHeader("Invalid Retry-After header: %s" % retry_after)
+            if retry_date_tuple[9] is None:  # Python 2
```
sethmlarson, line 272: Could we do this?

```python
if retry_date_tuple[9] is None:  # Python 2
    retry_date_tuple = retry_date_tuple[:8] + (0,) + retry_date_tuple[9:]
```

Also we should add a comment that Python 2 parsedate_tz() returns `None` for a timezone offset instead of 0 if no timezone is given.

sethmlarson:
An additional comment now that I know how the failure was found: Can we add a test case that fails without this patch?

src/urllib3/util/retry.py
hodbn: Done, I'll add the test and ping you :)

test/tz_fake.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Switch to a locally-known timezone specified by `tz`.
+    On exit, restore the previous timezone.
+    If `tz` is `None`, do nothing.
+    """
+    if tz is None:
+        yield
+        return
+
+    if not hasattr(time, "tzset"):
```
sethmlarson, line 20: We should document in a comment when this is the case (ie `non-Unix`)

test/tz_fake.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # Make sure the new timezone exists, at least in pytz
+    if tz not in pytz.all_timezones:
+        raise ValueError("Invalid timezone specified: %r" % (tz,))
+
+    # Get the current timezone
+    try:
+        old_tz = tzlocal.get_localzone().zone
+    except ValueError:
+        pytest.skip("Cannot determine current timezone")
```
sethmlarson, line 31: Should we error out here instead of skip?

dev-requirements.txt
```diff
@@ -15,3 +15,6 @@ gcp-devrel-py-tools==0.0.15
 
 # https://github.com/GoogleCloudPlatform/python-repo-tools/issues/23
 pylint<2.0;python_version<="2.7"
+
+pytz==2020.1
+tzlocal==2.1
```
sethmlarson, line 20: Never heard of `tzlocal`, could we use `python-dateutil` `tz.gettz()` instead if we're only using "get local timezone" functionality?

test/tz_fake.py
```diff
@@ -0,0 +1,37 @@
+import pytz
+from contextlib import contextmanager
+import time
+import os
+import pytest
+import tzlocal
+
+
+@contextmanager
+def fake_timezone_ctx(tz):
```
sethmlarson, line 10: Maybe instead of `fake_` prefix we could use `stub_` prefix instead to give a better indication what's going on?

test/tz_fake.py
hodbn: Done.

test/tz_fake.py
hodbn: Done.

dev-requirements.txt
hodbn: Done, using the current time to do that.

test/tz_fake.py
hodbn: Done, also changed the filename.

PR #1936, opened 2020-08-26
Blacken using black==20.8b1

hodbn:
Black v20.8b0/1 changed docstring formatting and line length.
This was created with `nox -s blacken`.


pquentin:
Thanks 👍

PR #1937, opened 2020-08-27
Include exceptions.pyi and fields.pyi

savarin:
Addresses https://github.com/urllib3/urllib3/issues/1897, as an alternative to https://github.com/urllib3/urllib3/pull/1931.

The previous PR https://github.com/urllib3/urllib3/pull/1931 was created to incrementally add `.pyi` files from `typeshed`, with the following steps:
1. Create `.pyi` files incrementally based on dep tree
2. Port `.pyi` files in from `typeshed` as is
3. Correct mypy errors flagged by `nox -s lint`
4. Commit all stubs for all `.pyi` files created

This PR is created to illustrate another possibility, where the focus would be on porting as much of the `.pyi` files from `typeshed` first. This involves only steps 1 - 3, i.e. changes made only where errors are raised.

After all the `.pyi` files have been ported from `typeshed`, step 4 would then be completed for all the `.pyi` files.


savarin:
@sethmlarson I'm a little concerned we might have been trying to do too much with the previous PR. I created this one in light of the objective of moving as all the `.pyi` files from `typeshed` first (and making changes only where there are explicit mypy errors), and once that's done fill out the `.pyi` files. This should help us move faster, and do so in a systematic way.

Steps 1 and 2 are captured in the commit https://github.com/urllib3/urllib3/pull/1937/commits/f1fd1c0bedcd73960784262af7cfd414607df6cc (ports from `typeshed` as is), and step 3 in the commit https://github.com/urllib3/urllib3/pull/1937/commits/b3db9acdd74586a012f20ca14e851531b8b5930b for `exceptions.pyi` and https://github.com/urllib3/urllib3/pull/1937/commits/842907173670f95dffedf422a01feef71fe206e4 for `fields.pyi`.

src/urllib3/exceptions.pyi
```diff
@@ -0,0 +1,55 @@
+from typing import Any, Optional, Union, Tuple, TYPE_CHECKING
```
savarin, line 1: From https://github.com/python/typeshed/blob/master/third_party/2and3/requests/packages/urllib3/exceptions.pyi.

src/urllib3/fields.pyi
```diff
@@ -0,0 +1,26 @@
+# Stubs for requests.packages.urllib3.fields (Python 3.4)
+
+from typing import Any, Callable, Dict, Optional
```
savarin, line 3: From `typeshed` at [fields.pyi](https://github.com/python/typeshed/blob/master/third_party/2and3/requests/packages/urllib3/fields.pyi).

sethmlarson:
Also unrelated to this PR but has to be completed as a part of type stubbing everything: we need a `py.typed` file that gets included in our `MANIFEST.in` so that Mypy knows we ship our own types now.

src/urllib3/fields.pyi
```diff
@@ -0,0 +1,26 @@
...
+
+def guess_content_type(filename: str, default: str) -> str: ...
+def format_header_param(name: str, value: str) -> str: ...
```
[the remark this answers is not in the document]
savarin: @sethmlarson Annoyingly with stub files, mypy only catches errors on a subsequent import. For example https://github.com/savarin/urllib3/compare/port-types-03...savarin:port-types-03-example?expand=1 would raise the following error.

```
src/urllib3/result.py:4: error: Argument 2 to "format_header_param" has incompatible type "int"; expected "str".
```

Types added for those two.

savarin:
@sethmlarson Changes incorporated.

I also have an idea on tracking progress. We can simply count the lines returned by `mypy` to get a sense how many errors we're reducing at each change, as per https://github.com/urllib3/urllib3/commit/11dabdbdb57a5556706b6286ec46da8854e0fdcd.

Remove all existing `.pyi` files: 1116
Current master: 1105
This change: 854

sethmlarson:
Hey @savarin wanted to thank you for all this work. Want to keep you up to date with what we're planning and how that relates to type hints.

Right now I'm planning on shipping type hint support in v2 which likely won't be released until mid/late 2021. For now you can continue with type stubs and when the `master` branch becomes the "v2" branch we'll migrate your type stubs into inline type hints. This means that type stubs won't get released in v1.26 because our API will become a **lot** simpler when we drop Python 2.7 support. So for now:

- You can continue as you're doing with `.pyi` files, we'll handle the rolling into inline type hints via a tool probably
- Focus on Python 3.6+ types, don't worry about `urllib3.packages.six` or code that's written explicitly for Python 2 for example.
- If you are discouraged by this I understand, we really value what you're doing and all hope you'll still be motivated to keep contributing. Let me know if you want to discuss more via email (`sethmichaellarson@gmail.com`)

savarin:
> Hey @savarin wanted to thank you for all this work. Want to keep you up to date with what we're planning and how that relates to type hints.
> 
> Right now I'm planning on shipping type hint support in v2 which likely won't be released until mid/late 2021. For now you can continue with type stubs and when the `master` branch becomes the "v2" branch we'll migrate your type stubs into inline type hints. This means that type stubs won't get released in v1.26 because our API will become a **lot** simpler when we drop Python 2.7 support. So for now:
> 
> * You can continue as you're doing with `.pyi` files, we'll handle the rolling into inline type hints via a tool probably
> * Focus on Python 3.6+ types, don't worry about `urllib3.packages.six` or code that's written explicitly for Python 2 for example.
> * If you are discouraged by this I understand, we really value what you're doing and all hope you'll still be motivated to keep contributing. Let me know if you want to discuss more via email (`sethmichaellarson@gmail.com`)

@sethmlarson Thanks for sharing and the kind words. Re: plans, it's not a problem. Focus on 3.6+ types makes a lot of sense. Let me know if you have any further thoughts/comments on this PR.

sethmlarson (approved):
LGTM

PR #1940, opened 2020-08-30
Change 'sanity-friendly' to 'user-friendly'

sethmlarson:
Also changed the GitHub description, open to other suggestions too :)


hodbn (approved):
Good idea :) I remember an Action that detects and monitor certain words.

pquentin (commented):
Thanks! I wanted to do that since forever

Issue #1885, opened 2020-06-01
Forbid use of cadata arg in pyOpenSSL backend


MarSoft:
The class `urllib3.contrib.pyopenssl.PyOpenSSLContext` as a `load_verify_locations()` method. This method accepts `cadata` as its third parameter.
If this parameter is given then it wraps the provided certificate data in `io.BytesIO()` and passes to `OpenSSL.SSL.Context.load_verify_locations()` as `cafile` param.
But `OpenSSL.SSL.Context.load_verify_locations` does not support a file-like object for `cafile` param, it expects only file path.

This seems to break intentions of #1804.

pquentin:
Hmm, this is a tricky one, thanks for pointing this out. As you said, pyOpenSSL does not support cadata, and only accepts cafile (a file path) or capath (a directory path). 

While we *could* write the data in a temporary file, I don't think urllib3 shoud write any kind of data to the filesystem. I think the best option is here is to refuse to handle the `cadata` argument, just like #1804 does it with Python < 2.7.9.

@sethmlarson @sigmavirus24 Thoughts?

sigmavirus24:
> While we _could_ write the data in a temporary file, I don't think urllib3 shoud write any kind of data to the filesystem

Agreed

>  I think the best option is here is to refuse to handle the `cadata` argument, just like #1804 does it with Python < 2.7.9.

Makes sense to me

sethmlarson:
I'm also in agreement here, no filesystem hacks instead fail loudly.

pquentin:
@MarSoft Would you be interested in working on this?

IvanLauLinTiong:
pyOpenSSL is deprecated and will be removed in future release version 2.x (https://github.com/urllib3/urllib3/issues/2691).

PR #1946, opened 2020-09-02
Forbid use of cadata arg in pyOpenSSL backend.

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/1885

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -57,7 +57,6 @@ except ImportError:
 
 
 from socket import timeout, error as SocketError
-from io import BytesIO
 
 try:  # Platform-specific: Python 2
     from socket import _fileobject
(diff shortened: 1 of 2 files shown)


src/urllib3/contrib/pyopenssl.py
```diff
@@ -446,6 +446,8 @@ def set_ciphers(self, ciphers):
         self._ctx.set_cipher_list(ciphers)
 
     def load_verify_locations(self, cafile=None, capath=None, cadata=None):
+        if cadata is not None:
+            raise ssl.SSLError("CA data not supported in PyOpenSSL")
```
sethmlarson, line 450: Does it make more sense to use `NotImplementedError` here?

hramezani: Yes, I think you are right.
I used `ssl.SSLError` because we have something similar in 
https://github.com/urllib3/urllib3/blob/b83cc4698e9c74baab1ce17500f170e7a5170d07/src/urllib3/util/ssl_.py#L131

pquentin (changes requested):
Can you please remove the `if cadata is not None:` line below? It should have been caught by coverage, but well I have to fix (or quarantine) flaky tests first.

hramezani:
@pquentin Thanks for the review. I removed the line.

pquentin (approved):
Thanks! I guess we can use `SSLError` here too, I like foolish consistency. :)

(By the way, I just realized that coverage would not have caught this, as we don't measure coverage in the PyOpenSSL backend.)

sethmlarson:
I think we may close this in favor of https://github.com/urllib3/urllib3/pull/1947 so we can actually support this use-case instead of erroring out. Thanks for this work, @hramezani!

PR #1951, opened 2020-09-07
Skip two flaky Windows tests

pquentin:
These tests tend to throw WinError 10053, 10054 and 10061 on Windows,
which breaks the continuous integration most of the time. Since we're
not testing anything Windows-specific here, it's okay to not run those
tests on Windows.

(It would have been better to fix the actual errors, but I don't know how to do that.)

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -143,6 +143,19 @@ def notPyPy2(test):
     return wrapper
 
 
+def notWindows(test):
+    """Skips this test on Windows"""
+
+    @six.wraps(test)
+    def wrapper(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


sethmlarson (dismissed):
Seems good, what are your thoughts on adding a [`notWindows`](https://github.com/urllib3/urllib3/pull/1524/files#diff-bf5f52200bbd5eee5274f847c2de762cR169) decorator?

sethmlarson (approved):
LGTM, will merge on CI pass

pquentin:
@sethmlarson Like this? It says "is flaky" though, not "does not run"

PR #1952, opened 2020-09-07
Fix LGTM alerts

hodbn:
1. Remove [unused except block](https://lgtm.com/projects/g/urllib3/urllib3/snapshot/f0c417aa08617b25541b23de0391d97adb92d2cf/files/src/urllib3/util/ssl_.py#x5bf6644a21af1868:1). The except block was added to account for `FileNotFoundError` in Python3 but it does not hit.
2. Remove [unnecessary assignment](https://lgtm.com/projects/g/urllib3/urllib3/snapshot/f0c417aa08617b25541b23de0391d97adb92d2cf/files/src/urllib3/_collections.py#xd6dec001decbd3d4:1). The value is never used.
3. Remove [redundant dictionary item](https://lgtm.com/projects/g/urllib3/urllib3/snapshot/f0c417aa08617b25541b23de0391d97adb92d2cf/files/src/urllib3/fields.py#xb8b9b767f000eb07:1). The item is declared twice.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -58,7 +58,6 @@ class RecentlyUsedContainer(MutableMapping):
             return item
 
     def __setitem__(self, key, value):
-        evicted_value = _Null
         with self.lock:
             # Possibly evict the existing value of 'key'
             evicted_value = self._container.get(key, _Null)
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -58,6 +58,7 @@ class RecentlyUsedContainer(MutableMapping):
             return item
 
     def __setitem__(self, key, value):
+        evicted_value = _Null
         with self.lock:
             # Possibly evict the existing value of 'key'
             evicted_value = self._container.get(key, _Null)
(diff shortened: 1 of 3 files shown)


src/urllib3/util/ssl_.py
```diff
@@ -349,14 +348,9 @@ def ssl_wrap_socket(
     if ca_certs or ca_cert_dir or ca_cert_data:
         try:
             context.load_verify_locations(ca_certs, ca_cert_dir, ca_cert_data)
-        except IOError as e:  # Platform-specific: Python 2.7
+        except IOError as e:
```
sethmlarson, line 351: I'm not sure if we want to be more strict in Python 3 here?

Also should we keep `except (IOError, OSError) as e` just to enhance confidence in this change even if it's a no-op?

src/urllib3/_collections.py
```diff
@@ -58,7 +58,6 @@ def __getitem__(self, key):
             return item
 
     def __setitem__(self, key, value):
-        evicted_value = _Null
```
sethmlarson, line 61: I actually don't agree that this is an improvement. Yes we "know" what `RLock.__enter__()` won't fail but it makes sense to define the `evicted_value` outside the block since it's used outside the block after it closes.

sethmlarson (changes requested):
Agree on 2/3 improvements. Surprised the flake8 doesn't find the duplicate dict literal keys.

src/urllib3/util/ssl_.py
hodbn: I meant, we could catch a FileNotFoundErro in Python3, but I removed that for now.

PR #1953, opened 2020-09-08
Use LONG_TIMEOUT for read with retries=0

pquentin:
This is done in other tests, and avoids timing out on the initial read.


sethmlarson (approved):
Nice!

PR #1956, opened 2020-09-09
Use LONG_TIMEOUT on another timeout-based test

pquentin:
While we do have a retry here, it's needed to make two queries, and cannot be wasted on a timeout due to continuous integration being slow.

Fixes errors like https://travis-ci.org/github/urllib3/urllib3/jobs/725437480.


pquentin:
One job failed to upload its report to Codecov, hence the coverage drop

https://travis-ci.org/github/urllib3/urllib3/jobs/725502399

sethmlarson (approved):
LGTM

PR #1958, opened 2020-09-10
Consume connections better in socket-level tests

hodbn:
Add a utility that allows better timing of test responses, instead of socket consuming.

In Windows, after the second operation on a remotely closed socket a `WSAECONNABORTED` is raised.
In our socket-level tests, we implement a dummy HTTP server that only consumes data from the client socket, therefore it can't know when to stop consuming and might close the socket too early. This might cause unnecessary `send/recv`s on the already closed socket, which in turn will raise `WSAECONNABORTED` when it could be avoided.

I've added the 2 tests that are currently skipped on Windows (since [#1951](https://github.com/urllib3/urllib3/pull/1951)).
This change should prevent `WSAECONNABORTED` when testing the aforementioned tests on Windows.

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -225,6 +225,9 @@ class IPv6HTTPDummyProxyTestCase(HTTPDummyProxyTestCase):
 class ConnectionMarker(object):
     """
     Marks an HTTP(S)Connection's socket after a request was made.
+
+    Helps a test server understand when a client finished a request,
+    without implementing a complete HTTP server.
     """
 
     MARK_FORMAT = b"$#MARK%04x*!"
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -1,4 +1,5 @@
 import threading
+from contextlib import contextmanager
 
 import pytest
 from tornado import ioloop, web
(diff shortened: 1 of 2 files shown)


pquentin:
@hodbn I merged master into your branch, since it was a prerequisite before merging this anyway

dummyserver/testcase.py
```diff
@@ -217,3 +218,61 @@ class IPv6HTTPDummyProxyTestCase(HTTPDummyProxyTestCase):
...
+
+    MARK_FORMAT = b"$#MARK%04x*!"
+
+    @classmethod
+    @contextmanager
+    def mark(cls, monkeypatch):
+        """
+        Mark connections under in that context.
+        """
+        from urllib3.connection import HTTPConnection
```
pquentin, line 236: Can you please move this import at the top of the file?

pquentin:
This looks like black magic to me, but is great if it works! Would like to have another pair of eyes before I merge it

dummyserver/testcase.py
hodbn: Done :)

hodbn:
> 
> 
> This looks like black magic to me, but is great if it works! Would like to have another pair of eyes before I merge it

I've added another sentence of docs, hopefully the class purpose is more clear now.

pquentin (approved):
Let's try that! Thanks.

PleasantMachine9:
I'm guessing this is the change that makes the CI flaky?
https://github.com/urllib3/urllib3/pull/1970/checks?check_run_id=1136185374

At least the MacOS tests failed 2/2 times, and first it was the 3.8, then the 3.6 test, and different ones, with the only common thing being that they were both socket-related (Connection Refused).

hodbn:
> I'm guessing this is the change that makes the CI flaky?
> https://github.com/urllib3/urllib3/pull/1970/checks?check_run_id=1136185374
> 
I don't think these failures are related to this change.
This change only affects `test_preserve_chunked_on_redirect` and `test_preserve_chunked_on_broken_connection` for now.

PR #1961, opened 2020-09-13
Get SSLTransport tests to 100% coverage.

jalopezsilva:
The PR has a single commit which adds two additional tests (one for recv_into without a buffer, and one for unbuffered binary access through makefile), and it also enables the shutdown test. For the shutdown test I mitigated the flakiness by adding synchronization between the client and server threads.

With the three tests, we should be at 100% coverage.

### DIFF
--- a/test/contrib/test_ssltransport.py
+++ b/test/contrib/test_ssltransport.py
@@ -381,38 +381,8 @@ class TlsInTlsTestCase(SocketDummyServerTestCase):
             assert e.type in [ssl.SSLError, ssl.CertificateError]
 
     @pytest.mark.timeout(PER_TEST_TIMEOUT)
-    def test_tls_in_tls_makefile_rw_binary(self):
-        """
-        Uses makefile with read, write and binary modes.
-        """
-        self.start_destination_server()
-        self.start_proxy_server()
-
-        sock = socket.create_connection(
-            (self.proxy_server.host, self.proxy_server.port)
-        )
-        with self.client_context.wrap_socket(
-            sock, server_hostname="localhost"
-        ) as proxy_sock:
-            with SSLTransport(
-                proxy_sock, self.client_context, server_hostname="localhost"
-            ) as destination_sock:
-
-                file = destination_sock.makefile("rwb")
-                file.write(sample_request())
-                file.flush()
-
-                response = bytearray(65536)
-                wrote = file.readinto(response)
-                assert wrote is not None
-                # Allocated response is bigger than the actual response, we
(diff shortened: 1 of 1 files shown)


test/contrib/test_ssltransport.py
```diff
@@ -408,6 +411,37 @@ def test_tls_in_tls_makefile_rw_binary(self):
...
+            (self.proxy_server.host, self.proxy_server.port)
+        )
+        with self.client_context.wrap_socket(
+            sock, server_hostname="localhost"
+        ) as proxy_sock:
+            with SSLTransport(
+                proxy_sock, self.client_context, server_hostname="localhost"
+            ) as destination_sock:
+
+                file = destination_sock.makefile("rwb", buffering=0)
```
pquentin, line 432: Can you please use `pytest.mark.parametrize`? This `buffering=0` argument appears to be the only change from `test_tls_in_tls_makefile_rw_binary`. That way, if we fix one thing in one test it'll be fixed in the other too.

pquentin (dismissed):
Thank you, this is great! Being at 100% coverage makes our life so much easier than being at 99.99%, because it can never go down for good reasons like removing tested code.

Only one small thing to fix before merging this

pquentin:
The two CI errors seem real, it looks like `test_unwrap_existing_socket` is still flaky

jalopezsilva:
Yeah, seems like it is still failing. No worries, I'm on it.

jalopezsilva:
Alright, everything is passing now and the test is behaving much better. Let me know if there's anything else missing. Thank you!

test/contrib/test_ssltransport.py
```diff
@@ -151,28 +151,39 @@ def test_unbuffered_text_makefile(self):
...
 
         def shutdown_handler(listener):
             sock = listener.accept()[0]
             ssl_sock = self.server_context.wrap_socket(sock, server_side=True)
+
             request = consume_socket(ssl_sock)
             validate_request(request)
-            sock = ssl_sock.unwrap()
-            sock.send(sample_response())
+            ssl_sock.send(sample_response())
```
sethmlarson, line 167: nit: should these be `sendall()` instead of `send()`?

sethmlarson (approved):
Seems good, thanks!  Left one comment/question

test/contrib/test_ssltransport.py
jalopezsilva: Good catch, as you said, I don't think it's making much of a difference with such small buffers but it doesn't hurt to be thorough.

PR #1960, opened 2020-09-12
Move Windows CI to Actions

hodbn:
#### Add Windows to Actions
Changed the workflow a bit to allow running all steps on both Windows/macOS.

#### Remove AppVeyor
Removed AppVeyor's configuration and badge. Instead, added CI workflow badge and marked 3.9-dev testing as experimental (failure won't affect status and badge).

#### TODO
Optimize workflow for speed.

PR #1962, opened 2020-09-13
[1.25] Backport all CI improvements to remove AppVeyor

sethmlarson:
Backport of #1960 and #1957


shazow (approved):

test/with_dummyserver/test_socketlevel.py
```diff
@@ -887,7 +887,7 @@ def socket_handler(listener):
                 retries=1,
                 release_conn=False,
                 preload_content=False,
```
pquentin, line 889: If including this is intended, there were two other CI improvements recently.

pquentin (dismissed):
Is the plan to support both 1.25 and 1.26 for some time?

sethmlarson:
No, this is just the last branch that still has mentions of AppVeyor and I've completely torn it down so if we were to need to make an emergency security release best to have this all ready to go :)

Beyond that situation the `1.25-series` branch shouldn't be needed.

pquentin (approved):

pquentin (approved):

PR #1964, opened 2020-09-15
Actually use Sphinx 3 on ReadTheDocs

pradyunsg:


sethmlarson:
You can unpin Furo here too

PR #1965, opened 2020-09-16
Move nox -s lint to GitHub Actions

pquentin:
Notice how much more verbose this is, I should try to find ways to factorize things a bit. Also, should we add this file to CODEOWNERS?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,21 @@ defaults:
     shell: bash
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v1
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install dependencies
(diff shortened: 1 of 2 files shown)


pradyunsg:
May I suggest adopting pre-commit for managing the linters, and using the pre-commit GitHub action?

sethmlarson:
Can add `.github/workflows` directory to CODEOWNERS once the deploy job is migrated to GitHub Actions :)

pquentin:
@sethmlarson does that mean you're approving this change? :)

pquentin:
@sethmlarson does that mean you're approving this change? :)

sethmlarson:
Oh! Yes I can! Only reason I didn't is because I thought there was more work to do here for some reason :) thanks for the nudge

sethmlarson (approved):

pradyunsg:
@sethmlarson Fair enough! I'll mention that `nox -s lint` that runs `pre-commit` is definitely straightforward -- https://github.com/pradyunsg/furo/blob/master/noxfile.py#L73-L82 -- and now I'mma stop being noisy about adopting pre-commit. :)

PR #1966, opened 2020-09-17
Fix autodoc and intersphinx links in documentation

sethmlarson:
Mostly fixes for CPython intersphinx since we're using Python 3.x versus 2.x and making sure our autodoc links are consistent.

Also removes mentions of `VerifiedHTTPSConnection` now that it's an alias of `HTTPSConnection`.

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -5,17 +5,19 @@ from .packages.six.moves.http_client import IncompleteRead as httplib_Incomplete
 
 
 class HTTPError(Exception):
-    "Base exception used by this module."
+    """Base exception used by this module."""
+
     pass
 
 
 class HTTPWarning(Warning):
-    "Base warning used by this module."
+    """Base warning used by this module."""
+
     pass
 
 
 class PoolError(HTTPError):
-    "Base exception for errors caused within a pool."
+    """Base exception for errors caused within a pool."""
 
     def __init__(self, pool, message):
         self.pool = pool
@@ -27,7 +29,7 @@ class PoolError(HTTPError):
 
 
 class RequestError(PoolError):
-    "Base exception for PoolErrors that have associated URLs."
+    """Base exception for PoolErrors that have associated URLs."""
 
     def __init__(self, pool, url, message):
         self.url = url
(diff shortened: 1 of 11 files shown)


src/urllib3/exceptions.py
```diff
@@ -209,14 +227,15 @@ class DependencyWarning(HTTPWarning):
...
+    """Response needs to be chunked in order to read it as chunks."""
+
     pass
 
 
 class BodyNotHttplibCompatible(HTTPError):
-    """
-    Body should be httplib.HTTPResponse like (have an fp attribute which
-    returns raw chunks) for read_chunked().
+    """ "
```
pquentin, line 236: Looks like you have an extra quote here

pquentin: Should we explain why this exception still as httplib in the name?

src/urllib3/util/timeout.py
```diff
@@ -19,20 +19,26 @@
 class Timeout(object):
     """Timeout configuration.
 
-    Timeouts can be defined as a default for a pool::
+    Timeouts can be defined as a default for a pool:
```
pquentin, line 22: Those code blocks don't show up in the doc: https://urllib3--1966.org.readthedocs.build/en/1966/reference/urllib3.util.html?highlight=timeout#urllib3.util.Timeout

pquentin (dismissed):
Thanks for tackling this. I like that we're replacing httplib with http.client!

src/urllib3/util/timeout.py
sethmlarson: Ah, I think I'm missing a `:` after `code-block:`

src/urllib3/exceptions.py
sethmlarson: eh, I'm pretty sure this is a pretty rare one and the docstring explains it enough?

sethmlarson:
@pquentin Thanks for the comments! I think I fixed everything up :)

pquentin (approved):
Looks good to me!

However you got affected to https://github.com/urllib3/urllib3/pull/1967, will close/reopen.

PR #1967, opened 2020-09-18
Use LONG_TIMEOUT on test with disabled retries

pquentin:
As seen in https://github.com/urllib3/urllib3/pull/1966/checks?check_run_id=1131651057. This fix is in the same vein of https://github.com/urllib3/urllib3/pull/1956 and https://github.com/urllib3/urllib3/pull/1953. While I could try something like `Retry(total=10, read=0)`, this seems risky as it could introduce other issues. Increasing `LONG_TIMEOUT` is more likely to Just Work™️.


sethmlarson (approved):
LGTM!

Issue #1769, opened 2019-12-06
UnicodeError while processing invalid redirect


lorien:
Urllib3 version: 1.25.6
```
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): 184.95.32.74:443
/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/connectionpool.py:1004: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  InsecureRequestWarning,
DEBUG:urllib3.connectionpool:https://184.95.32.74:443 "GET / HTTP/1.1" 302 0
DEBUG:urllib3.util.retry:Incremented Retry for (url='https://184.95.32.74'): Retry(total=2, connect=None, read=None, redirect=None, status=None)
INFO:urllib3.poolmanager:Redirecting https://184.95.32.74 -> https://.../
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): ...:443
Traceback (most recent call last):
  File "/home/web/web/domscan/.env/lib/python3.5/encodings/idna.py", line 165, in encode
    raise UnicodeError("label empty or too long")
UnicodeError: label empty or too long

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "test.py", line 6, in <module>
    mgr.urlopen('GET', 'https://184.95.32.74')
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/poolmanager.py", line 369, in urlopen
    return self.urlopen(method, redirect_location, **kw)
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/poolmanager.py", line 330, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/connectionpool.py", line 376, in _make_request
    self._validate_conn(conn)
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/connectionpool.py", line 994, in _validate_conn
    conn.connect()
  File "/home/web/web/domscan/.env/lib/python3.5/site-packages/urllib3/connection.py", line 334, in connect
[...]

PR #1968, opened 2020-09-18
Fix invalid host problem in create_connection function

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/1769 based on https://github.com/urllib3/urllib3/issues/1769#issuecomment-604386564 .

### DIFF
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -1,7 +1,11 @@
 from __future__ import absolute_import
 import socket
+
+from urllib3.exceptions import LocationParseError
+
 from .wait import NoWayToWaitForSocketError, wait_for_read
 from ..contrib import _appengine_environ
+from ..packages import six
 
 
 def is_connection_dropped(conn):  # Platform-specific
@@ -57,6 +61,13 @@ def create_connection(
     # us select whether to work with IPv4 DNS records, IPv6 records, or both.
     # The original create_connection function always returns all records.
     family = allowed_gai_family()
+
+    try:
+        host.encode("idna")
+    except UnicodeError:
+        six.raise_from(
+            LocationParseError(u"Host '%s' is not a valid IDNA label" % host), None
+        )
 
     for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
         af, socktype, proto, canonname, sa = res
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -65,8 +65,8 @@ def create_connection(
     try:
         host.encode("idna")
     except UnicodeError:
-        six.raise_from(
-            LocationParseError(u"Host '%s' is not a valid IDNA label" % host), None
+        return six.raise_from(
+            LocationParseError(u"'%s', label empty or too long" % host), None
         )
 
     for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -766,7 +766,7 @@ class TestUtil(object):
     @pytest.mark.parametrize("host", [".localhost", "...", "t" * 64])
     def test_create_connection_with_invalid_idna_labels(self, host):
         with pytest.raises(LocationParseError) as ctx:
-            create_connection((label, 80))
+            create_connection((host, 80))
         assert (
             str(ctx.value)
             == "Failed to parse: '%s', label empty or too long" % host
@@ -787,7 +787,7 @@ class TestUtil(object):
     def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, host):
         getaddrinfo.return_value = [(None, None, None, None, None)]
         socket.return_value = Mock()
-        create_connection((label, 80))
+        create_connection((host, 80))
 
 
 class TestUtilSSL(object):
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -763,18 +763,19 @@ class TestUtil(object):
         header_msg.seek(0)
         assert_header_parsing(client.parse_headers(header_msg))
 
-    @pytest.mark.parametrize("label", ["...", "t" * 64])
-    def test_create_connection_with_invalid_idna_labels(self, label):
+    @pytest.mark.parametrize("host", [".localhost", "...", "t" * 64])
+    def test_create_connection_with_invalid_idna_labels(self, host):
         with pytest.raises(LocationParseError) as ctx:
             create_connection((label, 80))
         assert (
             str(ctx.value)
-            == "Failed to parse: Host '%s' is not a valid IDNA label" % label
+            == "Failed to parse: '%s', label empty or too long" % host
         )
 
     @pytest.mark.parametrize(
-        "label",
+        "host",
         [
+            "a.example.com",
             "localhost.",
             "[dead::beef]",
             "[dead::beef%en5]",
@@ -783,7 +784,7 @@ class TestUtil(object):
     )
     @patch("socket.getaddrinfo")
     @patch("socket.socket")
-    def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, label):
+    def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, host):
         getaddrinfo.return_value = [(None, None, None, None, None)]
         socket.return_value = Mock()
         create_connection((label, 80))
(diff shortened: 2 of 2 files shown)


sethmlarson (changes requested):
I think this fix doesn't require adding a new dependency. See my comment in #1769.

hramezani:
[test_dotted_fqdn](https://github.com/urllib3/urllib3/blob/d560e21dea87e9a24b94ea04873a68ac1a5f03aa/test/with_dummyserver/test_https.py#L138) fails because of my changes with this error:
`urllib3.exceptions.LocationParseError: Failed to parse: Host 'localhost.' is not a valid IDNA label`

hramezani:
> [test_dotted_fqdn](https://github.com/urllib3/urllib3/blob/d560e21dea87e9a24b94ea04873a68ac1a5f03aa/test/with_dummyserver/test_https.py#L138) fails because of my changes with this error:
> `urllib3.exceptions.LocationParseError: Failed to parse: Host 'localhost.' is not a valid IDNA label`

@sethmlarson Is the `localhost.` a valid host? If so, how should we handle it?

hodbn:
> 
> 
> > [test_dotted_fqdn](https://github.com/urllib3/urllib3/blob/d560e21dea87e9a24b94ea04873a68ac1a5f03aa/test/with_dummyserver/test_https.py#L138) fails because of my changes with this error:
> > `urllib3.exceptions.LocationParseError: Failed to parse: Host 'localhost.' is not a valid IDNA label`
> 
> @sethmlarson Is the `localhost.` a valid host? If so, how should we handle it?

The validation should be there just to make sure the idna encoding will not fail later in `_socket.getaddrinfo`.
I think you can try to idna encode the host the beforehand: `host.encode('idna')`.

hramezani:
Thanks @hodbn for your suggestion :+1:

src/urllib3/util/connection.py
```diff
@@ -58,6 +61,11 @@ def create_connection(
     # The original create_connection function always returns all records.
     family = allowed_gai_family()
 
+    try:
+        host.encode("idna")
+    except UnicodeError:
+        raise LocationParseError("Host '%s' is not a valid IDNA label" % host)
```
hodbn, line 67: 1. I think the error message should say "is not a valid IDNA name".
2. Since `LocationParseError` is not a subclass of `UnicodeError` callers that relied on the previous behavior (`except UnicodeError`) will fail to catch the new error now.

test/test_poolmanager.py
```diff
@@ -83,6 +83,11 @@ def test_nohost(self, url):
         with pytest.raises(LocationValueError):
             p.connection_from_url(url=url)
 
+    def test_invalid_host(self):
```
hodbn, line 86: Can you rename it to a more appropriate name?

hodbn (changes requested):
Just a few comments :)

src/urllib3/util/connection.py
hramezani: > Since LocationParseError is not a subclass of UnicodeError callers that relied on the previous behavior (except UnicodeError) will fail to catch the new error now.

Yes, you are right. What should we do? raise `UnicodeError` instead?

hodbn: You can change `LocationParseError` to subclass `UnicodeError` as well.

Regarding my first comment, I see now that when we normalize non-ascii hosts we use the term "label" :raised_eyebrow: :
https://github.com/urllib3/urllib3/blob/6c429006c73da330b3c7e5b349a9a4b0eaf5d609/src/urllib3/util/url.py#L315
so I guess "label" is fine.

hramezani: I made the `LocationParseError` subclass of `UnicodeError`.

test/test_poolmanager.py
hramezani: Renamed to `test_host_with_invalid_idna_labels`

src/urllib3/exceptions.py
```diff
@@ -154,7 +154,7 @@ class LocationValueError(ValueError, HTTPError):
     pass
 
 
-class LocationParseError(LocationValueError):
+class LocationParseError(UnicodeError, LocationValueError):
```
sethmlarson, line 157: I'm recommending this not be treated like a breaking change by changing the exception type on the basis of:
- `UnicodeError` clearly isn't expected to be raised from this API.
- Our test suite doesn't expect this either, nor is it documented.
- Raising `LocationParseError` is what *should* have been happening all this time, thus I view this more as a bug fix.

So we can revert adding `UnicodeError` as a subclass.

src/urllib3/util/connection.py
sethmlarson, line 67: We can use `return six.raise_from(..., None)` here.

Also since the `UnicodeError` contains a nice description of what occurred can we use that message somehow for `LocationParseError`?

test/test_poolmanager.py
```diff
@@ -83,6 +83,11 @@ def test_nohost(self, url):
         with pytest.raises(LocationValueError):
             p.connection_from_url(url=url)
 
+    def test_host_with_invalid_idna_labels(self):
+        p = PoolManager(5)
+        with pytest.raises(LocationParseError):
+            p.urlopen("GET", "https://.../")
```
sethmlarson, line 89: I'd like to verify the message on the `LocationParseError` being raised here.

It'd also be nice to have test cases for labels that are too long (ie over 64 characters), for `localhost.`, IPv6 with and without zone IDs. Also technically you can have an unencoded `.` in an IPv6 Zone ID, so we need a case for that.

sethmlarson (changes requested):
Some comments for you:

src/urllib3/exceptions.py
hramezani: Reverted!

src/urllib3/util/connection.py
hramezani: I did it as you mentioned.

test/test_poolmanager.py
hramezani: I added a test for `labels that are too long` as you mentioned.
For `localhost.` we have already a [test](https://github.com/urllib3/urllib3/blob/fc89ea40881e78028571b723b233e3a8d7d7af75/test/with_dummyserver/test_https.py#L138) which passes through the newly added code.

For IPv6 tests, I don't know how to test it. With the following test:
```python
    def test_ipv6_host(self):
        p = PoolManager(5)
        p.urlopen("GET", "http://[dead::beef]")
```
I will get `urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='dead::beef', port=80)`. I think I need to mock something because `http://[dead::beef]` passes the validation part and it is valid host from validation point of view. Could you please help me?

hodbn: You are right, testing in `PoolManager` level will require mocking the underlying layers. You can have this test in a different file to single out the parsing validation code, look at test/test_util.py for example.

hramezani: Thanks @hodbn, I moved them to `test_util.py` and added missed tests.

hodbn: Looks great 👍 
Just a small test that @sethmlarson mentioned: "an unencoded . in an IPv6 Zone ID".
Also, I think that splitting parameter values to lines is better for readbility:
```python
@pytest.mark.parametrize(
    "label",
    [
        "localhost.",
        "[dead::beef]",
        "[dead::beef%en5]",
    ],
)
```

hramezani: @hodbn Thanks.
I added `"[dead::beef%en5.]"` for `an unencoded . in an IPv6 Zone ID`, Don't know is it a valid case or not?

hodbn: It should be valid.

(Did a brief look at the RFCs, it seems that [RFC6874](https://tools.ietf.org/html/rfc6874#section-2) states that `ZoneID = 1*( unreserved / pct-encoded )`, and the unreserved charaters are defined in [RFC3986](https://tools.ietf.org/html/rfc3986#section-2.3) to be `unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"`.)

sethmlarson (approved):
Thanks @hramezani for the patch and @hodbn for reviewing!

I made two little tweaks switching the word of the exception and adding a few more test cases. Will merge once CI passes :)

sethmlarson (approved):

PR #1969, opened 2020-09-18
Use default `dict` instead of OrderedDict on Python 3.7 and above

PleasantMachine9:
This should be free performance on python 3.7 and above, as OrderedDict is slow.
Any python interpreter is required to provide insertion ordered default dict if they implement 3.7 or above.

See:
- https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6
- https://mail.python.org/pipermail/python-dev/2017-December/151283.html

Also planning to use these collection classes in another change.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -16,7 +16,15 @@ except ImportError:  # Platform-specific: No threads available
             pass
 
 
-from collections import OrderedDict
+from sys import version_info
+
+# Since Python 3.7, `dict` must be ordered (language-level requirement)
+# so it is ok to use it instead of OrderedDict, and it is faster in CPython.
+if version_info >= (3, 7):
+    OrderedDictCls = dict
+else:
+    from collections import OrderedDict as OrderedDictCls
+
 from .exceptions import InvalidHeader
 from .packages.six import ensure_str, iterkeys, itervalues, PY3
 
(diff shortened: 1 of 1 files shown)


sethmlarson:
@PleasantMachine9 Looks like we're getting failures on `popitem()` with keyword arguments being used?

PleasantMachine9:
Yeah I checked a few tests on performance locally.

The generic `dict` only support LIFO. Could do something like:
```
class OrderedDictCls(dict):
    def popitem(self, last=True):
        if last:
            return super(OrderedDictCls, self).popitem()
        k = _Null
        for k in self:
            break
        if k is _Null:
            raise KeyError("popitem(): dictionary is empty")
        return k, self.pop(k)```

But according to `timeit`, this loses most of the performance benefits. At least it does make all kinds of operations slower, I guess because `dict` is pure C, and inheriting it from python adds costs.

PleasantMachine9:
I guess I will close the PR for now, since this is actually not a trivial change at all.

I don't have any good ideas that actually perform significantly better than OrderedDict when it comes to FIFO.

pquentin:
@PleasantMachine9 Thanks for trying, though!

(We're generally not too worried about performance as execution time in urllib3 is dominated by I/O, but if something is actually slow, as shown by benchmarks, we'll be happy to fix it. For example, @sethmlarson went to great lengths to fix an import time regression a few months ago.)

Issue #1898, opened 2020-06-29
TLS session resumption by default in urllib3


PleasantMachine9:
@PleasantMachine9 Before we start writing code can we discuss within an issue what TLS session resumption by default in urllib3 would look like (where do tickets live, under which conditions do we attempt to resume sessions, what exact performance gains are we achieving by adding this support) and once we nail that down we can start a PR that accomplishes that.

_Originally posted by @sethmlarson in https://github.com/urllib3/urllib3/pull/1886#issuecomment-651308799_

PleasantMachine9:
Some short initial comments below. It's a bit disorganised becuase it's hard to collect all details.

In the current impl. proposal, the TLS sessions would live on the HTTPSConnection level, not pool level. This would mean each connection in the pool would have its own SSLSession reference (or None). I think this is most in line logically with the concept of a TLS "session". 

Sharing it between connections could yield better performance in specific cases, but I think it would make the implementation way more complex (for example, one-time use session tickets exist and those would not be poolable, and this info is not exposed on the python API of openssl).

## where do tickets live
Few things to clarify:

> **(TLSv1.3)**
> 2.2.  Resumption and Pre-Shared Key (PSK)
>
>   Although TLS PSKs can be established out of band, PSKs can also be
>   established in a previous connection and then used to establish a new
>   connection ("session resumption" or "resuming" with a PSK).  Once a
>   handshake has completed, the server can send the client a PSK
>   identity that corresponds to a unique key derived from the initial
>   handshake (see Section 4.6.1).  The client can then use that PSK
>   identity in future handshakes to negotiate the use of the associated
>   PSK.  If the server accepts the PSK, then the security context of the
>   new connection is cryptographically tied to the original connection
>   and the key derived from the initial handshake is used to bootstrap
>   the cryptographic state instead of a full handshake.  In TLS 1.2 and
>   below, this functionality was provided by "session IDs" and "session
>   tickets" [RFC5077].  **Both mechanisms are obsoleted in TLS 1.3.**
>
>   PSKs can be used with (EC)DHE key exchange in order to provide
>   forward secrecy in combination with shared keys, or can be used
>   alone, at the cost of losing forward secrecy for the application
>   data.

- (C)Python's ssl library is a wrapper around the openssl project, which is written in C. The two C structs of interest are wrapped by python's `SSLContext` and `SSLSocket`.

[...]

PR #1970, opened 2020-09-18
Disable the use of session tickets on TLSv1.2 by default

PleasantMachine9:
See #1886  and #1898 

> Since currently session resumption is not supported by urllib3,
> there is no reason to request tickets from the server.
> It takes up extra bytes in transit (~200 bytes), and raises
> some minor security concerns.
> 
> See also: https://blog.filippo.io/we-need-to-talk-about-session-tickets/

Hopefully we can add support for resumption, but until then, this flag (NO_TICKET) has no reason to be off.

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -58,10 +58,11 @@ except ImportError:
 
 
 try:
-    from ssl import OP_NO_SSLv2, OP_NO_SSLv3, OP_NO_COMPRESSION
+    from ssl import OP_NO_SSLv2, OP_NO_SSLv3, OP_NO_COMPRESSION, OP_NO_TICKET
 except ImportError:
     OP_NO_SSLv2, OP_NO_SSLv3 = 0x1000000, 0x2000000
     OP_NO_COMPRESSION = 0x20000
+    OP_NO_TICKET = 0x4000
 
 
 # A secure default.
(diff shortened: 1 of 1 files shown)


PleasantMachine9:
> test_ssl_failed_fingerprint_verification 
> where MaxRetryError("HTTPSConnectionPool(host='localhost', port=49868): Max retries exceeded with url: / (Caused by NewConne...ection.HTTPSConnection object at 0x10c3e5190>: Failed to establish a new connection: [Errno 61] Connection refused'))") = <ExceptionInfo MaxRetryError tblen=4>.value

This looks like a flaky test case to me? I don't see how my change would affect this TC.

sethmlarson:
> This looks like a flaky test case to me? I don't see how my change would affect this TC.

Our CI is a bit flaky sometimes :) might not have been caused by your change.

sethmlarson (changes requested):
It may not say so in the Python documentation, but according to OpenSSL `OP_NO_TICKET` is meant for servers only?

https://www.openssl.org/docs/man1.1.1/man3/SSL_set_options.html

sethmlarson:
So I confirmed in Wireshark that for TLS 1.2 specifically setting `OP_NO_TICKET` does indeed disable the `session_ticket (35)` extension from being sent with ClientHello.

PleasantMachine9:
Practically speaking it wouldn't make sense for it to be server-side only, so it's probably a documentation issue in openssl.

As per the RFC detailing tickets for TLS 1.2, the _client_ is the first party to indicate whether they support tickets, by including the empty extension.

So it makes sense that we should be able to tell openssl not to send that extension.

PleasantMachine9:
Trying again for the TCs

sethmlarson (approved):
This looks good to me now, thanks for putting this together!

PR #1977, opened 2020-09-22
Terminate connection when custom verification fails (SecureTransport)

### LINKED ISSUE #1976: SecureTransport does not close unverified connections correctly
SecureTransport does not handle SSL custom-verification failure correctly. Instead of sending an alert and terminating the socket, the client socket leaks thus hanging the server socket.

#### Expected behavior
When custom-verification is enabled, `SecureTransport` should terminate the connection.

#### Actual behavior
While `SecureTransport` fails custom validation, an `SSLError` is raised but the socket itself is not terminated:
https://github.com/urllib3/urllib3/blob/d79e82a698d51dc169ad07795812c3936b11bbc8/src/urllib3/contrib/securetransport.py#L446-L449
this leads to a `conn.close()`:
https://github.com/urllib3/urllib3/blob/d79e82a698d51dc169ad07795812c3936b11bbc8/src/urllib3/connectionpool.py#L754-L760
but `conn.sock` is still `None`, it will only be assigned the socket *after* (and only if) the socket is wrapped successfully:
https://github.com/urllib3/urllib3/blob/d79e82a698d51dc169ad07795812c3936b11bbc8/src/urllib3/connection.py#L389

#### Reproduction
This snippet hangs the server eventually instead of failing `NUM_CONNECTIONS` times cleanly.
```python
import logging
import socket
import ssl
import threading

import trustme
import urllib3
from urllib3.contrib.securetransport import inject_into_urllib3, extract_from_urllib3

CERT_PATH = "/tmp/ca.pem"
SSL_CTX2 = ssl.SSLContext()
NUM_CONNECTIONS = 3

def server(s):
    for i in range(NUM_CONNECTIONS):
        c, _ = s.accept()
        try:
            logging.critical("Server iter")
            c = SSL_CTX2.wrap_socket(c, server_side=True)
            c.recv(1024)
            c.sendall(b"HTTP/1.1 404 Not Found\r\nConnection: close\r\n\r\n")
            c.close()
        except Exception as e:
            logging.critical("Server ex: %r" % (e,))

def connect_and_fail(port):
    with urllib3.HTTPSConnectionPool("localhost", port, cert_reqs="REQUIRED", ca_certs=CERT_PATH) as p:
        for i in range(NUM_CONNECTIONS):
            logging.critical("Client iter")
            try:
                p.request("GET", "/", retries=False)
                logging.critical("Client ex: %r" % (e,))

def main():
    logging.basicConfig(level=logging.INFO)
    # client's chain - ca
    ca = trustme.CA()
    ca.cert_pem.write_to_path(CERT_PATH)
    # server's chain - ca2
    ca2 = trustme.CA()
    server2_cert = ca2.issue_cert(u"localhost")
    server2_cert.configure_cert(SSL_CTX2)

    # start an SSL server with ca chain
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.bind(('', 0))
    port = s.getsockname()[1]
    s.listen(10)
    th = threading.Thread(target=server, args=(s, ))
    th.start()
    
    # inject SecureTransport and try to connect with ca2 chain
    inject_into_urllib3()
    connect_and_fail(port)
    extract_from_urllib3()
    th.join()
```

hodbn:
Closes #1976.
It was fixed by manually sending TLS alerts and then terminating the connection.

I've added a helper to send TLS alerts since I couldn't find any supported way of sending a TLS alert in SecureTransport.


src/urllib3/contrib/_securetransport/record.py
```diff
@@ -0,0 +1,32 @@
+import struct
+
+
+# TODO: docs
+TLS_PROTOCOL_VERSIONS = {
```
sethmlarson, line 5: Since we don't need to yet send more than the `UNKNOWN_CA` alert record can we condense this into a single function `build_tls_unknown_ca_alert(version)`? Then we can hardcode a lot of the `0x??` values within the function. If we decide we need to send different records we can go from there.

It'd be good to move this into `_securetransport/low_level.py` as well since it's definitely low level :P Don't think it needs a separate file yet.

src/urllib3/contrib/securetransport.py
```diff
@@ -396,11 +402,43 @@ def _custom_validate(self, verify, trust_bundle):
...
+            reason = "exception: %r" % (e,)
+
+        # SecureTransport does not send an alert nor shuts down the connection.
+        rec = build_tls_alert_record(
+            self.version(), TLS_ALERT_SEVERITY_FATAL, TLS_ALERT_DESCRIPTION_UNKNOWN_CA
+        )
+        self.socket.sendall(rec)
+        # close the connection immediately
+        opts = struct.pack("ii", 1, 0)
+        self.socket.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, opts)
```
sethmlarson, line 431: Can we have a comment about the struct that `SO_LINGER` accepts so the `struct.pack()` call makes more sense?

test/with_dummyserver/test_https.py
```diff
@@ -465,29 +465,33 @@ def test_assert_fingerprint_sha256(self):
...
-
-            def _test_request(pool):
-                with pytest.raises(MaxRetryError) as cm:
-                    pool.request("GET", "/", retries=0)
-                assert isinstance(cm.value.reason, SSLError)
-
-            _test_request(https_pool)
+            e = _test_request(https_pool)
+            assert "Fingerprints did not match." in str(e)
             https_pool._get_conn()
```
sethmlarson, line 483: Why are these calls to `_get_conn()` necessary?

sethmlarson (changes requested):
Thanks for this! I'm sad that we so much micromanaging of SecureTransport sockets here but :shrug: 

Definitely need to check the TLS alert record getting sent in Wireshark somehow.

test/with_dummyserver/test_https.py
hodbn: They were added in https://github.com/urllib3/urllib3/commit/e6adf684591b71c562a17b05b0df374cd39d76be.
I don't see why they are necessary, removed them for now.

src/urllib3/contrib/securetransport.py
hodbn: Added 👍

src/urllib3/contrib/_securetransport/record.py
hodbn: Done.

pquentin:
@hodbn I merged master into your branch to fix the conflicts caused by the isort pull request.

hodbn:
> Definitely need to check the TLS alert record getting sent in Wireshark somehow.

Now asserting that the SSL server error references the alert:
https://github.com/urllib3/urllib3/blob/d5d2a88602539a7bca6fc26c686eba93be8ac874/test/with_dummyserver/test_socketlevel.py#L1460

pquentin (approved):
Looks good, thanks! I ran the new test multiple thousands of times and it does not seem to be flaky on my machine.

sethmlarson:
@pquentin Thanks for reviewing and merging this!! 🥳

PR #1982, opened 2020-09-24
Update color scheme of banner to match logo

sethmlarson:
Our square logo and banner logo colors were out of sync, favored the square logo. Will see how it looks on the RTD preview.


pquentin:
👍

PR #1998, opened 2020-09-25
FUNDING.yml: Add Gitcoin

shazow:
AFAIK Github does not support Gitcoin natively so we'd have to add it as a Custom URL.


sethmlarson (approved):
+1

PR #2000, opened 2020-09-25
Rename Retry options and defaults

### LINKED ISSUE #1916: Replace method_whitelist and DEFAULT_REDIRECT_HEADERS_BLACKLIST parameter names for the Retry method
Ref: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html

Request to replace the following parameters in the `Retry` module with neutral terms that do not have any Slavery connotation.

`method_whitelist`
`DEFAULT_REDIRECT_HEADERS_BLACKLIST`

Many thanks.

sethmlarson:
Closes #1916

Still need to add some test cases on how setting both options interacts.

### DIFF
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -37,19 +37,19 @@ class RetryMeta(type):
     def DEFAULT_METHOD_WHITELIST(cls):
         warnings.warn(
             "Using 'Retry.DEFAULT_METHOD_WHITELIST' is deprecated and "
-            "will be removed in v2.0. Use 'Retry.DEFAULT_METHOD_ALLOWLIST' instead",
+            "will be removed in v2.0. Use 'Retry.DEFAULT_METHODS_ALLOWED' instead",
             DeprecationWarning,
         )
-        return cls.DEFAULT_METHOD_ALLOWLIST
+        return cls.DEFAULT_ALLOWED_METHODS
 
     @DEFAULT_METHOD_WHITELIST.setter
     def DEFAULT_METHOD_WHITELIST(cls, value):
         warnings.warn(
             "Using 'Retry.DEFAULT_METHOD_WHITELIST' is deprecated and "
-            "will be removed in v2.0. Use 'Retry.DEFAULT_METHOD_ALLOWLIST' instead",
+            "will be removed in v2.0. Use 'Retry.DEFAULT_ALLOWED_METHODS' instead",
             DeprecationWarning,
         )
-        cls.DEFAULT_METHOD_ALLOWLIST = value
+        cls.DEFAULT_ALLOWED_METHODS = value
 
     @property
     def DEFAULT_REDIRECT_HEADERS_BLACKLIST(cls):
@@ -151,12 +151,12 @@ class Retry(object):
         If ``total`` is not set, it's a good idea to set this to 0 to account
         for unexpected edge cases and avoid infinite retry loops.
 
-    :param iterable method_allowlist:
+    :param iterable allowed_methods:
         Set of uppercased HTTP method verbs that we should retry on.
 
         By default, we only retry on methods which are considered to be
         idempotent (multiple requests with the same parameters end with the
-        same state). See :attr:`Retry.DEFAULT_METHOD_ALLOWLIST`.
+        same state). See :attr:`Retry.DEFAULT_ALLOWED_METHODS`.
 
         Set to a ``False`` value to retry on any verb.
 
@@ -167,7 +167,7 @@ class Retry(object):
 
     :param iterable status_forcelist:
         A set of integer HTTP status codes that we should force a retry on.
-        A retry is initiated if the request method is in ``method_allowlist``
+        A retry is initiated if the request method is in ``allowed_methods``
         and the response status code is in ``status_forcelist``.
 
         By default, this is disabled with ``None``.
@@ -208,8 +208,8 @@ class Retry(object):
         request.
     """
 
-    #: Default methods to be used for ``method_allowlist``
-    DEFAULT_METHOD_ALLOWLIST = frozenset(
+    #: Default methods to be used for ``allowed_methods``
+    DEFAULT_ALLOWED_METHODS = frozenset(
         ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
     )
 
@@ -230,7 +230,7 @@ class Retry(object):
         redirect=None,
         status=None,
         other=None,
-        method_allowlist=_Default,
+        allowed_methods=_Default,
         status_forcelist=None,
         backoff_factor=0,
         raise_on_redirect=True,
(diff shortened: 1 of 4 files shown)


src/urllib3/util/retry.py
```diff
@@ -159,13 +209,13 @@ class Retry(object):
...
 
-    DEFAULT_METHOD_WHITELIST = frozenset(
+    DEFAULT_METHOD_ALLOWLIST = frozenset(
         ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
     )
 
     RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
 
-    DEFAULT_REDIRECT_HEADERS_BLACKLIST = frozenset(["Authorization"])
+    DEFAULT_REMOVE_HEADERS_ON_REDIRECT = frozenset(["Authorization"])
```
sigmavirus24, line 218: If the name isn't set in concrete, I'd suggest `DEFAULT_HEADER_REMOVELIST` which matches `DEFAULT_METHOD_ALLOWLIST` better and doesn't feel unnecessarily wordy

sethmlarson: None of the new names are set, these were my first attempt :) I picked the name to match the parameter `remove_headers_on_redirect`.

src/urllib3/util/retry.py
```diff
@@ -227,6 +296,23 @@ def new(self, **kw):
...
             respect_retry_after_header=self.respect_retry_after_header,
         )
+
+        # TODO: If already given in **kw we use what's given to us
+        # If not given we need to figure out what to pass. We decide
+        # based on whether our class has the 'method_whitelist' property
+        # and if so we pass the deprecated 'method_whitelist' otherwise
+        # we use 'method_allowlist'. Remove in v2.0
+        if "method_whitelist" not in kw and "method_allowlist" not in kw:
+            if "method_whitelist" in self.__dict__:
```
sethmlarson, line 306: I'm doing this instead of `hasattr(self, "method_whitelist")` because that triggers a `getattr()` call. [Thanks Hynek for the blog post](https://hynek.me/articles/hasattr/).

test/test_retry_deprecated.py
```diff
@@ -0,0 +1,428 @@
+import mock
```
sethmlarson, line 1: This is literally a copy-paste of `test_retry.py` with extra asserts about deprecated options. Designed so that we can move along with `test_retry.py` without having to think about deprecations much but also ensure that everything that used to work still does.

src/urllib3/util/retry.py
sigmavirus24: Yeah, I think the redirect bit is important. You're right. Probably should be something like `DEFAULT_HEADER_REMOVELIST_ON_REDIRECT`? Ugh, I don't like that much. `DEFAULT_REDIRECT_HEADER_REMOVELIST`? I think we want to convey as much information as possible. Then again, that's just changing one word there from what we have now. I don't mind rewording it altogether

sethmlarson: Naming continues to be the scourge of software. :skull_and_crossbones:

shazow: I like the current name. In a vacuum, I can imagine this being called "headers to remove on redirect", and we're prefixing DEFAULT_ to it. I agree it's less symmetrical to something like `DEFAULT_ALLOWED_METHODS` but that's okay.

src/urllib3/util/retry.py
```diff
@@ -159,13 +208,16 @@ class Retry(object):
         request.
     """
 
-    DEFAULT_METHOD_WHITELIST = frozenset(
+    #: Default methods to be used for ``method_allowlist``
+    DEFAULT_METHOD_ALLOWLIST = frozenset(
```
shazow, line 212: 0.5 vote for `DEFAULT_ALLOWED_METHODS` if we're not going for symmetry with the remove headers variable naming. (Does it just add confusion that it's not actually a list?)

src/urllib3/util/retry.py
```diff
@@ -27,6 +28,49 @@
 )
 
 
+# TODO: In v2 we can remove this sentinel and metaclass with deprecated options.
```
shazow, line 31: Would be handy to have somekind of "TODO: Remove in v2: ..." fixed comment string we can search for later.

shazow:
This is great, will feel good to purge this in v2.0 :)

src/urllib3/util/retry.py
sethmlarson: I figure `v2` is enough since any mention of v2 in pre-v2 should be suspect. I'll do a big removal PR once we cut a `1.x` branch

src/urllib3/util/retry.py
sethmlarson: That's true, I'll make this change.

sethmlarson:
Thanks for review comments @sigmavirus24 and @shazow, rerequesting reviews from both of yall with the changes :)

src/urllib3/util/retry.py
```diff
@@ -27,6 +28,49 @@
...
+_Default = object()
+
+
+class RetryMeta(type):
```
shazow, line 35: Optional nit: Maybe `RetryDeprecations` or `RetryWarnings` would be more descriptive.

src/urllib3/util/retry.py
```diff
@@ -340,15 +428,26 @@ def _is_read_error(self, err):
 
     def _is_method_retryable(self, method):
         """Checks if a given HTTP method should be retried upon, depending if
-        it is included on the method whitelist.
+        it is included on the method allowlist.
```
shazow, line 431:
```suggestion
        it is included in the allowed_methods
```

test/test_retry_deprecated.py
```diff
@@ -0,0 +1,469 @@
...
+    ConnectTimeoutError,
+    InvalidHeader,
+    MaxRetryError,
+    ReadTimeoutError,
+    ResponseError,
+    SSLError,
+)
+
+
+# TODO: Remove this entire file once deprecated Retry options are removed.
```
shazow, line 19:
```suggestion
# TODO: Remove this entire file once deprecated Retry options are removed in v2.
```

For grepping purposes. :)

test/test_retry_deprecated.py
```diff
@@ -0,0 +1,469 @@
+import mock
```
shazow, line 1:
```suggestion
# This is a copy-paste of test_retry.py with extra asserts about deprecated options. It will be removed for v2.
import mock
```

shazow (dismissed):
Looks good, only minor/optional suggestions. :)

src/urllib3/util/retry.py
sethmlarson: I'm going to pass on this one since it's "Retry metaclass", maybe I should make the class private though.

shazow: Maybe I just haven't written Python in too long, it has become weird to me to add type hinting into the name of things (rather than describe what they do).

sigmavirus24 (approved):
I started down a path of "There's got to be a better way than using a metaclass" for that and ended up in a dark dark place of Python I might have to turn into a library to make better.

shazow:
@sigmavirus24 git init deprecatelib3

sigmavirus24:
Tag line: break shit more often :metal:

shazow:
DEPRECATE EVERYTHING / FEAR NOTHING!

sethmlarson:
Thanks for the reviews @shazow and @sigmavirus24!

xmo-odoo:
@SethMichaelLarson I'm not too sure about the maintenance policy of urllib3: would you accept a PR adding `stacklevel=2` to the `warn()` call? At the moment, when it pops it signals from `Retry.__init__` which makes it difficult to know where the issue comes from (unless `-Werror` which I probably provides a traceback).

Using `stacklevel=2`, `warnings` *should* point to the caller, which is where the issue really comes from.

pquentin:
@xmo-odoo Sure, we would accept a PR that does this. It would have to against the 1.26.x branch. It's true that Python makes it hard to see where warnings come from.

PR #2001, opened 2020-09-27
Add documentation for HTTPS proxies updates.

jalopezsilva:
Following on the TLS-in-TLS work from patch #1923, this PR will be updating the necessary documentation.

I expanded the advanced usage section. Let me know if there's any other documentation I might need to edit. Added image to make review easier. 

Screenshot from 2020-09-27 13-00-08

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -127,15 +127,24 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
 be different depending on the type of proxy you selected and the destination
 you're contacting.
 
-When contacting a HTTP website through a HTTP or HTTPS proxy, the request will
-be forwarded with the `absolute URI
+When contacting a HTTP website through a HTTP proxy or a HTTPS proxy, the
+request will be forwarded with the `absolute URI
 <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_.
 
-When contacting a HTTPS website through a HTTP proxy, a TCP tunnel will be
-established with a HTTP CONNECT. Afterward a TLS connection will be established
-with the destination and your request will be sent.
+When contacting a HTTPS website through a HTTP or a HTTPS proxy, a TCP tunnel
+will be established with a HTTP CONNECT. Afterward a TLS connection will be
+established with the destination and your request will be sent.
 
-Contacting HTTPS websites through HTTPS proxies is currently not supported.
+For HTTPS proxies, a TLS-in-TLS tunnel will be established. The first TLS
+connection will be to the proxy, the second one to the destination. You can
+cutomize the ``SSLContext`` used for the proxy TLS connection through the
+``proxy_ssl_context`` argument of the :class:`~poolmanager.ProxyManager` class.
+
+For HTTPS proxies we also support forwarding your requests with an `absolute
+URI <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_ if the
+``use_forwarding_for_https`` argument is set. We **strongly recommend** you
+only use this option with trusted or corporate proxies as the proxy will have
+full visibility of your requests.
 
 For SOCKS, you can use :class:`~contrib.socks.SOCKSProxyManager` to connect to
 SOCKS4 or SOCKS5 proxies. In order to use SOCKS proxies you will need to
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -127,7 +127,7 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
 be different depending on the type of proxy you selected and the destination
 you're contacting.
 
-HTTP and HTTPS proxies
+HTTP and HTTPS Proxies
 ~~~~~~~~~~~~~~~~~~~~~~
 
 Both HTTP/HTTPS proxies support HTTP and HTTPS destinations. The only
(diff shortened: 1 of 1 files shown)


docs/advanced-usage.rst
```diff
@@ -127,15 +127,24 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
 be different depending on the type of proxy you selected and the destination
 you're contacting.
 
-When contacting a HTTP website through a HTTP or HTTPS proxy, the request will
-be forwarded with the `absolute URI
+When contacting a HTTP website through a HTTP proxy or a HTTPS proxy, the
```
sethmlarson, line 130: Maybe we can add two sections here, one for "HTTP Forwarding Proxies" and one for "HTTP Tunneling Proxies"?

Should also link to our SOCKS proxy documentation from here?

Also the link "When contacting a HTTP website through an HTTP proxy or HTTPS proxy" that's not correct is it? We should draw out a table comparing HTTP/HTTPS proxy w/ HTTP/HTTPS destination with and without `use_forwarding_for_https`. Make sure all prose aligns with this table.

sethmlarson (changes requested):
Good start, here's some comments:

docs/advanced-usage.rst
jalopezsilva: Good ideas! I made a couple of changes based on your feedback:

* I added sub-sections for SOCKS/HTTP/HTTPS.
* I clarified the behavior for all combinations of destination and proxy.

Let me know your thoughts!

jalopezsilva:
Adding a second screenshot of how the new docs look.

Screenshot from 2020-10-04 13-31-16

sethmlarson:
Btw you can view a live website with the built docs in the RTD GitHub Commit Status: https://urllib3--2001.org.readthedocs.build/en/2001/

docs/advanced-usage.rst
```diff
@@ -127,15 +127,46 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
...
+    be sent.
+
+* *HTTPS proxy - HTTP destination*
+    A TLS connection will be established to the proxy and later your request
+    will be forwarded with the `absolute URI
+    <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_.
+
+* *HTTPS proxy - HTTPS destination*
+    A TLS-in-TLS tunnel will be established.  The first TLS connection will be
+    to the proxy, the second one to the destination. You can customize the
```
sethmlarson, line 156: There's no mention of `CONNECT` here, should be TLS to proxy, CONNECT to establish a TCP tunnel, then TLS to destination

docs/advanced-usage.rst
```diff
@@ -127,15 +127,46 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
...
+    ``SSLContext`` used for the proxy TLS connection through the
```
sethmlarson, line 157: We can use 

```
:class:`ssl.SSLContext`
```

to link to SSLContext docs

docs/advanced-usage.rst
```diff
@@ -127,15 +127,46 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
...
+    ``proxy_ssl_context`` argument of the :class:`~poolmanager.ProxyManager`
+    class.
+
+For HTTPS proxies we also support forwarding your requests to HTTPS destinations with
+an `absolute URI <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_ if the
+``use_forwarding_for_https`` argument is set. We **strongly recommend** you
```
sethmlarson, line 163: My first reading of this had me thinking we strongly recommend "using this option". Maybe we should take the emphasis off strongly recommend and put it on "only use this option with trusted or corporate proxies"

also nit: "argument is set to `True`"

docs/advanced-usage.rst
```diff
@@ -127,15 +127,46 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
...
+only use this option with trusted or corporate proxies as the proxy will have
+full visibility of your requests.
+
+SOCKS proxies
```
sethmlarson, line 167: title case: Proxies

docs/advanced-usage.rst
```diff
@@ -127,15 +127,46 @@ You can connect to a proxy using HTTP, HTTPS or SOCKS. urllib3's behavior will
...
+~~~~~~~~~~~~~
 
-When contacting a HTTPS website through a HTTP proxy, a TCP tunnel will be
-established with a HTTP CONNECT. Afterward a TLS connection will be established
-with the destination and your request will be sent.
-
-Contacting HTTPS websites through HTTPS proxies is currently not supported.
```
sethmlarson, line 170: While we're modifying this section can you change the pip install to:

`$ python -m pip install urllib3[socks]`

and add this `note` section:

```
.. note::
   It is recommended to use ``socks5h://`` or ``socks4a://`` schemes in
   your ``proxy_url`` to ensure that DNS resolution is done from the remote
   server instead of client-side when connecting to a domain name.
```

Change the example to use `socks5h://`

sethmlarson (changes requested):
Looking good, here's a whole bunch of comments :)

jalopezsilva:
Wow, TIL: https://urllib3--2001.org.readthedocs.build/en/2001/advanced-usage.html#http-and-https-proxies

All updated, let me know if there's anything else!

sethmlarson (approved):
This is so great and I'm very excited to ship HTTPS+HTTPS proxies! :shipit:

PR #2002, opened 2020-09-27
Deprecate negotiating TLSv1 and TLSv1.1 by default

### LINKED ISSUE #1990: [v1.26] Deprecate TLS <1.2 on v1.26.x
If we detect "TLSv1" or "TLSv1.1" from any `SSLSocket.version()` we should be emitting a `DeprecationWarning` including the host

sethmlarson:
Closes #1990

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -398,6 +398,23 @@ class HTTPSConnection(HTTPConnection):
             ssl_context=context,
         )
 
+        # If we're using all defaults and the connection
+        # is TLSv1 or TLSv1.1 we throw a DeprecationWarning
+        # for the host.
+        if (
+            default_ssl_context
+            and self.ssl_version is None
+            and hasattr(self.sock, "version")
+            and self.sock.version() in {"TLSv1", "TLSv1.1"}
+        ):
+            warnings.warn(
+                "Negotiating TLSv1/TLSv1.1 by default is deprecated "
+                "and will be disabled in urllib3 v2.0.0. Connecting to"
+                "'%s' with '%s' can be enabled by explicitly opting-in "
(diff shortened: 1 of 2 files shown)


pquentin:
The approach looks good, thanks! This breaks other tests, though, we'd have to fix that first

sethmlarson:
@pquentin Yeah I couldn't run the TLS<1.2 tests on my local machine because Ubuntu disables it via OpenSSL config now (:tada:) and upon seeing how many failures there were threw my hands up and said "another time!" :)

sethmlarson:
@pquentin Ready for a review now that tests are passing. The macOS 3.8 failure is strange, I'm not seeing any logs?

pquentin:
If you click on "View raw logs", you can see that the failure is `test_timeout_errors_cause_retries`. Most of the flaky tests left are actually socket-level, and I don't know how to fix them properly yet. I'll re-run the tests.

The new warning makes running the TLS 1.0 and 1.1 tests quite noisy, is there a way to not display them in pytest?

sethmlarson:
Hmm that's true, we can potentially set `ssl_version=...` on them so it doesn't trigger the warning?

pquentin:
Yeah, if that sounds good to me if it's easy to do. (Sorry, haven't read the actual diff for now, I'm fighting with CI instead!)

Closing/reopening to rerun due to another socketlevel flaky test. (Tests are way more flaky when the US workday starts!)

pquentin (dismissed):
Thanks! This looks good, though I'd like to see what can be done to reduce the warnings shown when running tests.

sethmlarson:
@pquentin Yeah we might have to live with the noisiness for now, any other work-around is pretty painful. We'll get away from it after v2 though!

PR #2004, opened 2020-09-28
Enforce flake8-2020

pquentin:
We don't know if we'll ever have a Python 4 or if `six.PY3` will include Python 4+ anyway or if six will still be of any use by then, but I'd like avoiding having to think about it in reviews. :)

Related discussion: https://gitter.im/python-trio/general?at=5f72328f8fe6f1196369065b

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -111,7 +111,7 @@ def blacken(session):
 
 @nox.session
 def lint(session):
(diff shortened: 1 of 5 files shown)


sethmlarson (approved):
Seems good to me. I'll be okay with "Flake8 2020" at this point. ;)

PR #2005, opened 2020-09-29
Stop testing docs in Travis

pquentin:
Let's rely on the ReadTheDocs pull request check instead. This does mean that `nox -rs docs` won't get exercised by continuous integration, which is an acceptable compromise.


sethmlarson (approved):
LGTM

PR #2007, opened 2020-10-01
Sort imports

pquentin:
isort 5.x came out in July and plays well with black, so that sounded like a nice opportunity to sort the urllib3 imports that are, honestly, quite a mess, especially in tests. I think black wants to sort imports at some point, but until then, that would be one less thing to worry about when working on urllib3.

We don't have many pull requests open thank to the hard work of @sethmlarson and @hodbn, but I'll be happy to fix the conflicts for the active ones.

What do you think?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 
-from datetime import date
 import os
 import sys
+from datetime import date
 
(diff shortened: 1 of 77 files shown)


sethmlarson (approved):
LGTM, `nox -rs format` is going to take a second to get used to :)

shazow (commented):
Yay love formatters! :)

PR #2008, opened 2020-10-01
Fix flaky test by avoiding refused connection with short timeout

pquentin:
As seen in https://github.com/urllib3/urllib3/runs/1193400148

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -458,7 +458,7 @@ class TestSocketClosing(SocketDummyServerTestCase):
 
         self._start_server(socket_handler)
         with HTTPConnectionPool(
-            self.host, self.port, timeout=SHORT_TIMEOUT, retries=True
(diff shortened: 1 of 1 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -458,7 +458,7 @@ def socket_handler(listener):
 
         self._start_server(socket_handler)
         with HTTPConnectionPool(
-            self.host, self.port, timeout=SHORT_TIMEOUT, retries=True
+            self.host, self.port, timeout=LONG_TIMEOUT, retries=True
```
sethmlarson, line 461: Can we rename this test to match new retry option terminology?

sethmlarson (changes requested):
One comment then LGTM

test/with_dummyserver/test_socketlevel.py
pquentin: Done, thanks

sethmlarson (approved):
LGTM

PR #2009, opened 2020-10-01
Configure isort for editors too

pquentin:
This way, editors can pick up the profile to use (black).

This is actually quite important, as without that some noqa comment can get messed up and break lint, with no easy way to fix that.


sethmlarson (approved):
Nice catch here, thanks @pquentin!

Issue #1997, opened 2020-09-25
[v2] Drop support for Python 3.5


sethmlarson:

sethmlarson:
Completed in #2044

PR #2012, opened 2020-10-02
Drop Python 3.5

hramezani:
Fixes  https://github.com/urllib3/urllib3/issues/1997

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python-version: [2.7, 3.6, 3.7, 3.8]
         os: [macos-latest, windows-latest]
         experimental: [false]
         include:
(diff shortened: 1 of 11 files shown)


hramezani:
Should we remove `pypy3.5-6.0` from `.travis.yml`?

sethmlarson (changes requested):
All the changes in this PR look good, thank you for them. Two things:

- We can't merge this into `master` until we cut a `v1.26.x` release stream
- I think it'd be best to remove Python 2.7 first, then remove Python 3.5 second so we can drop all the conditional (`sys.version_info < (3, 5)`) sections entirely rather than increment them to `< (3, 6)` which might make them tougher to find in the future.

For now we can have this PR here waiting for that to happen if you'd like :)

pquentin:
> Should we remove pypy3.5-6.0 from .travis.yml?

We should actually use a newer PyPy that supports Python 3.6 instead, see https://github.com/urllib3/urllib3/pull/2013

pquentin:
@hramezani Sorry, we ended up merging #2044 instead who had better timing.

The good news is that we have now a clear cut list of issues to fix before we can release 2.0: https://github.com/urllib3/urllib3/issues?q=is%3Aopen+is%3Aissue+label%3A%22Breaking+Change%22+milestone%3Av2.0

PR #2018, opened 2020-10-05
Add SKIP_HEADER for skipping automatically added headers

### LINKED ISSUE #2014: Discussion around 1.26.0 header encoding
### Subject

In #1933, we started forcing byte encoded header values into ascii prior to sending requests. This is technically spec compliant but doesn't match usage we've seen in the wild.

Requests has cases where users are passing and receiving iso-8859-1 characters in their headers (particularly Location and User-Agent). These were previously working with bytes and round tripping appropriately with external servers.

After this change those calls are failing because `ensure_str` introduces a `decode` rather than `encode`, resulting in the stacktrace below. It looks like there was some [analysis](https://github.com/urllib3/urllib3/pull/1933#issuecomment-682583692) in #1933 referencing httplib already requiring encoding. This was previously working since re-encoding an encoded byte-string to ascii won't fail in Python 2, and won't be attempted in Python 3 since `bytes` don't have an encode method.

I wanted to open this to make sure we understand impact and discuss if this is an acceptable breaking change. My initial reaction is no, but we don't have a way to quantify usage so I'd like to get other opinions.

### Reproduction
**test.py**
```python
# -*- coding: utf-8 -*-

import json
import urllib3

user_agent = u'Test-Service-Client (Schönefeld - SXF) - 1.18.0'.encode('iso-8859-1')

http = urllib3.PoolManager()
r = http.request('GET', 'https://httpbin.org/get', headers={'User-Agent': user_agent})
print(r.status)
print(json.loads(r.data)) # Displays correctly in Python3, escaped string in Python 2
```

#### urllib3 1.25.10 output:
Python 2:
```
200
{'args: {}, 'headers': { ... u'User-Agent': u'Test-Service-Client (Sch\xf6nefeld - SXF) - 1.18.0'}, ... }
```

Python 3:
```
200
{'args': {}, 'headers': { ... 'User-Agent': 'Test-Service-Client (Schönefeld - SXF) - 1.18.0' }, ... }
```

#### urllib3 1.26.0.dev0 output:
```
 File "test.py", line 13, in <module>
    r = requests.get('https://httpbin.org/get', headers={'User-Agent': user_agent})
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/api.py", line 76, in get
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/api.py", line 61, in request
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/sessions.py", line 542, in request
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/sessions.py", line 655, in send
  File "/Users/nateprewitt/Work/OpenSource/requests/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connectionpool.py", line 710, in urlopen
    chunked=chunked,
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connectionpool.py", line 400, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/connection.py", line 221, in request
    elif headers["user-agent"] == SUPPRESS_USER_AGENT:
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/_collections.py", line 158, in __getitem__
    return ", ".join([six.ensure_str(v, "ascii") for v in val[1:]])
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/_collections.py", line 158, in <listcomp>
    return ", ".join([six.ensure_str(v, "ascii") for v in val[1:]])
  File "/Users/nateprewitt/Work/OpenSource/urllib3/src/urllib3/packages/six.py", line 955, in ensure_str
    s = s.decode(encoding, errors)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf6 in position 24: ordinal not in range(128)
```

### LINKED ISSUE #2021: Can't modify type of headers object within HTTPConnection.request()
Captured from these two comments:
- https://github.com/urllib3/urllib3/pull/2018#discussion_r499988292
- https://github.com/urllib3/urllib3/pull/2018#discussion_r500699677

Follow this logic:

- Get the list of headers
- If they're all bytes return `b", ".join(val[1:])`
- Otherwise return `", ".join(val[1:])`

This way if a user uses mixed types they still get the same explosion? I'd like a little poking around with what happens when httplib is passed both bytes and str headers on Python 2.x since it had strange behavior when doing so with HTTP method and URL.

cc @jalopezsilva

sethmlarson:
Closes #2014, Closes #2021

Changed the sentinel from `SUPPRESS_USER_AGENT` to `SKIP_HEADER` for skipping 3 of the automatically added headers `Accept-Encoding`, `Host`, and `User-Agent`.

cc @nateprewitt could you test this to ensure it works as you're expecting?

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -155,7 +155,7 @@ class HTTPHeaderDict(MutableMapping):
 
     def __getitem__(self, key):
         val = self._container[key.lower()]
-        return ", ".join([six.ensure_str(v, "ascii") for v in val[1:]])
+        return ", ".join(val[1:])
 
     def __delitem__(self, key):
         del self._container[key.lower()]
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -43,7 +43,6 @@ except NameError:  # Python 2:
         pass
 
 
-from ._collections import HTTPHeaderDict
 from ._version import __version__
 from .exceptions import (
     ConnectTimeoutError,
@@ -52,7 +51,7 @@ from .exceptions import (
     SystemTimeWarning,
 )
 from .packages.ssl_match_hostname import CertificateError, match_hostname
-from .util import SKIP_USER_AGENT, connection
+from .util import SKIP_HEADER, SKIPPABLE_HEADERS, connection
 from .util.ssl_ import (
     assert_fingerprint,
     create_urllib3_context,
@@ -213,12 +212,20 @@ class HTTPConnection(_HTTPConnection, object):
 
         return _HTTPConnection.putrequest(self, method, url, *args, **kwargs)
 
+    def putheader(self, header, *values):
+        """"""
+        if SKIP_HEADER not in values:
+            _HTTPConnection.putheader(self, header, *values)
+        elif six.ensure_str(header.lower()) not in SKIPPABLE_HEADERS:
+            raise ValueError(
+                "urllib3.util.SKIP_HEADER only supports 'Accept-Encoding', 'Host', and 'User-Agent'"
+            )
+
     def request(self, method, url, body=None, headers=None):
-        headers = HTTPHeaderDict(headers if headers is not None else {})
-        if "user-agent" not in headers:
+        if headers is None:
+            headers = {"User-Agent": _get_default_user_agent()}
+        elif "user-agent" not in (k.lower() for k in headers):
             headers["User-Agent"] = _get_default_user_agent()
-        elif SKIP_USER_AGENT in headers.get_all("user-agent"):
-            del headers["user-agent"]
         super(HTTPConnection, self).request(method, url, body=body, headers=headers)
 
     def request_chunked(self, method, url, body=None, headers=None):
@@ -226,16 +233,14 @@ class HTTPConnection(_HTTPConnection, object):
         Alternative to the common request method, which sends the
         body with chunked encoding and not as one block
         """
-        headers = HTTPHeaderDict(headers if headers is not None else {})
-        skip_accept_encoding = "accept-encoding" in headers
-        skip_host = "host" in headers
+        header_keys = set([k.lower() for k in headers or ()])
+        skip_accept_encoding = "accept-encoding" in header_keys
+        skip_host = "host" in header_keys
         self.putrequest(
             method, url, skip_accept_encoding=skip_accept_encoding, skip_host=skip_host
         )
-        if "user-agent" not in headers:
-            headers["User-Agent"] = _get_default_user_agent()
-        elif SKIP_USER_AGENT in headers.get_all("user-agent"):
-            del headers["user-agent"]
+        if "user-agent" not in header_keys:
+            self.putheader("User-Agent", _get_default_user_agent())
         for header, value in headers.items():
             self.putheader(header, value)
         if "transfer-encoding" not in headers:
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -52,7 +52,7 @@ from .exceptions import (
     SystemTimeWarning,
 )
 from .packages.ssl_match_hostname import CertificateError, match_hostname
-from .util import SUPPRESS_USER_AGENT, connection
+from .util import SKIP_USER_AGENT, connection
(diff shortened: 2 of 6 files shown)


src/urllib3/_collections.py
```diff
@@ -155,7 +155,7 @@ def __setitem__(self, key, val):
 
     def __getitem__(self, key):
         val = self._container[key.lower()]
-        return ", ".join([six.ensure_str(v, "ascii") for v in val[1:]])
+        return ", ".join(val[1:])
```
sethmlarson, line 158: So this will still explode if passed bytes, which is fine (???) because it's the same behavior as in 1.25.x and before. cc @jalopezsilva

jalopezsilva: All checks are now using ``getall`` which works around the problem that I saw. We won't see ``_getitem__`` throw if the User-Agent is bytes. 

That works but it feels fragile. Shouldn't we respect whatever type the user provided? If bytes were used, let's use bytes, otherwise use strings. 

You can do something like to keep to avoid throwing if bytes.
```
b", ".join
```

src/urllib3/connection.py
```diff
@@ -217,7 +217,7 @@ def request(self, method, url, body=None, headers=None):
         headers = HTTPHeaderDict(headers if headers is not None else {})
         if "user-agent" not in headers:
             headers["User-Agent"] = _get_default_user_agent()
-        elif headers["user-agent"] == SUPPRESS_USER_AGENT:
+        elif SKIP_USER_AGENT in headers.get_all("user-agent"):
```
hodbn, line 220: Maybe we can test that?
```python
headers.add("User-Agent", "Normal")
headers.add("User-Agent", SKIP_USER_AGENT)
assert skipped
```

I'll try to add a test later :)

src/urllib3/_collections.py
sethmlarson: Hmm, yeah we could maybe do something there? Basically follow this logic:

- Get the list of headers
- If they're all bytes return `b", ".join(val[1:])`
- Otherwise return `", ".join(val[1:])`

This way if a user uses mixed types they still get the same explosion? I'd like a little poking around with what happens when httplib is passed both bytes and str headers on Python 2.x since it had strange behavior when doing so with HTTP method and URL. Probably deserves it's own issue!

src/urllib3/connection.py
sethmlarson: Good idea!

sethmlarson:
Thanks @hodbn for the test case, I added you as a coauthor to the patch you sent me :)

src/urllib3/_collections.py
sethmlarson: I captured this issue here: https://github.com/urllib3/urllib3/issues/2021

src/urllib3/connection.py
```diff
@@ -213,29 +212,35 @@ def putrequest(self, method, url, *args, **kwargs):
...
+                "urllib3.util.SKIP_HEADER only supports 'Accept-Encoding', 'Host', and 'User-Agent'"
+            )
+
     def request(self, method, url, body=None, headers=None):
-        headers = HTTPHeaderDict(headers if headers is not None else {})
-        if "user-agent" not in headers:
+        if headers is None:
+            headers = {"User-Agent": _get_default_user_agent()}
+        elif "user-agent" not in (k.lower() for k in headers):
             headers["User-Agent"] = _get_default_user_agent()
```
hodbn, line 228: I think this flow might change the original `headers`. We even have a test against that in `test_no_user_agent_header` but not when no user agent is given.

hodbn (changes requested):

sethmlarson:
@hodbn Thanks for the review, good catch on the copy issue. `HTTPConnectionPool` was saving us with a `.copy()` call due to proxy headers but we can't rely on that at the `HTTPConnection` level. Added stricter test cases for this.

hodbn (approved):

PR #2020, opened 2020-10-07
Deprecate Retry.is_retry() in favor of Retry.should_retry()

### LINKED ISSUE #1994: [v1.26] Deprecate Retry.is_retry() in favor of new API
New API should be handed the HTTPResponse object, maybe can be named `Retry.is_retryable(response)`
On the HTTPResponse object we should add a canned `.request` object (namedtuple/slotted object?) which holds on to `method`, `url`, and `headers`. This object should be similar to the one we add in v2.0 with regards to tracing (see https://github.com/urllib3/urllib3/issues/1920)

In v1.26.x we should essentially pass everything into `Retry.is_retry(...)` the way we're doing now, maybe with an extra private flag to silence the `DeprecationWarning` and only allow returning True/False like we do now. In v2.0 however we'll allow returning a `Request` object that modifies how the next request will be sent (see https://github.com/urllib3/urllib3/issues/984, https://github.com/urllib3/urllib3/issues/1172, and https://github.com/urllib3/urllib3/issues/1445)

sethmlarson:
Closes #1994 

For v2.0 we're adding an API that let's you return `True` or `False`, **or a modified Request** from "is_retry()" which is passed the entire `HTTPResponse` object. This API will have access to request info via `HTTPResponse.request.method`, `.url`, `.headers`, etc. For v1.x we're simply making `HTTPResponse.request.method` available as a read-only value to support existing v1.x functionality.

We need to add that API so we can deprecate `Retry.is_retry()` on v1.x and make the migration path easier for users moving to v2.0 that implement custom `Retry.is_retry()` behavior. Most users won't see any difference here, the only users that will see a difference are ones that subclass `Retry` with their own `is_retry()` method implemented. Hopefully because of how inflexible the previous implementation was will prove to be useful to us here :)

Maybe for v2 we can also think about a `Retry.should_redirect()` method as well that contains the `method="GET"` modification for 303 status codes? :)

TODO:
- [ ] Do we really like the name `should_retry()`? I'm not super in love with it.
- [ ] Add a test case using custom `Retry.is_retry()` implementations

cc @sigmavirus24 @nateprewitt since this has implementations in authentication would love your input on this PR!

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -154,6 +154,20 @@ def _get_decoder(mode):
     return DeflateDecoder()
 
 
+class _HTTPResponseRequest(object):
+    # This class is only used by Retry.should_retry() API
+    # and will be expanded in urllib3 v2.0. For now it is read-only.
+
+    __slots__ = ("_response",)
+
+    def __init__(self, response):
+        self._response = response
+
+    @property
+    def method(self):
+        return self._response._request_method
+
+
 class HTTPResponse(io.IOBase):
     """
     HTTP Response container.
@@ -233,6 +247,7 @@ class HTTPResponse(io.IOBase):
         self._fp_bytes_read = 0
         self.msg = msg
         self._request_url = request_url
+        self._request_method = request_method
 
         if body and isinstance(body, (six.string_types, bytes)):
             self._body = body
@@ -313,6 +328,10 @@ class HTTPResponse(io.IOBase):
         if bytes are encoded on the wire (e.g, compressed).
         """
         return self._fp_bytes_read
+
+    @property
+    def request(self):
+        return _HTTPResponseRequest(self)
 
     def _init_length(self, request_method):
         """
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -828,8 +828,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             )
 
         # Check if we should retry the HTTP response.
-        has_retry_after = bool(response.getheader("Retry-After"))
-        if retries.is_retry(method, response.status, has_retry_after):
+        if retries.should_retry(response):
             try:
                 retries = retries.increment(method, url, response=response, _pool=self)
             except MaxRetryError:
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import email
+import inspect
 import logging
 import re
 import time
@@ -213,7 +214,7 @@ class Retry(object):
         ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
     )
 
-    #: Default status codes to be used for ``status_forcelist``
+    #: Status codes that will be retried if the 'Retry-After' header is present
     RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
 
     #: Default headers to be used for ``remove_headers_on_redirect``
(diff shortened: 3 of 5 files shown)


sigmavirus24:
So I think I'm understanding a couple different things we're trying to solve here and I want to propose an alternate solution.

It sounds like "Some users want custom Retry logic/behaviour but currently `Retry` doesn't make that easy, so we're changing the API in order to make it easier for someone to implement that via sub-classing." Is that accurate? It seems like this API change isn't strictly necessary except to achieve the goal of "easier customization of Retry logic/behaviour".

---

If I'm understanding that correctly, can we flip this on its head? I've grown very wary of having libraries that are intended to be sub-classed for stuff like this. Instead I'd rather it be an object that has a default implementation but which has an API and which is passed in to Retry that does that customized logic. We'd leave Retry's other API alone, and we'd allow folks to have some kind of "Resolver" or "Decider" that we talk to inside of the `Retry` class. It probably seems like more ceremony but there's nothing to stop folks from, e.g., subclassing the default implementation or easily passing something in. It makes this API change less intrusive and confusing (it seems even you're not fond of this Seth).

--- 

Separately, it'd be really helpful to understand who (if anyone) today is using the Retry class as a parent class and passing the sub-class in.

src/urllib3/response.py
```diff
@@ -154,6 +154,20 @@ def _get_decoder(mode):
     return DeflateDecoder()
 
 
+class _HTTPResponseRequest(object):
```
shazow, line 157: Could this be the beginnings of a prebuildable Request class?

shazow (commented):
+1 to what @sigmavirus24 said, makes sense to me.

Could provide a middleware hook where you get the original request/response, and you return a new request if it should retry.

src/urllib3/response.py
sigmavirus24: I, for one, would rather an immutable Request class that has a builder friend. I know this starts to veer more towards Java land with an object that exists to build another, but there are so many foot-guns in generating an HTTP request, I think a builder could provide the necessary guard-rails. At that point, we could have a method on a Request that pre-fills a builder for a redirect. Then since the Request objects are immutable (and ideally comparable) we can say `nextreq = decider.should_retry(...)` and verify `nextreq != req`. I think the latter is important in the event someone tries to stream/upload a file-like object and wants to retry that, we shouldn't let them.

sethmlarson: I proposed the idea of using a `dataclass` in Discord along with potentially using dataclasses for other objects like Retry and likely the "Trace" object attached to an HTTPResponse. We can make the object immutable to give it the same properties as you mention with `frozen=True`. For Python 3.6 we can require the `dataclasses` module on PyPI or vendor the library until we drop Python 3.6 in v2.

I'm curious about why you mention that we shouldn't allow retrying uploading file like objects? They should be rewindable, can you elaborate on that?

sethmlarson:
@sigmavirus24 The change you're suggesting is definitely less intrusive, we might not need to make any change within v1 then if it's only going to be additive.

src/urllib3/response.py
sigmavirus24: @sethmlarson So _most_ file-like objects are rewindable. However, people have been known to do things like try to stream one response into a different request directly because our `HTTPResponse` object behaves like a file. We can't rewind sockets and so a user should be forced to handle that logic themselves. Does that make sense? We can make educated guesses around when it's safe, but I'm pretty sure a unix socket isn't rewindable either, so even if it looks like a local file it might not be perfectly rewindable

sigmavirus24: As for why I'm -1 on dataclasses and +1 on attrs:

- The standard library drifts between versions meaning that we need to support the common denominator in something like this _or_ categorically use the backported PyPI module to get consistency. Specifically, I'm expecting someone to want to steal more of the good parts of `attrs` in 3.10 and beyond and for us to want to use those meaning downstream package distributors whining that we're using the backport version on 3.8 and 3.9 which is a waste of everyone's time

- Attrs is fast and far more battle-tested than it's poor substitute in the stdlib. It has far more features, and will never have the conflict with packagers of "But this is in the stdlib!"

sethmlarson:
So I think we're decided that this API as it's proposed here isn't what we want so will be closing this and proceeding with v1.26.

src/urllib3/response.py
sethmlarson: @sigmavirus24 I'm in favor of dataclasses over attrs:

* I'm not expecting to use any special features of attrs/dataclasses, mostly need them for the standardized way to mutate a frozen data object, basically if namedtuples had a mutator we'd use them. Also once we migrate to v2 I suspect we'll be dropping Python versions at a more regular cadence now that we've shed Python 2. We won't be using the backported `dataclasses` for new features, only to support Python 3.6.

* Our position as a much-depended on package requires us to be stable, taking on a dependency from PyPI w/ the CalVer versioning scheme is riskier for us compared to using stdlib imo, not sure if anything that attrs brings to the table that we'd use is worth that risk?

* Thankfully a lot of the re-packagers have `python3` to mean 3.7 or later nowadays, so the decision should already be made for them?

PR #2025, opened 2020-10-13
Fix test URLs containing whitespaces

skazi0:
They triggered CVE-2019-9740 checks added in python here [0].
The problematic test should fail because of invalid source address but it
failed earlier because of invalid request URL. Request URLs contained string
representation of the tested source address which can contain whitespaces.
E.g. "/source_address?(\'192.0.2.255\', 0)"
The source addresses seem to be there only for information and were added as
part of [1]. Removing them from the request URL makes the tests pass again.

[0] https://github.com/python/cpython/pull/12755
[1] https://github.com/urllib3/urllib3/pull/703

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -306,5 +306,8 @@ In chronological order:
 * [Ezzeri Esa] <https://github.com/savarin>
   * Ports and extends on types from typeshed
 
+* Jacek Tomasiak <jacek.tomasiak@gmail.com>
+  * Test improvements
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -707,7 +707,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
                 self.host, self.port, source_address=addr, retries=False
             ) as pool:
                 with pytest.raises(NewConnectionError):
-                    pool.request("GET", "/source_address?{0}".format(addr))
+                    pool.request("GET", "/source_address")
 
     def test_stream_keepalive(self):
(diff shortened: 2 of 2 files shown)


sethmlarson:
Without this PR our test suite appears to be passing, are you also seeing this or am I misunderstanding something?

skazi0:
It depends on the python version you have on your CIs. If it doesn't have the CVE-2019-9740 fix included, it will pass without problems.
Note that having unescaped whitespace in the URL is wrong even if it doesn't fail the tests :o)

sethmlarson:
We percent-encode whitespace before passing to `HTTPConnection` the same way as all characters that aren't valid in a request target.

We have the latest `3.9-dev` (and `3.8` and `3.7.9`) in our CI, those should contain the patch correct?

skazi0:
OK. Maybe this is not a proper fix for master then.
We've encountered this problem in older urllib3 versions.
Probably they didn't include the encoding change yet.
Can you post some reference to the change which introduced it?

sethmlarson:
Looks like [v1.24.3](https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1243-2019-05-01):

> Apply fix for CVE-2019-9740. (Pull #1591)

skazi0:
@sethmlarson the linked PR is just throwing exception if invalid URL is passed. Do you remember when the percent-encoding of whitespace was added in urllib3?

skazi0:
OK, I think I've found it: https://github.com/urllib3/urllib3/pull/1673
This is 1.25.4 and I hit the problem in some 1.22.
BTW, https://github.com/urllib3/urllib3/blob/master/src/urllib3/connectionpool.py#L66 is clearly not valid anymore since #1673 is merged.

sethmlarson:
@skazi0 Very true, we can remove that note.

PR #2034, opened 2020-11-02
Show Ubuntu 18.04 SOCKS test failure

pquentin:
On Ubuntu 16.04 and 18.04 the socket creation times out in PySocks instead of failing as expected.

It was hard enough to get to this point. :) The next step is to figure out why, and how it can be fixed. (Or maybe a timeout is fine too? I don't know anything about SOCKS.)

This issue is blocking the migration to GitHub Actions: they do not include 16.04 at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,14 +43,25 @@ jobs:
       env: NOX_SESSION=test-3.7
     - python: 3.8
       env: NOX_SESSION=test-3.8
+    - python: 3.8
+      env: NOX_SESSION=test-3.8
+      dist: bionic
+    - python: 3.8
+      env: NOX_SESSION=test-3.8
+      dist: focal
     - python: 3.9-dev
       env: NOX_SESSION=test-3.9
     - python: nightly
       env: NOX_SESSION=test-3.10
-    - python: pypy2.7-6.0
+    - python: pypy2.7-7.3.1
       env: NOX_SESSION=test-pypy
-    - python: pypy3.5-6.0
+    - python: pypy3.5-7.0
       env: NOX_SESSION=test-pypy
+    - python: pypy3.6-7.3.1
+      env: NOX_SESSION=test-pypy
+    - python: pypy3.6-7.3.1
+      env: NOX_SESSION=test-pypy
+      dist: bionic
 
     # Extras
     - python: 2.7
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -351,7 +351,7 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
         proxy_url = "socks5h://%s:%s" % (self.host, self.port)
         with socks.SOCKSProxyManager(proxy_url) as pm:
             with pytest.raises(NewConnectionError):
-                pm.request("GET", "http://example.com", retries=False)
+                pm.request("GET", "http://example.com", retries=False, timeout=1)
             evt.set()
 
     def test_socks_with_password(self):
@@ -609,7 +609,7 @@ class TestSOCKS4Proxy(IPV4SocketDummyServerTestCase):
         proxy_url = "socks4a://%s:%s" % (self.host, self.port)
         with socks.SOCKSProxyManager(proxy_url) as pm:
             with pytest.raises(NewConnectionError):
-                pm.request("GET", "http://example.com", retries=False)
+                pm.request("GET", "http://example.com", retries=False, timeout=1)
             evt.set()
 
     def test_socks4_with_username(self):


pquentin:
I edited the title because I'm actually migrating from 16.04 to 18.04. I can't remember the Ubuntu names!

So the difference between 16.04 and 18.04 is what getaddrinfo returns:

 * 16.04 : getaddrinfo(localhost, port) = [(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', port))]
 * 18.04: getaddrinfo(localhost, port) = [(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', port)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.1.1', port))]

We're calling getaddrinfo to support IPv4+IPv6, and try each IP in turn. (Yes, this would benefit from happy eyeballs.) The specific urllib3 test exercises failure, so all calls fail. But since the urllib3 test only expects one call, the test server does not answer the second call, and we get a timeout instead of a failure.

Why would getaddrinfo return the same thing twice on 18.04?

pquentin:
The answer is as simple as: on Ubuntu 18.04, localhost is listed twice in /etc/hosts.

This means I may have to handle retries in the SOCKS test, or I could monkeypath getaddrinfo.

hodbn:
> Why would getaddrinfo return the same thing twice on 18.04?

The results are a bit different, `127.0.0.1` vs `127.0.1.1`:
```python
[(<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', port)),
 (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.1.1', port))]
```

I don't think `localhost` should resolve to `127.0.1.1` on Ubuntu, only the machine's hostname should.
We should take a look at `/etc/hosts` and pass `-o log_cli=true` to pytest to confirm the input to `getaddrinfo` and the names mapping on the runner.

hodbn:
Either way, I think this is a bug.
If the network connection to the proxy succeeded, there is no need to connect to the proxy again with a different resolved address.

pquentin:
Travis + Ubuntu 16.04:

```
$ cat /etc/hosts
## Managed by Chef on packer-5ef350f5-b1b4-97f2-7591-153f7d417944.c.eco-emissary-99515.internal :heart_eyes_cat:
## cookbook:: travis_build_environment
##     file:: templates/default/etc/cloud/templates/hosts.tmpl.erb
127.0.1.1 travis-job-de6af4d6-aa2f-4944-887f-ac6d8a7c4b47 travis-job-de6af4d6-aa2f-4944-887f-ac6d8a7c4b47 ip4-loopback xenial64
127.0.0.1 localhost nettuno travis vagrant travis-job-de6af4d6-aa2f-4944-887f-ac6d8a7c4b47 
```

Travis + Ubuntu 18.04:

```
$ cat /etc/hosts
## Managed by Chef on packer-5efb3610-907d-b1cc-c3e4-6423ffe2ee52.c.eco-emissary-99515.internal :heart_eyes_cat:
## cookbook:: travis_build_environment
##     file:: templates/default/etc/cloud/templates/hosts.tmpl.erb
127.0.1.1 localhost localhost ip4-loopback 
127.0.0.1 localhost nettuno travis vagrant travis-job-08e10069-fbe7-4cb0-ab44-910190ef7ce4 
```

You can see localhost pointing to 127.0.1.1 too.

>  If the network connection to the resolved address #⁠1 of the proxy succeeded, retries should go to the same proxy address and not to resolved address #⁠2 of the proxy.

That's the whole point! We're testing failure here, not success.

hodbn:
> 
> > If the network connection to the resolved address #⁠1 of the proxy succeeded, retries should go to the same proxy address and not to resolved address #⁠2 of the proxy.
> 
> That's the whole point! We're testing failure here, not success.

Yeah, I'm looking at `test_proxy_rejection`. It injects a failure in the SOCKS negotiation.
I think that in case of a SOCKS failure (in contrary to a network connection issue), the retry should go to the same resolved address.

pquentin:
Oh, I see what you mean. Would you like to work on that? You're the socket expert :)

hodbn:
I'm not too familiar with the proxy implementation in urllib3, but I can give it a try :)

pquentin:
Hm, but wait, in that case we specify `retries=False`, so maybe the bug is that this setting is not passed to PySocks since it's not supported. In any case, even if we consider that it's a bug, it's a PySocks bug.

hodbn:
You are right, I don't see a retry option in `create_connnection`, and it seems that `socksocket.connect` is the SOCKS negotiation, and it is treated just like a connection error.
I think we can go around it by creating our own `getaddressinfo` loop and giving PySocks an `(ip, port)` tuple.

pquentin:
I worked around the issue in #2035, so this isn't blocking the migration to 18.04 and GitHub Actions anymore.

Regarding the actual issue, we could decide to:

 1. accept this as a limitation of PySocks,
 2. fix PySocks to a) retry to the same IP when the connection succeeded and b) add an option to prevent retries,
 3. if the license permits it, vendor/integrate PySocks to urllib3 (it's currently a single .py file), give it full urllib3 retry/timeout support and move it from contrib to core.

Historically @Anorov has been receptive to patches from urllib3 developers, so if we had infinite energy, I'd go for 2. (Ironically enough, the behavior that bothers us right now was added by such a developer in https://github.com/Anorov/PySocks/pull/34.)

But I don't have infinite energy, and I'm fine with option 1, but feel free to open an issue to do more than that! Closing this pull request.

Issue #2033, opened 2020-11-02
Migrate off of Travis CI


pquentin:
What @njsmith said here: https://github.com/python-trio/trio/issues/1785.

Here's the Travis CI announcement: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing. We could also request to be special cased if we don't migrate quickly enough.

PR #2035, opened 2020-11-04
Upgrade Travis Ubuntu and PyPy versions

pquentin:
This works around the hanging test that puzzled me in #2034, unblocks #2012 (cc @hramezani), and paves the way for GitHub Actions (#2033), where the same test was an issue too for a slighly different reason.

I haven't upgraded to Ubuntu 20.04 because while Python has `ssl.PROTOCOL_TLSv1` and `PROTOCOL_TLSv1_1`, Ubuntu 20.04 configured OpenSSL to only support TLS v1.2+ and I'm not sure how we want to figure that out from the tests.

### DIFF
--- a/test/contrib/test_socks.py
+++ b/test/contrib/test_socks.py
@@ -1,8 +1,11 @@
+from __future__ import absolute_import
+
 import socket
 import threading
 from test import SHORT_TIMEOUT
 
 import pytest
+import socks as py_socks
 
 from dummyserver.server import DEFAULT_CA, DEFAULT_CERTS
 from dummyserver.testcase import IPV4SocketDummyServerTestCase
@@ -85,6 +88,25 @@ def _address_from_socket(sock):
         return _read_exactly(sock, addr_len)
     else:
         raise RuntimeError("Unexpected addr type: %r" % addr_type)
+
+
+def _set_up_fake_getaddrinfo(monkeypatch):
+    # Work around https://github.com/urllib3/urllib3/pull/2034
+    # Nothing prevents localhost to point to two different IPs. For example, in the
+    # Ubuntu set up by GitHub Actions, localhost points both to 127.0.0.1 and ::1.
+    #
+    # In case of failure, PySocks will try the same request on both IPs, but our
+    # handle_socks[45]_negotiation functions don't handle retries, which leads either to
+    # a deadlock or a timeout in case of a failure on the first address.
+    #
+    # However, some tests need to exercise failure. We don't want retries there, but
+    # can't affect PySocks retries via its API. Instead, we monkeypatch PySocks so that
+    # it only sees a single address, which effectively disables retries.
+    def fake_getaddrinfo(addr, port, family, socket_type):
+        return [(socket.AF_INET, socket_type, family, "", (addr, port))]
+
+    monkeypatch.setattr(py_socks.socket, "getaddrinfo", fake_getaddrinfo)
+
 
 
 def handle_socks5_negotiation(sock, negotiate, username=None, password=None):
@@ -334,7 +356,8 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
             with pytest.raises(NewConnectionError):
                 pm.request("GET", "http://example.com", retries=False)
 
-    def test_proxy_rejection(self):
+    def test_proxy_rejection(self, monkeypatch):
+        _set_up_fake_getaddrinfo(monkeypatch)
         evt = threading.Event()
 
         def request_handler(listener):
@@ -429,7 +452,9 @@ class TestSocks5Proxy(IPV4SocketDummyServerTestCase):
             assert response.data == b""
             assert response.headers["Server"] == "SocksTestServer"
 
-    def test_socks_with_invalid_password(self):
+    def test_socks_with_invalid_password(self, monkeypatch):
+        _set_up_fake_getaddrinfo(monkeypatch)
+
         def request_handler(listener):
             sock = listener.accept()[0]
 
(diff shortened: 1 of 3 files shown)


test/contrib/test_socks.py
```diff
@@ -87,6 +90,24 @@ def _address_from_socket(sock):
         raise RuntimeError("Unexpected addr type: %r" % addr_type)
 
 
+def _set_up_fake_getaddrinfo(monkeypatch):
```
sethmlarson, line 93: Thoughts on making this a fixture so `monkeypatch` is passed automatically? Then we can do `@pytest.mark.usefixtures("mock_getaddrinfo")`

test/contrib/test_socks.py
```diff
@@ -87,6 +90,24 @@ def _address_from_socket(sock):
...
+    #
+    # In case of failure, PySocks will try the same request on both IPs, but our
+    # handle_socks[45]_negotiation functions don't handle retries, which leads either to
+    # a deadlock or a timeout in case of a failure on the first address.
+    #
+    # However, some tests need to exercise failure. We don't want retries there, but
+    # can't affect PySocks retries via its API. Instead, we monkeypatch PySocks so that
+    # it only sees a single address, which effectively disables retries.
+    def fake_getaddrinfo(addr, port, family, socket_type):
+        return [(socket.AF_INET, socket_type, family, "", (addr, port))]
```
sethmlarson, line 106: Instead of hard-coding this value, can we instead filter out all non-`socket.AF_INET` to ensure there aren't any IPv6 connect attempts?

sethmlarson (approved):
This is great, thanks for tracking down this issue! Very excited about upgrading our Linux builders.
Had a couple of questions that might spark additional commits.

A larger thought: it'd be nice to test SOCKS on IPv6 since we're currently only testing IPv4.

test/contrib/test_socks.py
pquentin: I think that explicit is better than implicit here: it's apparent when reading the test thank monkeypatching is going on. I prefer reserving `usefixtures` magic to situations where it's required. (But good call on noting that it's possible, I hadn't realized it!)

Now that you know my thoughts, please tell me if I should still do this changes, I'll be happy to.

test/contrib/test_socks.py
pquentin: Right, it would be an issue if localhost resolved to an IPv6 since we specify `AF_INET`. So I called `getaddrinfo()` and filtered out all non `socket.AF_INET`, and still returned a one-item list, which works locally.

However, I don't understand why I need the filtering! But I do. `::1` is first on my system, and TestSocks5Proxy.test_socks_with_invalid_password fails with `Failed to establish a new connection: [Errno 61] Connection refused'` if I don't filter out the IPv6. (The two other tests work fine.)

pquentin:
`test_source_address` failed on macOS + 2.7 with the same error than on GitHub Actions but I don't understand it.

```
 ____________________ TestConnectionPool.test_source_address ____________________
  File "/Users/runner/work/urllib3/urllib3/test/with_dummyserver/test_connectionpool.py", line 706, in test_source_address
    assert r.data == b(addr[0])
AssertionError: assert '::1' == '127.0.0.1'
  - ::1
  + 127.0.0.1
```

sethmlarson:
@pquentin But this PR doesn't modify anything to do with GitHub Actions, I'm confused as well...

test/contrib/test_socks.py
sethmlarson: This is fine, my only hangup was passing a pytest fixture as a parameter when it could instead be setup within a fixture that depends on the `monkeypatch` fixture.

test/contrib/test_socks.py
sethmlarson: Might be worth running Wireshark on your system to see which addresses, ports, etc get used when you run the test?

sethmlarson (approved):

test/contrib/test_socks.py
```diff
@@ -87,6 +91,26 @@ def _address_from_socket(sock):
...
+    # Ubuntu set up by GitHub Actions, localhost points both to 127.0.0.1 and ::1.
+    #
+    # In case of failure, PySocks will try the same request on both IPs, but our
+    # handle_socks[45]_negotiation functions don't handle retries, which leads either to
+    # a deadlock or a timeout in case of a failure on the first address.
+    #
+    # However, some tests need to exercise failure. We don't want retries there, but
+    # can't affect PySocks retries via its API. Instead, we monkeypatch PySocks so that
+    # it only sees a single address, which effectively disables retries.
+    def fake_getaddrinfo(addr, port, family, socket_type):
```
hodbn, line 106: I think the correct behavior is to "disable" PySocks retries altogether by resolving the proxy hostname before PySocks and giving it the already-resolved IP address.

sethmlarson: So you're saying instead of mocking we instead hand PySocks `host="127.0.0.1"`? I'm fine with that too.

pquentin:
Closing/reopening to wake up codecov

pquentin:
@sethmlarson Can you update the Travis required checks? `continuous-integration/travis-ci/pr` is no longer needed, we need to require `Travis CI - Pull Request` instead. I think this may be confusing codecov.

test/contrib/test_socks.py
pquentin: Agreed, but this would be a larger change that should not block this pull request, in my opinion.

pquentin:
Looks like the checks are correct in other PRs. Will create a new pull request

PR #2036, opened 2020-11-04
Skip TLS versions if disabled by OpenSSL config

sethmlarson:
Tested this on Ubuntu 20.04 and appears to work properly.
Unblocks another part of https://github.com/urllib3/urllib3/issues/2033, onwards to GHA!

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -14,6 +14,7 @@ try:
 except ImportError:
     brotli = None
 
+from dummyserver.testcase import HTTPSDummyServerTestCase
 from urllib3 import util
 from urllib3.exceptions import HTTPWarning
 from urllib3.packages import six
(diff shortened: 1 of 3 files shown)


pquentin (changes requested):
Note that this is required to use Ubuntu 20.04, but it's orthogonal to the switch from Travis to GitHub Actions, as both of them offer 18.04 and 20.04.

Should we use this mechanism for all TLS versions, including TLS 1.3? Or does that mean we risk not testing TLS 1.3 and not seeing it in the coverage?

Haven't looked into it, but all tests appear to break with:

```
ImportError while loading conftest '/home/travis/build/urllib3/urllib3/test/conftest.py'.
test/__init__.py:71: in <module>
    _sock = ssl_.ssl_wrap_socket(_sock, cert_reqs="NONE", ssl_version=_ssl_version)
.nox/test-3-8/lib/python3.8/site-packages/urllib3/util/ssl_.py:369: in ssl_wrap_socket
    context = create_urllib3_context(ssl_version, cert_reqs, ciphers=ciphers)
.nox/test-3-8/lib/python3.8/site-packages/urllib3/util/ssl_.py:308: in create_urllib3_context
/opt/python/3.8.0/lib/python3.8/ssl.py:720: in verify_mode
    super(SSLContext, SSLContext).verify_mode.__set__(self, value)
E   TypeError: an integer is required (got type str)
```

sethmlarson:
@pquentin Huh, I definitely ran locally and it worked but now running again today it fails with the error on CI. Oh well, I pushed the fix :)

We can do TLS 1.3 as well, but it'd require a slightly different method as there is no "PROTOCOL_TLSv1.3", you can't disable TLS 1.3 via SSLContext unless you specify a single version like `PROTOCOL_TLSv1.2`.

sethmlarson:
Welp, looks like we're hitting https://github.com/tornadoweb/tornado/issues/2751 on Windows, let me make this global variable into a fixture.

sethmlarson:
Okay, changed the `requires_tls_x` into fixtures so they can benefit from the windows event loop session fixture. Let me know what you think @pquentin

pquentin (approved):
Thanks!

sethmlarson:
Merging without Travis completed since all Python versions were tested on all OSes, only the specialized tests were not executed.

PR #2039, opened 2020-11-08
Fix improper argument order in SSLTransport. 

jalopezsilva:
There are two commits on this PR.

The first commit fixes a bug that was caused by an incorrect order of constructor arguments. The ``server_hostname`` parameter was not properly being passed through to the ``wrap_bio`` method call. This caused issues with SNI hints not properly working and it affected some servers. It was detected in production by a team and I have verified the fix works as expected.

The second commit addresses some comments @sethmlarson had made over Discord. We can chat more about them if needed.


sethmlarson (approved):
These changes are all awesome, thank you!

sethmlarson (approved):

PR #2044, opened 2020-11-11
Drop support for EOL Python 2.7 and 3.5

hugovk:
Fixes https://github.com/urllib3/urllib3/issues/883 (or at least does a lot for it).

This drops support for EOL Python 2.7, upgrades Python syntax for 3.5+ using pyupgrade, and removes a lot of old compatibility code including most six.

There's still some six I was't sure how to replace, like `ensure_str` and `reraise`. Maybe they should go in a utility file? And there's still a few other places that can be cleaned up, but this PR is already quite big so they could be done in followups.

One thing: I switched the Travis `NOX_SESSION=app_engine` job from 2.7 to 3.7, but it's now failing because pytest isn't finding any tests. Any ideas what's needed here?


sethmlarson:
Nice! Initial comment: can you upgrade the syntax to 3.6+ and drop Python 3.5? I know there's a PR already out for it but I'm thinking specifically about the .format() calls, would prefer going straight to f-strings to avoid git blame noise.

sethmlarson:
Also can you run black and isort on each commit so we can eventually rebase merge?

hugovk:
Python 3.5 dropped as well. Before merge, it might be worth squashing all these into a single commit (or using the squash and merge button in GitHub) to minimise the Git blame noise.

pquentin:
This is blocked by https://github.com/urllib3/urllib3/pull/2035 so that we can continue testing PyPy support for Python 3.x. We also need to understand and fix the appengine failure.

Thanks for keeping the commits separate for now, reviewing it all the changes together would have been impossible otherwise!

sethmlarson:
Gonna cc @crwilcox for the App Engine failure in case he has any insights :)

crwilcox:
At least locally I am seeing the same GAE failures on urllib3:master as this branch. Looking into it now.

crwilcox:
@sethmlarson I don't believe these tests are relevant for Python 3+ versions. This is "App Engine Standard Py2.7" stuff. 

The reason you aren't seeing tests is this chunk in conftest.py

```
def pytest_ignore_collect(path, config):
    """Skip App Engine tests in python 3 or if no SDK is available."""
    if "appengine" in str(path):
        if not six.PY2:
            return True
        if not os.environ.get("GAE_SDK_PATH"):
            return True
    return False
```

hugovk:
@crwilcox Thanks for checking! Can we remove the `test/appengine/` directory?

crwilcox:
@hugovk Please do 😊

hugovk:
Done and rebased to resolve conflict! Also simplified the nox install.

The required 2.7 and 3.5 "Expected — Waiting for status to be reported" will need to be removed from the branch protection rules.

src/urllib3/connection.py
```diff
@@ -253,7 +234,7 @@ def request_chunked(self, method, url, body=None, headers=None):
         self.endheaders()
 
         if body is not None:
-            stringish_types = six.string_types + (bytes,)
+            stringish_types = (str,) + (bytes,)
```
sethmlarson, line 237: This can be combined to `(str, bytes)` and moved into the `isinstance` call

src/urllib3/contrib/securetransport.py
```diff
@@ -230,7 +229,7 @@ def _read_callback(connection_id, data_buffer, data_length_pointer):
             while read_count < requested_length:
                 if timeout is None or timeout >= 0:
                     if not util.wait_for_read(base_socket, timeout):
-                        raise socket.error(errno.EAGAIN, "timed out")
+                        raise OSError(errno.EAGAIN, "timed out")
```
sethmlarson, line 232: Are these transformations valid?

src/urllib3/filepost.py
```diff
@@ -17,8 +14,7 @@ def choose_boundary():
     Our embarrassingly-simple replacement for mimetools.choose_boundary.
     """
     boundary = binascii.hexlify(os.urandom(16))
-    if not six.PY2:
-        boundary = boundary.decode("ascii")
+    boundary = boundary.decode("ascii")
```
sethmlarson, line 17: Can we combine all these lines into a single `return` statement? (also `.decode()` instead of `.decode("ascii")`)

src/urllib3/response.py
```diff
@@ -234,7 +231,7 @@ def __init__(
         self.msg = msg
         self._request_url = request_url
 
-        if body and isinstance(body, (six.string_types, bytes)):
+        if body and isinstance(body, ((str,), bytes)):
```
sethmlarson, line 234: This needs a fix

src/urllib3/util/url.py
```diff
@@ -403,7 +400,7 @@ def parse_url(url):
 
     # Ensure that each part of the URL is a `str` for
     # backwards compatibility.
-    if isinstance(url, six.text_type):
+    if isinstance(url, str):
```
sethmlarson, line 403: We can use `six.ensure_text` for all `ensure_func` here since text==str

src/urllib3/util/wait.py
```diff
@@ -37,36 +30,6 @@ class NoWayToWaitForSocketError(Exception):
 # So: on Windows we use select(), and everywhere else we use poll(). We also
 # fall back to select() in case poll() is somehow broken or missing.
 
-if sys.version_info >= (3, 5):
```
sethmlarson, line 40: This feels really cathartic to get rid of having written this code :)

sethmlarson (changes requested):
Two tiny comments, beyond that I've looked at everything and it looks fantastic. Thank you for all this!

src/urllib3/contrib/securetransport.py
hugovk: https://docs.python.org/3/library/exceptions.html#exception-hierarchy says:

> Changed in version 3.3: `EnvironmentError`, `IOError`, `WindowsError`, `socket.error`, `select.error` and `mmap.error` have been merged into `OSError`, and the constructor may return a subclass.

See also https://docs.python.org/3/whatsnew/3.3.html#pep-3151-reworking-the-os-and-io-exception-hierarchy, https://github.com/asottile/pyupgrade/issues/141

sethmlarson (dismissed):
LGTM, did you want to review as well @pquentin?

pquentin:
It looks like coverage is dropping, I wouldn't be comfortable merging this unless we stay at 100%

pquentin:
I also don't understand why we're removing the AppEngine tests but are keeping the AppEngine code? 

@crwilcox Do you know if should:
 * remove src/urllib3/contrib/appengine.py?
 * remove the code marked with `# Platform-specific: Appengine.` too?
 * remove `is_using_tunnel()`?

If not, should they be tested differently? (I know nothing about Google App Engine, sorry.)

hugovk:
> It looks like coverage is dropping, I wouldn't be comfortable merging this unless we stay at 100%


# _appengine_environ.py 

Coverage drops are in `_appengine_environ.py`:

image

* https://codecov.io/gh/urllib3/urllib3/src/1510dfadbc7a1d419d22a3a57919387ac13b1b53/src/urllib3/contrib/_appengine_environ.py

If `src/urllib3/contrib/appengine.py` is deleted, looks like `_appengine_environ.py` can also be deleted, or at least `is_prod_appengine_mvms()`.

#  url.py

And also:

image

* https://codecov.io/gh/urllib3/urllib3/src/1510dfadbc7a1d419d22a3a57919387ac13b1b53/src/urllib3/util/url.py

Is it safe to remove lines 272-273 or do we need a new test?

pquentin:
I think we want to continue supporting host as bytes, yes, even if this should be a really rare case and would probably never have been supported if we had started with Python 3. +1 for a test

crwilcox:
@pquentin the appengine code should be specifically for [App Engine Standard, Python 2.7](https://cloud.google.com/appengine/docs/standard/python). I can review the specific removals to make sure some of this isn't for other App Engine specific tweaks.

crwilcox:
@pquentin it appears to me htis is all for Python 2 specific AppEngine as APPENGINE_RUNTIME env var is "" on Python 3, and SERVER_SOFTWARE gunicorn.

The conditional logic of `/src/urllib3/contrib/_appengine_environ.py` makes it seems this code is only used in the Python 2 case.

hugovk:
# url.py

Coverage: I don't think it's possible to run `_normalize_host(host, scheme)` with a bytes `host`.

It's called in:

1. `util.url.parse_url`
2. `connectionpool._normalize_host`, called by `ConnectionPool.__init__`
3. `connectionpool._normalize_host`, called by `HTTPConnectionPool.is_same_host`

https://github.com/hugovk/urllib3/commit/04fd14e601943725300a39ec2f371f77b9cad1ec is an attempt (in another branch) to test those with a bytes `host`, but they all [fail with `TypeError`](https://github.com/hugovk/urllib3/runs/1397934629?check_suite_focus=true):

```
__ TestConnectionPool.test_same_host[http://google.com/-http://google.com/1] ___
  File "/Users/runner/work/urllib3/urllib3/test/test_connectionpool.py", line 86, in test_same_host
    with connection_from_url(a) as c:
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/connectionpool.py", line 1040, in connection_from_url
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/util/url.py", line 419, in get_host
    p = parse_url(url)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/util/url.py", line 354, in parse_url
    if not SCHEME_RE.search(url):
TypeError: cannot use a string pattern on a bytes-like object
_____________________ TestConnectionPool.test_bytes_host1 ______________________
  File "/Users/runner/work/urllib3/urllib3/test/test_connectionpool.py", line 91, in test_bytes_host1
    HTTPConnection(host)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/connection.py", line 108, in __init__
    _HTTPConnection.__init__(self, *args, **kw)
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/http/client.py", line 863, in __init__
    (self.host, self.port) = self._get_hostport(host, port)
  File "/Users/runner/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/http/client.py", line 897, in _get_hostport
    i = host.rfind(':')
TypeError: a bytes-like object is required, not 'str'
_____________________ TestConnectionPool.test_bytes_host2 ______________________
  File "/Users/runner/work/urllib3/urllib3/test/test_connectionpool.py", line 96, in test_bytes_host2
    c.is_same_host(b"http://google.com")
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/connectionpool.py", line 498, in is_same_host
    if url.startswith("/"):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str
___________________________ TestUtil.test_bytes_url ____________________________
  File "/Users/runner/work/urllib3/urllib3/test/test_util.py", line 169, in test_bytes_url
    parse_url(b"http://google.com")
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-6/lib/python3.6/site-packages/urllib3/util/url.py", line 354, in parse_url
    if not SCHEME_RE.search(url):
TypeError: cannot use a string pattern on a bytes-like object
```

Any other ideas? We could try testing `_normalize_host` directly, but looks like other `_internal` functions are not tested directly?

sethmlarson:
@crwilcox Just confirming so it's explicit, you're in favor of removing `contrib/_app_engine_environ.py`?

@hugovk If it's unreachable from public APIs we can remove it. My guess is the branch was only reachable from Python 2 because of str==bytes. I don't think we support bytes as host in Python 3.

crwilcox:
@sethmlarson 👍

hugovk:
Updated, coverage is 100% and ready for review!

pquentin:
@hugovk Fantastic! Can I ask you to open a new pull request for Google App Engine removal? This will make this one more palatable and I wouldn't want GAE removal to be in the same commit.

Sorry that I'm asking you to do more work here! Hopefully your clean separation of commits should make it easy.

@crwilcox Thanks a lot for your help here! We'll continue to support GAE in 1.x but it's good that we won't need to in 2.x.

sethmlarson:
(@hugovk Or a separate commit for GAE removal is fine too)

pquentin:
As discussed on Discord, we're going to take it over from here, by separating the GAE changes from the rest. Thanks @hugovk!

sethmlarson:
Okay I rearranged the commits into two groups and then squashed them all together. Took note of the diff counts before and after the force-push and they are the same so I'm pretty sure I got everything. Either way I have a backup of the whole repository if I did mess something up somehow.

@hugovk @pquentin let me know how this looks! I took a quick glance over everything and seems to be properly split.

hugovk:
Looks good, and confirmed the squashed code is identical to the earlier, unsquashed code. Thanks!

pquentin (approved):
This looks good to me, thanks! There are a bunch of remaining cleanups to do, eg. we still mention App Engine in the Advanced Usage docs, and we should remove six and ssl_match_hostname, but it would be better to do that in follow up pull requests.

pquentin:
@sethmlarson Can't merge until you or @shazow approve. Please remember that we don't want to squash the two remaining commits. :)

sethmlarson:
Merged successfully :) Thanks @hugovk!!

src/urllib3/response.py
asottile: fwiw, this is technically fine but looks nonsense to humans -- isinstance can take any type of recursive tuple (it's identical to what was being passed before)

graingert:
@hugovk you should consider running pyupgrade in CI with pre-commit

hugovk:
Yep, I do in my projects :)

What do the urllib3 maintainers think?

pquentin:
I'd be happy to try using that! Also, I know @sethmlarson agreed to explore pre-commit itself, so a good start would be to open a PR with pyupgrade as a pre-commit hook checked in GitHub Actions? Later on we could put flake8, isort and black checks in pre-commit too.

hugovk:
Please see PR #2079.

asottile:
fwiw you may want to consider https://pre-commit.ci (open, secret alpha, should be non-secret later today if I get the docs completed -- https://results..pre-commit.ci) -- it provides several advantages over github actions

PR #2049, opened 2020-11-11
Release 1.26.1

sethmlarson:
Small release to fix reported issues now that Requests has unpinned us in 2.25.0.

One change in this release isn't documented here: the addition of `HTTPHeaderDict` back to `urllib3.connection` isn't technically in our public API, however the change is impacting users of one of our dependents and we're making the fix as a courtesy.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,13 @@
 Changes
 =======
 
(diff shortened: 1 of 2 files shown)


shazow (approved):
:+1:

PR #2050, opened 2020-11-12
Upgrade PyPy to test Python 3.6

pquentin:
Recreated #2035 to get the correct GitHub checks.

This works around the hanging test that puzzled me in #2034, unblocks #2044, and paves the way for GitHub Actions (#2033), where the same test was an issue too for a slighly different reason.

I have upgraded to Ubuntu 20.04 since @sethmlarson fixed the TLS v1/1.1 issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 os: linux
-dist: xenial
+dist: focal
 
 before_install:
   - env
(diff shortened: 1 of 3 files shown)


pquentin:
We seem to hit https://github.com/nedbat/coveragepy/issues/1010. Closing/reopening

pquentin:
The only error left is macOS 2.7 which already fails on master. I'd suggest merging this, improving the behavior of our SOCKS code can always be done later on.

sethmlarson (approved):
LGTM, thanks Quentin!

PR #2061, opened 2020-11-15
SSL feature detection

pquentin:
A follow-up of #2044 specifically for the `ssl_.py` file. I've wanted to remove that SSLContext backport for a long time now!


sethmlarson (approved):
This PR also excites me a lot too! It's great to get rid of all these backported APIs and know that we can depend on modern TLS features :muscle:

PR #2062, opened 2020-11-15
Update dev-requirements.txt

pquentin:
We no longer support Python 2.7, so we can stop worrying about the
PySocks bug and can install latest version of pytest. We also remove
gcp-devrel-py-tool that was installed to test Google App Engine.

@sethmlarson Do you plan to set up dependabot? I was surprised by this pull request: https://github.com/urllib3/urllib3/pull/2032

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,14 +1,10 @@
-coverage~=5.0
-tornado==6.0.3
+coverage==5.3
+tornado==6.1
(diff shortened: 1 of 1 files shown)


sethmlarson:
Dependabot isn't configured for all updates, only security updates. I was surprised at first too!

sethmlarson (approved):
LGTM!

PR #2063, opened 2020-11-15
Remove Python 2 support from SecureTransport makefile function

hramezani:
follow-up of #2044


sethmlarson:
Are we actually able to drop `urllib3.packages.backports.makefile`? It mentions being from Python 3.5 so might be able to remove it all-together? Haven't done much looking but that's my initial thought.

pquentin (approved):
+1 for removing `backport_makefile` if possible, but that can happen in another pull request.

PR #2066, opened 2020-11-16
Remove Python 2 related code from src/urllib3/connectionpool.py

hramezani:
follow-up of #2044

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -321,19 +321,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
                 self, url, f"Read timed out. (read timeout={timeout_value})"
             )
 
-        # See the above comment about EAGAIN in Python 3. In Python 2 we have
-        # to specifically catch it and throw the timeout error
+        # See the above comment about EAGAIN in Python 3.
         if hasattr(err, "errno") and err.errno in _blocking_errnos:
-            raise ReadTimeoutError(
-                self, url, f"Read timed out. (read timeout={timeout_value})"
-            )
-
-        # Catch possible read timeouts thrown as SSL errors. If not the
-        # case, rethrow the original. We need to do this because of:
-        # http://bugs.python.org/issue10272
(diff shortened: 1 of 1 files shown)


src/urllib3/connectionpool.py
```diff
@@ -414,18 +397,7 @@ def _make_request(
 
         # Receive the response from the server
         try:
-            try:
```
sethmlarson, line 417: Woo, this is super great and I've been waiting to remove this since day 1 :tada:

src/urllib3/connectionpool.py
```diff
@@ -364,8 +353,7 @@ def _make_request(
         # Trigger any extra validation we need to do.
         try:
             self._validate_conn(conn)
-        except (SocketTimeout, BaseSSLError) as e:
```
sethmlarson, line 367: I'm wondering if there's a way that `BaseSSLError` could be raised here that we're not aware of?

hramezani: @sethmlarson I will check it.

hramezani: @sethmlarson 
I checked it and couldn't find a way for `BaseSSLError` but we can keep `BaseSSLError` if you prefer.

pquentin:
Tests passed but we hit https://github.com/nedbat/coveragepy/issues/1010 again on PyPy. Closing/reopening

```
Couldn't use data file '/home/travis/build/urllib3/urllib3/.coverage.travis-job-baf7673a-9ba9-46b7-be88-2e66473e48c2.5892.501915': Error binding parameter 0 - probably unsupported type.
```

src/urllib3/connectionpool.py
sethmlarson: Let's keep the `BaseSSLError` in case there's "bad MAC" errors we could catch here.

hramezani: Done!

sethmlarson (approved):
LGTM, will merge once CI passes

PR #2068, opened 2020-11-17
Replace codecs.open() with builtin open()

jdufresne:
This follows commit fbe9620784b130dcdccf43cd3105940152c4d762.

In Python 3, the feature set between these two functions is the same,
might as well use the builtin version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # This file is protected via CODEOWNERS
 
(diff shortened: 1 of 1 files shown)


sethmlarson:
We're planning on "allowing" installs on Python 2 for a while due to old pip versions not respecting 'Requires-Python' so I think we'll leave setup.py how it is for now. 

Thanks though!

PR #2069, opened 2020-11-17
Use Python3 super()

jdufresne:
Follows commit fbe9620784b130dcdccf43cd3105940152c4d762.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -102,7 +102,7 @@ class HTTPConnection(_HTTPConnection):
         self.proxy = kw.pop("proxy", None)
         self.proxy_config = kw.pop("proxy_config", None)
 
-        _HTTPConnection.__init__(self, *args, **kw)
+        super().__init__(*args, **kw)
 
     @property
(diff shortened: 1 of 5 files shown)


src/urllib3/connection.py
```diff
@@ -4,7 +4,7 @@
 import re
 import socket
 import warnings
-from http.client import HTTPConnection as _HTTPConnection
```
sethmlarson, line 7: Can we revert the change here? Ensures that HTTPConnection name can only be one thing throughout this file.

sethmlarson (changes requested):
Instead of doing this `super()` upgrade file by file can you do it all in one PR if there's upgrades to be done elsewhere too?

src/urllib3/connection.py
jdufresne: done.

jdufresne:
I think I got them all, but please let me know if you notice any more.

sethmlarson (approved):
LGTM, thanks!

PR #2070, opened 2020-11-17
Fail explicitly on unsupported Python versions 

### LINKED ISSUE #1987: [v2] Detect unsupported Python versions and raise a helpful ImportError
We'll need to ensure our `urllib3/__init__.py` is still Python 2.x compatible for early urllib3 v2.x releases. If we detect an unsupported Python version we raise a helpful error message recommending pinning to `urllib3<2`

pquentin:
Closes #1987 

Please don't squash the two commits, as they're not related at all and we could want to revert one but not the other.

This changes the name of all checks, which will require changing them in the UI again. But it will allow adding PyPy and Brotlit tests easily later on. I'll merge master in all existing pull requests to make sure this isn't too disruptive.

Note that this test makes sure than `__init__.py` is still importable by Python 2, and as long as it passes we can do anything we want. For example, we can merge #2068 from @jdufresne if we want to, since it's valid Python 2 syntax.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,8 +4,37 @@
 import codecs
 import os
 import re
+import sys
 
 from setuptools import setup
+
+CURRENT_PYTHON = sys.version_info[:2]
+REQUIRED_PYTHON = (3, 6)
+
+# This check and everything above must remain compatible with Python 2.7.
+if CURRENT_PYTHON < REQUIRED_PYTHON:
+    sys.stderr.write(
+        """
+==========================
+Unsupported Python version
+==========================
+This version of urllib3 requires Python {}.{}, but you're trying to
+install it on Python {}.{}.
+This may be because you are using a version of pip that doesn't
+understand the python_requires classifier. Make sure you
+have pip >= 9.0 and setuptools >= 24.2, then try again:
+    $ python -m pip install --upgrade pip setuptools
+    $ python -m pip install urllib3
+This will install the latest version of urllib3 which works on your
+version of Python. If you can't upgrade your pip (or Python), request
+an older version of urllib3:
+    $ python -m pip install "urllib3<2"
+""".format(
+            *(REQUIRED_PYTHON + CURRENT_PYTHON)
+        )
+    )
+    sys.exit(1)
+
 
 base_path = os.path.dirname(__file__)
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,10 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: google_brotli
+          - python-version: 2.7
+            os: ubuntu-latest
+            experimental: false
+            nox-session: unsupported_python2
 
     runs-on: ${{ matrix.os }}
     name: ${{ fromJson('{"macos-latest":"macOS","windows-latest":"Windows","ubuntu-latest":"Ubuntu"}')[matrix.os] }} ${{ matrix.python-version }} ${{ matrix.nox-session}}
--- a/noxfile.py
+++ b/noxfile.py
@@ -70,6 +70,20 @@ def test(session):
     tests_impl(session)
 
 
+@nox.session(python=["2.7"])
+def unsupported_python2(session):
+    # Can't check both returncode and output with session.run
+    process = subprocess.run(
+        ["python", "setup.py", "install"],
+        env={**session.env},
+        text=True,
+        capture_output=True,
+    )
+    assert process.returncode == 1
+    print(process.stderr)
+    assert "Unsupported Python version" in process.stderr
+
+
 @nox.session(python=["3"])
 def google_brotli(session):
     # https://pypi.org/project/Brotli/ is the Google version of brotli, so


src/urllib3/__init__.py
```diff
@@ -2,6 +2,36 @@
...
+    $ python -m pip install urllib3
+This will install the latest version of urllib3 which works on your
+version of Python. If you can't upgrade your pip (or Python), request
+an older version of urllib3:
+    $ python -m pip install "urllib3<2"
+""".format(
+            *(REQUIRED_PYTHON + CURRENT_PYTHON)
+        )
+    )
+    sys.exit(1)
```
sigmavirus24, line 32: I think `sys.exit` is probably the wrong approach. We can raise an exception to force things to stop. Why do we presume we want to always exit? That's making lots of dangerous assumptions that can really annoy users

sethmlarson: I also agree `sys.exit()` is too strong, was thinking `ImportError`?

sigmavirus24: I'd rather not `ImportError` because that can be confused with something being missing. Instead I'd suggest something like `UnsupportedRuntimeVersion` to make it really obvious.

src/urllib3/__init__.py
```diff
@@ -2,6 +2,36 @@
 Python HTTP library with thread-safe connection pooling, file post support, user friendly, and more
 """
```
graingert, line 4:
```suggestion
from __future__ import generator_stop  # urllib 3 needs python 3.6+
```

pquentin: @graingert We can't put this here because the top of the file needs to be valid Python 2.7, which this line isn't! This is an interesting option, but out of scope for this pull request

graingert: The user will see the comment in the traceback

graingert: eg like this:

```pytb
  File "urllib3/__init__.py", line 1
    from __future__ import generator_stop  # urllib3 needs python 3.6+
SyntaxError: future feature generator_stop is not defined
```

graingert: then when 3.6 is dropped you can upgrade to `from __future__ import annotations`

graingert:
@pquentin wouldn't it be cleaner to have this check in the `setup.py` - and only ship a `.py3.` wheel?

src/urllib3/__init__.py
sethmlarson: I'd prefer to be explicit, if a library did this it wouldn't be immediately obvious to me that my Python version isn't supported.

sethmlarson:
@graingert Some people likely vendor urllib3 source code into their own, having it here prevents that.

jdufresne:
> would it be cleaner to have this check in the setup.py - and only ship a .py3. wheel?

I was thinking the same thing. This is the approach used by some other projects, e.g. Django:

https://github.com/django/django/blob/ead37dfb580136cc27dbd487a1f1ad90c9235d15/setup.py#L10-L52

> Some people likely vendor urllib3 source code into their own, having it here prevents that.

If people are vendoring libraries into incompatible environments -- when those environments are made explicit by the library -- at some point, isn't that on the person doing the vendoring?

graingert:
> If people are vendoring libraries into incompatible environments -- when those environments are made explicit by the library -- at some point, isn't that on the person doing the vendoring?

Yes the vendor should also put the exception message in their setup.py (if they also don't trust requires_python)

sethmlarson:
We can put the message in `setup.py` and only ship a Python 3 wheel, then we won't use a custom exception and can use the `sys.exit(1)` route.

pquentin:
I have no idea why I put that in `__init__.py` instead of setup.py 😕
I actually was trying to do what Django does.

Thanks for the reviews! Will be able to fix on Monday, but if someone else wants to prepare a proper pull request in the meantime, please do!

pquentin:
@sethmlarson @sigmavirus24 @jdufresne @graingert Please take another look!

(Also, yes, we're using py3 only wheels)

sigmavirus24 (approved):

setup.py
```diff
@@ -4,9 +4,38 @@
...
+understand the python_requires classifier. Make sure you
+have pip >= 9.0 and setuptools >= 24.2, then try again:
+    $ python -m pip install --upgrade pip setuptools
+    $ python -m pip install urllib3
+This will install the latest version of urllib3 which works on your
+version of Python. If you can't upgrade your pip (or Python), request
+an older version of urllib3:
+    $ python -m pip install "urllib3<2"
+""".format(
+            *(REQUIRED_PYTHON + CURRENT_PYTHON)
```
graingert, line 33: This seems to be a bit of a hack to work around pyupgrade f-strings, it might be easier to hardcode these. Like the pip and setuptools versions

graingert: This constant is only used here, but the same value is hardcoded later in python_requires:

https://github.com/urllib3/urllib3/pull/2070/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R112

setup.py
```diff
@@ -4,9 +4,38 @@
...
+    sys.stderr.write(
+        """
+==========================
+Unsupported Python version
+==========================
+This version of urllib3 requires Python {}.{}, but you're trying to
+install it on Python {}.{}.
+This may be because you are using a version of pip that doesn't
+understand the python_requires classifier. Make sure you
+have pip >= 9.0 and setuptools >= 24.2, then try again:
```
graingert, line 25: Will the `setuptools >= 24.2` python_requires code path ever be hit? It will fail here first

pquentin: It will show up as `Requires-Python: >=3.6, <4` in the wheel METADATA file, which means clients with Python 3.5 or lower will download 1.x automatically

graingert: right they see the wheel, so `setuptools` is not hit here

setup.py
pquentin: pyupgrade or f-strings did not exist when this was written in 2016. I can hardcode 3.6, but `CURRENT_PYTHON` will stay dynamic

sethmlarson (approved):
The current implementation looks good to me, merge when ready.

setup.py
pquentin: `REQUIRED_PYTHON` is used above, so I would not gain much by inlining it. Sorry, given the two other approvals I decided against doing this

PR #2072, opened 2020-11-18
Remove backport for ipaddress

RatanShreshtha:


src/urllib3/packages/ssl_match_hostname/_implementation.py
```diff
@@ -3,17 +3,8 @@
 # Note: This file is under the PSF license as the code comes from the python
 # stdlib.   http://docs.python.org/3/license.html
 
+import ipaddress
 import re
```
sethmlarson, line 7: If `ipaddress` is only used here we should try to remove `ssl_match_hostname` and get a 2-for-1.

RatanShreshtha:
closing in favor of #2076

pquentin:
This should probably reopened if we don't reopen #2076 :)

PR #2076, opened 2020-11-19
Remove backports for ssl match hostname

RatanShreshtha:


src/urllib3/connection.py
```diff
@@ -30,11 +30,12 @@ class BaseSSLError(BaseException):
     SubjectAltNameWarning,
     SystemTimeWarning,
 )
-from .packages.ssl_match_hostname import CertificateError, match_hostname
 from .util import SKIP_HEADER, SKIPPABLE_HEADERS, connection
 from .util.ssl_ import (
+    CertificateError,
     assert_fingerprint,
     create_urllib3_context,
+    match_hostname,
```
pquentin, line 38: You can directly use `from ssl import CertificateError, match_hostname` under `import ssl` above

pquentin: In other words, we no longer need the code in our repo since this code is now in the `ssl` module

sethmlarson:
So I've been poking around with removing commonName support and using SSLContext to do hostname matching and I'm finding there are some things that need to get figured out before we remove this code. I think we should close this PR until that work is figured out.

pquentin:
@sethmlarson On the other hand this backport is the Python 3.5 version and is now unused. It seems that it may be easier to just reintroduce the functions we need when we work on removing commonName support?

By the way, is it https://github.com/urllib3/urllib3/issues/497? Any reason why it's no longer marked as v2?

PR #2077, opened 2020-11-19
Fix banner URL in README

hynek:
Recent changes moved the banner away from the images directory and the README doesn't render it anymore.


pquentin (approved):
Oops. Thanks for the fix!

PR #2078, opened 2020-11-19
Remove six

hramezani:
There is another possibility to remove the dependency to six.
we can move the required function of six like `ensure_str`, `ensure_binary`, ... and then remove `six.py`


src/urllib3/util/request.py
```diff
@@ -80,11 +79,13 @@ def make_headers(
         headers["connection"] = "keep-alive"
 
     if basic_auth:
-        headers["authorization"] = "Basic " + b64encode(b(basic_auth)).decode("utf-8")
+        headers["authorization"] = "Basic " + b64encode(
```
sethmlarson, line 82: `b64encode` can only return ASCII so we can do `b64encode(x.encode("latin-1")).decode()`. Also we can use an `f-string here instead of string concat

test/test_filepost.py
```diff
@@ -2,9 +2,9 @@
 
 from urllib3.fields import RequestField
 from urllib3.filepost import encode_multipart_formdata, iter_fields
-from urllib3.packages.six import b
 
 BOUNDARY = "!! test boundary !!"
+LATIN_ENCODED_BOUNDARY = BOUNDARY.encode("latin-1")
```
sethmlarson, line 7: Instead of `LATIN_ENCODED_BOUNDARY` maybe `BOUNDARY_BYTES` is a more descriptive name?

test/test_response.py
```diff
@@ -801,7 +800,7 @@ def test_invalid_chunk_length(self):
 
         orig_ex = ctx.value.args[1]
         assert isinstance(orig_ex, InvalidChunkLength)
-        assert orig_ex.length == six.b(fp.BAD_LENGTH_LINE)
+        assert orig_ex.length == fp.BAD_LENGTH_LINE.encode("latin-1")
```
sethmlarson, line 803: We should remove it and only do `.encode()` to not confuse future readers into thinking there's anything relevant about the "latin-1" encoding here.

test/with_dummyserver/test_connectionpool.py
```diff
@@ -700,7 +699,7 @@ def test_source_address(self):
                 self.host, self.port, source_address=addr, retries=False
             ) as pool:
                 r = pool.request("GET", "/source_address")
-                assert r.data == b(addr[0])
+                assert r.data == addr[0].encode("latin-1")
```
sethmlarson, line 702: Same as above, remove latin-1 encoding.

sethmlarson (changes requested):
Thanks for this! I left you some comments.

hramezani:
@sethmlarson Thanks for the review. I addressed your comments.

What is your idea about my comment:

> we can move the required function of six like ensure_str, ensure_binary, ... and then remove six.py

If you agree with moving these function out of `six.py` please give me some hint about the file that we need to move these function to it.

sethmlarson:
Definitely, maybe create a new module (`urllib3.util.util`??) since `urllib3.util` is unfortunately public.

hramezani:
> Definitely, maybe create a new module (`urllib3.util.util`??) since `urllib3.util` is unfortunately public.

I will try to make a PR for it.

sethmlarson:
Okay, so this PR is good to go then?

hramezani:
we can merge this one and make another one for the six functions. Or have them all in this patch.
Which one do you prefer @sethmlarson?

sethmlarson:
Let's bring everything into this PR, then we can label it "remove six"

test/test_filepost.py
```diff
@@ -2,9 +2,9 @@
 
 from urllib3.fields import RequestField
 from urllib3.filepost import encode_multipart_formdata, iter_fields
-from urllib3.packages.six import b
 
 BOUNDARY = "!! test boundary !!"
+BOUNDARY_BYTES = BOUNDARY.encode("latin-1")
```
sethmlarson, line 7: Can we remove the latin-1 here as well?

src/urllib3/util/util.py
```diff
@@ -0,0 +1,36 @@
+def ensure_binary(s, encoding="utf-8", errors="strict"):
```
sethmlarson, line 1: Thoughts on using `to_str` and `to_bytes` as both shorter and more descriptive?

The following are optimized such that `to_X()` on type `X` is fast and we only pay the "not isinstance()" tax when it's not `X`. Also calling without any arguments is a lot faster than with any arguments so we should optimize for this case:

```python
from typing import Union, Optional


def to_bytes(x: Union[str, bytes], encoding: Optional[str] = None, errors: Optional[str] = None) -> bytes:
    if isinstance(x, bytes):
        return x
    elif not isinstance(x, str):
        raise TypeError(f"not expecting type {type(x).__name__}")
    elif encoding or errors:
        return x.encode(encoding, errors=errors)
    return x.encode()


def to_str(x: Union[str, bytes], encoding: Optional[str] = None, errors: Optional[str] = None) -> str:
    if isinstance(x, str):
        return x
    elif not isinstance(x, bytes):
        raise TypeError(f"not expecting type {type(x).__name__}")
    elif encoding or errors:
        return x.decode(encoding, errors=errors)
    return x.decode()
```

That way `to_bytes()` will encode as ASCII with errors=strict

src/urllib3/util/url.py
```diff
@@ -210,7 +210,7 @@ def _encode_invalid_chars(component, allowed_chars, encoding="utf-8"):
     if component is None:
         return component
 
-    component = six.ensure_text(component)
+    component = ensure_str(component)
```
sethmlarson, line 213: This call to `to_str()` would need an `encoding="utf-8"` if implemented like below.

src/urllib3/util/url.py
```diff
@@ -396,7 +396,7 @@ def parse_url(url):
             path = None
 
     def ensure_type(x):
-        return x if x is None else six.ensure_text(x)
+        return x if x is None else ensure_str(x)
```
sethmlarson, line 399: Idle thought: I feel like this isn't necessary anymore since we only accept URLs that are `str` for Python 3 so URL components can only end up as `str`.

sethmlarson (commented):
A suggestion on name/implementation, would like your thoughts

src/urllib3/util/util.py
hramezani: @sethmlarson Thanks for the suggestion.
I would suggest using default values for `encoding` and `errors` because with 
```python
    elif encoding or errors:
        return x.decode(encoding, errors=errors)
````
If we pass just `encoding`, the condition will be True and we will face `TypeError` (because the `error` is None)
`TypeError: decode() argument 2 must be str, not None`

Comment (sethmlarson) [reply src/urllib3/util/util.py:1]: Ah yes, perhaps `x.encode(encoding or "ascii", errors=errors or "strict")`. Trying to detect the case when we can simple `x.decode()` since it's >4x faster than calling with any arguments and encoding/decoding is a common operation for us.

File: src/urllib3/util/url.py (hunk 7 above)
Comment (hramezani) [reply src/urllib3/util/url.py:213]: > would need an encoding="utf-8" if implemented like below.

Do you mean `encoding=encoding`? because the `_encode_invalid_chars` accepts an `encoding` param with default value to `"utf-8"`

Comment (sethmlarson) [reply src/urllib3/util/url.py:213]: Hm, that parameter has been doing basically nothing it seems. You can remove the parameter and at the end of the `_encode_invalid_chars()` change `encoded_component.decode(encoding)` to `.decode()` since percent-encoding implies the end result is ASCII.

File: src/urllib3/contrib/securetransport.py [hunk 9]
```diff
@@ -869,7 +867,7 @@ def set_alpn_protocols(self, protocols):
             raise NotImplementedError(
                 "SecureTransport supports ALPN only in macOS 10.12+"
             )
-        self._alpn_protocols = [six.ensure_binary(p) for p in protocols]
+        self._alpn_protocols = [util.util.to_bytes(p) for p in protocols]
```
sethmlarson, line 870: Let's expand the import to `from ..util.util import to_bytes` so it's similar to the rest of the usages

src/urllib3/contrib/pyopenssl.py
```diff
@@ -450,7 +449,7 @@ def load_cert_chain(self, certfile, keyfile=None, password=None):
         self._ctx.use_privatekey_file(keyfile or certfile)
 
     def set_alpn_protocols(self, protocols):
-        protocols = [six.ensure_binary(p) for p in protocols]
+        protocols = [util.util.to_bytes(p) for p in protocols]
```
sethmlarson, line 452: nit: Let's expand the import to `from ..util.util import to_bytes` so it's similar to the rest of the usages

src/urllib3/util/url.py
```diff
@@ -395,17 +395,14 @@ def parse_url(url):
...
             path = None
 
-    def ensure_type(x):
-        return x if x is None else six.ensure_text(x)
-
     return Url(
-        scheme=ensure_type(scheme),
-        auth=ensure_type(auth),
-        host=ensure_type(host),
+        scheme=scheme,
```
sethmlarson, line 399: :tada: This is so great!

src/urllib3/util/util.py
```diff
@@ -0,0 +1,37 @@
+from typing import Optional, Union
+
+
+def to_bytes(
```
sethmlarson, line 4: Since I wrote this code would be good to get another outside review (cc @pquentin @hodbn )

sethmlarson (dismissed):
besides the two nits and the call for reviews I'm happy with this change, goodbye six!

sethmlarson:
Oh also I wonder if there's some coverage loss because the `to_X()` `TypeError` case isn't covered. Maybe can throw in a couple test cases for our new utility methods? It's fine to mark `reraise` as not requiring code coverage since we lifted it from six.

src/urllib3/contrib/securetransport.py
hramezani: I get this error when I import it by `from ..util.util import to_bytes`

`Could not import SecureTransport: ModuleNotFoundError("No module named 'urllib3.packages.util'")`

src/urllib3/contrib/pyopenssl.py
hramezani: doc tests will fail with this change.

Maybe it is related to `from .. import util` that exists at the top of file

hramezani:
@sethmlarson Thanks for the review.

I added two tests to cover `TypeError` in the newly adde to `to_X()` functions.
Also, I ignored coverage for `def reraise` by adding it to the `exclude_lines ` section of `.coveragerc` (not sure that I did it right)

pquentin (changes requested):
Great work! I have a question about `to_bytes`

pquentin:
Also, to_bytes and to_str are not fully covered as you can see here: https://codecov.io/gh/urllib3/urllib3/src/3d79bccbda1486de6945d0a084b782a43c77890e/src/urllib3/util/util.py

Capture d’écran 2020-11-22 à 21 17 36

src/urllib3/util/util.py
```diff
@@ -0,0 +1,37 @@
...
+    x: Union[str, bytes], encoding: Optional[str] = None, errors: Optional[str] = None
+) -> bytes:
+    if isinstance(x, bytes):
+        return x
+    elif not isinstance(x, str):
+        raise TypeError(f"not expecting type {type(x).__name__}")
+    elif encoding or errors:
+        return x.encode(encoding or "ascii", errors=errors or "strict")
+    return x.encode()
```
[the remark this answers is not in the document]
hramezani: Since we want to call `x.encode()`, because it's faster (based on @sethmlarson [comment](https://github.com/urllib3/urllib3/pull/2078#discussion_r527957964). I think we can change it like this:

`return x.encode(encoding or "utf-8", errors=errors or "strict")`

hramezani:
> Also, to_bytes and to_str are not fully covered as you can see here: https://codecov.io/gh/urllib3/urllib3/src/3d79bccbda1486de6945d0a084b782a43c77890e/src/urllib3/util/util.py
> 
> Capture d’écran 2020-11-22 à 21 17 36

I added some tests to complete the coverage.

src/urllib3/util/util.py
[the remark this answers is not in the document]
sethmlarson: Wow, I didn't realize the default encoding for `str.encode()` was utf-8, thought it was ascii all this time. Thanks for bringing this to my attention!

In this case we will need to explicitly say `to_str("ascii")` in all the places I previously thought should be ASCII.

src/urllib3/contrib/pyopenssl.py
sethmlarson: This should be `to_bytes(p, "ascii")`

src/urllib3/contrib/securetransport.py
sethmlarson: `to_bytes(p, "ascii")`

src/urllib3/util/url.py
```diff
@@ -203,14 +203,14 @@ def split_first(s, delims):
...
-def _encode_invalid_chars(component, allowed_chars, encoding="utf-8"):
+def _encode_invalid_chars(component, allowed_chars):
     """Percent-encodes a URI component without reapplying
     onto an already percent-encoded component.
     """
     if component is None:
         return component
 
-    component = six.ensure_text(component)
+    component = to_str(component, encoding="utf-8")
```
sethmlarson, line 213: Can remove the `encoding=` here

src/urllib3/util/url.py
```diff
@@ -286,7 +286,7 @@ def _normalize_host(host, scheme):
                 else:
                     return host.lower()
             elif not IPV4_RE.match(host):
-                return six.ensure_str(
+                return to_str(
```
sethmlarson, line 289: `to_str(x, "ascii")`

sethmlarson (changes requested):
Sorry for the confusion, thank you @pquentin for correcting me. A few little comments left!

sethmlarson (dismissed):
This looks good to me, looking for a final review from @pquentin now :)

test/test_util.py
```diff
@@ -776,6 +777,41 @@ def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, hos
         socket.return_value = Mock()
         create_connection((host, 80))
 
+    @pytest.mark.parametrize(
+        "input,params,expected",
+        (
+            ("test", {}, "test"),  # str input
+            (b"test", {}, "test"),  # bytes input
+            (b"test", {"encoding": "utf-8"}, "test"),  # bytes input with utf-8
+            (b"test", {"encoding": "ascii"}, "test"),  # bytes input with asscii
```
pquentin, line 786:
```suggestion
            (b"test", {"encoding": "ascii"}, "test"),  # bytes input with ascii
```

test/test_util.py
```diff
@@ -776,6 +777,41 @@ def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, hos
...
+        ),
+    )
+    def test_to_str(self, input, params, expected):
+        assert to_str(input, **params) == expected
+
+    def test_to_str_error(self):
+        with pytest.raises(TypeError) as e:
+            to_str(1)
+        assert "not expecting type int" in str(e.value)
```
pquentin, line 795:
```suggestion
        with pytest.raises(TypeError, match="not expecting type int"):
            to_str(1)
```

test/test_util.py
```diff
@@ -776,6 +777,41 @@ def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, hos
...
+            ("test", {"encoding": "ascii"}, b"test"),  # bytes input with asscii
+        ),
+    )
+    def test_to_bytes(self, input, params, expected):
+        assert to_bytes(input, **params) == expected
+
+    def test_to_bytes_error(self):
+        with pytest.raises(TypeError) as e:
+            to_bytes(1)
+        assert "not expecting type int" in str(e.value)
```
pquentin, line 813:
```suggestion
        with pytest.raises(TypeError, match="not expecting type int"):
            to_bytes(1)
```

.coveragerc
```diff
@@ -22,6 +22,7 @@ exclude_lines =
     pass
     import
     raise
+    def reraise
```
pquentin, line 25: This file is for statements that occur multiple times, which is not the case here. And `reraise` is covered anyway, so we can remove that line.

```suggestion
```

src/urllib3/util/util.py
```diff
@@ -0,0 +1,37 @@
...
+def to_bytes(
+    x: Union[str, bytes], encoding: Optional[str] = None, errors: Optional[str] = None
+) -> bytes:
+    if isinstance(x, bytes):
+        return x
+    elif not isinstance(x, str):
+        raise TypeError(f"not expecting type {type(x).__name__}")
+    elif encoding or errors:
+        return x.encode(encoding or "utf-8", errors=errors or "strict")
+    return x.encode()
```
pquentin, line 13: You can remove one branch here:

```suggestion
    return x.encode(encoding or "utf-8", errors=errors or "strict")
```

src/urllib3/util/util.py
```diff
@@ -0,0 +1,37 @@
...
+def to_str(
+    x: Union[str, bytes], encoding: Optional[str] = None, errors: Optional[str] = None
+) -> str:
+    if isinstance(x, str):
+        return x
+    elif not isinstance(x, bytes):
+        raise TypeError(f"not expecting type {type(x).__name__}")
+    elif encoding or errors:
+        return x.decode(encoding or "utf-8", errors=errors or "strict")
+    return x.decode()
```
pquentin, line 25: You can remove one branch here:

```suggestion
    return x.decode(encoding or "utf-8", errors=errors or "strict")
```

test/test_util.py
```diff
@@ -776,6 +777,41 @@ def test_create_connection_with_valid_idna_labels(self, socket, getaddrinfo, hos
...
+
+    @pytest.mark.parametrize(
+        "input,params,expected",
+        (
+            (b"test", {}, b"test"),  # str input
+            ("test", {}, b"test"),  # bytes input
+            ("é", {}, b"\xc3\xa9"),  # bytes input
+            ("test", {"encoding": "utf-8"}, b"test"),  # bytes input with utf-8
+            ("test", {"encoding": "ascii"}, b"test"),  # bytes input with asscii
```
pquentin, line 804:
```suggestion
            ("test", {"encoding": "ascii"}, b"test"),  # bytes input with ascii
```

pquentin (changes requested):
Great work! We're nearly there, I only have cosmetic comments left.

pquentin:
This now looks good to *me*, but contains code that I wrote, so I need review from either or you now!

src/urllib3/util/util.py
hramezani: I think @sethmlarson suggested the branch in function to detect the case that we can call `x.encode()` because it's 4X faster. 
Take a look at this [comment](https://github.com/urllib3/urllib3/pull/2078#discussion_r527972377)

pquentin:
Weird, but I seem to reproduce. I guess a comment would be best. I removed that part from the last commit, please review

sethmlarson (approved):
@pquentin LGTM, merge when ready!

pquentin:
Closing/reopening as we ran into the coverage SQLite race condition again.

pquentin (approved):

pquentin:
Goodbye six 👋 You served us well

hramezani:
@sethmlarson @pquentin thanks for the review and your help.

PR #2080, opened 2020-11-20
Change inner container of HTTPHeaderDict to 'dict'

sethmlarson:
`dict` is more memory efficient compared to `OrderedDict` and still has all the properties and features we want.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -1,3 +1,5 @@
+import sys
+from collections import OrderedDict
 from collections.abc import Mapping, MutableMapping
 
 try:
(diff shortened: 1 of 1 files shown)


hramezani (approved):
LGTM :+1:

src/urllib3/_collections.py
```diff
@@ -12,7 +14,16 @@ def __exit__(self, exc_type, exc_value, traceback):
             pass
 
 
-from collections import OrderedDict
+# Starting in Python 3.7 the 'dict' class is guaranteed to be
+# ordered by insertion. This behavior was an implementation
+# detail in Python 3.6. OrderedDict is implemented in C in
```
pquentin, line 19: And crucially it's also true in PyPy.

Why can't we also use `dict` in Python 3.6 by the way? Is this for other implementations not in our CI that might not have the properties we want?

pquentin:
@hramezani If this looks good to you, squash and merge!

hramezani:
> @hramezani If this looks good to you, squash and merge!

It looks good to me but the merge button is disabled for me.

pquentin (approved):

pquentin:
Hmm, that's something we should fix then. cc @sethmlarson

src/urllib3/_collections.py
sethmlarson: @pquentin Yeah basically! Being conservative :)

sethmlarson:
> It looks good to me but the merge button is disabled for me.

@hramezani That's not right, I'll fix that issue now :)

sethmlarson:
@hramezani Can you look at a different green PR and see if you have the merge button now?

hramezani:
@sethmlarson I couldn't find a green PR for now. But let's check it later we have one :)

PR #2081, opened 2020-11-22
Support passing blocksize to PoolManager in python 3.7+

### LINKED ISSUE #2067: PoolManager does not support argument blocksize
### Subject

urllib3's PoolManager does not support `blocksize`. This seems to be a new argument accepted by Python's standard library's http clients since 3.7 (https://docs.python.org/3.6/library/http.client.html vs https://docs.python.org/3.7/library/http.client.html)

Adding `key_blocksize` to poolmanager.py's `_key_fields` fixes the issue but it probably should only be appended when it is actually supported.

### Environment

OS Linux-5.4.0-47-generic-x86_64-with-glibc2.29
Python 3.8.5
urllib3 1.26.2

### Steps to Reproduce

```
with PoolManager(blocksize = 8192) as pool:
    response = pool.urlopen("GET", "https://www.google.com")
```

### Expected Behavior

Should perform a GET request successfully.

### Actual Behavior

The following exception is thrown

```
  File "report.py", line 13, in <module>
    response = pool.urlopen("GET", "https://www.google.com")
  File "/home/geancarlo/redacted/poc/venv/lib/python3.8/site-packages/urllib3/poolmanager.py", line 364, in urlopen
    conn = self.connection_from_host(u.host, port=u.port, scheme=u.scheme)
  File "/home/geancarlo/redacted/poc/venv/lib/python3.8/site-packages/urllib3/poolmanager.py", line 245, in connection_from_host
  File "/home/geancarlo/redacted/poc/venv/lib/python3.8/site-packages/urllib3/poolmanager.py", line 258, in connection_from_context
  File "/home/geancarlo/redacted/poc/venv/lib/python3.8/site-packages/urllib3/poolmanager.py", line 124, in _default_key_normalizer
    return key_class(**context)
TypeError: __new__() got an unexpected keyword argument 'key_blocksize'
```

geancarlo:
Fix #2067


pquentin (changes requested):
I realize I asked you to write a new test, but reviewing this it now seems better to just add a `blocksize` to `connection_pool_kw` in `test_http_pool_key_fields` and `test_https_pool_key_fields`. This avoids adding two new tests and achieves the same result.

Sorry about that!

src/urllib3/poolmanager.py
```diff
@@ -62,6 +63,11 @@
     "key_server_hostname",  # str
 )
 
+# Version 3.7 introduces the parameter "blocksize" to
```
pquentin, line 66:
```suggestion
# Python 3.7 introduces the parameter "blocksize" to
```

(Writing "Python 3.7" makes it easier to grep.)

test/test_poolmanager.py
```diff
@@ -128,6 +129,29 @@ def test_http_pool_key_fields(self):
...
+        p = PoolManager()
+        conn_pools = [
+            p.connection_from_url("http://example.com/"),
+            p.connection_from_url("http://example.com:8000/"),
+            p.connection_from_url("http://other.example.com/"),
+        ]
+
+        for key, value in connection_pool_kw.items():
+            p.connection_pool_kw[key] = value
+            conn_pools.append(p.connection_from_url("http://example.com/"))
```
hramezani, line 145: We don't need this loop because we just have one key in `connection_pool_kw`. SO:

```suggestion
        p.connection_pool_kw["blocksize"] = 16384
        conn_pools.append(p.connection_from_url("http://example.com/"))
```

Also, we can remove variable `connection_pool_kw` at the beginning of test

test/test_poolmanager.py
```diff
@@ -165,6 +189,34 @@ def test_https_pool_key_fields(self):
...
+        conn_pools = [
+            p.connection_from_url("https://example.com/"),
+            p.connection_from_url("https://example.com:4333/"),
+            p.connection_from_url("https://other.example.com/"),
+        ]
+        # Asking for a connection pool with the same key should give us an
+        # existing pool.
+        dup_pools = []
+
+        for key, value in connection_pool_kw.items():
```
hramezani, line 206: Same as above.

hramezani:
I removed my comments in favor of @pquentin [comment](https://github.com/urllib3/urllib3/pull/2081#pullrequestreview-536081440)

src/urllib3/poolmanager.py
geancarlo: I have updated my fork. Thanks

hramezani (dismissed):
LGTM :+1:  Thanks!

CONTRIBUTORS.txt
```diff
@@ -306,5 +306,8 @@ In chronological order:
 * [Ezzeri Esa] <https://github.com/savarin>
   * Ports and extends on types from typeshed
 
+* [Geancarlo Rocha] <https://github.com/geancarlo>
+  * Support passing blocksize to PoolManager in python 3.7+
```
sethmlarson, line 310: nit: Capitalize "Python"

src/urllib3/poolmanager.py
```diff
@@ -62,6 +63,11 @@
     "key_server_hostname",  # str
 )
 
+# Python 3.7 introduces the parameter "blocksize" to
+# http.client.HTTPConnection's and http.client.HTTPSConnection's initializers
+if sys.version_info >= (3, 7):
+    _key_fields += ("key_blocksize",)  # int
```
sethmlarson, line 69: Within the `_default_key_normalizer` we should set the default to be `16384` instead of `None` so that users get a more desirable default.

Can we also get a test case that the normalizer sets `blocksize` properly given inputs like no `blocksize` set, blocksize=8192, blocksize=16384

sethmlarson (changes requested):
Thanks for this, a couple of comments.

src/urllib3/poolmanager.py
geancarlo: I'm passing `blocksize = 16384` to the pool class in _new_pool if it is not set in the `request_context`, but I'm afraid this may be too intrusive. What do you think?

src/urllib3/poolmanager.py
```diff
@@ -116,6 +123,10 @@ def _default_key_normalizer(key_class, request_context):
         if field not in context:
             context[field] = None
 
+    # Default key blocksize to 16384 if missing from the context
+    if _CONN_CLS_ACCEPTS_BLOCKSIZE and context.get("key_blocksize") is None:
```
PleasantMachine9, line 127: If `context` is a dict, maybe using `dict.setdefault` would be better? 

```
if _CONN_CLS_ACCEPTS_BLOCKSIZE:
    context.setdefault("key_blocksize", 16384)
```

would be the same as 
```
if _CONN_CLS_ACCEPTS_BLOCKSIZE and "key_blocksize" not in context:
    context["key_blocksize"] = 16384
```

This won't overwrite an explicitly set `None` of course, so if that's desired then current behavior can stay.

src/urllib3/poolmanager.py
```diff
@@ -193,6 +204,12 @@ def _new_pool(self, scheme, host, port, request_context=None):
         pool_cls = self.pool_classes_by_scheme[scheme]
         if request_context is None:
             request_context = self.connection_pool_kw.copy()
+        elif _CONN_CLS_ACCEPTS_BLOCKSIZE:
+            request_context = request_context.copy()
+
+        # Default blocksize to 16384 if missing from the context
+        if _CONN_CLS_ACCEPTS_BLOCKSIZE and request_context.get("blocksize") is None:
```
PleasantMachine9, line 211: Same as  Line 127

test/test_poolmanager.py
```diff
@@ -366,3 +373,42 @@ def test_pool_manager_no_url_absolute_form(self):
...
+        pool_blocksize_16384 = p.connection_from_url(
+            "http://example.com", {"blocksize": 16384}
+        )
+        assert pool_blocksize_16384.conn_kw["blocksize"] == 16384
+
+        pool_blocksize_not_set = p.connection_from_url("http://example.com")
+        assert pool_blocksize_not_set == pool_blocksize_16384
+
+        pool_blocksize_set_as_none = p.connection_from_url(
+            "http://example.com", {"blocksize": None}
```
PleasantMachine9, line 391: It is interesting what should happen here. It could be argued that explicit `None` should not be treated the same as an omission (->setting default), but I'm not sure what's better design.

sethmlarson:
I think there's some merging work to be done now that https://github.com/urllib3/urllib3/pull/2162 has been merged, could you take a look @geancarlo?

Issue #2065, opened 2020-11-16
Tracking breaking changes for v2.0


pquentin:
We need to record all noteworthy 2.0 changes as news fragments. Here's the full list: https://github.com/urllib3/urllib3/compare/1.26.x...main. I've listed all relevant changes below. Sorry, this is messy, I sometimes link to the merged commit, and sometimes to the PR. The merged commit is usually a better starting point as Seth usually rewords before merging.

- [x] https://github.com/urllib3/urllib3/commit/fbe9620784b130dcdccf43cd3105940152c4d762 Remove support for Python 2 and 3.5 
- [x] https://github.com/urllib3/urllib3/commit/6af53ac41757399607382bf10a92b288fdf4951d Remove support for Google App Engine Standard
- [x] https://github.com/urllib3/urllib3/commit/b6061f09741082bed3d8b7db10c25bf1ae683a43 Remove strict parameter (but tolerate it in https://github.com/urllib3/urllib3/commit/186e7f91c058d8a04e3b894d5580001e09666ee3)
- [x] https://github.com/urllib3/urllib3/commit/4ff35efb97f68c83b588c0e31d612178776be3d7 Change inner container of HTTPHeaderDict to 'dict' 
- [x] https://github.com/urllib3/urllib3/commit/c67c0949e9c91c7621ea718a7f297ecac7c3b79e Remove deprecated Retry options
- [x] https://github.com/urllib3/urllib3/commit/d39d6a6d01083280554fbe1a240bc85ddd88fdd3 Add BaseHTTPResponse base class 
- [x] https://github.com/urllib3/urllib3/commit/cad4850614d86b4db56a785bc9f984eed342f8be Change return type of HTTPResponse.getheaders() to match CPython
- [x] https://github.com/urllib3/urllib3/commit/c91545052595db80eac46c44d4319a90cc5370a3 Switch to Brotli C bindings for CPython, brotlicffi for non-CPython (#2099)
- [x] https://github.com/urllib3/urllib3/commit/4fadb28df00d0446a1e6b4254050c9ac03958e75 Attach errno to SocketError when raised from OpenSSL.SSL.SysCallError
- [x] https://github.com/urllib3/urllib3/commit/102352a0807c16d8ee3569cfd38be1ca2bfdd7c9 Replace deprecated socket.error with OSError
[...]

PR #2083, opened 2020-11-23
Add BaseHTTPResponse base class

### LINKED ISSUE #1543: HTTPResponse.getheaders() is not backwards compatible
The `getheaders()` method of the `HTTPResponse` class returns the dict-like `HTTPHeaderDict`, while the httplib version of that method returns a list of (header, value) tuples. This means that code iterating over the response as in `for header, value in response.getheaders()` will not work with urllib3 response objects unless `.items()` is called on the returned `HTTPHeaderDict` object. A comment near the `getheaders()` method indicates it's supposed to be backwards compatible.

### LINKED ISSUE #1986: [v2] Create BaseHTTPResponse API
- Create a response base class that still inherits from `io.IOBase` but has no mentions of httplib
- There may be some other APIs we can remove from the constructor like `msg`?
- Make `HTTPResponse` a subclass of `BaseHTTPResponse`
- Ensure that type hints for `.request()` of `BaseHTTPConnection` annotate as `BaseHTTPResponse` instead of `HTTPResponse`

sethmlarson:
- Adds the `BaseHTTPResponse` base type and moved many agnostic functions to the class
- Adds the `ContentDecoder` base type
- Changes the return type of `HTTPResponse.getheaders()` to a list of key-value pairs in line with `http.client.HTTPResponse`

Closes #1986 
Closes #1543

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import typing
 import zlib
 from contextlib import contextmanager
 from socket import error as SocketError
@@ -29,14 +30,19 @@ from .util.response import is_fp_closed, is_response_to_head
 log = logging.getLogger(__name__)
 
 
-class DeflateDecoder:
+class ContentDecoder:
+    def decompress(self, data: bytes) -> bytes:
+        raise NotImplementedError()
+
+    def flush(self) -> bytes:
+        raise NotImplementedError()
+
+
+class DeflateDecoder(ContentDecoder):
     def __init__(self):
         self._first_try = True
         self._data = b""
         self._obj = zlib.decompressobj()
-
-    def __getattr__(self, name):
-        return getattr(self._obj, name)
 
     def decompress(self, data):
         if not data:
@@ -60,6 +66,9 @@ class DeflateDecoder:
             finally:
                 self._data = None
 
+    def flush(self) -> bytes:
+        return self._obj.flush()
+
 
 class GzipDecoderState:
 
@@ -68,13 +77,10 @@ class GzipDecoderState:
     SWALLOW_DATA = 2
 
 
-class GzipDecoder:
+class GzipDecoder(ContentDecoder):
     def __init__(self):
         self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
         self._state = GzipDecoderState.FIRST_MEMBER
-
-    def __getattr__(self, name):
-        return getattr(self._obj, name)
 
     def decompress(self, data):
         ret = bytearray()
@@ -97,10 +103,13 @@ class GzipDecoder:
             self._state = GzipDecoderState.OTHER_MEMBERS
             self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
 
+    def flush(self) -> bytes:
+        return self._obj.flush()
+
 
 if brotli is not None:
 
-    class BrotliDecoder:
+    class BrotliDecoder(ContentDecoder):
         # Supports both 'brotlipy' and 'Brotli' packages
         # since they share an import name. The top branches
         # are for 'brotlipy' and bottom branches for 'Brotli'
@@ -117,7 +126,7 @@ if brotli is not None:
             return b""
 
 
-class MultiDecoder:
+class MultiDecoder(ContentDecoder):
     """
     From RFC7231:
         If one or more encodings have been applied to a representation, the
@@ -138,7 +147,7 @@ class MultiDecoder:
         return data
 
 
-def _get_decoder(mode):
+def _get_decoder(mode: str) -> ContentDecoder:
     if "," in mode:
         return MultiDecoder(mode)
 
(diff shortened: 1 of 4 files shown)


pquentin:
@sethmlarson "Ubuntu 3.[6789]" are required but it looks like the name of the jobs that ran is "CI / Ubuntu 3.[6789]". Can you please fix that?

src/urllib3/response.py
```diff
@@ -321,8 +321,8 @@ def isclosed(self):
         return self.closed
 
     # Compatibility methods for http.client.HTTPResponse
-    def getheaders(self) -> typing.MutableMapping[str, str]:
-        return self.headers
+    def getheaders(self) -> typing.List[typing.Tuple[str, str]]:
+        return list(self.headers.items())
```
pquentin, line 325: Mentioned that breaking change in https://github.com/urllib3/urllib3/issues/2065#issuecomment-731951335 for lack of a better place. We should probably start using towncrier :)

src/urllib3/response.py
```diff
@@ -60,6 +66,9 @@ def decompress(self, data):
             finally:
                 self._data = None
 
+    def flush(self) -> bytes:
+        return self._obj.flush()
```
pquentin, line 70: Wow, this is great! Much more explicit and actually less characters. The magic `__getattr__` always confused me

src/urllib3/response.py
```diff
@@ -151,7 +160,184 @@ def _get_decoder(mode):
...
+        DECODER_ERROR_CLASSES += (brotli.error,)
+
+    def __init__(
+        self,
+        headers=None,
+        status=0,
+        version=0,
+        reason="",
+        decode_content=True,
+        enforce_content_length=False,
```
pquentin, line 180: Should we avoid mentioning the default values here? It sounds likely we'll update one class and not the other one in the future.

Also, should those be kwargs only?

src/urllib3/response.py
```diff
@@ -208,20 +389,19 @@ def __init__(
...
-        if isinstance(headers, HTTPHeaderDict):
-            self.headers = headers
-        else:
-            self.headers = HTTPHeaderDict(headers)
-        self.status = status
-        self.version = version
-        self.reason = reason
-        self.decode_content = decode_content
         self.retries = retries
         self.enforce_content_length = enforce_content_length
```
pquentin, line 402: This line can be removed as it's already in `BaseHTTPResposne`

src/urllib3/response.py
```diff
@@ -298,9 +459,6 @@ def data(self):
     def connection(self):
         return self._connection
 
-    def isclosed(self):
-        return is_fp_closed(self._fp)
```
pquentin, line 302: Did you intend to stop using `is_fp_closed` here? `close()` does not even set `self.closed` to True

pquentin (changes requested):
Great work, thank you!

src/urllib3/response.py
sethmlarson: I meant to revert this, thanks for catching!

src/urllib3/response.py
sethmlarson: Oh yeah, we can do that now can't we :) I'll remove the defaults and add types+keyword-only.

src/urllib3/response.py
sethmlarson: Thanks for adding that. I'm open to using towncrier for tracking changes, can create an issue for this work.

src/urllib3/response.py
pquentin: I'm still seeing that extra line, is that expected?

pquentin (commented):
Tried reviewing again from my phone

src/urllib3/response.py
sigmavirus24: I think something that might be valuable here is creating annotation aliases. For example `HeaderPair = typing.Tuple[str, str]` it makes the annotations clearer when reading the code and when reusing that `Tupe[str, str]` pattern for Headers as it conveys intent

src/urllib3/response.py
sethmlarson: @pquentin I think I moved the `enforce_content_length` back into `HTTPResponse`, are things okay now with the diff?

pquentin: Ahh right I hadn't realized that.

To be honest this is why I don't like rebases during pull requests. Since GitHub doesn't have a way to show what changed during two pushes (unlike GitLab and tools like Gerrit), rebasing is complicating the work of the reviewer who has to read the whole diff again and can easily miss things like this.

I'd suggest just adding commits, which is fine since we squash commits on merge anyway.

pquentin (approved):
This looks good to me, thanks!

src/urllib3/response.py
sethmlarson: Sorry for the confusion, will do in the future.

PR #2086, opened 2020-11-24
Remove deprecated Retry options

### LINKED ISSUE #2058: [v2] Remove deprecated Retry options
Remove the options deprecated in https://github.com/urllib3/urllib3/pull/2000

pquentin:
Closes #2058 

Thank you @sethmlarson for making this is easy to do!


pquentin:
We hit https://github.com/nedbat/coveragepy/issues/1010 again, closing/reopening

sethmlarson (approved):
LGTM, thanks for removing all the options and mentions!

src/urllib3/util/retry.py
```diff
@@ -26,49 +25,7 @@
...
-
-
-class _RetryMeta(type):
-    @property
-    def DEFAULT_METHOD_WHITELIST(cls):
-        warnings.warn(
-            "Using 'Retry.DEFAULT_METHOD_WHITELIST' is deprecated and "
-            "will be removed in v2.0. Use 'Retry.DEFAULT_METHODS_ALLOWED' instead",
-            DeprecationWarning,
-        )
```
andersk, line 40: FYI, this deprecation warning hadn’t been visible, because it was emitted without a [`stacklevel`](https://docs.python.org/3/library/warnings.html#warnings.warn) and therefore [filtered by default](https://docs.python.org/3/library/warnings.html#default-warning-filter).

pquentin: @andersk I'm sorry that you haven't seen the warning, but I don't think stacklevel has anything to do about it? It just changes the way the stack trace is displayed

andersk: It changes where Python thinks the warning originated, which changes whether it will be filtered. For example, this only shows a deprecation warning for `g`:

```console
$ cat a.py
import b
b.f()
b.g()

$ cat b.py
import warnings
def f():
    warnings.warn("f", DeprecationWarning)
def g():
    warnings.warn("g", DeprecationWarning, stacklevel=2)

$ python3 a.py
/tmp/a.py:3: DeprecationWarning: g
  b.g()
```

pquentin: Oh, wow, thanks for taking the time to educate us, I had not realized "defeating the purpose" meant "you won't see it". Will keep this in mind for any future warnings we add!

Sorry about that

PR #2087, opened 2020-11-24
Replace codecs.open() with builtin open()

jdufresne:
Since commit 494213c96915c6e420d6cce9c59fa99a6b621f33, this code will no
longer be executed by Python 2.

This follows the work from commit
fbe9620784b130dcdccf43cd3105940152c4d762.

In Python 3, the feature set between these two functions is the same,
might as well use the builtin version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # This file is protected via CODEOWNERS
 
-import codecs
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


jdufresne:
This is reopening #2068 now that #2070 has landed.

pquentin (approved):
Thanks!

sethmlarson (approved):
LGTM

PR #2088, opened 2020-11-24
Streaming file uploads

### LINKED ISSUE #624: Twice memory usage in `encode_multipart_formdata`

pquentin:
Closes #624 

This is nothing more than #1030 rebased on top of the current master. I havan't checked memory usage yet, but would be happy to get a review. I'm not really satisified with the encode() function, is it possible to avoid it now that we're Python 3 only?


sigmavirus24:
Is there a reason not to use the streaming encode from the toolbelt and adapt it here to avoid the requests-isms (namely importing urllib3 from it's packages for the things in this very module)?

Also a chunk_size of `8192` is extremely slow and small for most people. http.client finally allowed for users to configure that in 3.9. As it stands this looks like an iterator which (iirc) we treated as chunked TE instead of a file with a content-length (and most folks want a content-length, not chunked TE).

sethmlarson:
@sigmavirus24 requests-toolbelt is Apache-2.0 licensed compared to our MIT, this presents a problem?

sigmavirus24:
I wrote all the code and am happy to contribute it here under MIT. Also I give y'all permission to just copy it under MIT

pquentin:
@sigmavirus24 Are you referring to https://github.com/requests/toolbelt/blob/master/requests_toolbelt/streaming_iterator.py? I'd be happy to include that instead. The only limitation that I see compared to this pull request is that you need to know the size in advance, but that's probably not a huge deal.

sigmavirus24:
I was referring to https://github.com/requests/toolbelt/blob/master/requests_toolbelt/multipart/encoder.py

PR #2089, opened 2020-11-24
Remove backport makefile.

hramezani:
@sethmlarson @pquentin Here is the PR for removing backport makefile


hramezani:
I think pytest hangs on some tests on `test/contrib/test_pyopenssl.py` . This happens sometime in my machine with this path

pquentin:
@hramezani The issue was subtle and took me some time to understand: we don't want to call makefile() on `self.socket` but on the actual WrappedSocket!

(I now need to fix the way I import socket.makefile because it confuses Sphinx)

src/urllib3/contrib/pyopenssl.py
```diff
@@ -386,8 +386,10 @@ def _drop(self):
         else:
             self._makefile_refs -= 1
 
-
-WrappedSocket.makefile = backport_makefile
+    def makefile(self, mode="r", buffering=None, *args, **kwargs):
+        text = socket_makefile(mode, buffering, *args, **kwargs)
```
hramezani, line 390: Shouldn't we pass `self` as the first arg here as well? I can see that you did it for `securetransport.py`

hramezani:
Thanks @pquentin for checking and preparing the solution

hramezani:
> @hramezani The issue was subtle and took me some time to understand: we don't want to call makefile() on `self.socket` but on the actual WrappedSocket!
> 
> (I now need to fix the way I import socket.makefile because it confuses Sphinx)

I just realized that it seems the `test_pyopenssl.py` skkipped by your change.
```
SKIPPED [260] test/contrib/test_pyopenssl.py:21: Could not import PyOpenSSL: ModuleNotFoundError("No module named 'socket.socket'; 'socket' is not a package")
```

pquentin:
Uh, right, this is why my fix didn't work, thanks for noticing that! This explain why trying to fix the readthedocs build made the tests fail. Back to the drawing board :)

hramezani:
Close in favor of https://github.com/urllib3/urllib3/pull/2097

PR #2094, opened 2020-11-26
Move type hints from .pyi files to source code itself

RatanShreshtha:
Since we are dropping support for python2 we can remove .pyi files for type hints and put type hints in the files directly


pquentin:
@RatanShreshtha Thanks for tackling this! However this pull request should also add the hints to the .py files. Currently it only removes them from the .pyi ones.

RatanShreshtha:
I am going to open another PR for that

pquentin:
I'm all for small pull requests but just removing those files is a step backwards, not forward, and having two PRs will make the review more difficult.

RatanShreshtha:
I am updating the PR with type hints

pquentin:
Also I guess adding the type hints in urllib3 require to first fix the mypy errors in the source itself?

noxfile.py
```diff
@@ -7,17 +7,7 @@
...
-    "src/urllib3/exceptions.py",
-    "src/urllib3/fields.py",
-    "src/urllib3/filepost.py",
-    "src/urllib3/packages/__init__.py",
-    "src/urllib3/packages/ssl_match_hostname/__init__.py",
-    "src/urllib3/packages/ssl_match_hostname/_implementation.py",
-    "src/urllib3/util/queue.py",
-    "src/urllib3/util/url.py",
-}
+TYPED_FILES = {}
```
pquentin, line 10: Why did you remove that? This means that we're ignoring all mypy errors

yezz123 (approved):

pquentin:
Closing in favor of pull requests like #2115, #2139 and #2148. They add types step by step. Feel free to do the same!

PR #2099, opened 2020-12-02
Switch to Brotli C bindings for CPython, brotlicffi for non-CPython

sethmlarson:
Closes #1620

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -334,7 +334,8 @@ Brotli Encoding
 
 Brotli is a compression algorithm created by Google with better compression
 than gzip and deflate and is supported by urllib3 if the
-`brotlipy <https://github.com/python-hyper/brotlipy>`_ package is installed.
+`Brotli <https://pypi.org/Brotli>`_ package or
+`brotlicffi <https://github.com/python-hyper/brotlicffi>`_ package is installed.
 You may also request the package be installed via the ``urllib3[brotli]`` extra:
 
 .. code-block:: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,10 +51,6 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: test-pypy
-          - python-version: 3.7
-            os: ubuntu-latest
-            experimental: false
-            nox-session: google_brotli
           - python-version: 2.7
             os: ubuntu-latest
             experimental: false
--- a/noxfile.py
+++ b/noxfile.py
@@ -83,15 +83,6 @@ def unsupported_python2(session):
     assert "Unsupported Python version" in process.stderr
 
 
-@nox.session(python=["3"])
-def google_brotli(session):
-    # https://pypi.org/project/Brotli/ is the Google version of brotli, so
-    # install it separately and don't install our brotli extra (which installs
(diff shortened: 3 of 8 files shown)


docs/advanced-usage.rst
```diff
@@ -334,7 +334,8 @@ Brotli Encoding
 
 Brotli is a compression algorithm created by Google with better compression
 than gzip and deflate and is supported by urllib3 if the
-`brotlipy <https://github.com/python-hyper/brotlipy>`_ package is installed.
+`Brotli <https://pypi.org/Brotli>`_ package or
+`brotlicffi <https://github.com/python-hyper/brotlicffi>`_ package is installed.
```
pquentin, line 338: Should this read "are installed"? Genuinely not sure as non-native English speaker

pquentin (changes requested):
Looks great! pypy `make_headers` failed because of that change though. I'm not sure what got installed as nox hides the `pip install` log

docs/advanced-usage.rst
```diff
@@ -334,7 +334,8 @@ Brotli Encoding
 
 Brotli is a compression algorithm created by Google with better compression
 than gzip and deflate and is supported by urllib3 if the
-`brotlipy <https://github.com/python-hyper/brotlipy>`_ package is installed.
+`Brotli <https://pypi.org/Brotli>`_ package or
```
mgorny, line 337: This is 404.

mgorny:
That said, I actually don't like the idea of conditionally depending on two disjoint implementations. This increases the maintenance burden on downstreams, and means that things can go out of sync easily (and therefore become broken on some implementations but not others).

docs/advanced-usage.rst
sethmlarson: Good catch!

sethmlarson:
@mgory Unfortunately this is the situation we've been in for a long time *implicitly* due to the namespace conflict between brotli and brotlipy. Felt it was better to make explicit instead.

Plus performance wins on CPython and means that CPython users will get the latest and greatest Brotli sooner.

This situation isn't unique to us either, celery and fonttools both do this split requirement and it seems to have worked fine for them. Using only the basic Decompressor API hopefully won't lead to breakages :)

mgorny:
But then, is there any reason to prefer brotlicffi on PyPy? FWICS the Python bindings from brotli build just fine on PyPy3.

I personally would prefer to use brotlicffi unconditionally everywhere because it supports linking against shared libbrotli now which is a major improvement over building multiple statically linked copies of it, and having to rebuild them every time upstream makes a release.

sethmlarson:
@mgorny Good question! I haven't measured the performance differences between `Brotli`/`brotlicffi` on PyPy3.

mgorny:
Anyway, I don't really mind as long as I can use brotlicffi without having to patch the package to support it.

docs/advanced-usage.rst
shazow: I think the current phrasing works, or could change packages to plural and then "are installed" would work. Either way. :)

shazow (approved):

sethmlarson:
Thanks all for reviews, I believe I fixed the brotli test on PyPy now :) Will merge once I get all green.

pquentin (approved):

PR #2102, opened 2020-12-08
Update find_unused_port to Python 3.9 version

pquentin:
This can only be more robust than before, and is the opportunity to remove one more "Python 2" mention.


sethmlarson:
@pquentin Thanks for this, it looks like 3.8 failed multiple tests with address in use: https://github.com/urllib3/urllib3/pull/2102/checks?check_run_id=1515711820?

sigmavirus24 (approved):
:+1: (once CI passes of course)

pquentin:
Right, the CI issue is server-side and unrelated

Issue #2104, opened 2020-12-09
Better error message when a proxy is provided without a scheme


sethmlarson:
Current message is `Not supported proxy scheme {scheme}`, instead we could have a better message when `scheme=None` (ie `HTTP_PROXY=127.0.0.1`. Message should recommend using `http://<proxy host>` or `https://<proxy host>`.

References:
- https://github.com/pypa/pip/issues/9216

PR #2106, opened 2020-12-09
Improve error message when a proxy is provided without a scheme.

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/2104

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -285,7 +285,14 @@ class ProxySchemeUnknown(AssertionError, URLSchemeUnknown):
     # TODO(t-8ch): Stop inheriting from AssertionError in v2.0.
 
     def __init__(self, scheme):
-        message = f"Not supported proxy scheme {scheme}"
+        # 'localhost' is here because our URL parser parses
+        # localhost:8080 -> scheme=localhost, remove if we fix this.
+        if scheme == "localhost":
+            scheme = None
+        if scheme is None:
+            message = "Proxy had no scheme, should start with http:// or https://"
+        else:
+            message = f"Proxy URL had unsupported scheme {scheme}, should use http:// or https://"
         super().__init__(message)
 
 
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -290,7 +290,7 @@ class ProxySchemeUnknown(AssertionError, URLSchemeUnknown):
         if scheme == "localhost":
             scheme = None
         if scheme is None:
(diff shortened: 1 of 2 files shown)


sethmlarson:
So i have a couple ideas! If given a scheme of None or "localhost" (!!!) we should give a generic "Proxy had no scheme, should start with http:// or https://" and if there's any other scheme we say "Proxy URL had unsupported scheme X, should use http:// or https://"

Now why "localhost"? Because if given a url like `localhost:8080` our URL parser thinks that's scheme=localhost!

src/urllib3/exceptions.py
```diff
@@ -285,7 +285,10 @@ class ProxySchemeUnknown(AssertionError, URLSchemeUnknown):
     # TODO(t-8ch): Stop inheriting from AssertionError in v2.0.
 
     def __init__(self, scheme):
-        message = f"Not supported proxy scheme {scheme}"
+        if scheme is None or scheme == "localhost":
```
sethmlarson, line 288: idea: add a branch like this so if we sometime fix our URL parser we know we can remove this hacky work-around. Also a comment about why `localhost` is being compared to a scheme might be warranted

```python
# 'localhost' is here because our URL parser parses localhost:8080 -> scheme=localhost, remove if we fix this
if scheme == "localhost":
    scheme = None
if scheme is None:
    message = ...
```

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -450,6 +451,24 @@ def test_scheme_host_case_insensitive(self):
...
+            ),
+            (
+                "ftp://google.com",
+                "Proxy URL had unsupported scheme ftp, should use http:// or https://",
+            ),
+        ],
+    )
+    def test_invalid_schema(self, url, error_msg):
+        with pytest.raises(ProxySchemeUnknown, match=error_msg):
+            proxy_from_url(url)
```
sethmlarson, line 470: Great idea here with using `proxy_from_url` here :+1:

src/urllib3/exceptions.py
hramezani: Thanks @sethmlarson for pointing. I changed it to your suggested code block.

sethmlarson (dismissed):
Made one change for `Proxy` -> `Proxy URL` in the no scheme message, otherwise LGTM

sethmlarson (approved):

sethmlarson:
Thanks @hramezani!

hramezani:
@sethmlarson Yes, I think. That would be great!

PR #2112, opened 2020-12-16
.github/FUNDING.yml: Add Github Sponsors

shazow:
I thiiiink this is correct.

Also lmk if you have thoughts on ordering.


sethmlarson (approved):
LGTM

PR #2113, opened 2020-12-16
Remove the fallback on commonName in match hostname function.

hramezani:


sethmlarson:
You can remove the `SubjectAltName` warning section within `urllib3.connection` as well, since now we'll be raising errors.

We'll keep the warning definition around for now though (perhaps with a deprecation?)

hramezani:
@sethmlarson Thanks for the review. I removed the warning and added a test

sethmlarson (approved):
LGTM, thank you!

PR #2114, opened 2020-12-17
Restore 3.10-dev CI

pquentin:
See https://github.com/urllib3/urllib3/pull/1922. CPython finally decided that there was nothing to fix for Python 3.10, and that the fix should be in packaging/wheel. Maybe there will be something to fix in Python 31.0! Anyway, we can install cryptography again on 3.10.

Since then urllib3 switched from Travis CI to GitHub Actions, so restoring the 3.10 build on Ubuntu implies using the deadsnakes workaround. It actually works really well.

I also sneaked in a change to only install Python 3 for the `unsupported_python2` test and decided to let GitHub Actions choose the specific Python version it wants to install.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,10 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: unsupported_python2
+          - python-version: 3.10-dev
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Yay deadsnakes, LGTM

Issue #1985, opened 2020-09-25
[v2] Create a BaseHTTPConnection APIs


sethmlarson:
- API should only implement methods like `connect`, `request`, `close`, etc that other urllib3 components use.
- Create a `BaseHTTPSConnection` API as well with `set_cert(...)` and extra kwargs in `__init__`
- Have `urllib3.HTTPConnection` and `HTTPSConnection` inherit from `BaseHTTPConnection` and `BaseHTTPSConnection`

#### Sub-issues

- https://github.com/urllib3/urllib3/issues/2648

sethmlarson:
Went through how `HTTP[S]ConnectionPool` uses `HTTP[S]Connection` and found the following methods and parameters in use:

```python
import socket
from urllib3.util import Timeout
from urllib3.response import BaseHTTPResponse

class BaseHTTPConnection:
    http_version: str
    sock: socket.socket
    auto_open: bool  # Whether the HTTPConnection was opened by proxy?
    timeout: Timeout
    is_verified: bool

    def connect(self) -> None: ...
    def request(self, method, target, headers, body) -> None: ...
    def request_chunked(self, method, target, headers, body) -> None: ...
    def getresponse(self) -> BaseHTTPResponse: ...
    def close(self) -> None: ...


class BaseHTTPSConnection(BaseHTTPConnection):
    ssl_version: int
    def set_cert(self) -> None: ...

```

Also was wondering if we can move the conn.sock timeout section before getresponse() into HTTPConnection?

```python

        # >>> Can all this block go into HTTPConnection.getresponse()?
        if conn.sock:
            # In Python 3 socket.py will catch EAGAIN and return None when you
            # try and read into the file pointer created by http.client, which
            # instead raises a BadStatusLine exception. Instead of catching
            # the exception and assuming all BadStatusLine exceptions are read
            # timeouts, check for a zero timeout before making the request.
            if read_timeout == 0:
                raise ReadTimeoutError(
                    self, url, f"Read timed out. (read timeout={read_timeout})"
                )
            if read_timeout is Timeout.DEFAULT_TIMEOUT:
                conn.sock.settimeout(socket.getdefaulttimeout())
            else:  # None or a value
                conn.sock.settimeout(read_timeout)

        # Receive the response from the server
        try:
            httplib_response = conn.getresponse()
        # ...
```

sethmlarson:
[...]

PR #2115, opened 2020-12-18
Inline typechecking for exceptions and add typechecking for collections

haikuginger:
This is a first step towards moving towards inline types for all of `urllib3`. For now, this just fixes and extends the existing types for the `exceptions` module, but adds typing for the `_collections` and `utils.response` modules.

Note that because not all of our codebase is covered by typechecking, we can be pretty sure that these types are internally consistent, but *not* that they're correct. We'll need a lot more type coverage before we can be sure of that. For now, existing tests pass with minimal modifications, which is possibly the best we can do.


src/urllib3/exceptions.py
```diff
@@ -15,26 +22,33 @@ class HTTPWarning(Warning):
     pass
 
 
+ReduceResult = Tuple[Callable[..., object], Tuple[object, ...]]
```
haikuginger, line 25: This isn't strictly complete, as technically-speaking, the callable must have the same number of arguments as the tuple has members, but Python's typing doesn't have a way to describe that relationship.

src/urllib3/response.py
```diff
@@ -460,7 +460,7 @@ def connection(self):
     def isclosed(self):
         return is_fp_closed(self._fp)
 
-    def tell(self):
+    def tell(self) -> int:
```
haikuginger, line 463: Because this method is called from a typed context, we had to type it, although we haven't addressed the rest of the module yet.

src/urllib3/util/response.py
```diff
@@ -53,43 +53,40 @@ def assert_header_parsing(headers):
...
-        # - MultipartInvariantViolationDefect:
-        #     A message claimed to be a multipart but no subparts were found.
-        defects = [
-            defect
-            for defect in defects
-            if not isinstance(
-                defect, (StartBoundaryNotFoundDefect, MultipartInvariantViolationDefect)
-            )
-        ]
+
```
haikuginger, line 57: I figured since we're going to start protecting this with typechecking we could remove the runtime check on if attributes exist. Can revert though.

src/urllib3/util/response.py
```diff
@@ -98,4 +95,5 @@ def is_response_to_head(response):
         used 'HEAD' as a method.
     """
     # FIXME: Can we do this somehow without accessing private httplib _method?
-    return response._method.upper() == "HEAD"
+    method_str = response._method  # type: str  # type: ignore
```
haikuginger, line 98: Ignore that the stubs don't know about this attribute and also cast it as a string.

src/urllib3/_collections.py
```diff
@@ -1,17 +1,44 @@
...
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+    overload,
+)
+
+if TYPE_CHECKING:
```
haikuginger, line 22: We assume that we can import `RLock` if we're in type-checking mode, since `threading` is apt to not exist only in environments where we're unlikely to be doing typechecking.

src/urllib3/_collections.py
```diff
@@ -28,10 +55,33 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+]
 
-class RecentlyUsedContainer(MutableMapping):
+
+def ensure_can_construct_http_header_dict(
+    potential: object,
+) -> Optional[ValidHttpHeaderSource]:
+    if isinstance(potential, HTTPHeaderDict):
+        return potential
+    elif isinstance(potential, Mapping):
```
haikuginger, line 74: If we have a `Mapping` or an `Iterable`, we pretty much assume that the library passing it to us knew what they were doing and made it the right shape.

src/urllib3/compat/rlock.py
```diff
@@ -0,0 +1,12 @@
+"""
+In this module, we shim out a context manager to be used as a compatibility
+layer if the system `threading` module doesn't have a real `RLock` available.
+"""
+
+
+class RLock:
```
jdufresne, line 7: This could be replaced with `nullcontext` when dropping the project is Python3.7+. Perhaps a TODO comment could say as much.

https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext

src/urllib3/_collections.py
```diff
@@ -45,54 +95,66 @@ class RecentlyUsedContainer(MutableMapping):
...
             return item
 
-    def __setitem__(self, key, value):
-        evicted_value = _Null
+    def __setitem__(self, key: _KT, value: _VT) -> None:
+        evicted_value = None
         with self.lock:
             # Possibly evict the existing value of 'key'
-            evicted_value = self._container.get(key, _Null)
+            if key in self._container:
```
sethmlarson, line 123: Is the below code safe to do? Saves a lookup of `key in container`

```python
try:
    evicted_value = self._container.pop(key)
except KeyError:
    if len(self._container) >= self._maxsize:
        _, evicted_value = self._container.popitem(last=False)

self._container[key] = value
```

sethmlarson (changes requested):
Thanks for this, some early comments more coming later.

src/urllib3/_collections.py
```diff
@@ -242,25 +335,39 @@ def extend(self, *args, **kwargs):
...
-            for key in other.keys():
-                self.add(key, other[key])
-        else:
+            for key, val in other.items():
+                self.add(key, val)
+        elif isinstance(other, Iterable):
+            other = cast(Iterable[Tuple[str, str]], other)
             for key, value in other:
                 self.add(key, value)
+        elif hasattr(other, "keys") and hasattr(other, "__getitem__"):
```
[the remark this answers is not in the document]
haikuginger: Can't without making `typing_extensions` a runtime dependency; added it as a dev dependency here. `@runtime_checkable` made it into the stdlib in 3.8, so we can't do that without either adding the dependency (which I'm fine with, to be clear, but we've tried to keep runtime deps down in the past) or waiting until we drop 3.7.

src/urllib3/_collections.py
haikuginger: Will take a peek.

src/urllib3/compat/rlock.py
haikuginger: You're definitely right, but if we're commenting all the TODOs we can make better once we're on newer versions of Python, there are going to be a lot of those. 😉

.coveragerc
```diff
@@ -26,3 +26,5 @@ exclude_lines =
     .*:.* # Python \d.*
     .* # Abstract
     .* # Defensive:
+    if TYPE_CHECKING:
+    \.\.\.
```
haikuginger, line 30: These allow no-coverage for:

* blocks that only evaluate during type checking, since they'll never execute at runtime
* the bodies of protocol stubs and overload stubs (`Ellipsis` is roughly equivalent to `pass` but is typically used in contexts where it doesn't actually ever execute directly)

src/urllib3/exceptions.py
```diff
@@ -48,16 +62,17 @@ class SSLError(HTTPError):
...
+        super().__init__(message, error)
         self.original_error = error
 
 
 class HTTPSProxyError(ProxyError):
     """Used only when establishing a TLS connection to a proxy"""
 
-    def __init__(self, message, error, *args):
-        super().__init__(message, error, *args)
+    pass
```
haikuginger, line 75: Removed the `__init__` that was added here; it's technically written incorrectly, since `super().__init__()` can't take `*args`, and removing that item results in an identical signature to its parent, so we just inherit that instead.

.coveragerc
haikuginger: Actually considering whether I need to adjust the `\.\.\.` regex to be more defensive about only excluding coverage on lines where it's the only item, since it can be used in function type signatures (e.g., a function that takes `Callable[..., None]` as a param)

haikuginger: Validated issue. This code was marked as coverage-not-required with `\.\.\.`:

```python3
def has_no_coverage(something: Callable[..., None]) -> None:
    return something()
```

Replacing regex with:
```regex
^\s*?\.\.\.\s*$
```

This checks that the only non-whitespace content in the whole line is an ellipsis, as when used for the contents of a class or function body.

franekmagiera:
Hi, took a look at `_collections.py`. Maybe it would make sense to include a comment that when `Mapping` or `Iterable` is passed to `HTTPHeaderDict` it is assumed that it is `Mapping[str, str]` and iterable of string pairs? Small thing, but the only one that caught my attention. Overall it looks OK to me, nice work :)

franekmagiera:
Also, regarding making `typing extensions` a runtime dependency - if #1985 was to be implemented using `Protocol ` and not abc, wouldn't it be necessary to have `typing extensions` as a runtime dependency?

haikuginger:
> Also, regarding making `typing extensions` a runtime dependency - if #1985 was to be implemented using `Protocol ` and not abc, wouldn't it be necessary to have `typing extensions` as a runtime dependency?

Are you talking about how we describe the API? If so, then not strictly speaking. If we implemented a protocol in an `if TYPE_CHECKING:` block, we could refer to it via string annotations in places where it's returned or needs to be passed in. E.g.:

```python3

if TYPE_CHECKING:
    class HasSomeValuesInAList(Protocol):
        that_list_i_told_you_about: List[int]

@dataclass
class ImplementationOfAbove:
    some_other_thing: str
    that_list_i_told_you_about: List[int]

def does_stuff_with_those_things(carries_vals: "HasSomeValuesInAList") -> None:
    for item in carries_vals.that_list_i_told_you_about:
        print(item)

# Valid
does_stuff_with_those_things(ImplementationOfAbove("the string", [1, 2, 3, 4, 5]))
```

Using string annotations results in a `ForwardRef` being used rather than the actual class in question, which means that the actual class in question can not exist at runtime and it'll be fine—as long as it's there when the typechecker goes looking.

haikuginger:
Tagging @SethMichaelLarson for re-review. I think this is ready to merge.

src/urllib3/compat/rlock.py
sethmlarson: Can this definition be placed elsewhere? I don't want to create new `urllib3.*` packages that aren't private to avoid confusing users on what our supported APIs are.

haikuginger: @SethMichaelLarson, how would you feel about a `urllib3.compat._shims` module to put stuff like this?

sethmlarson: (btw my handle is @sethmlarson) and moving to `urllib3._compat` would be enough, the `_*` shows it's private same as `_collections`.

src/urllib3/_collections.py
```diff
@@ -28,10 +55,38 @@ def __exit__(self, exc_type, exc_value, traceback):
...
 
 
-_Null = object()
-
-
-class RecentlyUsedContainer(MutableMapping):
+_KT = TypeVar("_KT")
+_VT = TypeVar("_VT")
+
+ValidHttpHeaderSource = Union[
```
sethmlarson, line 61: nit: casing as `HTTPHeader` instead of `HttpHeader` matches the casing elsewhere (ie `HTTPHeaderDict`)

src/urllib3/_collections.py
```diff
@@ -45,54 +100,67 @@ class RecentlyUsedContainer(MutableMapping):
...
-    def __init__(self, maxsize=10, dispose_func=None):
+    def __init__(
+        self, maxsize: int = 10, dispose_func: Optional[Callable[[_VT], None]] = None
+    ) -> None:
+        super().__init__()
         self._maxsize = maxsize
         self.dispose_func = dispose_func
-
-        self._container = self.ContainerCls()
+        self._container = OrderedDict()
```
sethmlarson, line 114: Adding to the breaking changes list "You can't set `RecentlyUsedContainer.ContainerCls` anymore". Totally fine with this change, this class is internal and shouldn't be mucked with.

src/urllib3/_collections.py
```diff
@@ -45,54 +100,67 @@ class RecentlyUsedContainer(MutableMapping):
...
             self._container[key] = value
 
-            # If we didn't evict an existing value, we might have to evict the
-            # least recently used item from the beginning of the container.
-            if len(self._container) > self._maxsize:
-                _key, evicted_value = self._container.popitem(last=False)
-
-        if self.dispose_func and evicted_value is not _Null:
+        # Release the lock on the pool, and dispose of the evicted value.
+        if evicted_value is not None and self.dispose_func:
```
sethmlarson, line 144: Isn't it now possible that an evicted `None` value won't have `dispose_func` called? Maybe we need a flag that is set if we actually evict anything. Not sure if this actually matters, but I do know `None` is inserted into ConnectionPools.

src/urllib3/_collections.py
```diff
@@ -242,25 +341,39 @@ def extend(self, *args, **kwargs):
...
         for key, value in kwargs.items():
             self.add(key, value)
 
-    def getlist(self, key, default=__marker):
+    @overload
+    def getlist(self, key: str) -> List[str]:
+        ...
+
+    @overload
+    def getlist(self, key: str, default: List[str]) -> List[str]:
```
sethmlarson, line 367: Wouldn't this override need to be `Union[List[str], DefaultType` where `DefaultType` is whatever type `default` is?

src/urllib3/_collections.py
```diff
@@ -274,34 +387,36 @@ def getlist(self, key, default=__marker):
...
         for key in self:
             val = self._container[key.lower()]
             yield val[0], ", ".join(val[1:])
 
-    def items(self):
-        return list(self.iteritems())
+    def items(self) -> HTTPHeaderDictItemView:
+        return HTTPHeaderDictItemView(self)
+
+    def has_value_for_header(self, header_name: str, potential_value: str) -> bool:
```
sethmlarson, line 419: Can this method be made private?

src/urllib3/compat/rlock.py
haikuginger: My bad @sethmlarson! Autofill got me...

src/urllib3/_collections.py
```diff
@@ -102,12 +170,57 @@ def clear(self):
...
 
-    def keys(self):
+    def keys(self) -> Set[_KT]:
         with self.lock:
-            return list(self._container.keys())
+            return set(self._container.keys())
 
 
-class HTTPHeaderDict(MutableMapping):
+class HTTPHeaderDictItemView(Set[Tuple[str, str]]):
```
sethmlarson, line 178: I'm confused why this is `Set[Tuple[str, str]]`?

src/urllib3/_collections.py
```diff
@@ -102,12 +170,57 @@ def clear(self):
...
+
+    This class conforms to the interface required by the MutableMapping ABC while
+    also giving us the nonstandard iteration behavior we want; items with duplicate
+    keys, ordered by time of first insertion.
+    """
+
+    headers: "HTTPHeaderDict"
+
+    def __init__(self, headers: "HTTPHeaderDict") -> None:
+        self.headers = headers
```
sethmlarson, line 207: Shouldn't `header` be private since this object is opaque?

sethmlarson (changes requested):
Thanks for the changes, more comments for you!

src/urllib3/util/response.py
sethmlarson: This is all fine since we have the `isinstance` check, so we don't have to worry about users sending weird objects into this function

src/urllib3/_collections.py
haikuginger: The type of the return value from `MutableMapping[_KT, _VT].items()` is a not-strictly-ordered `AbstractSet[Tuple[_KT, _VT]]`, so this needs to match that in order to conform.

src/urllib3/_collections.py
haikuginger: That's a good catch; this would not call the eviction function for some `evicted_value` if `_VT` is a supertype of `None`. Will play with some things.

src/urllib3/_collections.py
haikuginger: We can; we are calling it from another class though.

src/urllib3/_collections.py
haikuginger: Yeah, if we want to support some default that isn't `List[str]`, we'd need to do this:

```python3
@overload
def getlist(self, key: str, default: _DT) -> Union[List[str], _DT]]: ...
```
(in which case `Union[List[str], _DT]` collapses to `List[str]` in situations where `_DT` can be proven to be `List[str]`)

The question is whether we want to support passing a default that isn't `List[str]`; I can take a look through the existing codebase to see what we do and where; the most likely case is that we pass in `None` in some places, and at `typeshed` to see what they do for builtins that accept defaults.

src/urllib3/_private_compat/rlock.py
```diff
@@ -0,0 +1,12 @@
+"""
```
sethmlarson, line 1: Adding a new module requires a change to `setup.py:packages=[...]`, why not `urllib3._compat`?

src/urllib3/_collections.py
sethmlarson: Hmm, could we get the same properties from `typing.Collection`? Seems it is `Sized`, `Container`, and `Iterable`. It looks like `Set` is simply `Collection` with comparison methods and the `isdisjoint` method.

src/urllib3/_private_compat/rlock.py
haikuginger: Got it—you mean a single `_compat.py` file at the top level?

src/urllib3/_collections.py
haikuginger: No; because `HTTPHeaderDict` is `MutableMapping`, any `HTTPHeaderDict` `x` can be passed to any `Callable[[MutableMapping], object]` `f`, and `f` can expect `x.items()` to return an item that is `AbstractSet`, which `Collection` isn't enough to conform to.

I'm curious what the issue with using `Set` is? Note that this is the ABC `Set` and not the concrete stdlib `set` (which implements the `Set` interface).

sethmlarson: Ah TIL that `ItemsView` is a `Set`, thanks for showing me this. This looks good then, sorry for the confusion!

src/urllib3/_private_compat/rlock.py
sethmlarson: Yeah!

sethmlarson (approved):
LGTM, thanks for moving this mountain! :tada:

PR #2118, opened 2020-12-22
Bubble up errno from OpenSSL.SSL.SysCallError (#2117)

### LINKED ISSUE #2117: pyopenssl.py should explictly bubble up errno from OpenSSL.SSL.SysCallError
### Context

Several spots in pyopenssl catch OpenSSL.SSL.SysCallError and then reraise the exception as SocketError (an alias for socket.error).

Example code:

https://github.com/urllib3/urllib3/blob/83652e0fe9113fbe35f7c34ee71c2f7ed268c5f9/src/urllib3/contrib/pyopenssl.py#L278-L282

`raise SocketError(str(e))` passes all the information about the underlying exception as a string.  The string includes the underlying `errno`.  There's an opportunity to explicitly include the `errno` in the re-raised exception.  Libraries that use urllib3 can then extract the errno if it's present.

### Alternatives

No obvious alternatives.

### Contribution

Would you be willing to submit a PR?  Yes, a PR is forthcoming.

dossett:
Closes #2117

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -279,7 +279,7 @@ class WrappedSocket:
             if self.suppress_ragged_eofs and e.args == (-1, "Unexpected EOF"):
                 return b""
             else:
-                raise SocketError(str(e))
+                raise SocketError(e.args[0], str(e))
         except OpenSSL.SSL.ZeroReturnError:
             if self.connection.get_shutdown() == OpenSSL.SSL.RECEIVED_SHUTDOWN:
                 return b""
@@ -304,7 +304,7 @@ class WrappedSocket:
             if self.suppress_ragged_eofs and e.args == (-1, "Unexpected EOF"):
                 return 0
             else:
-                raise SocketError(str(e))
+                raise SocketError(e.args[0], str(e))
         except OpenSSL.SSL.ZeroReturnError:
             if self.connection.get_shutdown() == OpenSSL.SSL.RECEIVED_SHUTDOWN:
                 return 0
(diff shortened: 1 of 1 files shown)


src/urllib3/contrib/pyopenssl.py
```diff
@@ -279,7 +279,7 @@ def recv(self, *args, **kwargs):
             if self.suppress_ragged_eofs and e.args == (-1, "Unexpected EOF"):
                 return b""
             else:
-                raise SocketError(str(e))
+                raise SocketError(e.args[0], str(e))
```
dossett, line 282: SysCallError is always construct with a (errnum, message) tuple
https://www.pyopenssl.org/en/stable/api/ssl.html#OpenSSL.SSL.SysCallError

sigmavirus24:
`SocketError` here is `socket.error` which is an alias of [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError) which takes `errno` as the first argument (in case anyone else reviewing this was worried about what changing the parameters to the constructor would do)

dossett:
> `SocketError` here is `socket.error` which is an alias of [`OSError`](https://docs.python.org/3/library/exceptions.html#OSError) which takes `errno` as the first argument (in case anyone else reviewing this was worried about what changing the parameters to the constructor would do)

Thanks for calling that out @sigmavirus24.  I opened a separate issue and (soon) a PR to use `OSError` directly to make it all a  little clearer.

sigmavirus24 (approved):
This looks good to me

sethmlarson (approved):
LGTM

PR #2122, opened 2020-12-25
Remove outdated workaround for missing ipaddress

jdufresne:
Outdated since urllib3 dropped Python 2 support. The ipaddress module
has been available in the stdlib since Python 3.3 and all used functions
are available.


hramezani (approved):

sethmlarson (approved):
LGTM, thanks!

PR #2124, opened 2020-12-29
Use match in pytest.raises.

hramezani:

### DIFF
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -223,8 +223,8 @@ class TestRetry:
         assert "Caused by redirect" not in str(e.value)
 
         retry = Retry(total=1)
+        retry = retry.increment("POST", "/")
         with pytest.raises(MaxRetryError, match=ResponseError.GENERIC_ERROR) as e:
-            retry = retry.increment("POST", "/")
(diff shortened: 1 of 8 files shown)


test/test_retry.py
```diff
@@ -218,36 +216,32 @@ def test_disabled(self, total):
...
             retry = retry.increment("POST", "/", response=response)
             retry = retry.increment("POST", "/", response=response)
         assert "Caused by redirect" not in str(e.value)
-        msg = ResponseError.SPECIFIC_ERROR.format(status_code=500)
-        assert str(e.value.reason) == msg
 
         retry = Retry(connect=1)
-        with pytest.raises(MaxRetryError) as e:
+        with pytest.raises(MaxRetryError, match="conntimeout") as e:
             retry = retry.increment(error=ConnectTimeoutError("conntimeout"))
```
sethmlarson, line 242: I feel like this line should be moved outside of the `pytest.raises()`

sethmlarson (approved):
Thanks for this, looks good to me! I changed some of the retry tests to call a single `Retry.increment()` within a `pytest.raises` so we can test exactly which invocation should raise the error.

Issue #2123, opened 2020-12-29
When iterating over a `stream`ed response with amt=None, entire response is returned instead of streaming


djrobstep:
### Subject

While using requests, I noticed that passing `chunk_size=None` into iter_content returns the entire response when it should yield the streamed data as it arrives.

I've tracked this down to the `read(` method of `HTTPResponse`, and have a fix which simply changes this method to call `read1` instead of `read` on the underlying `_fp` when `amt=None`.

I'm happy to prepare a PR with this fix if that would be helpful (I have not written a proper test yet).

### Environment

OS macOS-10.14.6-x86_64-i386-64bit
Python 3.9.0
urllib3 1.26.2

### Steps to Reproduce

I have a reproducible example if this in the related requests issue: https://github.com/psf/requests/issues/5536

### Expected Behavior

Streamed response data should be yielded as it arrives.

### Actual Behavior

Entire response is returned.

sethmlarson:
Agree that there should be a mechanism to receive data as soon as its available without setting a tiny `amt`. We've had problems in the past with our `HTTPResponse` object claiming to be an `IOBase` but having slightly different behavior compared to an `IOBase` object.

djrobstep:
Hey, just wondering about the status of this one? It sounds like v2 is not far away, now might be an opportune time to incorporate changes on this (assuming this hasn't been fixed already)?

kyczawon:
Hi, any updates on this issue?

PR #2125, opened 2020-12-31
yield streamed bytes as soon as they arrive

djrobstep:
Right now, calling `read` with `amt = None` (or `requests.iter_content(chunk_size=None)`) seems to wait for the entire response contents to arrive before yielding, which contradicts the docs in both `urllib3` and `requests`. This is discussed in #2123.

For instance, this change makes the following correctly yield bytes as they arrive, provided `CHUNK_SIZE` is set to `-1`:

```
import requests

URL = 'https://httpbin.org/drip?duration=2'

r = requests.get(URL, stream=True)

for x in r.iter_content(chunk_size=CHUNK_SIZE):
    print(f'response: {x}')
```

The requests docs say this should already be happening if CHUNK_SIZE is set to `None`, but it doesn't:

> chunk_size must be of type int or None. A value of None will function differently depending on the value of stream. stream=True will read data as it arrives in whatever size the chunks are received. If stream=False, data is returned as a single chunk

And so do the docs for `HTTPResponse`'s `stream(` that this value is passed through to:

> How much of the content to read. The generator will return up to this much data per iteration, but may return less. This is particularly likely when using compressed data. However, the empty string will never be returned.

But in fact `None` forces the entire response to be read before yielding.

I was hesitant to make a change that would affect the existing behaviour of `None` or positive integers, so here you specify a chunk size of `-1` to activate this "stream on any bytes" behaviour. Hopefully this is suitable.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -583,7 +583,7 @@ class HTTPResponse(BaseHTTPResponse):
         :param amt:
             How much of the content to read. If specified, caching is skipped
             because it doesn't make sense to cache partial content as the full
-            response.
+            response. If -1, return bytes as soon as any become available.
 
         :param decode_content:
             If True, will attempt to decode the body based on the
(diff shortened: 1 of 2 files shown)


sethmlarson:
Sorry that this PR has stalled completely for over a year and now has conflicts. Unless there's someone on our team that wants to push this one over the finish line I may close it.

I'm not sure I love the `.read(-1)` being used to mean `.read1()`, shouldn't we instead implement `HTTPResponse.read1()`?

Silverfishnet (approved):

pquentin:
Agreed that `read1()` is the way to go here.

sethmlarson:
Going to close this PR in favor of one implementing read1().

Issue #2126, opened 2020-12-31
HTTPSConnection fails when assert_hostname is True


ddebeau:
### Subject

HTTPSConnection fails to check hostnames properly when `assert_hostname` is set to `True`.

### Environment

Describe your environment.
At least, paste here the output of:

```python
OS Darwin-19.6.0-x86_64-i386-64bit
Python 3.6.12
urllib3 1.25.11
```

### Steps to Reproduce

```python
import urllib3

manager = urllib3.PoolManager(assert_hostname=True)
r = manager.request('GET', 'https://google.com/')
print(r.status)
```

### Expected Behavior

```
200
```

### Actual Behavior

```python
Traceback (most recent call last):
  File "<redacted>/lib/python3.6/site-packages/urllib3/connectionpool.py", line 677, in urlopen
    chunked=chunked,
  File "<redacted>/lib/python3.6/site-packages/urllib3/connectionpool.py", line 381, in _make_request
    self._validate_conn(conn)
  File "<redacted>/lib/python3.6/site-packages/urllib3/connectionpool.py", line 978, in _validate_conn
    conn.connect()
  File "<redacted>/lib/python3.6/site-packages/urllib3/connection.py", line 397, in connect
    _match_hostname(cert, self.assert_hostname or server_hostname)
  File "<redacted>/lib/python3.6/site-packages/urllib3/connection.py", line 407, in _match_hostname
    match_hostname(cert, asserted_hostname)
  File "<redacted>/lib/python3.6/ssl.py", line 327, in match_hostname
    % (hostname, ', '.join(map(repr, dnsnames))))
[...]

PR #2127, opened 2020-12-31
Remove assert_hostname from hostname check

ddebeau:
Closes: https://github.com/urllib3/urllib3/issues/2126

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -306,5 +306,8 @@ In chronological order:
 * [Ezzeri Esa] <https://github.com/savarin>
   * Ports and extends on types from typeshed
 
+* [David Debeau] <[ddebeau@enthought.com]>
+  * Fix how HTTPSConnection handles assert_hostname
+
 * [Your name or handle] <[email or website]>
   * [Brief summary of your changes]
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -415,7 +415,7 @@ class HTTPSConnection(HTTPConnection):
             # the TLS library, this cannot always be done. So we check whether
             # the TLS Library still thinks it's matching hostnames.
             cert = self.sock.getpeercert()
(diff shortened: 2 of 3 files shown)


sethmlarson:
`assert_hostname=True` isn't supported and isn't documented anywhere as far as I can tell? `assert_hostname` can either be a hostname or `False` to disable it. Instead of removing we should instead document `assert_hostname` better?

ddebeau:
@sethmlarson Passing a bool to the underlying function `ssl.match_hostname` will result in an exception.

https://github.com/python/cpython/blob/b5711c940f70af89f2b4cf081a3fcd83924f3ae7/Lib/ssl.py#L371

sethmlarson:
I'm aware, I'm suggesting that passing `True` isn't a documented or supported input for `assert_hostname` so there isn't anything to fix/remove here aside from documenting usage better.

ddebeau:
I don't understand why you want to leave the possibility for a bool to be passed to `_match_hostname` when it will result in an exception. I don't see the purpose of having it there.

Accepting only `False` instead of `True` or `False` seems bizarre.

sethmlarson:
Closing this due to inactivity.

pquentin:
@sethmlarson For what it's worth I have the same question than @ddebeau above. Is the idea that we don't want to break our API?

sethmlarson:
It wouldn't be a breakage, I'd accept a PR that allows for `assert_hostname=True`

pquentin:
What I had not understood in your initial explanation, Seth, is that `assert_hostname` could also be a host. It's not only a boolean!

> Accepting only False instead of True or False seems bizarre.

@ddebeau This isn't true: we accept False, None and an actual hostname.

Actually, it would be better to document is as a Optional[str], an optional hostname: since the code uses `or`, it works too. And it would avoid the confusion. Accepting `True` isn't useful, there are many other places where the code will break by sending the wrong type. We should continue to accept `False` for backwards compatibility, though.

Issue #2105, opened 2020-12-09
Better error message when connecting to HTTPS proxy results in non-certificate connection errors


sethmlarson:
Many users attempting to connect to "HTTPS" proxies that receive errors like:

- RemoteDisconnected
- Connection reset by peer
- [SSL: WRONG_VERSION_NUMBER] wrong version number

we should somehow make it more obvious that the error is likely because their proxy wants to speak HTTP, not HTTPS.
Not sure by what exact method we can do this, potentially the `ProxyManager` can do something here?

References:
- https://github.com/pypa/pip/issues/9216

pquentin:
Do we understand why https://github.com/urllib3/urllib3/issues/1850 did not catch this? (This issue now has four thumbs up, by the way.)

sethmlarson:
@pquentin I think it's because Pip wasn't handling/enabling the new warning so wasn't visible to users.

jalopezsilva:
The ``ssl`` APIs make it difficult to clearly send a a consistent error for these cases. The best we can do probably is catch errors when we are trying to establish the TLS session with the proxy, and raise another exception with a better error: https://github.com/urllib3/urllib3/blob/master/src/urllib3/connection.py#L329

I'll take a stab at that unless anyone has better suggestions.

pradyunsg:
> @pquentin I think it's because Pip wasn't handling/enabling the new warning so wasn't visible to users.

Yea -- let me know if there's some way for pip to better expose urllib3's warnings to our end users.

sethmlarson:
This has been solved in https://github.com/urllib3/urllib3/pull/2434

pradyunsg:
Neato! Imma be annoying and ask: when's the release? 🙃

sethmlarson:
@pradyunsg if no blockers are found then next week is the plan.

PR #2133, opened 2021-01-11
Add exception wrapper for HTTPS proxy connection failure

jalopezsilva:
From https://github.com/urllib3/urllib3/issues/2105 we saw that it can be confusing for users when their connection fails to connect to a TLS proxy. The generated error is not consistent. To make it easier to diagnose I'm proposing we wrap the initial connection failure error with a custom exception to make it clearer when we're attempting to connect to an HTTPS proxy. 

Putting this up as an RFC, happy to adjust as needed.


pquentin (approved):
Thanks, this looks great!

@sethmlarson Do we want to backport this to 1.26.x?

PR #2134, opened 2021-01-11
Use coverage version with PyPy workaround

pquentin:
I thought we were using this already, but no. This should make PyPy builds much more reliable.


sethmlarson (approved):
:tada:

PR #2139, opened 2021-01-22
Add types to URL parser, remove get_host() and split_first()

sethmlarson:
- Make all `urllib3.util.url` regular expressions private to protect our internal APIs
- Change `Url` to a `typing.NamedTuple`
- Removed `get_host` and `split_first` deprecated APIs
- `idna` doesn't have type hints *yet*! But will soon :)


pquentin:
@sethmlarson Is this ready to review now that the coverage is 💯?

sethmlarson:
@pquentin Ready for review! :)

src/urllib3/util/url.py
```diff
@@ -1,20 +1,18 @@
 import re
-from collections import namedtuple
+import typing
```
pquentin, line 2: Can we be consistent between `import typing` and `from typing import *`? I slightly prefer the latter.

src/urllib3/util/url.py
```diff
@@ -164,46 +177,27 @@ def url(self):
...
-    min_delim = None
-    for d in delims:
-        idx = s.find(d)
-        if idx < 0:
-            continue
+@typing.overload
+def _encode_invalid_chars(
+    component: None, allowed_chars: typing.Container[str]
+) -> None:  # Abstract
+    ...
```
pquentin, line 195: TIL about typing.overload, thanks!

(It would have been cleaner to never call `_encode_invalid_chars` with a `None` component, but maybe that would have resulted in a lot of boilerplate in callers, and anyway let's not change the code when adding types.)

pquentin (approved):
Thanks!

Issue #2071, opened 2020-11-18
connection.py: BytesWarning: Comparison between bytes and string: if SKIP_HEADER not in values:


jdufresne:
### Subject

Starting with the 1.26 series, I started receiving a `BytesWarning` in my project's test suite.

```
.../urllib3/connection.py:218: BytesWarning: Comparison between bytes and string
  if SKIP_HEADER not in values:
```

I run my test suite with Python's `-b` option to help catch misuses of bytes and string objects. My hunch is that urllib3 is sometimes treating headers as bytes and other times as string and then comparing them with each other.

### Environment

Describe your environment.
At least, paste here the output of:

```python
import platform
import urllib3

print("OS", platform.platform())
print("Python", platform.python_version())
print("urllib3", urllib3.__version__)
```

```
OS Linux-5.8.16-200.fc32.x86_64-x86_64-with-glibc2.2.5
Python 3.8.6
urllib3 1.26.2
```

### Steps to Reproduce

This is reproducible using urllib3's own test suite:

Invoke using:

```
$ python -b -m pytest -W a test/with_dummyserver/test_socketlevel.py
...
test/with_dummyserver/test_socketlevel.py::TestProxyManager::test_connect_ipv6_addr
  /home/jon/.virtualenvs/urllib3/lib64/python3.9/site-packages/urllib3/connection.py:193: BytesWarning: Comparison between bytes and string
    if SKIP_HEADER not in values:
```

### Expected Behavior

No `ByteWarning` emitted.

### Actual Behavior

```
.../connection.py:193: BytesWarning: Comparison between bytes and string
    if SKIP_HEADER not in values:
```

sethmlarson:
Nice catch, we'll have to enable this in our test suite and remove the issue.

The-Compiler:
I'm seeing this as well FWIW - I did a quick bisect and this was introduced in 16b7b332fd1b84c2d465f11d17658c1e83d3f20f (#2018, "Add SKIP_HEADER for skipping automatically added headers").

mgorny:
Ping. This is causing a lot of cron spam for a lot of our users.

fred-vogt:
Breaks `python ... -bb ...`, until this is fixed, can be worked around by downgrading in pip:
```
urllib3==1.25.11
```

pquentin:
@mgorny Can you please clarify what Gentoo cron spam is?

[...]

PR #2141, opened 2021-01-25
Never compare bytes and str in putheader()

pquentin:
Fixes https://github.com/urllib3/urllib3/issues/2071. It's ugly, but I don't know how else to fix that. 

I can't enable `-bb` before https://github.com/python-hyper/brotlicffi/pull/177 is merged and released. (edit: disregard, there are other issues to fix before we can enable `-bb`.)

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -25,7 +25,7 @@ except (ImportError, AttributeError):  # Platform-specific: No SSL.
 from ._version import __version__
 from .exceptions import ConnectTimeoutError, NewConnectionError, SystemTimeWarning
 from .packages.ssl_match_hostname import CertificateError, match_hostname
-from .util import BYTES_SKIP_HEADER, SKIP_HEADER, SKIPPABLE_HEADERS, connection
+from .util import SKIP_HEADER, SKIPPABLE_HEADERS, connection
 from .util.ssl_ import (
     assert_fingerprint,
     create_urllib3_context,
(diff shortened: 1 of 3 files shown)


mgorny:
I agree it's ugly-ish but nevertheless kudos for finding *some* way to solve this. I've been pondering on it for some time and couldn't come up with anything even that good.

pquentin:
Actually with the second commit it's not that bad!

src/urllib3/connection.py
```diff
@@ -184,7 +184,7 @@ def putrequest(self, method, url, *args, **kwargs):
 
     def putheader(self, header, *values):
         """"""
-        if SKIP_HEADER not in values:
+        if not any(isinstance(v, str) and v == SKIP_HEADER for v in values):
```
sethmlarson, line 187: Thoughts on using `type(v) is str` instead of `isinstance()`? Also can we use the `-bb` option as you've done in brotlicffi to ensure these error in the future?

sethmlarson (commented):
This looks good, one thought:

src/urllib3/connection.py
sigmavirus24: Why `type(v) is str`?

pquentin: As mentioned in the initial comment, I can't enable `-bb` because of brotlicffi (and other issues that I haven't investiaged yet).

I think `isinstance()` is the pythonic way to do this, and it performs just as well as `type() is str` (consistently less than 100nsec as mesured by timeit on my laptop).

sethmlarson: Was looking for a speed win in this often-called function :) It's all good though, let's merge this!

sethmlarson (approved):
LGTM

PR #2148, opened 2021-01-29
Remove deprecated iter_fields(), add type hints to 'fields' and 'filepost' modules

sethmlarson:
More removals and more types!


src/urllib3/fields.py
```diff
@@ -76,7 +93,7 @@ def replacer(match):
     return result
 
 
-def format_header_param_html5(name, value):
+def format_header_param_html5(name: str, value: Union[str, bytes]) -> str:
```
pquentin, line 96: Should we use `_TYPE_FIELD_VALUE` here too, as we do in `RequestField.__init__`?

src/urllib3/fields.py
```diff
@@ -231,28 +272,34 @@ def render_headers(self):
...
 
+        :param content_disposition:
+            The 'Content-Disposition' of the request body. Defaults to 'form-data'
         :param content_type:
             The 'Content-Type' of the request body.
         :param content_location:
             The 'Content-Location' of the request body.
 
         """
-        self.headers["Content-Disposition"] = content_disposition or "form-data"
```
pquentin, line 248: Would it be better to encode the default value in the kwarg? This would avoid using `Optional` and would be consistent with `guess_content_type`.

I understand that we want to just add types and not change the code, but this type of small change is worth it in my opinion as it results in better types.

pquentin (dismissed):
This looks good to me, thanks! I left two suggestions, feel free to include them or not.

src/urllib3/fields.py
sethmlarson: I like this idea! We can keep functional compatibility + stay defensive by keeping the `... or "form-data"` but still recommend users only pass `str`.

pquentin: Hmm, I hadn't thought about keeping the default value in the code. It's s good point. That would mean keeping two copies of that default value, and they'll go out of sync eventually.

I'm now only in favor of this idea if we say the argument isn't Optional and accept that passing None will now break. Not sure if this is acceptable or not.

pquentin: Probably not acceptable. I added a commit with the other accepted change, and will merge if tests pass.

pquentin (approved):

pquentin:
All checks passed (https://github.com/urllib3/urllib3/pull/2148/checks), but GitHub won't admit it... Closing/reopening.

PR #2149, opened 2021-01-30
Add defensive test cases for brotlipy

sethmlarson:
brotlipy shadows the 'brotli' module namespace and we previously used brotlipy so to be defensive we should continue to test that 'brotlipy' works.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -83,6 +83,15 @@ def unsupported_python2(session):
     assert "Unsupported Python version" in process.stderr
 
 
+@nox.session(python=["3"])
+def test_brotlipy(session):
+    """Check that if 'brotlipy' is installed instead of 'brotli' or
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -99,6 +108,8 @@ def format(session):
     # Ensure that pre-commit itself ran successfully
     assert process.returncode in (0, 1)
 
+    lint(session)
```
sethmlarson, line 111: Not related to this PR directly but I noticed this got removed after the pre-commit case, imo it makes sense to call `lint` after `format` to ensure your changes post-format are good to go.

pquentin (commented):
Thanks for this! Sorry for the late review.

sethmlarson:
@pquentin No worries! This was a small change I was pretty confident merging :)

Issue #1993, opened 2020-09-25
[v2] Add a top-level urllib3.request() method


sethmlarson:
Use a default `PoolManager()` instance configured with certifi if it's available. Document it as not recommended for production use, recommended only for experimenting or quick scripts.

imba-tjd:
Would you consider adding a default `Accept-Encoding: gzip` or sth? As I understand `identity` means no compression.

sigmavirus24:
@imba-tjd this is completely unrelated to the original issue

imba-tjd:
Yes, it's related. I'm only thinking adding the header to this "quick scripts".

sigmavirus24:
I'd argue that changing the default for the library versus adding a convenience function as a proxy to a method on a commonly used object are two orthogonal feature requests. While this may have spurred the idea, they are not inextricably related and thus separate concerns

shazow:
I'm curious, what's the thinking behind this part?

> Document it as not recommended for production use, recommended only for experimenting or quick scripts.

Perhaps better to document that this is a module-global instance and side effects on this instance from other code will be retained across dependencies?

pquentin:
Done in #2150.

PR #2150, opened 2021-02-03
Add top level request method

franekmagiera:
My proposition for #1993 
Was not sure if this feature needed tests as it reuses already tested `PoolManager`.
Let me know what you think.

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -55,6 +55,9 @@ HTTP verb:
 The :ref:`request_data` section covers sending other kinds of requests data,
 including JSON, files, and binary data.
 
+.. note:: For quick scripts and experiments you can also use a top-level ``urllib3.request()``.
+    However, it is not recommended for production use!
+
 .. _response_content:
 
 Response Content
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -56,7 +56,10 @@ The :ref:`request_data` section covers sending other kinds of requests data,
 including JSON, files, and binary data.
 
 .. note:: For quick scripts and experiments you can also use a top-level ``urllib3.request()``.
-    However, it is not recommended for production use!
+    It uses a module-global ``DEFAULT_POOL`` ``PoolManager`` instance.
+    Because of that, its side effects could be shared across dependencies relying on it.
+    To avoid side effects, create a new ``PoolManager`` instance and use it instead.
+    In addition, the method skips the low level ``**urlopen_kw`` keyword arguments. It is not recommended for use in production.
 
 .. _response_content:
 
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -56,7 +56,7 @@ The :ref:`request_data` section covers sending other kinds of requests data,
 including JSON, files, and binary data.
 
 .. note:: For quick scripts and experiments you can also use a top-level ``urllib3.request()``.
-    It uses a module-global ``_DEFAULT_POOL`` ``PoolManager`` instance.
+    It uses a module-global ``PoolManager`` instance.
     Because of that, its side effects could be shared across dependencies relying on it.
     To avoid side effects, create a new ``PoolManager`` instance and use it instead.
     In addition, the method does not accept the low-level ``**urlopen_kw`` keyword arguments.
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -56,10 +56,11 @@ The :ref:`request_data` section covers sending other kinds of requests data,
 including JSON, files, and binary data.
 
 .. note:: For quick scripts and experiments you can also use a top-level ``urllib3.request()``.
-    It uses a module-global ``DEFAULT_POOL`` ``PoolManager`` instance.
+    It uses a module-global ``_DEFAULT_POOL`` ``PoolManager`` instance.
     Because of that, its side effects could be shared across dependencies relying on it.
     To avoid side effects, create a new ``PoolManager`` instance and use it instead.
-    In addition, the method skips the low level ``**urlopen_kw`` keyword arguments. It is not recommended for use in production.
+    In addition, the method does not accept the low-level ``**urlopen_kw`` keyword arguments.
+    System CA certificates are loaded on default.
 
 .. _response_content:
 
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -94,4 +94,4 @@ def request(method, url, fields=None, headers=None):
     The method does not accept low-level ``**urlopen_kw`` keyword arguments.
     """
 
-    return _DEFAULT_POOL.request(method, url, fields, headers)
+    return _DEFAULT_POOL.request(method, url, fields=fields, headers=headers)
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -37,6 +37,7 @@ __all__ = (
     "get_host",
     "make_headers",
     "proxy_from_url",
+    "request",
 )
 
 logging.getLogger(__name__).addHandler(NullHandler())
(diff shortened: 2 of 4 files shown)


shazow:
We recently had a discussion about this in our community Discord channel, here's the relevant snippet:
image

franekmagiera:
Thanks for your response. So, should I just change `_manager` to `_DEFAULT_POOL` and it would be good to go (assuming you are ok with the code)? Or are you still considering adding other features for this?

shazow:
I'll let @SethMichaelLarson make the final call here.

Personally I'd say let's make it `DEFAULT_POOL`, change the `urllib3.request(...)` method to be a stand-alone function that calls `DEFAULT_POOL.request(...)` while removing the `**urllib3_kw` args from the call signature.

```python
def request(...):
    return DEFAULT_POOL.request(...)
```

Also IMO, the docstring should explain that `urllib3.request` uses a module-global `DEFAULT_POOL` instance and that side effects could be shared across dependencies that rely on it. To avoid side effects, create a new PoolManager and use that instead.

I think more details about what the issue might be is better than just blankly telling people not to use it in production. :)

pquentin:
+1 to what @shazow said.

Also, we do need a test in `test/with_dummyserver/test_poolmanager.py` to make sure nobody removes this by accident. Untested, but it could look like this:

```
    def test_top_level_request(self):
        with PoolManager() as http:
            r = http.request("GET", f"{self.base_url}/")
            assert r.status == 200
            assert r.data == b"Dummy server!"
```

franekmagiera:
Thanks for the pointers, they have been really helpful, I added suggested changes.

I've got a check failing for Ubuntu 3.1-dev test-3.10 in `test_http.py`. for now I don't know what's causing the 'Address already in use error' - tornado's `bind_socket` is called with `None` as far as I checked and should allocate ports automatically.

Also Ubuntu 3.8 seems to be stuck on set up for some reason.

pquentin:
Closing/reopening to see what failures are real (they don't look related to your pull request).

docs/user-guide.rst
```diff
@@ -55,6 +55,12 @@ HTTP verb:
 The :ref:`request_data` section covers sending other kinds of requests data,
 including JSON, files, and binary data.
 
+.. note:: For quick scripts and experiments you can also use a top-level ``urllib3.request()``.
+    It uses a module-global ``DEFAULT_POOL`` ``PoolManager`` instance.
+    Because of that, its side effects could be shared across dependencies relying on it.
+    To avoid side effects, create a new ``PoolManager`` instance and use it instead.
+    In addition, the method skips the low level ``**urlopen_kw`` keyword arguments. It is not recommended for use in production.
```
sethmlarson, line 62: We can remove the note about not being able to use in production. It's production capable given you're okay with the above constraints.

src/urllib3/__init__.py
```diff
@@ -80,3 +82,28 @@ def disable_warnings(category=exceptions.HTTPWarning):
...
+# To avoid side effects create a new PoolManager instance and use it instead.
+try:
+    import certifi
+
+    DEFAULT_POOL = PoolManager(cert_reqs="CERT_REQUIRED", ca_certs=certifi.where())
+except ImportError:
+    DEFAULT_POOL = PoolManager()
+
+
+def request(method, url, fields=None, headers=None, **urlopen_kw):
```
sethmlarson, line 99: Shouldn't we omit `**urlopen_kw` in the signature if we're going to drop them on the floor anyways?

src/urllib3/__init__.py
```diff
@@ -80,3 +82,28 @@ def disable_warnings(category=exceptions.HTTPWarning):
...
     """
     warnings.simplefilter("ignore", category)
+
+
+# PoolManager for a top-level request method.
+# DEFAULT_POOL is a module global PoolManager instance,
+# therefore its side effects could be shared across dependencies relying on it.
+# To avoid side effects create a new PoolManager instance and use it instead.
+try:
+    import certifi
```
sethmlarson, line 92: I'm not thrilled about using `certifi` as a default given us trying to move away from it and to instead rely on system certificates. We use system certificates by default on Linux and kinda on Windows and not on macOS iirc. Not certain what to do about this since whatever we choose we're supporting for all of 2.x.

sethmlarson (changes requested):
Thanks for this change, I have some comments and questions.

sethmlarson:
Also `DEFAULT_POOL` could either be a **Pool**Manager or Connection**Pool** when looking at the name. Also PoolManager isn't exactly "configurable" past initialization, it's making me think we actually want it to be private? @shazow @pquentin?

docs/user-guide.rst
shazow: At the very least, if people find scenarios where this breaks "in production" in unexpected ways then I'd like to consider it a bug.

shazow:
@sethmlarson I don't feel very strongly about it. It's not like the name is a lie, it is indeed a pool! The kind of pool that we choose by default is an implementation/typing detail.

I'm onboard with starting more conservative though, `_DEFAULT_POOL` is fine with me.

pquentin:
There are good arguments either way.

Regarding CI, the tests actually passed (https://github.com/urllib3/urllib3/actions/runs/538104753) but GitHub is telling me that all checks are still expected on this PR. 🤦

docs/user-guide.rst
franekmagiera: Removed the recommendation to not use it in production.

src/urllib3/__init__.py
franekmagiera: Right, makes more sense, it is better to get an error when passing keyword arguments instead of having them silently dropped. Removed the ``**urlopen_kw`` from the signature.

src/urllib3/__init__.py
franekmagiera: Initially I used certifi because of the #1993 description, but I think it would be better to use system certificates by default. Removed certifi.

franekmagiera:
Thanks for the responses. Personally, I feel `_DEFAULT_POOL` is a natural choice for this PoolManager instance, although maybe not entirely reflecting what object it really is.

I made the suggested changes.

Also, would it make sense to have some kind of reload method for the `_DEFAULT_POOL`? Given that there is no access to this instance maybe having a method that would call ``__exit__`` and reinitialized the `_DEFAULT_POOL` could come in handy?

shazow:
Looking good to me!

>  Given that there is no access to this instance...

Does recent Python actually enforce limiting access to underscore-prefixed namespaces? I haven't Python'd in a while but back in the day it was merely a convention, we still had full access to everything.

In either case, IMO let's err on the side of a smaller API surface until it's clear that we need more.

docs/user-guide.rst
franekmagiera: Also added a note about system CA certificates being loaded by default.

franekmagiera:
Right, my mistake. It will only be omitted in case of `import *`.

sethmlarson (approved):
Did some minor changes manually, thanks for this change. Will merge once green!

pquentin:
@franekmagiera Thank you! Would you like to continue working on urllib3? Are there areas you're interested in specifically?

shazow:
@franekmagiera Also thank you from me! This has been something I felt urllib3 needed for a long time, I really appreciate you diving in and making it happen. :)

franekmagiera:
Thanks for support from your side ;)

Sure, I would be happy to tackle another issue. There is nothing specific I'm interested in. I guess most things related to v2 would be ok. Do you have any suggestions?

sethmlarson:
We have all the issues for v2 in [this milestone](https://github.com/urllib3/urllib3/milestone/6), if you're interested in type hints then working on https://github.com/urllib3/urllib3/issues/1897 incrementally by adding types to a single module would be a good issue.

PR #2151, opened 2021-02-05
Improve `Retry` warning

xmo-odoo:
By default, `warnings.warn` will point to the location of the `warnings.warn` call e.g. (with `logging.captureWarnings(True)`)
```
WARNING test py.warnings: /home/odoo/.pyenv/versions/odoo39/lib/python3.9/site-packages/urllib3/util/retry.py:252: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
  warnings.warn(
```

For this specific warning it's not useful, the issue really is with the whoever *calls* `Retry`, so that's where we want the warning message to point us: for the case above it would have pointed directly into [firebase-admin](https://github.com/firebase/firebase-admin-python/issues/515).

### DIFF
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -253,6 +253,7 @@ class Retry(object):
                 "Using 'method_whitelist' with Retry is deprecated and "
                 "will be removed in v2.0. Use 'allowed_methods' instead",
                 DeprecationWarning,
+                stacklevel=2,
             )
             allowed_methods = method_whitelist
         if allowed_methods is _Default:


xmo-odoo:
Aside: I noticed there were various other warnings triggered in the codebase, would it be worth going them (or at least the other `DeprecationWarning`) to evaluate whether they could benefit from a different stacklevel?

pquentin (approved):
Thanks! This is better.

And yes, making sure we always use the correct stacklevel would be useful.

pquentin:
@sethmlarson Can you please user your superpowers to merge this? The failures are unrelated, and I currently prefer using my limited time on v2 than fixing the build here

pquentin:
@sethmlarson ping :)

sethmlarson:
Sorry about that @pquentin, merging now!

PR #2154, opened 2021-02-10
Added management for unknown pool kwargs

TanyaEleventhGoddess:
As the title says, this change prevents errors resulting from changes in supported pool arguments.
If another script or library which depends on urllib3 supplies a deprecated or removed argument, it'll be simply ignored and removed

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -309,5 +309,8 @@ In chronological order:
 * [Franciszek Magiera] <https://github.com/franekmagiera>
   * Add top-level request method.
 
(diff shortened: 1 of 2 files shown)


pquentin:
@sethmlarson What are your thoughts on this idea? I guess the main downside is potentially giving a fall sense of security/functionality, but it can make supporting multiple urllib3 versions easier

sethmlarson:
@pquentin I think this is a good idea, especially if we are going to be adding new parameters in the future. It's a shame that any sort of warning mechanism might be deemed too noisy here. I guess being explicit with expanded types at least gives some indication.

sethmlarson:
Going to close since we haven't seen any movement here.

PR #2159, opened 2021-02-12
Remove always True condition

pquentin:
Now that we only support Python 3.6+, check_hostname always exists.


sethmlarson (approved):
LGTM, thanks!

PR #2161, opened 2021-02-15
Add types to reraise util.

hramezani:
This patch adds types to `reraise` function.

### DIFF
--- a/src/urllib3/util/util.py
+++ b/src/urllib3/util/util.py
@@ -1,4 +1,5 @@
-from typing import Optional, Union
+from types import TracebackType
+from typing import NoReturn, Optional, Type, Union
 
 
 def to_bytes(
@@ -25,7 +26,11 @@ def to_str(
     return x.decode()
 
 
-def reraise(tp, value, tb=None):
+def reraise(
+    tp: Optional[Type[BaseException]],
+    value: Optional[BaseException],
+    tb: Optional[TracebackType] = None,
+) -> NoReturn:
     try:
         if value is None:
             value = tp()


src/urllib3/util/util.py
```diff
@@ -25,7 +26,7 @@ def to_str(
     return x.decode()
 
 
-def reraise(tp, value, tb=None):
+def reraise(tp: type, value: Exception, tb: Optional[TracebackType] = None) -> NoReturn:
```
pquentin, line 29: This function comes from six, and [typeshed uses the following types](https://github.com/python/typeshed/blob/0b1cd5989669544866213807afa833a88f649ee7/stubs/six/six/__init__.pyi#L93-L95):


```python3
def reraise(
    tp: Optional[Type[BaseException]], value: Optional[BaseException], tb: Optional[types.TracebackType] = ...
) -> NoReturn: ...
```

Maybe we should just copy that?

hramezani: Thanks @pquentin for the suggestion. I will change it

pquentin (approved):

pquentin:
The tests passed (https://github.com/urllib3/urllib3/actions/runs/571065039) but GitHub does not see it. Closing/reopening to try fixing this mess

pquentin:
I don't understand why it says "waiting for status".

sethmlarson:
I also don't understand, from the actions screen the PR is passing. I'm going to merge this. @hramezani have you updated your fork with the change to `master` -> `main`

hramezani:
> @hramezani have you updated your fork with the change to `master` -> `main`

Yes, I did.

Issue #2158, opened 2021-02-12
Is HTTPHeaderDict a public API to make requests?


pquentin:
`HTTPHeaderDict` was initially designed to handle response headers, and is not documented for request headers.

 * Should it be documented?
 * How should it be imported? The current options are `from urllib3.response import HTTPHeaderDict` and `from urllib3._collections import HTTPHeaderDict`, and they don't feel right.

sigmavirus24:
The question is "should it be documented for request headers?" and if so, "Why?"

pquentin:
@sigmavirus24 Yes, that's my first bullet point. The context is https://github.com/urllib3/urllib3/pull/2145 where we're catering to the request use case.

sigmavirus24:
I mistyped. I meant should it be used for request headers by users

pquentin:
@sigmavirus24 Sorry, I'm still confused as it still looks like we're both asking the same question!

I don't have a strong opinion on the answer, and am wondering if anyone has an opinion on this minor issue.

sigmavirus24:
My 2¢ is that constructing a HTTPHeaderDict should be the only way to work with headers because basic dictionaries are deceptively simple.

sethmlarson:
My vote is that `HTTPHeaderDict` should be support for request headers, it's already happening in the wild and like @sigmavirus24 said it's actually a good way to work with headers in all their forms.

sethmlarson:
I think this has been answered with "Yes". Should we take the next step and make `HTTPHeaderDict` into a top-level object or leave things as-is?

pquentin:
@sethmlarson yeah, I think a good action here would be to allow `from urllib3 import HTTPHeaderDict` and document it for requests in the docs.

ubdussamad:
Hi @pquentin  , would adding HTTPHeaderDict as a top level object only involve adding `from .response import HTTPHeaderDict` and appending `HTTPHeaderDict` to the __all__ in the file [_init_.py](https://github.com/urllib3/urllib3/blob/main/src/urllib3/__init__.py)?

pquentin:
@ubdussamad Yes! But import it from its original location, `_collections`.

ubdussamad:
@pquentin  For the documentation, would including usage exmaples of  `HTTPHeaderDict` for setting http request headers be enough?
Also, [user-guide](https://github.com/urllib3/urllib3/blob/main/docs/user-guide.rst) is the file I am supposed to edit (for documentation) right?

[...]

PR #2162, opened 2021-02-15
1897 add type hints connection

franekmagiera:
Related to #1897 
Added type hints to `connection.py` and `util/connection.py`.
I also added some type hints for `ssl_.py`, `wait.py` and `proxy.py`, but only for the functions that are used in `connection.py`. I can complete the rest later or as a part of another PR, I didn't want to do it now because it already feels like there are a lot of changes here and wanted to get some feedback first.

There are still some issues, will comment the code shortly.

Also, I used `Mapping[str, str]` for `headers`. I saw recent discussions about using `HTTPHeaderDict` for requests. I think it would be ok to leave headers as `Mapping[str, str]` for connections and let some upper layer class (like ConnectionPool or PoolManager) worry about transforming `HTTPHeaderDict` to `Mapping[str, str]` - what do you think about that?


src/urllib3/connection.py
```diff
@@ -80,31 +108,57 @@ class HTTPConnection(_HTTPConnection):
...
+            )
 
-    @property
-    def host(self):
+    # https://github.com/python/mypy/issues/4125
+    # Mypy treats this as LSP violation, which is considered a bug.
+    # If `host` is made a property it violates LSP, because a writeable attribute is overriden with a read-only one.
+    # However, there is also a `host` setter so LSP is not violated.
+    # Potentailly, a `@host.deleter` might be needed depending on how this issue will be fixed.
+    @property  # type: ignore
```
franekmagiera, line 160: I think it is a mypy bug, hence `type: ignore`

src/urllib3/connection.py
```diff
@@ -160,22 +211,28 @@ def _new_conn(self):
```
franekmagiera, line 211: Line 210 - `NewConnectionError` expects a `ConnectionPool` - should it then be raised by `ConnectionPool` or this exception should be modified?

src/urllib3/connection.py
```diff
@@ -198,22 +257,37 @@ def putheader(self, header, *values):
...
+        method: str,
+        url: str,
+        body: Optional[_DefaultType] = None,
+        headers: Mapping[str, str] = {},
+        *,
+        encode_chunked: bool = False,
+    ) -> None:
+        # Avoid modifying the headers passed into .request()
+        headers = copy(headers)
+        # What about HTTPHeaderDict?
```
franekmagiera, line 271: Forgot to delete that comment ;)

src/urllib3/connection.py
```diff
@@ -422,17 +521,20 @@ def connect(self):
             cert = self.sock.getpeercert()
             _match_hostname(cert, self.assert_hostname or server_hostname)
```
franekmagiera, line 522: `self.assert_hostname` is of type `Optional[Union[str,bool]]`. Should users be allowed to pass `True` to `assert_hostname`? If so, `_match_hostname` should then be called with `server_hostname`? The other option I see is for `assert_hostname` to be of type `Optional[Union[Literal['False'], str]]`.

src/urllib3/connection.py
```diff
@@ -482,11 +584,11 @@ def _match_hostname(cert, asserted_hostname):
         )
         # Add cert to exception and reraise so client code can inspect
         # the cert when catching the exception, if they want to
-        e._peer_cert = cert
+        e._peer_cert = cert  # type: ignore
```
franekmagiera, line 587: Should I create a new Exception for this type? Like
```
MatchHostnameError(CertificateError):
    _peer_cert: _PeerCertRetType
```
?

src/urllib3/util/connection.py
```diff
@@ -1,11 +1,15 @@
 import socket
+from typing import List, Optional, Tuple
 
 from urllib3.exceptions import LocationParseError
 
 from .wait import wait_for_read
 
+SOCKET_GLOBAL_DEFAULT_TIMEOUT = socket._GLOBAL_DEFAULT_TIMEOUT  # type: ignore
```
franekmagiera, line 8: mypy returns an error - it cannot find socket._GLOBAL_DEFAULT_TIMEOUT, hence `type: ignore`. Should urllib3's own GLOBAL_DEFAULT_TIMEOUT be defined?

src/urllib3/util/connection.py
```diff
@@ -1,11 +1,15 @@
...
+SocketOptions = List[Tuple[int, int, int]]
```
franekmagiera, line 9: Should this be restricted to tuples of ints or the last value could also be str/bytes?

src/urllib3/util/ssl_.py
```diff
@@ -109,6 +114,9 @@ def assert_fingerprint(cert, fingerprint):
         Fingerprint as string of hexdigits, can be interspersed by colons.
     """
 
+    if cert is None:
```
franekmagiera, line 117: Probably will have to add a test for that.
`cert` can be None because `getpeercert` can return None.

src/urllib3/exceptions.pyi
```diff
@@ -18,6 +18,7 @@ class RequestError(PoolError):
 
 class SSLError(HTTPError): ...
```
sethmlarson, line 19: Types PRs should remove the `.pyi` files

src/urllib3/connection.py
```diff
@@ -3,23 +3,42 @@
...
+    Tuple,
+    Union,
+    cast,
+)
 
 from .util.proxy import create_proxy_ssl_context
 from .util.util import to_str
 
-try:  # Compiled with SSL?
+if TYPE_CHECKING:
```
sethmlarson, line 30: Isn't `BaseSSLError` required for being imported elsewhere? If so we can't have `TYPE_CHECKING` bounding it

src/urllib3/connection.py
```diff
@@ -55,6 +74,14 @@ class BaseSSLError(BaseException):
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 
 
+_DefaultType = Union[bytes, IO[Any], Iterable[bytes], str]  # Type for HTTP body.
```
sethmlarson, line 77: Let's have a more descriptive global here for HTTP body type

src/urllib3/connection.py
```diff
@@ -80,31 +107,59 @@ class HTTPConnection(_HTTPConnection):
       Or you may want to disable the defaults by passing an empty list (e.g., ``[]``).
     """
 
-    default_port = port_by_scheme["http"]
-
-    #: Disable Nagle's algorithm by default.
-    #: ``[(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]``
-    default_socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
```
sethmlarson, line 87: `default_socket_options` shouldn't be removed

src/urllib3/connection.py
sethmlarson: Has this bug been reported to `mypy`? If not might be something to report and get their thoughts on.

src/urllib3/connection.py
```diff
@@ -198,22 +258,41 @@ def putheader(self, header, *values):
...
             )
 
-    def request(self, method, url, body=None, headers=None):
+    def request(
+        self,
+        method: str,
+        url: str,
+        body: Optional[_DefaultType] = None,
+        headers: Optional[Mapping[str, str]] = None,
+        *,
```
sethmlarson, line 267: The `encode_chunked` parameter wasn't available before, I don't think we should add it.

src/urllib3/connection.py
```diff
@@ -230,22 +309,32 @@ def request_chunked(self, method, url, body=None, headers=None):
...
-                len_str = hex(len(chunk))[2:]
-                to_send = bytearray(len_str.encode())
-                to_send += b"\r\n"
-                to_send += chunk
-                to_send += b"\r\n"
-                self.send(to_send)
+                self.send_chunk(body)
+            else:
+                for chunk in body:
+                    self.send_chunk(chunk)
```
sethmlarson, line 315: Is this refactor necessary? Typically with type hint additions we don't make code changes if possible.

src/urllib3/connection.py
```diff
@@ -230,22 +309,32 @@ def request_chunked(self, method, url, body=None, headers=None):
...
+        to_send = bytearray(len_str.encode())
+        to_send += b"\r\n"
+        to_send += chunk
+        to_send += b"\r\n"
+        self.send(to_send)
+
+
+_PCTRTT = Tuple[Tuple[str, str], ...]
+_PCTRTTT = Tuple[_PCTRTT, ...]
+_PeerCertRetDictType = Dict[str, Union[str, _PCTRTTT, _PCTRTT]]
```
sethmlarson, line 335: We can maybe use `typing_extensions` and `TypedDict(..., total=False)` here to be more exact about the minimum types that come from `.getpeercert()`?

src/urllib3/connection.py
```diff
@@ -420,19 +524,22 @@ def connect(self):
             # the TLS library, this cannot always be done. So we check whether
             # the TLS Library still thinks it's matching hostnames.
             cert = self.sock.getpeercert()
-            _match_hostname(cert, self.assert_hostname or server_hostname)
+            _match_hostname(cert, self.assert_hostname or server_hostname)  # type: ignore
```
sethmlarson, line 527: Why is `type: ignore` needed here?

src/urllib3/connection.py
sethmlarson: I think leaving the `# type: ignore` here for now is fine.

src/urllib3/util/connection.py
sethmlarson: The type for `socket.socket.setsockopt()` appears to be `Revealed type is 'def (self: socket.socket, level: builtins.int, optname: builtins.int, value: Union[builtins.int, builtins.bytes])` so do `int/bytes` for last type.

src/urllib3/util/proxy.py
```diff
@@ -35,8 +40,12 @@ def connection_requires_http_tunnel(
 
 
 def create_proxy_ssl_context(
-    ssl_version, cert_reqs, ca_certs=None, ca_cert_dir=None, ca_cert_data=None
-):
+    ssl_version: Optional[int] = None,
+    cert_reqs: Optional[int] = None,
+    ca_certs: Optional[str] = None,
+    ca_cert_dir: Optional[str] = None,
+    ca_cert_data: Optional[str] = None,
```
sethmlarson, line 47: Can `ca_cert_data` be of type `bytes`? I'm not sure

src/urllib3/util/ssl_.py
```diff
@@ -33,32 +35,35 @@ def _is_ge_openssl_v1_1_1(
     )
 
 
-try:  # Do we have ssl at all?
+if TYPE_CHECKING:
```
sethmlarson, line 38: This `if` statement can be independent from the `try-except` below I believe? Basically we still want the `from ssl import ...` to happen even when type checking

test/with_dummyserver/test_connectionpool.py
```diff
@@ -313,10 +313,10 @@ def test_socket_options(self):
                 s.close()
 
     def test_disable_default_socket_options(self):
-        """Test that passing None disables all socket options."""
+        """Test that passing empty list disables all socket options."""
         # This test needs to be here in order to be run. socket.create_connection actually tries
         # to connect to the host provided so we need a dummyserver to be running.
-        with HTTPConnectionPool(self.host, self.port, socket_options=None) as pool:
+        with HTTPConnectionPool(self.host, self.port, socket_options=[]) as pool:
```
sethmlarson, line 319: Since it seems we're supporting `None` as well can we instead parametize this test for both `None` and `[]`?

sethmlarson (changes requested):
Thanks for this, lots of improvements! Here's a bundle of comments for you :)

src/urllib3/connection.py
franekmagiera: It's been an issue for some time now - https://github.com/python/mypy/issues/4125

src/urllib3/connection.py
franekmagiera: I think it is necessary here, because in the `if isinstance(body, (str, bytes))` branch mypy knows `(body,)` is `Tuple[Union[bytes, str]]` and doesn't allow the `body = (body,)` assignment as http body is `HTTPBody = Union[bytes, IO[Any], Iterable[bytes], str]`.

src/urllib3/connection.py
franekmagiera: Yes, it was added to `http.client` in `3.6`. If it is not added then `request` method violates LSP because of incompatible signatures. I can't think of a workaround that wouldn't require `# type: ignore` or a refactor. Having two ways to do the same thing is also not ideal. What about refactoring the `request` method to do all the work including sending chunked requests and then making `request_chunked` a convenience method for compatibility that would call `request`?

src/urllib3/connection.py
franekmagiera: This is because `self.assert_hostname` could be `True`. In this case should `_match_hostname` be called with `server_hostname`? Or `assert_hostname` could be of type `Optional[Union[Literal['False'], str]]` if `typing_extensions` are added, so 'Literal` is available.

src/urllib3/util/proxy.py
franekmagiera: Right, typeshed uses `cadata: Union[Text, bytes, None] = ...`

src/urllib3/connection.py
franekmagiera: Right, it is required by `connectionpool` and `response`. I assumed that type checking won't be performed in environments where there is no SSL, in that case should I just ignore everything in the `except` branch? Also, there are some issues regarding conditional imports in mypy https://github.com/python/mypy/issues/1153

src/urllib3/connection.py
franekmagiera: Right, got confused about that. Now I see that it is shared across all instances.

src/urllib3/connection.py
sethmlarson: I think this is a case where we want to violate LSP as our API in the future is different from `http.client.HTTPConnection`'s API and our subclassing for now is incidental (see BaseHTTPConnection). I'm comfortable having a `type: ignore` along with a comment about us intentionally violating LSP

src/urllib3/connection.py
```diff
@@ -198,22 +258,41 @@ def putheader(self, header, *values):
...
+        headers: Optional[Mapping[str, str]] = None,
+        *,
+        encode_chunked: bool = False,
+    ) -> None:
         if headers is None:
             headers = {}
         else:
             # Avoid modifying the headers passed into .request()
-            headers = headers.copy()
+            headers = copy(headers)
```
sethmlarson, line 274: Can we avoid using `copy.copy()`? Why does mypy balk at `headers = headers.copy()`?

src/urllib3/connection.py
sethmlarson: Can `HTTPBody` be expanded to also have `Iterable[str]` and then does the previous implementation work?

src/urllib3/connection.py
sethmlarson: Gotcha, related to https://github.com/urllib3/urllib3/issues/2126. And yeah let's use `Literal['False']` for now.

src/urllib3/connection.py
franekmagiera: Makes sense. Will add comments and type ignores.

By the way, do I understand it correctly that in the future when `urllib3.HTTPConnection` will inherit from `BaseHTTPConnection` it would have to contain `http.client.HTTPConnection` (namely, the relationship will be composition instead of inheritance)?

src/urllib3/connection.py
franekmagiera: This is because `headers` are of type `Mapping[str, str]` and `Mapping` does not have a `copy` method. Changing `headers` type would again violate LSP, so the same situation as with the `code_enchunked` discussion. So, probably makes sense to add comments and type ignores again and make headers to be `dict` for now? This also raises the question if in the `BaseHTTPConnection.request` should take headers as `dict` or `HTTPHeaderDict` (or union of both, which seems best I think). Related to #2158

src/urllib3/connection.py
sethmlarson: @franekmagiera Things will remain how they are for 2.x and then in 3.x we'll be using a different HTTP implementation `h2` or `h11` instead of only `http.client.HTTPConnection`. `BaseHTTPConnection` can be used as a parent class for the new HTTP connections.

franekmagiera:
@sethmlarson Thanks for doing the review! I made some changes, will address the rest tomorrow.

src/urllib3/util/connection.py
franekmagiera: Changed to `List[Tuple[int, int, Union[int, bytes]]]`

src/urllib3/util/ssl_.py
franekmagiera: `getpeercert` returns `None` only for server SSL sockets, so I added a test that directly tests the `assert_fingerprint` function in `test_ssl.py`

src/urllib3/connection.py
franekmagiera: Changed to `HTTPBody`

src/urllib3/connection.py
franekmagiera: Added the comment and type ignore for the `request` method.

src/urllib3/connection.py
franekmagiera: I changed the type of `request_chunked`'s `body` parameter to be `Union[None, HTTPBody, Tuple[Union[bytes, str]]` and brought back original implementation.

src/urllib3/connection.py
franekmagiera: Removed type ignore and changed `self.assert_hostname` type to include `Literal[False]` instead of `bool`

test/with_dummyserver/test_connectionpool.py
franekmagiera: Parameterized the test for `None` and `[]`

src/urllib3/connection.py
franekmagiera: Do you have something like:
```
class _PeerCertRetDictType(TypedDict, total=False):
    issuer: _PCTRTTT
    notAfter: str
    notBefore: str
    serialNumber: str
    subject: _PCTRTTT
    subjectAltName: _PCTRTT
    version: str
```
in mind?

src/urllib3/connection.py
```diff
@@ -23,6 +23,8 @@
     cast,
 )
 
+from typing_extensions import Literal
```
sethmlarson, line 26: Can't have `typing_extensions` be top-level, instead use `if typing.TYPE_CHECKING: from typing import Literal` and then string literals `"Literal[False]"`

franekmagiera: Right, forgot its just a dev dependency... Changed that.

KapJI:
@sethmlarson any updates on this one? Looking forward to see fully typed `urllib3` 🙂

src/urllib3/util/connection.py
sethmlarson: Nah, using a type ignore for this case is fine.

sethmlarson (approved):
LGTM, thank you for all of this work!!! Sorry it took so long for me to review, these days are busy :)

PR #2167, opened 2021-03-08
Bring coverage back to 100%

pquentin:
All calls to reraise() are in branches where value is truthy, so we can't reach that code. (I don't explain why this only showed up in coverage when we... added types to reraise, though.)

It seemed better than testing a case that we never hit in real code. (I guess that not using reraise at all would be better, but that seems difficult to achieve.)


pquentin:
The coverage *is* back to 100%, and the CI failure is https://github.com/urllib3/urllib3/issues/2171

sethmlarson (approved):
LGTM

PR #2169, opened 2021-03-08
Improve `assert_hostname` and `server_hostname` documentation

pquentin:
The default value of `assert_hostname` is `server_hostname`, so specifying `assert_hostname` for SNI isn't needed. Also, even if the IP of badssl.com probably isn't that stable, being able to actually run the example seems nicer than using a private IP.

Having a separate example for `assert_hostname` clarifies that it's not only about SNI. Again, I've used badssl.com since that allows to actually run the example.

Fixes https://github.com/urllib3/urllib3/issues/2126 (the type documentation will be handled by https://github.com/urllib3/urllib3/pull/2162)

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -232,14 +232,13 @@ you could connect to a server by IP using HTTPS like so::
 
     >>> import urllib3
     >>> pool = urllib3.HTTPSConnectionPool(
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Great example with `badssl.com`, thanks for this!

PR #2170, opened 2021-03-08
Fix last place that allowed a missing check_hostname

pquentin:
This is a followup of https://github.com/urllib3/urllib3/pull/2159


sethmlarson (approved):
LGTM

PR #2174, opened 2021-03-10
Bring coverage back to 100%

pquentin:
The real reason for this pull request is that we forgot to mark our `Rlock` as not covered, but I also used this opportunity to switch to a different `.coveragerc` string.

Indeed, it's true that `threading` support is "platform-specific" since it's optional in Python 3.6, but since it's always available in Python 3.7 (https://docs.python.org/3/library/threading.html), using "Python 3.6" will remind us to remove the shim when we drop support for Python 3.6 next year.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 else:
     try:
         from threading import RLock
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
LGTM! Didn't know Python 3.7 made the module non-optional, TIL :)

PR #2176, opened 2021-03-10
Always pass server_hostname to wrap_socket

pquentin:
Since Python 3.5, wrap_socket always allows a server_hostname to bepassed, even if OpenSSL does not have SNI. This is how our pyOpenSSL and SecureTransport backends work too.


sethmlarson (approved):
LGTM, this is a great simplification!

PR #2177, opened 2021-03-12
Move "no SNI if IP address" responsibilty to TLS backends

pquentin:
* The ssl module already avoids SNI when the host is an IP address
 * pyOpenSSL now does that too
 * We were already using SNI on IP addresses with SecureTransport anyway

In other words, this does not change anything, but is cleaner
(we no longer test for SecureTransport in ssl.py) and will allow us to
lean on ssl.SSLContext to match hostnames.


src/urllib3/util/__init__.py
```diff
@@ -11,6 +11,7 @@
     PROTOCOL_TLS,
     SSLContext,
     assert_fingerprint,
+    is_ipaddress,
```
sethmlarson, line 14: `urllib3.util` are public APIs, are we okay with having this function be public?

sethmlarson (approved):
The change itself looks good, great to simplify things! One thought:

src/urllib3/util/__init__.py
pquentin: Thanks, fixed! I hadn't realized that `util` was more public than `util.ssl_`. I certainly don't want to add more public APIs

pquentin:
The codecov check is red, but the coverage is still 100%: https://app.codecov.io/gh/urllib3/urllib3/compare/2177/commits

pquentin:
Was merged as part of #2178 by mistake, closing

PR #2178, opened 2021-03-12
Lean on SSLContext to verify hostnames when possible

### LINKED ISSUE #517: Lean on SSLContext to do HTTPS hostname verification
In Python 3.4 and 2.7.9 there is a `check_hostname` attribute on `SSLContext` which will have the SSLContext instance handle checking the hostname inside of the `do_handshake`. I think it would be good for urllib3 to rely on this where possible instead of doing the check itself. I think this would go good with the other efforts to use `SSLContext` as the "bag of configuration" for TLS stuff.

This can be detected by determining if the `SSLContext` object has a `check_hostname` attribute or not.

There is one downside, this relies on passing the hostname as part of `SSLContext().wrap_socket(server_name=)`. Originally this only worked if the OpenSSL had SNI enabled. However Python 3.4.3 and 2.7.9 will accept `server_name` even if SNI isn't enabled.

This would mean that the code a little ugly until you drop older versions of Python 3, it would look something like (paraphrased and super simplified to just be the ugly parts):

``` python
import ssl
import sys

# In reality these would use the SSLContext shim that urllib3 has in
# order to work on Python 2.6 too and 2.7 earlier than 

if ssl.HAS_SNI or sys.version_info[0] == 2 or sys.version_info[:3] >= (3, 4, 3):
    ctx = SSLContext(ssl.PROTOCOL_SSLv23)
    ctx.check_hostname = True
    sock = ctx.wrap_socket(sock, server_name="example.com")
else:
    ctx = SSLContext(ssl.PROTOCOL_SSLv23)
    sock = ctx.wrap_socket(sock)

    if not ssl.match_hostname(sock.getpeercert(), "example.com"):
        raise Error()
```

Maybe this is a bad idea? I think it would be great to lay the ground work for eventually moving the responsibility for checking hostnames into the stdlib, however you wouldn't actually be able to do that completely without the ugly if statement until 3.4.3 was your minimum supports Python 3 version.
##

pquentin:
Closes #517 

There are four situations where we can't lean on SSLContext to verify certs for us:

 * Python 3.6, where we need our custom code to reject commonName
 * fingerprints *and* alternate hostnames, as those aren't supported by Python's `ssl`
 * PyOpenSSL where there's no easy way to check for hostnames

So, given the preparatory work done in https://github.com/urllib3/urllib3/pull/2176 and https://github.com/urllib3/urllib3/pull/2177, this PR only adds one commit where we just ask the SSL backend to verify the hostname (SecureTransport was already doing it anyway.)

I made sure that invalid hostnames are still rejected with all three SSL backends.


pquentin:
Ubuntu build timed out, retrying

src/urllib3/util/ssl_.py
```diff
@@ -241,9 +241,10 @@ def create_urllib3_context(
         context.post_handshake_auth = True
 
     context.verify_mode = cert_reqs
-    # We do our own verification, including fingerprints and alternative
-    # hostnames. So disable it here
-    context.check_hostname = False
+    # We ask for verification here but it may be disabled in HTTPSConnection.connect
+    context.check_hostname = cert_reqs == ssl.CERT_REQUIRED
+    if hasattr(context, "hostname_checks_common_name"):
+        context.hostname_checks_common_name = False  # Python 3.7
```
sethmlarson, line 247: Should the `# Python ...` comment be on the `if` line to cover the whole branch?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1351,7 +1351,7 @@ def socket_handler(listener):
         with mock.patch("urllib3.util.ssl_.SSLContext", lambda *_, **__: context):
             self._start_server(socket_handler)
             with HTTPSConnectionPool(self.host, self.port) as pool:
-                with pytest.raises(MaxRetryError):
+                with pytest.raises(Exception):
```
sethmlarson, line 1354: Not a huge fan of asserting `Exception`, means this error could change any time and we wouldn't notice

src/urllib3/connection.py
```diff
@@ -362,6 +362,18 @@ def connect(self):
                 ssl_version=resolve_ssl_version(self.ssl_version),
                 cert_reqs=resolve_cert_reqs(self.cert_reqs),
             )
+            # In some cases, we want to verify hostnames ourselves
+            if (
```
sethmlarson, line 366: Love how this codifies all the situations where we disable `check_hostname` so nicely :star_struck:

sethmlarson (commented):
This is looking pretty good, a few comments for you:

src/urllib3/util/ssl_.py
pquentin: So the intent here is to remind us to remove this when we drop support for Python 3.6 while still making sure it's covered.

Now that I think about it more, the best thing to do is to rely on codecov telling us it's no longer covered when we drop Python 3.6.

test/with_dummyserver/test_socketlevel.py
pquentin: I usually avoid Pokemon exception handling too, but what we really want to test is that `load_default_certs` was called, and this continues to be true with this pull request.

Regarding the exception, we know that without a correct `SSLContext` things are going to fail at some point, and we currently fail with a `TypeError` instead of a `MaxRetryError`. Since we don't care about the actual exception type, I thought using `Exception` was fine.

Would you prefer using `with contextlib.suppress(Exception)`?

sethmlarson: No if all we care about is an exception is raised this is fine.

sethmlarson (approved):
I think all review comments are covered here, merge when ready! :tada:

pquentin:
Thanks! Merged

PR #2179, opened 2021-03-12
Use exact same SNI logic in pyOpenSSL and SecureTransport

pquentin:
This pull request builds on https://github.com/urllib3/urllib3/pull/2178 and adds a new dependency to match hostnames inside the pyOpenSSL backend. Hopefully, by the time we drop support for OpenSSL 1.0.2, this will be possible using pyOpenSSL directly.


pquentin:
macOS build timeoud, retrying

src/urllib3/contrib/pyopenssl.py
```diff
@@ -456,21 +457,33 @@ def wrap_socket(
...
+                server_hostname_bytes = server_hostname_bytes.encode("utf-8")
+            cnx.set_tlsext_host_name(server_hostname_bytes)
 
         cnx.set_connect_state()
 
         while True:
             try:
                 cnx.do_handshake()
+                if self.check_hostname:
+                    if util.is_ipaddress(server_hostname):
```
sethmlarson, line 473: What are we gaining by using `service-identity` package over our stricter verification of IP addresses from the standard library which only checks equality?

Since we're keeping around `match_hostname()` to disable commonName checking on Python 3.6 can we use `match_hostname()` here instead of bringing on a new dependency?

sethmlarson (commented):
Couple of questions about this change:

src/urllib3/contrib/pyopenssl.py
pquentin: Right, it would be much better to use our own code here, thanks!

pquentin:
> Couple of questions about this change

Did you have another question? I'm only seeing one

PR #2181, opened 2021-03-15
extend LifoQueue from python

dulmandakh:
This PR changes LifoQueue to extend from [queue.LifoQueue](https://github.com/python/cpython/blob/35d5068928ab5485e5f28b60b1e33062bc2c46cc/Lib/queue.py#L242), which uses list as queue internally. LifoQueue was introduced in 2.7 with https://github.com/python/cpython/commit/35641461db7c692877ffa4bbbfe31a525c81770e.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -37,7 +37,6 @@ from .request import RequestMethods
 from .response import BaseHTTPResponse, HTTPResponse
 from .util.connection import is_connection_dropped
 from .util.proxy import connection_requires_http_tunnel
-from .util.queue import LifoQueue
 from .util.request import set_file_position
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks!

pquentin (changes requested):
However, now that I think about it more, we don't need a `deque` for a LIFOQueue, and can remove this file to just use `LifoQueue` from Python directly

sethmlarson:
Made some changes myself, let me know what you think @pquentin

pquentin (approved):
Thanks!

sethmlarson:
Thanks to @dulmandakh for making us aware of this improvement :bow:

PR #2186, opened 2021-03-16
Actually reject common names

pquentin:
We were using Python's match_hostname by default, who accepts them.

The next step is to move ssl_match_hostname from packages to util, but that would make a nice first issue so I left it out.


test/with_dummyserver/test_https.py
```diff
@@ -779,6 +779,18 @@ class TestHTTPS_TLSv1_3(TestHTTPS):
     certs = TLSv1_3_CERTS
 
 
+class TestHTTPS_NoSAN:
```
sethmlarson, line 782: :tada:

sethmlarson (approved):
LGTM

PR #2188, opened 2021-03-16
Stop relying on stdlib for hostname matching

pquentin:
I can't make hostname_checks_common_name work, either locally or on GitHub Actions, so stop using the standard library to match hostnames for now.

https://bugs.python.org/issue43522 will hopefully help me understand what I'm missing.

cc @tiran

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -370,8 +370,10 @@ class HTTPSConnection(HTTPConnection):
                 # We still support OpenSSL 1.0.2, which prevents us from verifying
                 # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
                 or ssl_.IS_PYOPENSSL
-                # Python 3.6 can't disable commonName checks
-                or not hasattr(self.ssl_context, "hostname_checks_common_name")
+                # context.hostname_checks_common_name seems ignored, and it's more
+                # important to reject certs without SANs than to rely on the standard
+                # libary. See https://bugs.python.org/issue43522 for details.
+                or True
             ):
                 self.ssl_context.check_hostname = False
 


pquentin:
Ubuntu 3.7 failed on `OSError: [Errno 98] Address already in use`, retrying.

src/urllib3/connection.py
```diff
@@ -370,8 +370,10 @@ def connect(self):
                 # We still support OpenSSL 1.0.2, which prevents us from verifying
                 # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
                 or ssl_.IS_PYOPENSSL
-                # Python 3.6 can't disable commonName checks
-                or not hasattr(self.ssl_context, "hostname_checks_common_name")
```
tiran, line 374: The original code is wrong. The feature depends on Python version and OpenSSL version.

```
if getattr(ssl, "HAS_NEVER_CHECK_COMMON_NAME", False):
    ctx.hostname_checks_common_name = False
```

This changes the internal attribute ``ctx._host_flags`` from ``4`` to ``36`` (``HOSTFLAG_NO_PARTIAL_WILDCARDS|HOSTFLAG_NEVER_CHECK_SUBJECT``). Internally it sets the OpenSSL flag ``X509_CHECK_FLAG_NEVER_CHECK_SUBJECT`` with ``X509_VERIFY_PARAM_set_hostflags()``.

pquentin: Thanks for pointing this out! I saw it earlier today and initially though that it was the only problem, but it's not. I will certainly use HAS_NEVER_CHECK_COMMON_NAME when reverting this.

(I also realize that setting `hostname_checks_common_name` only sets an OpenSSL flag, which is why I believe I'm doing something wrong.)

sethmlarson:
Given the comments in the BPO we'll need to keep `match_hostname` until Python 3.10 is our minimum supported version. Thanks @pquentin and @tiran for tracking this down!

sethmlarson (approved):
LGTM, let's create an issue to track the detection when this feature is fixed.

PR #2193, opened 2021-03-19
Add test for idna encode when idna is not installed.

hramezani:
Add test to cover [ImportError when idna is not installed](https://github.com/urllib3/urllib3/blob/4f4679683074b2808c36b41b50b3ff8de93eb8be/src/urllib3/util/url.py#L296).
Also, I moved `ImportBlocker` and `ModuleStash` to `test/__init__.py` to be usable by other tests.

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -2,9 +2,10 @@ import io
 import logging
 import socket
 import ssl
+import sys
 import warnings
 from itertools import chain
-from test import notBrotli, onlyBrotli
+from test import ImportBlocker, ModuleStash, notBrotli, onlyBrotli
 from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


test/test_util.py
```diff
@@ -889,3 +890,26 @@ def test_ssl_wrap_socket_sni_none_no_warn(self):
         context, warn = self._wrap_socket_and_mock_warn(sock, None)
         context.wrap_socket.assert_called_once_with(sock, server_hostname=None)
         warn.assert_not_called()
+
+
+idna_blocker = ImportBlocker("idna")
+module_stash = ModuleStash("urllib3")
+
+
+class TestWithoutIdna:
```
sethmlarson, line 899: Because we only have on test case that uses this blocker we can put `setup_class` and `teardown_class` on `TestUtilWithoutIdna`. If there's another test case that needs it we can generalize then :)

test/test_util.py
```diff
@@ -889,3 +890,26 @@ def test_ssl_wrap_socket_sni_none_no_warn(self):
...
+        sys.meta_path.insert(0, idna_blocker)
+
+    def teardown_class(cls):
+        sys.meta_path.remove(idna_blocker)
+        module_stash.pop()
+
+
+class TestUtilWithoutIdna(TestWithoutIdna):
+    def test_parse_url_without_idna(self):
+        with pytest.raises(LocationParseError):
```
sethmlarson, line 914: Can we assert the error string for the `LocationParseError`

sethmlarson (changes requested):
This is a good idea to test for the missing `idna` case, a few comments:

hramezani:
Thanks @SethMichaelLarson, I've addressed your comments.

sethmlarson (approved):

PR #2194, opened 2021-03-28
util/request.py: mention ',br' as automatically added parameter

dilyanpalauzov:


pquentin (changes requested):
Thanks! Can you please mention that this happens if a brotli package is installed?

sigmavirus24 (approved):

PR #2197, opened 2021-04-12
raise FullPoolError if we try to _put_conn an extra connection in blocking mode

dmazin:
I have been looking at ConnectionPool's behavior in blocking mode and was confused by EmptyPoolError's message (it said the pool was full, which is sort of true, but needs a bit of explanation, which I added.

I was also confused by the absence of a FullPoolError if you try to _put_conn a conn in blocking mode. I added a FullPoolError.

I also added a couple of tests for blocking mode and one test for non-blocking mode.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -279,23 +279,28 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         """
         try:
             self.pool.put(conn, block=False)
+            return  # Everything is dandy, done.
         except AttributeError:
             # self.pool is None.
-            # Connection never got put back into the pool, close it.
-            conn and conn.close()
+            pass
         except queue.Full:
-            # This should never happen if self.block == True and you got the conn from self._get_conn
 
             # Connection never got put back into the pool, close it.
-            conn and conn.close()
+            if conn:
+                conn.close()
 
             if self.block:
+                # This should never happen if you got the conn from self._get_conn
                 raise FullPoolError(
                     self,
                     "Pool reached maximum size and no more connections are allowed.",
                 )
 
             log.warning("Connection pool is full, discarding connection: %s", self.host)
+
+        # Connection never got put back into the pool, close it.
+        if conn:
+            conn.close()
 
     def _validate_conn(self, conn):
         """
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -19,6 +19,7 @@ from .connection import (
 from .exceptions import (
     ClosedPoolError,
     EmptyPoolError,
+    FullPoolError,
     HeaderParsingError,
     HostChangedError,
     HTTPSProxyError,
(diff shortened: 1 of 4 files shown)


test/test_connectionpool.py
```diff
@@ -223,13 +223,19 @@ def test_max_connections(self):
 
             assert pool.num_connections == 1
 
-    def test_pool_edgecases(self):
+    def test_pool_edgecases(self):  # TODO rename this
```
dmazin, line 226: I think this case is meant to test that if you add 2 conns to a pool with pool.maxsize=1, the 2nd connection will get closed instead of getting put onto the queue.
Is that correct? If so, I'll rename the test case.

sethmlarson: Yeah you can rename this test, thanks

shaoster (changes requested):
This review shouldn't go throw.

src/urllib3/connectionpool.py
```diff
@@ -246,7 +247,7 @@ def _get_conn(self, timeout=None):
             if self.block:
                 raise EmptyPoolError(
                     self,
-                    "Pool reached maximum size and no more connections are allowed.",
+                    "No connection is available, and new connection cannot be created due to blocking mode.",
```
dmazin, line 250: I was very confused by the original error message. How can I be getting EmptyPoolError if the pool has reached maximum size? In reality, what I would tell the user is "hey, you are in blocking mode and you have already checked out all maxsize conns so we cannot create a new conn for you." This is what I changed the message to, except, you know, more formal-sounding.

src/urllib3/connectionpool.py
```diff
@@ -278,17 +279,23 @@ def _put_conn(self, conn):
...
+            # This should never happen if self.block == True and you got the conn from self._get_conn
 
-        # Connection never got put back into the pool, close it.
-        if conn:
-            conn.close()
+            # Connection never got put back into the pool, close it.
+            conn and conn.close()
+
+            if self.block:
+                raise FullPoolError(
```
dmazin, line 293: This block originally had a "This should never happen if self.block == True" comment. If this should never happen then if it *does* happen we should raise an exception.

sethmlarson (changes requested):
I'm always wary of reorganizing our pool management code. How about instead we add the exception within the `except queue.Full:` block and add an extra step to close the connection so we're not changing too much of the existing logic?

dmazin:
> I'm always wary of reorganizing our pool management code. How about instead we add the exception within the `except queue.Full:` block and add an extra step to close the connection so we're not changing too much of the existing logic?

@sethmlarson That makes way more sense than my initial approach. This PR is ready for another look.

dmazin:
@sethmlarson ping <3

test/test_connectionpool.py
```diff
@@ -223,19 +224,66 @@ def test_max_connections(self):
...
+            conn2.close = Mock()
+
+            pool._put_conn(conn1)
+            with pytest.raises(FullPoolError):
+                pool._put_conn(conn2)
+
+            assert conn1.close.called is False
+            assert conn2.close.called is True
+
+            assert conn1 == pool._get_conn()
```
sethmlarson, line 271: Should we do an `is` instead of `==` here?

sethmlarson (changes requested):
Looks mostly good, two thoughts for you:

test/test_connectionpool.py
dmazin: That totally makes sense to me, though other tests in this file use `==` for similar comparisons. Usually I would prefer to follow the convention of the rest of the file and perhaps leave a fixme.

src/urllib3/exceptions.py
```diff
@@ -169,6 +169,12 @@ class EmptyPoolError(PoolError):
     pass
 
 
+class FullPoolError(PoolError):
```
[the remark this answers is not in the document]
dmazin: Not sure I follow :/ Isn't FullPoolError included automatically?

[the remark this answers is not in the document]
sethmlarson: You're right, my bad!

sethmlarson (dismissed):
LGTM

sethmlarson (approved):

sethmlarson:
Thanks @dmazin for your first contribution and being patient through the review cycles :bow: Hope you're interested in contributing more, there's a lot of cool issues in the [v2.0 milestone](https://github.com/urllib3/urllib3/milestone/6). If not that's totally fine, thanks again!

dmazin:
> Thanks @dmazin for your first contribution and being patient through the review cycles :bow: Hope you're interested in contributing more, there's a lot of cool issues in the [v2.0 milestone](https://github.com/urllib3/urllib3/milestone/6). If not that's totally fine, thanks again!

@sethmlarson thank you so much for going out of your way to write such a welcoming message! now THAT'S how you foster community!

I am definitely wondering how to get move involved with 2.0 -- will check out the milestone

PR #2198, opened 2021-04-12
Move ssl_match_hostname to urllib3.utils.

hramezani:


src/urllib3/util/__init__.py
```diff
@@ -1,4 +1,5 @@
 # For backwards compatibility, provide imports that used to be here.
+from . import ssl_match_hostname
```
sethmlarson, line 2: We don't want to make this function available here as `urllib3.util` is public API.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 ignore = E501, E203, W503, W504
-exclude=./docs/conf.py,./src/urllib3/packages/*
+exclude=./docs/conf.py,./src/urllib3/util/ssl_match_hostname/*
```
sethmlarson, line 3: Let's allow linting on this file to simplify things.

sethmlarson (changes requested):
Thanks for this, instead of pulling the whole module over, can we instead move everything to a single file?

hramezani:
@SethMichaelLarson 
We have a mypy validation error 
`src/urllib3/connection.py:591: error: Call to untyped function "match_hostname" in typed context.`
Should I ignore validation for this line?

pquentin:
No, please don't ignore: we need to include the types from https://github.com/urllib3/urllib3/pull/2198/files#diff-e60901ca7e95a57497217ad9b1a4228bbe3319f6d264575a4f667ac349b015c7 inline in the new file, since we're now Python 3.6+. Do you need help with that?

hramezani:
> No, please don't ignore: we need to include the types from https://github.com/urllib3/urllib3/pull/2198/files#diff-e60901ca7e95a57497217ad9b1a4228bbe3319f6d264575a4f667ac349b015c7 inline in the new file, since we're now Python 3.6+. Do you need help with that?

We have the same in [connection.py](https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L330-L333). Should we keep them as is or import them from `src/urllib3/util/ssl_match_hostname.py`

pquentin:
> or import them from src/urllib3/util/ssl_match_hostname.py

That sounds better to me!

noxfile.py
```diff
@@ -14,9 +14,6 @@
     "src/urllib3/_collections.py",
     "src/urllib3/fields.py",
     "src/urllib3/filepost.py",
-    "src/urllib3/packages/__init__.py",
-    "src/urllib3/packages/ssl_match_hostname/__init__.py",
-    "src/urllib3/packages/ssl_match_hostname/_implementation.py",
```
hramezani, line 19: @pquentin Should we keep the new `src/util/ssl_match_hostname.py` in this list?

pquentin: We should yes!

pquentin:
This is looking quite good, but I'd like to fix the Python 3.10 tests (https://github.com/urllib3/urllib3/pull/2201) before doing a thorough review.

noxfile.py
hramezani: Ok, I will add type-hints to `src/util/ssl_match_hostname.py` then I will add the file to the list.

src/urllib3/util/__init__.py
```diff
@@ -37,6 +37,7 @@
     "make_headers",
     "resolve_cert_reqs",
     "resolve_ssl_version",
+    "ssl_match_hostname",
```
sethmlarson, line 40: Want to revert this change as well

src/urllib3/util/ssl_match_hostname.py
```diff
@@ -41,7 +49,7 @@ def _dnsname_match(dn, hostname, max_wildcards=1):
 
     # speed up common case w/o wildcards
     if not wildcards:
-        return dn.lower() == hostname.lower()
+        return bool(dn.lower() == hostname.lower())
```
sethmlarson, line 52: Is this for typing reasons? What does mypy think the result of `str == str` is?

src/urllib3/util/ssl_match_hostname.py
```diff
@@ -77,10 +85,10 @@ def _ipaddress_match(ipname, host_ip):
     # OpenSSL may add a trailing newline to a subjectAltName's IP address
     # Divergence from upstream: ipaddress can't handle byte str
     ip = ipaddress.ip_address(ipname.rstrip())
-    return ip == host_ip
+    return bool(ip == host_ip)
 
 
-def match_hostname(cert, hostname):
+def match_hostname(cert: _PeerCertRetType, hostname: str) -> None:
```
sethmlarson, line 91: Shouldn't `cert` be `_PeerCertRetDictType` here?

sethmlarson (changes requested):
Couple more comments for you

src/urllib3/util/ssl_match_hostname.py
hramezani: I copied it from https://github.com/urllib3/urllib3/pull/2198/files#diff-e60901ca7e95a57497217ad9b1a4228bbe3319f6d264575a4f667ac349b015c7 as @pquentin suggested.

src/urllib3/util/__init__.py
hramezani: Done!

hramezani:
Just rebased with main!

sethmlarson:
@hramezani It looks like we're [missing coverage](https://app.codecov.io/gh/urllib3/urllib3/compare/2198/tree/src/urllib3/util/ssl_match_hostname.py) by moving the file to utils, perhaps we can create a few test cases to ensure `match_hostname` is covered?

hramezani:
> @hramezani It looks like we're [missing coverage](https://app.codecov.io/gh/urllib3/urllib3/compare/2198/tree/src/urllib3/util/ssl_match_hostname.py) by moving the file to utils, perhaps we can create a few test cases to ensure `match_hostname` is covered?

@SethMichaelLarson I added some test to cover them

sethmlarson (approved):
This looks good to me now, once CI passes I'll merge :)

Issue #2192, opened 2021-03-18
Respect SSLContext.hostname_checks_common_name set manually by users


sethmlarson:
If a user sets `SSLContext.hostname_checks_common_name` we should respect their request and check commonName. This provides a fallback we can point to in case users rely on commonName support.

sethmlarson:
Adding `Breaking Changes` solely for being considered a blocker of v2.0

tiran:
The issue is now fixed in OpenSSL and Python has a workaround for previous OpenSSL versions, https://bugs.python.org/issue43522.

You either need

* Python 3.8.9+
* Python 3.9.3+
* Python 3.10

or

* OpenSSL 1.1.1l
* OpenSSL 3.0.0-alpha15

As of today none of the versions are released yet.

PR #2199, opened 2021-04-17
Add type hints for proxy, ssl_, ssltransport and wait modules

franekmagiera:
Continuation of #2162 

Added type hints for modules that `connection.py` uses, i.e. `util/proxy.py`, `util/ssl_.py`, `util/ssltransport.py` and `util/wait.py`.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -14,7 +14,6 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
-    Dict,
     Iterable,
     Mapping,
     NamedTuple,
@@ -51,6 +50,7 @@ from .exceptions import (
 from .packages.ssl_match_hostname import CertificateError, match_hostname
 from .util import SKIP_HEADER, SKIPPABLE_HEADERS, connection, ssl_
 from .util.ssl_ import (
+    PeerCertRetType,
     assert_fingerprint,
     create_urllib3_context,
     resolve_cert_reqs,
@@ -327,12 +327,6 @@ class HTTPConnection(_HTTPConnection):
         self.send(b"0\r\n\r\n")
 
 
-_PCTRTT = Tuple[Tuple[str, str], ...]
-_PCTRTTT = Tuple[_PCTRTT, ...]
-_PeerCertRetDictType = Dict[str, Union[str, _PCTRTTT, _PCTRTT]]
-_PeerCertRetType = Union[_PeerCertRetDictType, bytes, None]
-
-
 class HTTPSConnection(HTTPConnection):
     """
     Many of the parameters to this constructor are passed to the underlying SSL
@@ -586,7 +580,7 @@ class HTTPSConnection(HTTPConnection):
             )
 
 
-def _match_hostname(cert: _PeerCertRetType, asserted_hostname: str) -> None:
+def _match_hostname(cert: PeerCertRetType, asserted_hostname: str) -> None:
     try:
         match_hostname(cert, asserted_hostname)
     except CertificateError as e:
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -221,11 +221,9 @@ def create_urllib3_context(
         Constructed SSLContext object with specified options
     :rtype: SSLContext
     """
-    if SSLContext is not None:
-        context = SSLContext(ssl_version or PROTOCOL_TLS)
-    else:
-        raise TypeError("SSLContext is None - cannot create SSLContext (no ssl module)")
-
+    if SSLContext is None:
+        raise TypeError("Can't create an SSLContext object without an ssl module")
+    context = SSLContext(ssl_version or PROTOCOL_TLS)
     # Unless we're given ciphers defer to either system ciphers in
     # the case of OpenSSL 1.1.1+ or use our own secure default ciphers.
     if ciphers is not None or not USE_DEFAULT_SSLCONTEXT_CIPHERS:
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -5,13 +5,11 @@ import sys
 import warnings
 from binascii import hexlify, unhexlify
 from hashlib import md5, sha1, sha256
-from typing import Optional, Union
+from typing import TYPE_CHECKING, Dict, Optional, Tuple, Union, cast, overload
 
 from ..exceptions import ProxySchemeUnsupported, SNIMissingWarning, SSLError
 from .url import _BRACELESS_IPV6_ADDRZ_RE, _IPV4_RE
 
-SSLContext = None
-SSLTransport = None
 HAS_SNI = False
 IS_PYOPENSSL = False
 IS_SECURETRANSPORT = False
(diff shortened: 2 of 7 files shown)


src/urllib3/connection.py
```diff
@@ -327,12 +327,6 @@ def request_chunked(
         self.send(b"0\r\n\r\n")
 
 
-_PCTRTT = Tuple[Tuple[str, str], ...]
```
franekmagiera, line 330: Moved those to `util/ssl_.py` because both `connection.py` and `util/ssltransport.py` need them.

src/urllib3/util/ssl_.py
```diff
@@ -246,24 +276,62 @@ def create_urllib3_context(
     if (cert_reqs == ssl.CERT_REQUIRED or sys.version_info >= (3, 7, 4)) and getattr(
         context, "post_handshake_auth", None
     ) is not None:
-        context.post_handshake_auth = True
+        context.post_handshake_auth = True  # type: ignore
```
franekmagiera, line 279: As far as I understand this works for versions 3.7.4+, but typeshed includes it only for 3.8+ (SSLContext)

src/urllib3/util/ssl_.py
```diff
@@ -35,7 +33,7 @@ def _is_ge_openssl_v1_1_1(
     )
 
 
-try:  # Do we have ssl at all?
+if TYPE_CHECKING:  # Assume ssl module is there.
```
sethmlarson, line 36: We don't want these imports to be predicated on type checking being active. We should move the `if TYPE_CHECKING` to only cover the `typing_extensions` import.

src/urllib3/util/ssl_.py
```diff
@@ -127,7 +157,7 @@ def assert_fingerprint(cert: Optional[bytes], fingerprint: str) -> None:
 
     if not hmac.compare_digest(cert_digest, fingerprint_bytes):
         raise SSLError(
-            f'Fingerprints did not match. Expected "{fingerprint}", got "{hexlify(cert_digest)}".'
+            f'Fingerprints did not match. Expected "{fingerprint}", got "{hexlify(cert_digest)!r}".'
```
sethmlarson, line 160: `!r` isn't what we want here, we'd receive a message that says `... got "'abcdef'"` (notice the `"` followed by `'`)

src/urllib3/util/ssl_.py
```diff
@@ -251,19 +281,57 @@ def create_urllib3_context(
     context.verify_mode = cert_reqs
     # We ask for verification here but it may be disabled in HTTPSConnection.connect
     context.check_hostname = cert_reqs == ssl.CERT_REQUIRED
-    if hasattr(context, "hostname_checks_common_name"):
+    if sys.version_info >= (3, 7):
```
sethmlarson, line 284: This situation is more complex than this unfortunately (see https://github.com/urllib3/urllib3/issues/2192#issuecomment-821832963) so we can leave this as-is for now.

src/urllib3/util/ssltransport.py
```diff
@@ -89,15 +124,20 @@ def sendall(self, data, flags=0):
...
             raise ValueError("non-zero flags not allowed in calls to send")
-        response = self._ssl_io_loop(self.sslobj.write, data)
-        return response
+        return self._ssl_io_loop(self.sslobj.write, data)
 
     def makefile(
-        self, mode="r", buffering=None, encoding=None, errors=None, newline=None
-    ):
+        self,
+        mode: Union["_TextMode", "_BinaryMode"],
```
sethmlarson, line 134: Since the `socket` object is transparent to users do we need this level of granularity for our type hints (ie is it okay to use `str` here?)

src/urllib3/util/ssltransport.py
```diff
@@ -185,15 +244,45 @@ def _wrap_ssl_read(self, len, buffer=None):
...
+        self,
+        func: Callable[[int, Optional[bytearray]], bytes],
+        arg1: int,
+        arg2: Optional[bytearray],
+    ) -> bytes:
+        ...
+
+    def _ssl_io_loop(
+        self,
+        func: Callable[..., _ReturnValue],
```
sethmlarson, line 269: Nice :clap:

test/test_ssltransport.py
```diff
@@ -18,20 +18,10 @@
 def server_client_ssl_contexts():
     if hasattr(ssl, "PROTOCOL_TLS_SERVER"):
         server_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
-    else:
```
sethmlarson, line 21: Thanks for these removals! :tada:

sethmlarson (changes requested):
This is great, thanks! Here's some comments for you.

franekmagiera:
Thanks for doing the review! Addressed the comments.

src/urllib3/util/ssltransport.py
franekmagiera: `makefile` implementation already checks if mode contains only letters 'b', 'r' and 'w'. I think one thing that could go unnoticed is having duplicates like 'rrw' but it is rather unlikely so for simplicity I think it is ok to use `str`.

src/urllib3/util/ssl_.py
franekmagiera: Right, I see. Changed back to `hasattr` check.

src/urllib3/util/ssl_.py
franekmagiera: Changed to use `bytes.hex()` (introduced in Python 3.5)

src/urllib3/util/ssl_.py
franekmagiera: Refactored it and it uses some `# type: ignore`s for now, not sure if this is ok in this case

sethmlarson (changes requested):
So close! Just one nit and a question :)

src/urllib3/util/ssl_.py
```diff
@@ -210,7 +221,10 @@ def create_urllib3_context(
         Constructed SSLContext object with specified options
     :rtype: SSLContext
     """
-    context = SSLContext(ssl_version or PROTOCOL_TLS)
+    if SSLContext is not None:
+        context = SSLContext(ssl_version or PROTOCOL_TLS)
+    else:
+        raise TypeError("SSLContext is None - cannot create SSLContext (no ssl module)")
```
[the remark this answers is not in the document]
franekmagiera: Right, makes a lot more sense

src/urllib3/util/ssl_.py
```diff
@@ -278,7 +330,7 @@ def ssl_wrap_socket(
     key_password: Optional[str] = None,
     ca_cert_data: Union[None, str, bytes] = None,
     tls_in_tls: bool = False,
-) -> "ssl.SSLSocket":
+) -> Union["ssl.SSLSocket", "SSLTransportType"]:
```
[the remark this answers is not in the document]
franekmagiera: I think the issue is that `SSLTransport` is a variable - at the beginning it is set to `None` and then is possibly overwritten by an import - and is not valid as a type for mypy. I could not think of another approach other than to `import SSLTransport as SSLTransportType` and use it for type checking instead. Mypy returns this link when just `"SSLTransport"` is used: https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases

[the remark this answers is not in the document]
sethmlarson: Huh that's unfortunate, I guess if this makes mypy happy we can stick with it.

sethmlarson (approved):
This is fantastic, thank you for all of this! :rocket:

PR #2202, opened 2021-04-23
Add type hint to src/urllib3/request.py

hramezani:


src/urllib3/request.py
```diff
@@ -78,11 +85,16 @@ def request(
             )
         else:
             return self.request_encode_body(
-                method, url, fields=fields, headers=headers, **urlopen_kw
+                method, url, fields=fields, headers=headers, **urlopen_kw  # type: ignore
```
hramezani, line 88:
```
error: Argument 5 to "request_encode_body" of "RequestMethods" has incompatible type "**Dict[str, str]"; expected "bool"
```
I think it's related to [this issue](https://github.com/python/mypy/issues/5382)

src/urllib3/request.py
```diff
@@ -91,23 +103,23 @@ def request_encode_url(
         if headers is None:
             headers = self.headers
 
-        extra_kw = {"headers": headers}
+        extra_kw: Dict[str, Any] = {"headers": headers}
         extra_kw.update(urlopen_kw)
 
         if fields:
-            url += "?" + urlencode(fields)
+            url += "?" + urlencode(fields)  # type: ignore
```
hramezani, line 110: `error: Argument 1 to "urlencode" has incompatible type "Union[Sequence[Union[Tuple[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], RequestField]], Dict[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]]]"; expected "Union[Mapping[Any, Any], Mapping[Any, Sequence[Any]], Sequence[Tuple[Any, Any]], Sequence[Tuple[Any, Sequence[Any]]]]"`

src/urllib3/request.py
```diff
@@ -161,7 +174,7 @@ def request_encode_body(
                 )
             else:
                 body, content_type = (
-                    urlencode(fields),
+                    urlencode(fields),  # type: ignore
```
hramezani, line 177: `error: Argument 1 to "urlencode" has incompatible type "Union[Sequence[Union[Tuple[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], RequestField]], Dict[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]]]"; expected "Union[Mapping[Any, Any], Mapping[Any, Sequence[Any]], Sequence[Tuple[Any, Any]], Sequence[Tuple[Any, Sequence[Any]]]]"
`

sethmlarson (changes requested):
One thought on how to resolve a `# type: ignore`

src/urllib3/request.py
hramezani: > Since encoding into the URL only actually supports Union[Sequence[Tuple[str, Union[str, bytes]]], Mapping[str, Union[str, bytes]]] let's use this as our type hint for fields of request_encode_url?


If we do this, then we will face with error in [encode_multipart_formdata](https://github.com/urllib3/urllib3/blob/ed23062eef4900f4d4c7540f51062c94bb9dcf07/src/urllib3/request.py#L159)

`
error: Argument 1 to "encode_multipart_formdata" has incompatible type "Union[Sequence[Tuple[str, Union[str, bytes]]], Mapping[str, Union[str, bytes]]]"; expected "Union[Sequence[Union[Tuple[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], RequestField]], Dict[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]]]"
`

because `encode_multipart_formdata` uses the [_TYPE_FIELDS in filepost.py](https://github.com/urllib3/urllib3/blob/ed23062eef4900f4d4c7540f51062c94bb9dcf07/src/urllib3/filepost.py#L14).
Should we ignore this?

sethmlarson: `_TYPE_FIELDS` should be allowed in `request_encode_body`, add a `type: ignore` to the `urlencode()` if `encode_multipart` is false

hramezani:
@SethMichaelLarson anything to do?

src/urllib3/request.py
```diff
@@ -37,26 +43,31 @@ class RequestMethods:
...
-        encode_multipart=True,
-        multipart_boundary=None,
-        **kw,
+        method: str,
+        url: str,
+        body: Optional[HTTPBody] = None,
+        headers: Optional[Mapping[str, str]] = None,
+        encode_multipart: bool = True,
+        multipart_boundary: Optional[str] = None,
+        **kw: str,
```
sethmlarson, line 57: Are these really `str`? I feel like we should use `Any` for `**kw` and **`urlopen_kw`?

src/urllib3/request.py
```diff
@@ -37,26 +43,31 @@ class RequestMethods:
...
             "Classes extending RequestMethods must implement "
             "their own ``urlopen`` method."
         )
 
     def request(
-        self, method, url, fields=None, headers=None, **urlopen_kw
+        self,
+        method: str,
+        url: str,
+        fields: Optional[_TYPE_ENCODE_URL_FIELDS] = None,
```
sethmlarson, line 68: I'd rather allow the full range of fields here so users don't get type errors for passing in valid options for `request_encode_body`. Only wanted the `_TYPE_ENCODE_URL_FIELDS` to apply to `request_encode_url`.

sethmlarson (changes requested):
Last two changes then we're good to ship!

src/urllib3/request.py
hramezani: If I change it like:
`fields: Union[_TYPE_FIELDS, _TYPE_ENCODE_URL_FIELDS] = None,`

we still get this error:

```error: Argument "fields" to "request_encode_url" of "RequestMethods" has incompatible type "Union[Union[Sequence[Tuple[str, Union[str, bytes]]], Mapping[str, Union[str, bytes]], None], Union[Sequence[Union[Tuple[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], RequestField]], Dict[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]]]]"; expected "Union[Sequence[Tuple[str, Union[str, bytes]]], Mapping[str, Union[str, bytes]], None]"```

It says that expexted type is `Union[Sequence[Tuple[str, Union[str, bytes]]], Mapping[str, Union[str, bytes]], None]` which is exists in type.

any idea?

src/urllib3/request.py
hramezani: Changed!

sethmlarson (approved):
@hramezani I made one change and I think this looks good now. Let me know if this works for you and if you see any issues with my change.

PR #2204, opened 2021-04-26
Fix None/False confusion in is_connection_dropped

### LINKED ISSUE #2203: is_connection_dropped checks against None but uses False as default value for getattr
I happened to read this line and the code looks fishy. I did not otherwise verify the potential bug.

See implementation of `is_connection_dropped(conn: socket.socket) -> bool`:

https://github.com/urllib3/urllib3/blob/287052a16a59bcaba5772387de36fa9a49eb8378/src/urllib3/util/connection.py#L19-L23

If there is no property `sock` on `conn`, then we will call `wait_for_read(False, timeout=0.0)`, which e.g. may end up putting the `False` into the iterable passed to `select`.

Since this seemed to never have caused problems, the `sock = getattr(conn, "sock", False)` can probably be replaced with just `sock = conn.sock`.

Alternatives would be to replace the default (last argument of `getattr`) of `False` with `None` or replace the `if sock is None` with `if not sock`.

Oberon00:
Fixes #2203.

### DIFF
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -16,7 +16,7 @@ def is_connection_dropped(conn: socket.socket) -> bool:  # Platform-specific
     :param conn:
         :class:`http.client.HTTPConnection` object.
     """
-    sock = getattr(conn, "sock", False)
+    sock = getattr(conn, "sock", None)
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Thank you!

Issue #1988, opened 2020-09-25
[v2] Replace all string formatting with f-strings


sethmlarson:
Once we're Python 3.6+ we can start using f-strings, a very powerful and performant formatter.
There are likely other performance optimizations (such as regarding creating a hex value from an integer in chunking if I recall)

pquentin:
Closing, this was done in #1988.

PR #2205, opened 2021-05-01
Modify string formatting to use f-strings

franekmagiera:
Some small changes related to #1988 

`ResponseError.SPECIFIC_ERROR` from `exceptions` module is also used with `.format()` but I am not sure if it would make sense to, for example, refactor this to a method that accepts a `status_code` argument and returns an f-string formatted string?

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -255,10 +255,11 @@ class HTTPConnection(_HTTPConnection):
         if not any(isinstance(v, str) and v == SKIP_HEADER for v in values):
             super().putheader(header, *values)
         elif to_str(header.lower()) not in SKIPPABLE_HEADERS:
+            skippable_headers = "', '".join(
+                [str.title(header) for header in sorted(SKIPPABLE_HEADERS)]
+            )
             raise ValueError(
-                "urllib3.util.SKIP_HEADER only supports '{}'".format(
-                    "', '".join(map(str.title, sorted(SKIPPABLE_HEADERS)))
-                )
+                f"urllib3.util.SKIP_HEADER only supports '{skippable_headers}'"
             )
 
     # `request` method's signature intentionally violates LSP.
--- a/src/urllib3/contrib/_securetransport/bindings.py
+++ b/src/urllib3/contrib/_securetransport/bindings.py
@@ -54,9 +54,7 @@ version = platform.mac_ver()[0]
 version_info = tuple(map(int, version.split(".")))
 if version_info < (10, 8):
     raise OSError(
-        "Only OS X 10.8 and newer are supported, not {}.{}".format(
-            version_info[0], version_info[1]
-        )
+        f"Only OS X 10.8 and newer are supported, not {version_info[0]}.{version_info[1]}"
     )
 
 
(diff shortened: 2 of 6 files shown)


src/urllib3/connection.py
```diff
@@ -255,10 +255,11 @@ def putheader(self, header: str, *values: str) -> None:
         if not any(isinstance(v, str) and v == SKIP_HEADER for v in values):
             super().putheader(header, *values)
         elif to_str(header.lower()) not in SKIPPABLE_HEADERS:
+            skippable_headers = "', '".join(
```
franekmagiera, line 258: I would also be fine with having it inline in the f-string. Also would be fine with sticking to `map`, apparently if it does not use a `lambda` it is faster than list comprehension, which IMO is more readable and more pythonic than `map`.

pquentin:
We actually switched to f-strings in #2044 (thanks to pyupgrade) and forgot to close #1988.

As you've noticed there are other places where we can use f-strings, but it's not because we can that we should: I think most of the changes here make the code actually less readable.

src/urllib3/connection.py
pquentin: This is more readable, yes 👍

src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -54,9 +54,7 @@
 version_info = tuple(map(int, version.split(".")))
 if version_info < (10, 8):
     raise OSError(
-        "Only OS X 10.8 and newer are supported, not {}.{}".format(
-            version_info[0], version_info[1]
-        )
+        f"Only OS X 10.8 and newer are supported, not {version_info[0]}.{version_info[1]}"
```
pquentin, line 57: Not sure what to think, why not!

src/urllib3/util/url.py
```diff
@@ -23,7 +23,7 @@
 
 _IPV4_PAT = r"(?:[0-9]{1,3}\.){3}[0-9]{1,3}"
 _HEX_PAT = "[0-9A-Fa-f]{1,4}"
-_LS32_PAT = "(?:{hex}:{hex}|{ipv4})".format(hex=_HEX_PAT, ipv4=_IPV4_PAT)
+_LS32_PAT = f"(?:{_HEX_PAT}:{_HEX_PAT}|{_IPV4_PAT})"
```
pquentin, line 26: This seems worse to me

test/test_response.py
```diff
@@ -1035,11 +1035,8 @@ def _encode_chunk(self, chunk):
 
 class MockChunkedEncodingWithoutCRLFOnEnd(MockChunkedEncodingResponse):
     def _encode_chunk(self, chunk):
-        return "{:X}\r\n{}{}".format(
-            len(chunk),
-            chunk.decode(),
-            "\r\n" if len(chunk) > 0 else "",
-        )
+        crlf = "\r\n"
+        return f"{len(chunk):X}\r\n{chunk.decode()}{crlf if len(chunk) > 0 else ''}"
```
pquentin, line 1039: This seems worse to me

test/test_util.py
```diff
@@ -200,7 +200,7 @@ def test_parse_url_normalization(self, url, expected_normalized_url):
     def test_control_characters_are_percent_encoded(self, char):
         percent_char = "%" + (hex(ord(char))[2:].zfill(2).upper())
         url = parse_url(
-            "http://user{0}@example.com/path{0}?query{0}#fragment{0}".format(char)
+            f"http://user{char}@example.com/path{char}?query{char}#fragment{char}"
```
pquentin, line 203: This is better!

sethmlarson (approved):
LGTM, thanks!

franekmagiera:
Thanks for reviewing!

PR #2208, opened 2021-05-06
Complete src/urllib3/util/util.py type hint.

hramezani:
After https://github.com/urllib3/urllib3/commit/e9171e8d77b457e2c96fca37c89d68c518bec5f7, there are some mypy errors:
```
src/urllib3/util/util.py:35: error: Item "None" of "Optional[BaseException]" has no attribute "__traceback__"
src/urllib3/util/util.py:36: error: Item "None" of "Optional[BaseException]" has no attribute "with_traceback"
src/urllib3/util/util.py:37: error: Exception must be derived from BaseException
```
This patch fixes these errors.


sethmlarson (approved):
LGTM

PR #2209, opened 2021-05-06
Fix warnings in 3.10.0-beta.1

### LINKED ISSUE #2200: Python 3.10 ssl module deprecations
### Context

Heads up! I'm in the processing of cleaning up the ssl module. Python 3.10's ssl module will raise deprecation warnings for a lot of old cruft that has been deprecated since 3.6 and 3.7. For example  SSL 3.0 to TLS 1.1, ``OP_NO_SSL/TLS*``, and more old features are going away. You'll also need OpenSSL 1.1.1+ for testing.

* https://bugs.python.org/issue43669
* https://bugs.python.org/issue43880

### Alternatives

None

### Duplicate

No

### Contribution

Sorry, no time :)

sethmlarson:
Closes #2200

Will need to be backported to 1.26.x


sethmlarson:
[Looks like our test suite is failing in Python 3.10-beta.1](https://github.com/urllib3/urllib3/pull/2209/checks?check_run_id=2521440650)

sethmlarson:
@tiran Would appreciate your eyeballs to review this as well. Supporting `PROTOCOL_TLS`, `PROTOCOL_TLS_CLIENT`, along with enabled and disabled verifications turns out to be quite the hairy situation.

tiran:
Sure! I'll get back to you after PyCon.

mgorny:
@sethmlarson, could you tell me whether this is something that affects the runtime usage of urllib3 or merely the test suite? I'm wondering if we can unleash it on Python 3.10 users in Gentoo to unblock other packages and deal with the tests later.

src/urllib3/util/ssl_.py
```diff
@@ -223,7 +230,13 @@ def create_urllib3_context(
     """
     if SSLContext is None:
         raise TypeError("Can't create an SSLContext object without an ssl module")
-    context = SSLContext(ssl_version or PROTOCOL_TLS)
+
+    # PROTOCOL_TLS is deprecated in Python 3.10 so we pass PROTOCOL_TLS_CLIENT instead.
+    # If it's not available in 'ssl' the value is aliased to PROTOCOL_TLS anyways.
+    if ssl_version == PROTOCOL_TLS:
+        ssl_version = PROTOCOL_TLS_CLIENT
```
tiran, line 237: In fact all ``PROTOCOL_*`` except for ``PROTOCOL_TLS_CLIENT`` and ``PROTOCOL_TLS_SERVER`` are deprecated. OpenSSL has deprecated the other methods, too.

sethmlarson:
@mgorny The only thing this patch fixes is a `DeprecationWarning` emitted with `PROTOCOL_TLS` when using the default TLS config. This patch is for v2.x and will be backporting another patch for v1.26.x

mgorny:
@sethmlarson so I suppose it doesn't address the two timeouts? I mean, these two:

```
FAILED test/test_ssltransport.py::TlsInTlsTestCase::test_tls_in_tls_makefile_raw_rw_binary[None] - Failed: Timeout >60.0s
FAILED test/test_ssltransport.py::TlsInTlsTestCase::test_tls_in_tls_makefile_rw_text - Failed: Timeout >60.0s
```

sethmlarson:
@mgorny Ah those timeouts were fixed in https://github.com/urllib3/urllib3/pull/2218, I've just now pushed a backport commit to the 1.26.x branch.

mgorny:
Ah, cool. Then, do I understand correctly that if I backport that backport, I should be fine using that version on py3.10 against other packages, and the test failures are 'harmless'?

sethmlarson:
@mgorny Users on 3.10 will see deprecation warnings but beyond that yes the failures are harmless.

src/urllib3/util/ssl_.py
```diff
@@ -67,6 +67,13 @@ def _is_ge_openssl_v1_1_1(
     OP_NO_SSLv3 = 0x2000000  # type: ignore
     PROTOCOL_SSLv23 = PROTOCOL_TLS = 2
 
+# PROTOCOL_TLS_CLIENT is essentially PROTOCOL_TLS with additional
+# configuration so if it's not available we fallback on PROTOCOL_TLS.
+try:
+    from ssl import PROTOCOL_TLS_CLIENT
+except ImportError:
+    PROTOCOL_TLS_CLIENT = PROTOCOL_TLS
```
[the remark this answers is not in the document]
sethmlarson: For v2.0 we can do this, will have to keep the patch fairly similar to this for 1.26.x

src/urllib3/util/ssl_.py
```diff
@@ -223,7 +225,14 @@ def create_urllib3_context(
     """
     if SSLContext is None:
         raise TypeError("Can't create an SSLContext object without an ssl module")
-    context = SSLContext(ssl_version or PROTOCOL_TLS)
+
+    # PROTOCOL_TLS is deprecated in Python 3.10 so we pass PROTOCOL_TLS_CLIENT instead.
+    # If it's not available in 'ssl' the value is aliased to PROTOCOL_TLS anyways.
```
pquentin, line 230: This comment is no longer true: on Python 3.6+ PROTOCOL_TLS_CLIENT is always available

src/urllib3/util/ssl_.py
```diff
@@ -260,9 +269,19 @@ def create_urllib3_context(
...
         context.post_handshake_auth = True
 
-    context.verify_mode = cert_reqs
-    # We ask for verification here but it may be disabled in HTTPSConnection.connect
-    context.check_hostname = cert_reqs == ssl.CERT_REQUIRED
+    # The order of the below lines setting verify_mode and check_hostname
+    # matter due to safe-guards SSLContext has to prevent an SSLContext with
+    # check_hostname=True, verify_mode=NONE/OPTIONAL. This is made even more
+    # complex because we don't know whether PROTOCOL_TLS_CLIENT will be used
+    # or not so we don't know the initial state of the freshly created SSLContext.
```
pquentin, line 276: Dos the order still matter now that we're not calling any function inside those conditions?

src/urllib3/contrib/pyopenssl.py
```diff
@@ -81,6 +81,7 @@ class UnsupportedExtension(Exception):
 # Map from urllib3 to PyOpenSSL compatible parameter-values.
 _openssl_versions = {
     util.PROTOCOL_TLS: OpenSSL.SSL.SSLv23_METHOD,
+    PROTOCOL_TLS_CLIENT: OpenSSL.SSL.SSLv23_METHOD,
```
pquentin, line 84: nitpick: we now have three different ways of using `PROTOCOL` values for no great reason, can we unify?

Also, should we export `PROTOCOL_TLS_CLIENT` from `urllib3.util`?

src/urllib3/contrib/securetransport.py
```diff
@@ -111,7 +112,8 @@
 # TLSv1 and a high of TLSv1.2. For everything else, we pin to that version.
 # TLSv1 to 1.2 are supported on macOS 10.8+
 _protocol_to_min_max = {
-    util.PROTOCOL_TLS: (SecurityConst.kTLSProtocol1, SecurityConst.kTLSProtocol12)
+    util.PROTOCOL_TLS: (SecurityConst.kTLSProtocol1, SecurityConst.kTLSProtocol12),
+    PROTOCOL_TLS_CLIENT: (SecurityConst.kTLSProtocol1, SecurityConst.kTLSProtocol12),
```
pquentin, line 116: Same remark, let's use `util` or `ssl_`, not both

pquentin (dismissed):
Thanks! The diff is small and clean thanks to your efforts! Mentioned a few things to make it even better, but still approving as they're not blocking

src/urllib3/contrib/pyopenssl.py
sethmlarson: I actually don't like having `PROTOCOL_TLS` in `urllib3.util`, it's not our API (and technically it's a deprecated API now) so in v2 maybe we should remove it.

pquentin: Sounds good!

src/urllib3/util/ssl_.py
sethmlarson: Let's try without the order in 1.26.x, might just work! :crossed_fingers:

src/urllib3/util/ssl_.py
```diff
@@ -260,9 +268,13 @@ def create_urllib3_context(
     ) is not None:
         context.post_handshake_auth = True
 
-    context.verify_mode = cert_reqs
-    # We ask for verification here but it may be disabled in HTTPSConnection.connect
+    # This option may be disabled in HTTPSConnection.connect
     context.check_hostname = cert_reqs == ssl.CERT_REQUIRED
+
+    # PROTOCOL_TLS_CLIENT guarantees we have cert+hostname verification on
+    # so we turn these two options off in a specific order.
```
pquentin, line 275: What is order-specific here? I'm not following

pquentin (dismissed):
Awesome, this new version is even better and cleaner!

sethmlarson:
@tiran I'm seeing a failure when using the exact ordering of setting `check_hostname` and `verify_mode` on Windows + Python 3.6 which on GitHub Actions uses OpenSSL 1.0.2q: https://github.com/urllib3/urllib3/pull/2209/checks?check_run_id=2757846974#step:8:108

I'm going to set the ordering back to what was released in 1.26.x, but can you take a look at this and confirm that something wonky isn't happening?

graingert:
@sethmlarson it looks like OP_NO_SSLv2 is deprecated also https://github.com/urllib3/urllib3/blob/2ec06d1185f1642dfe8ff767a9ad3541bd0ddc39/src/urllib3/util/ssl_.py#L246-L248

sethmlarson:
@graingert Are we receiving warnings from our usage?

graingert:
> @graingert Are we receiving warnings from our usage?

No, but these attributes are deprecated

sethmlarson:
@graingert There's a lot of features that we use that are deprecated (see `ssl.wrap_socket`, `ssl.PROTOCOL_TLS`) because they aren't raising a warning and it'd be much harder to selectively use them in some Python versions and not in others. Even once they're removed I can see us aliasing them to a number and `|=` them to `options` the same way we do today.

tiran:
You should never ever use ``ssl.wrap_socket``. It does not support SNI and is horribly insecure because it does not verify hostnames. It raises a DeprecationWarning:

```
$ python3.10 -Werror
Python 3.10.0b2 (default, Jun  1 2021, 00:00:00) [GCC 11.1.1 20210428 (Red Hat 11.1.1-1)] on linux
>>> import ssl, socket
>>> ssl.wrap_socket(socket.socket())
  File "/usr/lib64/python3.10/ssl.py", line 1421, in wrap_socket
    warnings.warn(
DeprecationWarning: ssl module: wrap_socket is deprecated, use SSLContext.wrap_socket()
>>> ssl.SSLContext(ssl.PROTOCOL_TLS)
  File "/usr/lib64/python3.10/ssl.py", line 501, in __new__
    self = _SSLContext.__new__(cls, protocol)
DeprecationWarning: ssl module: PROTOCOL_TLS is deprecated
```

You also won't be able set the attributes with ``SSLContext.options``. The descriptor creates the deprecation warning and will soon raise an exception. You don't get the deprecation warning for ``OP_NO_SSLv2`` and ``OP_NO_SSLv3`` because ``OP_NO_SSLv2`` and ``OP_NO_SSLv3`` are already set by default. It is impossible to detect the case. (Also ``OP_NO_SSLv2`` is usually ``0`` these days).

sethmlarson:
> You should never ever use `ssl.wrap_socket`. It does not support SNI and is horribly insecure because it does not verify hostnames. It raises a DeprecationWarning:

Sorry! Typo, meant `ssl.match_hostname` which we have a vendored copy to support Python versions that don't have it / emits deprecation warnings.

> You also won't be able set the attributes with `SSLContext.options`. The descriptor creates the deprecation warnin...

Is it supposed to be raising a DeprecationWarning in Python 3.10-beta2?

>  ...and will soon raise an exception.

Since this isn't happening yet I'm going to wait until it happens. We're usually on top of changes to Python versions.

> You don't get the deprecation warning for `OP_NO_SSLv2` and `OP_NO_SSLv3` because `OP_NO_SSLv2` and `OP_NO_SSLv3` are already set by default. It is impossible to detect the case. (Also `OP_NO_SSLv2` is usually `0` these days).

This works in our favor, doesn't it? Since we're supporting Python versions in 1.26.x that don't do this it's helpful that we can blindly set these options. Having them be `0` is helpful in this case too.

For 2.x I'm definitely on board to remove all of this though, I believe that's being tracked here: https://github.com/urllib3/urllib3/issues/2233

Let me know if there's something we should be doing that we aren't on `1.26.x` or `main`, thank you for helping us with everything TLS @tiran :pray:

tiran:
> > You also won't be able set the attributes with `SSLContext.options`. The descriptor creates the deprecation warnin...
> 
> Is it supposed to be raising a DeprecationWarning in Python 3.10-beta2?

Yes, but it can only raise a DeprecationWarning when you add or remove an ``OP_NO_*`` constants. There is no way to detect a no-op. :(

PR #2210, opened 2021-05-07
Add type hints to urllib3.util.request

hramezani:

### DIFF
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -1,5 +1,5 @@
 from base64 import b64encode
-from typing import IO, Any, AnyStr, Dict, List, Optional, Union
+from typing import IO, AnyStr, Dict, List, Optional, Union
 
 from ..exceptions import UnrewindableBodyError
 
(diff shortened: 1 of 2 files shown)


src/urllib3/util/request.py
```diff
@@ -66,7 +67,7 @@ def make_headers(
         >>> make_headers(accept_encoding=True)
         {'accept-encoding': 'gzip,deflate'}
     """
-    headers = {}
+    headers: Dict[str, Any] = {}
```
sethmlarson, line 70: Why is the value type `Any` instead of `str`?

src/urllib3/util/request.py
```diff
@@ -98,7 +99,9 @@ def make_headers(
     return headers
 
 
-def set_file_position(body, pos):
+def set_file_position(
+    body: Union[BinaryIO, TextIO], pos: Optional[Union[int, object]]
```
sethmlarson, line 103: Is there any advantage to using `Union[BinaryIO, TextIO]` over `IO`?

sethmlarson (changes requested):
Two small comments then this looks good :)

src/urllib3/util/request.py
hramezani: I was wrong. `str` is the right type.

src/urllib3/util/request.py
hramezani: Just to make it clear. I've replaced it with `IO[AnyStr]`

sethmlarson (dismissed):
LGTM

sethmlarson (approved):

PR #2211, opened 2021-05-08
Add type hints to urllib3.util.timeout

hramezani:

### DIFF
--- a/src/urllib3/util/timeout.py
+++ b/src/urllib3/util/timeout.py
@@ -2,13 +2,17 @@ import time
 
 # The default socket timeout, used by httplib to indicate that no timeout was
 # specified by the user
-from socket import _GLOBAL_DEFAULT_TIMEOUT
+from socket import _GLOBAL_DEFAULT_TIMEOUT  # type: ignore
+from typing import Optional, Union
 
 from ..exceptions import TimeoutStateError
 
 # A sentinel value to indicate that no timeout was specified by the user in
 # urllib3
 _Default = object()
+
+
+_TYPE_TIMEOUT = Optional[Union[float, int, object]]
 
 
 class Timeout:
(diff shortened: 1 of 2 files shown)


src/urllib3/util/timeout.py
```diff
@@ -95,22 +98,29 @@ class Timeout:
     """
 
     #: A sentinel object representing the default timeout value
-    DEFAULT_TIMEOUT = _GLOBAL_DEFAULT_TIMEOUT
-
-    def __init__(self, total=None, connect=_Default, read=_Default):
+    DEFAULT_TIMEOUT: object = _GLOBAL_DEFAULT_TIMEOUT
```
sethmlarson, line 101: Make this `: _TYPE_TIMEOUT` so it's settable by users via `urllib3.util.Timeout.DEFAULT_TIMEOUT = ...`

src/urllib3/util/timeout.py
```diff
@@ -132,15 +142,15 @@ def _validate_timeout(cls, value, name):
                 "be an int, float or None."
             )
         try:
-            float(value)
+            float(value)  # type: ignore
```
sethmlarson, line 145: Let's add an `value = float(value)` here so we don't need to do the `# type: ignore` below?

src/urllib3/util/timeout.py
```diff
@@ -196,7 +206,7 @@ def start_connect(self):
         self._start_connect = current_time()
```
sethmlarson, line 206: We can actually use `time.monotonic` instead of `current_time` here (and elsewhere) because it's [always available in Python 3.5+](https://docs.python.org/3/library/time.html#time.monotonic)

sethmlarson (changes requested):
This looks good, a few comments:

src/urllib3/util/timeout.py
hramezani: if we change it to `value = float(value)`, we have to keep `# type: ignore` for this line. otherwise, we will get:

```src/urllib3/util/timeout.py:145: error: Argument 1 to "float" has incompatible type "object"; expected "Union[SupportsFloat, _SupportsIndex, str]"```

and [test_timeout_str](https://github.com/urllib3/urllib3/blob/26e9f27692193f898e78f88ea8169558ced30d12/test/test_util.py#L597) will fail because the numbers are int now.
But, Yes, we don't need below `# type: ignore` after that.

Should we change it?

File: src/urllib3/util/timeout.py (hunk 3 above)
Comment (hramezani) [reply src/urllib3/util/timeout.py:206]: Here is the [PR](https://github.com/urllib3/urllib3/pull/2214)

File: src/urllib3/util/timeout.py [hunk 4]
```diff
@@ -181,7 +192,7 @@ def clone(self):
         # detect the user default.
         return Timeout(connect=self._connect, read=self._read, total=self.total)
 
-    def start_connect(self):
+    def start_connect(self) -> Optional[float]:
```
sethmlarson, line 195: This function always returns `float` and is only `Optional[float]` due to `self._start_connect`'s type, should we do `self._start_connect = ret = time.monotonic(); return ret`?

sethmlarson (approved):
LGTM, thanks!

PR #2217, opened 2021-05-13
Update LICENSE.txt

Saiprasad16:


sethmlarson:
We don't need to update the year every year so I'm going to close this PR.

PR #2220, opened 2021-05-15
Start using SSLContext.minimum|maximum_version when available

### LINKED ISSUE #2110: Switch to setting SSLContext.minimum_version
The `ssl.OP_NO_TLSvX` options were deprecated in Python 3.7. These values are available on OpenSSL 1.1.0g+, if we detect they are available with an SSLContext that supports them we should set `SSLContext.minumum_version = TLSVersion.TLSv1_2`

sethmlarson:
Closes #2110
Closes https://github.com/urllib3/urllib3/issues/2373

- [x] Adding `minimum_version` and `maximum_version` converters to the PyOpenSSL and SecureTransport SSLContext implementations
- [x] ~Fixing the test suite for the new deprecation message~
- [x] Adding documentation about `ssl_minimum_version` and `ssl_maximum_version`, removing documentation about `ssl_version`.
- [x] ~Creating an issue about deprecating `ssl_version` once Python 3.6 support is dropped. Right now we're in a soft deprecation as if we were emitting deprecation warnings now it'd make life hard for any library supporting Python 3.6+ and using the option.~

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -515,31 +515,6 @@ class HTTPSConnection(HTTPConnection):
             tls_in_tls=tls_in_tls,
         )
 
-        # If we're using all defaults and the connection
-        # is TLSv1 or TLSv1.1 we throw a DeprecationWarning
-        # for the host.
-        if hasattr(self.sock, "version"):
-            tls_version = self.sock.version()
-        else:
-            tls_version = None
-
-        if (
-            default_ssl_context
-            and self.ssl_minimum_version is None
-            and self.ssl_maximum_version is None
-            and tls_version is not None
-            and tls_version in {"TLSv1", "TLSv1.1"}
-        ):
-            warnings.warn(
-                "Negotiating TLSv1/TLSv1.1 by default is deprecated "
-                "and will be disabled in urllib3 v2.0.0. Connecting to "
-                f"'{self.host}' with '{tls_version}' can be "
-                f"enabled by explicitly opting-in with "
-                f"'ssl_minimum_version=ssl.TLSVersion.{tls_version.replace('.', '_')}'",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-
         if self.assert_fingerprint:
             assert_fingerprint(
                 self.sock.getpeercert(binary_form=True), self.assert_fingerprint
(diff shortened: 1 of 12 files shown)


pquentin:
You can now remove the Python 3.6 specific code! This will make this easier

sethmlarson:
Definitely! I have some local changes that rebased this on `main` without the Python 3.6 fallbacks but still have tests to work though.

sethmlarson:
I think this is ready for review functionality-wise! For future reviewers please rebase merge this PR as it contains multiple features.

sethmlarson:
And now the feature is documented too, ready for a full review.

pquentin (changes requested):
Thanks! This is great. I've only looked at the code (not the tests or docs) on my phone, hopefully this initial review is still useful.

There are two legitimate coverage issues, can you please take a look?

src/urllib3/connection.py
```diff
@@ -377,6 +379,9 @@ def __init__(
         self.key_password = key_password
         self.ssl_context = ssl_context
         self.server_hostname = server_hostname
+        self.ssl_version = None
```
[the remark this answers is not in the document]
sethmlarson: `self.ssl_version` isn't defined anywhere else and I thought it was strange that we were assigning to an attribute not defined in `HTTPSConnection.__init__` within `HTTPSConnectionPool`. Obviously this works fine "because Python" but figured it'd be nicer to be explicit that `ssl_version` is indeed a valid attribute of `HTTPSConnection`.

src/urllib3/contrib/securetransport.py
```diff
@@ -165,6 +166,21 @@
...
+    ssl.TLSVersion.TLSv1_1: SecurityConst.kTLSProtocol11,
+    ssl.TLSVersion.TLSv1_2: SecurityConst.kTLSProtocol12,
+    ssl.TLSVersion.MAXIMUM_SUPPORTED: SecurityConst.kTLSProtocol12,
+}
+# Reverse mapping for mapping back to TLSVersion once set.
+_st_to_tls_version = {
+    v: k
+    for k, v in _tls_version_to_st.items()
+    if k not in (ssl.TLSVersion.MINIMUM_SUPPORTED, ssl.TLSVersion.MAXIMUM_SUPPORTED)
+}
```
pquentin, line 181: Do we use this somewhere?

pquentin (approved):
Thanks! Feel free to merge after looking at my only comment.

src/urllib3/contrib/securetransport.py
sethmlarson: Nope! Was used in a previous implementation but survived the change. Great catch, thank you :)

sethmlarson: Removed this construction, going to merge now :)

Issue #2222, opened 2021-05-16
Add parameters to top-level request() and RequestMethods.request types


sethmlarson:
`body` is a frequently used parameter, would be good to have it available in the mentioned places.

PR #2226, opened 2021-05-18
Added body parameter to top-level request() and RequestMethods.request()

venthur:
This adds the `body` parameter to the top-level request() and RequestMethods.request(). For both cases there is a simple echo-test that validates that this feature works.

For ReqestsMethos.requests() the `body` parameter was always implicitly available via the `urlopen_kw` -- so even if you remove the `body` parameter and that tiny piece of code that then adds it to `urlopen_kw` the test will not fail.

Closes: #2222


sethmlarson (approved):
LGTM, thank you!

PR #2229, opened 2021-05-18
fixed rules for naming variables

venthur:
Rules for naming variables:

    PoolManager and ProxyManager should be http
    ConnectionPool should be pool
    Connection should be conn
    HTTPResponse should be resp (ie resp = http.request("GET", "https://example.com"))
    Only use example.com or httpbin.org for example URLs

There was only a few occurrences to fix.

related: #2223

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -28,7 +28,7 @@ performance to increase this number::
     >>> import urllib3
     >>> http = urllib3.PoolManager(maxsize=10)
     # Alternatively
-    >>> http = urllib3.HTTPConnectionPool('google.com', maxsize=10)
+    >>> pool = urllib3.HTTPConnectionPool('google.com', maxsize=10)
 
 The behavior of the pooling for :class:`~connectionpool.ConnectionPool` is
 different from :class:`~poolmanager.PoolManager`. By default, if a new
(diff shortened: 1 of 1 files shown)


docs/advanced-usage.rst
```diff
@@ -117,8 +117,8 @@ You can use :class:`~poolmanager.ProxyManager` to tunnel requests through an
 HTTP proxy::
 
     >>> import urllib3
-    >>> proxy = urllib3.ProxyManager('http://localhost:3128/')
-    >>> proxy.request('GET', 'http://google.com/')
+    >>> http = urllib3.ProxyManager('http://localhost:3128/')
+    >>> http.request('GET', 'http://google.com/')
```
pquentin, line 121: @sethmlarson I think I liked the `proxy` name which makes it clear that the request is going through a proxy, but don't have any strong opinion. What's the rationale for using `http` here?

pquentin (approved):

docs/advanced-usage.rst
sethmlarson: I'm ++ on using `proxy` here, I'll update the rules.

sethmlarson (approved):
LGTM

sethmlarson:
Btw @venthur:
- If you're interested we have a public Discord where we're discussing this effort, if possible it'd be nice to coordinate these changes.
- Instead of writing "partially fixes #XYZ" in the issue write "related to #XYZ", etc. The words "fixes #XYZ" makes the issue get closed after I merge the PR.

PR #2231, opened 2021-05-18
Modified Documentation for v2

ubdussamad:
This PR is now close to complete.

It's up for review. Minor changes might be required.

This PR is related to the Issue: #2223

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -93,8 +93,14 @@ further parameterize pytest for local testing.
 For all valid arguments, check `the pytest documentation
 <https://docs.pytest.org/en/stable/usage.html#stopping-after-the-first-or-n-failures>`_.
 
-Documentation Guidelines
-------------------------
+Contributing to documentation
+-----------------------------
+
+You can build the docs locally using ``nox``:
+
+.. code-block:: bash
+
+  $ nox -rs docs
 
 While writing documentation you should follow these guidelines:
 
@@ -102,14 +108,14 @@ While writing documentation you should follow these guidelines:
 - Use double quotes for all strings. (Output, Declaration etc.)
 - Use keyword arguments everywhere except for method and url. (ie ``http.request("GET", "https://example.com", headers={...})`` )
 - Use HTTPS in URLs everywhere unless HTTP is needed.
-- Rules for naming variables:
+- Rules for code examples and naming variables:
 
-  - PoolManager should be http. (ie ``http = urllib3.PoolManager(...)``)
-  - ProxyManager should be proxy.
-  - ConnectionPool should be pool.
-  - Connection should be conn.
-  -  HTTPResponse should be resp.
-  -  Only use example.com or httpbin.org for example URLs
+  - ``PoolManager`` instances should be named ``http``. (ie ``http = urllib3.PoolManager(...)``)
+  - ``ProxyManager`` instances should be named ``proxy``.
+  - ``ConnectionPool`` instances should be named ``pool``.
+  - ``Connection`` instances should be named ``conn``.
+  - ``HTTPResponse`` instances should be named ``resp``.
+  -  Only use ``example.com`` or ``httpbin.org`` for example URLs
 
 - Comments within snippets should be useful, if what's being done is apparent
   (such as parsing JSON, making a request) then it can be skipped for that section.
@@ -121,21 +127,6 @@ While writing documentation you should follow these guidelines:
 - No whitespace is required between the sections as normally would be in case of isort.
 - Add print statements along with a comment below them showing the output, potentially compressed.
 - This helps users using the copy-paste button immediately see the results from a script.
-
-For testing, you can build the docs locally using:
-
-
-.. code-block:: bash
-
-  $ nox -rs docs
-
-Incase you're modifying a python file for documentation, please run the linter and see if it catches any errors.
-
-To run the linter locally run:
-
-.. code-block:: bash
-
-  $ nox -rs lint
 
 Releases
 --------
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -93,6 +93,50 @@ further parameterize pytest for local testing.
 For all valid arguments, check `the pytest documentation
 <https://docs.pytest.org/en/stable/usage.html#stopping-after-the-first-or-n-failures>`_.
 
+Documentation Guidelines
+------------------------
+
+While writing documentation you should follow these guidelines:
+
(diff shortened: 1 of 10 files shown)


docs/user-guide.rst
```diff
@@ -46,11 +49,16 @@ HTTP verb:
...
 .. code-block:: python
 
-    >>> r = http.request(
-    ...     'POST',
-    ...     'http://httpbin.org/post',
-    ...     fields={'hello': 'world'}
-    ... )
+    import urllib3
+
+    r = urllib3.request(
```
sethmlarson: The braces in this example don't line up horizontally, should be:

```python
resp = urllib3.request(
    ...
)
```

docs/user-guide.rst
```diff
@@ -119,11 +136,17 @@ to ``False``. By default HTTP responses are closed after reading all bytes, this
...
+    import urllib3
+
+    http = urllib3.PoolManager() # Creating a seprate PoolManager instance.
+
+    r = http.request('GET', 'https://example.com', preload_content=False)
+    r.auto_close = False
+
+    for line in io.TextIOWrapper(r):
+        print(line)
+    #<!doctype html> ...
```
sethmlarson, line 155: Let's remove the `# ...` here only because it's kinda clunky since we're in a loop.

docs/user-guide.rst
```diff
@@ -20,23 +20,26 @@ First things first, import the urllib3 module:
...
 To make a request use :meth:`~poolmanager.PoolManager.request`:
 
 .. code-block:: python
 
-    >>> r = http.request('GET', 'http://httpbin.org/robots.txt')
-    >>> r.data
-    b'User-agent: *\nDisallow: /deny\n'
+    import urllib3
+    # Send the GET request.
+    r = urllib3.request('GET', 'https://httpbin.org/robots.txt')
```
sethmlarson: May as well use proper names like `resp` instead of `r` if we're modifying the example

docs/user-guide.rst
```diff
@@ -20,23 +20,26 @@ First things first, import the urllib3 module:
...
+    # Print the returned data.
+    print(r.data)
```
sethmlarson: Let's put a blank line before every section so it's easier to parse visually

docs/user-guide.rst
```diff
@@ -20,23 +20,26 @@ First things first, import the urllib3 module:
...
+    #b'User-agent: *\nDisallow: /deny\n'
```
sethmlarson, line 45: Should be a space after every `#` on a comment line and some whitespace between every section. Also quotes should be Black formatted so double quotes. This example should look like this:

```python
    import urllib3

    # Creating a PoolManager instance for sending requests.
    http = urllib3.PoolManager()

    # Send the GET request and getting back a HTTPResponse object.
    resp = http.request("GET", "https://httpbin.org/robots.txt")

    # Print the returned data.
    print(resp.data)
    # b"User-agent: *\nDisallow: /deny\n"
```

sethmlarson (changes requested):
Thank you! Left a handful of comments to start off with, apply these across the board and I'll review again :)

docs/user-guide.rst
ubdussamad: Yeah, did that in the latest commit

docs/user-guide.rst
ubdussamad: Done.

docs/user-guide.rst
ubdussamad: Will do.

sethmlarson:
I pushed my own change, take a look at it and learn how the style should look for future PRs. If you are going to make more changes you'll have to do a `git pull` of the branch to grab my change too.

src/urllib3/util/url.py
```diff
@@ -149,12 +149,18 @@ def url(self) -> str:
...
-            ... '/path', 'query', 'fragment').url
-            'http://username:password@host.com:80/path?query#fragment'
+            import urllib3
+
+            U = urllib3.util.parse_url("https://google.com/mail/")
+
+            print(U.url)
+            # "http://google.com/mail/"
+
+            print( urllib3.util.Url("http", "username:password",
```
ubdussamad, line 159: I am totally not sure about the best way to format this.

docs/contributing.rst
```diff
@@ -93,6 +93,50 @@ further parameterize pytest for local testing.
 For all valid arguments, check `the pytest documentation
 <https://docs.pytest.org/en/stable/usage.html#stopping-after-the-first-or-n-failures>`_.
 
+Documentation Guidelines
```
ubdussamad, line 96: Check this out, I am not sure if this is what you wanted.

ubdussamad (commented):
Please check this out whenever you get time. Sorry, I know it's a lot to review at once. I tried the best I could but it'll likey have some errors.

shazow:
Wish there was a way to preview a rendered version of this automagically, Netlify has spoiled me. Looks alright so far. :)

sethmlarson:
@shazow There is though! https://urllib3--2231.org.readthedocs.build/en/2231

shazow:
@SethMichaelLarson Whoa the future is here! Thank you.

Personally I'm not a huuuuge fan of how the new formatting looks. I think it's okay and I wouldn't block it, but I would propose maybe we keep the "repl" format for the first opening block perhaps? And do the copypasta format for the rest.

image

vs

image

It could just be my brain but the copypasta format for the intro block looks somehow staggered and harder to grok the intent of what it's trying to say.

The rest of the blocks, in user guide etc, actually look quite comprehensible and nice.

ubdussamad:
> @SethMichaelLarson Whoa the future is here! Thank you.
> 
> ~Personally I'm not a huuuuge fan of how the new copypasta snippets feel in all cases. I think it's okay and I wouldn't block it, but~ (Actually I quite like the new formatting everywhere except the intro example.)
> 
> I would propose maybe we keep the "repl" format for the first opening block perhaps? And do the copypasta format for the rest.
> 
> image
> 
> vs
> 
> image
> 
> It could just be my brain but the copypasta format for the intro block looks somehow staggered and harder to grok the intent of what it's trying to say.
> 
> The rest of the blocks, in user guide etc, actually look quite comprehensible and nice.

@shazow, Indeed the console-style looks more vibrant. But I really don't have any opinion for/against its use in the docs. I like it either way.

Btw, do you want both the github readme & the front page of docs? Or just the github readme.

ubdussamad:
@sethmlarson @shazow  Okay, Ill push those changes to both the files.

sethmlarson (approved):
This looks good to me now, thanks for all of this! :tada:

PR #2232, opened 2021-05-18
Add type hints to urllib3.poolmanager

hramezani:
This is a draft PR to add type hint for `urllib3.poolmanager`. There are some mypy errors that we can ignore.
Here is the list of errors:

```
src/urllib3/poolmanager.py:9: error: Module 'urllib3.connectionpool' does not explicitly export attribute 'port_by_scheme'; implicit reexport disabled
src/urllib3/poolmanager.py:73: error: List or tuple literal expected as the second argument to namedtuple()
src/urllib3/poolmanager.py:125: error: Returning Any from function declared to return "PoolKey"
src/urllib3/poolmanager.py:125: error: "PoolKey" not callable
src/urllib3/poolmanager.py:305: error: No overload variant of "get" of "Mapping" matches argument type "PoolKey"
src/urllib3/poolmanager.py:305: note: Possible overload variant:
src/urllib3/poolmanager.py:305: note:     def get(self, key: int) -> Optional[Callable[[_VT], None]]
src/urllib3/poolmanager.py:305: note:     <1 more non-matching overload not shown>
src/urllib3/poolmanager.py:307: error: Returning Any from function declared to return "Union[HTTPConnectionPool, HTTPSConnectionPool]"
src/urllib3/poolmanager.py:310: error: Value of type "Optional[Dict[str, Any]]" is not indexable
src/urllib3/poolmanager.py:311: error: Value of type "Optional[Dict[str, Any]]" is not indexable
src/urllib3/poolmanager.py:312: error: Value of type "Optional[Dict[str, Any]]" is not indexable
src/urllib3/poolmanager.py:314: error: Invalid index type "PoolKey" for "RecentlyUsedContainer[int, Optional[Callable[[_VT], None]]]"; expected type "int"
src/urllib3/poolmanager.py:316: error: Returning Any from function declared to return "Union[HTTPConnectionPool, HTTPSConnectionPool]"
src/urllib3/poolmanager.py:369: error: Signature of "urlopen" incompatible with supertype "RequestMethods"
src/urllib3/poolmanager.py:388: error: "Mapping[str, str]" has no attribute "copy"
src/urllib3/poolmanager.py:400: error: Value of type variable "AnyStr" of "urljoin" cannot be "object"
src/urllib3/poolmanager.py:422: error: Argument "response" to "increment" of "Retry" has incompatible type "BaseHTTPResponse"; expected "Optional[HTTPResponse]"
src/urllib3/poolmanager.py:487: error: Function is missing a type annotation for one or more arguments
src/urllib3/poolmanager.py:509: error: Incompatible types in assignment (expression has type "Url", variable has type "None")
src/urllib3/poolmanager.py:512: error: Incompatible types in assignment (expression has type "ProxyConfig", variable has type "None")
src/urllib3/poolmanager.py:533: error: "None" has no attribute "host"
src/urllib3/poolmanager.py:533: error: "None" has no attribute "port"
src/urllib3/poolmanager.py:533: error: "None" has no attribute "scheme"
src/urllib3/poolmanager.py:553: error: Signature of "urlopen" incompatible with supertype "RequestMethods"
```

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -5,7 +5,7 @@ from typing import (
     Any,
     Callable,
     Dict,
-    List,
+    FrozenSet,
     Mapping,
     NamedTuple,
     Optional,
@@ -30,6 +30,7 @@ from .exceptions import (
 )
 from .request import RequestMethods
 from .response import BaseHTTPResponse
+from .util.connection import SocketOptions
 from .util.proxy import connection_requires_http_tunnel
 from .util.retry import Retry
 from .util.timeout import Timeout
@@ -64,29 +65,29 @@ SSL_KEYWORDS = (
 class PoolKey(NamedTuple):
     key_scheme: str
     key_host: str
-    key_port: int
-    key_timeout: Union[Timeout, float, int]
-    key_retries: Union[Retry, int]
-    key_block: bool
-    key_source_address: str
-    key_key_file: str
-    key_key_password: str
-    key_cert_file: str
-    key_cert_reqs: str
-    key_ca_certs: str
-    key_ssl_version: str
-    key_ca_cert_dir: str
+    key_port: Optional[int]
+    key_timeout: Optional[Union[Timeout, float, int]]
+    key_retries: Optional[Union[Retry, int]]
+    key_block: Optional[bool]
+    key_source_address: Optional[Tuple[str, int]]
+    key_key_file: Optional[str]
+    key_key_password: Optional[str]
+    key_cert_file: Optional[str]
+    key_cert_reqs: Optional[str]
+    key_ca_certs: Optional[str]
+    key_ssl_version: Optional[str]
+    key_ca_cert_dir: Optional[str]
     key_ssl_context: Optional["ssl.SSLContext"]
-    key_maxsize: int
-    key_headers: Mapping[str, str]
-    key__proxy: Url
-    key__proxy_headers: Mapping[str, str]
-    key__proxy_config: Type[ProxyConfig]
-    key_socket_options: List[Tuple[int, int, Union[int, str]]]
-    key__socks_options: Dict[str, Any]
-    key_assert_hostname: Union[bool, str]
-    key_assert_fingerprint: str
-    key_server_hostname: str
+    key_maxsize: Optional[int]
+    key_headers: Optional[FrozenSet[Tuple[str, str]]]
+    key__proxy: Optional[Url]
+    key__proxy_headers: Optional[FrozenSet[Tuple[str, str]]]
+    key__proxy_config: Optional[ProxyConfig]
+    key_socket_options: Optional[SocketOptions]
+    key__socks_options: Optional[FrozenSet[Tuple[str, str]]]
+    key_assert_hostname: Optional[Union[bool, str]]
+    key_assert_fingerprint: Optional[str]
+    key_server_hostname: Optional[str]
 
 
 def _default_key_normalizer(
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -1,9 +1,20 @@
 import collections
 import functools
 import logging
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Mapping,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 from urllib.parse import urljoin
 
-from ._collections import RecentlyUsedContainer
+from ._collections import _VT, RecentlyUsedContainer
 from .connection import ProxyConfig
 from .connectionpool import HTTPConnectionPool, HTTPSConnectionPool, port_by_scheme
 from .exceptions import (
@@ -13,9 +24,13 @@ from .exceptions import (
     URLSchemeUnknown,
 )
 from .request import RequestMethods
+from .response import BaseHTTPResponse
 from .util.proxy import connection_requires_http_tunnel
 from .util.retry import Retry
-from .util.url import parse_url
+from .util.url import Url, parse_url
+
+if TYPE_CHECKING:
+    from typing_extensions import Literal
 
 __all__ = ["PoolManager", "ProxyManager", "proxy_from_url"]
 
@@ -68,7 +83,9 @@ _key_fields = (
 PoolKey = collections.namedtuple("PoolKey", _key_fields)
 
 
-def _default_key_normalizer(key_class, request_context):
+def _default_key_normalizer(
+    key_class: PoolKey, request_context: Dict[str, Any]
+) -> PoolKey:
     """
     Create a pool key out of a request context dictionary.
 
@@ -110,6 +127,7 @@ def _default_key_normalizer(key_class, request_context):
         context["key_" + key] = context.pop(key)
 
     # Default to ``None`` for keys missing from the context
+    field: Tuple[str, ...]
     for field in key_class._fields:
         if field not in context:
             context[field] = None
@@ -166,25 +184,40 @@ class PoolManager(RequestMethods):
     proxy = None
     proxy_config = None
 
-    def __init__(self, num_pools=10, headers=None, **connection_pool_kw):
+    def __init__(
+        self,
+        num_pools: int = 10,
+        headers: Optional[Mapping[str, str]] = None,
+        **connection_pool_kw: Any,
+    ) -> None:
         super().__init__(headers)
         self.connection_pool_kw = connection_pool_kw
-        self.pools = RecentlyUsedContainer(num_pools, dispose_func=lambda p: p.close())
+        func: Callable[[Any], Any] = lambda p: p.close()
+        self.pools: RecentlyUsedContainer[int, Optional[Callable[[_VT], None]]]
+        self.pools = RecentlyUsedContainer(num_pools, dispose_func=func)
 
         # Locally set the pool classes and keys so other PoolManagers can
         # override them.
         self.pool_classes_by_scheme = pool_classes_by_scheme
         self.key_fn_by_scheme = key_fn_by_scheme.copy()
 
-    def __enter__(self):
+    def __enter__(self) -> "PoolManager":
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(
+        self, exc_type: object, exc_val: object, exc_tb: object
+    ) -> "Literal[False]":
         self.clear()
         # Return False to re-raise any potential exceptions
         return False
 
-    def _new_pool(self, scheme, host, port, request_context=None):
+    def _new_pool(
+        self,
+        scheme: str,
+        host: str,
+        port: int,
+        request_context: Optional[Dict[str, Any]] = None,
+    ) -> Union[HTTPConnectionPool, HTTPSConnectionPool]:
         """
         Create a new :class:`urllib3.connectionpool.ConnectionPool` based on host, port, scheme, and
         any additional pool keyword arguments.
@@ -194,7 +227,7 @@ class PoolManager(RequestMethods):
         connection pools handed out by :meth:`connection_from_url` and
         companion methods. It is intended to be overridden for customization.
         """
-        pool_cls = self.pool_classes_by_scheme[scheme]
+        pool_cls: Type[HTTPConnectionPool] = self.pool_classes_by_scheme[scheme]
         if request_context is None:
             request_context = self.connection_pool_kw.copy()
 
@@ -211,7 +244,7 @@ class PoolManager(RequestMethods):
 
         return pool_cls(host, port, **request_context)
 
-    def clear(self):
+    def clear(self) -> None:
         """
         Empty our store of pools and direct them all to close.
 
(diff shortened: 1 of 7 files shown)


src/urllib3/poolmanager.py
```diff
@@ -434,14 +469,14 @@ class ProxyManager(PoolManager):
...
-        proxy_headers=None,
-        proxy_ssl_context=None,
-        use_forwarding_for_https=False,
-        **connection_pool_kw,
-    ):
+        proxy_url: str,
+        num_pools: int = 10,
+        headers: Optional[Mapping[str, str]] = None,
+        proxy_headers: Optional[Mapping[str, str]] = None,
+        proxy_ssl_context = None,
```
hramezani, line 476: we can use `Optional["ssl.SSLContext"]` here but we have to import `ssl`.

src/urllib3/poolmanager.py
```diff
@@ -166,25 +184,40 @@ class PoolManager(RequestMethods):
...
+    def __init__(
+        self,
+        num_pools: int = 10,
+        headers: Optional[Mapping[str, str]] = None,
+        **connection_pool_kw: Any,
+    ) -> None:
         super().__init__(headers)
         self.connection_pool_kw = connection_pool_kw
-        self.pools = RecentlyUsedContainer(num_pools, dispose_func=lambda p: p.close())
+        func: Callable[[Any], Any] = lambda p: p.close()
```
sethmlarson, line 195: Can we call this `dispose_func`, also instead of a named lambda should we use a `def` here?

src/urllib3/poolmanager.py
```diff
@@ -166,25 +184,40 @@ class PoolManager(RequestMethods):
...
         self.key_fn_by_scheme = key_fn_by_scheme.copy()
 
-    def __enter__(self):
+    def __enter__(self) -> "PoolManager":
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(
+        self, exc_type: object, exc_val: object, exc_tb: object
+    ) -> "Literal[False]":
```
sethmlarson, line 209: Should we use `bool` here? Does `Literal[False]` have special meaning for mypy + `__exit__`?

src/urllib3/poolmanager.py
```diff
@@ -220,7 +253,13 @@ def clear(self):
         """
         self.pools.clear()
 
-    def connection_from_host(self, host, port=None, scheme="http", pool_kwargs=None):
+    def connection_from_host(
```
sethmlarson, line 256: Since HTTPConnectionPool and HTTPSConnectionPool don't have any differences in public APIs does it make sense to have a Union here and below?

src/urllib3/poolmanager.py
```diff
@@ -360,7 +407,7 @@ def urlopen(self, method, url, redirect=True, **kw):
             return response
 
         # Support relative URLs for redirecting.
-        redirect_location = urljoin(url, redirect_location)
+        redirect_location = str(urljoin(url, redirect_location))
```
sethmlarson, line 410: Why does this need a `str()` call? Shouldn't `urljoin(str, str)` be `str`?

src/urllib3/poolmanager.py
```diff
@@ -449,14 +496,14 @@ class ProxyManager(PoolManager):
...
         self,
-        proxy_url,
-        num_pools=10,
-        headers=None,
-        proxy_headers=None,
+        proxy_url: str,
+        num_pools: int = 10,
+        headers: Optional[Mapping[str, str]] = None,
+        proxy_headers: Optional[Mapping[str, str]] = None,
         proxy_ssl_context=None,
```
sethmlarson, line 503: `Optional["ssl.SSLContext"]`?

sethmlarson (changes requested):
Looks pretty good, a handful of comments for you!

src/urllib3/poolmanager.py
```diff
@@ -68,7 +83,9 @@
 PoolKey = collections.namedtuple("PoolKey", _key_fields)
```
sethmlarson, line 83: `PoolKey` will have to change into a `typing.NamedTuple`, I don't think that'll be too tough looking through `_key_fields`?

src/urllib3/poolmanager.py
```diff
@@ -243,22 +282,26 @@ def connection_from_host(self, host, port=None, scheme="http", pool_kwargs=None)
...
+        scheme = str(request_context["scheme"]).lower()
         pool_key_constructor = self.key_fn_by_scheme.get(scheme)
         if not pool_key_constructor:
             raise URLSchemeUnknown(scheme)
         pool_key = pool_key_constructor(request_context)
 
         return self.connection_from_pool_key(pool_key, request_context=request_context)
 
-    def connection_from_pool_key(self, pool_key, request_context=None):
+    def connection_from_pool_key(
```
sethmlarson, line 302: The `request_context` parameter here can't be `Optional` as it's used immediately below. This is a strange one, this API is only used in one place and in that place `request_context` is required. Maybe it should be required here as well? If we do this we need to mark it in the "breaking change" list as it technically is but almost no one should be using these APIs unless they know what they're doing so it should be fine :)

src/urllib3/poolmanager.py
hramezani: I think we have to use `Literal[False]`. otherwise we will get:
```
src/urllib3/poolmanager.py:210: error: "bool" is invalid as return type for "__exit__" that always returns False
src/urllib3/poolmanager.py:210: note: Use "typing_extensions.Literal[False]" as the return type or change it to "None"
src/urllib3/poolmanager.py:210: note: If return type of "__exit__" implies that it may return True, the context manager may swallow exceptions

```

src/urllib3/poolmanager.py
hramezani: Then we get ` Name 'ssl' is not defined` from mypy

src/urllib3/poolmanager.py
hramezani: > Shouldn't urljoin(str, str) be str?

Yes

> Why does this need a str() call? 

Because the type of `redirect_location` is `Union[bool, str]`  https://github.com/urllib3/urllib3/blob/2ef05adfcf001c85f1f95ae3fa078d16190851b0/src/urllib3/response.py#L202

I will change this line to 
```
redirect_location = urljoin(url, str(redirect_location))
```

hramezani:
@SethMichaelLarson I added a commit and addressed your comments.
Also, I added some `type: ignore`

src/urllib3/poolmanager.py
```diff
@@ -33,42 +56,42 @@
...
-    "key_server_hostname",  # str
-)
-
 #: The namedtuple class used to construct keys for the connection pool.
 #: All custom key schemes should include the fields in this key at a minimum.
-PoolKey = collections.namedtuple("PoolKey", _key_fields)
-
-
-def _default_key_normalizer(key_class, request_context):
+class PoolKey(NamedTuple):
```
sethmlarson, line 64: All of these keys except `scheme` and `host` should be optional. `headers`, `_proxy_headers`, and `_socks_options` should be FrozenSet[Tuple[<key>, <value>]] because they go through a normalizer.

src/urllib3/poolmanager.py
```diff
@@ -269,20 +321,22 @@ def connection_from_pool_key(self, pool_key, request_context=None):
         with self.pools.lock:
             # If the scheme, host, or port doesn't match existing open
             # connections, open a new ConnectionPool.
-            pool = self.pools.get(pool_key)
+            pool = self.pools.get(pool_key)  # type: ignore
```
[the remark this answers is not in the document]
hramezani: Wow! Great one :dart: 
By this we can remove the other `type: ignore` in this function as well.

sethmlarson:
I'm going to give this another complete review once I have access to a laptop, been on mobile for all these reviews and it makes it tough to see big picture :)

src/urllib3/util/connection.py
```diff
@@ -1,12 +1,12 @@
...
-from typing import List, Optional, Tuple, Union
+from typing import Optional, Sequence, Tuple, Union
 
 from urllib3.exceptions import LocationParseError
 
 from .wait import wait_for_read
 
 SOCKET_GLOBAL_DEFAULT_TIMEOUT = socket._GLOBAL_DEFAULT_TIMEOUT  # type: ignore
-SocketOptions = List[Tuple[int, int, Union[int, bytes]]]
+SocketOptions = Sequence[Tuple[int, int, Union[int, bytes]]]
```
sethmlarson, line 9: Since we're changing this to `Sequence` can we add a small test that makes sure sending for instance a tuple of socket options works okay?

src/urllib3/poolmanager.py
```diff
@@ -33,42 +56,42 @@
...
-    "key__proxy_headers",  # dict
-    "key__proxy_config",  # class
-    "key_socket_options",  # list of (level (int), optname (int), value (int or str)) tuples
-    "key__socks_options",  # dict
-    "key_assert_hostname",  # bool or string
-    "key_assert_fingerprint",  # str
-    "key_server_hostname",  # str
-)
-
 #: The namedtuple class used to construct keys for the connection pool.
```
sethmlarson, line 62: We can move this comment documentation into the class docstring now :tada:

sethmlarson (changes requested):
Okay here's your final review, a couple of questions too!

src/urllib3/poolmanager.py
sethmlarson: Can have `ssl` import within a `if typing.TYPE_CHECKING:` branch

src/urllib3/poolmanager.py
```diff
@@ -382,7 +437,7 @@ def urlopen(self, method, url, redirect=True, **kw):
                     kw["headers"].pop(header, None)
 
         try:
-            retries = retries.increment(method, url, response=response, _pool=conn)
+            retries = retries.increment(method, url, response=response, _pool=conn)  # type: ignore
```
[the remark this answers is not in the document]
hramezani: `src/urllib3/poolmanager.py:440: error: Argument "response" to "increment" of "Retry" has incompatible type "BaseHTTPResponse"; expected "Optional[HTTPResponse]"`

src/urllib3/poolmanager.py
```diff
@@ -243,22 +290,26 @@ def connection_from_host(self, host, port=None, scheme="http", pool_kwargs=None)
...
+        self, request_context: Dict[str, Any]
+    ) -> HTTPConnectionPool:
         """
         Get a :class:`urllib3.connectionpool.ConnectionPool` based on the request context.
 
         ``request_context`` must at least contain the ``scheme`` key and its
         value must be a key in ``key_fn_by_scheme`` instance variable.
         """
-        scheme = request_context["scheme"].lower()
+        scheme = str(request_context["scheme"]).lower()
```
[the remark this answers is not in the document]
hramezani: I just removed the `str` and no error. It seems I added it by mistake.

src/urllib3/poolmanager.py
```diff
@@ -348,7 +403,7 @@ def urlopen(self, method, url, redirect=True, **kw):
         kw["redirect"] = False
 
         if "headers" not in kw:
-            kw["headers"] = self.headers.copy()
+            kw["headers"] = dict(self.headers).copy()
```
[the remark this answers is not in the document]
hramezani: Ok, `# type: ignore` added.

src/urllib3/poolmanager.py
```diff
@@ -373,7 +428,7 @@ def urlopen(self, method, url, redirect=True, **kw):
         # Strip headers marked as unsafe to forward to the redirected location.
         # Check remove_headers_on_redirect to avoid a potential network call within
         # conn.is_same_host() which may use socket.gethostbyname() in the future.
-        if retries.remove_headers_on_redirect and not conn.is_same_host(
+        if retries.remove_headers_on_redirect and not conn.is_same_host(  # type: ignore
```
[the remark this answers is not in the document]
hramezani: The `is_same_host` didn't have a type hint. I added a type hint for the function and removed `# type: ignore`

src/urllib3/poolmanager.py
```diff
@@ -360,7 +415,7 @@ def urlopen(self, method, url, redirect=True, **kw):
             return response
 
         # Support relative URLs for redirecting.
-        redirect_location = urljoin(url, redirect_location)
+        redirect_location = urljoin(url, str(redirect_location))
```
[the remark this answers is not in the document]
hramezani: Thanks for the hint. Yes, I changed it like you said.

src/urllib3/util/connection.py
hramezani: I changed the `test_socket_options` test and add a tuple of socket options

src/urllib3/poolmanager.py
hramezani: Ok :+1:

hramezani:
@sethmlarson Thanks for the review. I addressed your comments.

sethmlarson (approved):
This looks fantastic, thanks for all of this great work!! 👏

Issue #1214, opened 2017-06-08
.coveragerc removes all raise statements from coverage checks


sethmlarson:
This was discovered in #1211 and affects a few places that ought to have coverage. See #1212 for a full list of modules that are affected. We should aim to remove `raise` from `.coveragerc` and have test coverage for all these places.

As @Lukasa said, this is a great place for a contributor to make a real difference in the quality of this code-base.

theodesp:
So basically we go through the https://codecov.io/gh/urllib3/urllib3/tree/master/src/urllib3 reports and check where the raise statements are not covered and provide test cases for that? If that's OK  I will try to push some PRs.

SethMichaelLarson:
Yeah that's the idea! :) Don't feel like you have to cover them all in one PR either. Committing a few at a time is totally fine.

PR #2234, opened 2021-05-19
Add coverage for all raise branches

venthur:
This PR tackles all of the missing `raise` statements missing from coverage.

Since I'm not super familiar with urllibs internals, you might find stuff you're not very happy with.

I also removed the `exclude_lines = raise` from `.coveragerc`, since the coverage should now be on par with the `main` branch.

Fixes: #1214

### DIFF
--- a/src/urllib3/util/timeout.py
+++ b/src/urllib3/util/timeout.py
@@ -146,11 +146,18 @@ class Timeout:
                 "int, float or None." % (name, value)
             )
 
-        if value <= 0:  # type: ignore
+        try:
+            if value <= 0:  # type: ignore
+                raise ValueError(
+                    "Attempted to set %s timeout to %s, but the "
+                    "timeout cannot be set to a value less "
+                    "than or equal to 0." % (name, value)
+                )
+        except TypeError:
+            # Python 3
             raise ValueError(
-                "Attempted to set %s timeout to %s, but the "
-                "timeout cannot be set to a value less "
-                "than or equal to 0." % (name, value)
+                "Timeout value %s was %s, but it must be an "
+                "int, float or None." % (name, value)
             )
 
         return value
--- a/src/urllib3/util/timeout.py
+++ b/src/urllib3/util/timeout.py
@@ -146,18 +146,11 @@ class Timeout:
                 "int, float or None." % (name, value)
             )
 
-        try:
-            if value <= 0:  # type: ignore
-                raise ValueError(
-                    "Attempted to set %s timeout to %s, but the "
-                    "timeout cannot be set to a value less "
-                    "than or equal to 0." % (name, value)
-                )
-        except TypeError:
-            # Python 3
+        if value <= 0:  # type: ignore
             raise ValueError(
-                "Timeout value %s was %s, but it must be an "
-                "int, float or None." % (name, value)
+                "Attempted to set %s timeout to %s, but the "
+                "timeout cannot be set to a value less "
+                "than or equal to 0." % (name, value)
             )
 
         return value
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -1,4 +1,3 @@
-import errno
 import http.client as httplib
 import ssl
 from http.client import HTTPException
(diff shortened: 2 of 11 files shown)


test/test_connectionpool.py
```diff
@@ -533,6 +535,21 @@ def _test(exception):
         _test(SocketError)
         _test(ProtocolError)
 
+    def test_raise_timeout_error(self):
```
sethmlarson, line 538: So it turns out the branch you're adding this for was [added 7 years ago](https://github.com/urllib3/urllib3/pull/493) and has been unchanged since, tbh I'm not even sure how to trigger it since we never knew it wasn't covered. The test case supposedly triggering it is `with_dummserver/test_socketlevel.py:test_https_connection_read_timeout`

test/test_response.py
```diff
@@ -49,6 +50,18 @@ def sock():
     s.close()
 
 
+class TestContentDecoder:
+    decoder = ContentDecoder()
+
+    def test_decompress_error(self):
+        with pytest.raises(NotImplementedError):
```
sethmlarson, line 57: What are your thoughts on adding `raise NotImplementedError` to the `.coveragerc` skip list? Then we don't need mostly no-op tests like this to cover those cases

test/test_response.py
```diff
@@ -745,6 +758,22 @@ def test_read_not_chunked_response_as_chunks(self):
         with pytest.raises(ResponseNotChunked):
             next(r)
 
+    def test_read_chunked_not_supported(self):
```
sethmlarson, line 761: We might be able to implement this one without mocks, try passing an `io.BytesIO` to `HTTPResponse` and reading from it with `read_chunked()`? Seems to be what is called out in the PR adding this functionality: https://github.com/urllib3/urllib3/pull/990

test/test_ssltransport.py
```diff
@@ -503,3 +503,65 @@ def test_constructor_params(self):
             mock.ANY, mock.ANY, server_hostname=server_hostname
         )
         assert not ssl_transport.suppress_ragged_eofs
+
+    def test_recv_no_flags_error(self):
```
sethmlarson, line 507: Since all of these are basically mocking out and using the same setup, how about combining them into one test that asserts all the things with `flags=...`?

sethmlarson (changes requested):
Wow, what a great patch! :tada: Thank you for this, I know it definitely took a while :) Here's a handful of comments for you:

src/urllib3/util/timeout.py
```diff
@@ -146,18 +146,11 @@ def _validate_timeout(
...
-        try:
-            if value <= 0:  # type: ignore
-                raise ValueError(
-                    "Attempted to set %s timeout to %s, but the "
-                    "timeout cannot be set to a value less "
-                    "than or equal to 0." % (name, value)
-                )
-        except TypeError:
-            # Python 3
+        if value <= 0:  # type: ignore
```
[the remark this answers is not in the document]
venthur: Oh, you're right. I've overlooked that case and was confused by the `float(value)` above which does not return anything.

test/test_connectionpool.py
```diff
@@ -533,6 +535,21 @@ def _test(exception):
...
+        with HTTPConnectionPool(host="localhost") as pool:
+            err = Mock()
+            err.errno = errno.EAGAIN
+            with pytest.raises(ReadTimeoutError):
+                pool._raise_timeout(err, "", 0)
+
+    def test_make_request_error(self):
```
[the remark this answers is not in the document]
venthur: This line here:

https://github.com/urllib3/urllib3/blob/13678970dc9bae35a596310e047aca7c27537181/src/urllib3/connectionpool.py#L337

test/test_response.py
venthur: Thanks, done.

test/test_connectionpool.py
venthur: i've moved this test to `with_dummserver/test_socketlevel.py:test_https_connection_read_timeout` and added a comment to make clear that we test both errors here.

sethmlarson (approved):
So much value added here, thank you! :bow:

Issue #2240, opened 2021-05-21
Cert verify on subjectAltName->"IP Address" not working when doing IPv6 via a proxy


derekhiggins:
### Subject
Verifying a https certificate by the subjectAltName->"IP Address" fails when the accessing a server by IPv6 address through a Proxy

in the case where a proxy is used the Host IP used for comparison is wrapped in square brackets

### Environment

```
[derekh@u07 urllib3]$ python
Python 3.6.8 (default, Aug 24 2020, 17:57:11) 
[GCC 8.3.1 20191121 (Red Hat 8.3.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> import urllib3
>>> 
>>> print("OS", platform.platform())
OS Linux-4.18.0-240.22.1.el8_3.x86_64-x86_64-with-centos-8.3.2011
>>> print("Python", platform.python_version())
Python 3.6.8
>>> print("urllib3", urllib3.__version__)
urllib3 1.24.2

Also reproduced on the master branch
>>> print("urllib3", urllib3.__version__)
urllib3 2.0.0.dev0

```

### Steps to Reproduce

This works (without a proxy)
```
> conn = urllib3.connection_from_url('https://[fd00:1101::7d50:2883:3df9:68ce]:9999', ca_certs='ca.crt')
> response = conn.request('GET', '/')
> print(response.data)
HELLO

```
Accessing the same URL via a proxy doesn't 
```
> proxy = urllib3.ProxyManager('http://[fd00:1101::1]:3128', ca_certs='ca.crt')
> conn = proxy.connection_from_url('https://[fd00:1101::7d50:2883:3df9:68ce]:9999')
> try:
>     response = conn.request('GET', '/', retries=0)
> except Exception as e:
>     print(e)
HTTPSConnectionPool(host='fd00:1101::7d50:2883:3df9:68ce', port=9999): Max retries exceeded with url: / (Caused by SSLError(CertificateError("hostname '[fd00:1101::7d50:2883:3df9:68ce]' doesn't match 'FD00:1101:0:0:7D50:2883:3DF9:68CE\\n'",),))

```

PR #2241, opened 2021-05-21
Striped square brackets from host IP

derekhiggins:
When comparing host IP to a cert subjectAltName->"IP Address"
ensure were strip square brackets from IPv6 addresses.

Fixes: https://github.com/urllib3/urllib3/issues/2240


derekhiggins:
I also looked into stripping out the brackets further up the stack
in  PoolManager.connection_from_host , but this causes the CONNECT to the proxy to break 
`1621610373.972      3 fd00:1101::1 TCP_TUNNEL/200 1050 CONNECT [fd00:1101::7d50:2883:3df9:68ce]:9999 - HIER_DIRECT/fd00:1101::7d50:2883:3df9:68ce -                                                               
` 
vs.
`1621606788.209    257 fd00:1101::1 TAG_NONE/503 0 CONNECT fd00:1101 - HIER_NONE/- -
`

hramezani (approved):
I am not sure about the right place to stripping out the brackets would be `match_hostname` function, but the implementation and tests look good to me.
Also, I think we need to backport it to 1.26.x

sethmlarson:
I wonder if there's a better place to do this within the proxy flow?

derekhiggins:
> I wonder if there's a better place to do this within the proxy flow?

The alternative I tried was to strip them out in PoolManager.connection_from_host but this failed.

Looking a bit deeper this appears to be because HTTPSConnection.set_tunnel requires the square brackets to be in place. Which I guess is why they were left in place in the first place. I could strip them out in connection_from_host and then put them
back in for the call to set_tunnel but I'm not sure if there are other places where I'd need to put them back.

derekhiggins:
Rebased

sethmlarson (approved):
This much more localized implementation looks good to me, thank you!

Would love an integration test too, especially one that tests proxies too (cc @jalopezsilva).

The integration test on 1.26.x will probably reveal a slightly more complex route for 1.26.x because iirc we don't unconditionally use our implementation of `ssl_match_hostname()` there. Will have to take a peek.

sethmlarson (approved):
LGTM

jalopezsilva (approved):
Looks good to me!

jalopezsilva:
@sethmlarson, should I add the proxy integration directly here or wait until we merge it?

sethmlarson:
@jalopezsilva Let's merge this then you can add the proxy integration after. I think this one may be non-trivial to backport, maybe you can look at this a bit too?

PR #2246, opened 2021-05-27
[1.26] Configure lint with pre-commit

pquentin:
I'd like to bring `1.26.x` CI closer to the `main` CI because we're making a lot 1.26.x releases and they are often stressful, in part because they tend to include CVEs, in part due to bad CI. https://twitter.com/sethmlarson/status/1397603959733243908

I'm starting with lint: this uses pre-commit just like in main but omits mypy and configures pyupgrade/black to use Python 2.7. Consider reviewing/merging the two commits separately due to the amount of noise black/pyupgrade introduce.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,21 +7,6 @@ defaults:
     shell: bash
 
 jobs:
-  lint:
-    runs-on: ubuntu-latest
-
-    steps:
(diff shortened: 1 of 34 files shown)


pquentin:
THis fails just like 1.26.x normally does, so is safe to merge :)

pquentin:
Closing in favor of #2247.

PR #2247, opened 2021-05-27
[1.26] Run CI with GitHub Actions only

pquentin:
This is a follow up of #2246 which runs all 1.26.x tests on GitHub Actions, which means pull requests won't need Travis anymore.


pquentin:
macOS 3.5 failure:
```
FAILED test/contrib/test_pyopenssl.py::TestSocketClosing::test_connection_closed_on_read_timeout_preload_false
```

Closing/reopening

src/urllib3/exceptions.py
```diff
@@ -22,7 +22,7 @@ class PoolError(HTTPError):
 
     def __init__(self, pool, message):
         self.pool = pool
-        HTTPError.__init__(self, "%s: %s" % (pool, message))
+        HTTPError.__init__(self, "{}: {}".format(pool, message))
```
sethmlarson, line 25: All of the changes from `%` to `.format()` aren't great because `%` is more performant and the same level of readable. I don't know if `pyupgrade` is configurable in this way?

src/urllib3/util/url.py
```diff
@@ -51,7 +51,7 @@
 ]
 
 UNRESERVED_PAT = r"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._!\-~"
-IPV6_PAT = "(?:" + "|".join([x % _subs for x in _variations]) + ")"
+IPV6_PAT = "(?:" + "|".join(x % _subs for x in _variations) + ")"
```
sethmlarson, line 54: `"".join([...])` is faster than `"".join(...)`

sethmlarson (changes requested):
This is exciting! I had a couple comments about some of the pyupgrade changes

src/urllib3/util/url.py
shazow: This is absolutely wild to me, but very true!

Had to see it for myself:

```ipython
In [6]: %timeit "(?:" + "|".join(x % _subs for x in _variations) + ")"
4.9 µs ± 26.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit "(?:" + "|".join([x % _subs for x in _variations]) + ")"
4.66 µs ± 46.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

pquentin:
I removed pyupgrade, which I only knew how to do by rebasing

sethmlarson (approved):
This is going to make both of our lives so much better, woo! :partying_face:

PR #2250, opened 2021-06-02
Improve JSON Support to request and response

### LINKED ISSUE #2243: Proposal: Improve support for JSON
Many APIs are simply JSON bodies over HTTP, it'd be nice if we could make that experience more enjoyable for most users.

On the request side:

- Raise a `ValueError` if `body` is given or if `fields` is given when within `request_encode_body`
- Set `Content-Type: application/json` via `headers.setdefault()` to avoid stepping on custom Content-Types.
- Encode the JSON to bytes and set as the `body` parameter, remove spaces after delimiters.
- Add `json` to `urllib3.request()`

On the response side:

- Add a function `json()` to `HTTPResponse`
- `HTTPResponse.json()` should parse the JSON via `json.loads(resp.data)` and return the parsed object.

On the documentation side:

- In the JSON section, recommend using `json=` instead of encoding your own JSON to bytes if you're not doing anything special
- Document how to use a custom JSON library/encoder like orjson by encoding to bytes and sending via `body=`
  and to pull `resp.data` for decoding JSON responses

V1NAY8:
Closes #2243 

Opening this as draft. 

Todo
- [x] Add documentation
- [x] Add tests 

@sethmlarson 
I hope those are the changes to be done. Can you please review this draft?
Let me know if any changes are required.

Also where should I write tests?

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -102,8 +102,8 @@ The :class:`~response.HTTPResponse` object provides
 JSON Content
 ~~~~~~~~~~~~
 
-JSON content can be loaded by using :attr:`~response.HTTPResponse.json`
-attribute of the request or by decoding and deserializing the 
+JSON content can be loaded by using :meth:`~response.HTTPResponse.json`
+method of the request or by decoding and deserializing the 
 :attr:`~response.HTTPResponse.data` attribute of the request:
 
 .. code-block:: python
@@ -280,9 +280,9 @@ JSON
 ~~~~
 
 You can send a JSON request by specifying the data as ``json`` argument,
-urllib3 automatically encodes data using ``json`` module with ``utf-8`` 
+urllib3 automatically encodes data using ``json`` module with ``UTF-8`` 
 encoding. Also by default ``"Content-Type"`` in headers is set to 
-"application/json"`` if not specified when calling
+``"application/json"`` if not specified when calling
 :meth:`~poolmanager.PoolManager.request`:
 
 .. code-block:: python
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -132,7 +132,6 @@ as the ``body``argument and setting the ``Content-Type`` header when calling
     print(json.loads(resp.data.decode("utf-8")))
     # {"origin": "127.0.0.1"}
 
-
 Binary Content
 ~~~~~~~~~~~~~~
 
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -102,7 +102,8 @@ The :class:`~response.HTTPResponse` object provides
 JSON Content
 ~~~~~~~~~~~~
 
-JSON content can be loaded by decoding and deserializing the
+JSON content can be loaded by using :attr:`~response.HTTPResponse.json`
+attribute of the request or by decoding and deserializing the 
 :attr:`~response.HTTPResponse.data` attribute of the request:
 
 .. code-block:: python
@@ -111,6 +112,9 @@ JSON content can be loaded by decoding and deserializing the
     import urllib3
 
     resp = urllib3.request("GET", "https://httpbin.org/ip")
+
+    resp.json()
+    # {"origin": "127.0.0.1"}
 
     print(json.loads(resp.data.decode("utf-8")))
     # {"origin": "127.0.0.1"}
@@ -275,8 +279,30 @@ dictionary in the ``fields`` argument provided to
 JSON
 ~~~~
 
-You can send a JSON request by specifying the encoded data as the ``body``
-argument and setting the ``Content-Type`` header when calling
+You can send a JSON request by specifying the data as ``json`` argument,
+urllib3 automatically encodes data using ``json`` module with ``utf-8`` 
+encoding. Also by default ``"Content-Type"`` in headers is set to 
+"application/json"`` if not specified when calling
+:meth:`~poolmanager.PoolManager.request`:
+
+.. code-block:: python
+
+    import urllib3
+
+    data = {"attribute": "value"}
+
+    resp = urllib3.request(
+        "POST",
+        "https://httpbin.org/post",
+        json=data, # Passing data without encoding
+    )
+
+    print(resp.json())
+    # {"attribute": "value"}
+
+
+Alternatively, you can send a JSON request by specifying the encoded data 
+as the ``body``argument and setting the ``Content-Type`` header when calling
 :meth:`~poolmanager.PoolManager.request`:
 
 .. code-block:: python
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -102,9 +102,11 @@ The :class:`~response.HTTPResponse` object provides
 JSON Content
 ~~~~~~~~~~~~
 
-JSON content can be loaded by using :meth:`~response.HTTPResponse.json`
-method of the request or by decoding and deserializing the 
-:attr:`~response.HTTPResponse.data` attribute of the request:
+You can send a JSON request by specifying the data as ``json`` argument,
+urllib3 automatically encodes data using ``json`` module with ``UTF-8`` 
+encoding. Also by default ``"Content-Type"`` in headers is set to 
+``"application/json"`` if not specified when calling
+:meth:`~poolmanager.PoolManager.request`:
 
 .. code-block:: python
 
@@ -113,11 +115,23 @@ method of the request or by decoding and deserializing the
 
     resp = urllib3.request("GET", "https://httpbin.org/ip")
 
-    resp.json()
+    print(resp.json())
     # {"origin": "127.0.0.1"}
+
+Alternatively, you can send a JSON request by specifying the encoded data 
+as the ``body``argument and setting the ``Content-Type`` header when calling
+:meth:`~poolmanager.PoolManager.request`:
+
+.. code-block:: python
+
+    import json
+    import urllib3
+
+    resp = urllib3.request("GET", "https://httpbin.org/ip")
 
     print(json.loads(resp.data.decode("utf-8")))
     # {"origin": "127.0.0.1"}
+
 
 Binary Content
 ~~~~~~~~~~~~~~
@@ -185,7 +199,7 @@ You can specify headers as a dictionary in the ``headers`` argument in :meth:`~p
         }
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["headers"])
+    print(resp.json()["headers"])
     # {"X-Something": "value", ...}
 
 Or you can use the ``HTTPHeaderDict`` class to create multi-valued HTTP headers:
@@ -207,7 +221,7 @@ Or you can use the ``HTTPHeaderDict`` class to create multi-valued HTTP headers:
         headers=headers
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["headers"])
+    print(resp.json()["headers"])
     # {"Accept": "application/json, text/plain", ...}
    
 
@@ -229,7 +243,7 @@ arguments as a dictionary in the ``fields`` argument to
         fields={"arg": "value"}
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["args"])
+    print(resp.json()["args"])
     # {"arg": "value"}
 
 For ``POST`` and ``PUT`` requests, you need to manually encode query parameters
@@ -248,9 +262,8 @@ in the URL:
     url = "https://httpbin.org/post?" + encoded_args
     resp = urllib3.request("POST", url)
 
-    print(json.loads(resp.data.decode("utf-8"))["args"])
+    print(resp.json()["args"])
     # {"arg": "value"}
-
 
 
 .. _form_data:
@@ -273,8 +286,10 @@ dictionary in the ``fields`` argument provided to
         fields={"field": "value"}
     )
     
-    print(json.loads(resp.data.decode("utf-8"))["form"])
+    print(resp.json()["form"])
     # {"field": "value"}
+
+.. _json:
 
 JSON
 ~~~~
@@ -283,26 +298,6 @@ You can send a JSON request by specifying the data as ``json`` argument,
 urllib3 automatically encodes data using ``json`` module with ``UTF-8`` 
 encoding. Also by default ``"Content-Type"`` in headers is set to 
 ``"application/json"`` if not specified when calling
-:meth:`~poolmanager.PoolManager.request`:
-
-.. code-block:: python
-
-    import urllib3
-
-    data = {"attribute": "value"}
-
-    resp = urllib3.request(
-        "POST",
-        "https://httpbin.org/post",
-        json=data, # Passing data without encoding
-    )
-
-    print(resp.json())
-    # {"attribute": "value"}
-
-
-Alternatively, you can send a JSON request by specifying the encoded data 
-as the ``body``argument and setting the ``Content-Type`` header when calling
 :meth:`~poolmanager.PoolManager.request`:
 
 .. code-block:: python
@@ -322,7 +317,7 @@ as the ``body``argument and setting the ``Content-Type`` header when calling
         headers={"Content-Type": "application/json"}
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["json"])
+    print(resp.json())
     # {"attribute": "value"}
 
 Files & Binary Data
@@ -350,7 +345,7 @@ approach as :ref:`form_data` and specify the file field as a tuple of
         }
     )
     
-    print(json.loads(resp.data.decode("utf-8"))["files"])
+    print(resp.json()["files"])
     # {"filefield": "..."}
 
 While specifying the filename is not strictly required, it's recommended in
@@ -385,7 +380,7 @@ recommended to set the ``Content-Type`` header:
         headers={"Content-Type": "image/jpeg"}
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["data"])
+    print(resp.json()["data"])
     # data:application/octet-stream;base64,...
 
 .. _ssl:
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -102,7 +102,7 @@ The :class:`~response.HTTPResponse` object provides
 JSON Content
 ~~~~~~~~~~~~
 JSON content can be loaded by :meth:`~response.HTTPResponse.json` 
-method of the request:
+method of the response:
 
 .. code-block:: python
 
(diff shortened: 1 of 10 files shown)


V1NAY8:
An Example of Regular expression usage to remove whitespaces after delimiters
```python
>>> import re
>>> x = r.data
>>> x
b'{\n  "args": {}, \n  "data": "{\\"attribute\\":\\"value\\"}", \n  "files": {}, \n  "form": {}, \n  "headers": {\n    "Accept-Encoding": "identity", \n    "Content-Length": "21", \n    "Content-Type": "application/json", \n    "Host": "httpbin.org", \n    "User-Agent": "python-urllib3/2.0.0.dev0", \n    "X-Amzn-Trace-Id": "Root=1-60b750f0-086f7f880502163a4bd7656c"\n  }, \n  "json": {\n    "attribute": "value"\n  }, \n  "origin": "103.108.117.119", \n  "url": "http://httpbin.org/post"\n}\n'
>>> re.sub(b"[^\\S\r\n\t]", b"", x)    
b'{\n"args":{},\n"data":"{\\"attribute\\":\\"value\\"}",\n"files":{},\n"form":{},\n"headers":{\n"Accept-Encoding":"identity",\n"Content-Length":"21",\n"Content-Type":"application/json",\n"Host":"httpbin.org",\n"User-Agent":"python-urllib3/2.0.0.dev0",\n"X-Amzn-Trace-Id":"Root=1-60b750f0-086f7f880502163a4bd7656c"\n},\n"json":{\n"attribute":"value"\n},\n"origin":"103.108.117.119",\n"url":"http://httpbin.org/post"\n}\n'
>>>
```

pquentin:
@V1NAY8 One clarification regarding `orjson`: we don't want to add a new dependency (we don't have nay bu default) or write any code using it, we just want to document how it can be used if for some reason the `json` standard module is too slow.

src/urllib3/request.py
```diff
@@ -84,7 +88,20 @@ def request(
...
+
+        if json is not None:
+            if headers is None:
+                headers = self.headers.setdefault("Content-Type", "application/json")
+            # orjson is strict UTF-8
+            body = orjson.dumps(json)
+
         if body is not None:
+            # Remove spaces for delimiters after encoding
+            body = re.sub(b"[^\\S\r\n\t]", b"", body)
```
pquentin, line 104: You don't need a regex here, just pass `delimiters=(',', ':')` to `json.dumps` (documented here: https://docs.python.org/3/library/json.html#json.dump)

V1NAY8:
Ok, I'll remove `orjson` and implement using `json`, so that we won't have any new package.

src/urllib3/request.py
pquentin: (And this should only be done in the `json` case, there's no reason to touch an existing `body`)

V1NAY8: Fixed this. Thanks
Regex is always fun!

src/urllib3/request.py
```diff
@@ -84,6 +86,16 @@ def request(
...
         urlopen_kw["request_url"] = url
 
+        if json is not None and body is not None:
+            raise TypeError(
+                "request got values for both 'json' and 'body', can only specify one. Prefer json"
+            )
+
+        if json is not None:
+            if headers is None:
+                headers = self.headers.setdefault("Content-Type", "application/json")  # type: ignore
```
pquentin, line 96: We need to support `HTTPHeaderDict`, so you'll have to check if the header is defined before setting it instead of using `setdefault `.

src/urllib3/response.py
```diff
@@ -455,6 +457,13 @@ def data(self):
         if self._fp:
             return self.read(cache_content=True)
 
+    def json(self) -> Mapping[str, str]:
+        res = self.data
+        if isinstance(res, bytes):
+            return json.loads(res.decode("utf-8"))
+        else:
+            return json.loads(res)
```
pquentin, line 465: The `isinstance` check is not necessary since `self.data` is always bytes. (And `json.loads` does this already internally anyway.). This is enough:

```suggestion
        return json.loads(self.data)
```

pquentin (changes requested):
Do you need help for tests and docs?

V1NAY8:
@pquentin Can you tell me in which file to write tests. Ill write them. Ill try to add documentation. 
Then we can make necessary changes to it, to look good :)

src/urllib3/request.py
V1NAY8: Hmm, So the logic should be , If `headers is None`, then `setdefault`
Else if headers is HTTPHeaderDict, then check if it has Content-Type (No, then `setdefault`)
else if it is not HTTPHeaderDict and Not None then throw error
 ?

pquentin:
You can add tests after the top level request tests in test/with_dummyserver/test_poolmanager.py

src/urllib3/request.py
```diff
@@ -84,6 +86,16 @@ def request(
 
         urlopen_kw["request_url"] = url
 
+        if json is not None and body is not None:
+            raise TypeError(
+                "request got values for both 'json' and 'body', can only specify one. Prefer json"
```
shazow, line 91:
```suggestion
                "request got values for both 'json' and 'body', can only specify one."
```

IMO "Prefer json" is confusing. Probably best to cut it, or could add a link to https://urllib3.readthedocs.io/en/latest/user-guide.html (I don't think we do links in exceptions anywhere else yet, but I've grown to be a big fan of this in other projects).

shazow:
@V1NAY8 Thank you for working on this! I'm excited to get this added. :)

src/urllib3/request.py
V1NAY8: Ok, I just realized what `setdefault` does on empty dict.

V1NAY8: @pquentin  or @shazow . I'm a little confused on this part

Does ``defined before``means like this.
```python
import urllib3
header = urllib3.HTTPHeaderDict(cookie="foo, bar")
http = urllib3.PoolManager(headers=header)
data = {"attribute": "value"}
r = http.request(
    "POST",
    "http://httpbin.org/post",
    json=data
)
```
So self.headers can be `HTTPHeaderDict` or `{}`
In either case, we can by default set it as 
```python
if headers is None:
    headers  = self.headers
    headers["Content-Type"] = "application/json"
```

Or am I looking this wrong ? 🤔

src/urllib3/request.py
```diff
@@ -88,12 +88,14 @@ def request(
...
         if json is not None and body is not None:
             raise TypeError(
-                "request got values for both 'json' and 'body', can only specify one. Prefer json"
+                "request got values for both 'json' and 'body', can only specify one."
             )
 
         if json is not None:
             if headers is None:
-                headers = self.headers.setdefault("Content-Type", "application/json")  # type: ignore
+                headers = self.headers
```
shazow, line 96: I think this needs to be `self.headers.copy()` or similar, otherwise we'll be mutating the self's default headers in the next line that will persist to future requests.

src/urllib3/request.py
```diff
@@ -88,12 +88,14 @@ def request(
...
+                headers["Content-Type"] = "application/json"
```
shazow, line 97: We probably also want to provide a Content-Type header even if headers are passed in _but_ Content-Type is not passed in. Ideally in a case-insensitive way. :)

Maybe normalizing headers to a HTTPHeaderDict early on is a good strategy which would take care of normalizing the keys? Not sure!

(A test for this would be great)

src/urllib3/response.py
```diff
@@ -455,6 +457,9 @@ def data(self):
         if self._fp:
             return self.read(cache_content=True)
 
+    def json(self) -> Mapping[str, str]:
```
bluetech, line 460: I think this should be `-> Any`, because nothing restricts the response to only JSON objects.

Also, I think a docstring would be nice here, for the API reference. In particular, I would mention the exception this can raise (`json.JSONDecodeError`); since this is user-controlled input, a lot of users of this function would want to handle it.

src/urllib3/request.py
```diff
@@ -84,6 +86,18 @@ def request(
...
+            raise TypeError(
+                "request got values for both 'json' and 'body', can only specify one."
+            )
+
+        if json is not None:
+            if headers is None:
+                headers = self.headers
+                headers["Content-Type"] = "application/json"
+
+            body = json_serializer.dumps(json, separators=(",", ":")).encode("utf-8")
```
bluetech, line 99: May I suggest using `ensure_ascii=False` here? IMO there is no good reason to escape to ASCII except for uglifying the output.

src/urllib3/request.py
```diff
@@ -68,6 +69,7 @@ def request(
         body: Optional[HTTPBody] = None,
         fields: Optional[_TYPE_FIELDS] = None,
         headers: Optional[Mapping[str, str]] = None,
+        json: Optional[HTTPBody] = None,
```
bluetech, line 72: `HTTPBody` is not the correct type for this. It is possible to to provide some type safety with something like `Union[None, bool, float, str, Sequence[Any], Mapping[str, Any]]` but without support for recursive types in mypy it is not possible to provide an exact type (hence the Any's). So IMO, you can just go with `object`.

BTW, `None` is a valid input to `json.dumps` so you might want to use some sentinel object instead of `None` for the default.

pquentin:
@V1NAY8 Do you need help on this?

V1NAY8:
Yep a little, I am trying to understand the reviews given. Just not sure what to do on the `HTTPHeaderDict` part.

src/urllib3/request.py
pquentin: This isn't about empty or not, but whether `Content-Type` is already set or not.

In a `dict`, `setdefault` only updates `Content-Type` if it does not exist already. This is exactly the behavior we want.

However, `HTTPHeaderDict` does not have a `setdefault` method. Maybe we can add it, but if we don't we can't call `setdefault` here, we need to do this manually:

```
if "Content-Type" not in headers:
    headers ["Content-Type"] = "application/json"
```

And that's it! The code will work for both dict and HTTPHeaderDict.

Does that make sense?

V1NAY8: Yes. It makes sense. Thank you so much

shazow (commented):
Looking good! Though there is a conflict with main now, mind doing a rebase? :)

src/urllib3/request.py
```diff
@@ -1,3 +1,4 @@
+import json as json_serializer
```
sethmlarson, line 1: Can we alias this as a private name (ie `_json`) so it doesn't appear to be a public member of `urllib3.request`?

src/urllib3/request.py
```diff
@@ -68,6 +69,7 @@ def request(
         body: Optional[HTTPBody] = None,
         fields: Optional[_TYPE_FIELDS] = None,
         headers: Optional[Mapping[str, str]] = None,
+        json: Any = None,
```
sethmlarson, line 72: Should we use Optional[Any]? I believe the two are different somehow

src/urllib3/request.py
```diff
@@ -84,6 +86,21 @@ def request(
...
+
+        if json is not None:
+            if headers is None:
+                headers = self.headers.copy()
+            if not ("content-type" in map(str.lower, headers.keys())):
+                headers["Content-Type"] = "application/json"
+
+            body = json_serializer.dumps(
+                json, separators=(",", ":"), ensure_ascii=False
+            ).encode("utf-8")
```
sethmlarson, line 102: Unsure if we need to configure anything special in the encode() case like `surrogatepass`? What does Requests do?

src/urllib3/response.py
```diff
@@ -1,9 +1,11 @@
 import io
+import json
```
sethmlarson, line 2: Let's alias this import too so we don't have a name conflict with `HTTPResponse.json()`

src/urllib3/response.py
```diff
@@ -455,6 +457,13 @@ def data(self):
         if self._fp:
             return self.read(cache_content=True)
 
+    def json(self) -> Any:
```
sethmlarson, line 460: Because this function relies on `data` we can implement under `BaseHTTPResponse`

test/with_dummyserver/test_poolmanager.py
```diff
@@ -378,6 +379,31 @@ def test_top_level_request_with_body(self):
         assert r.status == 200
         assert r.data == b"test"
 
+    @pytest.mark.parametrize(
+        "headers", [None, {"content-Type": "application/json"}, {"attribute": "value"}]
```
sethmlarson, line 383: Would be good to have tests for different casings of "Content-Length" and with an HTTPHeaderDict too.

We should also make some assertation that the content-length header is sent and if already set to a value won't be overwritten.

sethmlarson (changes requested):
Thanks for this, here's a few more suggestions

src/urllib3/request.py
V1NAY8: Nope this is their logic, they are not using `errors="surrogatepass"`
```python
        if not data and json is not None:
            # urllib3 requires a bytes-like body. Python 2's json.dumps
            # provides this natively, but Python 3 gives a Unicode string.
            content_type = 'application/json'

            try:
              body = complexjson.dumps(json, allow_nan=False)
            except ValueError as ve:
              raise InvalidJSONError(ve, request=self)

            if not isinstance(body, bytes):
                body = body.encode('utf-8')
```

src/urllib3/response.py
V1NAY8: Sure

test/with_dummyserver/test_poolmanager.py
V1NAY8: Sure

sethmlarson (commented):
More various comments that I created but never submitted:

docs/user-guide.rst
```diff
@@ -269,8 +273,30 @@ dictionary in the ``fields`` argument provided to
 JSON
 ~~~~
 
-You can send a JSON request by specifying the encoded data as the ``body``
-argument and setting the ``Content-Type`` header when calling
+You can send a JSON request by specifying the data as ``json`` argument,
```
bluetech, line 276: A bunch of the code examples in the previous sections use JSON and handle it like

```py
    print(json.loads(resp.data.decode("utf-8"))["headers"])
```

These can be simplified with the new `json()` method. Perhaps this section should be moved up and then the following examples could use it.

src/urllib3/response.py
```diff
@@ -220,6 +221,16 @@ def get_redirect_location(self) -> typing.Union[str, "Literal[False]"]:
     def data(self) -> bytes:
         raise NotImplementedError()
 
+    def json(self) -> Any:
+        """
+        Parses the body of the HTTP response as UTF-8, UTF-16, or UTF-32 encoded JSON using the standard library :func:`json.loads` function.
```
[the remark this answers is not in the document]
pquentin: You're right. But we can't do this anyway, as the `encoding` parameter of `json.loads` was removed in Python 3.9. However, since we could support additional parsers in the future, including parsers that only support UTF-8 such as `orjson`, I would avoid documenting this at all.

```suggestion
        Parses the body of the HTTP response as JSON.
```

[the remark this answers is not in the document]
bluetech: I propose to enforce UTF-8, to follow the latest RFC and because I don't like "surprises" like my client accepting UTF-16/UTF-32 when it is not expected. Your point on orjson is also good, and I think enforcing UTF-8 would make it easier to use alternative JSON decoders consistently.

So I would do

```py
data = self.data.decode('utf-8')
return _json.loads(data)
```

In the docstring I would mention that the function can raise `UnicodeDecodeError` or `json.JSONDecodeError`.

orjson could skip the decode step.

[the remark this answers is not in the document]
pquentin: OK, that sounds good to me!

V1NAY8:
@pquentin I have done all the changes requested, 
Also changed all the references in tests, docs to use `json()` instead of `json.loads(r.data.decode())`

Let me know if any thing else is required.

pquentin (changes requested):
Sorry, this is not a full review, but I have a few comments already.

Can you please look into the failing tests? I think you just have to `json=None` in the correct place.

docs/user-guide.rst
```diff
@@ -296,7 +317,7 @@ argument and setting the ``Content-Type`` header when calling
         headers={"Content-Type": "application/json"}
     )
 
-    print(json.loads(resp.data.decode("utf-8"))["json"])
+    print(resp.json())
     # {"attribute": "value"}
```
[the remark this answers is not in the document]
V1NAY8: So, the above one is for `response` and the below one is for how to send a `request` am I missing something ?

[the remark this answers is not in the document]
pquentin: OK, but both sections start with "You can send a JSON request by specifying the data as ``json`` argument,". `JSON Content` should only talk about receiving JSON, not sending JSON.

src/urllib3/request.py
```diff
@@ -68,6 +69,7 @@ def request(
         body: Optional[_TYPE_BODY] = None,
         fields: Optional[_TYPE_FIELDS] = None,
         headers: Optional[Mapping[str, str]] = None,
+        json: Optional[Any] = None,
```
euri10, line 72: despite `Any` being the go-to solution with mypy unable to recursively type, I found this quite interesting:
https://github.com/python/typing/issues/182#issuecomment-870697089

maybe something to keep in mind

src/urllib3/request.py
```diff
@@ -84,6 +86,21 @@ def request(
...
+                "request got values for both 'body' and 'json' parameters which are mutually exclusive"
+            )
+
+        if json is not None:
+            if headers is None:
+                headers = self.headers.copy()  # type: ignore
+            if not ("content-type" in map(str.lower, headers.keys())):
+                headers["Content-Type"] = "application/json"  # type: ignore
+
+            body = _json.dumps(json, separators=(",", ":"), ensure_ascii=False).encode(
```
euri10, line 100: do we really want / need to eliminate whitespaces ?

V1NAY8: It was part of requirement. I thought it was used to reduce size of payload.

src/urllib3/request.py
vanburgerberg: Check this https://github.com/python/typing/issues/182#issuecomment-893657366

sethmlarson: I think we'll go with `Any` for now until typing has made a decision about a standardized JSON type.

pquentin (changes requested):
OK, I'm confident I've covered everything now. :) Sorry for the long delay.

sethmlarson (dismissed):
Going to give my +1 on this PR now, thanks to everyone involved. Merge when other reviewers are ready.

bluetech (dismissed):

sethmlarson:
Okay now with a changelog entry this is ready for merge, @V1NAY8 could you rebase this on `main`? I think the only conflicting change you need is the `_request_methods.py` rename.

V1NAY8:
I have rebased this PR

Now mypy is failing in `handlers.py` because of the following logic:
```python
        if json is not None:
            self.headers = headers
            self.body = json
        else:
            self.headers = headers or [("Content-type", "text/plain")]
```
If headers are passed as None, we have logic to automatically set them to `("Content-type", "application/json")` We assert this scenario too

Since headers can be None => mypy is throwing Error

If I set the headers to headers or [("Content-type", "application/json")]. This will never hit the 
This part of code in `_request_methods.py`
```python
            if not ("content-type" in map(str.lower, headers.keys())):
                headers["Content-Type"] = "application/json"  # type: ignore
```
Because by default this mock API is going to set the headers.

This will result in less code coverage. So, can we type ignore at line 41 at `handlers.py`?

sethmlarson:
@V1NAY8 I think you can add ` or [("Content-type", "application/json")]` to the `headers` line in the `if json is not None` in `dummyserver/handlers.py`. I don't think we'll lose coverage by making this change since the line you're mentioning is on the request side, whereas the handler logic is to do with the response.

sethmlarson (approved):
LGTM, will merge after CI passes.

sethmlarson:
:tada::partying_face::tada: This is incredible, thank you to everyone involved on this feature! Can't wait to start using it myself in v2.0.

V1NAY8:
Thank you @sethmlarson, @pquentin @bluetech @shazow @euri10 @vanburgerberg for the review, guidance, and patience!

💯

shazow:
Woohoo!

PR #2251, opened 2021-06-02
Add type hints to urllib3.connectionpool

hramezani:
```
src/urllib3/connectionpool.py:1037: error: Argument 1 to "HTTPSConnectionPool" has incompatible type "Optional[str]"; expected "str"
src/urllib3/connectionpool.py:1039: error: Argument 1 to "HTTPConnectionPool" has incompatible type "Optional[str]"; expected "str"
src/urllib3/connectionpool.py:12: error: Module 'urllib3.connection' does not explicitly export attribute 'HTTPException'; implicit reexport disabled
src/urllib3/connectionpool.py:235: error: Argument "timeout" to "HTTPConnection" has incompatible type "object"; expected "Optional[float]"
src/urllib3/connectionpool.py:386: error: Incompatible types in assignment (expression has type "object", variable has type "Optional[float]")
src/urllib3/connectionpool.py:448: error: "HTTPConnection" has no attribute "_http_vsn_str"
src/urllib3/connectionpool.py:450: error: "HTTPResponse" has no attribute "length"
src/urllib3/connectionpool.py:673: error: Argument 1 to "set_file_position" has incompatible type "Union[bytes, IO[Any], Iterable[bytes], str, None]"; expected "IO[Any]"
src/urllib3/connectionpool.py:680: error: Incompatible types in assignment (expression has type "object", variable has type "Optional[float]")
src/urllib3/connectionpool.py:932: error: Argument "cert_reqs" to "set_cert" of "HTTPSConnection" has incompatible type "Union[int, str, None]"; expected "Optional[int]"
src/urllib3/connectionpool.py:938: error: Incompatible types in assignment (expression has type "Union[int, str, None]", variable has type "Optional[int]")
src/urllib3/connectionpool.py:939: error: Incompatible return value type (got "urllib3.connection.HTTPSConnection", expected "http.client.HTTPSConnection")
src/urllib3/connectionpool.py:968: error: Non-overlapping identity check (left operand type: "Type[HTTPSConnection]", right operand type: "Type[DummyConnection]")
src/urllib3/connectionpool.py:982: error: Argument "timeout" has incompatible type "object"; expected "Optional[float]".```

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -5,7 +5,6 @@ import socket
 import sys
 import warnings
 from http.client import HTTPResponse as _HttplibHTTPResponse
-from http.client import HTTPSConnection as _HttplibHTTPSConnection
 from socket import timeout as SocketTimeout
 from typing import TYPE_CHECKING, Any, Mapping, Optional, Type, Union, overload
 
@@ -232,7 +231,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         conn = self.ConnectionCls(
             host=self.host,
             port=self.port,
-            timeout=self.timeout.connect_timeout,
+            timeout=self.timeout.connect_timeout,  # type: ignore
             **self.conn_kw,
         )
         return conn
@@ -256,6 +255,10 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         try:
             conn = self.pool.get(block=self.block, timeout=timeout)
+
+        except AttributeError:  # self.pool is None
+            raise ClosedPoolError(self, "Pool is closed.")
+
         except queue.Empty:
             if self.block:
                 raise EmptyPoolError(
@@ -294,6 +297,9 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             try:
                 self.pool.put(conn, block=False)
                 return  # Everything is dandy, done.
+            except AttributeError:
+                # self.pool is None.
+                pass
             except queue.Full:
 
                 # Connection never got put back into the pool, close it.
@@ -383,7 +389,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
 
         timeout_obj = self._get_timeout(timeout)
         timeout_obj.start_connect()
-        conn.timeout = timeout_obj.connect_timeout
+        conn.timeout = timeout_obj.connect_timeout  # type: ignore
 
         # Trigger any extra validation we need to do.
         try:
@@ -677,7 +683,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             timeout_obj = self._get_timeout(timeout)
             conn = self._get_conn(timeout=pool_timeout)
 
-            conn.timeout = timeout_obj.connect_timeout
+            conn.timeout = timeout_obj.connect_timeout  # type: ignore
 
             is_new_proxy_conn = self.proxy is not None and not getattr(
                 conn, "sock", None
@@ -918,7 +924,7 @@ class HTTPSConnectionPool(HTTPConnectionPool):
         self.assert_hostname = assert_hostname
         self.assert_fingerprint = assert_fingerprint
 
-    def _prepare_conn(self, conn: HTTPSConnection) -> _HttplibHTTPSConnection:
+    def _prepare_conn(self, conn: HTTPSConnection) -> HTTPConnection:
         """
         Prepare the ``connection`` for :meth:`urllib3.util.ssl_wrap_socket`
         and establish the tunnel if proxy is used.
@@ -953,9 +959,9 @@ class HTTPSConnectionPool(HTTPConnectionPool):
 
         conn.connect()
 
-    def _new_conn(self) -> _HttplibHTTPSConnection:  # type: ignore
+    def _new_conn(self) -> HTTPConnection:
         """
-        Return a fresh :class:`http.client.HTTPSConnection`.
+        Return a fresh :class:`urllib3.connection.HTTPConnection`.
         """
         self.num_connections += 1
         log.debug(
@@ -965,7 +971,7 @@ class HTTPSConnectionPool(HTTPConnectionPool):
             self.port or "443",
         )
 
-        if not self.ConnectionCls or self.ConnectionCls is DummyConnection:
+        if not self.ConnectionCls or self.ConnectionCls is DummyConnection:  # type: ignore
             raise SSLError(
                 "Can't connect to HTTPS URL because the SSL module is not available."
             )
@@ -979,7 +985,7 @@ class HTTPSConnectionPool(HTTPConnectionPool):
         conn = self.ConnectionCls(
             host=actual_host,
             port=actual_port,
-            timeout=self.timeout.connect_timeout,
+            timeout=self.timeout.connect_timeout,  # type: ignore
             cert_file=self.cert_file,
             key_file=self.key_file,
             key_password=self.key_password,
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -4,15 +4,20 @@ import queue
 import socket
 import sys
 import warnings
+from http.client import HTTPResponse as _HttplibHTTPResponse
+from http.client import HTTPSConnection as _HttplibHTTPSConnection
 from socket import timeout as SocketTimeout
+from typing import TYPE_CHECKING, Any, Mapping, Optional, Type, Union
 
 from .connection import (
     BaseSSLError,
     BrokenPipeError,
     DummyConnection,
+    HTTPBody,
     HTTPConnection,
     HTTPException,
     HTTPSConnection,
+    ProxyConfig,
     VerifiedHTTPSConnection,
     port_by_scheme,
 )
@@ -47,9 +52,15 @@ from .util.url import _normalize_host as normalize_host
 from .util.url import parse_url
 from .util.util import to_str
 
+if TYPE_CHECKING:
+    from typing_extensions import Literal
+
 log = logging.getLogger(__name__)
 
 _Default = object()
+
+
+_TYPE_TIMEOUT = Union[Timeout, int, float, object]
 
 
 # Pool objects
@@ -64,10 +75,10 @@ class ConnectionPool:
        target URIs.
     """
 
-    scheme = None
+    scheme: Optional[str] = None
     QueueCls = queue.LifoQueue
 
-    def __init__(self, host, port=None):
+    def __init__(self, host: str, port: Optional[int] = None) -> None:
         if not host:
             raise LocationValueError("No host specified.")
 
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/urllib3/connectionpool.py [hunk 1]
```diff
@@ -64,29 +72,31 @@ class ConnectionPool:
...
     """
 
-    scheme = None
+    scheme: Optional[str] = None
     QueueCls = queue.LifoQueue
 
-    def __init__(self, host, port=None):
-        if not host:
+    def __init__(self, host: Optional[str], port: Optional[int] = None) -> None:
+        if host is None:
```
sethmlarson, line 79: This changes functionality for empty hosts, we should keep this branch the same.

src/urllib3/connectionpool.py
```diff
@@ -64,29 +72,31 @@ class ConnectionPool:
        target URIs.
     """
 
-    scheme = None
+    scheme: Optional[str] = None
     QueueCls = queue.LifoQueue
 
-    def __init__(self, host, port=None):
-        if not host:
+    def __init__(self, host: Optional[str], port: Optional[int] = None) -> None:
```
sethmlarson, line 78: Any reason to have `Optional[str]` instead of `str` if we always error on this case?

src/urllib3/connectionpool.py
```diff
@@ -150,22 +160,22 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
...
-        timeout=Timeout.DEFAULT_TIMEOUT,
-        maxsize=1,
-        block=False,
-        headers=None,
-        retries=None,
-        _proxy=None,
-        _proxy_headers=None,
-        _proxy_config=None,
-        **conn_kw,
+        host: Optional[str],
```
sethmlarson, line 168: same as above, do we need Optional?

src/urllib3/connectionpool.py
```diff
@@ -150,22 +160,22 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
...
+        port: Optional[int] = None,
+        timeout: object = Timeout.DEFAULT_TIMEOUT,
```
sethmlarson, line 170: Given how often we're doing `timeout` maybe we can add an alias for the type? `_TYPE_TIMEOUT`?

src/urllib3/connectionpool.py
```diff
@@ -224,7 +234,7 @@ def _new_conn(self):
         )
         return conn
 
-    def _get_conn(self, timeout=None):
+    def _get_conn(self, timeout: Optional[int] = None) -> HTTPConnection:
```
sethmlarson, line 237: Should this timeout be `float`?

src/urllib3/connectionpool.py
```diff
@@ -633,8 +654,8 @@ def urlopen(
         # have to copy the headers dict so we can safely change it without those
         # changes being reflected in anyone else's copy.
         if not http_tunnel_required:
-            headers = headers.copy()
-            headers.update(self.proxy_headers)
+            headers = headers.copy()  # type: ignore
```
sethmlarson, line 657: Ugh, this type: ignore is annoying because our APIs are clearly favoring `dict` instead of `Mapping`. Can't wait to switch to `HTTPHeaderDict` everywhere...

src/urllib3/connectionpool.py
```diff
@@ -733,7 +754,9 @@ def urlopen(
                 # to throw the connection away unless explicitly told not to.
                 # Close the connection, set the variable to None, and make sure
                 # we put the None back in the pool to avoid leaking it.
-                conn = conn and conn.close()
+                if conn:
```
sethmlarson, line 757: Way more readable! :sparkles:

src/urllib3/connectionpool.py
```diff
@@ -848,26 +871,26 @@ class HTTPSConnectionPool(HTTPConnectionPool):
...
+        timeout: Union[Timeout, int, float, object] = Timeout.DEFAULT_TIMEOUT,
+        maxsize: int = 1,
+        block: bool = False,
+        headers: Optional[Mapping[str, str]] = None,
+        retries: Optional[Union[Retry, bool, int]] = None,
+        _proxy: Optional[Url] = None,
+        _proxy_headers: Optional[Mapping[str, str]] = None,
+        key_file: Optional[str] = None,
+        cert_file: Optional[str] = None,
+        cert_reqs: Optional[int] = None,
```
sethmlarson, line 885: `Union[str, int]`?

src/urllib3/connectionpool.py
```diff
@@ -848,26 +871,26 @@ class HTTPSConnectionPool(HTTPConnectionPool):
...
+        key_password: Optional[str] = None,
+        ca_certs: Optional[str] = None,
+        ssl_version: Optional[int] = None,
```
sethmlarson, line 888: `Union[str, int]`?

src/urllib3/connectionpool.py
```diff
@@ -892,7 +915,7 @@ def __init__(
         self.assert_hostname = assert_hostname
         self.assert_fingerprint = assert_fingerprint
 
-    def _prepare_conn(self, conn):
+    def _prepare_conn(self, conn: HTTPSConnection) -> ClientHTTPSConnection:
```
sethmlarson, line 918: Why `ClientHTTPSConnection` instead of `HTTPSConnection` here?

src/urllib3/connectionpool.py
```diff
@@ -922,12 +945,12 @@ def _prepare_proxy(self, conn):
...
         conn.set_tunnel(self._proxy_host, self.port, self.proxy_headers)
 
-        if self.proxy.scheme == "https":
+        if self.proxy and self.proxy.scheme == "https":
             conn.tls_in_tls_required = True
 
         conn.connect()
 
-    def _new_conn(self):
+    def _new_conn(self) -> ClientHTTPSConnection:  # type: ignore
```
sethmlarson, line 953: Same as above, why are we using `ClientHTTPSConnection`?

src/urllib3/connectionpool.py
```diff
@@ -1013,7 +1036,7 @@ def connection_from_url(url, **kw):
         return HTTPConnectionPool(host, port=port, **kw)
 
 
-def _normalize_host(host, scheme):
+def _normalize_host(host: Optional[str], scheme: Optional[str]) -> Optional[str]:
```
sethmlarson, line 1039: `host` should be `str` here and return `str`?

src/urllib3/response.py
```diff
@@ -3,7 +3,9 @@
 import typing
 import zlib
 from contextlib import contextmanager
+from http.client import HTTPResponse as ClientHTTPResponse
```
sethmlarson, line 6: Let's import this as `_HttplibHTTPResponse`

sethmlarson (changes requested):
Another tough file, nice work! :clap: Here's a bunch of comments and questions for you:

src/urllib3/connectionpool.py
hramezani: Without `Optional`, we get
`src/urllib3/connectionpool.py:1036: error: Argument 1 to "HTTPConnectionPool" has incompatible type "Optional[str]"; expected "str"`

src/urllib3/connectionpool.py
hramezani: we have `Return a fresh :class:`http.client.HTTPSConnection`.` in docstring.
Isn't it right?

src/urllib3/connectionpool.py
hramezani: because `util.url._normalize_host` returns `Optional[str]`

src/urllib3/connectionpool.py
hramezani: changed reverted.

src/urllib3/connectionpool.py
hramezani: changed to str

src/urllib3/connectionpool.py
hramezani: Yes, `_TYPE_TIMEOUT` added.

src/urllib3/connectionpool.py
hramezani: changed.

src/urllib3/connectionpool.py
hramezani: Done

src/urllib3/connectionpool.py
hramezani: Done

src/urllib3/response.py
hramezani: Done

hramezani:
Thanks @sethmlarson for the review.
I addressed your comments and aswered some of them. we have to think about the errors as well.

src/urllib3/connectionpool.py
sethmlarson: Ah gotcha, so a lot of this stems from there. Let's add an `@override` for `_normalize_host` where `None` returns `None` and `str` returns `str`? Then if we specify `str` here we don't have to return `Optional[str]`, does this solve the problem?

src/urllib3/connectionpool.py
hramezani: Should we revert it to `Optional[int]`?
After changing to `Optional[Union[int, str]]` we got:
`src/urllib3/connectionpool.py:932: error: Argument "cert_reqs" to "set_cert" of "HTTPSConnection" has incompatible type "Union[int, str, None]"; expected "Optional[int]"`

src/urllib3/connectionpool.py
hramezani: after this change we got:

`src/urllib3/connectionpool.py:938: error: Incompatible types in assignment (expression has type "Union[int, str, None]", variable has type "Optional[int]")`

src/urllib3/connectionpool.py
```diff
@@ -9,7 +9,7 @@
 from socket import timeout as SocketTimeout
 from typing import TYPE_CHECKING, Any, Mapping, Optional, Type, Union, overload
 
-from .connection import (
+from .connection import (  # type: ignore
```
sethmlarson, line 12: Are we allowed to put this on the `HTTPException` directly?

src/urllib3/util/request.py
```diff
@@ -103,6 +103,20 @@ def make_headers(
...
 
+@overload
+def set_file_position(
+    body: IO[Any], pos: Optional[Union[int, object]]
+) -> Optional[Union[int, object]]:
+    ...
+
+
+@overload
+def set_file_position(
```
sethmlarson, line 114: Hmm, we might have to remove the `@overload` here and just go with `body: Any, pos: Optional[Union[int, object]] -> Optional[Union[int, object]]`

sethmlarson (changes requested):
Two things then I think this is good?

sethmlarson (approved):
LGTM, thanks much!!

PR #2255, opened 2021-06-07
Add type hints to urllib3.response

hramezani:
```
src/urllib3/response.py:152: error: Returning Any from function declared to return "bytes"
src/urllib3/response.py:212: error: Argument 1 to "HTTPHeaderDict" has incompatible type "Optional[Mapping[bytes, bytes]]"; expected "Union[HTTPHeaderDict, Mapping[str, str], Iterable[Tuple[str, str]], HasGettableStringKeys, None]"
src/urllib3/response.py:441: error: Incompatible types in assignment (expression has type "Union[bytes, IO[Any], Iterable[bytes], str]", variable has type "Optional[HTTPResponse]")
src/urllib3/response.py:726: error: Incompatible types in assignment (expression has type "HTTPHeaderDict", variable has type "HTTPMessage")
src/urllib3/response.py:730: error: Argument "headers" to "HTTPResponse" has incompatible type "HTTPMessage"; expected "Optional[Mapping[AnyStr, AnyStr]]"
src/urllib3/response.py:753: error: "Callable[[IOBase], bool]" has no attribute "__get__"
src/urllib3/response.py:753: error: Returning Any from function declared to return "bool"
src/urllib3/response.py:796: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "fp"
src/urllib3/response.py:796: error: Item "None" of "Optional[HTTPResponse]" has no attribute "fp"
src/urllib3/response.py:808: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:808: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:810: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:810: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:813: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:813: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:817: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:817: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:818: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:818: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:822: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:822: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:823: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:823: error: Item "None" of "Optional[HTTPResponse]" has no attribute "_safe_read"
src/urllib3/response.py:825: error: Returning Any from function declared to return "bytes"
src/urllib3/response.py:864: error: Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "fp"
src/urllib3/response.py:864: error: Item "None" of "Optional[HTTPResponse]" has no attribute "fp"
src/urllib3/response.py:86: error: Incompatible types in assignment (expression has type "None", variable has type "bytes")
src/urllib3/response.py:888: error: "HTTPResponse" has no attribute "fp"
src/urllib3/response.py:94: error: Incompatible types in assignment (expression has type "None", variable has type "bytes").```

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -313,7 +313,7 @@ class IncompleteRead(HTTPError, httplib_IncompleteRead):
     partial: int
     expected: int
 
-    def __init__(self, partial: int, expected: Optional[int]) -> None:
+    def __init__(self, partial: int, expected: int) -> None:
         super().__init__(partial, expected)
 
     def __repr__(self) -> str:
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -32,13 +32,9 @@ _TYPE_REDUCE_RESULT = Tuple[Callable[..., object], Tuple[object, ...]]
 class PoolError(HTTPError):
     """Base exception for errors caused within a pool."""
 
-    pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]]
+    pool: "ConnectionPool"
 
-    def __init__(
-        self,
-        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
-        message: str,
-    ) -> None:
+    def __init__(self, pool: "ConnectionPool", message: str) -> None:
         self.pool = pool
         super().__init__(f"{pool}: {message}")
 
@@ -50,14 +46,9 @@ class PoolError(HTTPError):
 class RequestError(PoolError):
     """Base exception for PoolErrors that have associated URLs."""
 
-    url: Optional[str]
+    url: str
 
-    def __init__(
-        self,
-        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
-        url: Optional[str],
-        message: str,
-    ) -> None:
+    def __init__(self, pool: "ConnectionPool", url: str, message: str) -> None:
         self.url = url
         super().__init__(pool, message)
 
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -32,9 +32,13 @@ _TYPE_REDUCE_RESULT = Tuple[Callable[..., object], Tuple[object, ...]]
 class PoolError(HTTPError):
     """Base exception for errors caused within a pool."""
 
-    pool: "ConnectionPool"
+    pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]]
 
-    def __init__(self, pool: "ConnectionPool", message: str) -> None:
+    def __init__(
+        self,
+        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
+        message: str,
+    ) -> None:
         self.pool = pool
         super().__init__(f"{pool}: {message}")
 
@@ -46,9 +50,14 @@ class PoolError(HTTPError):
 class RequestError(PoolError):
     """Base exception for PoolErrors that have associated URLs."""
 
-    url: str
+    url: Optional[str]
 
-    def __init__(self, pool: "ConnectionPool", url: str, message: str) -> None:
+    def __init__(
+        self,
+        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
+        url: Optional[str],
+        message: str,
+    ) -> None:
         self.url = url
         super().__init__(pool, message)
 
@@ -304,7 +313,7 @@ class IncompleteRead(HTTPError, httplib_IncompleteRead):
     partial: int
     expected: int
 
-    def __init__(self, partial: int, expected: int) -> None:
+    def __init__(self, partial: int, expected: Optional[int]) -> None:
         super().__init__(partial, expected)
 
     def __repr__(self) -> str:
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -1,22 +1,39 @@
 import io
 import logging
-import typing
 import zlib
 from contextlib import contextmanager
+from http.client import HTTPMessage as _HttplibHTTPMessage
 from http.client import HTTPResponse as _HttplibHTTPResponse
 from socket import timeout as SocketTimeout
-from typing import Any, Type
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AnyStr,
+    Generator,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
 try:
     try:
-        import brotlicffi as brotli
+        import brotlicffi as brotli  # type: ignore
     except ImportError:
-        import brotli
+        import brotli  # type: ignore
 except ImportError:
     brotli = None
 
 from ._collections import HTTPHeaderDict
-from .connection import BaseSSLError, HTTPException
+from .connection import (  # type: ignore
+    BaseSSLError,
+    HTTPBody,
+    HTTPConnection,
+    HTTPException,
+)
 from .exceptions import (
     BodyNotHttplibCompatible,
     DecodeError,
@@ -30,9 +47,12 @@ from .exceptions import (
     SSLError,
 )
 from .util.response import is_fp_closed, is_response_to_head
+from .util.retry import Retry
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from typing_extensions import Literal
+
+    from urllib3.connectionpool import HTTPConnectionPool
 
 log = logging.getLogger(__name__)
 
@@ -46,12 +66,12 @@ class ContentDecoder:
 
 
 class DeflateDecoder(ContentDecoder):
-    def __init__(self):
+    def __init__(self) -> None:
         self._first_try = True
         self._data = b""
         self._obj = zlib.decompressobj()
 
-    def decompress(self, data):
+    def decompress(self, data: bytes) -> bytes:
         if not data:
             return data
 
@@ -85,11 +105,11 @@ class GzipDecoderState:
 
 
 class GzipDecoder(ContentDecoder):
-    def __init__(self):
+    def __init__(self) -> None:
         self._obj = zlib.decompressobj(16 + zlib.MAX_WBITS)
         self._state = GzipDecoderState.FIRST_MEMBER
 
-    def decompress(self, data):
+    def decompress(self, data: bytes) -> bytes:
         ret = bytearray()
         if self._state == GzipDecoderState.SWALLOW_DATA or not data:
             return bytes(ret)
@@ -120,14 +140,14 @@ if brotli is not None:
         # Supports both 'brotlipy' and 'Brotli' packages
         # since they share an import name. The top branches
         # are for 'brotlipy' and bottom branches for 'Brotli'
-        def __init__(self):
+        def __init__(self) -> None:
             self._obj = brotli.Decompressor()
             if hasattr(self._obj, "decompress"):
-                self.decompress = self._obj.decompress
+                setattr(self, "decompress", self._obj.decompress)
             else:
-                self.decompress = self._obj.process
+                setattr(self, "decompress", self._obj.process)
 
-        def flush(self):
+        def flush(self) -> bytes:
             if hasattr(self._obj, "flush"):
                 return self._obj.flush()
             return b""
@@ -142,13 +162,13 @@ class MultiDecoder(ContentDecoder):
         they were applied.
     """
 
-    def __init__(self, modes):
+    def __init__(self, modes: str) -> None:
         self._decoders = [_get_decoder(m.strip()) for m in modes.split(",")]
 
-    def flush(self):
+    def flush(self) -> bytes:
         return self._decoders[0].flush()
 
-    def decompress(self, data):
+    def decompress(self, data: bytes) -> bytes:
         for d in reversed(self._decoders):
             data = d.decompress(data)
         return data
@@ -173,17 +193,17 @@ class BaseHTTPResponse(io.IOBase):
         CONTENT_DECODERS += ["br"]
     REDIRECT_STATUSES = [301, 302, 303, 307, 308]
 
-    DECODER_ERROR_CLASSES = (IOError, zlib.error)
+    DECODER_ERROR_CLASSES: Tuple[Any, ...] = (IOError, zlib.error)
     if brotli is not None:
         DECODER_ERROR_CLASSES += (brotli.error,)
 
     def __init__(
         self,
         *,
-        headers: typing.Optional[typing.Mapping[typing.AnyStr, typing.AnyStr]] = None,
+        headers: Optional[Mapping[AnyStr, AnyStr]] = None,
         status: int,
         version: int,
-        reason: str,
+        reason: Optional[str],
         decode_content: bool,
     ) -> None:
         if isinstance(headers, HTTPHeaderDict):
@@ -202,9 +222,9 @@ class BaseHTTPResponse(io.IOBase):
         if "chunked" in encodings:
             self.chunked = True
 
-        self._decoder: typing.Optional[ContentDecoder] = None
+        self._decoder: Optional[ContentDecoder] = None
 
-    def get_redirect_location(self) -> typing.Union[str, "Literal[False]"]:
+    def get_redirect_location(self) -> Union[Optional[str], "Literal[False]"]:
         """
         Should we redirect and where to?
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/urllib3/response.py [hunk 1]
```diff
@@ -449,19 +469,21 @@ def drain_conn(self) -> None:
             pass
 
     @property
-    def data(self):
+    def data(self) -> Optional[Union[bytes, str]]:
```
sethmlarson, line 472: Should this be `-> bytes`?

src/urllib3/response.py
```diff
@@ -217,39 +237,39 @@ def get_redirect_location(self) -> typing.Union[str, "Literal[False]"]:
         return False
 
     @property
-    def data(self) -> bytes:
+    def data(self) -> Optional[Union[bytes, str]]:
```
sethmlarson, line 240: Should this be -> bytes?

src/urllib3/response.py
```diff
@@ -472,11 +494,11 @@ def tell(self) -> int:
         """
         return self._fp_bytes_read
 
-    def _init_length(self, request_method):
+    def _init_length(self, request_method: Optional[str]) -> Optional[int]:
         """
         Set initial length value for Response content if available.
         """
-        length = self.headers.get("content-length")
+        length: Optional[Union[int, str]] = self.headers.get("content-length")
```
sethmlarson, line 501: Having this typed here as `Optional[Union[int, str]]` instead of `Optional[str]` is causing a `str()` call below to be necessary. Let's have this variable be named `content_length` and keep `length` as it is elsewhere but define it above as `length: Optional[int]`?

src/urllib3/response.py
```diff
@@ -865,7 +896,7 @@ def read_chunked(self, amt=None, decode_content=None):
                 self._original_response.close()
 
     @property
-    def url(self) -> str:
+    def url(self) -> Optional[Union[str, "Literal[False]"]]:
```
sethmlarson, line 899: When would this be `False`?

src/urllib3/util/retry.py
```diff
@@ -19,16 +18,21 @@
...
 # Data structure for representing the metadata of requests that result in a retry.
-RequestHistory = namedtuple(
-    "RequestHistory", ["method", "url", "error", "status", "redirect_location"]
-)
+class RequestHistory(NamedTuple):
+    method: Optional[str]
+    url: Optional[str]
+    error: Optional[Exception]
+    status: Optional[int]
+    redirect_location: Optional[Union[str, "Literal[False]"]]
```
sethmlarson, line 35: When is this `False`? I think we don't add a `redirect_location` if `False` is given

sethmlarson (changes requested):
Looks pretty good, few comments for you:

src/urllib3/response.py
```diff
@@ -173,17 +193,17 @@ class BaseHTTPResponse(io.IOBase):
         CONTENT_DECODERS += ["br"]
     REDIRECT_STATUSES = [301, 302, 303, 307, 308]
 
-    DECODER_ERROR_CLASSES = (IOError, zlib.error)
+    DECODER_ERROR_CLASSES: Tuple[Any, ...] = (IOError, zlib.error)
```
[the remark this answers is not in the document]
hramezani: Then we have 
```
src/urllib3/response.py:196: error: Incompatible types in assignment (expression has type "Tuple[Type[OSError], Type[error]]", variable has type "Tuple[BaseException, ...]")
src/urllib3/response.py:315: error: Exception type must be derived from BaseException
```

src/urllib3/response.py
hramezani: we have `return self._body` in this function
https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L455

And `_body` can be `Union[str, bytes]` https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L414-L415
Also, The function doesn't have a return(I added a `return None`) 
So, I changed the return type to `Optional[Union[bytes, str]]`

src/urllib3/response.py
hramezani: to be compatible with the `HTTPResponse.data` property.

src/urllib3/response.py
hramezani: Then we have `UnboundLocalError: local variable 'length' referenced before assignment` error in 
https://github.com/urllib3/urllib3/blob/2ec06d1185f1642dfe8ff767a9ad3541bd0ddc39/src/urllib3/response.py#L525

when `content-length` is `None`.

src/urllib3/response.py
hramezani: changed to `Optional[str]`

src/urllib3/util/retry.py
hramezani: Yes, You are right. I changed it to `Optional[str]` and changed the `increment` function to handle it.

src/urllib3/exceptions.py
```diff
@@ -286,7 +295,7 @@ class IncompleteRead(HTTPError, httplib_IncompleteRead):
     partial: int
     expected: int
```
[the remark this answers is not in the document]
hramezani: I reverted the constructor to accept `int` and changed the `read` function to fix it.

src/urllib3/response.py
sethmlarson: I was going to try to explain in text but it's way faster to just copy-paste and make the edit:

```python
length: Optional[int]
content_length: Optional[str] = self.headers.get("content-length")

if content_length is not None:
    if self.chunked:
        # This Response will fail with an IncompleteRead if it can't be
        # received as chunked. This method falls back to attempt reading
        # the response before raising an exception.
        log.warning(
            "Received response with both Content-Length and "
            "Transfer-Encoding set. This is expressly forbidden "
            "by RFC 7230 sec 3.3.2. Ignoring Content-Length and "
            "attempting to process response as Transfer-Encoding: "
            "chunked."
        )
        return None

    try:
        # RFC 7230 section 3.3.2 specifies multiple content lengths can
        # be sent in a single Content-Length header
        # (e.g. Content-Length: 42, 42). This line ensures the values
        # are all valid ints and that as long as the `set` length is 1,
        # all values are the same. Otherwise, the header is invalid.
        lengths = {int(val) for val in content_length.split(",")}
        if len(lengths) > 1:
            raise InvalidHeader(
                "Content-Length contained multiple "
                "unmatching values (%s)" % content_length
            )
        length = lengths.pop()
    except ValueError:
        length = None
    else:
        if length < 0:
            length = None

else:  # if content_length is None
    length = None

# Convert status to int for comparison
# In some cases, httplib returns a status of "_UNKNOWN"
try:
    status = int(self.status)
except ValueError:
    status = 0

# Check for responses that shouldn't include a body
if status in (204, 304) or 100 <= status < 200 or request_method == "HEAD":
    length = 0

return length
```

Basically the change is the `else: length=None`, does this work?

bluetech (commented):
Great improvements. I left some suggestions.

src/urllib3/exceptions.py
```diff
@@ -29,9 +29,13 @@ class HTTPWarning(Warning):
 class PoolError(HTTPError):
     """Base exception for errors caused within a pool."""
 
-    pool: "ConnectionPool"
+    pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]]
 
-    def __init__(self, pool: "ConnectionPool", message: str) -> None:
+    def __init__(
+        self,
+        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
```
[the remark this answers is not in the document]
sethmlarson: The base class has so few attributes, almost to the point of not being useful anymore. Is this why we're doing this construction @hramezani?

src/urllib3/response.py
hramezani: after changing the return type to `bytes`, I got:
```
src/urllib3/response.py:475: error: Incompatible return value type (got "Union[str, bytes]", expected "bytes")
src/urllib3/response.py:478: error: Incompatible return value type (got "Optional[bytes]", expected "bytes")
src/urllib3/response.py:480: error: Incompatible return value type (got "None", expected "bytes")
```

We have to ignore them for now.

src/urllib3/response.py
```diff
@@ -772,14 +802,14 @@ def _update_chunk_length(self):
...
-    def _handle_chunk(self, amt):
+    def _handle_chunk(self, amt: Optional[int]) -> bytes:
         returned_chunk = None
         if amt is None:
             chunk = self._fp._safe_read(self.chunk_left)
             returned_chunk = chunk
             self._fp._safe_read(2)  # Toss the CRLF at the end of the chunk.
             self.chunk_left = None
-        elif amt < self.chunk_left:
+        elif self.chunk_left is not None and amt < self.chunk_left:
```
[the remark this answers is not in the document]
hramezani: I am not sure but I think it can be `None`. because based on the code [_update_chunk_length](https://github.com/urllib3/urllib3/blob/dfd8b580adb4f87388310c11b5c76d989bd6c4e3/src/urllib3/response.py#L761) will be called before [_handle_chunk](https://github.com/urllib3/urllib3/blob/dfd8b580adb4f87388310c11b5c76d989bd6c4e3/src/urllib3/response.py#L775), and there is:
```
            return
```
So, it seems it can be `None`.

src/urllib3/response.py
```diff
@@ -173,17 +193,17 @@ class BaseHTTPResponse(io.IOBase):
...
-    DECODER_ERROR_CLASSES = (IOError, zlib.error)
+    DECODER_ERROR_CLASSES: Tuple[Any, ...] = (IOError, zlib.error)
     if brotli is not None:
         DECODER_ERROR_CLASSES += (brotli.error,)
 
     def __init__(
         self,
         *,
-        headers: typing.Optional[typing.Mapping[typing.AnyStr, typing.AnyStr]] = None,
+        headers: Optional[Mapping[AnyStr, AnyStr]] = None,
```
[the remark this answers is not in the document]
hramezani: Then we get:
```
src/urllib3/response.py:211: error: Argument 1 to "HTTPHeaderDict" has incompatible type "Union[Mapping[str, str], Mapping[bytes, bytes], None]"; expected "Union[HTTPHeaderDict, Mapping[str, str], Iterable[Tuple[str, str]], HasGettableStringKeys, None]"
```

src/urllib3/response.py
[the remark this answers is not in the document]
hramezani: Done!

src/urllib3/response.py
hramezani: Thanks @sethmlarson, I've replaced your cone block

src/urllib3/response.py
```diff
@@ -217,39 +237,39 @@ def get_redirect_location(self) -> typing.Union[str, "Literal[False]"]:
...
     @property
-    def connection(self):
+    def connection(self) -> Optional[HTTPConnection]:
         raise NotImplementedError()
 
     def stream(
-        self, amt: int = 2 ** 16, decode_content: typing.Optional[bool] = None
-    ) -> typing.Generator[bytes, None, None]:
+        self, amt: int = 2 ** 16, decode_content: Optional[bool] = None
+    ) -> Generator[bytes, None, None]:
```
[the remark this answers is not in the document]
hramezani: Done!

src/urllib3/response.py
```diff
@@ -375,23 +395,23 @@ class is also compatible with the Python standard library's :mod:`io`
...
-        connection=None,
-        msg=None,
-        retries=None,
-        enforce_content_length=False,
-        request_method=None,
-        request_url=None,
-        auto_close=True,
-    ):
+        body: HTTPBody = "",
+        headers: Optional[Mapping[AnyStr, AnyStr]] = None,
```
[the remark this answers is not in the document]
hramezani: Done!

src/urllib3/response.py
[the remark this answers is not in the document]
hramezani: Done!

src/urllib3/exceptions.py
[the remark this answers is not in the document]
hramezani: I think we don't need this change anymore. SO, I've reverted this.

src/urllib3/exceptions.py
```diff
@@ -43,9 +47,14 @@ def __reduce__(self) -> ReduceResult:
 class RequestError(PoolError):
     """Base exception for PoolErrors that have associated URLs."""
 
-    url: str
+    url: Optional[str]
 
-    def __init__(self, pool: "ConnectionPool", url: str, message: str) -> None:
+    def __init__(
+        self,
+        pool: Optional[Union["ConnectionPool", "HTTPConnectionPool"]],
```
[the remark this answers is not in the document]
hramezani: This one also reverted.

hramezani:
@sethmlarson @bluetech Thanks for your review. I've addressed some of your comments. 
There are some errors like:
```
"HTTPResponse" has no attribute "fp"
Item "HTTPResponse" of "Optional[HTTPResponse]" has no attribute "_safe_read"
```

I think we have to ignore them.

src/urllib3/response.py
```diff
@@ -221,35 +240,35 @@ def data(self) -> bytes:
...
         raise NotImplementedError()
 
     def read_chunked(
         self,
-        amt: typing.Optional[int] = None,
-        decode_content: typing.Optional[bool] = None,
-    ) -> bytes:
+        amt: Optional[int] = None,
+        decode_content: Optional[bool] = None,
+    ) -> Generator[bytes, None, None]:
```
sethmlarson, line 271: Can we do `Iterator[bytes]` here as well?

src/urllib3/response.py
```diff
@@ -317,26 +338,24 @@ def readable(self) -> bool:
 
     def readinto(self, b: bytearray) -> int:
         temp = self.read(len(b))
-        if len(temp) == 0:
+        if temp is None or len(temp) == 0:
```
sethmlarson, line 341: Since `read()` now returns only `bytes` can we remove the `is None` check here?

src/urllib3/response.py
```diff
@@ -852,7 +887,7 @@ def read_chunked(self, amt=None, decode_content=None):
                     yield decoded
 
             # Chunk content ends with \r\n: discard it.
-            while True:
+            while True and self._fp is not None:
```
sethmlarson, line 890: We can reduce this to `while self._fp is not None:`?

src/urllib3/util/retry.py
```diff
@@ -461,7 +463,9 @@ def increment(
             if redirect is not None:
                 redirect -= 1
             cause = "too many redirects"
-            redirect_location = response.get_redirect_location()
+            response_redirect_location = response.get_redirect_location()
```
sethmlarson, line 466: Can we remove this if we add a type annotation to `redirect_location` up above like so:

`redirect_location: Optional[str] = None`

sethmlarson (changes requested):
Few more comments, this is nearly there I think!

src/urllib3/util/retry.py
hramezani: No, because the return type of `get_redirect_location` is `Union[str, None, Literal[False]]`.
And if we change it to `redirect_location: Union[Optional[str], "Literal[False]"] = None` then we get 
`Argument 5 to "RequestHistory" has incompatible type "Union[str, None, Literal[False]]"; expected "Optional[str]"`
and we need to change `RequestHistory`

src/urllib3/response.py
hramezani: done!

src/urllib3/response.py
hramezani: done

src/urllib3/response.py
hramezani: done

sethmlarson:
This looks good to me now @hramezani , can you add the type: ignore comments in one commit so I can review them then we can get this merged :)

hramezani:
> This looks good to me now @hramezani , can you add the type: ignore comments in one commit so I can review them then we can get this merged :)

Added!

sethmlarson (approved):
LGTM, thank you!!

PR #2261, opened 2021-06-11
Add type hints to urllib3.contrib.ntlmpool

hramezani:

### DIFF
--- a/src/urllib3/contrib/ntlmpool.py
+++ b/src/urllib3/contrib/ntlmpool.py
@@ -6,7 +6,7 @@ Issue #10, see: http://code.google.com/p/urllib3/issues/detail?id=10
 
 from http.client import HTTPSConnection
 from logging import getLogger
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Union
 
 from ntlm import ntlm  # type:ignore
 
@@ -14,6 +14,7 @@ from urllib3.response import BaseHTTPResponse
 
 from .. import HTTPSConnectionPool
 from ..connection import HTTPBody
+from ..util.retry import Retry
 
 log = getLogger(__name__)
 
@@ -110,7 +111,7 @@ class NTLMConnectionPool(HTTPSConnectionPool):
         url: str,
         body: Optional[HTTPBody] = None,
         headers: Optional[Dict[str, str]] = None,
-        retries: int = 3,
+        retries: Union[Retry, bool, int] = 3,
         redirect: bool = True,
         assert_same_host: bool = True,
     ) -> BaseHTTPResponse:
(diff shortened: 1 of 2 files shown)


src/urllib3/contrib/ntlmpool.py
```diff
@@ -64,7 +68,7 @@ def _new_conn(self):
 
         # Remove the reference to the socket, so that it can not be closed by
         # the response object (we want to keep the socket open)
-        res.fp = None
+        res.fp = None  # type: ignore
```
hramezani, line 71: ignore because `error: "HTTPResponse" has no attribute "fp"`

src/urllib3/contrib/ntlmpool.py
```diff
@@ -96,20 +100,20 @@ def _new_conn(self):
                 raise Exception("Server rejected request: wrong username or password")
             raise Exception(f"Wrong server response: {res.status} {res.reason}")
 
-        res.fp = None
+        res.fp = None  # type: ignore
```
hramezani, line 103: ignore because `error: "HTTPResponse" has no attribute "fp"`

src/urllib3/contrib/ntlmpool.py
```diff
@@ -96,20 +100,20 @@ def _new_conn(self):
...
-        body=None,
-        headers=None,
-        retries=3,
-        redirect=True,
-        assert_same_host=True,
-    ):
+        method: str,
+        url: str,
+        body: Optional[HTTPBody] = None,
+        headers: Optional[Dict[str, str]] = None,
```
sethmlarson, line 112: Mappin[str, str]?

src/urllib3/contrib/ntlmpool.py
```diff
@@ -96,20 +100,20 @@ def _new_conn(self):
...
+        retries: int = 3,
```
sethmlarson, line 113: Should this be Union[Retry, bool, int] like elsewhere?

sethmlarson (changes requested):
Nice, few thoughts:

src/urllib3/contrib/ntlmpool.py
```diff
@@ -6,10 +6,15 @@
 
 from http.client import HTTPSConnection
 from logging import getLogger
+from typing import Any, Dict, Optional, Union
 
-from ntlm import ntlm
+from ntlm import ntlm  # type:ignore
+
+from urllib3.response import BaseHTTPResponse
```
sethmlarson, line 13: This should be a relative import

sethmlarson (commented):
Missed one thing then we're ready to merge

src/urllib3/contrib/ntlmpool.py
hramezani: Done!

sethmlarson (approved):
LGTM

PR #2262, opened 2021-06-11
Adding preload_content,retries,redirect and timout argumets to the top level request method.

ubdussamad:
I've tried to add the following arguments to the top-level request method.

1. `preload_content`
2. `redirect`
3. `retries`
4. `timeout`

I've also written some tests for it. One of the tests is simply a mockup.


src/urllib3/__init__.py
```diff
@@ -94,6 +94,10 @@ def request(
     body: Optional[HTTPBody] = None,
     fields: Optional[_TYPE_FIELDS] = None,
     headers: Optional[Mapping[str, str]] = None,
+    preload_content: Optional[bool] = True,
+    redirect: Optional[bool] = True,
+    retries: Optional[Union[Retry, bool, int]] = None,
```
ubdussamad, line 99: I've left the `retries` to be `None`, it's by default `None` in the file [`connectionpool.py`](https://github.com/urllib3/urllib3/blob/main/src/urllib3/connectionpool.py#L629) at line#629 and afterwards it's converted to a proper `retries` object.

test/with_dummyserver/test_poolmanager.py
```diff
@@ -378,6 +379,60 @@ def test_top_level_request_with_body(self):
...
+
+    def test_top_level_request_with_retries(self):
+        r = request("GET", f"{self.base_url}/redirect", retries=False)
+        assert r.status == 303
+
+        r = request("GET", f"{self.base_url}/redirect", retries=3)
+        assert r.status == 200
+
+    def test_top_level_request_with_timeout(self):
+        with mock.patch("urllib3.poolmanager.RequestMethods.request") as mockRequest:
```
ubdussamad, line 417: Is mocking with context manager okay here? Like I can do this with a mock decorator well.

shazow: IMO context manager is preferred. :D

src/urllib3/__init__.py
```diff
@@ -102,4 +106,14 @@ def request(
     The method does not accept low-level ``**urlopen_kw`` keyword arguments.
     """
```
shazow, line 107: I wonder if we're getting to the point where we should more explicitly (and repeatedly) document all of our parameters here?

It's a pain to have to click through N layers of "... and the rest of the args come from X" (top-level request -> PoolManager request -> ConnectionPool urlopen -> HTTPResponse from_httplib).

I feel like the top-level request helper will be a great entry-point for most people and learning about what the various parameters do off the bat would help beginners a lot.

(Also don't think this is blocking of this PR, but would be nice)

shazow (approved):
Looks good to me! Would be great to get another pair of :eyes: just in case. :)

test/with_dummyserver/test_poolmanager.py
ubdussamad: Okey-dok! :)

src/urllib3/__init__.py
ubdussamad: Yes, that'd be a great idea. Btw apart from the added args, I found a few more args like `decode_content` , `enforce_content_length` etc. But I am not sure if they have any importance or not.

shazow: Yea let's not be super exhaustive about adding all the extra args, we can be incremental about that. :)

sethmlarson (approved):
LGTM

PR #2265, opened 2021-06-12
Add --ignore-missing-imports flag to mypy run command.

hramezani:


src/urllib3/util/url.py
```diff
@@ -300,7 +300,7 @@ def _normalize_host(
 def _idna_encode(name: str) -> bytes:
     if name and any([ord(x) > 128 for x in name]):
         try:
-            import idna  # type: ignore
+            import idna
```
sethmlarson, line 303: Doesn't the latest idna include type hints?

sethmlarson (commented):
I'm unsure about this one, I think I prefer the explicit `type: ignore` on third party modules without type hints?

src/urllib3/util/url.py
hramezani: I think it does. but we need to install it in `idna` in the `lint` environment.

sethmlarson: Let's do that!

hramezani: Ok, I will create a new PR for this.

PR #2266, opened 2021-06-14
Add idna to mymy validation function in noxfile

hramezani:


sethmlarson (approved):
LGTM! We should probably explore using lockfiles per Nox session now that Dependabot supports the format :)

PR #2267, opened 2021-06-14
Strict param

### LINKED ISSUE #2263: Tolerate the strict parameter being passed via connection_from_context()
If we test Requests against urllib3 `main` branch we receive the following:

```
  File "/home/sethmlarson/urllib3/src/urllib3/poolmanager.py", line 354, in connection_from_url
    return self.connection_from_host(
  File "/home/sethmlarson/urllib3/src/urllib3/poolmanager.py", line 294, in connection_from_host
  File "/home/sethmlarson/urllib3/src/urllib3/poolmanager.py", line 310, in connection_from_context
  File "/home/sethmlarson/urllib3/src/urllib3/poolmanager.py", line 143, in _default_key_normalizer
    return key_class(**context)
TypeError: __new__() got an unexpected keyword argument 'key_strict'
```

This started in https://github.com/urllib3/urllib3/pull/2064. One of the goals of v2 is to be mostly API compatible so downstream projects can immediately start using v2. Should we add the following within `PoolManager.connection_from_context`?

```python
        if "strict" in request_context:
            import warnings

            warnings.warn(
                "The 'strict' parameter is no longer needed on Python 3+. "
                "This will raise an error in urllib3 v3.0.0.",
                DeprecationWarning
            )
            request_context.pop("strict")
```

to maintain API compatibility with Requests and other libraries that support Python 2?

Another small thought is to add a simple integration test with `requests.get(...)` to our test suite so we can track that we don't lose complete API compatibility.

hramezani:
Closes #2263

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -1,5 +1,6 @@
 import socket
 from test import resolvesLocalhostFQDN
+from unittest.mock import patch
 
 import pytest
 
@@ -266,6 +267,32 @@ class TestPoolManager:
         assert 1 == len(p.pools)
         assert pool is other_pool
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
+
+    @patch("urllib3.poolmanager.PoolManager.connection_from_pool_key")
+    def test_connection_from_context_strict_param(self, connection_from_pool_key):
+        p = PoolManager()
+        context = {
+            "scheme": "http",
+            "host": "example.com",
+            "port": 8080,
+            "strict": True,
+        }
+        with pytest.warns(DeprecationWarning) as record:
+            p.connection_from_context(context)
+
+        assert 1 == len(record)
+        msg = (
+            "The 'strict' parameter is no longer needed on Python 3+. "
+            "This will raise an error in urllib3 v3.0.0."
+        )
+        assert record[0].message.args[0] == msg
+
+        _, kwargs = connection_from_pool_key.call_args
+        assert kwargs["request_context"] == {
+            "scheme": "http",
+            "host": "example.com",
+            "port": 8080,
+        }
 
     def test_custom_pool_key(self):
         """Assert it is possible to define a custom key function."""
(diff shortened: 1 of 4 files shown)


test/test_integration.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+import requests
+
+
+class TestIntegration:
+    def test_requests_integration(self):
+        with pytest.warns(DeprecationWarning) as record:
```
sethmlarson, line 7: Can we move this under `with_dummyserver` and make an http and https request to avoid a real network call?

test/test_poolmanager.py
```diff
@@ -267,6 +268,32 @@ def test_http_connection_from_context_case_insensitive(self):
         assert pool is other_pool
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
 
+    @patch("urllib3.poolmanager.PoolManager.connection_from_pool_key")
+    def test_connection_from_context_strict_param(self, connection_from_pool_key):
```
sethmlarson, line 272: This is perfect!

sethmlarson (commented):
Woo! Requests compatibility on every PR 👏 One tiny change then we'll get this merged.

test/test_integration.py
hramezani: I've created a new file `test_integration.py` under `with_dummyserver` and added tests for http and https

test/test_poolmanager.py
```diff
@@ -267,6 +268,32 @@ def test_http_connection_from_context_case_insensitive(self):
...
+        p = PoolManager()
+        context = {
+            "scheme": "http",
+            "host": "example.com",
+            "port": "8080",
```
sethmlarson, line 277: Tiny change: port should be an int here

test/with_dummyserver/test_integration.py
```diff
@@ -0,0 +1,35 @@
+import pytest
```
sethmlarson, line 1: 🎉

sethmlarson (commented):
One more thing I should have noticed last time, after that let's merge this.

test/test_poolmanager.py
hramezani: Done!

sethmlarson (approved):
LGTM!

PR #2271, opened 2021-06-16
Add type hints to urllib3.contrib.socks

hramezani:

### DIFF
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -71,18 +71,18 @@ except ImportError:
 
 try:
     from typing import TypedDict
+
+    class _TYPE_SOCKS_OPTIONS(TypedDict):
+        socks_version: int
+        proxy_host: Optional[str]
+        proxy_port: Optional[str]
+        username: Optional[str]
+        password: Optional[str]
+        rdns: bool
+
+
 except ImportError:  # Python 3.7
-    _TYPE_SOCKS_OPTIONS = Dict[str, Any]
-else:
-    _TYPE_SOCKS_OPTIONS = TypedDict(  # type: ignore
-        "SocksOptions",
-        socks_version=int,
-        proxy_host=Optional[str],
-        proxy_port=Optional[str],
-        username=Optional[str],
-        password=Optional[str],
-        rdns=bool,
-    )
+    _TYPE_SOCKS_OPTIONS = Dict[str, Any]  # type: ignore
 
 
 class SOCKSConnection(HTTPConnection):
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -71,17 +71,18 @@ except ImportError:
 
 try:
     from typing import TypedDict
-except ImportError:
-    from typing_extensions import TypedDict  # Python 3.7
-
-
-class SocksOptions(TypedDict):
-    socks_version: int
-    proxy_host: Optional[str]
-    proxy_port: Optional[str]
-    username: Optional[str]
-    password: Optional[str]
-    rdns: bool
+except ImportError:  # Python 3.7
+    _TYPE_SOCKS_OPTIONS = Dict[str, Any]
+else:
+    _TYPE_SOCKS_OPTIONS = TypedDict(  # type: ignore
+        "SocksOptions",
+        socks_version=int,
+        proxy_host=Optional[str],
+        proxy_port=Optional[str],
+        username=Optional[str],
+        password=Optional[str],
+        rdns=bool,
+    )
 
 
 class SOCKSConnection(HTTPConnection):
@@ -89,7 +90,9 @@ class SOCKSConnection(HTTPConnection):
     A plain-text HTTP connection that connects via a SOCKS proxy.
     """
 
-    def __init__(self, _socks_options: SocksOptions, *args: Any, **kwargs: Any) -> None:
+    def __init__(
+        self, _socks_options: _TYPE_SOCKS_OPTIONS, *args: Any, **kwargs: Any
+    ) -> None:
         self._socks_options = _socks_options
         super().__init__(*args, **kwargs)
 
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -69,14 +69,28 @@ try:
 except ImportError:
     ssl = None  # type: ignore
 
+try:
+    from typing import TypedDict  # >=3.8
+except ImportError:
+    from typing_extensions import TypedDict  # <=3.7
+
+
+class SocksOptions(TypedDict):
+    socks_version: int
+    proxy_host: Optional[str]
+    proxy_port: Optional[str]
+    username: Optional[str]
+    password: Optional[str]
+    rdns: bool
+
 
 class SOCKSConnection(HTTPConnection):
     """
     A plain-text HTTP connection that connects via a SOCKS proxy.
     """
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        self._socks_options = kwargs.pop("_socks_options")
+    def __init__(self, _socks_options: SocksOptions, *args: Any, **kwargs: Any) -> None:
+        self._socks_options = _socks_options
         super().__init__(*args, **kwargs)
 
     def _new_conn(self) -> "socks.socksocket":
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -39,7 +39,7 @@ with the proxy:
 """
 
 try:
-    import socks
+    import socks  # type: ignore
 except ImportError:
     import warnings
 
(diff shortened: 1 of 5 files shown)


src/urllib3/contrib/socks.py
```diff
@@ -66,23 +67,23 @@
...
+    ssl = None  # type: ignore
 
 
 class SOCKSConnection(HTTPConnection):
     """
     A plain-text HTTP connection that connects via a SOCKS proxy.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
sethmlarson, line 78: Let's add an explicit `_sock_options` parameter to the signature. We can also use `typing.TypedDict` to make `_sock_options` really explicit.

src/urllib3/exceptions.py
```diff
@@ -29,9 +30,11 @@ class HTTPWarning(Warning):
 class PoolError(HTTPError):
     """Base exception for errors caused within a pool."""
 
-    pool: "ConnectionPool"
+    pool: Union["ConnectionPool", "SOCKSConnection"]
 
-    def __init__(self, pool: "ConnectionPool", message: str) -> None:
+    def __init__(
+        self, pool: Union["ConnectionPool", "SOCKSConnection"], message: str
```
sethmlarson, line 36: Hmmm, this is unfortunate.

Is there a better error we could be raising from `_new_conn()`? What do regular `HTTPConnection` objects raise in the case of non-timeout failure?

Might be a good time for a breaking change.

sethmlarson (changes requested):
Nice! Shame there's no types in PySocks, another reason it'd be cool if we could help with maintenance. Here's since comments for you:

src/urllib3/exceptions.py
pquentin: Regular `HTTPConnection` objects raise `NewConnectionError` too.

Why are we changing `PoolError` instead of just `NewConnectionError`? In both `HTTPConnection._new_conn` and `SOCKSConnection._new_conn` the `pool` parameter is actually `self`, so not a pool object.

src/urllib3/contrib/socks.py
```diff
@@ -66,23 +67,37 @@
 try:
     import ssl
 except ImportError:
-    ssl = None
+    ssl = None  # type: ignore
+
+try:
+    from typing import TypedDict  # >=3.8
+except ImportError:
+    from typing_extensions import TypedDict  # <=3.7
```
pquentin, line 75: Can we explicit write "Python 3.7"? It's much easier to grep when we'll be removing support for 3.7

src/urllib3/exceptions.py
```diff
@@ -160,10 +161,14 @@ class ConnectTimeoutError(TimeoutError):
     pass
 
 
-class NewConnectionError(ConnectTimeoutError, PoolError):
+class NewConnectionError(ConnectTimeoutError, HTTPError):
     """Raised when we fail to establish a new connection. Usually ECONNREFUSED."""
 
-    pass
+    pool: "SOCKSConnection"
```
pquentin, line 167: I think you'll have to use the parent class `HTTPConnection` when typing connectionpool.py, so let's use that right away?

It's also a bit weird to use `pool` instead of `conn` but I guess changing that would be a breaking change

pquentin (changes requested):
Thanks! Two small comments

src/urllib3/exceptions.py
sethmlarson: Could change to `conn` and add a deprecated property `pool`?

pquentin: Since the exception is no longer a PoolError I would go ahead and switch to conn. Let's just document that as part of the 2.0 breaking changes.

This is the kind of subtle issues that get fixed when you add types :)

src/urllib3/exceptions.py
```diff
@@ -160,10 +161,14 @@ class ConnectTimeoutError(TimeoutError):
     pass
 
 
-class NewConnectionError(ConnectTimeoutError, PoolError):
+class NewConnectionError(ConnectTimeoutError, HTTPError):
     """Raised when we fail to establish a new connection. Usually ECONNREFUSED."""
 
-    pass
+    pool: "HTTPConnection"
```
pquentin, line 167: Can you please update `pool` to `conn` here and below?

pquentin (dismissed):

src/urllib3/contrib/socks.py
```diff
@@ -66,23 +67,37 @@
...
-    ssl = None
+    ssl = None  # type: ignore
+
+try:
+    from typing import TypedDict
+except ImportError:
+    from typing_extensions import TypedDict  # Python 3.7
+
+
+class SocksOptions(TypedDict):
```
sethmlarson, line 78: We may have to define this structure under the `try: from typing import TypedDict` section. Since it's not a public type we can name it `_TYPE_SOCKS_OPTIONS`.

src/urllib3/contrib/socks.py
```diff
@@ -66,23 +67,37 @@
 try:
     import ssl
 except ImportError:
-    ssl = None
+    ssl = None  # type: ignore
+
+try:
+    from typing import TypedDict
+except ImportError:
+    from typing_extensions import TypedDict  # Python 3.7
```
sethmlarson, line 75: We don't depend on `typing_extensions` so can't use it here. In this case we can set `_TYPE_SOCKS_OPTIONS` to `Dict[str, Any]`? Is mypy on Python 3.6 okay with that?

src/urllib3/exceptions.py
hramezani: I will do this and I am going to add a deprecated property as @sethmlarson suggested here https://github.com/urllib3/urllib3/pull/2271#discussion_r653544455.

Shouldn't we go for deprecation?

sethmlarson (changes requested):
Some thoughts on the `_socks_options` types

src/urllib3/contrib/socks.py
sethmlarson: This PR actually highlights that we maybe shouldn't have typing-extensions in dev-requirements.txt if it's not being used because this failure wouldn't be caught in CI until release integration tests.

src/urllib3/exceptions.py
pquentin: As I mentioned above I don't think a deprecation makes sense unless we keep the `PoolError` inheritance. It's possible, but not worth it in my opinion. I'll let Seth decide

sethmlarson: My thought was to have a deprecation only on the accessor property `pool` and to remove pool completely from the constructor as it's a much bigger burden.

The likelyhood of someone accessing the actual object attached to the error is higher than someone using a keyword argument to construct an instance of this error (likely we're mostly the only ones creating these errors).

Maybe we can have a more aggressive deprecation period for these types of changes too, maybe say "in a later 2.x release" instead of "3.0.0"?

src/urllib3/contrib/socks.py
hramezani: I can remove it in a seperate PR

hramezani: Here is the PR https://github.com/urllib3/urllib3/pull/2275

test/test_exceptions.py
```diff
@@ -45,3 +46,16 @@ def test_header_parsing_errors(self):
 
         assert "defects" in str(hpe)
         assert "unparsed_data" in str(hpe)
+
+
+class TestNewConnectionError:
+    def test_pool_property_deprication_warning(self):
```
sethmlarson, line 52: Typo "deprecation"

test/test_exceptions.py
```diff
@@ -45,3 +46,16 @@ def test_header_parsing_errors(self):
...
+        err = NewConnectionError(HTTPConnectionPool("localhost"), "test")
```
sethmlarson, line 53: I think you meant to pass an HTTPConnection here?

test/test_exceptions.py
```diff
@@ -45,3 +46,16 @@ def test_header_parsing_errors(self):
...
+        with pytest.warns(DeprecationWarning) as record:
+            err.pool
```
sethmlarson, line 55: So we're doing something with this property can we `assert err.pool is err.conn`?

sethmlarson (changes requested):
Nearly there! Handful of small tweaks then we can merge this.

test/test_exceptions.py
pquentin: We should probably run mypy on tests at some point!

src/urllib3/contrib/socks.py
```diff
@@ -66,23 +67,40 @@
...
+    class _TYPE_SOCKS_OPTIONS(TypedDict):
+        socks_version: int
+        proxy_host: Optional[str]
+        proxy_port: Optional[str]
+        username: Optional[str]
+        password: Optional[str]
+        rdns: bool
+
+
+except ImportError:  # Python 3.7
```
sethmlarson, line 84: Should be "Python 3.6" here

sethmlarson (commented):
(disregard this)

sethmlarson (approved):
LGTM, thank you!!! 🎉

pquentin (approved):

PR #2272, opened 2021-06-17
Fix remove_headers_on_redirect feature modifying passed-in headers in-place

bluetech:
In general I'd say it's expected in Python that when passing a dict to
to an argument like `headers` that it won't be modified in-place and
could be reused safely for further requests.

Previously the remove_headers_on_redirect feature didn't uphold this. Do
a copy to make sure that it does.

This commit additionally *avoids* an unneeded copy when the
`remove_headers_on_redirect` code isn't reached.

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -415,7 +415,7 @@ class PoolManager(RequestMethods):
         kw["redirect"] = False
 
         if "headers" not in kw:
-            kw["headers"] = self.headers.copy()  # type: ignore
+            kw["headers"] = self.headers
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
This change looks mostly good, I'm sad we don't have type coverage on `urlopen` but I know we've had troubles with .copy() not being an attribute of `Mapping[str, str]` but headers are almost certainly a dict at this point.

I think we're okay? Can another reviewer (or future Seth with a desktop) take a peek too?

pquentin (approved):
Thanks!

PR #2273, opened 2021-06-17
Remove confusing unused argument from dummyserver SocketServerThread

bluetech:
Actually a dynamic port is used and the argument is ignored.


sethmlarson (approved):
LGTM thank you!

Issue #2254, opened 2021-06-06
Allow specifying pool headers which are not overridden by request-specific headers


bluetech:
This is a feature request + proposal for a small thing that tripped me.

### Context

> What are you trying to do?

I have a service which creates a service-scoped `PoolManager` and uses it for all external requests. I set some headers when I create it, for example

```py
pm = urllib3.PoolManager(headers={
    'User-Agent': 'my-service',
    'Connection': 'keep-alive',
    'Accept': '*/*,
}
```

Generally, I want these headers to be used in all requests. (So far so good).

Now I have a specific request which requires an extra header or overrides some default header.

> How do you expect to do it?

```py
pm.request('POST', 'https://foo.bar', headers={
    'Content-Type': 'application/json',
}, ...)
```

This does the wrong thing than what I want - it overrides the headers instead of adding to them.

> Is it something you currently cannot do?

It is possible to do as follows:

```py
pm.request('POST', 'https://foo.bar', headers={
    **pm.headers,
    'Content-Type': 'application/json',
}, ...)
```

The problem is that it is easy to forget to do.

> Is this related to an existing issue/problem?

Not that I can find.

### Proposal

Add `extra_headers` argument to `PoolManager` which is not entirely overridden by `request.headers`, only per-header overridden (with preference given to the request). `PoolManager.headers` remains as "default headers" (the existing behavior).

### Alternatives

> Can you achieve the same result doing it in an alternative way?

1. Keep using the manual merging. Easy to forget.

2. Change `pm.urlopen(headers=...)` itself to merge instead of override. This is what `requests` does. This is a big breaking change so probably no go. Also doesn't allow to override when actually wanted.

3. Add `pm.urlopen(extra_headers={...})` which merges. Seems still too easy to forget. I think putting it on the PoolManager is better.

4. Add some "merging strategy" (merge/override) parameter. Seems too complicated.

### Contribution

> Would you be willing to submit a PR?

Yes.

sethmlarson:
The `**pool.headers` construction interests me, maybe we can document that? Does that work with `HTTPHeaderDict`?

sigmavirus24:
So you want to merge request headers into the pre-existing headers. That's how Requests works today and how other clients work too so the ergonomics make sense

wimglenn:
A possible proposal is for

    pm.request('POST', 'https://foo.bar', headers=pm.headers | more_headers, ...)

More details in the linked PR.

wimglenn:
[...]

Issue #2242, opened 2021-05-22
Support repeated request headers using HTTPHeaderDict


jkupferer:
### Context

What are you trying to do?

I am trying to pass repeated headers such as the `Impersonate-Group` header used by the kubernetes API. (https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation)

How do you expect to do it?

I would like to be able to call the urllib3 request method and pass the parameter `headers` as an HTTPHeaderDict.

Is it something you currently cannot do?

It is nearly supported. It is broken for requests with request body because of the implementation of `request_encode_body`:

https://github.com/urllib3/urllib3/blob/main/src/urllib3/request.py#L174
```
        extra_kw: Dict[str, Any] = {"headers": {}}
```

https://github.com/urllib3/urllib3/blob/main/src/urllib3/request.py#L196
```
        extra_kw["headers"].update(headers)
```

These lines force headers to be a plain python dictionary even when an HTTPHeaderDict is passed.

Is this related to an existing issue/problem?

Not that I am aware of.

### Alternatives

Can you achieve the same result doing it in an alternative way?

Not that I know of

Is the alternative considerable?

### Duplicate

Has the feature been requested before?
If so, please provide a link to the issue.

Not that I know of.

### Contribution

Would you be willing to submit a PR?
_(Help can be provided if you need assistance submitting a PR)_

Yes

shazow:
+1, I believe this used to be a feature we supported at one point! Specifically, you could pass in a list of tuples which `dict.update(...)` accepted.

If the only thing blocking this is the new type annotations, makes sense to fix that. Let's wait for Seth/Quentin to give guidance on the best way to move forward though. :)

sethmlarson:
So `HTTPHeaderDict` for requests is an interesting one, because it makes total sense to use but it really wasn't meant to be a public API in v1.x. Users are still able to import it from `urllib3.connection` and `urllib3.response` and use it without knowing it isn't a public API though. But now when you try to pass that `HTTPHeaderDict` as request headers in v1.x it... almost works? And that's an unfortunate thing for us cuz now our users are using something that isn't tested anywhere!

So starting in v2.0 we're making `HTTPHeaderDict` a first-class citizen and actually the recommended way of managing headers for both requests and responses.

Basically all of this is to say we don't support duplicate headers with any released version, sorry! My recommendation is to try using commas to combine the header values. You can also use an `HTTPHeaderDict` to do this for you like so:

```python
[...]

PR #2274, opened 2021-06-17
Add a test for using HTTPHeaderDict with PoolManager(headers)

bluetech:
Add a simple test for `HTTPHeaderDict` in conjunction with `PoolManager`, and the `**headers` syntax (see https://github.com/urllib3/urllib3/issues/2254#issuecomment-860777164).

### DIFF
--- a/test/with_dummyserver/test_poolmanager.py
+++ b/test/with_dummyserver/test_poolmanager.py
@@ -6,7 +6,7 @@ import pytest
 
 from dummyserver.server import HAS_IPV6
 from dummyserver.testcase import HTTPDummyServerTestCase, IPv6HTTPDummyServerTestCase
-from urllib3 import HTTPResponse, request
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks!

pquentin (approved):
The coverage issue is not caused by the change as it only adds a test.

Are you interested in doing the breaking change suggested in #2254?

bluetech:
Thanks for reviewing @pquentin.

> Are you interested in doing the breaking change suggested in #2254?

I don't think a breaking change there would be acceptable -- it would be silent and have a significant effect. I started trying to add the `extra_headers` solution but it would require some cleanups before it could be done properly.

sethmlarson:
@bluetech [This issue](https://github.com/urllib3/urllib3/issues/2242) is related to this PR as well, basically it's `HTTPHeaderDict` all the way down so we can support duplicate headers without making it a comma.

PR #2276, opened 2021-06-18
 Remove python2 compat stuff from test_socketlevel.py 

bluetech:
The `TestBrokenHeaders` test has a skip on condition involving `mimetools` which no longer exists by this name in Python 3, so I assume it's no longer relevant.

The change makes isort reformat all of the imports so I split the formatting to a separate commit. I recommend looking at the second commit for the actual change here.


pquentin (approved):
Thanks! Nice catch.

The coverage on the actual codecov UI is indeed 100, it's just not correctly reported here.

PR #2277, opened 2021-06-18
Sync create_connection with socket module

bluetech:
Bring in cycle-breaking which reduces garbage for the GC.

For reference: https://github.com/python/cpython/blob/83c9dad8da5fc90b717eef683304aaa49448615a/Lib/socket.py#L807


sethmlarson (approved):
Great work, this is definitely an improvement. Thank you!

PR #2278, opened 2021-06-19
Couple of ntlmpool changes

bluetech:
First commit: Make ntlmpool use urllib3's HTTPSConnection instead of httplib's.  The main urllib3 doesn't use http.client.HTTPSConnection, but rolls its own instead. ntlmpool should follow suit.

Second commit:  Respect connection timeout in ntlmpool.

It seems to me like ntlmpool is missing a lot of additional stuff in its `_new_conn` compared to `HTTPSConnectionPool` but I only added timeout.

I should note I don't use ntlmpool at all, I only saw this when looking at how urllib3 uses httplib.

### DIFF
--- a/src/urllib3/contrib/ntlmpool.py
+++ b/src/urllib3/contrib/ntlmpool.py
@@ -3,13 +3,12 @@ NTLM authenticating pool, contributed by erikcederstran
 
 Issue #10, see: http://code.google.com/p/urllib3/issues/detail?id=10
 """
-
-from http.client import HTTPSConnection
 from logging import getLogger
 
 from ntlm import ntlm
 
 from .. import HTTPSConnectionPool
+from ..connection import HTTPSConnection
 
 log = getLogger(__name__)
 
--- a/src/urllib3/contrib/ntlmpool.py
+++ b/src/urllib3/contrib/ntlmpool.py
@@ -49,7 +49,11 @@ class NTLMConnectionPool(HTTPSConnectionPool):
         req_header = "Authorization"
         resp_header = "www-authenticate"
 
-        conn = HTTPSConnection(host=self.host, port=self.port)
+        conn = HTTPSConnection(
+            host=self.host,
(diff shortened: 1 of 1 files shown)


pquentin (approved):
This module would need an actual user, we don't know if it works or not, so it probably doesn't.

However the changes make sense.

pquentin:
Closing/reopening to see if the unrelated failure for a newly added test is random or not.

sethmlarson:
This module isn't tested anywhere in our test suite and as far as I know doesn't have any users. We've received [a total of two issues](https://github.com/urllib3/urllib3/issues?q=is%3Aissue+NTLMConnectionPool) in the entire project lifetime that have mentioned `NTLMConnectionPool`.

Without tests or users it's hard to tell if making any changes will break anything or if anyone will actually benefit.

I propose instead we should invest time in finding a maintainer/stakeholder for this module or we remove it in v2.0. What are everyone's thoughts here?

bluetech:
As mentioned, I don't use ntlm or windows in any way, but given that

- The module is not covered
- It is not clear even which pypi package is needed for it
- It is under `contrib`
- It has fallen into disrepair (e.g. timeout/ssl/other options not being respected)
- According to [Wikipedia](https://en.wikipedia.org/wiki/NT_LAN_Manager#Availability_and_use_of_NTLM), "Since 2010, Microsoft no longer recommends NTLM in applications"
- Seems like it's little used, if at all

I'd say we should remove it.

If the 1.26.x branch is still active, perhaps we can deprecate it there first and see if anyone complains? (And if they do, help them move it to a third-party package...)

sethmlarson:
I think we're in agreement that this PR isn't necessary. I've [opened a ticket to deprecate `NTLMConnectionPool`](https://github.com/urllib3/urllib3/issues/2282) in 1.26.x and if we don't hear anything we can remove in `main`.

Jjajsja (approved):

Issue #2283, opened 2021-06-20
Make top-level request() function parameters keyword-only


sethmlarson:
- Make all parameters after `method` and `url` for `urllib3.request()` keyword only by adding `*,` (see [PEP-3102](https://www.python.org/dev/peps/pep-3102/))
- Add a test case for the error that's raised when a parameter isn't used as a keyword

aitoehigie:
I think I can handle this.

fvlima:
I was doing this, but I had doubts about the best place to put the test case

aitoehigie:
> I was doing this, but I had doubts about the best place to put the test case

There is a test directory. I believe you will need to create a test_request.py file in it. Going by the convention in that directory

fvlima:
> There is a test directory. I believe you will need to create a test_request.py file in it. Going by the convention in that directory

yep, but I thought that would be strange to have just one test case in this `test_request.py`. It was exactly what I was doing...

fvlima:
I'm pushing the PR and we can discuss there

fvlima:
https://github.com/urllib3/urllib3/pull/2284

PR #2284, opened 2021-06-20
Add keywork arg to request #2283

fvlima:
Make all parameters after method and url for urllib3.request() keyword only by adding *

Closes https://github.com/urllib3/urllib3/issues/2283

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Changes
   deprecated names for ``format_multipart_header_param``. #2257
 * The ``RequestField`` ``header_formatter`` parameter is deprecated in
   favor of overriding the ``_render_part`` method. #2257
+* Make all parameters after `method` and `url` for `urllib3.request()` 
+  keyword only. #2283
 
 
 1.26.5 (2021-05-26)
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -91,6 +91,7 @@ _DEFAULT_POOL = PoolManager()
 def request(
     method: str,
     url: str,
+    *,
     body: Optional[HTTPBody] = None,
     fields: Optional[_TYPE_FIELDS] = None,
     headers: Optional[Mapping[str, str]] = None,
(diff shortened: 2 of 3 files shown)


src/urllib3/request.py
```diff
@@ -65,6 +65,7 @@ def request(
         self,
         method: str,
         url: str,
+        *,
```
sethmlarson, line 68: We can't change this one unfortunately, backwards compatibility needs to be maintained for existing APIs.

test/test_request.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
sethmlarson, line 1: The test cases look good but I think there's an existing place for tests of `urllib3.request()`

test/test_request.py
```diff
@@ -0,0 +1,24 @@
...
+
+from urllib3 import request
+
+from dummyserver.testcase import HTTPDummyServerTestCase
+
+
+class TestRequest(HTTPDummyServerTestCase):
+    @classmethod
```
sethmlarson, line 9: I don't think we need the `setup_method`

sethmlarson (changes requested):
Thanks for this! A few comments for you then we can get this merged.

test/test_request.py
fvlima: Great, can you show me? I have doubts to the best place to put this test

src/urllib3/request.py
fvlima: Ok, I forgot about the backward compatibility... removing it

test/test_request.py
sethmlarson: In the file `test/with_dummyserver/test_poolmanager.py` search for `test_top_level_request`. Under that test case should work!

sethmlarson: We only need the extra `TypeError` case btw, the positive case is already there.

test/test_request.py
fvlima: Done

test/test_request.py
fvlima: Done

src/urllib3/request.py
fvlima: Done

sethmlarson (approved):
LGTM, thanks!

PR #2289, opened 2021-06-21
Add DeprecationWarning for urllib3.contrib.nltmpool module

### LINKED ISSUE #2282: Deprecate NTLMConnectionPool in 1.26.x
As was mentioned in https://github.com/urllib3/urllib3/pull/2278#issuecomment-864414599 and https://github.com/urllib3/urllib3/pull/2278#issuecomment-864450016 we're moving to remove `NTLMConnectionPool` and the `urllib3.contrib.nltmpool` module from urllib3 in v2.0 if we don't find a new maintainer for the module (perhaps as a third-party package ie `urllib3-ntlmpool`?)

- The module is not covered by our test suite.
- It is not clear even which pypi package is needed for it.
- It has fallen into disrepair (e.g. timeout/ssl/other options not being respected).
- According to Wikipedia, "Since 2010, Microsoft no longer recommends NTLM in applications"
- Seems like it's not used often, if at all.

In the `1.26.x` branch we should unconditionally raise a `DeprecationWarning` when the module is imported. Should link to this issue with a call to action to comment in the issue if they are a user. This should help us better discover who (if any) our users are here so we can better make a decision.

imkaka:
Closes #2282 

Please let me know for further improvements in the PR.
1. The words used in the warning message can be definitely improved.
2. Should we add tests? because I couldn't find the test file for this module and also as we will be removing the whole module in v2.0 tests are required?
3. Please suggest any other changes I might need to do on this.

### DIFF
--- a/src/urllib3/contrib/ntlmpool.py
+++ b/src/urllib3/contrib/ntlmpool.py
@@ -13,15 +13,12 @@ from ntlm import ntlm
 from .. import HTTPSConnectionPool
 from ..packages.six.moves.http_client import HTTPSConnection
 
-
 warnings.warn(
     "The 'urllib3.contrib.nltmpool' module is deprecated and will be removed "
-    "in a later urllib3 v2.0 release for a possible third-party library as "
-    "urllib3 is not able to support it properly due to reasons listed in the "
-    "issue: https://github.com/urllib3/urllib3/issues/2282, If you are a user "
-    "of this module please have your say in the same issue above.",
+    "in urllib3 v2.0 release, urllib3 is not able to support it properly due "
+    "to reasons listed in issue: https://github.com/urllib3/urllib3/issues/2282. "
+    "If you are a user of this module please comment in the mentioned issue.",
     DeprecationWarning,
-    stacklevel=2,
 )
 
 log = getLogger(__name__)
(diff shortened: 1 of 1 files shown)


src/urllib3/contrib/ntlmpool.py
```diff
@@ -20,6 +32,15 @@ class NTLMConnectionPool(HTTPSConnectionPool):
     Implements an NTLM authentication version of an urllib3 connection pool
     """
 
+    warnings.warn(
```
sethmlarson, line 35: You can remove this warning, everything is covered by the one above.

src/urllib3/contrib/ntlmpool.py
```diff
@@ -5,13 +5,25 @@
...
 from logging import getLogger
 
 from ntlm import ntlm
 
 from .. import HTTPSConnectionPool
 from ..packages.six.moves.http_client import HTTPSConnection
 
+
+warnings.warn(
+    "The 'urllib3.contrib.nltmpool' module is deprecated and will be removed "
```
sethmlarson, line 18: - "...will be removed in urllib3 v2.0..."
- Remove the mention of the third-party library. We haven't made a decision here yet.
- "If you are a user of this module please comment in the mentioned issue."

src/urllib3/contrib/ntlmpool.py
```diff
@@ -5,13 +5,25 @@
...
+    "in a later urllib3 v2.0 release for a possible third-party library as "
+    "urllib3 is not able to support it properly due to reasons listed in the "
+    "issue: https://github.com/urllib3/urllib3/issues/2282, If you are a user "
+    "of this module please have your say in the same issue above.",
+    DeprecationWarning,
+    stacklevel=2,
```
sethmlarson, line 24: Remove the `stacklevel=2`. We're already in the global scope so no need to make the warning appear to have come from a different level in the call stack.

sethmlarson (changes requested):
Thanks for this! I left some comments for you:

sethmlarson:
I don't think tests are necessary here.  Also can you install `nox` and run `nox -rs format` before submitting your updates? That'll make sure your code is properly formatted to pass our linter.

src/urllib3/contrib/ntlmpool.py
imkaka: Okey

bluetech (dismissed):

imkaka:
Thank you @sethmlarson for the PR review, I have tried to address all the comments, please have a look.

sethmlarson (approved):
LGTM, thank you for this contribution! :tada:

PR #2293, opened 2021-06-21
Add type hints to urllib3.contrib._securetransport.low_level

hramezani:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -9,6 +9,7 @@ import nox
 # ignored.
 TYPED_FILES = {
     "src/urllib3/contrib/__init__.py",
+    "src/urllib3/contrib/_securetransport/bindings.py",
     "src/urllib3/contrib/_securetransport/low_level.py",
     "src/urllib3/contrib/ntlmpool.py",
     "src/urllib3/contrib/socks.py",
--- a/noxfile.py
+++ b/noxfile.py
@@ -9,6 +9,7 @@ import nox
 # ignored.
 TYPED_FILES = {
     "src/urllib3/contrib/__init__.py",
+    "src/urllib3/contrib/_securetransport/low_level.py",
     "src/urllib3/contrib/ntlmpool.py",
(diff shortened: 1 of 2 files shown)


src/urllib3/contrib/_securetransport/low_level.py
```diff
@@ -34,7 +46,7 @@ def _cf_data_from_bytes(bytestring):
     )
 
 
-def _cf_dictionary_from_tuples(tuples):
+def _cf_dictionary_from_tuples(tuples: List[Tuple[Any, ...]]) -> CFDictionary:
```
sethmlarson, line 49: Type should be `List[Tuple[Any, Any]]`

src/urllib3/contrib/_securetransport/low_level.py
```diff
@@ -123,10 +135,12 @@ def _cf_string_to_unicode(value):
         string = buffer.value
     if string is not None:
         string = string.decode("utf-8")
-    return string
+    return string  # type: ignore
```
sethmlarson, line 138: Can `string` somehow be `None` here?

noxfile.py
```diff
@@ -9,6 +9,7 @@
 # ignored.
 TYPED_FILES = {
     "src/urllib3/contrib/__init__.py",
+    "src/urllib3/contrib/_securetransport/low_level.py",
```
sethmlarson, line 12: Forgot to mention it in the previous PR, can we add the `bindings.py` to this list? Just so we run mypy on it even if it's basically skipped.

sethmlarson (changes requested):
Looks pretty good, few comments then we can merge.

hramezani:
@sethmlarson Thanks for the review.
I addressed your comments related to type hints(force pushed to the first commit)
And added another commit to add `bindings.py` to `TYPED_FILES` list

sethmlarson (approved):
LGTM :rocket:

PR #2296, opened 2021-06-22
Include py.typed file in package to comply with PEP 561 

johnthagen:
See https://www.python.org/dev/peps/pep-0561/

`mypy` and other type checkers use the presence of a `py.typed` file in the root of the package to signal that the package has inline type annotations to be used for type checking. Including this file allows users of this library to properly type check their usage of this package.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,4 @@
-include README.rst CHANGES.rst LICENSE.txt dev-requirements.txt Makefile
+include README.rst CHANGES.rst LICENSE.txt dev-requirements.txt Makefile src/urllib3/py.typed
 recursive-include dummyserver *
 recursive-include test *
 recursive-include docs *
--- a/setup.py
+++ b/setup.py
@@ -104,6 +104,9 @@ setup(
         "urllib3.contrib._securetransport",
         "urllib3.util",
     ],
+    package_data={
+        "urllib3": ["py.typed"],
+    },
     package_dir={"": "src"},
     requires=[],
     python_requires=">=3.6, <4",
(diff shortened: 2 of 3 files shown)


sethmlarson (changes requested):
Thanks for this! Can we create an integration test that ensures that `mypy` can recognize our types? Basically I'm worried that somehow `py.typed` wouldn't be included in the dists.

Could be something simple like:

- A file that has `import urllib3; urllib3.request("GET", "https://example.com")`
- A file (`ci/downstream/mypy.sh`) that installs `mypy`, `python -m pip install .`, and then runs `mypy` on the Python file above.

What do you think?

johnthagen:
If we do this, we'll want to test both the `.tar.gz` sdist and the `.whl` wheel, since both use slightly different mechanisms to include package data.

I've never seen another project add this kind of test before, but I can see that it could have some utility. The test would probably need to add a _failing_ type hint and check that `mypy` failed, since I believe if `py.typed` wasn't there, `mypy` would skip over `urllib3` entirely and things would "pass". 

So something like

```py
import urllib3; urllib3.request(0.0, False)
```

sethmlarson:
@johnthagen If you use `mypy --strict` Mypy will complain about not having types for an import iirc?

johnthagen:
@sethmlarson Yes, though you can selectively disable any of the `strict` settings.

What I normally do for my projects is to configure `mypy` like this:

```toml
[tool.mypy]
ignore_missing_imports = true
strict = true
```

An example can be viewed here: https://github.com/johnthagen/python-blueprint/blob/master/pyproject.toml#L5

johnthagen:
Yes, `mypy --strict` could be a way to validate that `urllib3` has the correct types exposed.

sethmlarson (approved):
LGTM, thank you!

Issue #2235, opened 2021-05-21
Map all keyword arguments to the top level `request` wrapper.


ubdussamad:
Right now, only a small number of keyword arguments are supported in the top-level wrapper `urllib3.request` function.
It wraps the `poolManager.request()` function by creating a default `poolManager` instance.
It was added to make the interface even simpler for users.

Currently supported kwArgs are: `body=None, fields=None, headers=None`.

The ones which are **not** supported: `preload_content, timeout, retries, redirect` and maybe some more which I might've missed.

This reduces the quality of the wrapper function as the user might not know what else they could do with this method and also makes adding type hints harder which is something we don't want in v2.

The wrapper method resides in the file [__init.py__.](https://github.com/urllib3/urllib3/blob/main/src/urllib3/__init__.py)

The original `request` method from the `poolManager` instance catches these kwArgs through **kwArgs.

If you need to look at an example of adding one, @venthur recently added the body parameter to the request function. (#2226)

ubdussamad:
@sethmlarson Should it be labeled contributor friendly? It'll attract more people.

csurfer:
@ubdussamad Looks like your recent PR takes care of all the requisites. Is something left to be done for this issue?

sethmlarson:
I think there are still more to work on here, might take a bit of discovery! One I can think of is `chunked`, I'm sure there are plenty more.

ubdussamad:
Hi @csurfer , yes there are a couple more, some of them are mentioned in my PR, for others you might have to a bit of digging.

sethmlarson:
I'm going to mark this as closed, I don't think there are any other parameters to be forwarded. Thanks everyone!

PR #2297, opened 2021-06-22
#2235, add chunked, body_pos and assert_same_host arguments to the to…

rootart:
…p level request method

### DIFF
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -99,6 +99,9 @@ def request(
     redirect: Optional[bool] = True,
     retries: Optional[Union[Retry, bool, int]] = None,
     timeout: Optional[Union[Timeout, float, int]] = 3,
+    chunked: bool = False,
+    body_pos: Optional[Union[int, object]] = None,
+    assert_same_host: bool = True,
 ) -> BaseHTTPResponse:
     """
     A convenience, top-level request method. It uses a module-global ``PoolManager`` instance.
@@ -117,4 +120,7 @@ def request(
         redirect=redirect,
         retries=retries,
         timeout=timeout,
+        chunked=chunked,
+        body_pos=body_pos,
+        assert_same_host=assert_same_host,
     )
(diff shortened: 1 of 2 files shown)


test/with_dummyserver/test_poolmanager.py
```diff
@@ -469,8 +469,20 @@ def test_top_level_request_with_timeout(self):
...
                 retries=None,
                 timeout=2.5,
+                chunked=False,
+                body_pos=None,
+                assert_same_host=True,
             )
 
+    def test_top_level_request_with_chunked(self):
+        res = request(
+            "GET", f"{self.base_url}/chunked", chunked=True, preload_content=False
```
sethmlarson, line 479: `chunked` refers to how the request body is encoded, not the response body. You need to specify `body=<something>` for it to do something.

src/urllib3/__init__.py
```diff
@@ -99,6 +99,9 @@ def request(
     redirect: Optional[bool] = True,
     retries: Optional[Union[Retry, bool, int]] = None,
     timeout: Optional[Union[Timeout, float, int]] = 3,
+    chunked: bool = False,
+    body_pos: Optional[Union[int, object]] = None,
```
sethmlarson, line 103: I'm unsure if we need `body_pos` and `assert_same_host` defined on the high-level helper as both are low level concepts. `body_pos` is used to signal initial body location in the case of rewinding on redirects, `assert_same_host` only matters at the `ConnectionPool` level or lower (`PoolManager` actually sets `assert_same_host=False` automatically in `urlopen`).

At this high level `redirects=True|False` is the appropriate mechanism to control cross-host redirects.

sethmlarson (changes requested):
Thanks for this, some thoughts for you.

Also if you're looking for more potential options, I believe `decode_content=True` would be an effective one!

sethmlarson:
@rootart Are you still interested in contributing this change?

sethmlarson:
Going to close this PR for inactivity.

PR #2299, opened 2021-06-23
Moved the metadata into setup.cfg

KOLANICH:
Added pyproject.toml
Used a well-known external dependency already used by multiple projects to extract version.
`setup.py` is kept for now, but unneeded and can be removed in later commits.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,5 +4,7 @@
 .github/CODEOWNERS          @sethmlarson @shazow
 src/urllib3/_version.py     @sethmlarson @shazow
 setup.py                    @sethmlarson @shazow
+setup.cfg                   @sethmlarson @shazow
+pyproject.toml              @sethmlarson @shazow
 ci/                         @sethmlarson @shazow
 .travis.yml                 @sethmlarson @shazow
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,8 +19,8 @@ jobs:
           python-version: 3.7
       - name: Check packages
         run: |
-          python3.7 -m pip install wheel twine rstcheck;
-          python3.7 setup.py sdist bdist_wheel;
+          python3.7 -m pip install wheel twine rstcheck build;
+          python3.7 -m build;
           rstcheck README.rst CHANGES.rst
           python3.7 -m twine check dist/*
   test:
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -2,6 +2,6 @@
 
 set -exo pipefail
 
-python3 -m pip install --upgrade twine wheel
-python3 setup.py sdist bdist_wheel
+python3 -m pip install --upgrade twine wheel build
+python3 -m build
 python3 -m twine upload dist/* -u $PYPI_USERNAME -p $PYPI_PASSWORD --skip-existing
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = ["setuptools", "wheel", "read_version[toml] >= 0.3"]
+build-backend = "setuptools.build_meta" 
+
+[tool.read_version]
+version = "src.urllib3._version:__version__"
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,22 +1,61 @@
+[metadata]
+name = urllib3
+author = Andrey Petrov
+author_email = andrey.petrov@shazow.net
+license = MIT
+description = HTTP library with thread-safe connection pooling, file post, and more.
+keywords = urllib, httplib, threadsafe, filepost, http, https, ssl, pooling
+url = https://urllib3.readthedocs.io/
+long_description = file: README.rst, CHANGES.rst
+long_description_content_type = text/x-rst
+classifiers =
+    Environment :: Web Environment
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3 :: Only
+    Programming Language :: Python :: Implementation :: CPython
+    Programming Language :: Python :: Implementation :: PyPy
+    Topic :: Internet :: WWW/HTTP
+    Topic :: Software Development :: Libraries
+project_urls =
+    Documentation = https://urllib3.readthedocs.io/
+    Code = https://github.com/urllib3/urllib3
+    Issue tracker = https://github.com/urllib3/urllib3/issues
+
+[options]
+package_dir=
+    = src
+packages =
+    urllib3
+    urllib3.contrib
+    urllib3.contrib._securetransport
+    urllib3.util
+python_requires = >=3.6, <4
+setup_requires = setuptools; wheel; toml; read_version >= 0.3
+
+[options.extras_require]
+brotli =
+    brotli>=1.0.9; platform_python_implementation == 'CPython'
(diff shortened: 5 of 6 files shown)


pquentin:
Thank you for your contribution! So we all would like to use the new and shiny packaging features, but we need to tread carefully to make sure we still support old Python and pip versions and that installation does not get slower.

For example, we made sure to fail gracefully if anyone tried to install urllib3 2.0 using Python 2.7, and you removed that code.

* I guess the first thing here to do is to establish what is the oldest pip/setuptools versions we support. Then we'll see what we can do.
* It's unclear to me when exactly we're going to fetch read_version, but we don't want to install any dependency at run time our install time.

sethmlarson (changes requested):
Thanks for opening this. It's always unfortunate having to "thumbs-down" a first contribution but I'm not aware of a way to move this change forward.

The error for Python 2.7 and urllib3 v2.0 isn't implementable without `setup.py` and we definitely want that error. If you and @pquentin agree I think we should close this PR.

Perhaps in the future though we can upgrade our packaging config once we're fairly confident we've swayed people using old tooling+Python versions to pin their version to `urllib3<2`?

Thanks again, I hope this won't discourage future contributions to urllib3, we always appreciate them from our community. :heart:

KOLANICH:
>The error for Python 2.7  isn't implementable without setup.py

Building a wheel (calling `setup.py`) should be decoupled from installing a wheel (calling `pip install --upgrade urllib.whl`). It is not quite correct to not to allow to build wheels on versions that don't support running them, and it is not quite correct not to allow running wheels on the versions they cannot be built on. These are different versions. Checking that a package can be run on the current version is done by pip. Checking that a package can be built on a version is just unneeded, one just builds and if it builds, it is OK. So the check is unneeded.

Also we haven't removed `setup.py` yet.

>and urllib3 v2.0

I don't understand what yoummean.

sethmlarson:
> It is not quite correct to not to allow to build wheels on versions that don't support running them, and it is not quite correct not to allow running wheels on the versions they cannot be built on. These are different versions.

> Checking that a package can be run on the current version is done by pip.

You'd be surprised how many people run super ancient versions of `pip` (knowingly or not).

Running the following query for yesterday's downloads...

```
SELECT details.installer.version, COUNT(*) AS c
FROM `the-psf.pypi.file_downloads`
WHERE (
  DATE(timestamp) = "2021-06-22"
  AND file.project = 'urllib3'
  AND details.installer.name = 'pip'
) GROUP BY details.installer.version
ORDER BY c DESC
```

pip version | count
-- | --
21.1.2 | 1779466
9.0.3 | 723906
21.0.1 | 376439
19.3.1 | 308286
20.1.1 | 290001
9.0.1 | 244688
20.0.2 | 236729
20.2.2 | 220853
20.3.4 | 173425
20.1 | 173018
18.1 | 172428
19.0.3 | 141841
21.1.1 | 139982
20.2.4 | 127098
20.2.3 | 123579
8.1.1 | 78091
6.1.1 | 75527
20.3.3 | 72689
19.1.1 | 69723
19.2.3 | 44421
20.3.1 | 43908
10.0.1 | 23456
8.1.2 | 16553
21.1 | 13917
20.2.1 | 12363
20.2 | 10899
18.0 | 10764
19.3 | 8943
21.0 | 8293
19.0.2 | 7880
19.2.1 | 7686
19.1 | 6424
1.5.4 | 6205
7.1.2 | 3441
19.0.1 | 3393
20.3 | 2994
7.1.0 | 2993
19.2.2 | 2172
20.0.dev0 | 1980
9.0.2 | 1294
1.5.6 | 1277
10.0.0 | 1218
19.0 | 1206
20.0.1 | 1185
21.0.dev0 | 853
9.0.0 | 637
20.3b1 | 544
19.2 | 422
20.3.2 | 182
6.0.8 | 175
20.2b1 | 174
10.0.0b2 | 54
8.0.2 | 23
7.0.1 | 22
6.0 | 7
21.1.dev0 | 3
20.0 | 2
20.1b1 | 2
7.0.3 | 2
6.0.6 | 2
1.5.5 | 2
7.1.1 | 1

- Total downloads: 5,563,422
- pip<9 downloads: 184,321 (3.31%)

So 3.3% of our installs come from pip versions that don't understand metadata to specify supported Python versions. We want those users to have a better experience than a `SyntaxError` on importing `urllib3` once we release breaking changes dropping support for Python <3.5.

> > and urllib3 v2.0
> 
> I don't understand what you mean.

You can read more here: https://urllib3.readthedocs.io/en/latest/v2-roadmap.html

venthur:
I'd also argue that putting almost everything into setup.cfg, but setup.py still has to be there otherwise `pip install -e .` doesn't work, doesn't really make sense. I'd rather keep the setup.py until there's a proper declarative solution that allows us to remove setup.py. I'm thinking pyproject.toml here. PEP518 and PEP621 are there and it seems that on the long run this fill will replace setup.py and setup.cfg (and maybe requirements.txt) but we're not there yet.

KOLANICH:
>but setup.py still has to be there otherwise pip install -e . doesn't work

Not quite true, in the latest `pip` it should work because of the patch by me.

>PEP518 and PEP621 are there and it seems that on the long run this fill will replace setup.py and setup.cfg

I guess `setuptools`, which looks stagnating (implementing support of PEP621 (without tests) looks like can take at most 2 days of full time work, but it has not been done yet), can be replaced by competing tools in the most of packages earlier. Anyway, if there is an existing `setup.cfg` where all the metadata resides, it will be much easier to move the metadata from it to `pyproject.toml`: it can be done fully automatically. `setup.cfg` support is present for more than 5 years and we can get the benefits it brings right now, not waiting for the events of indefinitely remote future.

pquentin:
You can't use the README file directly, because it contains a `raw` directive that works on GitHub but not on PyPI. This is what those lines were about:

https://github.com/urllib3/urllib3/blob/dfd8b580adb4f87388310c11b5c76d989bd6c4e3/setup.py#L47-L60

And this is why the CI currently fails. (And you also have a conflict.)

> Not quite true, in the latest pip it should work because of the patch by me.

I see this was released two months ago. By default Python versions won't come up with this version of pip, and this can be confusing. But this is mostly about contributor experience, not install/use experience, so it may not be a big issue. Anyway, the `raw` issue above is a bigger problem.

pquentin:
By the way, the latest Python version we support in the `main` branch is Python 3.6, and that is [bundled with pip 9.0.1 and setuptools 28.8.0](https://github.com/python/cpython/commit/756b45ab6e0d8d88f985cf449298145244bd317b), so I'd say anything that pip 9.0.1+ and setuptools 28.8.0+ support is fair game in the `main` branch. (Assuming we keep the Python 2 specific code in setup.py, that is.)

pquentin:
Since v2 is going to be Python 3.7+, we can assume pip 10 and setuptools 28.8 now: https://github.com/python/cpython/blob/v3.7.0/Lib/ensurepip/__init__.py.

To be clear, I'd be happy with a pull request that moves the declarative parts of setup.py to setup.cfg and switches to python -m build. But let's keep the code that checks for Python 2, that massages the README and that reads the version for now.

sethmlarson:
There hasn't been any movement on this PR in a few months. We'd still accept what @pquentin mentioned but in it's current state we can't accept this PR. If you'd like to pick up this work @KOLANICH ping us here and we can reopen or you can start a new PR. Thanks again!

bluetech:
Anyone doing this might be interested in [this tool](https://github.com/asottile/setup-py-upgrade) by asottile which does some of this automatically.

KOLANICH:
@bluetech, thanks. I used https://github.com/gvalkov/setuptools-py2cfg , but not quite it, but my fork of it with some of the stuff not yet merged into upstream https://github.com/KOLANICH-tools/setuptools-py2cfg/tree/private .

KOLANICH:
>There hasn't been any movement on this PR in a few months. We'd still accept what @pquentin mentioned but in it's current state we can't accept this PR. If you'd like to pick up this work @KOLANICH ping us here and we can reopen or you can start a new PR. Thanks again!

This PR main purpose is to eliminate `setup.py` completely. If it is not eliminated, it won't bring security benefit.

PR #2302, opened 2021-06-23
Docs: add print results where missing

euri10:
related to #2223 , bullet point 10


docs/user-guide.rst
```diff
@@ -354,7 +360,7 @@ recommended to set the ``Content-Type`` header:
     )
 
     print(json.loads(resp.data.decode("utf-8"))["data"])
-    # b"..."
+    # data:application/octet-stream;base64,...
```
euri10, line 363: here we load an image and print the decoded data, this is a string, not bytes

pquentin (approved):
Thank you!

PR #2303, opened 2021-06-23
Add type hints to urllib3.contrib.securetransport

hramezani:
These errors are related to `urllib3/contrib/_securetransport/bindings.py`
Because we ignored type hints on this file.
```
src/urllib3/contrib/securetransport.py:204: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:227: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:235: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:241: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:247: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:261: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:288: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:294: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:300: error: Returning Any from function declared to return "int"
src/urllib3/contrib/securetransport.py:440: error: Returning Any from function declared to return "int"
```


src/urllib3/contrib/securetransport.py
```diff
@@ -500,31 +523,33 @@ def handshake(
...
     # Copy-pasted from Python 3.5 source code
-    def _decref_socketios(self):
+    def _decref_socketios(self) -> None:
         if self._io_refs > 0:
             self._io_refs -= 1
         if self._closed:
             self.close()
 
-    def recv(self, bufsiz):
+    def recv(self, bufsiz: int) -> List[bytes]:
```
sethmlarson, line 544: This should be `-> bytes`?

src/urllib3/contrib/securetransport.py
```diff
@@ -500,31 +523,33 @@ def handshake(
                 result = Security.SSLHandshake(self.context)
 
                 if result == SecurityConst.errSSLWouldBlock:
-                    raise socket.timeout("handshake timed out")
+                    raise socket.timeout("handshake timed out")  # type: ignore
```
sethmlarson, line 526: Curious about this one. Mypy doesn't like `socket.timeout`?

sethmlarson (changes requested):
Thanks for this, comments for you!

As far as the `SecurityConstant.err*` not being typed as `int`s, maybe we could move the value into `_securetransport/low_level.py` since it's not technically a binding? It's in a private namespace so people *shouldn't* be expecting it to exist in any certain place?

src/urllib3/contrib/securetransport.py
hramezani: I think so but if I change it to `bytes` I get
`src/urllib3/contrib/securetransport.py:548: error: Incompatible return value type (got "List[bytes]", expected "bytes")`

Is it related to `data = buffer[:bytes_read]`?

src/urllib3/contrib/securetransport.py
hramezani: `src/urllib3/contrib/securetransport.py:526: error: Argument 1 to "timeout" has incompatible type "str"; expected "int"`

src/urllib3/contrib/securetransport.py
```diff
@@ -111,8 +130,8 @@
 # TLSv1 and a high of TLSv1.2. For everything else, we pin to that version.
 # TLSv1 to 1.2 are supported on macOS 10.8+
 _protocol_to_min_max = {
-    util.ssl_.PROTOCOL_TLS: (SecurityConst.kTLSProtocol1, SecurityConst.kTLSProtocol12),
-    util.ssl_.PROTOCOL_TLS_CLIENT: (
+    util.ssl_.PROTOCOL_TLS: (SecurityConst.kTLSProtocol1, SecurityConst.kTLSProtocol12),  # type: ignore
```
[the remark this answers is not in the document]
hramezani: `src/urllib3/contrib/securetransport.py:133: error: Module has no attribute "PROTOCOL_TLS"; maybe "PROTOCOL_TLS_CLIENT"?`

hramezani:
@sethmlarson Thanks for your review and great suggestion.
I've added a commit to move `SecurityConst` to `urllib3.contrib._securetransport.low_level`

src/urllib3/contrib/securetransport.py
sethmlarson: Probably? Might require a cast on the `return`?

sethmlarson (changes requested):
Almost there, few more comments for you.

src/urllib3/contrib/securetransport.py
```diff
@@ -171,7 +187,9 @@ def extract_from_urllib3():
     util.ssl_.USE_DEFAULT_SSLCONTEXT_CIPHERS = orig_util_USE_SYSTEM_SSL_CIPHERS
 
 
-def _read_callback(connection_id, data_buffer, data_length_pointer):
+def _read_callback(
+    connection_id: int, data_buffer: int, data_length_pointer: List[int]
```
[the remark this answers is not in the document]
hramezani: hmmm, Then we would have error like:
```
src/urllib3/contrib/securetransport.py:232: error: Unsupported target for indexed assignment ("Sequence[int]")
src/urllib3/contrib/securetransport.py:237: error: Unsupported target for indexed assignment ("Sequence[int]")
src/urllib3/contrib/securetransport.py:285: error: Unsupported target for indexed assignment ("Sequence[int]")
src/urllib3/contrib/securetransport.py:290: error: Unsupported target for indexed assignment ("Sequence[int]")
```

src/urllib3/contrib/securetransport.py
hramezani: Done!

src/urllib3/contrib/securetransport.py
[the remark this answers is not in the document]
sethmlarson: Ah we're writing to these buffers too, hmm can we use `bytearray` instead then?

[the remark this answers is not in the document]
hramezani: Great! Thanks!

sethmlarson (approved):
LGTM

PR #2304, opened 2021-06-24
[1.26] Run integration tests in GitHub Actions

pquentin:
A couple of notes:

 * I used nox instead of a collection of bash scripts to run the tests
 * I didn't bother trying to reduce the repetition between the two tests, following the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).
 * GitHub Actions does not have the fancy `if` Travis syntax: we can decide to run something conditionally based on target branches or modified files, [that's all](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions). Which is why I'm modifying `_version.py` here.
 * Modifying `_version.py` also allowed me to make sure we were really testing the urllib3 from the source, not something from pip. But I'm not necessarily suggesting adopting this `+dev` scheme in the future, even if it's handy in Trio.
 * I started with the 1.26 branch as we want to release it and it's more complicated, but plan to forward port to main if this gets accepted.


ci/requests.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/tests/test_requests.py b/tests/test_requests.py
```
sethmlarson, line 1: Hmm, is there something we can do to fix this upstream?

.github/workflows/integration.yml
```diff
@@ -0,0 +1,35 @@
+name: Downstream Integration
```
sethmlarson, line 1: Totally an aesthetic thing, but the status section is super verbose about these, can we make them more like the existing GHA statuses?

.travis.yml
```diff
@@ -22,43 +11,19 @@ notifications:
 
 env:
   global:
-    - PYTHONWARNINGS=always::DeprecationWarning
```
sethmlarson, line 25: Are we losing the `PYTHONWARNINGS=always::DeprecationWarning` with this change or is it enabled elsewhere?

sethmlarson (changes requested):
Woo! Very excited about this, thank you @pquentin! Here's a few comments:

ci/requests.patch
pquentin: Honestly I haven't investigated, but I will at least open an issue and spend a few minutes investigating

.github/workflows/integration.yml
pquentin: Any suggestions for a shorter name? Is "Downstream" too long too?

pquentin: (I agree that the current output is ugly but didn't know how to preserve clarity)

.travis.yml
pquentin: This is only useful for running our own tests and is now specified in noxfile.py

src/urllib3/_version.py
```diff
@@ -1,2 +1,2 @@
 # This file is protected via CODEOWNERS
-__version__ = "1.26.5"
+__version__ = "1.26.5+dev"
```
sethmlarson, line 2: Just clarifying that before merging this we're reverting this change?

.github/workflows/integration.yml
sethmlarson: I think "Downstream" is fine here

src/urllib3/_version.py
pquentin: Right, I reverted it.

.github/workflows/integration.yml
pquentin: Done

pquentin: Done

sethmlarson (approved):
LGTM

PR #2305, opened 2021-06-25
Add dns lookup error

### LINKED ISSUE #1003: DNS lookup error exception

euri10:
continuation of #1008 and #1827
fixes #1003 

WIP, looking mostly for some CI love at the moment

### DIFF
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -60,11 +60,8 @@ def create_connection(
     except UnicodeError:
         raise LocationParseError(f"'{host}', label empty or too long") from None
 
-    try:
-        res = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
-    except socket.gaierror as e:
-        raise e
-    for af, socktype, proto, canonname, sa in res:
+    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
+        af, socktype, proto, canonname, sa = res
         sock = None
         try:
             sock = socket.socket(af, socktype, proto)
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -62,8 +62,8 @@ def create_connection(
 
     try:
         res = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
-    except socket.gaierror:
-        raise
+    except socket.gaierror as e:
+        raise e
     for af, socktype, proto, canonname, sa in res:
         sock = None
         try:
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -60,8 +60,11 @@ def create_connection(
     except UnicodeError:
         raise LocationParseError(f"'{host}', label empty or too long") from None
 
-    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
-        af, socktype, proto, canonname, sa = res
+    try:
+        res = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+    except socket.gaierror:
+        raise
+    for af, socktype, proto, canonname, sa in res:
         sock = None
         try:
             sock = socket.socket(af, socktype, proto)
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -44,6 +44,7 @@ from ._version import __version__
 from .exceptions import (
     ConnectTimeoutError,
     HTTPSProxyError,
+    NameResolutionError,
     NewConnectionError,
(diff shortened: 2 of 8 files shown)


pquentin:
Thank you for tackling this! Please let us know when you'd like a review

euri10:
it seems raising socket.gaierror does not yield the same message on various platforms, so I had to remove the specific match in `with pytest.raises(socket.gaierror),  match=......):` in the `test.test_util.TestUtil.test_dnsresolver_error` because the CI [failed](https://github.com/urllib3/urllib3/runs/2912067602#step:8:100) on this test

windows: [Errno 11001] getaddrinfo failed in windows
linux: [Errno -2] Name or service not known

The docs say about the [error](https://docs.python.org/3/library/socket.html#socket.gaierror) that:
> The numeric error value will match one of the EAI_* constants defined in this module

so given the source [here](https://github.com/python/cpython/blob/bb3e0c240bc60fe08d332ff5955d54197f79751c/Modules/addrinfo.h#L58-L71) I'm not sure how to interpret the `11001` win error, I'm not familiar enough with how windows errors are "translated", would like some pointers here
```
#define EAI_ADDRFAMILY   1      /* address family for hostname not supported */
#define EAI_AGAIN        2      /* temporary failure in name resolution */
#define EAI_BADFLAGS     3      /* invalid value for ai_flags */
#define EAI_FAIL         4      /* non-recoverable failure in name resolution */
#define EAI_FAMILY       5      /* ai_family not supported */
#define EAI_MEMORY       6      /* memory allocation failure */
#define EAI_NODATA       7      /* no address associated with hostname */
#define EAI_NONAME       8      /* hostname nor servname provided, or not known */
#define EAI_SERVICE      9      /* servname not supported for ai_socktype */
#define EAI_SOCKTYPE    10      /* ai_socktype not supported */
#define EAI_SYSTEM      11      /* system error returned in errno */
#define EAI_BADHINTS    12
#define EAI_PROTOCOL    13
#define EAI_MAX         14
```

pquentin:
I'm pretty sure `EAI_*` is Unix or Linux-only, Windows uses completely different errors, as you've seen. See https://docs.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2 which tells us that `11001` is `WSAHOST_NOT_FOUND`

euri10:
I think it's OK for a review

I would have liked to add a more fine-grained test (with a match in the test.test_util.TestUtil.test_dnsresolver_error) which checks for the gaierror reason, but it's too OS specific, and I'm sure it adds lots of value.

I also tried to take into account [this comment](https://github.com/urllib3/urllib3/pull/1827#discussion_r401218612) from a previous attempt, but I found that if you do that, you'll get the NameResolutionError switched to a ProtocolError in [here](https://github.com/urllib3/urllib3/blob/3d8c93d96938afea4a7a9490b41d07d999d348c6/src/urllib3/connectionpool.py#L751-L752) because gaierror is a sublass of OSError (not sure I get this correctly though, I begin so I have the exception layout not quite well in mind !)

src/urllib3/util/connection.py
```diff
@@ -60,8 +60,11 @@ def create_connection(
     except UnicodeError:
         raise LocationParseError(f"'{host}', label empty or too long") from None
 
-    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
-        af, socktype, proto, canonname, sa = res
+    try:
+        res = socket.getaddrinfo(host, port, family, socket.SOCK_STREAM)
+    except socket.gaierror as e:
+        raise e
```
pquentin, line 66: @sigmavirus24 Do you remember what you concern was in https://github.com/urllib3/urllib3/pull/1827#pullrequestreview-388532636?

pquentin (approved):
This looks good to me! But I would like to know what bothered @sigmavirus24 in the previous PR

src/urllib3/util/connection.py
sigmavirus24: I think I'm confused as to why we're re-raising an exception with no other handling. Is it to shorten the traceback stack? What is the purpose of catching and then raising? (I'm on mobile right now so correct me if I'm missing something)

pquentin: You're right, this change itself seems useless. We could reraise as a `NameResolutionError` here instead or drop this change

pquentin (approved):
Thanks!

pquentin:
@euri10 Can you please look into this? We indeed want to retry name resolution errors

euri10:
Sure, I'll try come up with a test for this and resolution later today !

PR #2311, opened 2021-06-26
Update setup.cfg

scardracs:
As reference: https://bugs.gentoo.org/798606

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -5,11 +5,11 @@ max-line-length=99
 
 [metadata]
 license_file = LICENSE.txt
-provides-extra =
+provides_extra =
     secure
     socks
     brotli
-requires-dist =
+requires_dist =
     pyOpenSSL>=0.14; extra == 'secure'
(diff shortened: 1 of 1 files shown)


venthur:
Hi!

Thanks for the pull request. I was trying to find out which of the two versions is the correct one, but I couldn't find anything in the setuptools documentation. Do you have a link that would back, that this is the correct way?

scardracs:
Hi,

in Gentoo we have a command that checks for possible warnings such the setuptools one. Specifically the PR I've opened fix these 2 warnings:

* QA Notice: setuptools warnings detected:
 * 
 * 	Usage of dash-separated 'provides-extra' will not be supported in future versions. Please use the underscore name 'provides_extra' instead
 * 	Usage of dash-separated 'requires-dist' will not be supported in future versions. Please use the underscore name 'requires_dist' instead

venthur (approved):

sethmlarson:
Thanks @scardracs and @venthur!

PR #2312, opened 2021-06-29
Name lookup error on retry

euri10:
> We indeed want to retry name resolution errors

humble attempt at solving the above, I'm not sure if I correctly understood the original intent, in my mind it translates into the diff below:
```diff
+        elif isinstance(err, NameResolutionError):
+            return True
```

I also hesitated a lot where to add some tests, and while checking where that `_is_connection_error` function was called it seemed logical to put in the proxymanager tests where retry are used a lot, but I'm certain there might be other places where this could be interesting to have as well, I'm not familiar enough with the code base so I'd appreciate pointers on that if that's required

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -379,7 +379,7 @@ class UnrewindableBodyError(HTTPError):
     pass
 
 
-class NameResolutionError(HTTPError):
+class NameResolutionError(HTTPError, socket.gaierror):
     """Raised when host name resolution fails."""
 
     def __init__(self, host: str, reason: socket.gaierror):
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -382,9 +382,6 @@ class UnrewindableBodyError(HTTPError):
 class NameResolutionError(HTTPError, socket.gaierror):
     """Raised when host name resolution fails."""
 
-    original_error: Exception
-
-    def __init__(self, host: str, error: Exception) -> None:
-        message = f"Failed to resolve '{host}' ({error})"
-        HTTPError.__init__(self, message, error)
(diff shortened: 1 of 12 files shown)


src/urllib3/util/retry.py
```diff
@@ -344,6 +345,8 @@ def _is_connection_error(self, err: Exception) -> bool:
         """
         if isinstance(err, ProxyError):
             err = err.original_error
+        elif isinstance(err, NameResolutionError):
+            return True
         return isinstance(err, ConnectTimeoutError)
```
sethmlarson, line 350: Thoughts on `return isinstance(err, (ConnectTimeoutError, NameResolutionError))`?

src/urllib3/exceptions.py
```diff
@@ -382,6 +382,9 @@ class UnrewindableBodyError(HTTPError):
 class NameResolutionError(HTTPError, socket.gaierror):
     """Raised when host name resolution fails."""
 
-    def __init__(self, host: str, reason: socket.gaierror):
-        message = f"Failed to resolve '{host}' ({reason})"
-        HTTPError.__init__(self, message)
+    original_error: Exception
+
+    def __init__(self, host: str, error: Exception) -> None:
```
sethmlarson, line 387: Not sure if we need `original_error` on this exception?

src/urllib3/util/retry.py
euri10: funny how when I see this I felt like it was so obvious I dont get why I did not write it in the first place

src/urllib3/exceptions.py
euri10: I removed it

pquentin:
Closing per discussion in #1003

PR #2315, opened 2021-07-03
Change mypy session to log all errors.

### LINKED ISSUE #1897: Start shipping our own type stubs

hramezani:
Closes #1897

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -4,41 +4,6 @@ import subprocess
 
 import nox
 
-# Whenever type-hints are completed on a file it should be added here so that
-# this file will continue to be checked by mypy. Errors from other files are
-# ignored.
-TYPED_FILES = {
-    "src/urllib3/contrib/__init__.py",
-    "src/urllib3/contrib/_securetransport/bindings.py",
-    "src/urllib3/contrib/_securetransport/low_level.py",
-    "src/urllib3/contrib/ntlmpool.py",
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -159,13 +124,9 @@ def mypy(session):
     # Ensure that mypy itself ran successfully
     assert process.returncode in (0, 1)
 
-    for line in process.stdout.split("\n"):
```
sethmlarson, line 162: Can we change all this logic into a `session.run(mypy --strict src/)`?

sethmlarson (commented):
So exciting!! Looks like there's a lint failure too.

noxfile.py
hramezani: What kind of change do you mean?

sethmlarson: Instead of capturing the output of mypy and doing all this filtering we now don't expect any mypy errors so in theory can use `session.run()` directly?

sethmlarson (commented):
Amazing to see a ✅ next to the lint job, great work @hramezani 👏👏👏

sethmlarson (approved):

PR #2316, opened 2021-07-04
Run integration tests in GitHub Actions

pquentin:
Let's see how this goes


pquentin:
So the botocore tests run correctly! But we have four requests failures:

 * `test_urllib3_pool_connection_closed` (not sure why yet, but it seems legit)
 * `TestRequests.test_proxy_error` (possibly due to the bug mentioned by @bluetech in https://github.com/urllib3/urllib3/issues/1003#issuecomment-870600144)
 * `TestRequests.test_https_warnings` and `TestRequests.test_auth_is_stripped_on_http_downgrade` due to commonName removal, it seems

Pretty encouraging, actually!

sethmlarson:
It's both exciting and frightening seeing commonName not work 😀

pquentin:
Cycling to get the `ProxyError` fix

pquentin:
Nice, that fixed it. There's one more error to understand, the other two are definitely due to commonName!

pquentin:
The two `commonName` errors were different: one was legit, another was not, so I opened https://github.com/psf/requests/pull/5869.

sethmlarson:
https://github.com/psf/requests/pull/5869 has been merged :tada: Now we only have one commonName failure and one related to the `PoolKey`?

pquentin:
That's right! And now that 1/ `PoolKey`issue is fixed and 2/ I marked the `commonName` test as `xfail`, the integration tests are working!

They're quite stable in my experience and faster than PyPy, should I include them in every pull request, at least until we release 2.0?

sethmlarson:
Yes let's run them on every PR. We don't want any unexpected drifts suprising us near release time as we've seen has already happened.

pquentin:
@sethmlarson Three related PRs later, this is ready for the actual review! (I haven't managed to request a review from my phone)

sethmlarson (approved):
LGTM, thanks for all of your efforts here! :muscle:

PR #2318, opened 2021-07-08
Remove a few `# type: ignore`

pquentin:
I figured out there would be a few quick wins now that everything is typed. (Note: `import a as a` is a way to tell mypy that a should be considered as a reexport, without having to define `__all__`.)

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -7,7 +7,7 @@ import sys
 import warnings
 from copy import copy
 from http.client import HTTPConnection as _HTTPConnection
-from http.client import HTTPException  # noqa: F401
+from http.client import HTTPException as HTTPException  # noqa: F401
 from socket import timeout as SocketTimeout
 from typing import (
     IO,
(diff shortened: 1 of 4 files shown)


hramezani (approved):
LGTM!

src/urllib3/connectionpool.py
```diff
@@ -744,17 +744,18 @@ def urlopen(  # type: ignore
             # Discard the connection for these exceptions. It will be
             # replaced during the next _get_conn() call.
             clean_exit = False
+            new_e: Exception = e
```
pquentin, line 747: @sethmlarson @hramezani What do you think about that `new_e` name? I had to use another name because mypy does not support assignments that change the type, but I did not find a good name. Now we have `e`, `new_e` and `err`.

hramezani: Yes, you are right. I think we have to do it like this.

sigmavirus24:
What's the point of avoiding `__all__`?

src/urllib3/connection.py
```diff
@@ -7,7 +7,7 @@
 import warnings
 from copy import copy
 from http.client import HTTPConnection as _HTTPConnection
-from http.client import HTTPException  # noqa: F401
+from http.client import HTTPException as HTTPException  # noqa: F401
```
sigmavirus24, line 10: Note, if you just used `__all__` you could also get rid of the `# noqa` here

sethmlarson (approved):
I'm happy with the current PR.

Using `__all__` in all modules is a good idea though, maybe we can tackle that in another PR as it'd take more analysis across the whole package?

PR #2321, opened 2021-07-08
Update mypy to 0.910

hramezani:

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -305,7 +305,8 @@ class IncompleteRead(HTTPError, httplib_IncompleteRead):
     expected: int
 
     def __init__(self, partial: int, expected: int) -> None:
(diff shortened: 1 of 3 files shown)


src/urllib3/exceptions.py
```diff
@@ -305,7 +305,8 @@ class IncompleteRead(HTTPError, httplib_IncompleteRead):
     expected: int
 
     def __init__(self, partial: int, expected: int) -> None:
-        super().__init__(partial, expected)
```
hramezani, line 308: `Argument 1 to "__init__" of "IncompleteRead" has incompatible type "int"; expected "bytes"`

sethmlarson: Huh, `partial` is definitely supposed to be `bytes` but we're using it with `int`?

sethmlarson (approved):
LGTM

PR #2322, opened 2021-07-09
Remove leading white spaces when parsing input

BanzaiMan:
When a user passes strings such as "<tt>&nbsp;&nbsp;&nbsp;&nbsp;https&colon;//google.com</tt>", `parse_url` silently fails. If this method is used by another utility, it is often difficult for the user to understand that the leading white space is there, leading to some lost time.

This PR tries to be friendly and removes the leading white space from the user input to avoid this situation. #653 talks about removing white spaces in fragments, but I didn't see anything that addresses the user input in entirety.

Additionally, the trailing white spaces in the user inputs should be kept because they may need to be encoded.

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -365,6 +365,7 @@ def parse_url(url: str) -> Url:
     if not url:
         # Empty
         return Url()
+    url = url.lstrip()
 
     source_url = url
     if not _SCHEME_RE.search(url):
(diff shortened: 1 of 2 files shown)


bluetech:
IMO this is better done by the application if it wants to. A library should be strict, this way it can accommodate all types of users.

sethmlarson:
Yeah I'm kinda in agreement with @bluetech, WHATWG-URL technically specifies stripping whitespace before parsing but that behavior is already achievable for projects wanting to sanitize user input. In a way WHATWG-URL is specifying behavior for browsers along with "the rest of us" but urllib3 doesn't have the same "user" as someone typing something into the omnibar.

Going to close this for now, thank you for opening though!

PR #2323, opened 2021-07-11
Remove unnecessary elif after return from urllib3.util.util

hramezani:

### DIFF
--- a/src/urllib3/util/util.py
+++ b/src/urllib3/util/util.py
@@ -9,7 +9,7 @@ def to_bytes(
         return x
     elif not isinstance(x, str):
         raise TypeError(f"not expecting type {type(x).__name__}")
-    elif encoding or errors:
+    if encoding or errors:
         return x.encode(encoding or "utf-8", errors=errors or "strict")
     return x.encode()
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
It may be unnecessary but when skimming the code I think the elif is easier rather than a skim implying that those conditions might compound

src/urllib3/util/util.py
```diff
@@ -7,9 +7,9 @@ def to_bytes(
 ) -> bytes:
     if isinstance(x, bytes):
         return x
-    elif not isinstance(x, str):
+    if not isinstance(x, str):
```
sethmlarson, line 10: I would argue the `elif not isinstance(x, str)` here is easier to read since this and the above block are semantically linked. Agree that the bottom `if` statement shouldn't be an `elif` though!

src/urllib3/util/util.py
```diff
@@ -19,9 +19,9 @@ def to_str(
 ) -> str:
     if isinstance(x, str):
         return x
-    elif not isinstance(x, bytes):
+    if not isinstance(x, bytes):
```
sethmlarson, line 22: Same as above

sethmlarson (approved):
LGTM

PR #2327, opened 2021-07-13
Fix RecentlyUsedContainer setitem when maxsize is 0

### LINKED ISSUE #2326: Fix `RecentlyUsedContainer` regression
#2115 added typing to the `_collections` module. Unfortunately, it also introduced a regression that breaks `test_urllib3_pool_connection_closed` in the requests test suite, as noticed in https://github.com/urllib3/urllib3/pull/2316/checks?check_run_id=3049608329. `__setitem__` is implemented like this in 1.26.x:

https://github.com/urllib3/urllib3/blob/ba95e9eac73452d3bccfb5413b00d9a4fe3e4c31/src/urllib3/_collections.py#L61-L74

And it's now implemented like this in main:

https://github.com/urllib3/urllib3/blob/baba8c2b4aabda8907a777039baa3f8758e4333a/src/urllib3/_collections.py#L133-L155

When `len(self._container)` == `self._maxsize` == 0, `self._container[key] = value` is now called *after* `self._container.popitem(last=False)`, which is why that `popitem` call raises an exception.

Applying this hacky diff fixes the requests test:

```diff
diff --git a/src/urllib3/_collections.py b/src/urllib3/_collections.py
index 32d5330e..798ba432 100644
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -139,15 +139,15 @@ class RecentlyUsedContainer(Generic[_KT, _VT], MutableMapping[_KT, _VT]):
                 # size of the pool. Because accessing a key should move it to
                 # the end of the eviction line, we pop it out first.
                 evicted_item = key, self._container.pop(key)
+                self._container[key] = value
             except KeyError:
+                self._container[key] = value
                 if len(self._container) >= self._maxsize:
                     # If we didn't evict an existing value, and we've hit our maximum
                     # size, then we have to evict the least recently used item from
                     # the beginning of the container.
                     evicted_item = self._container.popitem(last=False)
 
-            # Finally, insert the new value.
-            self._container[key] = value
 
         # After releasing the lock on the pool, dispose of any evicted value.
         if evicted_item is not None and self.dispose_func:
```

@haikuginger What are your thoughts on this? Is it possible to revert to the original logic?

pquentin:
Co-authored-by: Seth Michael Larson <sethmichaellarson@gmail.com>

Closes #2326 (where the code review happened)

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -139,15 +139,16 @@ class RecentlyUsedContainer(Generic[_KT, _VT], MutableMapping[_KT, _VT]):
                 # size of the pool. Because accessing a key should move it to
                 # the end of the eviction line, we pop it out first.
                 evicted_item = key, self._container.pop(key)
+                self._container[key] = value
             except KeyError:
-                if len(self._container) >= self._maxsize:
+                # When the key does not exist, we insert the value first so that
+                # evicting works in all cases, including when self._maxsize is 0
+                self._container[key] = value
+                if len(self._container) > self._maxsize:
                     # If we didn't evict an existing value, and we've hit our maximum
                     # size, then we have to evict the least recently used item from
                     # the beginning of the container.
                     evicted_item = self._container.popitem(last=False)
-
-            # Finally, insert the new value.
-            self._container[key] = value
 
         # After releasing the lock on the pool, dispose of any evicted value.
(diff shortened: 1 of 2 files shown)


sethmlarson (approved):
LGTM, thanks!

PR #2329, opened 2021-07-15
style(__init__): Added typing for StreamHandler

SaidBySolo:
This pr makes my linter less complaining.


SaidBySolo:
Not backward compatible

pquentin:
What do you mean, "not backward compatible"?

SaidBySolo:
> What do you mean, "not backward compatible"?
The ``SteamHandler[TextIO]`` syntax does not work in version 3.8.

pquentin:
I see, thanks. No, unfortunately we support Python 3.6+

PR #2330, opened 2021-07-15
Generate changelog with towncrier

### LINKED ISSUE #2085: Start using towncrier for tracking changes
This will make sure that we don't miss anything. Asking if anyone on our team has configured towncrier in the past :)

pquentin:
I also added the changelog to the docs, inspired by pip (even though they use a plugin to reformat their changes).

Here's what this looks like for now: https://urllib3--2330.org.readthedocs.build/en/2330/changelog.html

Screenshot 2021-07-16 at 09 56 29

The goal of this PR is to reproduce the existing changelog with towncrier, and make sure it looks correctly in docs. In follow-up PRs we can go through the 2.0 commits and add them to newsfragments.

Closes #2085

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -199,6 +199,7 @@ class BaseHTTPResponse(io.IOBase):
         version: int,
         reason: Optional[str],
         decode_content: bool,
+        request_url: Optional[str],
     ) -> None:
         if isinstance(headers, HTTPHeaderDict):
             self.headers = headers
@@ -208,6 +209,7 @@ class BaseHTTPResponse(io.IOBase):
         self.version = version
         self.reason = reason
         self.decode_content = decode_content
+        self.request_url: Optional[str]
 
         self.chunked = False
         tr_enc = self.headers.get("transfer-encoding", "").lower()
@@ -412,6 +414,7 @@ class HTTPResponse(BaseHTTPResponse):
             version=version,
             reason=reason,
             decode_content=decode_content,
+            request_url=request_url,
         )
 
         self.retries = retries
@@ -423,7 +426,10 @@ class HTTPResponse(BaseHTTPResponse):
         self._original_response = original_response
         self._fp_bytes_read = 0
         self.msg = msg
-        self._request_url = request_url
+        if self.retries is not None and self.retries.history:
+            self._request_url = self.retries.history[-1].redirect_location
+        else:
+            self._request_url = request_url
 
         if body and isinstance(body, (str, bytes)):
             self._body = body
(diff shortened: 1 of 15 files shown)


src/urllib3/response.py
```diff
@@ -906,6 +906,10 @@ def url(self) -> Optional[str]:
         else:
             return self._request_url
 
+    @url.setter
+    def url(self, url: str) -> None:
+        self._request_url = url
+
```
pquentin, line 912: This is a hacky workaround to allow Sphinx to set the URL:

https://github.com/sphinx-doc/sphinx/blob/564439e66d81693edd08fa5973a14151af976ca8/sphinx/ext/intersphinx.py#L118-L120

"hacky" because it does not work when there are redirects. (We should first agree that whether we want to allow this or not. I think we do.)

sigmavirus24 (approved):
I love this. Towncrier's a pleasure to use in Twine

pquentin:
@sigmavirus24 Thank you for your support, and thanks for mentioning twine! This gave me the good idea or renaming the obscure `newsfragment` directory into `changelog`.

src/urllib3/response.py
sethmlarson: I think allowing this is fine.

sethmlarson (dismissed):
Woo, this is very nice. Thanks for tackling this Quentin! :tada:

Are we okay to merge now or is another test case required for setting `url`?

pquentin:
> Are we okay to merge now or is another test case required for setting url?

I'm not happy with the way this works in the `if self.retries is not None and self.retries.history:` case. But now that I look at the 1.26.x code, `url` was just not a thing in the past: we had `geturl()`. So maybe it's fine to say that Sphinx is at fault here and should not attach an `url` attribute? (They have a really fast release cycle, and I can probably get that fix in.)

bluetech (approved):
I've had a good experience with towncrier in pytest. It's nice for the PR submitter (can update the changelog without fear of conflicts), the release manager (doesn't need to create the changelog themselves), and for reviewers (the changelog is nice to have when reviewing).

I recommend adding a README file to the `changelog` directory explaining various things like
- The format of the filename (PR/issue number, category, rst)
- Which tense to use (present, past, etc.) with examples.
- Which categories (removal, feature, etc.) are possible and when to use them.

Even though it duplicates the contributing docs, it's very helpful for a quick reference, even for regular contributors.

One slight annoyance I've had is when submitting a PR without an issue, you don't know the PR ID before submitting it, so must do a push + immediate force push for the changelog name. Don't know if there's a fix for that, but it's pretty minor anyway.

sethmlarson (approved):
Very excited about this! I only have one question:

src/urllib3/_version.py
```diff
@@ -1,2 +1 @@
-# This file is protected via CODEOWNERS
```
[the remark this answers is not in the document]
pquentin: It's not useful until we have automated releases. I added it back anyway as it wasn't related to that pull request.

pquentin:
Please take another look:

- I rebased to cleanly separate the towncrier and url changes. (Please don't squash.)
- I added a README in the changelog directory, using pytest as an inspiration.
- I addressed Seth's comment.

bluetech (approved):

PR #2331, opened 2021-07-17
WIP: Add multipart/form-data streaming encoder

sigmavirus24:


sethmlarson:
Very exciting, thanks for opening this! Let me know when you're ready for reviews.

src/urllib3/fields.py
```diff
@@ -12,7 +13,7 @@
     cast,
 )
 
-_TYPE_FIELD_VALUE = Union[str, bytes]
+_TYPE_FIELD_VALUE = Union[str, bytes, BinaryIO]
```
sigmavirus24, line 16: The docstring for all of this shows using a file here so this was missing but it had the knock on effect of rewriting the header params which were overloading the definition of a "FIELD_VALUE"

src/urllib3/filepost.py
```diff
@@ -71,9 +71,10 @@ def encode_multipart_formdata(
 
         if isinstance(data, int):
             data = str(data)  # Backwards compatibility
-
-        if isinstance(data, str):
+        elif isinstance(data, str):
             writer(body).write(data)
+        elif isinstance(data, BinaryIO):
```
sigmavirus24, line 76: Again, we called out accepting files but weren't handling them?

test/with_dummyserver/test_connectionpool.py
```diff
@@ -281,6 +282,22 @@ def test_unicode_upload(self) -> None:
...
+                "upload_key": "value",
+                "upload_file": (
+                    "filename.txt",
+                    open(__file__, "rb"),
+                    {"Content-Type": "text/plain"},
+                ),
+            }
+        )
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            r = pool.request("POST", "/upload", headers=body.request_headers, body=body)
```
sigmavirus24, line 298: I suspect this will fail mypy but I thought it'd be good to have an end-to-end test here. It's also failing against the dummyserver for some reason I haven't figured out quite yet, but I'm working on that

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+
+from .. import _collections
+from .encoder import encode_with
+
+if typing.TYPE_CHECKING:
+    from .. import response as _response
+
+
+@typing.overload
+def _split_on_find(content: str, bound: str) -> typing.Tuple[str, str]:
```
sethmlarson, line 13: Can we use .partition() instead of this function?

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+class NonMultipartContentTypeError(Exception):
+    pass
+
+
+def _header_parser(
+    headers: bytes, encoding: str
+) -> typing.Sequence[typing.Tuple[str, str]]:
+    string = headers.decode(encoding)
+    items = email.parser.HeaderParser().parsestr(string).items()
+    items = typing.cast(typing.List[typing.Tuple[str, str]], items)
```
sethmlarson, line 42: If it's only changing to match the return value you can use a type ignore to avoid a call to typing.cast()

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+class NonMultipartContentTypeError(Exception):
+    pass
+
+
+def _header_parser(
+    headers: bytes, encoding: str
+) -> typing.Sequence[typing.Tuple[str, str]]:
+    string = headers.decode(encoding)
```
sethmlarson, line 40: I'm not sure if we allow users to set the headers encoding elsewhere, will need to check.

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+    encoding.
+
+    The encoding may be overridden by specifying ``part.encoding = '...'``.
+    """
+
+    def __init__(self, content: bytes, encoding: str):
+        self.encoding = encoding
+        headers: typing.Dict[str, str] = {}
+        # Split into header section (if any) and the content
+        if b"\r\n\r\n" in content:
```
sethmlarson, line 60: Can we use partition and check the middle element of the returned 3-tuple to avoid doing two scans for this bytestring?

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+    @staticmethod
+    def _fix_first_part(part: bytes, boundary_marker: bytes) -> bytes:
+        bm_len = len(boundary_marker)
+        if boundary_marker == part[:bm_len]:
+            return part[bm_len:]
+        else:
+            return part
+
+    def _parse_body(self, content: bytes) -> None:
+        boundary = b"".join((b"--", self.boundary))
```
sethmlarson, line 138: Used a bytes literal + boundary

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+
+        def test_part(part: bytes) -> bool:
+            return (
+                part != b""
+                and part != b"\r\n"
+                and part[:4] != b"--\r\n"
+                and part != b"--"
+            )
+
+        parts = content.split(b"".join((b"\r\n", boundary)))
```
sethmlarson, line 152: Bytes literal + boundary

src/urllib3/fields.py
sethmlarson: Hmm, I'm not sure if files were supported. Iirc it was `open (x).read()`? Need to check on PC

src/urllib3/multipart/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""Multipart support for urllib3."""
+
+from .decoder import (
```
sethmlarson, line 3: Should we hide the encoder/decoder submodules and reassign the names in this module? (Ie from .encoder import x as x)

sethmlarson (commented):
Thanks for this! Here's some initial comments from mobile:

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
+"""Logic for parsing and decomposing a multipart response body."""
+import email.parser
+import typing
+
+from .. import _collections
+from .encoder import encode_with
+
+if typing.TYPE_CHECKING:
```
hramezani, line 8: I to make the typing compatible with other files, would suggest to don't use `typing.x` and importing them as `from typing import x,y,...`

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,165 @@
...
+        from requests_toolbelt import MultipartDecoder
+
+        decoder = MultipartDecoder(content, content_type)
+        for part in decoder.parts:
+            print(part.headers['content-type'])
+
+    For both these usages, there is an optional ``encoding`` parameter. This is
+    a string, which is the name of the unicode codec to use (default is
+    ``'utf-8'``).
+
```
hramezani, line 104: This blank line can be removed
```suggestion
```

sethmlarson (changes requested):
More comments for you, this time from the comfort of a PC.

A general comment, if it's possible I'd like to move away from the `.len` property. You likely understand the historical significance more than I but if there's a way we can accomplish this with only using `__len__` and `.tell()`+`.seek()` that'd be best imo

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+
+def readable_data(
+    data: typing.Union[typing.AnyStr, typing.BinaryIO], encoding: str
+) -> typing.Union[typing.BinaryIO, _CustomBytesIO]:
+    """Coerce the data to an object with a ``read`` method."""
+    if hasattr(data, "read"):
+        data = typing.cast(typing.BinaryIO, data)
+        return data
+
+    data = typing.cast(typing.BinaryIO, typing.AnyStr)
```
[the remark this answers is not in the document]
sigmavirus24: It's wrong, not unnecessary 🙃

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+        return f"{self.len}"
+
+    @property
+    def request_headers(self) -> typing.Dict[str, str]:
+        return {
+            "Content-Type": self.content_type,
+            "Content-Length": self.content_length,
+        }
+
+    def getall(self) -> bytes:
```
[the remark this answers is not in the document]
sigmavirus24: I received complaints on toolbelt about not having something that didn't imply a need for a read size :shrug: I don't care either way

src/urllib3/multipart/decoder.py
sigmavirus24: Yeah, apparently `HeaderParser().parsestr()` returns a `Mapping[str, Any]` but I don't know why it uses that :shrug:

Generally, I find `cast` to be the explicit and `type: ignore` to be the implicit and the former explains why we're doing it, the latter looks like we're satisfying a computer gone wrong

src/urllib3/multipart/decoder.py
sigmavirus24: So to be very clear, this isn't Request headers, in `multipart/form-data` each part can have its own headers. And we're letting the user specify the encoding for handling the response bytes, so why wouldn't we use that for handling the headers too?

src/urllib3/multipart/decoder.py
sigmavirus24: I find that importing things from the typing module pollutes namespaces. I can consolidate to that here but I really hate the practice of importing from typing rather than importing typing or aliasing the import typing to something shorter

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+        self,
+        fields: Fields,
+        boundary: typing.Optional[str] = None,
+        encoding: str = "utf-8",
+    ):
+        #: Boundary value either passed in by the user or created
+        self.boundary_value: str = boundary or uuid.uuid4().hex
+
+        # Computed boundary
+        self.boundary: str = f"--{self.boundary_value}"
```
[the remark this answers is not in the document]
sigmavirus24: Generally speaking, this isn't there for the users, but we can make these private/read-only as necessary. It's just more to write everything we use frequently as `_{name}` because we don't trust users to be intelligent and not muck with things.

src/urllib3/multipart/encoder.py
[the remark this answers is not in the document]
sethmlarson: Let's stick with `.read()` for now, thanks!

src/urllib3/multipart/encoder.py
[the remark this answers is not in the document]
sethmlarson: Yeah, users will be users. :shrug: At least if these attributes are private we can say "told you so" if it ends up being a problem later.

src/urllib3/multipart/decoder.py
hramezani: I can see both ways of (`import typing` or `from typing import x`) in different projects.
I am not against any of them. I would suggest going for one of them in all projects.

BTW, I am ok with `import typing` as well. I can update the rest of the files of the project in a PR.

@sethmlarson what do you think?

src/urllib3/multipart/decoder.py
sethmlarson: Looks like `email.parser.HeaderParser().parsestr()` returns `email.message.Message` which looks like it's trying to be a `Mapping`.

[From typeshed](https://github.com/python/typeshed/blob/master/stdlib/email/message.pyi) it looks like `email.message.Message.items()` returns `List[Tuple[str, Any]]`, which explains why we need the cast. Carry on!

src/urllib3/multipart/decoder.py
sethmlarson: Ah gotcha, that makes more sense since this `encoding` applies to the entire response body. Thanks for making this right in my head!

src/urllib3/fields.py
sigmavirus24: Hm, so I guess requests always papered over this: https://github.com/psf/requests/blob/d09659997cd1e3eca49a07c59ece5557071c0ab9/requests/models.py#L160

I don't know how we consolidate these APIs though in such a way that folks don't need to read the entire file into memory.

sigmavirus24:
> A general comment, if it's possible I'd like to move away from the `.len` property. You likely understand the historical significance more than I but if there's a way we can accomplish this with only using `__len__` and `.tell()`+`.seek()` that'd be best imo

So today, urllib3 is more-or-less architecture agnostic.  We know from the harassment of the cryptography project for adopting Rust that there are people still operating older architectures and compiling things from scratch for themselves to make that work. On the face of things, I'd guess 90+% of our users are not on 32-bit architectures but we had a fair number of users of requests + urllib3 that were on 32-bit architectures that couldn't use the MultipartEncoder because `__len__` had restrictions (even on Py3) due to the underlying system. I think we can definitely migrate to `__len__` but I thoroughly suspect that we should head this off by explicitly documenting the difference. The other thing is that `MultipartEncoder` _probably_ doesn't need anything other than being able to calculate it's own length for the headers - unless we want users to be able to pass this into Requests as well in which case we need to provide something for them.

Also, I haven't done a deep reconsideration of the design here just the minimal effort to migrate it here. This was designed to work best within Requests and its idiosyncrasies.

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+        # Pre-encoded boundary
+        self._encoded_boundary: bytes = b"".join(
+            [
+                encode_with(self.boundary, self.encoding),
+                encode_with("\r\n", self.encoding),
+            ]
+        )
+
+        #: Fields provided by the user
+        self.fields = fields
```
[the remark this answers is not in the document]
sigmavirus24: It's so much extra boilerplate to do what? What motivates you wanting significantly more code to make things read-only here? Is the idea if this is read-only users won't search for the private attributes storing the data and then will report bugs that updating the private attributes didn't do what they expected versus what happened changing the undocumented attributes that are 'public' and writable? How adversarial a relationship do you want to have with our users?

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+
+
+class FileNotSupportedError(Exception):
+    """The file provided by the user is not supportable."""
+
+
+P = typing.TypeVar("P", bound="Part")
+
+
+class Part:
```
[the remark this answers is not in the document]
sigmavirus24: I don't think we can since we _want_ to cache the length of the part and compute it and I don't think it makes sense to have a tuple that has that as an attribute since you're so opposed to the idea of having an attribute for that.

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+import contextlib
+import io
+import os
+import typing
+import uuid
+
+from .. import fields
+
+
+class FileNotSupportedError(Exception):
```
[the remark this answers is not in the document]
sigmavirus24: I think I got rid of its use

src/urllib3/multipart/decoder.py
sethmlarson: Let's switch to the namedspaced usage with `typing.X`.

src/urllib3/multipart/decoder.py
sigmavirus24: Done

src/urllib3/multipart/decoder.py
sigmavirus24: We can use `libcst` to do it automagically for us if we want.

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+    @property
+    def content_type(self) -> str:
+        return f"multipart/form-data; boundary={self.boundary_value}"
+
+    @property
+    def content_length(self) -> str:
+        return f"{self.len}"
+
+    @property
+    def request_headers(self) -> typing.Dict[str, str]:
```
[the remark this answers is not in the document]
sigmavirus24: Done

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+        self._buffer = _CustomBytesIO(encoding=encoding)
+
+        # Pre-compute each part's headers
+        self._prepare_parts()
+
+        # Load boundary into buffer
+        self._write_boundary()
+
+    @property
+    def len(self) -> int:
```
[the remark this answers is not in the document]
sigmavirus24: Switched this to `__len__`

src/urllib3/multipart/encoder.py
```diff
@@ -0,0 +1,676 @@
...
+
+    def _calculate_length(self) -> int:
+        """
+        This uses the parts to calculate the length of the body.
+
+        This returns the calculated length so __len__ can be lazy.
+        """
+        boundary_len = len(self.boundary)  # Length of --{boundary}
+        # boundary length + header length + body length + len('\r\n') * 2
+        self._len = (
```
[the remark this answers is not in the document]
sigmavirus24: Done

pquentin:
Regarding CI, there are two issues left:

 * 35 lines are not covered, which is against our policy of 100% coverage. We can always exclude those lines from coverage, but it would be nice to add tests.
 * The docs is complaining due to our new nitpicky settings. @hramezani Given your experience here, would you mind taking a look at the errors and tell what you think? For reference I copied them below.

```
/home/docs/checkouts/readthedocs.org/user_builds/urllib3/envs/2331/lib/python3.7/site-packages/urllib3/multipart/decoder.py:docstring of urllib3.multipart.decoder.MultipartDecoder.from_response:: WARNING: py:class reference target not found: _response.HTTPResponse
/home/docs/checkouts/readthedocs.org/user_builds/urllib3/envs/2331/lib/python3.7/site-packages/urllib3/multipart/decoder.py:docstring of urllib3.multipart.decoder.MultipartDecoder.from_response:: WARNING: py:class reference target not found: urllib3.multipart.decoder.MD
/home/docs/checkouts/readthedocs.org/user_builds/urllib3/envs/2331/lib/python3.7/site-packages/urllib3/multipart/decoder.py:docstring of urllib3.multipart.MultipartDecoder.parts:: WARNING: py:class reference target not found: urllib3.multipart.decoder.BodyPart
```

pquentin:
The docs are fine now, coverage is left. (And also make sure all concerns are addressed)

spacether:
What's the status of this? Can it be merged? Openapi generator users want this and we use this lib.

nioncode:
@sigmavirus24 are you still working on this? I'm thinking of taking this over to get this over the line.

pquentin:
@nioncode Please do. The main task is figuring out what is left to do.

skydiver1980 (approved):

ecerulm:
@sigmavirus24 this has been open for  3 years almost, shall we close this or you have any plan to revive this PR?

sigmavirus24:
The project wants to land it. I don't have the time for it and others have done work intermittently to get it over the line

PR #2332, opened 2021-07-20
Drop support for Python 3.6

sethmlarson:
This PR only effects the v2.x development branch, v1.26.x will continue to support Python 3.6.

The reason for this PR is two-fold:
- We're unlikely to release v2.0 before Python 3.6 being EOL in December of 2021.
- There are multiple features in the v2.0 timeline that will be dramatically easier to implement without having to think about Python 3.6 (see https://github.com/urllib3/urllib3/pull/2220 for an in-progress example)

I think removing now is best to not impede development and if by some stroke of luck we complete all v2.0 deliverables in the Python 3.6 timeline we'll re-evaluate.

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -13,6 +13,7 @@ from .url import _BRACELESS_IPV6_ADDRZ_RE, _IPV4_RE
 SSLContext = None
 SSLTransport = None
 HAS_SNI = False
+HAS_NEVER_CHECK_COMMON_NAME = False
 IS_PYOPENSSL = False
 IS_SECURETRANSPORT = False
 ALPN_PROTOCOLS = ["http/1.1"]
@@ -44,6 +45,7 @@ try:  # Do we have ssl at all?
     import ssl
     from ssl import (  # type: ignore
         CERT_REQUIRED,
+        HAS_NEVER_CHECK_COMMON_NAME,
         HAS_SNI,
         OP_NO_COMPRESSION,
         OP_NO_TICKET,
@@ -278,7 +280,7 @@ def create_urllib3_context(
         context.check_hostname = False
         context.verify_mode = cert_reqs
 
-    if hasattr(context, "hostname_checks_common_name"):
+    if HAS_NEVER_CHECK_COMMON_NAME:
         context.hostname_checks_common_name = False
 
     # Enable logging of TLS session keys via defacto standard environment variable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,26 +13,26 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
-      - name: Set up Python 3.7
+      - name: Set up Python 3
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3
       - name: Check packages
         run: |
-          python3.7 -m pip install wheel twine rstcheck;
-          python3.7 setup.py sdist bdist_wheel;
+          python3 -m pip install wheel twine rstcheck;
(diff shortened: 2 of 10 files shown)


pquentin:
Thanks! This looks good, now we need to remove the Python 3.6 specific code, eg. threading is always available, RLock exists, and any other `Python 3.6` comments.

pquentin:
Deleted my previous comment as I hadn't seen all the code, sorry.

Barring the codecov issue, this looks good but it would be nice if our type experts @haikuginger and @hramezani could take a look as types are evolving rapidly. I think @haikuginger mentioned in the past that 3.6 made things more difficult type wise, and that RLock wasn't the only issue.

And I guess mentioning @hugovk is always a good idea when you stop supporting a Python version :p

shazow (commented):
Looks sensible. I'm not intimately familiar with these details but let me know if you require my approval. :)

haikuginger:
This looks correct in general; I'm just going to re-acquaint myself with why the `RLock` compat code was 3.6-only before approving.

haikuginger (approved):
LGTM; this is exciting! Found the doc where `threading` was made non-optional in 3.7.

pquentin:
I'll look into the HOSTFLAG_NEVER_CHECK_SUBJECT issue. (Its availability depends on OpenSSL, even on CPython.)

hramezani:
I think we have `typing.OrderedDict` in `Python > 3.7`.
So, I think we can `typing.OrderedDict` use in https://github.com/urllib3/urllib3/blob/8c905b09b2fef0d254e4350e52bf301348b2a8b0/src/urllib3/_collections.py#L112

hramezani:
I think we can remove `nox > Session test-3.6 was successful.` and a line for `3.10` maybe from https://github.com/urllib3/urllib3/blob/5b61732f2403c378bb60b451d95ce2835f6b5485/docs/contributing.rst

hramezani:
There is an `Optimized for Python 3.6+`  section in https://github.com/urllib3/urllib3/blob/2ef05adfcf001c85f1f95ae3fa078d16190851b0/docs/v2-roadmap.rst
which may need some change.

src/urllib3/util/ssl_.py
```diff
@@ -278,7 +280,7 @@ def create_urllib3_context(
         context.check_hostname = False
         context.verify_mode = cert_reqs
 
-    if hasattr(context, "hostname_checks_common_name"):
+    if HAS_NEVER_CHECK_COMMON_NAME:
         context.hostname_checks_common_name = False
```
sethmlarson, line 284: Should we also defensively swallow `AttributeError` on this line?

pquentin: I'm reluctant to do this: this behavior is clearly documented in Python 3.7+ and PyPy and CPython share the same simple implementation.

This also won't be just False in 2.0 now that we know exactly what CPython and OpenSSL versions work correctly.

pquentin:
There's another error message that changed on PyPy, that I still need to fix, but the `hostname_checks_common_name` bug on our end is fixed.

pquentin:
CI is green again, we just need to address the comments from @hramezani now.

hramezani:
@SethMichaelLarson @pquentin I can address them if you want.

pquentin:
Yes, please do! You should be allowed to push to Seth's branch directly.

I also noticed another change needed in `test_ssltransport.py` (search for 3.7)

hramezani:
@sethmlarson It seems I don't have permission to push. I've created a new PR https://github.com/urllib3/urllib3/pull/2336

hramezani:
Closing in favor of https://github.com/urllib3/urllib3/pull/2336

PR #2335, opened 2021-07-21
Add decode_content argument to the top level request method.

### LINKED ISSUE #2325: Add decode_content parameter to top-level APIs
Like the title says, add `decode_content` to the top-level APIs `urllib3.request()`.

See https://github.com/urllib3/urllib3/commit/ddf7361ac0467431a2f3df6ba346c9c506c29d56 for an example.

imkaka:
Closes #2325 

I was not sure about the exact way to test this, found out `/encodingrequest` and tried to test when `decode_content=False` where we won't we raising `DecodeError`.
Please let me know if it needs improvements.


test/with_dummyserver/test_poolmanager.py
```diff
@@ -424,6 +424,16 @@ def test_top_level_request_with_preload_content(self):
         r.data
         assert r.connection is None
 
+    def test_top_level_request_with_decode_content(self):
```
sethmlarson, line 427: Can we change this test to make a request with `Accept-Encoding: gzip` and `decode_content=False` and see that we get back gzipped bytes, then make another request with `decode_content=True` and see we get back decompressed bytes

sethmlarson (changes requested):
Thanks for opening this!! Sorry I didn't get a review on it sooner, one comment to improve the test case and we should be good to go.

test/with_dummyserver/test_poolmanager.py
imkaka: Thanks @sethmlarson 
It's done, please review when you get some time :)

sethmlarson (approved):
Thank you for this, what a great first contribution! :sparkles: 

If you enjoyed contributing this we'd love to help you find your next issue to work on. There's a [list of issues for v2.0](https://github.com/urllib3/urllib3/milestone/6) that are our current highest priority. If you need help finding an issue that's good for you drop a message on Discord.

Again thank you very much, hope to see you more! :)

PR #2336, opened 2021-07-22
Drop support for Python 3.6

hramezani:

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -12,13 +12,9 @@ from typing import (
     MutableMapping,
     NoReturn,
     Optional,
-    Set,
-    Tuple,
-    TypeVar,
-    Union,
-    cast,
-    overload,
 )
+from typing import OrderedDict as OrderedDictType
+from typing import Set, Tuple, TypeVar, Union, cast, overload
 
 if TYPE_CHECKING:
 
@@ -90,7 +86,7 @@ class RecentlyUsedContainer(Generic[_KT, _VT], MutableMapping[_KT, _VT]):
         ``dispose_func(value)`` is called.  Callback which will get called
     """
 
-    _container: "OrderedDict[_KT, _VT]"
+    _container: OrderedDictType[_KT, _VT]
     _maxsize: int
     dispose_func: Optional[Callable[[_VT], None]]
     lock: RLock
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -245,7 +245,7 @@ class HTTPHeaderDict(MutableMapping[str, str]):
 
     def __init__(self, headers: Optional[ValidHTTPHeaderSource] = None, **kwargs: str):
         super().__init__()
-        self._container = dict()  # 'dict' is insert-ordered in Python 3.7+
+        self._container = {}  # 'dict' is insert-ordered in Python 3.7+
         if headers is not None:
             if isinstance(headers, HTTPHeaderDict):
                 self._copy_from(headers)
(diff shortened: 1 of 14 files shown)


src/urllib3/_collections.py
```diff
@@ -268,7 +245,7 @@ class HTTPHeaderDict(MutableMapping[str, str]):
 
     def __init__(self, headers: Optional[ValidHTTPHeaderSource] = None, **kwargs: str):
         super().__init__()
-        self._container = _ordered_dict()
+        self._container = dict()  # 'dict' is insert-ordered in Python 3.7+
```
sethmlarson, line 248: nit: Can do an empty dict literal here

sethmlarson (approved):
LGTM, thanks for this. One nit for you:

pquentin:
@sethmlarson can you please remove the PyPy build from the required checks? The name has changed to 3.7 and it's unreliable enough to not be required.

V1NAY8:
@hramezani  Just did some digging Should we also update these comments or its underlying code?
https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/connection.py#L486

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/util/ssl_.py#L195

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/util/ssl_.py#L383

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/util/ssl_match_hostname.py#L1

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/util/ssl_match_hostname.py#L12

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/contrib/pyopenssl.py#L286

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/contrib/securetransport.py#L535

hramezani:
Thanks @V1NAY8 for checking.
I think we can remove `(requires Python3.4+) `.
But, I think we can keep `# Copy-pasted from Python 3.5 source code `.

What do you think @sethmlarson @pquentin?

pquentin:
@hramezani Agreed.

Regarding https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/ssl_match_hostname.py, it's from Python 3.5 now, not 3.3, see https://github.com/urllib3/urllib3/commit/f620d997134708b09560ca5797aa79a59a2ef4c0. Also, it was modified to remove commonName support, which should be mentioned.

hramezani:
Anything to do here? @sethmlarson Have you checked @pquentin [comment](https://github.com/urllib3/urllib3/pull/2336#issuecomment-885527422)?

sethmlarson (approved):
Thanks all for carrying this so far! I removed `pypy3` from the required checks on GitHub for the `main` branch, is there anything else I'm missing @pquentin and @hramezani?

test/test_ssltransport.py
```diff
@@ -367,13 +367,10 @@ def test_wrong_sni_hint(self):
         with self.client_context.wrap_socket(
             sock, server_hostname="localhost"
         ) as proxy_sock:
-            with pytest.raises(Exception) as e:
+            with pytest.raises(ssl.SSLError):
```
pquentin, line 370: Can we catch https://docs.python.org/3/library/ssl.html#ssl.SSLCertVerificationError instead?

sethmlarson (approved):

pquentin (approved):

PR #2339, opened 2021-07-25
[v2] Remove urllib3.contrib.ntlmpool module

bluetech:
The module is likely broken and unlikely to be in use.

The module has been deprecated in urllib3 1.26.6 release and an advance
warning was given in its changelog.

Refs #2282.


sethmlarson (approved):
This is awesome, thanks much! :rocket:

PR #2340, opened 2021-07-25
Explicitly chain or don't chain exceptions

### LINKED ISSUE #2313: [v2] Chaining exceptions
### Context

Currently urllib3 doesn't chain exceptions explicitly when wrapping exceptions (this feature wasn't available in Python 2 AFAIK). For example:

https://github.com/urllib3/urllib3/blob/1831327b881880ed871f96f56a6977d360042e1b/src/urllib3/response.py#L539-L557

If the `SocketTimeout` happens, then it renders as

```
... SocketTimeout stacktrace ...

... ReadTimeoutError stacktrace ...
```

Explicit chaining would be

```py
 except SocketTimeout as e:
     raise ReadTimeoutError(self._pool, None, "Read timed out.") from e
```

then the error is the nicer

```
... SocketTimeout stacktrace ...

... ReadTimeoutError stacktrace ...
```

Alternatively, if we don't want to expose the cause exception, can do

```py
 except SocketTimeout:
     raise ReadTimeoutError(self._pool, None, "Read timed out.") from None
```

then only the `ReadTimeoutError` is shown.

IMO, the first case looks like a bug to the user, and, if this issue is resolved, would indicate an unexpected exception in urllib3.
So I suggest going over all cases of exception wrapping in urllib3 and changing them to either explicitly chain or suppress the wrapped exception.

There's is a question whether an exception's `__cause__` could be considered part of the API promise. IMO, there is no such expectation and we shouldn't guarantee this. But if we do, then we'd want to be a lot more picky with chaining vs. suppressing.

### Contribution

Yes.

bluetech:
Explicitly chain exceptions (or disable the chaining) instead of creating implicit "During handling of the above exception, another exception occurred" which look like a bug.

Since urllib3 is now Python 3 only, we can use the built-in syntax for this.

Methodology: grepped for `raise` and went over each case, deciding whether to chain or `from None`. When in doubt I preferred to chain.

Fixes #2313.


src/urllib3/contrib/_securetransport/bindings.py
```diff
@@ -416,8 +416,8 @@ def load_cdll(name: str, macos10_16_path: str) -> CDLL:
     CoreFoundation.CFStringRef = CFStringRef
     CoreFoundation.CFDictionaryRef = CFDictionaryRef
 
-except AttributeError:
-    raise ImportError("Error initializing ctypes")
+except AttributeError as e:
+    raise ImportError("Error initializing ctypes") from e
```
sethmlarson, line 420: This can be `from None`

src/urllib3/contrib/_securetransport/low_level.py
```diff
@@ -109,7 +109,7 @@ def _create_cfstring_array(lst: List[bytes]) -> CFMutableArray:
     except BaseException as e:
         if cf_arr:
             CoreFoundation.CFRelease(cf_arr)
-        raise ssl.SSLError(f"Unable to allocate array: {e}")
+        raise ssl.SSLError(f"Unable to allocate array: {e}") from e
```
sethmlarson, line 112: This can be `from None`

sethmlarson (changes requested):
Nice! Only have a couple of comments for you then this LGTM:

src/urllib3/contrib/_securetransport/bindings.py
bluetech: Done! (The other one to)

sethmlarson (dismissed):
LGTM, thanks for this!!

PR #2342, opened 2021-07-25
 Fix bad error handling given an invalid trust bundle in contrib.securetransport

### LINKED ISSUE #2338: Possible bug in securetransport _cert_array_from_pem() error handling
While working on an unrelated change, I noticed this line, which *seems* to miss a `raise` after the free in the `except` block. I may be wrong; I didn't try at all to understand what the code does, it just looks wrong to me so thought I'd report it.

https://github.com/urllib3/urllib3/blob/21945b84b0707a3143b09ade5c59e8dbf6a69a0f/src/urllib3/contrib/_securetransport/low_level.py#L186-L206

bluetech:
Fixes #2338.

### DIFF
--- a/src/urllib3/contrib/_securetransport/low_level.py
+++ b/src/urllib3/contrib/_securetransport/low_level.py
@@ -202,6 +202,7 @@ def _cert_array_from_pem(pem_bundle: bytes) -> CFArray:
         # We only want to do that if an error occurs: otherwise, the caller
         # should free.
         CoreFoundation.CFRelease(cert_array)
+        raise
 
     return cert_array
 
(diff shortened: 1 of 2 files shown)


bluetech:
I abused the CI a bit to run the test without the fix, indeed it [crashes](https://github.com/urllib3/urllib3/pull/2342/checks?check_run_id=3155486299) the interpreter with `Fatal Python error: Illegal instruction`. After the fix the test passes.

sethmlarson (approved):
LGTM! Thanks for finding and fixing this. We should probably backport to 1.26.x as well.

PR #2343, opened 2021-07-25
Use a lower timeout on CI jobs

### LINKED ISSUE #2341: ci: putting a time limit on jobs
Some CI runs got stuck and only failed after 6 hours, example: https://github.com/urllib3/urllib3/actions/runs/1064318368

Even though it's free it's seems like a big waste, so I propose we set a time limit on all jobs. I think 10 minutes is reasonable (from a glance it seems most jobs take 1m-2m). The `timeout-minutes: ...` directive does this.

bluetech:
If a CI job gets stuck, wait 10 or 20 minutes instead of 6 hours.

Fixes #2341.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ defaults:
 jobs:
   package:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
 
     steps:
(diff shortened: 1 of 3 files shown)


sethmlarson (approved):
LGTM, thanks!

PR #2352, opened 2021-07-29
Add mypy configuration file

bluetech:
This allows text editors, standalone `mypy` runs, etc. to get the correct configuration.

The flags correspond to `--strict` as of mypy 0.910 + `--show-error-codes`.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -156,7 +156,7 @@ def mypy(session):
     session.install("idna>=2.0.0")
     session.install("cryptography>=1.3.4")
     session.run("mypy", "--version")
-    session.run("mypy", "--strict", "src/urllib3")
+    session.run("mypy", "src/urllib3")
 
 
 @nox.session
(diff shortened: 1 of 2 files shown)


hramezani (approved):

sethmlarson:
@bluetech Should we keep `--strict` in Nox in case additional options are added that are auto-ticked by `--strict`? Maybe that's not a concern for us?

bluetech:
Since urllib3 pins the mypy version (as it should), using `--strict` would alert the person updating mypy when a new strict flag is added, which is nice. But I think that once we have a config file, I prefer `nox` to be fully consistent with it, and hope that the one who updates mypy would notice the new flag.

BTW, initially I also added `warn_unreachable = True`, which is in my experience very useful, but it triggers several warnings so will do it separately.

sethmlarson (approved):
Sounds good, thank you!

Issue #2252, opened 2021-06-03
Investigate removing `object` from the `Union[...]` of many `timeout` types


sethmlarson:
`object` is only there because of the `socket` module "default" sentinel value which is kind-of an unfortunate hack if you ask me.

It's causing a lot of problems type-wise when going from urllib3 timeout to http.client timeouts, let's take a peek if we can instead remove `object` and add `type: ignore` only where the socket default timeout sentinel is used.

pquentin:
Related discussion: https://github.com/encode/httpx/issues/1384

hramezani:
I've tested it on my own setup. by removing `object`, we will have to add 9 `type: ignore` and 9 `type: ignore` will be removed.
I can create a PR if you agree

PR #2353, opened 2021-07-30
Remove object from Timeout types.

hramezani:
Closes https://github.com/urllib3/urllib3/issues/2252
Thanks, @sethmlarson for the solution.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -6,7 +6,7 @@ import sys
 import warnings
 from http.client import HTTPResponse as _HttplibHTTPResponse
 from socket import timeout as SocketTimeout
-from typing import TYPE_CHECKING, Any, Mapping, Optional, Type, Union, overload
+from typing import TYPE_CHECKING, Any, Mapping, Optional, Type, Union, cast, overload
 
 from .connection import (
     _TYPE_BODY,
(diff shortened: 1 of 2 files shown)


src/urllib3/connectionpool.py
```diff
@@ -56,10 +56,10 @@
 
 log = logging.getLogger(__name__)
 
-_Default = object()
+_Default: float = cast(float, object())
 
 
-_TYPE_TIMEOUT = Union[Timeout, int, float, object]
+_TYPE_TIMEOUT = Union[Timeout, int, float]
```
bluetech, line 62: Unrelated to the PR: in python typing (unlike in runtime), `int` is actually a sub-type of `float`, so the `int` here is redundant. However it makes to keep for clarity.

src/urllib3/connectionpool.py
```diff
@@ -56,10 +56,10 @@
 
 log = logging.getLogger(__name__)
 
-_Default = object()
+_Default: float = cast(float, object())
```
bluetech, line 59: Some other options:

- Mark the sentinels as `Any`, then you don't need the cast, but the default parameter assignment still works. And people would perhaps be less tempted to use them as actual floats.

- Add a real sentinel type (e.g. a one-item enum) and add it to the types and all. This is completely type safe but somewhat annoying.

I think all approaches are probably fine.

bluetech (approved):

src/urllib3/connectionpool.py
hramezani: Thanks, @bluetech for the review and suggestion.
Yeah, you are right. we have to add a real sentinel type (based on the discussion in the dev channel)

hramezani:
Closing per @bluetech suggestion and discussion with other maintainers)

Issue #2344, opened 2021-07-25
Type check tests?


bluetech:
Currently the tests are not typed and not type checked. In my experience, doing so is useful, foremost as a way for testing the typing of the main code itself. But others think it adds too much friction when writing tests.

I propose we do it, as a way to gain some confidence in our own typing. There are two ways to go about it:

1. Do a sweep and add `-> None` to all test functions, and possibly fixture arguments etc.
2. Use mypy `check_untyped_defs = True` for the tests. This means mypy will type check even when the test is not annotated.

I personally prefer 2 because I prefer to be strict and explicit, but don't mind either way.

sethmlarson:
This sounds like a good idea, maybe we can adopt a similar construction for incrementally adding types like in the package? cc @hramezani

hramezani:
Sounds good to me as well.
I can help on this topic

bluetech:
WIP here: https://github.com/bluetech/urllib3/commits/tests-typing

sethmlarson:
Closing this issue as we've completed type hints on all tests!

PR #2354, opened 2021-08-01
Add type hints to dummyserver

bluetech:
As part of #2344. The tests use dummyserver, so type it first.

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -8,6 +8,7 @@ import zlib
 from datetime import datetime, timedelta
 from http.client import responses
 from io import BytesIO
+from typing import Dict, Optional, Sequence, Tuple, Union
 from urllib.parse import urlsplit
 
 from tornado import httputil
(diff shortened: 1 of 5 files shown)


dummyserver/handlers.py
```diff
@@ -19,33 +20,43 @@
...
 
-            request_handler.write(body)
+
+RETRY_TEST_NAMES: Dict[str, int] = collections.defaultdict(int)
 
 
-RETRY_TEST_NAMES = collections.defaultdict(int)
+def request_params(request: httputil.HTTPServerRequest) -> Dict[str, bytes]:
+    params = {}
+    for k, v in request.arguments.items():
```
sethmlarson, line 57: This is changing where we get `request.params` to `request.arguments`, it looks like these arguments can come from both the querystring or the body, is that what we're intending here? (I can't find a reference to `HTTPServerRequest.params` in the Tornado docs)

dummyserver/testcase.py
```diff
@@ -121,8 +137,13 @@ class HTTPDummyServerTestCase:
     host_alt = "127.0.0.1"  # Some tests need two hosts
     certs = DEFAULT_CERTS
 
+    io_loop: ClassVar[ioloop.IOLoop]
```
sethmlarson, line 140: TIL about `ClassVar`, nice!

sethmlarson (approved):
LGTM, one question for you though about a potential change in behavior.

dummyserver/handlers.py
bluetech: I should have explained this part. Previously the `_call_method` function monkey-patched the `params` attribute onto the request like so:

```py
        req.params = {}
        for k, v in req.arguments.items():
            req.params[k] = next(iter(v))
```

but mypy doesn't like this, so instead of assigning and using `request.params`, I use `params = request_params(request))` as needed.

PR #2355, opened 2021-08-01
Fix several problems in test/benchmark.py

bluetech:
- `urllib.urlopen` became `urllib.request.urlopen` in Python 3.

- Adapt sys path for the src/ layout.

- `/apis/contact` URL returns 404 now, which raises an exception in the
  urllib3 benchmark.

### DIFF
--- a/test/benchmark.py
+++ b/test/benchmark.py
@@ -7,9 +7,9 @@ urllib to demonstrate the usefulness of connection re-using.
 
 import sys
 import time
-import urllib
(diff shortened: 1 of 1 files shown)


sethmlarson:
I'm actually in favor of removing this file, I don't think it's been used by anyone since 2014 and possibly even longer ago. Is this okay with others?

bluetech:
I'll send a PR to remove it unless someone beats me to it.

PR #887, opened 2016-06-09
New docs

jonparrott:
For #884.

Preview here: http://urllib3.theadora.io/

So far:
* Switched to alabaster theme.
* First run at the homepage.
* Started on the user guide.

For the moment I just wanted to give @shazow a chance to look it over & a way to keep track of my progress, not quite ready for feedback yet.

PR #2357, opened 2021-08-03
Added example of custom ssl context usage

### LINKED ISSUE #2320: Document how to use custom SSLContext objects
### Context

What are you trying to do?
Provide an example of how to configure urllib3 to enable Partial Certificate chain validation.

How do you expect to do it? Document update with an example or Provided as a standard option/flag in Poolmanager.

Is it something you currently cannot do? Right

Is this related to an existing issue/problem? Yes. There is no option to configure enabling of partial chain validation.

### Alternatives

### Duplicate

Has the feature been requested before? No idea.

### Contribution

Would you be willing to submit a PR? Unplanned.

euri10:
FIxes #2320 
Basically it is a rip-off of a section in the previously deleted  (in #887) `docs/security.rst`, I just adapted the example adding a flag as per our conversation !

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -458,7 +458,7 @@ It is possible to provide a specific `SSL Context objects`_ directly to urllib3
 to exercise fine-grained control over the SSL configuration used in urllib3.
 
 For the purposes of compatibility, if you intend to customise an ``SSLContext``
-object we *strongly* recommend you obtain one from the following factory
+object we recommend you obtain one from the following factory
 function:
 
 .. autofunction:: urllib3.util.ssl_.create_urllib3_context
@@ -471,6 +471,7 @@ and then makes a HTTPS request to Google:
 .. code-block:: python
 
     import ssl
+
     from urllib3 import PoolManager
     from urllib3.util.ssl_ import create_urllib3_context
 
@@ -478,8 +479,8 @@ and then makes a HTTPS request to Google:
     ctx.load_default_certs()
     ctx.options |= ssl.OP_NO_RENEGOTIATION
 
-    p = PoolManager(ssl_context=ctx)
-    r = p.urlopen('GET', 'https://www.google.com/')
+    with PoolManager(ssl_context=ctx) as p:
+        r = p.urlopen("GET", "https://www.google.com/")
 
 Any customisation that is possible with the ``SSLContext`` object is possible here.
 
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -459,8 +459,8 @@ of compatibility, we recommend you obtain one from
 
 Once you have a context object, you can mutate it to achieve whatever effect
 you'd like. For example, the code below loads the default SSL certificates, sets
-the :data:`ssl.OP_NO_TICKET<python:ssl.OP_NO_TICKET>` flag that isn't set by
-default, and then makes a HTTPS request:
+the :data:`ssl.OP_ENABLE_MIDDLEBOX_COMPAT<python:ssl.OP_ENABLE_MIDDLEBOX_COMPAT>`
+flag that isn't set by default, and then makes a HTTPS request:
 
 .. code-block:: python
 
@@ -471,7 +471,7 @@ default, and then makes a HTTPS request:
 
     ctx = create_urllib3_context()
     ctx.load_default_certs()
-    ctx.options |= ssl.OP_NO_TICKET
+    ctx.options |= ssl.OP_ENABLE_MIDDLEBOX_COMPAT
 
     with PoolManager(ssl_context=ctx) as pool:
         pool.request("GET", "https://www.google.com/")
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -448,3 +448,39 @@ To enable this simply define environment variable `SSLKEYLOGFILE`:
 
 Then configure the key logfile in `Wireshark <https://wireshark.org>`_, see
 `Wireshark TLS Decryption <https://wiki.wireshark.org/TLS#TLS_Decryption>`_ for instructions.
+
+Custom SSL Contexts
+-------------------
+
+.. versionadded:: 1.17.0
+
+It is possible to provide a specific `SSL Context objects`_ directly to urllib3
+to exercise fine-grained control over the SSL configuration used in urllib3.
+
+For the purposes of compatibility, if you intend to customise an ``SSLContext``
+object we *strongly* recommend you obtain one from the following factory
+function:
+
+.. autofunction:: urllib3.util.ssl_.create_urllib3_context
+
+Once you have a context object, you can mutate that to achieve whatever affect
+you'd like. For example, the code below loads the default SSL certificates on
+Linux systems, sets the `OP_NO_RENEGOTIATION <https://docs.python.org/3/library/ssl.html#ssl.OP_NO_RENEGOTIATION>`_  flag that isn't set by default,
+and then makes a HTTPS request to Google:
+
+.. code-block:: python
+
+    import ssl
+    from urllib3 import PoolManager
+    from urllib3.util.ssl_ import create_urllib3_context
+
+    ctx = create_urllib3_context()
+    ctx.load_default_certs()
+    ctx.options |= ssl.OP_NO_RENEGOTIATION
+
+    p = PoolManager(ssl_context=ctx)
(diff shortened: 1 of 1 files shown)


venthur (approved):
LGTM

docs/advanced-usage.rst
```diff
@@ -448,3 +448,39 @@ To enable this simply define environment variable `SSLKEYLOGFILE`:
...
+Custom SSL Contexts
+-------------------
+
+.. versionadded:: 1.17.0
+
+It is possible to provide a specific `SSL Context objects`_ directly to urllib3
+to exercise fine-grained control over the SSL configuration used in urllib3.
+
+For the purposes of compatibility, if you intend to customise an ``SSLContext``
+object we *strongly* recommend you obtain one from the following factory
```
pquentin, line 461: We can remove "strongly". When this was written the alternative was to create your own SSLContext, which is insecure, but now the alternative is to use `ssl.create_default_context` which is OK.

docs/advanced-usage.rst
```diff
@@ -448,3 +448,39 @@ To enable this simply define environment variable `SSLKEYLOGFILE`:
...
+and then makes a HTTPS request to Google:
+
+.. code-block:: python
+
+    import ssl
+    from urllib3 import PoolManager
+    from urllib3.util.ssl_ import create_urllib3_context
+
+    ctx = create_urllib3_context()
+    ctx.load_default_certs()
```
pquentin, line 478: This line is already called by `create_urllib3_context`, please remove it.

docs/advanced-usage.rst
```diff
@@ -448,3 +448,39 @@ To enable this simply define environment variable `SSLKEYLOGFILE`:
...
+    ctx.options |= ssl.OP_NO_RENEGOTIATION
+
+    p = PoolManager(ssl_context=ctx)
+    r = p.urlopen('GET', 'https://www.google.com/')
```
pquentin, line 482: Can you please use the context manager form?

pquentin (changes requested):
Thanks! This looks good, we just need to modernize it slightly. Can you please format the code snippet with black and isort?

docs/advanced-usage.rst
euri10: this is not and without the snippet fails: 
```    
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 696, in urlopen
    httplib_response = self._make_request(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 397, in _make_request
    self._validate_conn(conn)
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1009, in _validate_conn
    conn.connect()
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connection.py", line 496, in connect
    self.sock = ssl_wrap_socket(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 416, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(sock, context, tls_in_tls, server_hostname)
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 460, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/home/lotso/.asdf/installs/python/3.9.5/lib/python3.9/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/home/lotso/.asdf/installs/python/3.9.5/lib/python3.9/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/home/lotso/.asdf/installs/python/3.9.5/lib/python3.9/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)
  File "<input>", line 10, in <module>
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/poolmanager.py", line 423, in urlopen
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 786, in urlopen
    return self.urlopen(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 786, in urlopen
    return self.urlopen(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 786, in urlopen
    return self.urlopen(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 756, in urlopen
    retries = retries.increment(
  File "/home/lotso/PycharmProjects/urllib3/.nox/test-3-9/lib/python3.9/site-packages/urllib3/util/retry.py", line 495, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))  # type: ignore
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='www.google.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)')))

```

euri10:
Implemented the requested changes @pquentin except for https://github.com/urllib3/urllib3/pull/2357#discussion_r684758018

pquentin (commented):
Sorry, I had so many comments when looking at the result that I ended up rewriting the whole thing. I also changed the option used in the example as it only made sense for a server.

pquentin:
The result is here: https://urllib3--2357.org.readthedocs.build/en/2357/advanced-usage.html#custom-ssl-contexts

Note that `create_default_context` docstring does not appear in the API Reference. I tried adding `ssl_` but I think the trailing underscores prevents it from being added.

docs/advanced-usage.rst
```diff
@@ -448,3 +448,34 @@ To enable this simply define environment variable `SSLKEYLOGFILE`:
...
+-------------------
+
+You can exercise fine-grained control over the urllib3 SSL configuration by
+providing a :class:`ssl.SSLContext <python:ssl.SSLContext>` object. For purposes
+of compatibility, we recommend you obtain one from
+:func:`~urllib3.util.ssl_.create_urllib3_context`.
+
+Once you have a context object, you can mutate it to achieve whatever effect
+you'd like. For example, the code below loads the default SSL certificates, sets
+the :data:`ssl.OP_NO_TICKET<python:ssl.OP_NO_TICKET>` flag that isn't set by
```
sethmlarson, line 462: [The default `create_ssl_context` does set `OP_NO_TICKET`](https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/ssl_.py#L258), maybe we should change this example?

sethmlarson (dismissed):
This looks good, thanks for working on this! Just one thought before merging:

docs/advanced-usage.rst
pquentin: Ugh, right, that's why we didn't include it in the first place. I'm actually surprised to see all the things that can be done without a custom SSL Context.

Honestly, the reason most users wanted to do that in requests was to add back an insecure cipher, but I'm not sure you'll be happy to see that documented. :D

The other thing left are verify_flags, which is what prompted this whole thing, but I don't understand well what they do.

sethmlarson: If we're just looking for an option we can use as an example, maybe `OP_CIPHER_SERVER_PREFERENCE`?

pquentin: Thanks, fixed! It's only available starting with Python 3.8+, but at least it's not set by default.

sethmlarson (approved):
This PR is good, I'll note that we're recommending users use a member in a "technically private" module `urllib3.util.ssl_`. This is one API that I'm kinda happy users are using though, because they get brought on the ride of what "should" be a default configured `SSLContext` (and we reserve the ability to add to that definition in future versions to protect our users).

PR #2359, opened 2021-08-05
Test type hints

hramezani:
I am not sure about `Callable[..., Any]` that I used in decorators.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -36,6 +36,8 @@ if TYPE_CHECKING:
 
 
 _RT = TypeVar("_RT")  # return type
+_TestFuncT = TypeVar("_TestFuncT", bound=Callable[..., Any])
+
 
 # We need a host that will not immediately close the connection with a TCP
 # Reset.
@@ -108,61 +110,46 @@ def setUp() -> None:
     warnings.simplefilter("ignore", HTTPWarning)
 
 
-def notWindows(test: Callable[..., _RT]) -> Callable[..., _RT]:
+def notWindows() -> Callable[[_TestFuncT], _TestFuncT]:
     """Skips this test on Windows"""
-
-    @functools.wraps(test)
-    def wrapper(*args: Any, **kwargs: Any) -> _RT:
-        msg = f"{test.__name__} does not run on Windows"
-        if platform.system() == "Windows":
-            pytest.skip(msg)
-        return test(*args, **kwargs)
-
-    return wrapper
+    return pytest.mark.skipif(
+        platform.system() == "Windows",
+        reason="Test does not run on Windows",
+    )
 
 
-def onlyBrotli() -> Callable[..., Any]:
+def onlyBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
     return pytest.mark.skipif(
         brotli is None, reason="only run if brotli library is present"
     )
 
 
-def notBrotli() -> Callable[..., Any]:
+def notBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
     return pytest.mark.skipif(
         brotli is not None, reason="only run if a brotli library is absent"
     )
 
 
-def onlySecureTransport(test: Callable[..., _RT]) -> Callable[..., _RT]:
+def onlySecureTransport() -> Callable[[_TestFuncT], _TestFuncT]:
     """Runs this test when SecureTransport is in use."""
-
-    @functools.wraps(test)
-    def wrapper(*args: Any, **kwargs: Any) -> _RT:
-        msg = f"{test.__name__} only runs with SecureTransport"
-        if not ssl_.IS_SECURETRANSPORT:
-            pytest.skip(msg)
-        return test(*args, **kwargs)
-
-    return wrapper
+    return pytest.mark.skipif(
+        not ssl_.IS_SECURETRANSPORT,
+        reason="Test only runs with SecureTransport",
+    )
 
 
-def notSecureTransport(test: Callable[..., _RT]) -> Callable[..., _RT]:
+def notSecureTransport() -> Callable[[_TestFuncT], _TestFuncT]:
     """Skips this test when SecureTransport is in use."""
-
-    @functools.wraps(test)
-    def wrapper(*args: Any, **kwargs: Any) -> _RT:
-        msg = f"{test.__name__} does not run with SecureTransport"
-        if ssl_.IS_SECURETRANSPORT:
-            pytest.skip(msg)
-        return test(*args, **kwargs)
-
-    return wrapper
+    return pytest.mark.skipif(
+        ssl_.IS_SECURETRANSPORT,
+        reason="Test does not run with SecureTransport",
+    )
 
 
 _requires_network_has_route = None
 
 
-def requires_network(test: Callable[..., _RT]) -> Callable[..., _RT]:
+def requires_network() -> Callable[[_TestFuncT], _TestFuncT]:
     """Helps you skip tests that require the network"""
 
     def _is_unreachable_err(err: Exception) -> bool:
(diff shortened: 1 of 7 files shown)


test/__init__.py
```diff
@@ -5,14 +5,16 @@
...
 import warnings
+from types import ModuleType
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
 
 import pytest
 
 try:
     try:
-        import brotlicffi as brotli
+        import brotlicffi as brotli  # type: ignore
```
bluetech, line 15: Suggestion (take it or leave it): When using `type: ignore`, I prefer to also specify the error code, e.g. `type: ignore[import]` (the error code is mentioned in the mypy error). This way, if the reason for the error changes, the ignore no longer works, which is good. It also helps the reader understand what is being ignored.

test/__init__.py
```diff
@@ -65,10 +71,10 @@ def _can_resolve(host):
         return False
 
 
-def has_alpn(ctx_cls=None):
+def has_alpn(ctx_cls: Optional[type] = None) -> bool:
```
bluetech, line 74: This can be `Optional[Type["ssl.SSLContext"]]` I believe.

test/__init__.py
```diff
@@ -84,25 +90,25 @@ def has_alpn(ctx_cls=None):
 RESOLVES_LOCALHOST_FQDN = _can_resolve("localhost.")
 
 
-def clear_warnings(cls=HTTPWarning):
+def clear_warnings(cls: type = HTTPWarning) -> None:
```
bluetech, line 93: Can be `Type[Warning]`

test/__init__.py
```diff
@@ -84,25 +90,25 @@ def has_alpn(ctx_cls=None):
...
     new_filters = []
-    for f in warnings.filters:
+    for f in warnings.filters:  # type: ignore
```
bluetech, line 95: If you feel like it, can send a patch to typeshed adding `filters` to `warnings.pyi` as an undocumented attribute.

test/__init__.py
```diff
@@ -84,25 +90,25 @@ def has_alpn(ctx_cls=None):
...
 
 
-def setUp():
+def setUp() -> None:
     clear_warnings()
     warnings.simplefilter("ignore", HTTPWarning)
 
 
-def notWindows(test):
+def notWindows(test: Callable[..., Any]) -> Callable[..., Any]:
```
bluetech, line 107: I think we want to make these decorators generic, so the type of callable they receive is the type of callable they return. Otherwise, the underlying test type becomes Any and (I think) isn't checked.

test/__init__.py
```diff
@@ -231,35 +239,37 @@ def wrapper(*args, **kwargs):
...
         self._target.removeHandler(self._handler)
 
-    def __enter__(self):
+    def __enter__(self) -> List[logging.LogRecord]:
         self.install()
         return self.records
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(
+        self, exc_type: object, exc_value: object, traceback: object
```
bluetech, line 271: If you want to be pedantic, this can be something like `exc_type: Type[BaseException], exc_value: BaseException, traceback: types.TracebackType`. Hmm or actually, with `Optional` around all of them...

test/port_helpers.py
```diff
@@ -108,5 +111,5 @@ def bind_port(sock, host=HOST):
             sock.setsockopt(socket.SOL_SOCKET, socket.SO_EXCLUSIVEADDRUSE, 1)
 
     sock.bind((host, 0))
-    port = sock.getsockname()[1]
+    port: int = sock.getsockname()[1]
```
bluetech, line 114: I suggest adding an `assert isinstance(port, int)` instead of this annotation, then if for some reason the type *isn't* int, it will be caught instead of silently being incorrect.

bluetech (commented):
Nice! I left some comments, LGTM otherwise.

test/__init__.py
hramezani: Good suggestion. I've updated all `type: ignore` in this path.
I will create a patch to update all other `type: ignore`.
Or maybe it can be a good contributor-friendly issue.
what do you think @sethmlarson

sethmlarson: It'd be great if we could have one person go through and commit all of these at once so probably not great for a new contributor. Could you handle these?

test/__init__.py
```diff
@@ -292,19 +309,22 @@ class ModuleStash:
...
+        self, namespace: str, modules: Dict[str, ModuleType] = sys.modules
+    ) -> None:
         self.namespace = namespace
         self.modules = modules
-        self._data = {}
+        self._data: Dict[str, ModuleType] = {}
 
-    def stash(self):
-        self._data[self.namespace] = self.modules.pop(self.namespace, None)
+    def stash(self) -> None:
```
sethmlarson, line 319: We're losing the safety of a KeyError here, what was causing the previous logic to fail?

sethmlarson (changes requested):
Thanks for this! LGTM with @bluetech comments applied. One additional comment I had too.

test/__init__.py
hramezani: Sure. I will create a PR for this

test/__init__.py
hramezani:
```
test/__init__.py:322: error: Argument 2 to "pop" of "MutableMapping" has incompatible type "None"; expected Module  [arg-type]
```
Shall I revert this and add `type: ignore`?

sethmlarson: Yeah I think we can add a type: ignore for this instance, thanks for clarifying :)

sethmlarson: Actually nevermind, this is fine as is! I was reviewing on mobile so didn't get full context.

hramezani:
@bluetech Thanks for your review.
I've addressed your comments.

test/__init__.py
```diff
@@ -111,23 +121,23 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def onlyBrotli():
+def onlyBrotli() -> Callable[..., Any]:
```
bluetech, line 124: Here I think it can be

```py
def onlyBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
```

(again, untested)

test/port_helpers.py
```diff
@@ -108,5 +111,6 @@ def bind_port(sock, host=HOST):
...
```
bluetech, line 114: So now the `: int` should not be necessary so I suggest removing it.

bluetech (commented):
I left some suggestions for the decorators, I'm not sure if they'd work, but worth a try.

test/port_helpers.py
hramezani: ah, sorry!

bluetech:
Hrr one of my comments got lost somehow. I suggested adding

```py
_TestFuncT = TypeVar("_TestFuncT", bound=Callable[..., Any])
```

and using it

```py
def notWindows(test: _TestFuncT) -> _TestFuncT:
```

this way the test type is preserved exactly, including kwargs, defaults, etc. that Callable can't currently express.

test/__init__.py
hramezani: I've tested this one. seems ok!

hramezani:
> Hrr one of my comments got lost somehow. I suggested adding
> 
> ```python
> _TestFuncT = TypeVar("_TestFuncT", bound=Callable[..., Any])
> ```
> 
> and using it
> 
> ```python
> def notWindows(test: _TestFuncT) -> _TestFuncT:
> ```
> 
> this way the test type is preserved exactly, including kwargs, defaults, etc. that Callable can't currently express.

What about `def wrapper` type annotation?
with `def wrapper(*args: Any, **kwargs: Any) -> Any:`, we will get 
```test/__init__.py:123: error: Incompatible return value type (got "Callable[[VarArg(Any), KwArg(Any)], Any]", expected "_TestFuncT")  [return-value]```

**Review by sethmlarson — APPROVED** [review]
LGTM pending @bluetech's comments

**Comment by bluetech** [comment]
> What about def wrapper type annotation?

I think it's fair to cast it to `_TestFuncT`.

I tried this:

```py
def notWindows(test: _TestFuncT) -> _TestFuncT:
    """Skips this test on Windows"""
    condition = platform.system() == "Windows"
    reason = f"{test.__name__} does not run on Windows"
    return pytest.mark.skipif(condition, reason=reason)(test)
```

unfortunately the last line becomes Any due to some pytest/mypy issue.

If we're willing to drop the `{test.__name__}` part from them message (which IMO is somewhat redundant, since pytest already shows the test name), then we can just make it like the others, that is

```py

def notWindows() -> Callable[[_TestFuncT], _TestFuncT]:
    """Skips this test on Windows"""
    condition = platform.system() == "Windows"
    reason = "Test does not run on Windows"
    return pytest.mark.skipif(condition, reason=reason)
```

hramezani:
@bluetech Thanks for your suggestion. I've updated the decorators with your suggestion.
But we still have a problem with mac os tests. I am investigating.
I didn't update `withPyOpenSSL` decorator because it's run `pyopenssl.inject_into_urllib3()` before test execution and `pyopenssl.extract_from_urllib3()` after.

bluetech:
>  But we still have a problem with mac os tests. I am investigating.

Right. Just from looking at the logs, I don't see what might be causing it. Let me know if you want me to take a deeper look.

hramezani:
> > But we still have a problem with mac os tests. I am investigating.
> 
> Right. Just from looking at the logs, I don't see what might be causing it. Let me know if you want me to take a deeper look.

Here is @sethmlarson comment:
> it's cause the skipif is evaluated immediately? Instead of evaluated at call time.

I've also proposed the problem in discord to other devs. it would be great if you can take a look @bluetech 
Thanks!

bluetech:
Ah, I see now how `ssl_.IS_SECURETRANSPORT` is mutated so the failure makes more sense.

Simplest way I can think to make it work is a little hack to make pytest evaluate the condition lazily:

```diff
diff --git a/test/__init__.py b/test/__init__.py
index 0dd71046..907c3a07 100644
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -6,7 +6,17 @@ import socket
 import sys
 import warnings
 from types import ModuleType, TracebackType
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Type, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Type,
+    TypeVar,
+    cast,
+)
 
 import pytest
 
@@ -133,7 +143,7 @@ def notBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
 def onlySecureTransport() -> Callable[[_TestFuncT], _TestFuncT]:
     """Runs this test when SecureTransport is in use."""
     return pytest.mark.skipif(
-        not ssl_.IS_SECURETRANSPORT,
+        lazy_condition(lambda: not ssl_.IS_SECURETRANSPORT),
         reason="Test only runs with SecureTransport",
     )
 
@@ -141,7 +151,7 @@ def onlySecureTransport() -> Callable[[_TestFuncT], _TestFuncT]:
 def notSecureTransport() -> Callable[[_TestFuncT], _TestFuncT]:
     """Skips this test when SecureTransport is in use."""
     return pytest.mark.skipif(
-        ssl_.IS_SECURETRANSPORT,
+        lazy_condition(lambda: ssl_.IS_SECURETRANSPORT),
         reason="Test does not run with SecureTransport",
     )
 
@@ -182,9 +192,18 @@ def requires_network() -> Callable[[_TestFuncT], _TestFuncT]:
     )
 
 
+# Hack to make pytest evaluate a condition at test runtime instead of collection time.
+def lazy_condition(condition: Callable[[], bool]) -> bool:
+    class LazyCondition:
+        def __bool__(self) -> bool:
+            return condition()
+
+    return cast(bool, LazyCondition())
+
+
 def requires_ssl_context_keyfile_password() -> Callable[[_TestFuncT], _TestFuncT]:
     return pytest.mark.skipif(
-        not ssl_.IS_SECURETRANSPORT,
+        lazy_condition(lambda: not ssl_.IS_SECURETRANSPORT),
         reason="Test requires password parameter for SSLContext.load_cert_chain()",
     )
 
```

Should probably add support for lambdas in pytest directly (I remember some discussion on that), but it's not currently supported. There are stringified conditions (which pytest `eval`s) but I really dislike those :)

bluetech (approved):

hramezani:
Thanks, @bluetech for your great patch.
@sethmlarson Could you please take a final look at this?

sethmlarson:
Taking a look now! :eyes:

sethmlarson (approved):
LGTM, this is great! :tada:

PR #2360, opened 2021-08-08
Make return type of {PoolManager,*ConnectionPool}.__enter__ more precise

bluetech:
Particularly for `ConnectionPool` subclasses, e.g. `HTTPConnnectionPool`, previously the static type of `pool` in

```py
with HTTPConnectionPool(..) as pool:
    ...
```

would be `ConnectionPool` instead of the more useful `HTTPConnectionPool` subtype.


hramezani (approved):
LGTM!

sethmlarson (approved):
LGTM

PR #2361, opened 2021-08-08
Add initial Towncrier newsfragments for 2.0 commits

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/2065 except for the "Lean on SSLContext" issue, I can't remember what the current state of support there is.

### DIFF
--- a/changelog/2082.feature.rst
+++ b/changelog/2082.feature.rst
@@ -1,1 +1,1 @@
-Changed default TLS ciphers being used to ciphers configured by the system for OpenSSL 1.1.1+ and SecureTransport.
+Stopped configuring default ciphers for OpenSSL 1.1.1+ and SecureTransport as their own defaults are already secure.
--- a/changelog/2082.feature.rst
+++ b/changelog/2082.feature.rst
@@ -1,0 +1,1 @@
+Changed default TLS ciphers being used to ciphers configured by the system for OpenSSL 1.1.1+ and SecureTransport.
--- a/changelog/2083.feature.rst
+++ b/changelog/2083.feature.rst
@@ -1,1 +1,1 @@
-Added ``urllib3.response.BaseHTTPResponse`` type.
+Added ``urllib3.response.BaseHTTPResponse`` class. All future response classes will be subclasses of ``BaseHTTPResponse``.
--- a/changelog/2083.feature.rst
+++ b/changelog/2083.feature.rst
@@ -1,0 +1,1 @@
+Added ``urllib3.response.BaseHTTPResponse`` type.
--- a/changelog/2099.feature.rst
+++ b/changelog/2099.feature.rst
@@ -1,3 +1,5 @@
 Changed ``urllib3[brotli]`` extra to install Google's Brotli C bindings on CPython.
 
-Continue to use Brotli CFFI bindings on non-CPython implementations.
+The Brotli CFFI bindings are still used on non-CPython implementations like PyPy.
+This change was made to increase performance for CPython as CFFI is typically
+is less performant than C extensions.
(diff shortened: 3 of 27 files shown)


changelog/2336.removal.rst
```diff
@@ -0,0 +1 @@
+Removed support for Python 3.6
```
pquentin, line 1: Should we merge this with the 2.7 removal?

pquentin (changes requested):
Thanks!

 * Is 2213 empty on purpose?
 * Should we add trailing dots at the end of our sentences? We've done that more recently, and https://keepachangelog.com/en/1.0.0/ does it too.

pquentin:
Also, looking at https://urllib3--2361.org.readthedocs.build/en/2361/changelog.html I realized that misc entries are hidden on purpose, which is not what we want.

changelog/2336.removal.rst
sethmlarson: Probably, although I wonder if it makes sense to keep the issues separate and then do this merge come release time so we capture each PR that contained the change?

sethmlarson:
> Is 2213 empty on purpose?

Woops! Good catch.
 
> Should we add trailing dots at the end of our sentences? We've done that more recently, and https://keepachangelog.com/en/1.0.0/ does it too.

Yeah let's do this to be consistent, good suggestion!

changelog/2336.removal.rst
pquentin: Sounds good, let's merge at release time

sethmlarson:
Alright, updated the newsfragments that were mentioned and moved everything that was `misc` to a different heading.

changelog/2082.feature.rst
```diff
@@ -0,0 +1 @@
+Changed default TLS ciphers being used to ciphers configured by the system for OpenSSL 1.1.1+ and SecureTransport.
```
pquentin, line 1: In this case, starting with "Changed" makes the changelog harder to understand, I had to go back to the PR to understand what it did.

In particular, the current version does not make clear that we're not defining new defaults, we're just letting OpenSSL 1.1.1 and SecureTransport do their thing. Also, I guess Linux distributions can change the OpenSSL 1.1.1 ciphers, which is why we're using the vague "system" term?

```suggestion
Stopped configuring default ciphers for OpenSSL 1.1.1+ and SecureTransport as their own defaults are already secure.
```

changelog/2113.removal.rst
```diff
@@ -0,0 +1 @@
+Removed fallback on certificate ``commonName`` in ``match_hostname()`` function.
```
pquentin, line 1: Since this is probably the most breaking change:

 * Can we mention that we're now looking at "Subject Alternative Name"?
 * Can we name drop RFC 2818 from May 2000?
 * Can it be disabled or not?

pquentin (changes requested):
Thanks a lot for this. Finally took the time to check everything! I left a few comments.

I would avoid closing #2065 until we have a check for missing changelog entries. This will also remind me that the SSLContext change should be documented.

changelog/2083.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``urllib3.response.BaseHTTPResponse`` type.
```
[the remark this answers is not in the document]
sethmlarson: Good idea!

changelog/2099.feature.rst
```diff
@@ -0,0 +1,3 @@
+Changed ``urllib3[brotli]`` extra to install Google's Brotli C bindings on CPython.
```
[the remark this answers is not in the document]
sethmlarson: Probably, good idea!

pquentin (approved):
Thanks!

PR #2362, opened 2021-08-09
Add type hints for noxfile.py

jdufresne:
Help ensure correct API usage and potential misuse as the library
evolves (none detected).


sethmlarson (approved):
Nice! Thank you, hadn't even thought to do that but it's a great idea :)

PR #2364, opened 2021-08-10
Add type hints to test.tz_stub

hramezani:

### DIFF
--- a/test/tz_stub.py
+++ b/test/tz_stub.py
@@ -2,13 +2,14 @@ import datetime
 import os
 import time
 from contextlib import contextmanager
+from typing import Generator, Optional
 
 import pytest
 from dateutil import tz
 
 
 @contextmanager
-def stub_timezone_ctx(tzname):
+def stub_timezone_ctx(tzname: Optional[str]) -> Generator[None, None, None]:
     """
     Switch to a locally-known timezone specified by `tzname`.
     On exit, restore the previous timezone.
@@ -32,6 +33,7 @@ def stub_timezone_ctx(tzname):
     if local_tz is None:
         raise OSError("Cannot determine current timezone")
     old_tzname = datetime.datetime.now(local_tz).tzname()
+    assert old_tzname is not None
 
     os.environ["TZ"] = tzname
     time.tzset()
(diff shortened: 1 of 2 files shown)


sethmlarson (dismissed):
LGTM

sethmlarson:
GitHub Actions are broken right now: https://www.githubstatus.com/

hramezani:
> GitHub Actions are broken right now: https://www.githubstatus.com/

Thanks for letting me know. I will try to run them again tomorrow!

hramezani:
@sethmlarson `mypy --install-types --non-interactive` failed with this error:
`error: Can't determine which types to install with no files to check (and no cache from previous mypy run)`
So, I've installed `types-python-dateutil` based on `mypy` suggestion.
```
test/tz_stub.py:8: note: Hint: "python3 -m pip install types-python-dateutil"
test/tz_stub.py:8: note: (or run "mypy --install-types" to install all missing stub packages)
```
Please take another look.
Thanks!

test/tz_stub.py
```diff
@@ -36,5 +37,5 @@ def stub_timezone_ctx(tzname):
     os.environ["TZ"] = tzname
     time.tzset()
     yield
-    os.environ["TZ"] = old_tzname
+    os.environ["TZ"] = old_tzname  # type: ignore[assignment]
```
bluetech, line 40: What is the type of `old_tzname` here?

bluetech:
I wonder if we can use the standard `zoneinfo` module instead of `dateutil`. The dateutil usage here seems kinda optional, like some sanity test IIUC. So we can either just remove it for py<3.9 or use `backports.zoneinfo` for older versions.

test/tz_stub.py
hramezani: here is the error message from mypy:
`Incompatible types in assignment (expression has type "Optional[str]", target has type "str")  [assignment]`

The `old_tzname` comes from `old_tzname = datetime.datetime.now(local_tz).tzname()`.
`tzname()` return type is `Optional[str]`

bluetech: Ha got it. I just checked, if we assign `None` to `os.environ` it raises a `TypeError`. So I think what we should do is `assert old_tzname is not None` right after obtaining it, to clarify our assumption and avoid the type-ignore.

hramezani: Good suggestion. Thanks!
I've updated the patch based on your suggestion.

bluetech (approved):

Issue #2103, opened 2020-12-08
Make sure README looks good in GitHub dark mode


pquentin:
GitHub just added dark mode support in its app and web site. We had this issue in the docs but could use fancy CSS there. In the case of GitHub I guess the only option is to add a white background to the logo?

ZurMaD:
Not important but can be better

![image](https://user-images.githubusercontent.com/28235457/103101261-b0713000-45e4-11eb-9fd7-fc34734bfbcb.png)

boyuan12:
Can I try to work on this issue?

sethmlarson:
@boyuan12 You certainly can!

chetanugale:
@boyuan12 @sethmlarson did we fix this issue or can I work on it?

sethmlarson:
@chetanugale I believe this issue is available for you to work on :)

chetanugale:
thanks @sethmlarson 👍

PR #2366, opened 2021-08-12
Make sure README looks good in GitHub dark mode

chetanugale:
Closes https://github.com/urllib3/urllib3/issues/2103

Light mode:
light_theme
Dark mode:
dark_theme


pquentin:
Thanks for this! It's the only thing we can do to improve the situation. However, I'm not sure what to think. In "Dark dimmed" mode, the old logo is not that bad. But hard to read in "Default dark". The new logo is readable in all settings, but hard on the eyes.

What do others think?

(Note that we'll have to adjust details like the border of the name of the file if this gets accepted, but let's not discuss those details right now.)

sethmlarson:
Yeah I think a thin white border around the lettering+symbols would be enough here and probably would look more deliberate as well :)

pquentin:
This looks great! Do we need to make the image slightly bigger so that the bottom lettering of the b isn't cut off?

sethmlarson:
Agree with Quentin about the tiny bit of cutoff. Looks fantastic!! 🤩

sethmlarson:
Screenshot from 2021-08-12 07-59-06

That looks really really good, thank you!

sethmlarson:
@chetanugale Not to be too nitpicky but can we increase the background a tiny bit more? I think there's a little cutoff on the right edge of the 3 but after that it's perfect :muscle:

sethmlarson:
Now there's some cutoff on the left, if you need to we can increase the size of the canvas beyond 400 :)

sethmlarson:
@chetanugale As an aside I'm planning on tweeting this out once it's merged, do you have a Twitter handle I can tag to credit?

pquentin:
Should we call this file `banner_github.svg`? That way we don't touch the existing dark/white logo in the docs.

Or should we use this in the docs too?

chetanugale:
looks good?
Screenshot 2021-08-12 at 6 38 57 PM

sethmlarson:
Yeah we should leave the banner used by the docs the same, this can be `banner-readme.svg` or `banner-github.svg`

sethmlarson:
@chetanugale I think the `u` is still barely cutoff by 1 pixel on the left.

sethmlarson (approved):
LGTM, thank you so much for this!

chetanugale:
> @chetanugale As an aside I'm planning on tweeting this out once it's merged, do you have a Twitter handle I can tag to credit?

**@duratan_it_is**

PR #2367, opened 2021-08-12
Add type hints to test.conftest

hramezani:


test/conftest.py
```diff
@@ -90,16 +102,16 @@ def ipv6_san_server(tmp_path_factory):
 
 
 @pytest.fixture
-def stub_timezone(request):
+def stub_timezone(request: pytest.FixtureRequest) -> Generator[None, None, None]:
     """
     A pytest fixture that runs the test with a stub timezone.
     """
-    with stub_timezone_ctx(request.param):
+    with stub_timezone_ctx(request.param):  # type: ignore[attr-defined]
```
hramezani, line 109: `request` type should be `SubRequest` which is in `_pytest.fixtures.SubRequest` and does not exposed in `pytest`.
@bluetech Should I create a PR and add `SubRequest` in https://github.com/pytest-dev/pytest/blob/main/src/pytest/__init__.py?

bluetech: We have an existing issue for this: https://github.com/pytest-dev/pytest/issues/8763

For now I think what you did is best :+1:

test/conftest.py
```diff
@@ -90,16 +102,16 @@ def ipv6_san_server(tmp_path_factory):
...
         yield
 
 
 @pytest.fixture(scope="session")
-def supported_tls_versions():
+def supported_tls_versions() -> Set[Optional[str]]:
```
bluetech, line 114: Given this is a shared session fixture, it will be safer to use `AbstractSet` instead of `Set`, to ensure no user mutates it, affecting all other users.

Also, why is the `Optional` needed? (can `_sock.version()` be None?)

bluetech (approved):
LGTM, with one suggestion.

test/conftest.py
hramezani: > Given this is a shared session fixture, it will be safer to use AbstractSet instead of Set, to ensure no user mutates it, affecting all other users.

Good to know. Thanks!

> why is the Optional needed? (can _sock.version() be None?)

Yes, `_sock.version()` returns `Optional[str])`

bluetech (approved):

PR #2369, opened 2021-08-12
Remove leftover type stubs

pquentin:
All types are now inline.


sethmlarson (approved):
LGTM, good catch here!

PR #2370, opened 2021-08-12
Install types-pyOpenSSL in mypy nox session

hramezani:


bluetech:
A bit of an issue with this is https://github.com/python/typeshed/pull/5847 -- namely, `types-pyopenssl` depends on `types-cryptography` which seems to override the actual `cryptography` type stubs, as seen here in the additional `type: ignore` on `cryptography` types. Not sure what's the best thing to do, maybe better to just keep the status quo until this is sorted out hopefully soon.

sethmlarson:
Yeah I agree with @bluetech here, we'll wait for typeshed to figure out what it's going to do before we go down a specific road.

PR #2372, opened 2021-08-13
Improve request debug log readability

pktiuk:
Hello,  
I think it would greatly improve readability of debug log printed after making request.

This log in current form may be not readable for users of this library  
```
DEBUG:urllib3.connectionpool:https://mywebsite.com:443 "GET /configuration HTTP/1.1" 200 1405
```
It is not obvious what these numbers at the end mean (at least for users not familiar with this library)

I think something like this would be better
```
DEBUG:urllib3.connectionpool:https://mywebsite.com:443/configuration "GET HTTP/1.1" status:200 len:1405
```

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -450,12 +450,12 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             raise
 
         log.debug(
-            '%s://%s:%s "%s %s %s" %s %s',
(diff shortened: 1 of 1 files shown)


pquentin:
For what it's worth, we're following the Common Log Format here: https://en.wikipedia.org/wiki/Common_Log_Format. This is not specific to urllib3.

pktiuk:
That explains a lot.  
If this is the case, then it is better to leave it as it is.

PR #2374, opened 2021-08-16
Add type hints to test.test_wait

hramezani:

### DIFF
--- a/test/test_wait.py
+++ b/test/test_wait.py
@@ -1,7 +1,9 @@
 import signal
 import threading
 import time
-from socket import socketpair
+from socket import socket, socketpair
+from types import FrameType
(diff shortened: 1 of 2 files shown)


test/test_wait.py
```diff
@@ -14,22 +16,25 @@
     wait_for_write,
 )
 
+_TYPE_SOCKET_PAIR = Tuple[socket, socket]
```
bluetech, line 19: Take it or leave it: since tests don't need to worry about privacy, the `_` is not really necessary.

bluetech (approved):
LGTM!

sethmlarson (approved):
LGTM, thank you!

PR #2375, opened 2021-08-17
Add type hints to test.test_retry

hramezani:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -176,6 +176,7 @@ def mypy(session: nox.Session) -> None:
         "test/__init__.py",
         "test/conftest.py",
         "test/port_helpers.py",
+        "test/test_retry.py",
         "test/test_wait.py",
         "test/tz_stub.py",
     )
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -1,7 +1,9 @@
+from typing import Optional
 from unittest import mock
 
 import pytest
 
+from urllib3.connectionpool import ConnectionPool
 from urllib3.exceptions import (
     ConnectTimeoutError,
     InvalidHeader,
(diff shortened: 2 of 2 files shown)


src/urllib3/exceptions.py
```diff
@@ -32,9 +32,9 @@ class HTTPWarning(Warning):
 class PoolError(HTTPError):
     """Base exception for errors caused within a pool."""
 
-    pool: "ConnectionPool"
+    pool: Optional["ConnectionPool"]
```
bluetech, line 35: Preexisting: this line looks redundant to me, unless I'm missing something.

src/urllib3/exceptions.py
```diff
@@ -32,9 +32,9 @@ class HTTPWarning(Warning):
...
 
-    def __init__(self, pool: "ConnectionPool", message: str) -> None:
+    def __init__(self, pool: Optional["ConnectionPool"], message: str) -> None:
```
bluetech, line 37: Given that we only need to change this due to the tests, I infer that the actual code never passes `None` here, so it's better to fix the tests instead of pessimizing the type for all users.

Maybe adding 

```py
DUMMY_POOL = ConnectionPool("dummy")
```

to the test file and passing it instead of `None`?

bluetech (commented):
LGTM, except one point.

sethmlarson (commented):
LGTM, I agree with @bluetech on the `pool` parameter.

src/urllib3/exceptions.py
hramezani: hmmm, @sethmlarson Could you please elaborate on it?

src/urllib3/exceptions.py
hramezani: Great suggestion. Thanks!

sethmlarson: This was the suggestion I wanted implemented :)

sethmlarson (approved):

PR #2376, opened 2021-08-17
Remove redundant attrs from exception classes

hramezani:
Based on @bluetech comment https://github.com/urllib3/urllib3/pull/2375#discussion_r690249917


src/urllib3/exceptions.py
```diff
@@ -329,8 +312,6 @@ class InvalidChunkLength(HTTPError, httplib_IncompleteRead):
 
     partial: int
```
sethmlarson, line 313: Are these needed? We can add `self.partial: int = response.tell()` for example?

hramezani: You are right. Added!

bluetech (approved):

PR #2379, opened 2021-08-18
Collect mypy requirements in their own file

### LINKED ISSUE #2377: Collect mypy requirements in their own file
We currently install mypy dependencies one by one:

https://github.com/urllib3/urllib3/blob/0a4839af7103e9ad71d26bf5ba7914e765577efe/noxfile.py#L162-L169

It was fine initially, but we currently install 8 dependencies that way. That's slower than necessary and isn't a format tools like dependabot can understand.

Instead, put those dependencies in a file called `mypy-requirements.txt` and install them just like we install dev-requirements.txt:

https://github.com/urllib3/urllib3/blob/0a4839af7103e9ad71d26bf5ba7914e765577efe/noxfile.py#L23

V1NAY8:
Closes #2377

### DIFF
--- a/mypy-requirements.txt
+++ b/mypy-requirements.txt
@@ -1,0 +1,8 @@
+mypy==0.910
+idna>=2.0.0
+cryptography>=1.3.4
+tornado>=6.1
+pytest>=6.2
+trustme==0.9.0
+types-python-dateutil
+nox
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks!

hramezani (approved):

PR #2380, opened 2021-08-18
Fix type hints import in urllib3.util.url

hramezani:
To make the import compatible with other files, I looked in the code for `typing.*` and fix them.


sethmlarson (approved):
LGTM

Issue #2368, opened 2021-08-12
Handle `https:///`, `https:/` and `https:\` as `https://`


pquentin:
### Subject

parse_url in urllib3 mishandles certain uses of extra slash or backslash (such as `https:///`, `https:/`, `https:\`) and interprets the URI as a relative path. Browsers accept extra slash or backslash after the protocol, and treat it as a normal slash, while parse_url sees it as a relative path.

Full disclosure: this was originally sent to us as a security vulnerability, but as the report shows, with those URLs we don't detect any hosts. That means that any request with those URLs will fail with "urllib3.exceptions.LocationValueError: No host specified.". So no SSRF is possible.

### Steps to Reproduce

```
import urllib3
url1=urllib3.util.parse_url('https:///www.attacker.com/a/b')
print("HOSTNAME:  " +str(url1.host))
print("PATH    :  " +str(url1.path))
url2=urllib3.util.parse_url('https:/www.attacker.com/a/b')
print("HOSTNAME:  " +str(url2.host))
print("PATH    :  " +str(url2.path))
url3=urllib3.util.parse_url('https:\www.attacker.com/a/b')
print("HOSTNAME:  " +str(url3.host))
print("PATH    :  " +str(url3.path))
```

### Expected Behavior

Do what browsers do: all those URLs should be treated as `https://www.attacker.com/a/b`.

### Actual Behavior

```
HOSTNAME:  None
PATH    :  /www.attacker.com/a/b
HOSTNAME:  None
PATH    :  /www.attacker.com/a/b
HOSTNAME:  None
PATH    :  /%5Cwww.attacker.com/a/b
```

chetanugale:
@pquentin can I pick this issue?

pquentin:
Sure!

timmc:
While some browsers do "fix up" URLs in that way, that is contrary to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), which makes no such allowance. Adding such variations increases the chance of a parser mismatch vulnerability as seen in Claroty's excellent [paper on URL parser confusion](https://claroty.com/wp-content/uploads/2022/01/Exploiting-URL-Parsing-Confusion.pdf).

The WHATWG guidance on URLs does suggest treating backslashes as slashes, but it is driven by observed usage in browsers, and is subject to change. As such, it is not a stable standard that can be targeted for correctness and compliance.

[...]

PR #2384, opened 2021-08-20
Handle https:///, https:/ and https:\ as https:// 

chetanugale:
Issue : https://github.com/urllib3/urllib3/issues/2368

Tested with :
url1=urllib3.util.parse_url('https:///www.attacker.com/a/b')
print("HOSTNAME:  " +str(url1.host))
print("PATH    :  " +str(url1.path))
url2=urllib3.util.parse_url('https:/www.attacker.com/a/b')
print("HOSTNAME:  " +str(url2.host))
print("PATH    :  " +str(url2.path))
url3=urllib3.util.parse_url('https:\www.attacker.com/a/b')
print("HOSTNAME:  " +str(url3.host))
print("PATH    :  " +str(url3.path))

response:

HOSTNAME:  www.attacker.com
PATH    :  /a/b
HOSTNAME:  www.attacker.com
PATH    :  /a/b
HOSTNAME:  www.attacker.com
PATH    :  /a/b


pquentin:
Regardless of the code here, it's interesting that the requests test suite is using `http:/badproxyurl:3128` as an URL that should fail to parse. We'll have to submit a fix there to use a different URL!

chetanugale:
@pquentin @SethMichaelLarson any inputs for this PR ?

pquentin:
@chetanugale We're sorry that we haven't reviewed this yet. We know that we should, but URL parsing changes are always scary as we had many vulnerabilities here in the past, and it's easy to make subtle mistakes. Unfortunately I can't promise any timeline for a review here. :(

timmc:
I don't think this change should be accepted -- see comments on issue: https://github.com/urllib3/urllib3/issues/2368#issuecomment-1086726748

sethmlarson:
I'm not confident enough to make a call that this is the right thing to do, so I'm going to decline it for now. Sorry it took so long to reach this conclusion. See https://github.com/urllib3/urllib3/issues/2368#issuecomment-1086755912 for more info.

Issue #2386, opened 2021-08-22
Fix socks.py coverage drop


pquentin:
We use codecov to make sure we stay at 100% coverage, which is incredibly valuable to catch bugs and maintain high code quality. This means that each line needs to be covered by at least one run. And this is no longer the case for this line:

https://github.com/urllib3/urllib3/blob/25ed51f319a3c6920bb5bd959924ca80925ee120/src/urllib3/contrib/socks.py#L144-L146

This looks like defensive code and a comment above says "This is fragile as hell", but it would be nice to understand what happened before using a "Defensive" coverage pragma.

Here's a [good report](https://codecov.io/gh/urllib3/urllib3/src/0267e766a49f0de76b3e0b5b2dac6f8f29492b67/src/urllib3/contrib/socks.py) happened on August 13th:

<img width="907" alt="Capture d’écran 2021-08-22 à 21 10 54" src="https://user-images.githubusercontent.com/42327/130363924-8529f013-ad83-4144-b0fe-e95024844e95.png">

We can see that 4 builds went through that line.

The [first bad report](https://codecov.io/gh/urllib3/urllib3/src/0a4839af7103e9ad71d26bf5ba7914e765577efe/src/urllib3/contrib/socks.py) happened on August 15th:

<img width="878" alt="Capture d’écran 2021-08-22 à 21 14 14" src="https://user-images.githubusercontent.com/42327/130364014-6b48678b-73c8-43e5-9dc2-28ce6f35f1ce.png">

Here's the diff between those commits: https://github.com/urllib3/urllib3/compare/c674f48d59affef387d8f2ea817e1ae4d696e4da...0a4839af7103e9ad71d26bf5ba7914e765577efe

To be continued

bluetech:
I figured we need to find the test which is supposed to cover this line. So I pulled your revert PR #2388 which reports 100% coverage and added a `breakpoint()` above the uncovered line, but it didn't trigger. So I'm not sure how it's covered - maybe some test that's skipped here.

If you don't mind, I'll use your PR to try to figure it out, namely I'll add some `raise BaseException('BOOM')` and see which test fails.

pquentin:
I don't mind, otherwise I wouldn't allow other maintainers to modify my PRs. :)

That sounds like a great plan!

bluetech:
That didn't work either - 100% and all CI tests green. I will look at this more tomorrow.

bluetech:
[...]

PR #2388, opened 2021-08-22
Revert socks.py exception chaining

pquentin:
Related to #2386

This should not be merged as is, but this pull request restores the 100% coverage. Now someone will be able to understand what's going on.


bluetech (approved):
As discussed in the issue I think we should go ahead with this (removing `from e` from this one branch).

I left the bisection commits for reference, but when merging the PR should probably be squashed.

sethmlarson (approved):
Going to go ahead and merge this, thanks @pquentin and @bluetech for this exploration!

PR #2389, opened 2021-08-23
Add type hints to test.test_poolmanager

hramezani:

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -353,7 +353,7 @@ class PoolManager(RequestMethods):
         return pool
 
     def connection_from_url(
-        self, url: str, pool_kwargs: Optional[Dict[str, Any]] = None
+        self, url: Optional[str], pool_kwargs: Optional[Dict[str, Any]] = None
     ) -> HTTPConnectionPool:
         """
         Similar to :func:`urllib3.connectionpool.connection_from_url`.
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -353,7 +353,7 @@ class PoolManager(RequestMethods):
         return pool
 
     def connection_from_url(
-        self, url: Optional[str], pool_kwargs: Optional[Dict[str, Any]] = None
+        self, url: str, pool_kwargs: Optional[Dict[str, Any]] = None
     ) -> HTTPConnectionPool:
         """
         Similar to :func:`urllib3.connectionpool.connection_from_url`.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -185,6 +185,9 @@ class Retry:
     #: Maximum backoff time.
     BACKOFF_MAX = 120
 
+    # Backward compatibility; assigned outside of the class.
+    DEFAULT: "Retry"
+
     def __init__(
         self,
         total: Optional[Union[bool, int]] = 10,
@@ -256,7 +259,7 @@ class Retry:
     ) -> "Retry":
         """ Backwards-compatibility for the old retries format."""
         if retries is None:
-            retries = default if default is not None else cls.DEFAULT  # type: ignore[attr-defined]
+            retries = default if default is not None else cls.DEFAULT
 
         if isinstance(retries, Retry):
             return retries
@@ -508,4 +511,4 @@ class Retry:
 
 
 # For backwards compatibility (equivalent to pre-v1.9):
-Retry.DEFAULT = Retry(3)  # type: ignore[attr-defined]
+Retry.DEFAULT = Retry(3)
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -4,7 +4,16 @@ import re
 import time
 from itertools import takewhile
 from types import TracebackType
-from typing import TYPE_CHECKING, Any, Collection, NamedTuple, Optional, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    ClassVar,
+    Collection,
+    NamedTuple,
+    Optional,
+    Tuple,
+    Union,
+)
 
 from ..exceptions import (
     ConnectTimeoutError,
@@ -186,7 +195,7 @@ class Retry:
     BACKOFF_MAX = 120
 
     # Backward compatibility; assigned outside of the class.
-    DEFAULT: "Retry"
+    DEFAULT: ClassVar["Retry"]
 
     def __init__(
         self,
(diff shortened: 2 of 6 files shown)


test/test_poolmanager.py
```diff
@@ -247,16 +250,16 @@ def test_http_connection_from_host_case_insensitive(self):
...
-    def test_assert_hostname_and_fingerprint_flag(self):
+    def test_assert_hostname_and_fingerprint_flag(self) -> None:
         """Assert that pool manager can accept hostname and fingerprint flags."""
         fingerprint = "92:81:FE:85:F7:0C:26:60:EC:D6:B3:BF:93:CF:F9:71:CC:07:7D:0A"
         p = PoolManager(assert_hostname=True, assert_fingerprint=fingerprint)
         pool = p.connection_from_url("https://example.com/")
         assert 1 == len(p.pools)
-        assert pool.assert_hostname
-        assert fingerprint == pool.assert_fingerprint
+        assert pool.assert_hostname  # type: ignore[attr-defined]
```
bluetech, line 259: Why does mypy not know about these attributes?

test/test_poolmanager.py
```diff
@@ -322,13 +329,13 @@ def test_override_pool_kwargs_url(self):
             "http://example.com/", pool_kwargs=pool_kwargs
         )
 
-        assert retry.Retry.DEFAULT == default_pool.retries
+        assert retry.Retry.DEFAULT == default_pool.retries  # type: ignore[attr-defined]
```
bluetech, line 332: Why does mypy not know about this attribute?

src/urllib3/poolmanager.py
```diff
@@ -353,7 +353,7 @@ def connection_from_pool_key(
         return pool
 
     def connection_from_url(
-        self, url: str, pool_kwargs: Optional[Dict[str, Any]] = None
+        self, url: Optional[str], pool_kwargs: Optional[Dict[str, Any]] = None
```
bluetech, line 356: Does `connection_from_host` has any chance of succeeding with an empty URL? If not, then it seems better to prevent None here.

test/test_poolmanager.py
hramezani: Here it's defined in `util.retry` for backwards compatibility like this:
https://github.com/urllib3/urllib3/blob/d8eabf304190c62766e05c79ef44be7ff94cfbcb/src/urllib3/util/retry.py#L511
and we ignored type check for it there.

src/urllib3/poolmanager.py
hramezani: With empty `Url`, the `host` attr will be `None`. by passing `None` to `connection_from_host`, it will raise `LocationValueError`
https://github.com/urllib3/urllib3/blob/d8eabf304190c62766e05c79ef44be7ff94cfbcb/src/urllib3/poolmanager.py#L293

Also, `connection_from_host` accepts `None`:
https://github.com/urllib3/urllib3/blob/d8eabf304190c62766e05c79ef44be7ff94cfbcb/src/urllib3/poolmanager.py#L277

test/test_poolmanager.py
bluetech: Ah got it. It seems like maybe something we can drop for 2.0? The comment says "For backwards compatibility (equivalent to pre-v1.9)" which was a long time ago...

If not, I think we can still add a type annotation for it like

```py
class Retry:
    # Backward compatibility; assigned outside of the class.
    DEFAULT: "Retry"
```

but the ignore is also fine.

hramezani: @sethmlarson what do you think? can we drop it?

src/urllib3/poolmanager.py
bluetech: Got it. The Optional looks fine to me then, but the question I'm asking myself is which type would be more useful here? I don't think anyone wants to pass a special value (None) which will immediately raise -- presumably anyone who is using typing would prefer mypy to warn them about it already instead of raising at runtime.

If what I'm saying makes sense, the `if None raise LocationValueError(...)` check can still remain, but only as a runtime sanity check (and will probably require ignoring an `unreachable` mypy error *inside* the function, but that's an expected ignore).

hramezani: Yeah, I think we can revert this change and ignore mypy check it test.

test/test_poolmanager.py
hramezani: `assert_hostname` and `assert_fingerprint` are attributes of `connectionpool.HTTPSConnectionPool`
https://github.com/urllib3/urllib3/blob/d8eabf304190c62766e05c79ef44be7ff94cfbcb/src/urllib3/connectionpool.py#L873
but we return `HTTPSConnectionPool` from `poolmanager.PoolManager.connection_from_url`
https://github.com/urllib3/urllib3/blob/d8eabf304190c62766e05c79ef44be7ff94cfbcb/src/urllib3/poolmanager.py#L355

We have to change the return type to `HTTPSConnectionPool` for `connection_from_url` and other related functions in `poolmanager.py`

bluetech: Got it

> We have to change the return type to HTTPSConnectionPool for connection_from_url and other related functions in poolmanager.py

I don't think we can do it since it depends on the scheme which is a runtime value.

I think the only thing to improve here is to replace the ignores with an `assert isinstance(pool, HTTPSConnectionPool)`, that would make things clear.

bluetech (dismissed):
LGTM! There's one comment pending but I'm fine with whatever you decide there.

test/test_poolmanager.py
sethmlarson: I think we should keep `Retry.DEFAULT` and add the type annotation+comment that @bluetech suggested. Thanks for discussing this!

hramezani: Thanks, @bluetech for your suggestion. it removed 5 type ignore from our codebase.

src/urllib3/util/retry.py
```diff
@@ -185,6 +185,9 @@ class Retry:
     #: Maximum backoff time.
     BACKOFF_MAX = 120
 
+    # Backward compatibility; assigned outside of the class.
+    DEFAULT: "Retry"
```
bluetech, line 189: It just occurred to me, that this actually needs to be `DEFAULT: ClassVar["Retry"]`, since it's defined once on the class, not on every instance. I forgot about it, sorry!

hramezani: Done!

src/urllib3/util/retry.py
```diff
@@ -185,6 +194,9 @@ class Retry:
     #: Maximum backoff time.
     BACKOFF_MAX = 120
 
+    # Backward compatibility; assigned outside of the class.
+    DEFAULT: ClassVar["Retry"]
```
sethmlarson, line 198: Nice call with `ClassVar`!

sethmlarson (approved):
LGTM

PR #2393, opened 2021-08-27
feat: make backoff_max configurable with default of 120

barrett-schonefeld:
I work in the IBM Cloud Developer Experience group, and we use this package in our Python SDK "core" library (an enabling layer that is a dependency of our various IBM Cloud Python SDKs).

This message is adapted from an issue we posted in a package we use as a dependency for our _**Node**_ SDK "core" library. As a reference, the message is adapted from [this issue](https://github.com/JustinBeckwith/retry-axios/issues/164).

We want to use the exponential backoff policy, but also be able to cap the backoff time by some user-configurable amount. It looks like the package supports a default `BACKOFF_TIME` of 120. I would like to utilize this `BACKOFF_TIME` of 120 as the default value and allow users to configure this value as well.

This contribution is parallel to the contribution we made to the Node package in [this PR](https://github.com/JustinBeckwith/retry-axios/pull/165).

### DIFF
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -153,7 +153,7 @@ class Retry:
 
         seconds. If the backoff_factor is 0.1, then :func:`.sleep` will sleep
         for [0.0s, 0.2s, 0.4s, ...] between retries. It will never be longer
-        than :attr:`Retry.BACKOFF_MAX`.
+        than :attr:`Retry.backoff_max`.
 
         By default, backoff is disabled (set to 0).
 
@@ -191,8 +191,8 @@ class Retry:
     #: Default headers to be used for ``remove_headers_on_redirect``
     DEFAULT_REMOVE_HEADERS_ON_REDIRECT = frozenset(["Authorization"])
 
-    #: Maximum backoff time.
-    BACKOFF_MAX = 120
+    #: Default maximum backoff time.
+    DEFAULT_BACKOFF_MAX = 120
 
     # Backward compatibility; assigned outside of the class.
     DEFAULT: ClassVar["Retry"]
@@ -208,6 +208,7 @@ class Retry:
         allowed_methods: Optional[Collection[str]] = DEFAULT_ALLOWED_METHODS,
         status_forcelist: Optional[Collection[int]] = None,
         backoff_factor: float = 0,
+        backoff_max: float = DEFAULT_BACKOFF_MAX,
         raise_on_redirect: bool = True,
         raise_on_status: bool = True,
         history: Optional[Tuple[RequestHistory, ...]] = None,
@@ -230,6 +231,7 @@ class Retry:
         self.status_forcelist = status_forcelist or set()
         self.allowed_methods = allowed_methods
         self.backoff_factor = backoff_factor
+        self.backoff_max = backoff_max
         self.raise_on_redirect = raise_on_redirect
         self.raise_on_status = raise_on_status
         self.history = history or ()
@@ -249,6 +251,7 @@ class Retry:
             allowed_methods=self.allowed_methods,
             status_forcelist=self.status_forcelist,
             backoff_factor=self.backoff_factor,
+            backoff_max=self.backoff_max,
             raise_on_redirect=self.raise_on_redirect,
             raise_on_status=self.raise_on_status,
             history=self.history,
@@ -293,7 +296,7 @@ class Retry:
             return 0
 
         backoff_value = self.backoff_factor * (2 ** (consecutive_errors_len - 1))
-        return float(min(self.BACKOFF_MAX, backoff_value))
+        return float(min(self.backoff_max, backoff_value))
 
     def parse_retry_after(self, retry_after: str) -> float:
         seconds: float
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -176,6 +176,9 @@ class TestRetry:
         retry = retry.increment(method="GET")
         assert retry.get_backoff_time() == max_backoff
 
+        retry = retry.increment(method="GET")
+        assert retry.get_backoff_time() == max_backoff
+
     def test_zero_backoff(self) -> None:
         retry = Retry()
         assert retry.get_backoff_time() == 0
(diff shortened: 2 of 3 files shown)


src/urllib3/util/retry.py
```diff
@@ -191,8 +191,8 @@ class Retry:
     #: Default headers to be used for ``remove_headers_on_redirect``
     DEFAULT_REMOVE_HEADERS_ON_REDIRECT = frozenset(["Authorization"])
 
-    #: Maximum backoff time.
-    BACKOFF_MAX = 120
+    #: Default maximum backoff time.
+    DEFAULT_BACKOFF_MAX = 120
```
sethmlarson, line 195: This change makes sense, however I think we'd need a deprecation cycle in v1.26.x to remove `BACKOFF_MAX` safely in v2.0. Perhaps we can do that since it's likely not used often.

In 1.26.x I'm thinking we can go the route of other deprecated Retry classvars: https://github.com/urllib3/urllib3/pull/2000

src/urllib3/util/retry.py
```diff
@@ -230,6 +231,7 @@ def __init__(
         self.status_forcelist = status_forcelist or set()
         self.allowed_methods = allowed_methods
         self.backoff_factor = backoff_factor
+        self.backoff_max = backoff_max
```
sethmlarson, line 234: Need to also add this to `__new__()`

src/urllib3/util/retry.py
```diff
@@ -208,6 +208,7 @@ def __init__(
         allowed_methods: Optional[Collection[str]] = DEFAULT_ALLOWED_METHODS,
         status_forcelist: Optional[Collection[int]] = None,
         backoff_factor: float = 0,
+        backoff_max: float = self.DEFAULT_BACKOFF_MAX,
```
sethmlarson, line 211: Should be `DEFAULT_BACKOFF_MAX` without `self`

sethmlarson (changes requested):
I think this makes sense to add. Having class-level setting be the only method to configure this isn't great because it impacts all usages of urllib3. The current CI looks like it's failing along with `lint`, you should run `nox -s format` to make sure your code is formatted properly and doesn't have type errors.

src/urllib3/util/retry.py
barrett-schonefeld: Ok, great!

To be sure I understand the process, I will open a PR to deprecate `BACKOFF_MAX`. The deprecation change will be included in `v1.26.x`.

Then this PR (including `DEFAULT_BACKOFF_MAX`) will be introduced in `2.0.0`.

src/urllib3/util/retry.py
barrett-schonefeld: Fixed!

src/urllib3/util/retry.py
barrett-schonefeld: Fixed!

barrett-schonefeld:
Hey, folks, checking in on this PR!

How might I help get this work ready to be merged?

Also, would it be incorrect to leave `BACKOFF_MAX` as `BACKOFF_MAX` instead of changing this to `DEFAULT_BACKOFF_MAX`, so that we may more quickly release this update to make the backoff max configurable?

sethmlarson:
@barrett-schonefeld Thanks for the patience here, to answer your questions:

> Would it be incorrect to leave BACKOFF_MAX as BACKOFF_MAX instead of changing this to DEFAULT_BACKOFF_MAX

I'm fine with keeping this value as `BACKOFF_MAX` but since it's going in v2.0 it might be good to get the naming right.

> ...so that we may more quickly release this update to make the backoff max configurable?

Unfortunately the 1.x release stream isn't receiving releases including new non-critical features. We're focusing completely on v2.0 so this likely won't be released until then. We don't have a date we're targetting for v2.0 as there [still some large TODOs on our plate](https://github.com/urllib3/urllib3/issues?q=is%3Aopen+is%3Aissue+milestone%3Av2.0+label%3A%22Breaking+Change%22).

test/test_retry.py
```diff
@@ -155,6 +155,27 @@ def test_backoff(self) -> None:
...
+        retry = retry.increment(method="GET")
+        assert retry.backoff_factor == 0.2
+        assert retry.total == 98
+        assert retry.get_backoff_time() == 0.4  # Start backoff
+
+        retry = retry.increment(method="GET")
+        assert retry.get_backoff_time() == 0.8
+
+        retry = retry.increment(method="GET")
+        assert retry.get_backoff_time() == max_backoff
```
sethmlarson, line 177: Should add an additional case that after another `.increment()` our `get_backoff_time()` is still `max_backoff`

sethmlarson (changes requested):
One comment, once we have the PR targeting 1.26.x we can merge this.

In addition we'll want to create a newsfragment, see this directory for examples on how to do so: https://github.com/urllib3/urllib3/tree/main/changelog Let me know if you have questions about this.

test/test_retry.py
barrett-schonefeld: Added

barrett-schonefeld:
> In addition we'll want to create a newsfragment

Updated the test and add to the changelog!

pquentin (changes requested):
Can you please use past tense in the changelog entry as mentioned in https://github.com/urllib3/urllib3/blob/main/changelog/README.rst? (Wasn't clear from the contributing docs, so I updated them: https://github.com/urllib3/urllib3/pull/2415)

As you know, we'll also need the PR targeting the 1.26.x branch deprecating the old attribute first. You can look at #2000 for inspiration.

sethmlarson (commented):
After implementing @pquentin's comments (normalize changelog entry, open 1.26.x PR) this PR should be ready to merge.

sethmlarson:
@barrett-schonefeld We're still interested in landing this, but we haven't heard from you in some time, are you still interested in this PR?

pquentin:
Right, please open a new PR, we'll close this one. Thanks!

pyrooka:
I opened the new PRs (https://github.com/urllib3/urllib3/pull/2494, https://github.com/urllib3/urllib3/pull/2495) yesterday, so this can be closed.

PR #2395, opened 2021-08-27
Complete type annotations of several files

jdufresne:


src/urllib3/_collections.py
```diff
@@ -255,7 +255,7 @@ def __init__(self, headers: Optional[ValidHTTPHeaderSource] = None, **kwargs: st
         if kwargs:
             self.extend(kwargs)
 
-    def __setitem__(self, key: str, val: str) -> None:
+    def __setitem__(self, key: Union[bytes, str], val: str) -> None:
```
sethmlarson, line 258: So this is true but I'm not sure if we should recommend it? Wasn't there some discussion about `bytes` headers somewhere, can't find it now.

test/contrib/test_socks.py
```diff
@@ -294,16 +311,17 @@ def request_handler(listener):
...
+        def request_handler(listener: socket.socket) -> None:
             sock = listener.accept()[0]
 
             handler = handle_socks5_negotiation(sock, negotiate=False)
             addr, port = next(handler)
 
             assert addr == b"example.com"
             assert port == 80
-            handler.send(True)
+            with pytest.raises(StopIteration):
```
sethmlarson, line 323: Why is this added?

sethmlarson (changes requested):
This is really great, thanks for this! A few comments for you below:

src/urllib3/__init__.py
```diff
@@ -22,9 +22,11 @@
 
 __author__ = "Andrey Petrov (andrey.petrov@shazow.net)"
 __license__ = "MIT"
-__version__ = __version__
 
 __all__ = (
+    "__author__",
```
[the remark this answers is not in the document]
jdufresne: That makes sense! I added it here for the sake of `docs/conf.py`, but I think it would be better to continue excluding it from type checking.

src/urllib3/_collections.py
jdufresne: Is the preference to?

1. Add ignore comments to all uses
2. Only handle str, update tests, implementation and the type signature?

test/contrib/test_socks.py
jdufresne: The original implementation of `handle_socks*_negotiation` would yield multiple types of varying types. Sometimes a (addr, port) tuple and other times a bool. The bool looks like it was added to avoid handling a `StopIteration` exception.The yielded value is not actually used. We could define the generator to yield `Union[Tuple, bool]` however, this complicates the call sites as now the type checker will complain that `bool` is not iterable. I found it simpler to only yield a single type `(addr, port)` and in other code branches to stop the generator, hence `StopIteration`.

Is there a preference for a different approach?

src/urllib3/_collections.py
jdufresne: I went with option 2 in #2397 but am very open to discussing other options.

jdufresne: I have reverted the changes in `src/urllib3/_collections.py`. They now continue to be typed for `str` only even though the implementation allows bytes. I have added `# type: ignore` comments to the necessary tests with a comment. Thanks for clarifying.

sethmlarson: I appreciate you taking initiative here, we may end up deprecating using `bytes` for this case to push a little harder towards using our APIs as intended. We can handle that separately though.

test/contrib/test_socks.py
sethmlarson: Your approach is much better, thanks for making this change :+1:

sethmlarson (approved):
LGTM, thanks for all this great work :muscle: Is `jon.dufresne@gmail.com` the best way to reach you?

jdufresne:
> Is jon.dufresne@gmail.com the best way to reach you?

Yup. :slightly_smiling_face:

PR #2397, opened 2021-08-29
Remove support for passing a bytes key to ``HTTPHeaderDict``

jdufresne:


jdufresne:
Motivation for this change started in thread: https://github.com/urllib3/urllib3/pull/2395#discussion_r697938359

sethmlarson:
Even though this is a "major" release we're trying to minimize breaking changes while opening up potential for future improvements. Supporting byte keys is both fortunately (because imo it's the correct way to do HTTP headers) and unfortunately (strings are easy to work with, users prefer them, it's what our API has historically been) widely used. So we're kinda soft-disallowing them in v2.0 via type hints but still secretly accepting them.

jdufresne:
Thanks for clarifying. I have updated #2395 with this in mind and will close this PR.

PR #2399, opened 2021-08-31
Add type hints to test.with_dummyserver.test_chunked_transfer

hramezani:

### DIFF
--- a/test/with_dummyserver/test_chunked_transfer.py
+++ b/test/with_dummyserver/test_chunked_transfer.py
@@ -1,3 +1,6 @@
+import socket
+from typing import List, Optional, Union
+
 import pytest
 
 from dummyserver.testcase import (
@@ -14,10 +17,10 @@ pytestmark = pytest.mark.flaky
 
 
 class TestChunkedTransfer(SocketDummyServerTestCase):
-    def start_chunked_handler(self):
+    def start_chunked_handler(self) -> None:
         self.buffer = b""
 
-        def socket_handler(listener):
+        def socket_handler(listener: socket.socket) -> None:
             sock = listener.accept()[0]
 
             while not self.buffer.endswith(b"\r\n0\r\n\r\n"):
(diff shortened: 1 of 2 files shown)


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -33,9 +36,9 @@ def socket_handler(listener):
 
         self._start_server(socket_handler)
 
-    def test_chunks(self):
+    def test_chunks(self) -> None:
         self.start_chunked_handler()
-        chunks = ["foo", "bar", "", "bazzzzzzzzzzzzzzzzzzzzzz"]
+        chunks = [b"foo", b"bar", b"", b"bazzzzzzzzzzzzzzzzzzzzzz"]
```
bluetech, line 41: It does seem like `request_chunked` explicitly supports `str` chunks:

https://github.com/urllib3/urllib3/blob/cff8b5ba3d2ba9338e31c102e1e91273124df297/src/urllib3/connection.py#L326-L327

I'm not sure if this is a case "we support it but don't do this, so type annotation forbid this" or "this is supported so `_TYPE_BODY` should allow it". To me chunked + str sounds like a bad idea but I don't really know.

bluetech (commented):
I left a comment regarding the str -> bytes changes, other than that LGTM.

test/with_dummyserver/test_chunked_transfer.py
hramezani: @sethmlarson @pquentin Could you please elaborate on this?

pquentin: I'd lean for "we support it but don't do this, so type annotation forbid this".

@sethmlarson thoughts?

sethmlarson: ++ on support with type annotations forbidding it.

test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -33,22 +36,30 @@ def socket_handler(listener):
 
         self._start_server(socket_handler)
 
-    def test_chunks(self):
+    @pytest.mark.parametrize(
```
sethmlarson, line 39: :muscle: Love this!

sethmlarson (approved):
LGTM, thanks for the revisions!

PR #2401, opened 2021-09-01
Add type hints to test.with_dummyserver.test_connectionpool

hramezani:

### DIFF
--- a/src/urllib3/_request_methods.py
+++ b/src/urllib3/_request_methods.py
@@ -1,4 +1,14 @@
-from typing import Any, Dict, Mapping, Optional, Sequence, Tuple, Union
+from typing import (
+    Any,
+    Dict,
+    Generic,
+    Mapping,
+    Optional,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+)
 from urllib.parse import urlencode
 
 from .connection import _TYPE_BODY
@@ -12,7 +22,10 @@ _TYPE_ENCODE_URL_FIELDS = Union[
 ]
 
 
-class RequestMethods:
+BaseHTTPResponseT = TypeVar("BaseHTTPResponseT", bound=BaseHTTPResponse)
+
+
+class RequestMethods(Generic[BaseHTTPResponseT]):
     """
     Convenience mixin for classes who implement a :meth:`urlopen` method, such
     as :class:`urllib3.HTTPConnectionPool` and
@@ -55,7 +68,7 @@ class RequestMethods:
         encode_multipart: bool = True,
         multipart_boundary: Optional[str] = None,
         **kw: Any,
-    ) -> BaseHTTPResponse:  # Abstract
+    ) -> BaseHTTPResponseT:  # Abstract
         raise NotImplementedError(
             "Classes extending RequestMethods must implement "
             "their own ``urlopen`` method."
@@ -69,7 +82,7 @@ class RequestMethods:
         fields: Optional[_TYPE_FIELDS] = None,
         headers: Optional[Mapping[str, str]] = None,
         **urlopen_kw: Any,
-    ) -> BaseHTTPResponse:
+    ) -> BaseHTTPResponseT:
         """
         Make a request using :meth:`urlopen` with the appropriate encoding of
         ``fields`` based on the ``method`` used.
@@ -107,7 +120,7 @@ class RequestMethods:
         fields: Optional[_TYPE_ENCODE_URL_FIELDS] = None,
         headers: Optional[Mapping[str, str]] = None,
         **urlopen_kw: str,
-    ) -> BaseHTTPResponse:
+    ) -> BaseHTTPResponseT:
         """
         Make a request using :meth:`urlopen` with the ``fields`` encoded in
         the url. This is useful for request methods like GET, HEAD, DELETE, etc.
@@ -132,7 +145,7 @@ class RequestMethods:
         encode_multipart: bool = True,
         multipart_boundary: Optional[str] = None,
         **urlopen_kw: str,
-    ) -> BaseHTTPResponse:
+    ) -> BaseHTTPResponseT:
         """
         Make a request using :meth:`urlopen` with the ``fields`` encoded in
         the body. This is useful for request methods like POST, PUT, PATCH, etc.
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -27,7 +27,7 @@ from .exceptions import (
     ProxySchemeUnknown,
     URLSchemeUnknown,
 )
-from .response import BaseHTTPResponse
+from .response import BaseHTTPResponse, HTTPResponse
 from .util.connection import _TYPE_SOCKET_OPTIONS
 from .util.proxy import connection_requires_http_tunnel
 from .util.retry import Retry
@@ -168,7 +168,7 @@ key_fn_by_scheme = {
 pool_classes_by_scheme = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
 
 
-class PoolManager(RequestMethods):
+class PoolManager(RequestMethods[HTTPResponse]):
     """
     Allows for arbitrary requests while transparently keeping track of
     necessary connection pools for you.
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -1144,8 +1144,9 @@ class TestRetry(HTTPDummyServerTestCase):
                 "GET", "/successful_retry", headers=headers, retries=retry
             )
             assert resp.status == 200
-            assert resp.retries.total == 1  # type: ignore[attr-defined]
-            assert resp.retries.history == (  # type: ignore[attr-defined]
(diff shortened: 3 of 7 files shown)


test/with_dummyserver/test_connectionpool.py
```diff
@@ -229,20 +230,20 @@ def test_wrong_specific_method(self):
             r = pool.request("POST", "/specific_method", fields={"method": "GET"})
             assert r.status == 400, r.data
 
-    def test_upload(self):
+    def test_upload(self) -> None:
         data = "I'm in ur multipart form-data, hazing a cheezburgr"
-        fields = {
```
bluetech, line 234: Looking at `_TYPE_FIELDS` it seems like it should allow this, why did this need to change?

test/with_dummyserver/test_connectionpool.py
```diff
@@ -1112,28 +1135,28 @@ def test_retry_reuse_safe(self):
...
         with HTTPConnectionPool(self.host, self.port) as pool:
             headers = {"test-name": "test_retry_return_in_response"}
             retry = Retry(total=2, status_forcelist=[418])
             resp = pool.request(
                 "GET", "/successful_retry", headers=headers, retries=retry
             )
             assert resp.status == 200
-            assert resp.retries.total == 1
-            assert resp.retries.history == (
+            assert resp.retries.total == 1  # type: ignore[attr-defined]
```
bluetech, line 1146: Which attr is not defined here, and why?

bluetech (commented):
Couple of questions, looks good otherwise!

test/with_dummyserver/test_connectionpool.py
hramezani: Yeah, it seems so. but without these change we get:
```
test/with_dummyserver/test_connectionpool.py:243: error: Argument "fields" to "request" of "RequestMethods" has incompatible type "Dict[str, Sequence[str]]"; expected "Union[Sequence[Union[Tuple[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], RequestField]], Dict[str, Union[Union[str, bytes], Tuple[str, Union[str, bytes]], Tuple[str, Union[str, bytes], str]]], None]"  [arg-type]
```

Do you have any idea what is the reason?

test/with_dummyserver/test_connectionpool.py
hramezani: it was `retries` and the `request` method returned `BaseHTTPResponse`.
I've added a commit and changed the return type to `HTTPResponse`
Thanks for bringing this.

src/urllib3/_request_methods.py
```diff
@@ -55,7 +55,7 @@ def urlopen(
         encode_multipart: bool = True,
         multipart_boundary: Optional[str] = None,
         **kw: Any,
-    ) -> BaseHTTPResponse:  # Abstract
+    ) -> HTTPResponse:  # Abstract
```
sethmlarson, line 58: We don't want to have too specific a type here, returning `BaseHTTPResponse` is intentional.

test/with_dummyserver/test_connectionpool.py
sethmlarson: Agreed here, it's really unfortunate that `str` can be inferred as `Sequence[str]`

src/urllib3/_request_methods.py
bluetech: Perhaps `RequestMethods` can be made generic in the response type, with the type var bound by `BaseHTTPResponse`. Then each class can extend with its actual response type? (I didn't check if that makes sense)

Aside: I personally dislike mixin classes like `RequestMethods`, but that's another discussion...

pquentin: (RequestMethods is definitely confusing.)

hramezani: @sethmlarson I've reverted my commit for changing the type.

@bluetech
> Perhaps RequestMethods can be made generic in the response type, with the type var bound by BaseHTTPResponse. Then each class can extend with its actual response type? 

Could you please give me an example or explain how to do it?

test/with_dummyserver/test_connectionpool.py
hramezani: Thanks, @bluetech, I've added explicit annotation for `fields`.

src/urllib3/_request_methods.py
bluetech: Change `RequestMethods` to

```py
BaseHTTPResponseT = TypeVar('BaseHTTPResponseT', bound=BaseHTTPResponse)

class RequestMethods(Generic[BaseHTTPResponseT]):
    def request(...) -> BaseHTTPResponseT: ...
```

and then for each mixing-in inherit from `RequestMethods[TheActualBaseHTTPResponseTypeForThisClass]`.

Again, didn't check if it would actually help, but it allows each class to decide which response type it wants for the mixed-in methods.

hramezani: Thanks @bluetech for your code example. I've changed it as you suggested

src/urllib3/poolmanager.py
```diff
@@ -168,7 +168,7 @@ def _default_key_normalizer(
 pool_classes_by_scheme = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
 
 
-class PoolManager(RequestMethods):
+class PoolManager(RequestMethods[HTTPResponse]):
```
bluetech, line 171: This one I'm not sure about -- if both users of `RequestMethods` use `HTTPResponse` then why does `BaseHTTPResponse` exist?

Looking into `BaseHTTPResponse` I found #1986 where @sethmlarson mentions

> Ensure that type hints for .request() of BaseHTTPConnection annotate as BaseHTTPResponse instead of HTTPResponse

but currently there is no `BaseHTTPConnection` so I'm guessing this work is still in progress?

sethmlarson: The reason we want to have `BaseHTTPConnection` and `BaseHTTPResponse` is so we can decouple ourselves from the httplib APIs that we're not guaranteeing to be available for alternate HTTP protocol implementations. So for PoolManager and ConnectionPool we shouldn't be typing the responses/connections to be `HTTPResponse`, they should be `BaseHTTPResponse` so users know which APIs will be supported into the future.

Does this make sense? If not we can revisit this, the primary goal is to decouple our API from httplib.

bluetech: Makes sense.

The relevant point for this PR is how does `PoolManager` fit into this:

- Will be used as is with some configuration which will decide which backend `request` etc. use.
    - Configuration is dynamic: cannot make `PoolManager` generic in the backend.
    - Configuration is static (defined when the `PoolManager` is constructed): can use a generic.
- Will have `BasePoolManager` with subclasses for each backend.
- Will have different methods for different backends.
- Or maybe the library should be designed such that the backend never needs to "leak" i.e. all functionality should be present on the base types and the sub-types never need to be exposed in any public API.
- Something else

If this is not decided yet, then we should be careful not to add typing that will lock a specific approach, i.e. probably defer these parts of the PR.

sethmlarson: @bluetech I think the fourth item you listed is what we're aiming for now:

> library should be designed such that the backend never needs to "leak" i.e. all functionality should be present on the base types and the sub-types never need to be exposed in any public API.

I'm not sure if generics are needed for this to work out fine, I'm really happy someone is putting extra brain cycles on this problem though.

pquentin:
We have a genuine coverage drop in `encode_multipart_formdata` which accepts integer data for backwards compatibility purposes. Yet another mypy finding!

ksnip_20210917-091453

That was added in 2010, so I'd be in favor of removing this but documenting it in the changelog.

hramezani:
@pquentin Thanks for finding this. I added a commit to drop it.

pquentin:
Yeah, I think we should open a requests PR to test that an integer raises a TypeError. (And test separately for None I suppose?). But @sethmlarson and @sigmavirus24 know more about requests, let see what they think first.

sigmavirus24:
> But @sethmlarson and @sigmavirus24 know more about requests, let see what they think first.

The change to requests should be to ensure that Requests itself does this conversion. It's likely unfortunately relied upon by someone

pquentin:
Thanks! In that case it may be simpler to just continue supporting it

sigmavirus24:
I don't know if urllib3 users rely on it. Just that requests users pass chaos-monkey level shit to the library

test/with_dummyserver/test_connectionpool.py
sethmlarson: We should add `retries` to `BaseHTTPResponse`.

sethmlarson:
We'll continue supporting integers here, thanks @sigmavirus24 for replying for Requests users.

sethmlarson:
I've opened a PR which tries to accomplish the same as this PR sans-RequestMethods generics: https://github.com/urllib3/urllib3/pull/2418

sethmlarson:
Closing this PR in favor of the merged https://github.com/urllib3/urllib3/pull/2418, thanks all!

PR #2402, opened 2021-09-03
Add type hints to test.with_dummyserver.test_https

hramezani:

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -7,6 +7,7 @@ import ssl
 import sys
 import tempfile
 import warnings
+from pathlib import Path
 from test import (
     LONG_TIMEOUT,
     SHORT_TIMEOUT,
@@ -16,10 +17,13 @@ from test import (
     requires_ssl_context_keyfile_password,
     resolvesLocalhostFQDN,
 )
+from test.conftest import ServerConfig
+from typing import List, Optional
 from unittest import mock
 
 import pytest
 import trustme
+from py._path.local import LocalPath
 
 import urllib3.util as util
 import urllib3.util.ssl_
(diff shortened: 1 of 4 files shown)


sethmlarson (approved):
LGTM, thanks!

test/with_dummyserver/test_https.py
```diff
@@ -74,31 +78,31 @@
...
-    def tls_version(self):
+    def tls_version(self) -> "ssl.TLSVersion":
         if self.tls_protocol_name is None:
             return pytest.skip("Skipping base test class")
         try:
             from ssl import TLSVersion
         except ImportError:
             return pytest.skip("ssl.TLSVersion isn't available")
-        return getattr(TLSVersion, self.tls_protocol_name.replace(".", "_"))
+        return getattr(TLSVersion, self.tls_protocol_name.replace(".", "_"))  # type: ignore[no-any-return]
```
bluetech, line 93: Enum types implement `__getitem__`  and the typing probably knows about it, so I think this can be

```py
return TLSVersion[self.tls_protocol_name.replace(".", "_")]
```

test/with_dummyserver/test_https.py
```diff
@@ -16,10 +17,13 @@
     requires_ssl_context_keyfile_password,
     resolvesLocalhostFQDN,
 )
+from test.conftest import ServerConfig
+from typing import List, Optional
 from unittest import mock
 
 import pytest
 import trustme
+from py._path.local import LocalPath
```
bluetech, line 26: `py._path` is private; the intended usage is `import py` and `path: py.path.local`.

BTW, some time in the future pytest is going to deprecate `py.path.local` fixtures (`tmpdir` and `tmpdir_factory`) in favor of pathlib fixtures (`tmp_path` and `tmp_path_factory`).

test/with_dummyserver/test_https.py
```diff
@@ -272,7 +276,7 @@ def test_context_combines_with_ca_certs(self):
                 assert not warn.called, warn.call_args_list
 
     @notSecureTransport()  # SecureTransport does not support cert directories
-    def test_ca_dir_verified(self, tmpdir):
+    def test_ca_dir_verified(self, tmpdir: Path) -> None:
```
bluetech, line 279: This should be `tmpdir: py.path.local` (though might as well replace it with `tmp_path: Path` as mentioned above)

bluetech (commented):
Couple of post-merge comments I noticed

hramezani:
Thanks @bluetech for the comments. I will address them

test/with_dummyserver/test_https.py
pquentin: Definitely, we only used py.path for Python 2 support.

PR #2405, opened 2021-09-06
[docs] Fixed first delay value when retrying.

facundobatista:
The first delay is the backoff factor itself, not 0, as...
```
>>> .1 * 2 ** (1 - 1)
0.1
```


pquentin:
This is incorrect, we only start sleeping after the second try, as mentioned a few lines above your change.

PR #2414, opened 2021-09-16
Add type hints to test.with_dummyserver.test_proxy_poolmanager

hramezani:

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -1,7 +1,7 @@
 import socket
 import threading
 from contextlib import contextmanager
-from typing import Any, Callable, ClassVar, Generator, Iterable, Optional
+from typing import Any, Callable, ClassVar, Dict, Generator, Iterable, Optional
 
 import pytest
 from tornado import httpserver, ioloop, web
@@ -179,23 +179,32 @@ class HTTPSDummyServerTestCase(HTTPDummyServerTestCase):
 class HTTPDummyProxyTestCase:
     io_loop: ClassVar[ioloop.IOLoop]
 
-    http_host = "localhost"
-    http_host_alt = "127.0.0.1"
+    http_host: ClassVar[str] = "localhost"
+    http_host_alt: ClassVar[str] = "127.0.0.1"
     http_server: ClassVar[httpserver.HTTPServer]
     http_port: ClassVar[int]
+    http_url: ClassVar[str]
+    http_url_alt: ClassVar[str]
 
-    https_host = "localhost"
-    https_host_alt = "127.0.0.1"
-    https_certs = DEFAULT_CERTS
+    https_host: ClassVar[str] = "localhost"
+    https_host_alt: ClassVar[str] = "127.0.0.1"
+    https_certs: ClassVar[Dict[str, Any]] = DEFAULT_CERTS
     https_server: ClassVar[httpserver.HTTPServer]
     https_port: ClassVar[int]
+    https_url: ClassVar[str]
+    https_url_alt: ClassVar[str]
 
-    proxy_host = "localhost"
-    proxy_host_alt = "127.0.0.1"
+    proxy_host: ClassVar[str] = "localhost"
+    proxy_host_alt: ClassVar[str] = "127.0.0.1"
     proxy_server: ClassVar[httpserver.HTTPServer]
     proxy_port: ClassVar[int]
+    proxy_url: ClassVar[str]
     https_proxy_server: ClassVar[httpserver.HTTPServer]
     https_proxy_port: ClassVar[int]
+    https_proxy_url: ClassVar[str]
+
+    certs_dir: ClassVar[str] = ""
+    bad_ca_path: ClassVar[str] = ""
 
     server_thread: ClassVar[threading.Thread]
 
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -1032,7 +1032,7 @@ class HTTPSConnectionPool(HTTPConnectionPool):
             )
 
 
-def connection_from_url(url: str, **kw: Any) -> ConnectionPool:
+def connection_from_url(url: str, **kw: Any) -> HTTPConnectionPool:
     """
     Given a url, return an :class:`.ConnectionPool` instance of its host.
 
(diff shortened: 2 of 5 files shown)


dummyserver/testcase.py
```diff
@@ -181,19 +181,28 @@ class HTTPDummyProxyTestCase:
...
     proxy_host = "localhost"
     proxy_host_alt = "127.0.0.1"
     proxy_server: ClassVar[httpserver.HTTPServer]
     proxy_port: ClassVar[int]
+    proxy_url: ClassVar[str]
     https_proxy_server: ClassVar[httpserver.HTTPServer]
     https_proxy_port: ClassVar[int]
+    https_proxy_url: ClassVar[str]
+
+    certs_dir = ""
```
sethmlarson, line 204: Should these be `ClassVar[str]` as well?

src/urllib3/poolmanager.py
```diff
@@ -537,7 +543,7 @@ class ProxyManager(PoolManager):
 
     def __init__(
         self,
-        proxy_url: str,
+        proxy_url: Union[ConnectionPool, str],
```
sethmlarson, line 546: This should be `HTTPConnectionPool` since we use `.scheme` below.

src/urllib3/poolmanager.py
```diff
@@ -547,8 +553,10 @@ def __init__(
     ) -> None:
 
         if isinstance(proxy_url, HTTPConnectionPool):
-            proxy_url = f"{proxy_url.scheme}://{proxy_url.host}:{proxy_url.port}"
-        proxy = parse_url(proxy_url)
+            str_proxy_url = f"{proxy_url.scheme}://{proxy_url.host}:{proxy_url.port}"
+        else:
+            str_proxy_url = cast(str, proxy_url)
```
sethmlarson, line 558: If we do the above change can we remove the `cast`?

sethmlarson (changes requested):
Two comments, otherwise this looks great! Thank you :)

dummyserver/testcase.py
hramezani: as we don't have `ClassVar` for the other class variables with default values in this file, I didn't define it by `ClassVar`
for example:
https://github.com/urllib3/urllib3/blob/7c9ca88b7dd5fe6b0e9fac52f72202edd74bd8b2/dummyserver/testcase.py#L191

Should we change them as well?

src/urllib3/poolmanager.py
hramezani: Then we have an error in `test_oldapi`
https://github.com/urllib3/urllib3/blob/7c9ca88b7dd5fe6b0e9fac52f72202edd74bd8b2/test/with_dummyserver/test_proxy_poolmanager.py#L165

because `connection_from_url` returns `ConnectionPool` and `ProxyManager` accepts `HTTPConnectionPool`

```
test/with_dummyserver/test_proxy_poolmanager.py:168: error: Argument 1 to "ProxyManager" has incompatible type "ConnectionPool"; expected "Union[HTTPConnectionPool, str]"  [arg-type]
```

dummyserver/testcase.py
bluetech: It looks like they get set on the `cls` so they should be `ClassVar` as well. If `ClassVar` is not used then it is taken to be a default for instance variables.

src/urllib3/poolmanager.py
sethmlarson: Bleh, `ConnectionPool` is both useless and getting in the way. Makes me want to make it an empty class and move everything to `HTTPConnectionPool`.

dummyserver/testcase.py
hramezani: Ok, Thanks for the explanation. Changed them to `ClassVar`

src/urllib3/poolmanager.py
hramezani: Shall we change it to `HTTPConnectionPool` and ignore the mypy error in test?

sethmlarson: Change both here and `connection_from_url` to be `HTTPConnectionPool`.

hramezani: Done!

hramezani:
There are some mypy errors related ‍to `test.with_dummyserver.test_poolmanager` that will be fixed after we merged https://github.com/urllib3/urllib3/pull/2420

sethmlarson (approved):
LGTM

src/urllib3/poolmanager.py
```diff
@@ -537,7 +537,7 @@ class ProxyManager(PoolManager):
 
     def __init__(
         self,
-        proxy_url: str,
+        proxy_url: Union[HTTPConnectionPool, str],
```
bluetech, line 540: It's funky that an argument called `proxy_url` can be an `HTTPConnectionPool` but I'm sure there's a story there :grin:

src/urllib3/connectionpool.py
```diff
@@ -1032,7 +1032,7 @@ def _validate_conn(self, conn: HTTPConnection) -> None:
             )
 
 
-def connection_from_url(url: str, **kw: Any) -> ConnectionPool:
+def connection_from_url(url: str, **kw: Any) -> HTTPConnectionPool:
```
bluetech, line 1035: I don't know if `connection_from_url` wants to keep the option for itself to someday return a `ConnectionPool` which is not an `HTTPConnectionPool`. If it does, then this change would interfere; if not, it's good.

bluetech (approved):

src/urllib3/poolmanager.py
sethmlarson: Hmm yeah maybe we shouldn't advertise this.

src/urllib3/connectionpool.py
sethmlarson: ++ Yeah these functions aren't what we're recommending anymore, I think it's safe to say we'll only be returning `HTTPConnectionPool`.

src/urllib3/poolmanager.py
sethmlarson: @hramezani Can we remove the type hint for `HTTPConnectionPool` since it's a fallback to cover old behavior? Instead we can add `type: ignore` to the test suite.

sethmlarson (changes requested):
One more thing that @bluetech mentioned that makes much more sense. Thanks @bluetech!

sethmlarson (approved):
LGTM, thanks for iterating on this! Also thanks to @bluetech for pointing out that we shouldn't be advertising the clunky old API :)

Issue #2378, opened 2021-08-18
Fix missing link targets in docs


pquentin:
Since the BaseHTTPResponse split in #2083, various links in the docs are broken. Here's how the [Response Content](https://urllib3.readthedocs.io/en/stable/user-guide.html#response-content) looks like:

<img width="580" alt="Screenshot 2021-08-18 at 11 34 23" src="https://user-images.githubusercontent.com/42327/129857043-3f8ddf59-fa22-4c27-b8c9-9333d4a2194c.png">

There are other cases, eg. the API Reference doesn't include the top-level functions, so `disable_warnings()` or `request()` are not documented.

To fix everything in one go, we need to enable Sphinx's `nitpicky` mode. Despite the vague name, it only does one thing: warn about missing references. And since we treat Sphinx warnings as errors, this will allow us to fix those missing references and prevent any regressions in the future.

Here's how Trio configures this `nitpicky` mode: https://github.com/python-trio/trio/blob/cb48b33a42b09dde955617f93923ed8f5fa02426/docs/source/conf.py#L39-L57

The process here, for each of the 78 failures is to figure out if:

 * the link is wrong and should be fixed,
 * the function exists and should be in the API Reference (in that case, add it somewhere in docs/reference),
 * or the link should be added to `nitpicky_ignore`.

It's OK to open a PR that fixes only a few issues!

V1NAY8:
`-W` flag has to be removed in `noxfile.py` in `docs` method, otherwise it stops at the first error

pquentin:
@V1NAY8 Do you still intend to work on this?

V1NAY8:
Nope. It's good for anyone else to pick :)

V1NAY8:
@pquentin I did some work on this. This is still open for anyone else to contribute for other missing links

RatanShreshtha:
I can pick this up

pquentin:
@RatanShreshtha It turns out @V1NAY8 opened https://github.com/urllib3/urllib3/pull/2416, so if he's still interested please focus on a different file

PR #2416, opened 2021-09-17
Fix missing links in docs

V1NAY8:
Helps in resolving #2378


pquentin:
Sorry, the build fails due to https://github.com/urllib3/urllib3/pull/2425, I restored my branch for now.

docs/user-guide.rst
```diff
@@ -30,7 +30,7 @@ so that you don't have to:
 
     http = urllib3.PoolManager()
 
-To make a request use :meth:`~poolmanager.PoolManager.request`:
+To make a request use :meth:`~request.RequestMethods.request`:
```
pquentin, line 33:
```suggestion
To make a request use :meth:`~urllib3.PoolManager.request`:
```

This only changes the warning, but you still have a warning because this does not link to anywhere in the documentation. The fix in this case is to add `:inherited-members:` to `urllib3.PoolManager` in `docs/reference/urllib3.poolmanager.rst`. And since it's documented as `urllib3.PoolManager`, I've found that using that in the text allows Sphinx to find it, hence the suggestion.

docs/user-guide.rst
```diff
@@ -49,7 +49,7 @@ To make a request use :meth:`~poolmanager.PoolManager.request`:
 ``request()`` returns a :class:`~response.HTTPResponse` object, the
 :ref:`response_content` section explains how to handle various responses.
 
-You can use :meth:`~poolmanager.PoolManager.request` to make requests using any
+You can use :meth:`~request.RequestMethods.request` to make requests using any
```
pquentin, line 52: Same as above.

docs/user-guide.rst
```diff
@@ -82,8 +82,8 @@ Response Content
 ----------------
 
 The :class:`~response.HTTPResponse` object provides
-:attr:`~response.HTTPResponse.status`, :attr:`~response.HTTPResponse.data`, and
-:attr:`~response.HTTPResponse.headers` attributes:
+:attr:`~response.BaseHTTPResponse.status`, :attr:`~response.HTTPResponse.data`, and
+:attr:`~response.BaseHTTPResponse.headers` attributes:
```
pquentin, line 86: This case is different from `request()`. Unlike `RequestMethods` we may want to show the inheritance explicitly in the documentation (i.e. document `headers` in `BaseHTTPResponse`, and `data` in `HTTPResponse`), so this may be correct. (Note: this change not lower the number of warnings because that attribute is not documented and `BaseHTTPResponse` isn't currently part of the reference. Here's how you can [document attributes](https://github.com/python-trio/trio/blob/4b0388b7c9442e9bea5259460cd64687bff8c26b/trio/_subprocess.py#L86-L107))

@sethmlarson Do we want to document `BaseHTTPResponse` separately or should each subclass documentation contain the inherited members docs?

docs/user-guide.rst
```diff
@@ -166,7 +166,7 @@ Request Data
 Headers
 ~~~~~~~
 
-You can specify headers as a dictionary in the ``headers`` argument in :meth:`~poolmanager.PoolManager.request`:
+You can specify headers as a dictionary in the ``headers`` argument in :meth:`~urllib3.request`:
```
pquentin, line 169: `urllib3.request` is a `func`, not a `meth`. This applies to all the changes below.

It's also not in the reference. To do that, you could add a `docs/reference/urllib3.rst` file:

```
urllib3 top level
=================

.. autofunction:: urllib3.request
```

and add `urllib3` in `reference/index.rst`.

docs/conf.py
```diff
@@ -82,3 +82,7 @@
 }
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
+
+# Warn about all references to unknown targets
+nitpicky = True
```
pquentin, line 87: `nitpicky = True` is correct, but please remove it as we don't want to include until all links are fixed.

pquentin (changes requested):

docs/user-guide.rst
pquentin: Dicussed a bit offline with Seth, and he confirmed that we don't want `BaseHTTPResponse` to be in the actual docs.

pquentin:
@V1NAY8 Are you still planning to work on this? Need any help?

V1NAY8:
HI @pquentin Finding time is tough these days. I hope someone else can pick this. :)

sethmlarson:
@V1NAY8 Are you still interested in landing this one? No worries if there's no time to.

V1NAY8:
Let me give it a shot once again!

docs/user-guide.rst
```diff
@@ -484,7 +484,7 @@ to :meth:`~poolmanager.PoolManager.request`:
     )
 
     print(type(resp))
-    # <class "urllib3.response.HTTPResponse">
+    # <class "urllib3.urllib3.response.HTTPResponse">
```
sethmlarson, line 487: This addition isn't needed.

docs/user-guide.rst
```diff
@@ -508,7 +508,7 @@ instance which lets you specify separate connect and read timeouts:
     )
 
     print(type(resp))
-    # <urllib3.response.HTTPResponse>
+    # <urllib3.urllib3.response.HTTPResponse>
```
sethmlarson, line 511: Same comment as above.

dev-requirements.txt
```diff
@@ -10,3 +10,5 @@ cryptography==35.0.0
 backports.zoneinfo==0.2.1;python_version<"3.9"
 # https://github.com/twisted/towncrier/pull/356
 git+https://github.com/twisted/towncrier.git
+nox
```
sethmlarson, line 13: Let's remove these since they're not needed for this PR.

sethmlarson (commented):
Thanks for this, a couple small comments:

PR #2417, opened 2021-09-17
[1.26] Strip brackets from IPv6 hosts before certificate hostname matching

pquentin:
Backport of #2241 

Not working yet, but can be a starting point.


pquentin:
I now understand what Seth was saying: we don't always use our own ssl_match_hostname function in 1.26, so modifying it doesn't help.

PR #2418, opened 2021-09-17
Add type hints to test.with_dummyserver.test_connectionpool

sethmlarson:
Created as an alternative to https://github.com/urllib3/urllib3/pull/2401 by adding `retries` to `BaseHTTPResponse` and making `_TYPE_FIELDS` use `Mapping` instead of `Dict`.


hramezani (approved):
LGTM :+1:. Thanks for taking care!

test/with_dummyserver/test_connectionpool.py
```diff
@@ -263,23 +265,23 @@ def test_request_method_body(self):
...
         fieldname = "myfile"
         filename = "\xe2\x99\xa5.txt"
         data = "\xe2\x99\xa5".encode()
         size = len(data)
 
-        fields = {
+        fields: Dict[str, _TYPE_FIELD_VALUE_TUPLE] = {
             "upload_param": fieldname,
             "upload_filename": filename,
-            "upload_size": size,
```
bluetech, line 275: I wonder what is the problem here - is an int not allowed by the types?

bluetech (approved):
LGTM, with one question.

The `retries` change is the sort of thing I had hoped type-checking the tests would uncover so that's good :)

sethmlarson:
No need to rereview, will merge once CI passes.

PR #2424, opened 2021-09-23
Remove integration tests for botocore with Python 2.7

### LINKED ISSUE #2423: Remove integration tests for Botocore with Python 2.7
Botocore dropped support for Python 2.7 in July so we don't have to do integration testing with Botocore+Python 2.7 on the 1.26.x branch any longer.

Reference: https://github.com/urllib3/urllib3/pull/2422

imkaka:
Closes #2423 

@sethmlarson Please let me know if anything else is needed here.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -12,10 +12,10 @@ jobs:
       matrix:
         python-version: [2.7, 3.9]
         downstream: [botocore, requests]
-      exclude:
-        # excludes botocore in Python 2.7
-        - python-version: 2.7
-          downstream: botocore
+        exclude:
+          # excludes botocore in Python 2.7
+          - python-version: 2.7
+            downstream: botocore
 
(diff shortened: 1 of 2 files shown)


.github/workflows/integration.yml
```diff
@@ -12,6 +12,11 @@ jobs:
       matrix:
         python-version: [2.7, 3.9]
         downstream: [botocore, requests]
+      exclude:
+        # excludes botocore in Python 2.7
+        - python-version: 2.7
+          downstream: botocore
+
```
pquentin, line 19: This should be under `matrix`, the indentation is currently incorrect

imkaka: @pquentin Thanks, fixed it please have a look.

pquentin (approved):
Thanks! This works now (tested in my own fork).

sethmlarson (approved):
LGTM

PR #2427, opened 2021-09-24
Upgrade Python 3.10.0-beta.4 to 3.10.0-rc.2

cclauss:
https://www.python.org/downloads/release/python-3100rc2

https://github.com/actions/python-versions/releases

PR #2428, opened 2021-09-24
Fixed typo errors and comments.

MalikIdreesHasanKhan:


pquentin:
I merged #2426 and #2427 as they came first, don't conflate concerns, and don't contain irrelevant changes or unneeded fixes (such as function and Enterprises). I don't understand what was your intent here.

PR #2429, opened 2021-09-24
[1.26] Add exception wrapper for HTTPS proxy connection failure

pquentin:
This backports #2133 and could help with various reports we're receiving.


sethmlarson (approved):
LGTM, thank you!

PR #2430, opened 2021-09-24
Remove redundant requests integration test

pquentin:
We now run the full requests test suite on every pull request. It caught more issues than this simple test and is actually reliable.


hramezani (approved):
👍

sethmlarson (approved):
LGTM, thanks!

PR #2431, opened 2021-09-24
Bump linters

pquentin:


src/urllib3/util/retry.py
```diff
@@ -235,7 +235,7 @@ def __init__(
         self.history = history or ()
         self.respect_retry_after_header = respect_retry_after_header
         self.remove_headers_on_redirect = frozenset(
-            [h.lower() for h in remove_headers_on_redirect]
+            h.lower() for h in remove_headers_on_redirect
```
pquentin, line 238: This is technically slower but unlikely to matter in practice

hramezani (approved):

src/urllib3/util/retry.py
sethmlarson: Which tool does this change come from? We should file an issue that this construction shouldn't be modified.

sethmlarson: My eyes failed me, don't know how I read this as a `str.join()` case. This is likely faster.

sethmlarson (approved):
LGTM

PR #2433, opened 2021-09-24
Lean on SSLContext on CPython versions that support it

### LINKED ISSUE #2191: Enable hostname_checks_common_name on Python versions that support it
We'll have to wait on the resolution of https://bugs.python.org/issue43522, and then based on that decide how we detect the feature:

 * use an hypothetical new flag?
 * 3.10?
 * 3.7.x, 3.8.x, 3.9.x and 3.10 since this fix will be backported?

We'll have to wait for https://github.com/python/cpython/pull/24899 to land and to be in the 3.10 we use in GitHub Actions, probably when 3.10.0a7 get released.

pquentin:
Closes #2191

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
           - python-version: 3.9.2
             os: ubuntu-latest
             experimental: false
-            nox-session: test-3.8
+            nox-session: test-3.9
           - python-version: 3.10.0-rc.2
             os: ubuntu-latest
             experimental: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,11 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: test_brotlipy
+          # Test CPython with a broken hostname_checks_common_name
+          - python-version: 3.9.2
+            os: ubuntu-latest
+            experimental: false
+            nox-session: test-3.8
           - python-version: 3.10.0-rc.2
             os: ubuntu-latest
             experimental: false
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -67,6 +67,8 @@ try:  # Do we have ssl at all?
     )
     PROTOCOL_SSLv23 = PROTOCOL_TLS
 
+    # Setting SSLContext.hostname_checks_common_name = False didn't work before CPython
+    # 3.8.9, 3.9.3, and 3.10" (but OK on PyPy) or OpenSSL 1.1.1l / OpenSSL 3.0.0-alpha15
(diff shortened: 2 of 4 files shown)


src/urllib3/util/ssl_.py
```diff
@@ -67,6 +67,20 @@ def _is_ge_openssl_v1_1_1(
     )
     PROTOCOL_SSLv23 = PROTOCOL_TLS
 
+    # https://github.com/urllib3/urllib3/issues/2192#issuecomment-821832963
```
sethmlarson, line 70: Can we put a brief description of why we're doing this so readers don't have to click into a URL for context?

Something short like "SSLContext.hostname_checks_common_name = False didn't work before CPython 3.8.9, 3.9.3, and 3.10"

sethmlarson (dismissed):
This change LGTM, two small comments. Merge when ready!

.github/workflows/ci.yml
```diff
@@ -45,6 +45,11 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: test_brotlipy
+          # Test CPython with a broken hostname_checks_common_name
+          - python-version: 3.9.2
+            os: ubuntu-latest
+            experimental: false
+            nox-session: test-3.8
```
sethmlarson, line 52: Should be test-3.9?

pquentin: Nice catch. With the fix, we do run the tests with 3.9.2. Wasn't noticed by coverage as Python 3.7 still covers this for now.

src/urllib3/util/ssl_.py
pquentin: Thanks, added.

pquentin:
Requesting a new review as I'm now checking the OpenSSL version after all.

src/urllib3/util/ssl_.py
```diff
@@ -36,6 +36,19 @@ def _is_ge_openssl_v1_1_1(
...
+def _is_ge_openssl_v1_1_1l(
+    openssl_version_text: str, openssl_version_number: int
+) -> bool:
+    """Returns True for OpenSSL 1.1.1l+ (>=0x101010cf)
+    LibreSSL reports a version number of 0x20000000 for
+    OpenSSL version number so we need to filter out LibreSSL.
+    """
+    return (
+        not openssl_version_text.startswith("LibreSSL")
+        and openssl_version_number >= 0x101010CF
```
pquentin, line 48: How do I know this? I currently have 1.1.1l on my laptop.

I considered using OPENSSL_VERSION_INFO which is less obscure and more fitting for Python code (which can compare tuples easily) but was lazy. Please tell me if you'd prefer that.

pquentin:
Requesting a new review as the code changed significantly

pquentin:
I've added OpenSSL but this is now an "and", not an "or". Will rework this.

PR #2434, opened 2021-09-24
Don't unconditionally wrap errors with HTTPSProxyError

sethmlarson:
Noticed that when we unconditionally wrap all exceptions with `HTTPSProxyError` we're likely to report inability to connect via TLS when TLS isn't always the problem (DNS, timeouts, etc) so instead we wrap only when we are certain the error is TLS related.

Also adds documentation to help users debug a common scenario where they specify an HTTPS proxy but are actually talking to an HTTP proxy. Will update the error message with the link to this fragment in the docs after docs build.

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -239,9 +239,9 @@ and not ``https://``:
 
 .. code-block:: bash
 
-     # Check environment variables in bash
+     # Check your existing environment variables in bash
      $ env | grep "_PROXY"
-     HTTP_PROXY=https://127.0.0.1:8888
+     HTTP_PROXY=http://127.0.0.1:8888
      HTTPS_PROXY=https://127.0.0.1:8888  # <--- This setting is the problem!
      
      # Make the fix in your current session and test your script
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -210,6 +210,78 @@ an `absolute URI <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_ if the
 **only use this option with trusted or corporate proxies** as the proxy will have
 full visibility of your requests.
 
+.. _https_proxy_error_http_proxy:
+
+How to fix HTTPSProxyError?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


docs/advanced-usage.rst
```diff
@@ -210,6 +210,78 @@ an `absolute URI <https://tools.ietf.org/html/rfc7230#section-5.3.2>`_ if the
...
+to look are environment variables like ``HTTP_PROXY``, ``HTTPS_PROXY``, and ``ALL_PROXY``.
+
+Ensure that the values for all of these environment variables starts with ``http://``
+and not ``https://``:
+
+.. code-block:: bash
+
+     # Check environment variables in bash
+     $ env | grep "_PROXY"
+     HTTP_PROXY=https://127.0.0.1:8888
```
jalopezsilva, line 244: This one would also be problematic right? You might have meant to be ``http://``

src/urllib3/connection.py
```diff
@@ -597,6 +593,36 @@ def _match_hostname(cert: _TYPE_PEER_CERT_RET, asserted_hostname: str) -> None:
         raise
 
 
+def _should_wrap_https_proxy_error(err: Exception) -> bool:
+    """Detects if the error should be wrapped into :class:`urllib3.exceptions.HTTPSProxyError`"""
+    if isinstance(err, ProxyError):
+        err = err.original_error
```
jalopezsilva, line 599: Not covered by tests?

jalopezsilva (commented):
The documentation changes look great and I'm hopeful they will reduce the constant noise we get with misconfigured proxies.  Thanks also for fixing the wrapping on HTTPSProxyError, I might have been overzealous when I did it the first time :sweat_smile: 

Looks like there are some legit failures with "TestProxyManager::test_https_proxymanager_connected_to_http_proxy"?

docs/advanced-usage.rst
sethmlarson: You're right! I'll clarify this, great catch.

sethmlarson:
Thanks for the review @jalopezsilva! I tried to fix the Windows `ConnectionAbortedError` failure, we'll see if that fixes things?

jalopezsilva (approved):
Perfect, thank you!

PR #2436, opened 2021-09-29
Add type hints to test.with_dummyserver.test_socketlevel

hramezani:

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -9,6 +9,7 @@ import socket
 import ssl
 import tempfile
 from collections import OrderedDict
+from pathlib import Path
 from test import (
     LONG_TIMEOUT,
     SHORT_TIMEOUT,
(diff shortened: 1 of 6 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1653,29 +1664,30 @@ def _test_broken_header_parsing(self, headers, unparsed_data_check=None):
...
+                    and isinstance(record.args, tuple)
                     and pool._absolute_url("/") == record.args[0]
                 ):
                     if (
                         unparsed_data_check is None
                         or unparsed_data_check in record.getMessage()
                     ):
                         return
-            self.fail("Missing log about unparsed headers")
+            self.fail("Missing log about unparsed headers")  # type: ignore[attr-defined]
```
bluetech, line 1675: This type error is actually correct -- this is probably a leftover from a unittest -> pytest migration, but pytest doesn't support `self.fail`. Can be replaced with `pytest.fail`, or just `assert False, "the message"`.

(There is another instance of this in this file)

bluetech (dismissed):
Looks good to me, just one comment!

test/with_dummyserver/test_socketlevel.py
hramezani: Thanks, @bluetech. Changed it to `pytest.fail`

pquentin (approved):

PR #2442, opened 2021-10-04
Bump requirements

pquentin:
I know coverage is not going to fix our codecov problems, but hey, it does not hurt to use the latest and greatest and encourage projects that drop Python 2.7.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,12 +1,12 @@
-coverage==5.5
+coverage==6.0
 tornado==6.1
 PySocks==1.7.1
-pytest==6.2.4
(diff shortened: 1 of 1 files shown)


hramezani:
I've added a separate [commit](https://github.com/urllib3/urllib3/pull/2436/commits/6c5ee09b3beab13eeecab7d8e13f60a7ff934be6) to https://github.com/urllib3/urllib3/pull/2436 to remove the `no-untyped-call type ignores`. I think we can merge this PR after we merged my PR.

hramezani (approved):

hramezani (approved):

PR #2443, opened 2021-10-05
Add type hints to test.test_ssltransport

hramezani:


test/test_ssltransport.py
```diff
@@ -37,24 +51,43 @@ def sample_request(binary=True):
...
+    ...
+
+
+@overload
+def sample_response(binary: "Literal[False]") -> str:
+    ...
+
+
+@overload
+def sample_response(binary: bool = False) -> Union[bytes, str]:
```
sethmlarson, line 73: Do we need this last `@overload` or is this covered by the function definition?

src/urllib3/util/ssltransport.py
```diff
@@ -100,7 +100,10 @@ def recv(self, buflen: int = 1024, flags: int = 0) -> Union[int, bytes]:
         return self._wrap_ssl_read(buflen)
 
     def recv_into(
```
sethmlarson, line 102: Hmmm, Python doesn't allow passing `None` for it's `buffer` argument of `socket.recv_into()`? We probably shouldn't allow this either? Thoughts on this @jalopezsilva?

sethmlarson (changes requested):
This is great, two things I saw:

test/test_ssltransport.py
hramezani: I think yes. otherwise we will get:
```
test/test_ssltransport.py:86: error: No overload variant of "sample_response" matches argument type "bool"  [call-overload]
test/test_ssltransport.py:86: note: Possible overload variants:
test/test_ssltransport.py:86: note:     def sample_response(binary: Literal[True] = ...) -> bytes
test/test_ssltransport.py:86: note:     def sample_response(binary: Literal[False]) -> str

```

test/test_ssltransport.py
```diff
@@ -27,7 +31,17 @@ def server_client_ssl_contexts():
     return server_context, client_context
 
 
-def sample_request(binary=True):
+@overload
+def sample_request(binary: "Literal[True]" = True) -> bytes:
```
bluetech, line 35: In overloads (and stub files) it is customary not to write `...` instead of the actual default values. This is because the default value itself is not meaningful at the type level, only that it exists, and this avoids duplication and is a bit shorter.

test/test_ssltransport.py
```diff
@@ -37,24 +51,43 @@ def sample_request(binary=True):
     return request if binary else request.decode("utf-8")
 
 
-def validate_request(provided_request, binary=True):
+def validate_request(
+    provided_request: bytearray, binary: Union["Literal[True]", "Literal[False]"] = True
```
bluetech, line 55: There's a shorthand for a union of literals:

```suggestion
    provided_request: bytearray, binary: "Literal[False, True]" = True
```

test/test_ssltransport.py
bluetech: I *think* that if you change here to `Literal[True, False]` it should work. That would mean that the function could only be called with a statically-known True or False, but since this is only used in tests, that seems like a good thing.

hramezani: @bluetech do you mean changing the last overload for the function? or changing `binary` variable type to `Literal[True, False]` in the function itself and removing the last overload?

bluetech: The latter. I'm not sure if it will work but worth a try.

hramezani: Thanks for the clarification. I've tried it but got the same error.
Let's keep it like this.

Issue #2439, opened 2021-10-01
Vendor ssl_match_hostname to avoid Python 3.10 deprecation warning


pquentin:
As reported by @AdamWill in https://github.com/urllib3/urllib3/issues/2200#issuecomment-931759912, we're issuing deprecation warnings for using `ssl_match_hostname` on Python 3.10:

```
/usr/lib/python3.10/site-packages/urllib3/connection.py:542: DeprecationWarning: ssl.match_hostname() is deprecated
  match_hostname(cert, asserted_hostname)
```

We should probably fix that as we're relying on users fixing 1.26.x deprecation warnings before moving to 2.0. In that case, the fix is vendoring `ssl_match_hostname` as `SSLContext` alternatives does not support all our needs. Look at this `main` commit for inspiration of what should be done in the `1.26.x` branch: https://github.com/urllib3/urllib3/pull/2198

Pradhvan:
I can give it a shot. 😄

AdamWill:
Thanks! I would've sent a PR but it was a bit tricky to figure out how to attack it with no prior knowledge of the project's architecture, sorry :)

pquentin:
This is up for grabs again :)

chetanugale:
@pquentin can I pick this issue?

sethmlarson:
@chetanugale It's yours!

sethmlarson:
@chetanugale @Pradhvan Because there's some time sensitivity to this PR I'd recommend both of you work on this and whoever has a PR that's ready to merge first will be merged. Not trying to turn this into a race, I just know that I'm also guilty of committing to an OSS contribution but not quite sure when it'll get done due to real life concerns.

Thank you both for being interested in this, it's very appreciated. Hope that either way you'll both be interested in contributing other issues as well :bow:

PR #2448, opened 2021-10-07
[1.26] Vendor ssl.match_hostname to avoid Python 3.10 deprecation warning

chetanugale:
Fixes : https://github.com/urllib3/urllib3/issues/2439


pquentin:
After discussing this privately with @sethmlarson, I ended using another strategy: I just ignore the F821 error in the line where it happens. Introducing a `six` import in a file will cause pain to distributions that will need to rewrite this line to `import six` as they don't use our six.py file. Let's avoid them the trouble.

PR #2441, opened 2021-10-03
[1.26] Don't unconditionally wrap proxy connection errors with HTTPSProxyError

sethmlarson:
This is a pretty simple cherry-pick, will likely fail on Python 2.7. Will update after failures.

Relates #2434

PR #2455, opened 2021-10-09
Use ProxyError to warn against HTTPS proxy issues

sethmlarson:
Created as an alternative to https://github.com/urllib3/urllib3/pull/2133. Discussed this with @jalopezsilva and I wasn't sure if adding an additional exception `HTTPSProxyError` captured anything extra over a normal `ProxyError`. `ProxyError` exists to differentiate between an error that occurred on the proxy versus the target/origin.

If merged would need to be backported to 1.26.x and would close https://github.com/urllib3/urllib3/pull/2441.

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -212,10 +212,10 @@ full visibility of your requests.
 
 .. _https_proxy_error_http_proxy:
 
-How to fix HTTPSProxyError?
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


sethmlarson:
After reading all of our proxy tunneling vs. forwarding code and where the connection to the proxy occurs I'm convinced we need more tests and to take an attempt at simplifying that logic in v2.0.

sethmlarson:
Closed in favor of https://github.com/urllib3/urllib3/pull/2482

PR #2458, opened 2021-10-11
Add type hints to test.test_response

hramezani:
There are mypy errors that I think we can ignore. Here is list of erros:
```
test/test_response.py:77: error: Argument 1 to "HTTPResponse" has incompatible type "None"; expected "Union[bytes, IO[Any], Iterable[bytes], str]"  [arg-type]
test/test_response.py:323: error: Incompatible types in assignment (expression has type "int", variable has type "Optional[HTTPResponse]")  [assignment]
test/test_response.py:353: error: Argument 1 to "BufferedReader" has incompatible type "HTTPResponse"; expected "RawIOBase"  [arg-type]
test/test_response.py:365: error: Argument 1 to "BufferedReader" has incompatible type "HTTPResponse"; expected "RawIOBase"  [arg-type]
test/test_response.py:370: error: Argument 1 to "BufferedReader" has incompatible type "HTTPResponse"; expected "RawIOBase"  [arg-type]
test/test_response.py:383: error: Argument 1 to "BufferedReader" has incompatible type "HTTPResponse"; expected "RawIOBase"  [arg-type]
test/test_response.py:679: error: Argument 1 to "HTTPResponse" has incompatible type "MockHTTPRequest"; expected "Union[bytes, IO[Any], Iterable[bytes], str]"  [arg-type]
test/test_response.py:814: error: Non-overlapping equality check (left operand type: "int", right operand type: "bytes")  [comparison-overlap]
```

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -810,6 +810,11 @@ class TestResponse:
             next(resp.read_chunked())
 
         orig_ex = ctx.value.args[1]
+        msg = (
+            "(\"Connection broken: InvalidChunkLength(got length b'ZZZ\\\\r\\\\n', 0 bytes read)\", "
+            "InvalidChunkLength(got length b'ZZZ\\r\\n', 0 bytes read))"
+        )
+        assert str(ctx.value) == msg
         assert isinstance(orig_ex, InvalidChunkLength)
         assert orig_ex.length == fp.BAD_LENGTH_LINE.encode()
 


sethmlarson:
Let's ignore `httplib.HTTPResponse.[chunked,chunk_left,length]` and open a PR on typeshed to add these types.

We can also ignore the `MockSock`.

For the `HTTPResponse` not being `IO[bytes]`, can we type it as `io.BytesIO` instead of `io.IOBase` since `.read()` always returns `bytes`? I wonder if we'll see more errors by doing this?

hramezani:
> Let's ignore `httplib.HTTPResponse.[chunked,chunk_left,length]`

Done!

> and open a PR on typeshed to add these types.

I will try to do it later

> For the HTTPResponse not being IO[bytes], can we type it as io.BytesIO instead of io.IOBase since .read() always returns bytes? I wonder if we'll see more errors by doing this?

I've changed it and it added a new ignore

sethmlarson:
> I've changed it and it added a new ignore and broke some tests. shall we revert the change?

Yes let's revert this and add type: ignores. We're so close to the IO API but not quite...

hramezani:
> This error looks legitimate? Can you dig into this one more? I think the rest can be all ignored.

It got fixed by changing `InvalidChunkLength.length` type to bytes

hramezani:
@sethmlarson Thanks for the review. I've reverted the changes and ignore other errors.

PR #2460, opened 2021-10-15
Convert non-relative imports

TimPansino:
## Overview

* Convert non-relative imports to match relative importing style of the rest of the source code.

## Context

When distributed inside other code bases, the relative importing is critical to making certain urllib3 uses the correct source code and doesn't accidentally import from another installed version of the package. This causes potential incompatibilities as well as resurfacing a compatibility issue with gevent for our codebase.


sethmlarson:
@TimPansino Ah! Would you be willing to make the fix to the 1.26.x branch as well?

PR #2462, opened 2021-10-17
Add type hits to test.test_connectionpool

hramezani:


bluetech:
> Yes, I've created the PR to share mypy errors with others. Yeah, I think we can ignore most of them.

Perhaps replace direct assignments e.g. `conn1.close = Mock()` with `mock.patch` calls? Then typing wouldn't complain.

PR #2470, opened 2021-10-22
Mention pool size when discarding connections

Streadz:
Note: Need to follow the rest of the [maintainer's guide](https://urllib3.readthedocs.io/en/latest/contributing.html) and then will get appropriate review.

I've encountered scenarios where developers are confused about this message because they believe their connection pool size is sufficiently large. In many cases, it turns out that they may not be properly reading their connection pool size settings file, and thus the default connection pool size is being used. I think adding some transparency to this message (by including the connection pool size) would help folks using the library unblock themselves more quickly.


Streadz:
@pquentin Let me know if all looks good from your end!

pquentin:
@Streadz It's not you, it's us. :) Haven't got time for a proper review yet, sorry. The 1.26.8 and 2.0 releases are taking a lot of maintainer time

Streadz:
hi @pquentin, just checking in -- are you still backlogged working on the up and coming releases?

test/test_connectionpool.py
```diff
@@ -35,6 +36,8 @@
 
 from .test_response import MockChunkedEncodingResponse, MockSock
 
+LOGGER = logging.getLogger(__name__)
+
```
pquentin, line 40: Are those two lines needed?

test/test_connectionpool.py
```diff
@@ -225,6 +228,32 @@ def test_max_connections(self) -> None:
...
+                    pool._put_conn(conn2)
+
+            assert conn1_close.called is False
+            assert conn2_close.called is True
+
+            assert conn1 == pool._get_conn()
+            assert conn2 != pool._get_conn()
+
+            # Primary objective is making sure the pool size is explicitly listed
+            assert "Connection pool size: 1" in caplog.text
```
pquentin, line 255: Can you please just modify the next test to add this line just after `assert conn2_close.called is True`? And remove this new test.

test/test_connectionpool.py
```diff
@@ -225,6 +228,32 @@ def test_max_connections(self) -> None:
 
             assert pool.num_connections == 1
 
+    def test_put_conn_message_when_pool_is_full_nonblocking(self, caplog: Any) -> None:
```
pquentin, line 231: Please use `pytest.LogCaptureFixture` instead

pquentin (changes requested):
Thanks for the ping, and sorry for the delay.

A few test improvements are needed, then we'll merge it. You will then have to backport this to the `1.26.x` branch, as `main` is for the upcoming 2.0 release

test/test_connectionpool.py
Streadz: Will do!

test/test_connectionpool.py
Streadz: Not if using the pytest line

test/test_connectionpool.py
Streadz: Sure thing! I originally made it separate because I thought you wanted everything to be completely decoupled (in terms of test purposes), but I agree it makes sense here

Streadz:
> Thanks for the ping, and sorry for the delay.
> 
> A few test improvements are needed, then we'll merge it. You will then have to backport this to the 1.26.x branch, as main is for the upcoming 2.0 release

Sounds good @pquentin, started backporting PR [here](https://github.com/urllib3/urllib3/pull/2497)

pquentin (approved):
Thanks! Right, in that case decoupling does not make sense, as in my opinion we're not testing something different.

Issue #2140, opened 2021-01-24
docs: Add example for cookies


shazow:
Would be great to have at least a basic example of doing cookies with urllib3 in our docs.

At minimum, https://urllib3.readthedocs.io/en/latest/search.html?q=cookies should yield some result.

kiwids0220:
import urllib3


http = urllib3.PoolManager()
cookies = {'Cookie':'cookie name, you need to change this=xxxxxxxxxxxxxxxx'}
r= http.request('METHOD, you need to change this', 
                'some URL, you need to change this'
                headers=cookies)
print(r.data)

#######
import urllib3


http = urllib3.PoolManager()
cookies = {'Cookie':'session=f3efe9db-ea4d-4c9c-b716-69dc51498002.ckQgoSDFg-r6YW3xMUiKNTbBNk4'}
r= http.request('GET', 
                'http://xxx.xxx.xxx.xxx/admin/challenges/new',
                headers=cookies)
print(r.data)


above is a cookie template format. and a real example.

daboshh:
Is it possible to get cookie data out of response, so that in the next request we can use it?

sethmlarson:
You'd have to manually do this with a cookiejar currently, our `HTTPResponse` is compatible to be used by the standard library cookiejar. urllib3 doesn't have helpers currently for this.

daboshh:
I think I figured out why I'm having a problem.
When connection is initiated, in the reply I receive
`Set-Cookie: JSESSIONID=xxxxxxxxxxxxxxxxxxxxxxxx`
On a redirect, server expects me me to have it, but because parameter is ignored and no set, the next packet is sent without it raising error: `(Caused by ResponseError('too many redirects'))`
Because session cannot be established.

I'm not sure if it's a bug or it's the way it is. I try cookiejar.
Maybe someone finds this useful.

PR #2474, opened 2021-10-27
Add cookiejar documentation

AlexanderDavid:
Added documentation for handling cookies with the standard library cookiejar #2140

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -253,10 +253,12 @@ Cookies provided by the server are stored in the ``Set-Cookie`` header:
     print(resp.headers["Set-Cookie"])
     # session=f3efe9db; Path=/
 
-Alternatively, the `cookiejar <https://docs.python.org/3/library/http.cookiejar.html>`_ class from
-the standard library can be used to manage cookies. This facilitates the handling of more complex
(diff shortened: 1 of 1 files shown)


pquentin:
For reference if someone wants to work on this next the text was:

```
You can also use the `cookiejar <https://docs.python.org/3/library/http.cookiejar.html>`_
library for more complex use cases. This library was created for use with ``urllib``,
so some slight concessions are required to integrate it with ``urllib3``. First, a request
object is needed for many of the functions. This request object isn't used for parsing
cookies, so a dummy ``urllib`` request object may be used. Second, the cookies need
to be explicitly included in the request's header.

.. code-block:: python
    import urllib
    import urllib3
    from http.cookiejar import CookieJar
    cj = CookieJar()
    # Dummy request object
    req = urllib.request.Request("http://httpbin.org/cookies/set/foo/bar")
    resp = urllib3.request("GET", "http://httpbin.org/cookies/set/foo/bar", redirect=False)
    cj.extract_cookies(resp, req)
    resp = urllib3.request("GET",
                           "http://httpbin.org/cookies",
                           headers={"Cookie": ";".join([c.name + "=" + c.value for c in cj])})
    print(resp.json())
    # {'cookies': {'foo': 'bar'}}
```

PR #2484, opened 2021-11-12
Test development versions of Python

pquentin:
Closes #2483

The original pull request had two issues:
 * "nightly" is not an actual deadsnakes version. Per https://launchpad.net/~deadsnakes/+archive/ubuntu/nightly, this PPA provides nightly builds for all Python versions. We could test 3.7-dev, for example. I removed the nightly action.
 * We need to tell nox about "3.11"

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,10 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: test-3.9
+          - python-version: 3.11-dev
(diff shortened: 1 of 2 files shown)


pquentin:
Running the test suite under 3.11 was way slower than other versions, that's surprising. (But still faster than PyPy.)

sethmlarson (approved):
Thanks!!

sethmlarson:
Regarding the slowness, I wonder if the builds aren't optimized? Not sure.

PR #2485, opened 2021-11-12
[WIP] Respect 'SSLContext.hostname_checks_common_name' when available and set

### LINKED ISSUE #2192: Respect SSLContext.hostname_checks_common_name set manually by users

sethmlarson:
Closes #2192

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -24,6 +24,7 @@ from typing import (
 
 if TYPE_CHECKING:
     from typing_extensions import Literal, NoReturn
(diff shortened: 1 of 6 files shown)


hramezani:
I've created a new PR and rebased your patch and fixed the mypy errors
https://github.com/urllib3/urllib3/pull/2518

sethmlarson:
Thanks @hramezani!

PR #2487, opened 2021-11-15
Keep full backtrace when raising LocationParseError in parse_url

elfprince13:
Recently ran into an issue where I needed to modify the system `url.py` to add this additional debug info and figured it made sense to make everyone else's lives easier as well.

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -413,8 +413,8 @@ def parse_url(url: str) -> Url:
         if normalize_uri and fragment:
             fragment = _encode_invalid_chars(fragment, _FRAGMENT_CHARS)
 
-    except (ValueError, AttributeError) as parse_exc:
-        raise LocationParseError(source_url) from parse_exc
+    except (ValueError, AttributeError) as e:
+        raise LocationParseError(source_url) from e
 
     # For the sake of backwards compatibility we put empty
     # string values for path if there are any defined values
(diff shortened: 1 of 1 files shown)


src/urllib3/util/url.py
```diff
@@ -413,8 +413,8 @@ def parse_url(url: str) -> Url:
         if normalize_uri and fragment:
             fragment = _encode_invalid_chars(fragment, _FRAGMENT_CHARS)
 
-    except (ValueError, AttributeError):
-        raise LocationParseError(source_url) from None
+    except (ValueError, AttributeError) as parse_exc:
```
sethmlarson, line 416: I think this makes sense, could we change this to `as e:` to match elsewhere in the codebase?

Also could you give an example backtrace of what this would look like in action?

sethmlarson (dismissed):

src/urllib3/util/url.py
elfprince13:
```python3
>>> urllib3.util.url.parse_url('http://foo:1000000000')
  File "/usr/lib/python3/dist-packages/urllib3/util/url.py", line 368, in parse_url
    auth, host, port = SUBAUTHORITY_RE.match(authority).groups()
AttributeError: 'NoneType' object has no attribute 'groups'

  File "/usr/lib/python3/dist-packages/urllib3/util/url.py", line 392, in parse_url
    raise LocationParseError(source_url) from parse_exc
urllib3.exceptions.LocationParseError: Failed to parse: http://foo:1000000000
>>> 
```

elfprince13: (requested change made in 72ee1db)

pquentin (approved):

PR #2490, opened 2021-11-18
Try removing Codecov

sethmlarson:
From https://hynek.me/til/ditch-codecov-python, let's try to get rid of Codecov.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,12 +8,7 @@ omit =
     *urllib3/contrib/securetransport.py
     *urllib3/contrib/_securetransport/*
 
-[paths]
-source =
-    src/urllib3
-    .nox/*/lib/python*/site-packages/urllib3
-    .nox\*\Lib\site-packages\urllib3
-    .nox/*/site-packages/urllib3
+relative_files = True
 
 [report]
 exclude_lines =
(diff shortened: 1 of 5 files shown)


nedbat:
This [works if you specify paths to combine](https://github.com/nedbat/urllib3/actions/runs/1567764239):
```
[paths]
source =
    src/urllib3
    */urllib3
    *\urllib3
```

TBH, relative_files is still kind of spotty.

nedbat:
I don't understand why this is needed: https://github.com/urllib3/urllib3/pull/2490/files#diff-fad2fa83c042cce646b2c1a0455e9b60c22dd5f105ed8859ce85dcf361c07fbdR11-R13

sethmlarson:
@nedbat Thanks for commenting here, that was my attempt at making `.coverage` files names not collide when uploaded as an artifact to GitHub Actions.

nedbat:
If you use `parallel = True` in the .coveragerc, they will get unique names anyway, so you shouldn't need that.

sethmlarson:
@nedbat Right, I tried implementing without the parallel runs because I wasn't sure if that was messing with `relative_path` somehow but gave up on that train of thought, forgot to remove the random token though!

sethmlarson:
Success! Thanks @nedbat for your help with this!

PR #2493, opened 2021-11-24
Cache installed dependencies in CI

sethmlarson:
See: https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/


pquentin (approved):
Thanks! Awesome to see how easy it is to enable now.

Issue #1445, opened 2018-09-19
Feature Request: Pass Response to `Retry.is_retry()`


Mr0grog:
I’m working on some tools for using the [Internet Archive’s Wayback](https://web.archive.org) Machine APIs and it’d be really helpful if I could write a custom `Retry` class that can look at the the response headers to determine whether a retry is warranted. In particular, when requesting the snapshot of a URL, it’s possible to get an error code that might be coming from the wayback machine or might be coming from the snapshot itself (i.e. the original URL returned the error code when it was scraped). You can determine which of these situations you’re dealing with by looking for the `Memento-Datetime` header, but [`Retry.is_retry()`](https://github.com/urllib3/urllib3/blob/f4efcca559cc753f878e0a73e28dd6e795d0d052/src/urllib3/util/retry.py#L304-L318), which seems like the right place to do this, doesn’t have access to that info.

It feels like there ought to be plenty of other odd cases where more info about the response would be helpful, so I’m wondering if passing it directly to `Retry.is_retry()` would be a good idea. Since this gets called before `Retry.increment()` adds request info to the history, passing the URL might be nice, too.

I’d love to change this signature:

```py
def is_retry(self, method, status_code, has_retry_after=False):
```

to:

```py
def is_retry(self, method, url, response):
```

(`status_code` and `has_retry_after` can both be obtained from the `response` object, and can even be handled slightly more efficiently by not asking for them before they’re needed, since `is_retry()` returns early in some cases.)

SethMichaelLarson:
Unfortunately I'm doubtful we'll be able to change that method to remove the `status_code` in the signature. Adding `response` is more likely but it would break users that subclass the `Retry` object, I'm unsure of how often that occurs.

An easier way to accomplish this for you may be to use the Requests library along with some request-toolbelt code to handle a custom retry/redirect?

Mr0grog:
Hmmm, it doesn’t seem like there’s anything in toolbelt for retrying or that would cover a case like this. It also feels a bit hacky to wrap my retry/backoff logic *around* `urlopen()` (or `requests.Session.send()`) when it feels like this *ought* to fit neatly in the scope of the `Retry` class. I do understand the need not to make a breaking change to public API, though.

Would it be reasonable to add `url` and `response` as optional kwargs to the end? I suppose it would be kind of ugly and involve including data that the default implementation never needs, but it would make this use case possible.

---

[...]

PR #2500, opened 2021-12-11
Allow Retry class to determine whether to retry based on response

EnricoMi:
The `Retry` class allows library users to implement their own retry logic. The library calls into `Retry.is_retry` to determine if a request is to be retried based on the status code. Sometimes the status code is not sufficient to identify a retry-able error (#1445). The actual response, which includes the status code, the reason, response headers and payload, among others, is sometimes needed.

Replacing the `is_retry` method with `is_response_retry` allows users to get hold of the actual response. This change keeps `is_retry` to avoid breaking existing code.


EnricoMi:
@sethmlarson to be clear: this is not intended for 2.0 but the next 1.x release. So I presume this should target branch `1.26.x`, right?

sethmlarson:
@EnricoMi We're not adding new features to 1.x as there won't be a 1.27 release. All new features must go into 2.0.

EnricoMi:
So no quick fix. And a non-breaking fix is not required. What is the time horizon of the 2.0 release.

sethmlarson:
@EnricoMi We don't have a date in mind, we're trying to finish [these blocking issues](https://github.com/urllib3/urllib3/milestone/6) before a release though.

changelog/1445.feature.rst
```diff
@@ -1 +1 @@
-Added is_response_retry variation of Retry.is_retry that considers `HTTPResponse`.
\ No newline at end of file
+Added `is_response_retry` variation of `Retry.is_retry` that considers `HTTPResponse`.
```
sigmavirus24, line 1:
```suggestion
Added ``is_response_retry`` variation of ``Retry.is_retry`` that considers ``HTTPResponse``.
```

reStructuredText doesn't use the same pre-formatted identifier as Markdown. It uses two backticks

EnricoMi: Thanks!

EnricoMi:
Should `is_retry` be marked as deprecated? Or shall both methods co-exist in the future?

JannKleen:
@EnricoMi is there any update on this? I need my `Retry`-class to support a non-standard Retry header and having this merged would make life easier ;)

ollie-bell:
Now that 2.0 is in pre-release, is this feature going to be available soon? Thanks!

EnricoMi:
@sethmlarson rebased with latest `main`. Is this now good to go?

EnricoMi:
This is sad...

skydiver1980 (approved):

JannKleen:
@sethmlarson as someone interested in using it, is there anything missing in order to get this merged?

ecerulm:
@EnricoMi can you resolve the merge conflict?

EnricoMi:
@ecerulm conflicts resolved

D-stefaang:
https://github.com/urllib3/urllib3/pull/3233 introduces a cap to the `Retry-After` header.
We've seen servers returning 429s with `Retry-After=3600` which is way more than the 2 minutes max backoff.
Such responses should be marked immediately as non-retriable... and this PR would allow us to do so.

PR #2506, opened 2021-12-16
Update mypy to 0.920

hramezani:
we still have some mypy errors because of `IncompleteRead.partial` and  `InvalidChunkLength.partial`. the `httplib.IncompleteRead.partial` type is `bytes` but we defined as `int`
```
src/urllib3/exceptions.py:293: error: Incompatible types in assignment (expression has type "int", variable has type "bytes")  [assignment]
src/urllib3/exceptions.py:297: error: Incompatible types in string interpolation (expression has type "bytes", placeholder has type "Union[int, float, SupportsInt]")  [str-format]
src/urllib3/exceptions.py:297: error: Incompatible types in string interpolation (expression has type "Optional[int]", placeholder has type "Union[int, float, SupportsInt]")  [str-format]
src/urllib3/exceptions.py:307: error: Incompatible types in assignment (expression has type "int", base class "IncompleteRead" defined the type as "bytes")  [assignment]
test/test_response.py:780: error: Non-overlapping equality check (left operand type: "bytes", right operand type: "Literal[0]")  [comparison-overlap]
```

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -134,7 +134,7 @@ class TestingApp(RequestHandler):
 
     def source_address(self, request: httputil.HTTPServerRequest) -> Response:
         """Return the requester's IP address."""
-        return Response(request.remote_ip)
+        return Response(request.remote_ip)  # type: ignore[arg-type]
 
     def set_up(self, request: httputil.HTTPServerRequest) -> Response:
         params = request_params(request)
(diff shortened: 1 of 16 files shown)


sethmlarson:
We can ignore all the errors related to IncompleteRead, does that covers all the errors?

hramezani:
> We can ignore all the errors related to partial types, does that covers all the errors?

Yes, I added a commit and ignored them. now we are green again!

sethmlarson (approved):
This is excellent, thanks for staying on top of Mypy updates!

PR #2509, opened 2021-12-30
Allow setuptools to install towncrier and fix our CI

sigmavirus24:
CI is presently broken trying on install towncrier: https://github.com/urllib3/urllib3/runs/4667770822?check_suite_focus=true


sethmlarson (approved):
LGTM, thanks for this!

PR #2511, opened 2021-12-30
Update CI workflow for the new year

sigmavirus24:
We don't need to set things up twice in a row (which we were doing in
restoring the cache) and let's use pypa/build instead of python setup.py
since that's better supported and the future


shazow (commented):
Looks good to me, but I'll let Seth approve since I don't have any experience with modern Python CI stuff. :)

sethmlarson (approved):
LGTM, thank you!

Issue #2291, opened 2021-06-21
[v2] Use httplib's chunked encoding support?


bluetech:
`urllib3.HTTPConnection` has a method [`request_chunked`](https://github.com/urllib3/urllib3/blob/89ca490bfc15b7b4762676b5b51bf57e8912d80f/src/urllib3/connection.py#L287) which sends the body with chunked encoding. [Since Python 3.6](https://github.com/python/cpython/commit/3c0d0baf2badfad7deb346d1043f7d83bb92691f), httplib supports this directly, with very similar-looking code. Since v2 now requires Python 3.6, I was wondering if we might want to use this.

I haven't checked if there are any gotchas there, this is more of a placeholder issue since I couldn't find an existing issue discussing this.

sethmlarson:
Kind-of related to this, I was wondering if we should instead add a `chunked` parameter to `HTTPConnection.request()` and shim `.request_chunked()` to simply call `return self.request(..., chunked=True)`. Reasoning being that the rest of our API looks like this, seems more natural to have `.request(..., chunked=True)` everywhere instead of at every level until you get to `HTTPConnection` and suddenly a shift to two different methods?

sethmlarson:
The above change would also bring us closer to the stdlib's implementation but keep things urllib3-flavored, we're already married to the `chunked` parameter (and I like it better than `encode_chunked` tbh)

PR #2516, opened 2022-01-03
Add test cases for framing of request bodies

sethmlarson:
Some partial work on https://github.com/urllib3/urllib3/issues/2291, wanted to start by adding some test cases for the existing request body framing behaviors.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1,6 +1,7 @@
 # TODO: Break this module up into pieces. Maybe group by functionality tested
 # rather than the socket level-ness of it.
 import errno
+import io
 import os
 import os.path
 import select
@@ -8,6 +9,7 @@ import shutil
 import socket
 import ssl
 import tempfile
+import time
 from collections import OrderedDict
 from pathlib import Path
 from test import (
@@ -19,7 +21,7 @@ from test import (
     resolvesLocalhostFQDN,
 )
 from threading import Event
-from typing import Any, List, Optional
+from typing import Any, Generator, List, Optional
 from typing import OrderedDict as OrderedDictType
 from typing import Tuple, Union
 from unittest import mock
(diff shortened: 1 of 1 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1930,7 +1932,7 @@ def socket_handler(listener: socket.socket) -> None:
         from urllib3.connectionpool import log
 
         with mock.patch.object(log, "warning") as log_warning:
-            with HTTPConnectionPool(self.host, self.port, timeout=3) as pool:
+            with HTTPConnectionPool(self.host, self.port, timeout=0.5) as pool:
```
pquentin, line 1935: We have multiple tests with timeout=3 here, but we should probably be using LONG_TIMEOUT instead, which is longer in CI.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1939,3 +1941,163 @@ def socket_handler(listener: socket.socket) -> None:
...
+
+        with HTTPConnectionPool(self.host, self.port, timeout=3, retries=False) as pool:
+            resp = pool.request(method, "/", chunked=chunked, body=body)
+            assert resp.status == 200
+
+        sent_bytes = bytes(buffer)
+        assert sent_bytes.count(b":") == 5
+        assert b"Host: localhost:" in sent_bytes
+        assert b"Accept-Encoding: identity\r\n" in sent_bytes
+        assert b"Transfer-Encoding: chunked\r\n" in sent_bytes
```
pquentin, line 2033: Why is this true when `chunked` is false?

pquentin (changes requested):
Fantastic, thanks! Two questions

test/with_dummyserver/test_socketlevel.py
sethmlarson: `chunked=True` essentially means "force chunked" and `chunked=False` (for both urllib3 and stdlib) means use `content-length` if you can find the length of `body` otherwise use `transfer-encoding: chunked`.

pquentin (approved):
Thanks!

PR #2517, opened 2022-01-03
Add missing brackets to __get__ bound method call in HTTPResponse.closed

jpy-git:
We're doing some work in typeshed to update the typing of function attributes and spotted that urllib3 uses `__get__` to get a bound method but is missing the brackets to call that method, hence why mypy is complaining that it expects `bool` but gets `MethodType`: https://github.com/python/typeshed/pull/6804#issuecomment-1004431743

I removed those type ignores to see if they're still needed after this change (can re-add them if needed) 😄

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -770,7 +770,7 @@ class HTTPResponse(BaseHTTPResponse):
     @property
     def closed(self) -> bool:
         if not self.auto_close:
-            return io.IOBase.closed.__get__(self)  # type: ignore[no-any-return, attr-defined]
+            return io.IOBase.closed.__get__(self)()
         elif self._fp is None:
             return True
         elif hasattr(self._fp, "isclosed"):


jpy-git:
Nevermind I stand corrected:
```
  File "/Users/runner/work/urllib3/urllib3/test/test_response.py", line 384, in test_io_not_autoclose_bufferedreader
    assert list(reader) == [b"hello\n", b"world"]
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-7/lib/python3.7/site-packages/urllib3/response.py", line 773, in closed
    return io.IOBase.closed.__get__(self)()
TypeError: 'bool' object is not callable
```
Will take a closer look 😄

jpy-git:
Ahh this highlights that the `MethodType` in typeshed doesn't account for @property, will close this as there's no issue here 👍

PR #2518, opened 2022-01-04
Respect 'SSLContext.hostname_checks_common_name' when available and set

### LINKED ISSUE #2192: Respect SSLContext.hostname_checks_common_name set manually by users

### LINKED ISSUE #2486: Try our test suite against PyPy 7.3.6 (Python 3.8 beta)
We currently test against pypy-3.7 but last month PyPy 7.3.6 added beta support for Python 3.8. @pquentin psubmitted an issue](https://foss.heptapod.net/pypy/pypy/-/issues/3539) that was fixed and released in this beta version, we should test the beta version before a stable release to see if the fix was successful.

We don't necessarily need to run this test suite on every PR yet (although that is welcomed, not sure how feasible it is though?)

hramezani:
Closes #2192
Closes #2486
Continue to https://github.com/urllib3/urllib3/pull/2485

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -25,6 +25,8 @@ from typing import (
 if TYPE_CHECKING:
     from typing_extensions import Literal
 
+    from .util.ssl_ import _TYPE_PEER_CERT_RET_DICT
+
 from .util.proxy import create_proxy_ssl_context
 from .util.timeout import _DEFAULT_TIMEOUT, _TYPE_TIMEOUT, Timeout
 from .util.util import to_bytes, to_str
@@ -50,7 +52,6 @@ from .exceptions import (
 )
 from .util import SKIP_HEADER, SKIPPABLE_HEADERS, connection, ssl_
 from .util.ssl_ import (
-    _TYPE_PEER_CERT_RET,
     assert_fingerprint,
     create_urllib3_context,
     resolve_cert_reqs,
@@ -537,8 +538,23 @@ class HTTPSConnection(HTTPConnection):
             and not context.check_hostname
             and self.assert_hostname is not False
         ):
-            cert = self.sock.getpeercert()
-            _match_hostname(cert, self.assert_hostname or server_hostname)
+            cert: "_TYPE_PEER_CERT_RET_DICT" = self.sock.getpeercert()  # type: ignore[assignment]
+
+            # Need to signal to our match_hostname whether to use 'commonName' or not.
+            # If we're using our own constructed SSLContext we explicitly set 'False'
+            # because PyPy hard-codes 'True' from SSLContext.hostname_checks_common_name.
+            if default_ssl_context:
+                hostname_checks_common_name = False
+            else:
+                hostname_checks_common_name = (
+                    getattr(context, "hostname_checks_common_name", False) or False
+                )
+
+            _match_hostname(
+                cert,
+                self.assert_hostname or server_hostname,
+                hostname_checks_common_name,
+            )
 
         self.is_verified = context.verify_mode == ssl.CERT_REQUIRED or bool(
             self.assert_fingerprint
@@ -583,9 +599,13 @@ class HTTPSConnection(HTTPConnection):
         )
 
 
-def _match_hostname(cert: _TYPE_PEER_CERT_RET, asserted_hostname: str) -> None:
+def _match_hostname(
+    cert: Optional["_TYPE_PEER_CERT_RET_DICT"],
+    asserted_hostname: str,
+    hostname_checks_common_name: bool = False,
+) -> None:
     try:
-        match_hostname(cert, asserted_hostname)
+        match_hostname(cert, asserted_hostname, hostname_checks_common_name)
     except CertificateError as e:
         log.warning(
             "Certificate did not match expected hostname: %s. Certificate: %s",
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -100,6 +100,18 @@ def no_san_server(
         yield cfg
 
 
+@pytest.fixture()
+def no_san_server_with_different_commmon_name(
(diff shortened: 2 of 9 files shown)


sethmlarson:
The failures look legitimate, I'm curious how 3.7+macOS is passing and Windows/Linux 3.7 are failing.

hramezani:
> he failures look legitimate, I'm curious how 3.7+macOS is passing and Windows/Linux 3.7 are failing.
Yes, it should fail on `3.7+macOS` as well.

BTW, I added a commit to fix the failing test. not sure we have to fix the test or change the code to pass the test :thinking: ?

hramezani:
@sethmlarson I've added another test to cover this line https://github.com/urllib3/urllib3/pull/2518/files#diff-23ca5b18a566c65407a9b8502ad3037867e6cedf50f464e386e79bf506e84851R135

src/urllib3/connection.py
```diff
@@ -488,17 +489,18 @@ def connect(self) -> None:
...
-                self.assert_fingerprint
-                or self.assert_hostname
-                # We still support OpenSSL 1.0.2, which prevents us from verifying
-                # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
-                or ssl_.IS_PYOPENSSL
-                or not ssl_.HAS_NEVER_CHECK_COMMON_NAME
-            ):
-                self.ssl_context.check_hostname = False
+
+        # In some cases, we want to verify hostnames ourselves
```
sethmlarson, line 493: I don't think we want to modify every `ssl_context` we get passed, only the default one that we create ourselves?

test/conftest.py
```diff
@@ -100,6 +100,18 @@ def no_san_server(
         yield cfg
 
 
+@pytest.fixture()
+def no_san_server_with_diffrent_commmon_name(
```
sethmlarson, line 104: :star_struck:

test/with_dummyserver/test_https.py
```diff
@@ -974,11 +975,18 @@ def test_can_validate_san(self, san_server: ServerConfig) -> None:
             assert r.status == 200
 
     def test_common_name_without_san_fails(self, no_san_server: ServerConfig) -> None:
```
sethmlarson, line 977: This test should continue to pass without the modifications IMO, since our default ssl_context should have `hostname_checks_common_name = False` set.

test/with_dummyserver/test_https.py
```diff
@@ -988,6 +996,68 @@ def test_common_name_without_san_fails(self, no_san_server: ServerConfig) -> Non
                 e.value
             ) or "no appropriate subjectAltName" in str(e.value)
 
+    def test_common_name_without_san_with_differen_common_name(
```
sethmlarson, line 999: nit: `different`*

test/with_dummyserver/test_https.py
```diff
@@ -988,6 +996,68 @@ def test_common_name_without_san_fails(self, no_san_server: ServerConfig) -> Non
...
+        self, no_san_server_with_diffrent_commmon_name: ServerConfig
+    ) -> None:
+        ctx = urllib3.util.ssl_.create_urllib3_context()
```
sethmlarson, line 1002: These test cases are excellent, nice work :muscle:

sethmlarson (changes requested):
Thanks for this, here's a couple comments for you:

test/with_dummyserver/test_https.py
hramezani: I've changed the code to pass this test.
Now, this test fails on `pypy` because [_is_bpo_43522_fixed](https://github.com/urllib3/urllib3/blob/49848a178c1ad108d311bdf904bd63aa3fd593e0/src/urllib3/util/ssl_.py#L59:L68) just works for `cpython`.
Shall we include `pypy` in this function?

hramezani:
Thanks @sethmlarson for the review.
I've addressed them and left a [comment](https://github.com/urllib3/urllib3/pull/2518#discussion_r781560121)

test/with_dummyserver/test_https.py
sethmlarson: Hmm, yeah can we figure out which versions of PyPy have that fix and go off of that? I guess I'd prefer being conservative and returning `False` from that function anywhere where we're not **certain** that setting `SSLContext.hostname_checks_common_name = False` works.

Can we somehow create a unit test for our function to see that the value of `urllib3.util.ssl_.HAS_NEVER_CHECK_COMMON_NAME` is the same as what `SSLContext.hostname_checks_common_name`'s behavior is?

This might be a thing we can do outside this PR, but for now let's return `False` for PyPy and lean on our own hostname checking implementation.

test/conftest.py
sethmlarson, line 104: nit: different*

test/with_dummyserver/test_https.py
pquentin: For what it's worth, I've intentionally checked for CPython only in `_is_bpo_43522_fixed`, because on PyPy, `ssl.HAS_NEVER_CHECK_COMMON_NAME` is supposed to always be reliable when present, see https://github.com/urllib3/urllib3/issues/2191#issuecomment-902618341 and https://foss.heptapod.net/pypy/pypy/-/issues/3539.

sethmlarson: Okay so we can leave the implementation as-is, thanks Quentin!

sethmlarson (dismissed):
Looks awesome, thanks @hramezani!

sethmlarson:
Thanks @hramezani for this PR :muscle:

PR #2519, opened 2022-01-04
Add sentinel type for _FAILEDTELL

### LINKED ISSUE #2512: Change set_file_position() type to use proper sentinel with _FAILEDTELL
Currently `_FAILEDTELL` in `urllib3.util.request` uses `object` as a sentinel type, we should give this object the same treatment as `_DEFAULT_TIMEOUT` and switch to a `Final[Enum]`. This type should be used wherever `body_pos` is used.

cc @hramezani

imkaka:
Fixes #2512 

Took the idea from #2473, Please let me know if it needs improvements.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -41,7 +41,7 @@ from .exceptions import (
 from .response import BaseHTTPResponse, HTTPResponse
 from .util.connection import is_connection_dropped
 from .util.proxy import connection_requires_http_tunnel
-from .util.request import _TYPE_FAILEDTELL, set_file_position
+from .util.request import _TYPE_BODY_POSITION, set_file_position
 from .util.response import assert_header_parsing
 from .util.retry import Retry
 from .util.ssl_match_hostname import CertificateError
@@ -547,7 +547,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         pool_timeout: Optional[int] = None,
         release_conn: Optional[bool] = None,
         chunked: bool = False,
-        body_pos: _TYPE_FAILEDTELL = None,
+        body_pos: _TYPE_BODY_POSITION = None,
         **response_kw: Any,
     ) -> BaseHTTPResponse:
         """
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -547,7 +547,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         pool_timeout: Optional[int] = None,
         release_conn: Optional[bool] = None,
         chunked: bool = False,
-        body_pos: _TYPE_BODY_POSITION = None,
+        body_pos: Optional[_TYPE_BODY_POSITION] = None,
         **response_kw: Any,
     ) -> BaseHTTPResponse:
         """
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -41,7 +41,7 @@ from .exceptions import (
 from .response import BaseHTTPResponse, HTTPResponse
 from .util.connection import is_connection_dropped
 from .util.proxy import connection_requires_http_tunnel
-from .util.request import set_file_position
+from .util.request import _TYPE_FAILEDTELL, set_file_position
 from .util.response import assert_header_parsing
 from .util.retry import Retry
 from .util.ssl_match_hostname import CertificateError
@@ -547,7 +547,7 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         pool_timeout: Optional[int] = None,
         release_conn: Optional[bool] = None,
         chunked: bool = False,
-        body_pos: Optional[Union[int, object]] = None,
+        body_pos: _TYPE_FAILEDTELL = None,
         **response_kw: Any,
     ) -> BaseHTTPResponse:
         """
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -26,13 +26,13 @@ else:
     ACCEPT_ENCODING += ",br"
 
 
-class _TYPE_DEFAULT(Enum):
+class _TYPE_FAILEDTELL(Enum):
     token = 0
 
 
-_FAILEDTELL: "Final[_TYPE_DEFAULT]" = _TYPE_DEFAULT.token
+_FAILEDTELL: "Final[_TYPE_FAILEDTELL]" = _TYPE_FAILEDTELL.token
 
-_TYPE_FAILEDTELL = Optional[Union[int, _TYPE_DEFAULT]]
+_TYPE_BODY_POSITION = Optional[Union[int, _TYPE_FAILEDTELL]]
 
 
 def make_headers(
@@ -114,7 +114,7 @@ def make_headers(
     return headers
 
 
-def set_file_position(body: Any, pos: _TYPE_FAILEDTELL) -> _TYPE_FAILEDTELL:
+def set_file_position(body: Any, pos: _TYPE_BODY_POSITION) -> _TYPE_BODY_POSITION:
     """
     If a position is provided, move file to that point.
     Otherwise, we'll attempt to record a position for future use.
@@ -132,7 +132,7 @@ def set_file_position(body: Any, pos: _TYPE_FAILEDTELL) -> _TYPE_FAILEDTELL:
     return pos
 
 
-def rewind_body(body: IO[AnyStr], body_pos: _TYPE_FAILEDTELL) -> None:
+def rewind_body(body: IO[AnyStr], body_pos: _TYPE_BODY_POSITION) -> None:
     """
     Attempt to rewind body to a certain position.
     Primarily used for request redirects and retries.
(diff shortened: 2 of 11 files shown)


test/test_util.py
```diff
@@ -531,8 +531,6 @@ def test_rewind_body_bad_position(self) -> None:
         # Pass non-integer position
         with pytest.raises(ValueError):
             rewind_body(body, body_pos=None)
-        with pytest.raises(ValueError):
-            rewind_body(body, body_pos=object())
```
nateprewitt, line 535: I think we want to keep this test, just change what's passed to `rewind_body`. It would be `_FAILEDTELL` in this case.

src/urllib3/util/request.py
```diff
@@ -123,7 +132,7 @@ def set_file_position(
     return pos
 
 
-def rewind_body(body: IO[AnyStr], body_pos: Optional[Union[int, object]]) -> None:
+def rewind_body(body: IO[AnyStr], body_pos: _TYPE_FAILEDTELL) -> None:
```
nateprewitt, line 135: I think we may still want the `Union` here and replace the generic `object` with the type of the new sentinel type.

src/urllib3/connectionpool.py
```diff
@@ -547,7 +547,7 @@ def urlopen(  # type: ignore[override]
         pool_timeout: Optional[int] = None,
         release_conn: Optional[bool] = None,
         chunked: bool = False,
-        body_pos: Optional[Union[int, object]] = None,
+        body_pos: _TYPE_FAILEDTELL = None,
```
nateprewitt, line 550: I _think_ we may want to keep the `Union` here as well. If we're going to define a variable for the composed type, the name probably needs to be more specific to `body_pos` rather than _FAILEDTELL. That's a side effect that we should try to minimize in the public typing. It's never intended for a user to pass this.

nateprewitt (changes requested):
Thanks for the PR, @imkaka! I think we're off to a great start. I left a few comments about naming and type composition. Please let me know if you have thoughts or questions about them.

src/urllib3/connectionpool.py
nateprewitt: I read through the other timeout PR and think I have a better grasp on what was done there. I'll enumerate the concepts we want to have defined with this change.

`_FAILEDTELL`: This is a sentinel value we created to denote we'd tried to get the position of our body object by calling `tell()` but it failed for whatever reason. It should never be passed as an argument by the user.

`_TYPE_FAILEDTELL`: This is the type of the sentinel object itself (_FAILEDTELL). I'm not sure if we want to conflate defaults with this concept. It may be more accurate to use either `_TYPE_SENTINEL` or `_TYPE_FAILEDTELL` in place of `_TYPE_DEFAULT`.

For how we handle this with `body_pos`, I think the type should be something closer to `Optional[Union[int, _TYPE_FAILEDTELL]]`. We can either use this as a standalone value or name it something like `_TYPE_BODY_POSITION`. I don't think we want to type any public interfaces with `_TYPE_FAILEDTELL` to avoid confusion.

test/test_util.py
imkaka: `rewind_body` does not support the `object()` type anymore, so the test will pass but `mypy` will cry so removed this.

src/urllib3/connectionpool.py
imkaka: yes, this makes sense will rename based on this.

test/test_util.py
nateprewitt: It never supported `object()` but the purpose of this test was to ensure a generic "sentinel-like" object would cause a value error rather than being interpreted as `_FAILEDTELL`. Perhaps it's a little less relevant now, but probably still be validated.

It also points to another interesting issue. The typing of `rewind_body` is wrong and appears to have been adjusted to appease these tests. `body_pos` should not be Optional on `rewind_body`, so we may need to fix that as well.

imkaka: What do you suggest here then @nateprewitt ?

nateprewitt: If we're running `mypy` over the test suite, I'll defer to @SethMLarson, @pquentin on how they want to handle it. It does seem like there should be a way to support testing unsupported types without creating incorrect type signatures.

My proposal would be to determine how we can support this with mypy, update the signature on `rewind_body` to remove the `Optional` definition, and leave this test untouched.

imkaka: > If we're running `mypy` over the test suite, I'll defer to @sethmlarson, @pquentin on how they want to handle it. It does seem like there should be a way to support testing unsupported types without creating incorrect type signatures.

Okey


Removing `Optional` won't solve the `mypy` error as `object()` is not supported in `Union[int, _TYPE_FAILEDTELL]`.

Going ahead with that will impact this test case instead.

```py

with pytest.raises(ValueError):
    rewind_body(body, body_pos=None)
```

nateprewitt: Correct, I was noting the `None` test case above _appears_ to be what introduced Optional to `rewind_body` in the first place.

Both test cases should be supported without having to mutilate the typing. I'll need to do some more reading on what our options are for this.

nateprewitt: It may be as simple as using `# type: ignore` in these two test calls. I'll defer to the other contributors on if that's acceptable as I haven't been involved in previous typing discussions.

sethmlarson: Definitely keep the test cases and use `# type: ignore[arg-type]` to make Mypy happy (or whatever Mypy says is the error code).

src/urllib3/util/request.py
```diff
@@ -21,7 +25,14 @@
...
 
-_FAILEDTELL = object()
+
+class _TYPE_FAILEDTELL(Enum):
+    token = 0
+
+
+_FAILEDTELL: "Final[_TYPE_FAILEDTELL]" = _TYPE_FAILEDTELL.token
+
+_TYPE_BODY_POSITION = Optional[Union[int, _TYPE_FAILEDTELL]]
```
nateprewitt, line 35: Apologies for my earlier comment, I misspoke. This should be `Union[int, _TYPE_FAILEDTELL]`. This allows us to accurately use the type in places where it isn’t actually optional.

src/urllib3/connectionpool.py
```diff
@@ -547,7 +547,7 @@ def urlopen(  # type: ignore[override]
         pool_timeout: Optional[int] = None,
         release_conn: Optional[bool] = None,
         chunked: bool = False,
-        body_pos: Optional[Union[int, object]] = None,
+        body_pos: _TYPE_BODY_POSITION = None,
```
nateprewitt, line 550: urlopen should never be taking a `_FAILEDTELL` so this is more accurately `Optional[int]`.

src/urllib3/util/request.py
```diff
@@ -103,9 +114,7 @@ def make_headers(
     return headers
 
 
-def set_file_position(
-    body: Any, pos: Optional[Union[int, object]]
-) -> Optional[Union[int, object]]:
+def set_file_position(body: Any, pos: _TYPE_BODY_POSITION) -> _TYPE_BODY_POSITION:
```
nateprewitt, line 117: This would now be `Optional[_TYPE_BODY_POSITION]`. It returns a non-Optional `_TYPE_BODY_POSITION`.

src/urllib3/util/request.py
```diff
@@ -123,7 +132,7 @@ def set_file_position(
     return pos
 
 
-def rewind_body(body: IO[AnyStr], body_pos: Optional[Union[int, object]]) -> None:
+def rewind_body(body: IO[AnyStr], body_pos: _TYPE_BODY_POSITION) -> None:
```
nateprewitt, line 135: This would remain the same, but the None test would also need the ignore clause.

sethmlarson (dismissed):
Looks good to me, will let @nateprewitt do final review+merge.

nateprewitt (commented):
Thanks @imkaka! I think the original goal of adding the new sentinel value is done. 

I’m still seeing some typing issues that need to be worked out. I’ve left comments if you’d like to fix those, otherwise, I’m happy with the change and can address the typing in a follow up PR. Let me know if you have a preference.

imkaka:
> I’m still seeing some typing issues that need to be worked out. I’ve left comments if you’d like to fix those, otherwise, I’m happy with the change and can address the typing in a follow up PR. Let me know if you have a preference.

Thank you @nateprewitt, I will address the typing issues as well in the same PR.

src/urllib3/util/request.py
imkaka: This won't work with following error, as we are trying to assign our sentinel `_FAILEDTELL` to pos.

```
error: Incompatible types in assignment (expression has type "_TYPE_FAILEDTELL", variable has type "Optional[int]")  [assignment]
```

imkaka: I think we need to keep `Optional[_TYPE_BODY_POSITION]` only instead of `Optional[int]` to keep it uniform.

Yes, `urlopen` should not take `_FAILEDTELL`, but it is taking the `body_pos` and typing wise above solution is not bad.

nateprewitt (approved):
Great, thanks for taking a look, @imkaka! This looks good to me ⛵️

PR #2520, opened 2022-01-05
Add authority property to the Url class

nateprewitt:
This is a proposal for a new property on the `Url` class to expose the concept of "authority" as defined in [RFC 3986 § 3.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2). This comes from a recent discovery in Requests (psf/requests#6027) when swapping our use of [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) for urllib3's [`parse_url`](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#urllib3.util.parse_url). The `netloc` attribute provided by urllib3 differs from the standard library in that it's missing userinfo (or auth info) which may surprise some users.

This newly introduced `authority` property will return the composed userinfo, host, and port of a given URI. This matches the behavior of the current `netloc` provided by urllib and provides an alternative to users composing their own.

### DIFF
--- a/changelog/2520.feature.rst
+++ b/changelog/2520.feature.rst
@@ -1,0 +1,3 @@
+Added the ``authority`` property to the Url class as per RFC 3986 3.2.
+This property should be used in place of ``netloc`` for users who
+want to include the userinfo (auth) component of the URI.
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -129,8 +129,29 @@ class Url(
         return uri
 
     @property
+    def authority(self) -> Optional[str]:
+        """
+        Authority component as defined in RFC 3986 3.2.
+        This includes userinfo (auth), host and port.
+
+        i.e.
+            userinfo@host:port
+        """
+        userinfo = self.auth
+        netloc = self.netloc
+        if netloc is None or userinfo is None:
+            return netloc
+        else:
+            return f"{userinfo}@{netloc}"
+
+    @property
     def netloc(self) -> Optional[str]:
-        """Network location including host and port"""
+        """
+        Network location including host and port.
+
+        If you need the equivalent of urllib.parse's ``netloc``,
+        use the ``authority`` property instead.
+        """
         if self.host is None:
             return None
         if self.port:
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -6,9 +6,10 @@ import sys
 import warnings
 from itertools import chain
 from test import ImportBlocker, ModuleStash, notBrotli, onlyBrotli
-from typing import TYPE_CHECKING, Dict, NoReturn, Optional, Tuple, Union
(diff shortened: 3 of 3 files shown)


test/test_util.py
```diff
@@ -360,6 +360,20 @@ def test_request_uri(self, url: str, expected_request_uri: str) -> None:
         returned_url = parse_url(url)
         assert returned_url.request_uri == expected_request_uri
 
+    url_authority_map = [
+        ("http://user:pass@google.com/mail", "user:pass@google.com"),
```
sethmlarson, line 364: Whenever I'm testing URLs I try to hit the follow variants:
- single label hostname (`localhost`)
- multi label hostname
- ipv4
- ipv6

And then all of the above with and without a port.

src/urllib3/util/url.py
```diff
@@ -128,9 +128,27 @@ def request_uri(self) -> str:
 
         return uri
 
+    @property
+    def authority(self) -> Optional[str]:
+        """
+        Network location including userinfo, host and port.
```
sethmlarson, line 134: Not sure if we should describe it as the "network location"? Honestly it's still strange to me that netloc included authentication.

Let's also leave a reference to where this component is defined (ie RFC 3986 3.2)

src/urllib3/util/url.py
```diff
@@ -128,9 +128,27 @@ def request_uri(self) -> str:
...
+
+        i.e.
+            userinfo@host:port
+        """
+        if self.netloc is None or self.auth is None:
```
sethmlarson, line 139: Since a lot of these properties are calculated properties can we use variables like `auth = self.auth` and `netloc = self.netloc`.

changelog/2520.feature.rst
```diff
@@ -0,0 +1 @@
+Added the authority property to the Url class
```
sethmlarson, line 1: Can we add a note here about how this property should be used instead of `netloc`.

test/test_util.py
sethmlarson: Since this is supposed to be the same as urlparse().netloc can we also verify that each test case that starts with `http://` matches between these two APIs?

sethmlarson (changes requested):
Thanks! Some initial comments for you:

test/test_util.py
nateprewitt: I left the constraint a little looser looking for `startswith('http')` instead of fully qualified `http://`. My goal is to be greedier with new test cases with things like `https` and `http+`. That can fail at PR time if that match is wrong, and we'll address it then, rather that think we added a test case but it's silently skipped.

Happy to change if there's consensus we'd rather tighten the scope.

sethmlarson (approved):
Looks good to me, thanks so much for this!!

Issue #2015, opened 2020-10-03
keylog_filename triggering error when SSLKEYLOGFILE is not set


mattbasta:
### Subject

Requests are failing with the error:

```
urllib3.exceptions.ProtocolError: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

The line that raises this exception is the following (at the bottom of the stacktrace):

```
File "/Users/basta/.local/share/virtualenvs/venv-OmZ6DiXY/lib/python3.8/site-packages/urllib3/util/ssl_.py", line 302, in create_urllib3_context
    context.keylog_filename = os.environ.get("SSLKEYLOGFILE")
```

Editing this file to print `os.environ.get("SSLKEYLOGFILE")` shows that this is empty.

This issue started happening after upgrading from Python 3.7 to 3.8.

### Environment

This issue is happening for me on macOS Catalina 10.15.2.

```python
import platform
import urllib3

print("OS", platform.platform())
print("Python", platform.python_version())
print("urllib3", urllib3.__version__)
```

returns

```
>>> import platform
>>> import urllib3
>>>
>>> print("OS", platform.platform())
OS macOS-10.15.2-x86_64-i386-64bit
>>> print("Python", platform.python_version())
Python 3.8.5
>>> print("urllib3", urllib3.__version__)
urllib3 1.25.10
```

### Steps to Reproduce

This issue is always reproducible when making any requests using the `stripe-python` library. I don't have access to other machines to try to reproduce it elsewhere. However, it does not occur in my production environment (Linux) on the same Python version (3.8).

### Expected Behavior

I would expect this line to no-op rather than raising. Replacing `if hasattr(context, "keylog_filename"):` with `if hasattr(context, "keylog_filename") and os.environ.get("SSLKEYLOGFILE"):` in `ssl_.py` eliminates the issue.

### Actual Behavior

The noted exception is raised.

sethmlarson:
Thanks for opening this, this looks like a bug. We'll get a fix in and make a 1.25.11 release.

Would you be willing to provide a patch and test case? If not no problem.

mattbasta:
@sethmlarson I'm happy to put up my little patch as a PR. I'm not so sure how to test it, though, since the problem only seems to occur on my one machine running macOS (and I can't possibly be the only person using Stripe's SDK on python 3.8!).

If the fix alone is enough, I can have that up soon. You're welcome to use it as a guide or take it over or whatever.

[...]

PR #2525, opened 2022-01-08
Check keylog_filename existence

Ghaaf:
Linked to #2015
If the file path doesn't exist it throw following exception:

> WARNING: Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'ProtocolError('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))'

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -410,7 +410,7 @@ def create_urllib3_context(
     # 'SSLKEYLOGFILE', if the feature is available (Python 3.8+). Skip empty values.
     if hasattr(context, "keylog_filename"):
         sslkeylogfile = os.environ.get("SSLKEYLOGFILE")
-        if sslkeylogfile:
+        if sslkeylogfile and os.path.isfile(sslkeylogfile):
             context.keylog_filename = sslkeylogfile
 
     return context


sethmlarson (changes requested):
Thanks for reporting this, however I think raising an error instead of silently doing nothing is the right call for this scenario. What type of error gets raised when you configure a `ca_certs` that doesn't exist? We should try to match that behavior.

We'll also need a test case confirming the behavior.

Ghaaf:
> Thanks for reporting this, however I think raising an error instead of silently doing nothing is the right call for this scenario. What type of error gets raised when you configure a `ca_certs` that doesn't exist? We should try to match that behavior.
> 
> We'll also need a test case confirming the behavior.

I was stuck with `pip` on my windows machine for a while because the `SSLKEYLOGFILE` points to a non-existing file path on my Windows environment, so I failed to install any package using pip, but this warning all the time. Spending some time to fix the issue, I found the problem on `site-packages\pip\_vendor\urllib3\util\ssl_.py`, changing my environment path to an existing file name, fixed the problem. I'm sure about the change, but unclear which package needs to fix the issue (pip or urllib3).
My test case was `Windows 7`, `Python3.8` with the latest `pip`(`21.3.1` at the moment) contains `urllib3 1.26.7`

sethmlarson:
@Ghaaf The fix should definitely go in urllib3. Then all users of urllib3, Requests, pip, etc would benefit from its inclusion.

sethmlarson:
Going to close this for the reasons listed above. If you'd like to land this PR in urllib3 please follow this comment: https://github.com/urllib3/urllib3/pull/2525#pullrequestreview-847304993

PR #2526, opened 2022-01-10
Improve typing

hramezani:
- Update mypy to 0.931
- Change return type to NoReturn where it makes sence

Note: The changing return type to `NoReturn` is not related to new version of mypy


sethmlarson (approved):
Awesome, thank you!!

PR #2528, opened 2022-01-14
Start testing against PyPy 3.8

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/2486


hramezani:
The `pypy3.8` problem might be because of the default value of `hostname_checks_common_name` which is `True`

pquentin:
Also, even if the PyPy builds are not required to merge, in practice they are because the coverage job is required and won't run if PyPy fails. Is there a way to tell the coverage job that it does not have to wait for PyPy builds?

Issue #2190, opened 2021-03-18
Port CVE-2021-28363 fix to urllib3 v2


pquentin:
The relevant commit is https://github.com/urllib3/urllib3/commit/8d65ea1ecf6e2cdc27d42124e587c1b83a3118b0.

We have a problem though: in v2 we dropped support for common name, and that fix does not remove support for common name, especially given we can't use `hostname_checks_common_name`.

That means we'll have to call `match_hostname` manually in the proxy case too and add a test. Labelling "Breaking Change" to make sure we don't release v2 without that.

V1NAY8:
@pquentin / @sethmlarson  Can you guide me on the functionality here? I'll try to implement this and to write tests.

-  I think the logic should go in `_connect_tls_proxy`. 
- So, if we have the `ssl_context` already specified, we need not do anything. 
- Else, the `ssl_context` is created using `create_proxy_ssl_context`, after that  we have to add logic to check `_match_hostname` with hostname and certificate. 
- If Yes we set `self.ssl_context.check_hostname` to `True` else to `False` 

I will try this.

pquentin:
Figuring what to do is 90% of the work for me, I didn't take time to sit down and focus on this yet. 

I would suggest however to start with the test from https://github.com/urllib3/urllib3/commit/8d65ea1ecf6e2cdc27d42124e587c1b83a3118b0. I would merge a PR with that test separated in two and annotated with pytest's xfail.

pquentin:
(To be clear, I'm not suggesting your approach is wrong! I don't want to discourage you from implementing it, but I can't confirm or deny that it's correct.)

sethmlarson:
Other PRs we may need to forward-port:
- https://github.com/urllib3/urllib3/pull/2407
- https://github.com/urllib3/urllib3/pull/2419

PR #2529, opened 2022-01-16
Change proxy TLS flow to match destination TLS flow

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/2190 Part of https://github.com/urllib3/urllib3/issues/2530

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -483,7 +483,7 @@ class HTTPSConnection(HTTPConnection):
                 SystemTimeWarning,
             )
 
-        self.sock, self.is_verified = _ssl_wrap_socket_and_match_hostname(
+        sock_and_verified = _ssl_wrap_socket_and_match_hostname(
             sock=sock,
             cert_reqs=self.cert_reqs,
             ssl_version=self.ssl_version,
@@ -501,6 +501,8 @@ class HTTPSConnection(HTTPConnection):
             assert_hostname=self.assert_hostname,
             assert_fingerprint=self.assert_fingerprint,
         )
+        self.sock = sock_and_verified.socket
+        self.is_verified = sock_and_verified.is_verified
         self._connecting_to_proxy = False
 
     def _connect_tls_proxy(self, hostname: str, sock: socket.socket) -> "ssl.SSLSocket":
@@ -511,7 +513,7 @@ class HTTPSConnection(HTTPConnection):
             ProxyConfig, self.proxy_config
         )  # `_connect_tls_proxy` is called when self._is_using_tunnel() is truthy.
         ssl_context = proxy_config.ssl_context
-        ssl_sock, self.proxy_is_verified = _ssl_wrap_socket_and_match_hostname(
+        sock_and_verified = _ssl_wrap_socket_and_match_hostname(
             sock,
             cert_reqs=self.cert_reqs,
             ssl_version=self.ssl_version,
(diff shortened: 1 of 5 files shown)


src/urllib3/connection.py
```diff
@@ -26,8 +26,8 @@
     from typing_extensions import Literal
 
     from .util.ssl_ import _TYPE_PEER_CERT_RET_DICT
+    from .util.ssltransport import SSLTransport as SSLTransportType
```
sethmlarson, line 29: Should maybe only import this on type checking time.

src/urllib3/connection.py
```diff
@@ -478,125 +483,161 @@ def connect(self) -> None:
...
+    ssl_context: Optional["ssl.SSLContext"],
+    tls_in_tls: bool = False,
+) -> Tuple[Union["ssl.SSLSocket", "SSLTransport"], bool]:
+    """Logic for constructing an SSLContext from all TLS parameters, passing
+    that down into ssl_wrap_socket, and then doing certificate verification
+    either via hostname or fingerprint. This function exists to guarantee
+    that both proxies and targets have the same behavior when connecting via TLS.
+    """
+
+    # Wrap socket using verification with the root certs in
```
sethmlarson, line 561: This comment is meaningless, should remove it.

src/urllib3/connection.py
```diff
@@ -478,125 +483,161 @@ def connect(self) -> None:
...
+        or assert_hostname
+        # We still support OpenSSL 1.0.2, which prevents us from verifying
+        # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
+        or ssl_.IS_PYOPENSSL
+        or not ssl_.HAS_NEVER_CHECK_COMMON_NAME
+    ):
+        context.check_hostname = False
+
+    # Try to load OS default certs if none are given.
+    # Works well on Windows.
```
jalopezsilva, line 590: Any reason we explicitly mention this? I would have assumed it worked in Windows without the comment.

src/urllib3/connection.py
```diff
@@ -478,125 +483,161 @@ def connect(self) -> None:
...
+    key_password: Optional[str],
+    ca_certs: Optional[str],
+    ca_cert_dir: Optional[str],
+    ca_cert_data: Union[None, str, bytes],
+    assert_hostname: Union[None, str, "Literal[False]"],
+    assert_fingerprint: Optional[str],
+    server_hostname: Optional[str],
+    ssl_context: Optional["ssl.SSLContext"],
+    tls_in_tls: bool = False,
+) -> Tuple[Union["ssl.SSLSocket", "SSLTransport"], bool]:
```
jalopezsilva, line 554: Not a huge fan of returning tuples, should we make it a small struct instead?

Advantages are: it will be easier to extend if we need it, return doesn't depend on ordering of the parameters.

jalopezsilva (commented):
Some minor comments, change looks good.

src/urllib3/connection.py
sethmlarson: I think back when it was originally added (by me iirc) the function itself mentioned that it mostly "worked good on Windows". Not sure why I felt like I had to copy that over though. I think it's good to remove and is self-explanatory.

src/urllib3/connection.py
sethmlarson: Yeah, we should return a namespaced object, even if it's a namedtuple.

pquentin (commented):
small nit I noticed when reading this

src/urllib3/connection.py
```diff
@@ -120,6 +120,10 @@ class HTTPConnection(_HTTPConnection):
     #: Whether this connection verifies the host's certificate.
     is_verified: bool = False
 
+    #: Whether this proxy connection (if used) verifies the proxy host's
+    #: certificate. If no HTTPS proxy is being used will be ``None``.
```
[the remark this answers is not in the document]
sethmlarson: This is intentional to make the comment a docstring as is done above for `is_verified`

jalopezsilva (approved):
Looking great, thanks for making the changes!

Issue #2534, opened 2022-01-18
PoolManager.urlopen fails with TypeError for http connection if the PoolManager is instantiated with server_hostname


kajinamit:
### Subject

PoolManager.urlopen fails with TypeError for http connection if the PoolManager is instantiated with server_hostname

### Environment

```
>>> import platform
>>> import urllib3
>>> 
>>> print("OS", platform.platform())
OS Linux-5.11.22-100.fc32.x86_64-x86_64-with-glibc2.2.5
>>> print("Python", platform.python_version())
Python 3.8.10
>>> print("urllib3", urllib3.__version__)
urllib3 1.26.8

```

### Steps to Reproduce

Create a PoolManager with server_hostname and call urlopen method of the instance.
```
>>> import urllib3
>>> manager = urllib3.PoolManager(server_hostname='foo')
>>> manager.urlopen('GET', 'http://127.0.0.1') 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/tkajinam/virtualenv/foo/lib/python3.8/site-packages/urllib3/poolmanager.py", line 375, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/home/tkajinam/virtualenv/foo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 692, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/home/tkajinam/virtualenv/foo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 281, in _get_conn
    return conn or self._new_conn()
  File "/home/tkajinam/virtualenv/foo/lib/python3.8/site-packages/urllib3/connectionpool.py", line 235, in _new_conn
    conn = self.ConnectionCls(
  File "/home/tkajinam/virtualenv/foo/lib/python3.8/site-packages/urllib3/connection.py", line 130, in __init__
    _HTTPConnection.__init__(self, *args, **kw)
TypeError: __init__() got an unexpected keyword argument 'server_hostname'
```

### Expected Behavior

TypeError should not be raised and a request is sent

### Actual Behavior

TypeError is raised and a request is not sent

kajinamit:
FYI: This was initially reported in https://github.com/cdent/gabbi/issues/309 .

kajinamit:
Just adding some more results...

TypeError is not raised when I use https. This is because HTTPSConnection DOES accept the server_hostname argument.
[...]

PR #2535, opened 2022-01-18
Fix TypeError caused by interface mismatch

kajinamit:
The PoolManager class takes the connection_pool_kw argument which
defines global parameters used for the connection pools managed by that
manager. Some of them are passed to the conn_kw argument to instantiate
a connection.

The problem here is that the PoolManager class instantiates different
classes of connection pools according to the protocol used. This
results in the issue caused by the different arguments in connection
classes(HTTPConnection vs HTTPSConnection).

This change introduces an additional kwargs to these connection classes
so that these classes can accept arguments which are not implemented.

Resolves: #2534

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -138,6 +138,7 @@ class HTTPConnection(_HTTPConnection):
         ] = default_socket_options,
         proxy: Optional[str] = None,
         proxy_config: Optional[ProxyConfig] = None,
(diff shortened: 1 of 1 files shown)


rabi:
https://github.com/urllib3/urllib3/pull/2537 better fix?

kajinamit:
> #2537 better fix?

Yes. That's much better and looks like a right approach.

PR #2537, opened 2022-01-18
Add server_hostname to SSL_KEYWORDS

rabi:
It is only relevant for HTTPSConnection.

Resolves: #2534

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -55,6 +55,7 @@ SSL_KEYWORDS = (
     "ca_cert_dir",
     "ssl_context",
     "key_password",
+    "server_hostname",
 )
 # Default value for `blocksize` - a new parameter introduced to
 # http.client.HTTPConnection & http.client.HTTPSConnection in Python 3.7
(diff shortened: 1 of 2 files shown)


kajinamit (approved):

cdent:
Looks like black wants a minor change before all tests will pass, but this looks good to me and fixes the issues being seen in gabbi (where a PoolManager can be used with mixed schemes).

sethmlarson (approved):
Thank you for this change, could you add a test case that catches this error so we don't regress in the future?

pquentin:
(We'll also have to backport this to the 1.26.x branch))

sethmlarson:
Very true! Our team can handle the backport after this PR is merged.

rabi:
>  could you add a test case that triggers the error so we don't regress in the future

Added a test. Hope that would suffice for now.

hramezani (approved):

sethmlarson (approved):
LGTM, thanks!

hramezani:
@sethmlarson I will create a patch to backport it to the `1.26.x`.

PR #2542, opened 2022-01-20
Add server_hostname to SSL_KEYWORDS

hramezani:
Backport https://github.com/urllib3/urllib3/commit/2459900b10f54326745a597cc1cf2bf8a18f0d56


sethmlarson (approved):
LGTM, thanks!

kajinamit:
Hi. Could you create a new release which includes the fix ?
Also, it would be helpful if we can backport this to 1.24.x and 1.25.x, since the problem was initially introduced in 1.24.0.

Issue #2538, opened 2022-01-18
Use static build system and reproducible builds


sethmlarson:
- [x] Switch to a build system that supports PEP 517 (I'm leaning towards Flit)
- [x] Change our long description to be simpler and static inside of `pyproject.toml` instead of based on README.rst. Keep existing README.rst as-is but don't distribute with the source (ie remove from MANIFEST.in)
- [x] Start setting `SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)` when building wheels. Add this recommendation to our release checklist (https://reproducible-builds.org/docs).
  - [x] Verify that the above works by having two maintainers build the wheel+sdist separately, both should be byte-for-byte identical.

sethmlarson:
@shazow and I have independently built commit https://github.com/urllib3/urllib3/commit/8ab5eb83a0bf359f5372a8eff2b07f46fe34203d using the provided release instructions and received a SHA256 of:

`96b40223c768231969979ec1a7afade8dca7d34ccff3b55fc172a59bfbea3487  dist/urllib3-2.0.0.dev0-py3-none-any.whl`

Success! Closing this now.

PR #2545, opened 2022-01-21
Convert README to Markdown

pquentin:
The main impetus for this was to remove the reST parsing in setup.py, which is an important step towards #2538.

I've also used this as an opportunity to use our real dark mode logo, by abusing the fact that https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/ was implemented with CSS. (This could have been done in reST too, but I only found out when working on this change.)

This also removes the Tidelift logo as I didn't try using a Markdown table.

:warning: We need to upload a packaged version of this to test.pypi.org to make sure it renders correctly there.


sigmavirus24:
There's a much easier to support reproducible builds without using markdown. You could have started specifying things in setup.cfg the way others do with `file:README.rst` and removed it from here. That's an even bigger step towards static metadata

sethmlarson:
@sigmavirus24 We'll definitely be doing the `readme:README` config. Our issue is rST doesn't allow embedding HTML unlike markdown so our `setup.py` had logic which stripped the HTML out of rST before building the package.

shazow:
What issues does markdpwn introduce for reproducible builds?

pquentin:
> Our issue is rST doesn't allow embedding HTML unlike markdown 

It's supported in GitHub, but not in PyPI, which is why we stripped that in setup.py.

> What issues does markdown introduce for reproducible builds?

Why are you asking that? Markdown does not introduce issues, hence the switch.

shazow:
@pquentin Er that was a question for @sigmavirus24's statement about

> There's a much easier to support reproducible builds without using markdown. [...]

sigmavirus24:
> @pquentin Er that was a question for @sigmavirus24's statement about
> 
> > There's a much easier to support reproducible builds without using markdown. [...]

Was trying to say "without switching to markdown" markdown poses no issues

pquentin:
I had to restore the logo with white outlines because PyPI does not support `gh-dark-mode-only` / `gh-light-mode-only`. I found a hack with a markdown image inside a `<h1>` tag that works:

 * GitHub: https://github.com/pquentin/urllib3/tree/markdown-readme
 * PyPI: https://test.pypi.org/project/urllib3-markdown-pquentin/2.0.0.dev0/

sethmlarson (approved):
Looks great, that's awesome that we can have our banner image on PyPI finally! :tada: Thanks @pquentin

PR #2549, opened 2022-01-25
Switch to Flit as a build-system

sethmlarson:
Part of https://github.com/urllib3/urllib3/issues/2538


pquentin:
I've fixed `twine check` that was expecting a long description, and added a new check called `unsupported_setup_py` to avoid baffling users that are still calling setup.py install. Please tell me what you think.

@mgorny I've just read https://blogs.gentoo.org/mgorny/2021/11/07/the-future-of-python-build-systems-and-gentoo/, and am wondering: how much pain is this going to cause to distributions if we ship 2.0 with this switch to flit later this year?

mgorny:
As long as it's flit, it's not a problem for us. Even if we can't use PEP 517 for some reason, pyproject2setuppy should be able to handle it.

shazow (dismissed):

PR #2550, opened 2022-01-25
Remove > signs from links in README

franekmagiera:
Removed `>` signs from two links in README so they are clickable


sethmlarson (approved):
Great catch, thank you!

nateprewitt (approved):
Thanks @franekmagiera!

PR #2552, opened 2022-01-26
Stop linking to codecov in README

### LINKED ISSUE #2551: Stop linking to codecov in README
We currently have a coverage badge in our README:

image

We're very proud that we had 100% coverage for years. Not only it's very useful to us, but it highlights that urllib3 has high code quality and exhaustive tests. However, this badge currently links to codecov. We stopped using codecov, though, because it was too flaky, see https://github.com/urllib3/urllib3/pull/2502.

Since we want to keep the 100% badge, let's keep it in the README by inlining it in `docs/_static` and removing the link to codecov. It won't be live anymore but we enforce that 100% coverage and it was like that for years, so it's not like we're going to deceive anyone.

imkaka:
Closes  #2551


nateprewitt (approved):
Thanks @imkaka!

sigmavirus24 (commented):
Am I reading the SVG correctly (from my phone) that it will always show 100% coverage?

nateprewitt:
@sigmavirus24 we’re still [running coverage](https://github.com/urllib3/urllib3/runs/4952946422?check_suite_focus=true) to ensure 100%, there just isn’t an external source to populate the number.

I agree it is a bit questionable at first glance, but we do have an enforcement mechanism. I’m not sure if there are other options to tie the badge to that job other than a flat pass/fail.

sigmavirus24:
Yeah, my point is that in the future it may be reasonable (though some people (not necessarily associated with urllib3) seem to think 100% code coverage is the only acceptable option) to lower that threshold and then the icon becomes inaccurate. I don't know that there's a good way to ensure they don't fall out of sync. I suppose we just go along

README.md
```diff
@@ -8,7 +8,7 @@
   <a href="https://pypi.org/project/urllib3"><img alt="PyPI Version" src="https://img.shields.io/pypi/v/urllib3.svg?maxAge=86400" /></a>
   <a href="https://pypi.org/project/urllib3"><img alt="Python Versions" src="https://img.shields.io/pypi/pyversions/urllib3.svg?maxAge=86400" /></a>
   <a href="https://discord.gg/urllib3"><img alt="Join our Discord" src="https://img.shields.io/discord/756342717725933608?color=%237289da&label=discord" /></a>
-  <a href="https://codecov.io/gh/urllib3/urllib3"><img alt="Coverage Status" src="https://img.shields.io/codecov/c/github/urllib3/urllib3.svg" /></a>
+  <img alt="Coverage Status" src="./docs/_static/coverage.svg" />
```
sethmlarson, line 11: Since this README is used on PyPI, this link would likely break after being uploaded. Instead let's use:

`https://img.shields.io/badge/coverage-100%25-success`

And we can maintain an `href` and link to our CI as is done below.

sethmlarson (changes requested):
Thanks for this, two things:

pquentin:
> Yeah, my point is that in the future it may be reasonable [...] to lower that threshold and then the icon becomes inaccurate.

@sigmavirus24 for what it's worth I don't agree. I'm not saying that every software project should have 100% coverage, on the contrary, all other projects I contribute to have less than that. In fact, even urllib3 doesn't actually have 100% coverage: we exclude the the alternate SSL backends and various branches in the main code.

However, when you get to 100% cocerage, I'm saying that staying at 100% is worth it. You already made the hard work, and staying at 100% is easy. It lets you catch various issues and removing covered code cannot lower coverage, so coverage that gets under 100% is always actionable.

So in the future we could have more code that is excluded from coverage, but I don't think going below 100% would be reasonable.

README.md
imkaka: Sure.

pquentin: The shields.io URL is a great catch, thanks @sethmlarson!

pquentin:
@imkaka Oops, I clicked on a button that merged `main` into your branch, make sure to pull before adding a new commit

imkaka:
> @imkaka Sorry, I clicked on a button that merged `main` into your branch, make sure to pull before adding a new commit

Sorry, I read this after adding the commit, do you want me to do anything for the commit here?

sigmavirus24:
> n fact, even urllib3 doesn't actually have 100% coverage: we exclude the alternate SSL backends and various branches in the main code.

Yeah, in that case the 100% number is even more of a disingenuous advertisement to users. There are absolutely things that they may rely on (knowingly or not) that are not well tested and can become unstable as a result.

sigmavirus24 (approved):

sethmlarson (approved):
LGTM, thanks!

Issue #2543, opened 2022-01-20
[1.26] Coverage dropped below 100% when vendoring ssl_match_hostname


pquentin:
https://github.com/urllib3/urllib3/pull/2540 shows that in the 1.26.x branch, the following lines are not covered:

https://github.com/urllib3/urllib3/blob/b1f60e44d43b13e5272d5b6003f125af9c25c8ad/src/urllib3/util/ssl_match_hostname.py#L118-L122

This was never covered, we just notice it now because we vendored that code. We need a test.

https://github.com/urllib3/urllib3/blob/b1f60e44d43b13e5272d5b6003f125af9c25c8ad/src/urllib3/util/ssl_match_hostname.py#L150

It seems like this line is covered in the main branch, maybe adding a `raise ValueError()` and running the main tests will show how it's covered? (It's surprising because we have the same tests in both branches.)

In the meantime, since this blocks merging 1.26 PRs, I'd suggest using a nocover pragma.

@chetanugale @hramezani This issue may be of interest to you as you've touched those files in the past.

hramezani:
Thanks @pquentin for the issue. 
I will check later uncovered line in  `ssl_match_hostname.py` 
 https://github.com/urllib3/urllib3/blob/b1f60e44d43b13e5272d5b6003f125af9c25c8ad/src/urllib3/util/ssl_match_hostname.py#L150

> In the meantime, since this blocks merging 1.26 PRs, I'd suggest using a nocover pragma.

Sound good to me for now!

hramezani:
To cover `UnicodeError` except:
https://github.com/urllib3/urllib3/blob/b1f60e44d43b13e5272d5b6003f125af9c25c8ad/src/urllib3/util/ssl_match_hostname.py#L118-L122

As [UnicodeError  is a subclass of ValueError](https://docs.python.org/3.8/library/exceptions.html#UnicodeError), and in both `except` blocks we have `host_ip = None`, I would suggest to merge this two blocks:

```
except (UnicodeError, ValueError): 
```

Or remove `UnicodeError` block,
Or move `UnicodeError` above the `ValueError` and add a test for it.

What do you think @pquentin ?

hramezani:
@pquentin I think we can close this now!

sethmlarson:
Yes we can! Thanks @hramezani

PR #2554, opened 2022-01-28
[1.26] Complete test coverage of urllib3.util.ssl_match_hostname.match_hostname.

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/2543

### DIFF
--- a/test/test_connection.py
+++ b/test/test_connection.py
@@ -28,6 +28,12 @@ class TestConnection(object):
         asserted_hostname = "foo"
         _match_hostname(cert, asserted_hostname)
 
+    def test_match_hostname_ipaddress_none(self):
+        cert = {"subjectAltName": [("DNS", "foo")]}
+        asserted_hostname = "foo"
+        with mock.patch("urllib3.util.ssl_match_hostname.ipaddress", None):
+            assert _match_hostname(cert, asserted_hostname) is None
+
     def test_match_hostname_mismatch(self):
         cert = {"subjectAltName": [("DNS", "foo")]}
         asserted_hostname = "bar"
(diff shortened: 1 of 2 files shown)


test/test_connection.py
```diff
@@ -28,6 +28,21 @@ def test_match_hostname_match(self):
         asserted_hostname = "foo"
         _match_hostname(cert, asserted_hostname)
 
+    def test_match_hostname_ipaddress_none(self):
+        cert = {"subjectAltName": [("DNS", "foo")]}
+        asserted_hostname = "foo"
+        with mock.patch("urllib3.util.ssl_match_hostname.ipaddress", None):
+            assert _match_hostname(cert, asserted_hostname) is None
+
+    def test_match_hostname_invalid_ipaddress(self):
```
sethmlarson, line 37: This test case shouldn't be necessary? There shouldn't be any case where the `ipaddress` module is a string.

sethmlarson (changes requested):
One comment, otherwise LGTM:

test/test_connection.py
hramezani: Yes, but this test covers https://github.com/urllib3/urllib3/blob/048c0ba15cec8c7dd787670bcafd243a54590982/src/urllib3/util/ssl_match_hostname.py#L126

nateprewitt: That's probably easily solved with a docstring. Having to work that out later is going to be difficult.

sethmlarson: Yeah, I'm not convinced we should test that branch, I don't think it's triggerable during regular execution. Let's add a `# Defensive:` comment to the `else` instead?

hramezani: Ok, i've removed the test and added `Defensive` in else block

sethmlarson (approved):
LGTM!!! 100% coverage restored.

PR #2557, opened 2022-02-08
Give the developer a hint if the pool is not available for a long time

SadPencil:
When someone forget to release the connections, the program might hang for a very long time.
This PR added a logging at debug level to give a hint.

-------
The code below shows how I got stuck and wasted several hours:
```python
http = PoolManager(block=True)
for url in urls:
    # Surprisingly, the connection can not be released by using 'with' keyword,
    # while no exceptions are raised because of this.
    # Thus it will finally get hang, without any hints.
    # It takes so much time to figure out it is due to the pool instead of handshakes
    with http.request('GET', url, preload_content=False) as r:
        do_something(r)
```

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -258,7 +258,16 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             raise ClosedPoolError(self, "Pool is closed.")
 
         try:
-            conn = self.pool.get(block=self.block, timeout=timeout)
+            if self.block and timeout is None:
+                # Don't wait and hang forever. Give the developer a hit if they forget to release the connection.
+                while True:
+                    try:
+                        conn = self.pool.get(block=self.block, timeout=60)
+                        break
+                    except queue.Empty:
+                        log.debug('Still waiting for available connections. Did you forget to release them?')
+            else:
+                conn = self.pool.get(block=self.block, timeout=timeout)
 
         except AttributeError:  # self.pool is None
             raise ClosedPoolError(self, "Pool is closed.") from None  # Defensive:


src/urllib3/connectionpool.py
```diff
@@ -258,7 +258,16 @@ def _get_conn(self, timeout: Optional[float] = None) -> HTTPConnection:
...
         try:
-            conn = self.pool.get(block=self.block, timeout=timeout)
+            if self.block and timeout is None:
+                # Don't wait and hang forever. Give the developer a hit if they forget to release the connection.
+                while True:
+                    try:
+                        conn = self.pool.get(block=self.block, timeout=60)
+                        break
+                    except queue.Empty:
+                        log.debug('Still waiting for available connections. Did you forget to release them?')
```
sigmavirus24, line 268: So in production, if the pool (or pools) is just particularly active, we could spam users' logs? I get that this is debug, but some people do enable urllib3 debug logs purposefully. What happens if there are multiple pools having similar issues? There's no indication as to which pool is having trouble and there's no indication as to how to fix this if you're using a library built atop urllib3 because it's not actionable.

sigmavirus24:
Why is it surprising that we cannot release a connection if we haven't consumed the response returned on that connection?  Do you expect `__exit__` to consume it all and then release the connection?  I believe the real problem is that you were surprised by this and we need to either improve that behaviour or make it easier for you to understand what's happening there without logging a vague message that many people will see with no clue how to action given the sheer popularity of urllib3

SadPencil:
> 

Yes. I do expect `__exit__` consume and release all resources because that is what `with` do.

sigmavirus24:
> Yes. I do expect `__exit__` consumes and releases all resources

So you'd be unsurprised if your code stopped at the `__exit__` while it consumed potentially gigs (or more) of data from the socket to be able to return it to the pool? In fact, some servers can slow trickle data for long periods of time - days even - to create a denial of service. You're asking for `__exit__` to become just such a vector for a Denial of Service attach.

SadPencil:
> So you'd be unsurprised if your code stopped at the __exit__ while it consumed potentially gigs (or more) of data from the socket to be able to return it to the pool? In fact, some servers can slow trickle data for long periods of time - days even - to create a denial of service. You're asking for __exit__ to become just such a vector for a Denial of Service attack.

Well, since calling `release_conn()` consumes a lot of data as you said which make performing cleanup at `__exit__` unacceptable, should an exception or a warning be raised at `__enter__` and `__exit__` as they should not be invoked in any circumstance?

SadPencil:
> > > So you'd be unsurprised if your code stopped at the `__exit__` while it consumed potentially gigs (or more) of data from the socket to be able to return it to the pool? In fact, some servers can slow trickle data for long periods of time - days even - to create a denial of service. You're asking for` __exit__` to become just such a vector for a Denial of Service attack.
> > 
> > 
> > Well, since calling `release_conn()` consumes a lot of data as you said which make performing cleanup at `__exit__` unacceptable, should an exception or a warning be raised at `__enter__` and `__exit__` as they should not be invoked in any circumstance?
> 
> That's not what I said. This is only the case with preload_content. And it's a potential issue, not a guaranteed one. You've implicitly opted to handle this yourself with preload_content=False

Well then, when `preload_content=True`, the user does not need to do any clean-ups and therefore the context management is unnecessary. A `with` statement is unneeded in this case. As for `preload_content=False`, the user needs to call `release_conn()` by hand, then someone might choose to use `with` statement as if it is a file object. But consider its drawback, no clean-ups will be done at `__exit__`, therefore a `with` statement is not proper in this case (misleading). Therefore `__enter__` and `__exit__` should not be invoked in any circumstance and a warning should be given.

sethmlarson:
I think I'm going to decline this PR in the state that it's in. Perhaps a better way to work around this issue is with better documentation+examples?

PR #2562, opened 2022-02-12
typing.TypedDict is always available on Python 3.7+

sethmlarson:


sethmlarson:
For some reason I thought TypedDict was added in 3.7, we need this until we drop 3.7.

PR #2568, opened 2022-02-16
Update _request_methods.py

omidekz:
update ```request(..., headers={...})``` (the request header) with ```self.headers``` when is not ```None```


sethmlarson:
This would break compatibility with how urllib3 has worked in the past. We're unlikely to take this direct approach. See: https://github.com/urllib3/urllib3/issues/2254

PR #2569, opened 2022-02-25
Match docs with github page

bt-nia:
Fixed the broken example in the docs index page


sethmlarson:
`urllib3.request()` is an unreleased feature, the documentation you're referencing is `latest` which isn't the default docs URL when you visit https://urllib3.readthedocs.io. I'm going to close this as this example is expected.

PR #2572, opened 2022-02-25
[1.26] Test how headers are folded and sent with .add()

sethmlarson:
Part of https://github.com/urllib3/urllib3/issues/2242, adds tests to 1.26.x for how headers are sent and folded by default.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1668,7 +1668,7 @@ class TestHeaders(SocketDummyServerTestCase):
             sock.send(b"HTTP/1.1 200 OK\r\n\r\n")
             sock.close()
 
-        headers = urllib3.HTTPHeaderDict()
+        headers = HTTPHeaderDict()
         headers.add("A", "1")
         headers.add("B", "2")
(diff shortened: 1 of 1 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1656,6 +1656,35 @@ def socket_handler(listener):
...
+        headers.add("A", "4")
+        headers.add("C", "5")
+
+        self._start_server(socket_handler)
+
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            r = pool.request("GET", "/", retries=0, headers=headers)
+            assert r.status == 200
+
+        assert b"\r\nA: 1, 4\r\nB: 2\r\nC: 3, 5\r\n" in data_sent[0]
```
pquentin, line 1686: I have two entries for C locally:

```
\r\nA: 1\r\nA: 4\r\nB: 2\r\nC: 3\r\nC: 5\r\n
```

sethmlarson:
We've done enough discovery here, I've updated https://github.com/urllib3/urllib3/issues/2242 with the list of todos for the `main` branch.

PR #2585, opened 2022-03-25
GitHub no longer supports the git:// protocol

verhovsky:

PR #2586, opened 2022-03-25
[1.26] Fix git installation instructions

verhovsky:
GitHub no longer supports the git:// protocol

backports #2585 and #2587 to 1.26


pquentin (approved):
Thanks!

PR #2594, opened 2022-04-12
[1.26] Upgrade black to latest version that supports Python 2

pquentin:
This allows us to get the same formatting as for Python 3.

We also pin click because black uses a private click API in 21.12b0.


hramezani (approved):

nateprewitt (approved):

sethmlarson (approved):
Thanks Quentin!

Issue #2583, opened 2022-03-22
Demote `Retrying .. after connection broken by ..` log warning to info


bluetech:
### Subject

When `urlopen` retries a request, it issues a warning log:

https://github.com/urllib3/urllib3/blob/37a67739ba9a7835886555744e9dabe3009e538a/src/urllib3/connectionpool.py#L819-L824

I have things set up such that when a warning (or above) is emitted, I get notified.
In my opinion, a request being retried does not warrant a warning, because it is *expected* - network conditions are not ideal, retries are going to happen. If the retry succeeded, great. If it didn't, it will raise an exception which will be handled some way, which *then* may warrant my attention.

This is not hard to workaround, but nonetheless, I think it will be better to change.

I propose:

1. Changing for 2.x (I can send a PR).
2. Backporting to 1.26.x (but can understand if this is not deemed acceptable).

### Environment

```
OS Linux-5.16.14-arch1-1-x86_64-with-glibc2.35
Python 3.9.9
urllib3 1.26.8
```

### Steps to Reproduce

(This is with current urllib3 main):

```py
import logging
import urllib3

logging.basicConfig(level=logging.INFO)

try:
    urllib3.request("GET", "http://example.com:9999", retries=2)
except urllib3.exceptions.MaxRetryError:
    pass
```

### Expected Behavior

The following is printed:

```
INFO:urllib3.connectionpool:Retrying (Retry(total=1, connect=None, read=None, redirect=0, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe812e89d20>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /
INFO:urllib3.connectionpool:Retrying (Retry(total=0, connect=None, read=None, redirect=0, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe812e89ab0>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /
```

### Actual Behavior

The following is printed:

```
WARNING:urllib3.connectionpool:Retrying (Retry(total=1, connect=None, read=None, redirect=0, status=None)) after connection broken by 'NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f49c7831c60>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /
[...]

PR #2595, opened 2022-04-14
Allow to change Retry message log level

sathieu:
This is part of https://github.com/urllib3/urllib3/issues/2583, which can be closed once the default is changed from `WARNING` to `INFO`.


sigmavirus24:
As I said I'm the linked issue, I'm very much against this approach

sethmlarson:
Agreeing with @sigmavirus24, we're going to decline this as a solution to the issue.

sathieu:
@sigmavirus24 @sethmlarson Sorry, I thought you were against the callback approach, but not against the argument approach.

PR #2599, opened 2022-04-30
Stop applying requests patch

pquentin:
Not only it broke since requests switched to black, but the test seems to
be passing in CI now?


pquentin:
So a different test is failing now:

```
FAILED tests/test_requests.py::TestRequests::test_https_warnings
```

pquentin:
Ugh no this isn't a new one, this one is supposed to fail. I got confused because 1.26.x is applying a different patch.

PR #2600, opened 2022-04-30
Fix requests patch after switch to black

pquentin:
Here's the requests pull request where the switch happened: https://github.com/psf/requests/pull/6095


sethmlarson (approved):
LGTM, thanks Quentin!

hramezani (approved):

PR #2601, opened 2022-04-30
[1.26] Fix requests downstream test

pquentin:
The patch was not working anymore since requests switched to black. But
it was no longer needed anyway as requests dropped Python 2 and the
`test_pyopenssl_redirect` test passes on Python 3.

We currently only run those tests on releases, but you can see them in my fork here: https://github.com/pquentin/urllib3/runs/6239879635?check_suite_focus=true


sethmlarson (approved):
Nice, happy we don't have to carry this patch any more.

Issue #2614, opened 2022-05-27
Backport #2613 to 1.26.x


sethmlarson:
This is a tracking issue for backporting #2613 to the 1.26.x branch.

PR #2616, opened 2022-05-30
[1.26] Fix incorrect error message during connection errors to HTTPS websites using a HTTP proxy

hramezani:
Fixes https://github.com/urllib3/urllib3/issues/2614

As we don't have `_wrap_proxy_error` in `1.26.x`, I had to add a minimal version `_wrap_proxy_error` to make the change able to test and also don't change the old behaviour

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -17,7 +17,6 @@ from .connection import (
     HTTPException,
     HTTPSConnection,
     VerifiedHTTPSConnection,
-    _wrap_proxy_error,
     port_by_scheme,
 )
 from .exceptions import (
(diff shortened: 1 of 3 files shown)


src/urllib3/connectionpool.py
```diff
@@ -768,13 +769,7 @@ def _is_ssl_error_message_from_http_proxy(ssl_error):
...
                 and _is_ssl_error_message_from_http_proxy(e)
             ):
-                e = ProxyError(
-                    "Your proxy appears to only use HTTP and not HTTPS, "
-                    "try changing your proxy URL to be HTTP. See: "
-                    "https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html"
-                    "#https-proxy-error-http-proxy",
-                    SSLError(e),
-                )
+                e = _wrap_proxy_error(e, conn.proxy)
```
sethmlarson, line 772: Instead of using an empty string for the message let's only wrap the `SSLError` with a `ProxyError` if `conn.proxy.scheme == 'https'`?

sethmlarson (changes requested):

src/urllib3/connectionpool.py
hramezani: What do you mean by `let's only wrap the SSLError with a ProxyError`?
if we do it like `ProxyError(SSLError(err))`, It will return error because `ProxyError` needs `message` arg.

Also, I think you meant `conn.proxy.scheme != 'https'`

sethmlarson: I'm suggesting adding the `conn.proxy.scheme == 'https'` to the `if` condition starting on line 767. That way we only wrap with `ProxyError` with our fancy message if the proxy is HTTPS.

hramezani: Ok, I got it now!
I've added `_wrap_proxy_error` to make it similar to the one that we have on `main` and also easy to test.
As we have [test_https_proxymanager_connected_to_http_proxy](https://github.com/urllib3/urllib3/blob/9397c61c7f447d466487ee8a4b6b2a44cbff0366/test/with_dummyserver/test_socketlevel.py#L1165) I think the changing the `if` condition make sense.

sethmlarson (approved):
LGTM!

PR #2621, opened 2022-06-02
Use 3.11-dev as it's "latest tag" not "latest commit"

sethmlarson:


pquentin (approved):
LGTM! Thanks. We may want to switch to setup-python@v3 at some point, even if v2 is working well.

PR #2623, opened 2022-06-04
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -2,6 +2,9 @@ name: Downstream
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   integration:
     strategy:


sethmlarson:
Interesting, I didn't know about this feature. However, we don't use any sort of tokens/environment variables to my knowledge as a part of our integration testing. Am I missing something that this setting protects against?

pquentin:
I like the idea. It's unobtrusive, explicit and adds defense in depth. If one day we do add a secret for whatever reason we'll be happy to have this.

However, if we adopt it, it should be added to all workflows.

sethmlarson:
Closing as duplicate of https://github.com/urllib3/urllib3/pull/2683

PR #2624, opened 2022-06-08
Zstandard decoder implementation #1992

### LINKED ISSUE #1992: [v2] Add support for Content-Encoding: zstd
See: https://tools.ietf.org/html/rfc8478

### Minimum requirements

:moneybag: **You can get paid to complete this issue! [Please read the docs for more information](https://urllib3.readthedocs.io/en/latest/contributing.html#getting-paid-for-your-contributions).**

- [ ] Evaluate all zstandard packages that are available for Python, which makes sense for us to support? One, multiple?
- [ ] Add support for a `ZstandardDecoder`, don't enable `zstd` in `Accept-Encoding` by default though.
- [ ] Add documentation for the feature
- [ ] Add test cases for the feature, see `BrotliDecoder` and `Content-Encoding: brotli` tests for inspiration.

mamico:
I've evaluated https://github.com/animalize/pyzstd and https://github.com/indygreg/python-zstandard, they are really close in terms of features and activities.

I initially implemented for both libraries, but chose pyzstd because I was unable to raise the exception with python-zstandard in streaming mode (conform to the test https://github.com/urllib3/urllib3/compare/main...mamico:zstd?expand=1#diff-f853b537efd0c07ee4e2affffe963ac50d718f5894aa4e6f4e57c5f98ad98599R283).

Closes #1992

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -564,7 +564,7 @@ Here's an example using zstd encoding via the ``Accept-Encoding`` header:
 
     urllib3.request(
         "GET",
-        "https://facebook.com/",
+        "https://www.facebook.com/",
         headers={"Accept-Encoding": "zstd"}
     )
 
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -544,7 +544,7 @@ Zstandard Encoding
 `Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
 is a compression algorithm created by Facebook with better compression
 than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
-and is supported by urllib3 if the `pyZstd <https://pypi.org/pyzstd>`_ package is installed.
+and is supported by urllib3 if the `python-zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
 You may also request the package be installed via the ``urllib3[zstd]`` extra:
 
 .. code-block:: bash
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -551,6 +551,11 @@ You may also request the package be installed via the ``urllib3[zstd]`` extra:
 
     $ python -m pip install urllib3[zstd]
 
+.. note::
+
+    Zstandard support in urllib3 requires using v0.18.0 or later of the ``zstandard`` package.
+    If the version installed is less than v0.18.0 then Zstandard support won't be enabled.
+
 Here's an example using zstd encoding via the ``Accept-Encoding`` header:
 
 .. code-block:: python
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -544,7 +544,7 @@ Zstandard Encoding
 `Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
 is a compression algorithm created by Facebook with better compression
 than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
-and is supported by urllib3 if the `zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
+and is supported by urllib3 if the `zstandard package <https://pypi.org/project/zstandard/>`_ is installed.
 You may also request the package be installed via the ``urllib3[zstd]`` extra:
 
 .. code-block:: bash
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -225,7 +225,7 @@ starts with ``http://`` instead of ``https://``:
 
      # Do this:
      http = urllib3.ProxyManager("http://...")
-
+     
      # Not this:
      http = urllib3.ProxyManager("https://...")
 
@@ -243,11 +243,11 @@ and not ``https://``:
      $ env | grep "_PROXY"
      HTTP_PROXY=http://127.0.0.1:8888
      HTTPS_PROXY=https://127.0.0.1:8888  # <--- This setting is the problem!
-
+     
      # Make the fix in your current session and test your script
      $ export HTTPS_PROXY="http://127.0.0.1:8888"
      $ python test-proxy.py  # This should now pass.
-
+     
      # Persist your change in your shell 'profile' (~/.bashrc, ~/.profile, ~/.bash_profile, etc)
      # You may need to logout and log back in to ensure this works across all programs.
      $ vim ~/.bashrc
@@ -452,7 +452,7 @@ server. For example if the server requires TLS 1.0 you'd configure urllib3 like 
 
     import ssl
     import urllib3
-
+    
     http = urllib3.PoolManager(
         ssl_minimum_version=ssl.TLSVersion.TLSv1
     )
@@ -500,7 +500,7 @@ understand the risks and wish to disable these warnings, you can use :func:`~url
 .. code-block:: python
 
     import urllib3
-
+    
     urllib3.disable_warnings()
 
 Alternatively you can capture the warnings with the standard :mod:`logging` module:
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -225,7 +225,7 @@ starts with ``http://`` instead of ``https://``:
 
      # Do this:
      http = urllib3.ProxyManager("http://...")
-     
+
      # Not this:
      http = urllib3.ProxyManager("https://...")
 
@@ -243,11 +243,11 @@ and not ``https://``:
      $ env | grep "_PROXY"
      HTTP_PROXY=http://127.0.0.1:8888
      HTTPS_PROXY=https://127.0.0.1:8888  # <--- This setting is the problem!
-     
+
      # Make the fix in your current session and test your script
      $ export HTTPS_PROXY="http://127.0.0.1:8888"
      $ python test-proxy.py  # This should now pass.
-     
+
      # Persist your change in your shell 'profile' (~/.bashrc, ~/.profile, ~/.bash_profile, etc)
      # You may need to logout and log back in to ensure this works across all programs.
      $ vim ~/.bashrc
@@ -449,10 +449,10 @@ For the best security it's a good idea to set this value to the version of TLS t
 server. For example if the server requires TLS 1.0 you'd configure urllib3 like so:
 
 .. code-block:: python
-    
+
     import ssl
     import urllib3
-    
+
     http = urllib3.PoolManager(
         ssl_minimum_version=ssl.TLSVersion.TLSv1
     )
@@ -500,7 +500,7 @@ understand the risks and wish to disable these warnings, you can use :func:`~url
 .. code-block:: python
 
     import urllib3
-    
+
     urllib3.disable_warnings()
 
 Alternatively you can capture the warnings with the standard :mod:`logging` module:
@@ -537,6 +537,32 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
         "https://www.google.com/",
         headers={"Accept-Encoding": "br"}
     )
+
+Zstandard Encoding
+---------------
+
+`Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
+is a compression algorithm created by Facebook with better compression
+than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
+and is supported by urllib3 if the `pyZstd <https://pypi.org/pyzstd>`_ package is installed.
+You may also request the package be installed via the ``urllib3[zstd]`` extra:
+
+.. code-block:: bash
+
+    $ python -m pip install urllib3[zstd]
+
+Here's an example using zstd encoding via the ``Accept-Encoding`` header:
+
+.. code-block:: python
+
+    import urllib3
+
+    urllib3.request(
+        "GET",
+        "https://facebook.com/",
+        headers={"Accept-Encoding": "zstd"}
+    )
+
 
 Decrypting Captured TLS Sessions with Wireshark
 -----------------------------------------------
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -544,7 +544,7 @@ Zstandard Encoding
 `Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
 is a compression algorithm created by Facebook with better compression
 than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
-and is supported by urllib3 if the `python-zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
+and is supported by urllib3 if the `zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
 You may also request the package be installed via the ``urllib3[zstd]`` extra:
 
 .. code-block:: bash
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -449,7 +449,7 @@ For the best security it's a good idea to set this value to the version of TLS t
 server. For example if the server requires TLS 1.0 you'd configure urllib3 like so:
 
 .. code-block:: python
-
+    
     import ssl
     import urllib3
     
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -539,7 +539,7 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
     )
 
 Zstandard Encoding
----------------
+------------------
 
 `Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
 is a compression algorithm created by Facebook with better compression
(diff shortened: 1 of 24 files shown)


sethmlarson:
Thanks for opening this! That's intriguing that you couldn't raise an error during streaming mode for `python-zstandard`. Of the [three projects listed on the Zstandard page](https://facebook.github.io/zstd/#other-languages) I'd been hoping we could make that one work. Is this something the maintainers of that project are aware of?

sethmlarson:
Also @mamico if you're not aware that issue is [eligible to be paid after completion](https://urllib3.readthedocs.io/en/latest/contributing.html#getting-paid-for-your-contributions). Are you interested in pursuing that? If so let us know and we can help with the rest of the process too!

mamico:
> Thanks for opening this! That's intriguing that you couldn't raise an error during streaming mode for `python-zstandard`. Of the [three projects listed on the Zstandard page](https://facebook.github.io/zstd/#other-languages) I'd been hoping we could make that one work. Is this something the maintainers of that project are aware of?

That's my problem:

```
>>> import zstandard
>>> ctx = zstandard.ZstdDecompressor()
>>> zfoo = zstandard.compress(b"foo")
```

decompressing all the data is fine:

```
>>> obj = ctx.decompressobj()
>>> obj.decompress(zfoo)
b'foo'
```

decompressing in streaming mode is fine too

```
>>> obj = ctx.decompressobj()
>>> b''.join([obj.decompress(good[i:i+1]) for i in range(len(zfoo))])
b'foo'
```

But, if I send something wrong (short), I am not able to know that there is a problem.

```
>>> obj = ctx.decompressobj()
>>> obj.decompress(b'bar')
b''
```

It is different if the data is enough

```
>>> obj = ctx.decompressobj()
>>> obj.decompress(b'fooooooooooooo')
zstd.ZstdError: zstd decompressor error: Unknown frame descriptor
```

Instead, `pyzstd` provides the `at_frame_edge` and` needs_input` flags which help to recognize problems.

If you think that is ok to change the test and raise exception only when there is enough data, it's easy for me to change the library. 

We could also leave the current implementation as a fallback: so that, for example, if you installed `pyzstd` it still works, but `python-zstandard` is the supported library.

sethmlarson:
@mamico Looked at the issue tracker for python-zstandard, would this feature request be the one needed to implement this? https://github.com/indygreg/python-zstandard/issues/161

src/urllib3/response.py
```diff
@@ -148,6 +153,24 @@ def flush(self) -> bytes:
...
 
 
+if zstd is not None:
+
+    class ZstdDecoder(ContentDecoder):
+        def __init__(self) -> None:
+            self._obj = zstd.EndlessZstdDecompressor()
+
+        def decompress(self, data: bytes) -> bytes:
+            return self._obj.decompress(data) if data else data  # type: ignore[no-any-return]
```
sethmlarson, line 163: This line seems more complex than it needs to be, maybe we can extract the empty case into it's own block for better readability?

sethmlarson (commented):
This PR looks great content-wise, here's a handful of comments for you to get this one across the finish line once when we decide on the python-zstandard issue:

pquentin:
> @mamico Looked at the issue tracker for python-zstandard, would this feature request be the one needed to implement this? [indygreg/python-zstandard#161](https://github.com/indygreg/python-zstandard/issues/161)

Given the thumbs up on that comment I thought that it was the issue, so I mentioned it to @indygreg. He kindly implemented this feature, apparently spending multiple hours on the project to help adoption here. I tried his fix and it did not help. Looking more closely, I don't think `unused_data` is what we're looking for:

```
>>> import zstandard as zstd
>>> zstd.__version__
'0.18.0.dev0'
>>> obj = zstd.ZstdDecompressor().decompressobj()
>>> obj.decompress(b"foo")
b''
>>> obj.unused_data
b''
```

@indygreg Please accept my apologies for requesting work that was not ultimately useful to us. :(

To make sure that decompressing b"foo" fails like in the other backends, we can address this in two places:

 * python-zstandard: ask to raise an exception for this case, like all other backends do (adding the same attributes as pyzstd seems like an inferior solution to me)
 * urllib3: argue that the test is not realistic, and only test a longer string like `b"foooooooooooooo"` is fine

To avoid requesting more work to @indygreg (and because there must be a reason why the zstd libraries don't raise in this case), I think using a longer test string is fine, achieves the same result in practice, and allows us to use the better Python Zstandard library.

mamico:
I've changed the library used from `pyzstd` to `python-zstandard` and made changes according to the review (thank for detailed work).

src/urllib3/response.py
```diff
@@ -148,6 +153,22 @@ def flush(self) -> bytes:
...
+
+    class ZstdDecoder(ContentDecoder):
+        def __init__(self) -> None:
+            self._obj = zstd.ZstdDecompressor().decompressobj()
+
+        def decompress(self, data: bytes) -> bytes:
+            if data == b"":
+                return b""
+            else:
+                return self._obj.decompress(data)  # type: ignore[no-any-return]
```
pquentin, line 166: For consistency with other decoders, can we please use an early return and check for all falsy values?

```suggestion
            if not data:
                return data

            return self._obj.decompress(data)  # type: ignore[no-any-return]
```

test/__init__.py
```diff
@@ -146,6 +151,18 @@ def notBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
     )
 
 
+def onlyZstd() -> Callable[[_TestFuncT], _TestFuncT]:
+    return pytest.mark.skipif(
+        zstd is None, reason="only run if zstandard library is present"
```
pquentin, line 156: Unlike brotlipy, we only support one Zstandard library:

```suggestion
        zstd is None, reason="only run if python-zstandard is installed"
```

test/__init__.py
```diff
@@ -146,6 +151,18 @@ def notBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
...
+    )
+
+
+def notZstd() -> Callable[[_TestFuncT], _TestFuncT]:
+    return pytest.mark.skipif(
+        zstd is not None, reason="only run if a zstandard library is absent"
```
pquentin, line 162: Unlike brotlipy, we only support one Zstandard library:

```suggestion
        zstd is None, reason="only run if python-zstandard is installed"
```

test/test_response.py
```diff
@@ -254,6 +254,38 @@ def test_decode_brotli_error(self) -> None:
...
+                break
+        assert ret == b"foobarbaz"
+
+    @onlyZstd()
+    def test_decode_zstd_error(self) -> None:
+        fp = BytesIO(b"fooooooooooooooo")
+
+        with pytest.raises(DecodeError):
+            HTTPResponse(fp, headers={"content-encoding": "zstd"})
+
```
pquentin, line 288: Can we use pytest.mark.paramettrize with marks as done in `test_make_headers` to avoid duplicating those tests a fourth time? (Could be a follow-up PR if you prefer.)

pquentin (changes requested):
Thanks! This looks good. I left a few comment to make sure [the contribution is polished](https://sage.thesharps.us/2014/09/01/the-gentle-art-of-patch-review/).

test/test_response.py
mamico: Ok, for a follow-up.

docs/advanced-usage.rst
```diff
@@ -538,6 +538,32 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
...
     )
 
+Zstandard Encoding
+------------------
+
+`Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
+is a compression algorithm created by Facebook with better compression
+than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
+and is supported by urllib3 if the `python-zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
+You may also request the package be installed via the ``urllib3[zstd]`` extra:
```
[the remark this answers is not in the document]
sethmlarson: Brotli uses its Content-Encoding value `br` so we could do `zstd` to continue that tradition.

indygreg:
I'm unsure how much work it would be, but if you care about achieving peak performance, I would recommend not using `zstandard`'s stdlib compatible API for decompression (`dctx.decompressionobj()`). Instead, something like the _stream reader_ interface (https://python-zstandard.readthedocs.io/en/latest/decompressor.html#zstddecompressionreader) would be better.

The reason is because the stdlib API requires excessive memory allocations and copies. Since the `decompress()` API doesn't know the final output size, it can't allocate a fixed output buffer for the return value. Instead, it allocates several buffers and concatenates them to return a final value. With APIs like `read(n)`, you know how many bytes are requested and can decompress directly into the buffer that is ultimately returned. For pathological cases, the overhead of the excessive buffer allocations and copies with the stdlib API can be noticeable.

Modern versions of Python do define an optional `max_length` argument to `decompress()` to address this issue. But I don't have it implemented in `zstandard` yet and I suspect a lot of software in the wild doesn't use it in order to maintain compatibility with older Python versions while keeping source code simpler. The `io.RawIOBase` interface is a better model for applying filtering on a stream anyway.

Unfortunately, it looks like `urllib3`'s `ContentDecoder` interface uses the inefficient stdlib decompression API design. So porting away is probably non-trivial. But I thought I'd make you aware of this known performance issue. It's really too bad the compression APIs in the standard library are poorly designed from a performance perspective.

mamico:
> I'm unsure how much work it would be, but if you care about achieving peak performance, I would recommend not using `zstandard`'s stdlib compatible API for decompression (`dctx.decompressionobj()`). 

@indygreg what do you think about https://github.com/indygreg/python-zstandard/pull/141 proposed by @animalize (the author of pyzstd)?  Can’t that be the solution to use a simple API with good performance?

animalize:
> But, if I send something wrong (short), I am not able to know that there is a problem.

In `zstandard`, the same problem is that when the data is incomplete, there is no way to know.

Another problem is when there are mutiple frames, only the first is decompressed.
But this can be supported in the future.

> The reason is because the stdlib API requires excessive memory allocations and copies.

`zstandard` is using `.write()` with `"y#"` format, it generates a `bytes` intermediate object (extra copy).

In `pyzstd`, `compress_stream()` and `decompress_stream()` use a `memoryview` to avoid extra copy when writing. You may improve this.

> Since the decompress() API doesn't know the final output size, it can't allocate a fixed output buffer for the return value. Instead, it allocates several buffers and concatenates them to return a final value.

If the decompressed size is unknown, and user needs to get the whole decompressed data as a `bytes` object, they will also implement such function themselves, but it may be less efficient and error-prone.

test/__init__.py
```diff
@@ -151,15 +151,16 @@ def notBrotli() -> Callable[[_TestFuncT], _TestFuncT]:
     )
 
 
-def onlyZstd() -> Callable[[_TestFuncT], _TestFuncT]:
+def hasZstd() -> Callable[[_TestFuncT], _TestFuncT]:
```
pquentin, line 154: Why change `only` to `has`? `only` was fine, and consistent with other fixtures like `onlySecureTransport`

mamico: Changed back to `only`.

animalize:
> In zstandard, the same problem is that when the data is incomplete, there is no way to know.

This statement is wrong, it has a `.eof` flag.
You can use it to raise exception:
```python
        def flush(self) -> bytes:
            ret = self._obj.flush()
            if not self._obj.eof:
                raise Exception()
            return ret
```

pquentin:
Good find @animalize!

@indygreg OK now we have a good reason to ask for a release, I confirmed that the code above lets us raise correctly with `b"foo"` as input (which after all seemed important to us as bad settings and short responses are common on the web). The `eof` attribute was only added when you worked on `unused_data` though: https://github.com/indygreg/python-zstandard/commit/1dc39cfea821893268176dae754355ec19609ce0.

pquentin:
Thanks! Can you please upload a source distribution too to fix the PyPy and CPython 3.11 builds? (They were working with 0.17.0)

src/urllib3/response.py
```diff
@@ -148,6 +153,24 @@ def flush(self) -> bytes:
...
+
+        def decompress(self, data: bytes) -> bytes:
+            if not data:
+                return b""
+            return self._obj.decompress(data)  # type: ignore[no-any-return]
+
+        def flush(self) -> bytes:
+            ret = self._obj.flush()
+            if not self._obj.eof:
+                raise DecodeError("Invalid zstandard data")
```
animalize, line 170: Strictly speaking, the message can be this: zstd data ends in an incomplete frame.

sethmlarson (commented):
One thought: since we can't guarantee correctness using zstandard<0.17 we should update our version ranges for zstandard in the `zstd` extra to be `>=0.18` and check `zstandard.__version__` as a part of detecting whether we expose `ZstdDecoder`. Add this to the documentation as well.

indygreg:
> Thanks! Can you please upload a source distribution too to fix the PyPy and CPython 3.11 builds? (They were working with 0.17.0)

0.18.0 sdist uploaded.

src/urllib3/response.py
animalize: It's a little long, how about "zstd data is incomplete"?

animalize:
Should `DeflateDecoder` / `GzipDecoder` / `BrotliDecoder` check `.eof` flag as well?
In current code, they don't check it.

sethmlarson:
@animalize I'm not sure! Maybe those decoders not checking for `eof` and us receiving no reports of issues is a sign users aren't receiving truncated streams? Our usage of `eof` in ZstdDecoder is to work-around the [issue you've raised](https://github.com/facebook/zstd/issues/3169) but I don't think we're as worried about partial streams there as we are about any "amount of data that's definitely not zstandard" not raising an error.

sethmlarson:
@mamico Went ahead and push my review comments since they were very small and the only outstanding work to get this one past the finish line. Let me know if you have concerns here.

docs/advanced-usage.rst
```diff
@@ -538,6 +538,37 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
         headers={"Accept-Encoding": "br"}
     )
 
+Zstandard Encoding
+------------------
+
+`Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
+is a compression algorithm created by Facebook with better compression
+than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
+and is supported by urllib3 if the `python-zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
```
pquentin, line 547: Realized thanks to the note below that we should call this package `zstandard` after its name on PyPI.

```suggestion
and is supported by urllib3 if the `zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
```

image

docs/advanced-usage.rst
```diff
@@ -538,6 +538,37 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
...
+
+Here's an example using zstd encoding via the ``Accept-Encoding`` header:
+
+.. code-block:: python
+
+    import urllib3
+
+    urllib3.request(
+        "GET",
+        "https://facebook.com/",
```
pquentin, line 567: Let's avoid a redirect:

```suggestion
        "https://www.facebook.com/",
```

docs/reference/urllib3.response.rst
```diff
@@ -30,3 +30,4 @@ representation.
 .. autoclass:: urllib3.response.DeflateDecoder
 .. autoclass:: urllib3.response.GzipDecoder
 .. autoclass:: urllib3.response.MultiDecoder
+.. autoclass:: urllib3.response.ZstdDecoder
```
pquentin, line 33: I think having ZstdDecoder before MultiDecoder would be nicer.

image

pquentin (approved):
I left a few nits, please merge after you've considered them.

docs/advanced-usage.rst
```diff
@@ -538,6 +538,37 @@ Here's an example using brotli encoding via the ``Accept-Encoding`` header:
         headers={"Accept-Encoding": "br"}
     )
 
+Zstandard Encoding
+------------------
+
+`Zstandard <https://datatracker.ietf.org/doc/html/rfc8878>`_
+is a compression algorithm created by Facebook with better compression
+than brotli, gzip and deflate (see `benchmarks <https://facebook.github.io/zstd/#benchmarks>`_)
+and is supported by urllib3 if the `zstandard <https://pypi.org/project/zstandard/>`_ package is installed.
```
sethmlarson, line 547: Sphinx doesn't like having two links with the same text, maybe we include the word "package" within the hyperlink text here?

sethmlarson:
The failure on macOS+Python 3.10 looks unrelated, an exit code of `-14` but not during a test case that changed. Curious. I don't think we need to block this PR on that failure.

sethmlarson (approved):
This looks wonderful, thanks to everyone who contributed to this PR! @mamico @animalize @indygreg :bow:

PR #2626, opened 2022-06-09
Make GitHub Actions run tests with OpenSSL 3.0 on Ubuntu 22.04

### LINKED ISSUE #2477: Understand OpenSSL 3.0 failures
OpenSSL 3 is supported by Python, we should try running urllib3's test suite against Python using OpenSSL 3.

Easiest route might be to use Docker in GitHub Actions with Fedora's "rawhide" image:
- https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-actions
- https://hub.docker.com/_/fedora

### Minimum requirements

- [ ] Add to the `ci` GitHub Action in the `main` branch to test against a platform with access to OpenSSL 3.
- [x] Create a pytest marker `@requires_openssl_lt_3` which adds `pytest.mark.xfail` to tests when Python is compiled with OpenSSL >= 3.0.0
- [x] For each failing test on CI figure out why the failure is occurring, document each failure with a comment and add the above pytest marker to the test.
- [x] Create a GitHub issue which summarizes all the problems that are currently failing tests

illia-v:
This makes GitHub Actions run tests on both Ubuntu 20.04 (OpenSSL 1.1.1) and Ubuntu 22.04 (OpenSSL 3).

Resolves #2477.

---

BTW, `matrix.include` and `matrix.exclude` do not support lists, I tried.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
             experimental: true
             nox-session: test-3.11
         exclude:
-          # Ubuntu 22.04 comes with OpenSSL 3.0, so only CPython => 3.9 is compatible with it
+          # Ubuntu 22.04 comes with OpenSSL 3.0, so only CPython >= 3.9 is compatible with it
           # https://github.com/python/cpython/issues/83001
           - python-version: "3.7"
             os: ubuntu-22.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
             experimental: true
             nox-session: test-3.11
         exclude:
-          # Ubuntu 22.04 comes with OpenSSL 3.0, so only CPython >= 3.9 is compatible with it
+          # Ubuntu 22.04 comes with OpenSSL 3.0, so only CPython 3.9+ is compatible with it
           # https://github.com/python/cpython/issues/83001
           - python-version: "3.7"
             os: ubuntu-22.04
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,6 +93,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
+          cache: pip
 
       - name: Set Up Python (Development version) - ${{ matrix.python-version }}
         uses: deadsnakes/action@v2.0.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,11 +65,11 @@ jobs:
             os: ubuntu-20.04  # CPython 3.9.2 is not available for ubuntu-22.04.
             experimental: false
             nox-session: test-3.9
-          - python-version: 3.11-dev
+          - python-version: "3.11.0-beta - 3.11.0"
             os: ubuntu-20.04
             experimental: true
             nox-session: test-3.11
-          - python-version: 3.11-dev
+          - python-version: "3.11.0-beta - 3.11.0"
             os: ubuntu-22.04
             experimental: true
             nox-session: test-3.11
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,11 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        os: [macos-latest, windows-latest, ubuntu-latest]
+        os:
+          - macos-latest
+          - windows-latest
+          - ubuntu-20.04  # OpenSSL 1.1.1
+          - ubuntu-22.04  # OpenSSL 3.0
         experimental: [false]
         nox-session: ['']
         include:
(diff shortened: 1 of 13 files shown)


nateprewitt:
Looking at the logs for the successful 3.0 build with brotli we see:

```
cryptography's compiled against OpenSSL: OpenSSL 1.1.1l  24 Aug 2021
cryptography's linked OpenSSL: OpenSSL 1.1.1l  24 Aug 2021
Python's OpenSSL: OpenSSL 3.0.2 15 Mar 2022
```

Do we have a concrete way to verify we're actually using OpenSSL 3.0 for everything?

illia-v:
@nateprewitt that's a very good point, thanks.

I made Github Actions compile cryptography, this is fixing the failed run too 🎉 
CPython runs on Linux became slower, but they are still faster than PyPy. So I hope this update is acceptable.

noxfile.py
```diff
@@ -19,7 +20,19 @@ def tests_impl(
     extras: str = "socks,secure,brotli",
     byte_string_comparisons: bool = True,
 ) -> None:
+    is_github_actions = "GITHUB_ACTIONS" in os.environ
+
     # Install deps and the package itself.
+    # Compile cryptography with the system OpenSSL and libffi libraries
+    # on Linux CPython runs of GitHub Actions instead of downloading its
+    # wheel.
+    is_cpython_on_linux = (
```
sethmlarson, line 29: I like the direction this is heading, I wonder if there's a way to detect if we're running against a Python with OpenSSL 3.0 (maybe `python -c 'import ssl; print(ssl.OPENSSL_VERSION[0])'` and capturing the output with `nox.Session.run(..., stdout=...)`? Then if we detect OpenSSL 3.0 we run `session.install("--no-binary=:all:", "cryptography")` to install from source as you're already doing now.

noxfile.py
```diff
@@ -19,7 +20,19 @@ def tests_impl(
     extras: str = "socks,secure,brotli",
     byte_string_comparisons: bool = True,
 ) -> None:
+    is_github_actions = "GITHUB_ACTIONS" in os.environ
```
sethmlarson, line 23: Since we likely want the same installs locally and in GitHub Actions I don't think we need to change our behavior whether we're in CI or not.

sethmlarson (commented):
Thanks for creating this PR, it's already in really great shape! A few comments for you:

A potential solution for Nate's suggestion, perhaps we need to examine the output of `python -m OpenSSL.debug` and `ssl.OPENSSL_VERSION` to make sure they're matching at least? Still leaves the whole "make sure it's OpenSSL 3.0" issue open but consistency between ssl and cryptography modules is a start.

Also if you're interested in being paid for this contribution once it's merged and have questions about that process let us know, we're happy to help! :)

.github/workflows/ci.yml
```diff
@@ -49,21 +53,36 @@ jobs:
...
-            os: ubuntu-latest
+            os: ubuntu-20.04
+            experimental: true
+            nox-session: test-3.11
+          - python-version: 3.11-dev
+            os: ubuntu-22.04
             experimental: true
             nox-session: test-3.11
+        exclude:
+          # Ubuntu 22.04 comes with OpenSSL 3.0, so only CPython => 3.9 is compatible with it
```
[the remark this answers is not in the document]
illia-v: Sure, that was a typo 🙂

illia-v:
I've just discovered that PyPI wheel of cryptography 37.0.2 is compiled against OpenSSL 3.0.3. I think most end users use it together with Python compiled against OpenSSL 1.1.1, it works and cryptography seems not to be concerned about the inconsistency. Therefore, maybe it is better to stick with the wheel instead of compiling the package and accept inconsistency in some environments. What do you think?

nateprewitt:
Yeah, I think that seems fair.

sethmlarson:
@illia-v I think 3.11-dev works differently on GHA than it does on pyenv. It's a symlink to the latest release including prereleases instead of the latest git commit on the 3.11 branch.

illia-v:
> @illia-v I think 3.11-dev works differently on GHA than it does on pyenv. It's a symlink to the latest release including prereleases instead of the latest git commit on the 3.11 branch.

I am not sure I understand what you mean. But this message led me to unset `nox-session` so that 3.11 is consistent with 3.7-3.10 and it fixed the error 🙂

illia-v:
Hmm, the error was fixed in a run in my repo https://github.com/illia-v/urllib3/runs/6843925665?check_suite_focus=true, but failed here again. This is very weird, I'll continue the investigation later.

sethmlarson:
@illia-v Try removing the "cache: pip" line in the test Github Action? I've seen something similar to this to do with caching.

illia-v:
> @illia-v Try removing the "cache: pip" line in the test Github Action? I've seen something similar to this to do with caching.

Nice, this fixed the error. Thanks for the tip!

Cache was restored from the same key in Ubuntu 20.04 and 22.04: `setup-python-Linux-python-3.11.0-beta.3-pip-60d374041db12689c3d346e149aa43744f06aa781ab1c8cbbe70e0b09b98dd70`.

Cryptography requires the cffi package. The latter does not have a wheel for CPython 3.11 on PyPI. So a CI job in Ubuntu 20.04 built and cached it, that is why the 22.04 job restoring the cache was failing with `ImportError: libffi.so.7: cannot open shared object file: No such file or directory`.

sethmlarson (dismissed):
I've added back the cache directive now that the cache is busted. Might be worth creating an issue on the `actions/setup-python` repo for adding the OS and version to the cache key to avoid these kinds of problems.

I'm adding my approval to this approach but would like @pquentin's review before we merge.

sethmlarson:
Hmm, cache must not be busted after all? Wish there was a better way to clear it. For now @illia-v  you can force push your branch without my commit so we can get back to things passing.

illia-v:
@sethmlarson I do not think the cache has been busted.
But even if it was, the error would happen the next time a new cache was restored. The same cache archive is shared by both Ubuntu 20.04 and 22.04 jobs of the same Python version, and cached CFFI needs a specific version of shared libffi that is different in the Ubuntu versions (Ubuntu 20.04 has `libffi.so.7`, Ubuntu 22.04 has `libffi.so.8`.)

This error does not happen for Python <= 3.11 because PyPI provides a CFFI wheel that does not depend on the shared library.

sethmlarson (approved):
LGTM

.github/workflows/ci.yml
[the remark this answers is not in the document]
pquentin: nit on the nit: I like the "CPython 3.9+" convention, that way grepping for "Python 3.9" will return this line

.github/workflows/ci.yml
```diff
@@ -49,21 +53,36 @@ jobs:
             experimental: false
             nox-session: unsupported_setup_py
           - python-version: "3.9"
-            os: ubuntu-latest
+            os: ubuntu-20.04
+            experimental: false
+            nox-session: test_brotlipy
```
pquentin, line 58: In the interest of reducing our test matrix to the minimum, I think we should only test brotli with one configuration, as brotli has nothing to do with OpenSSL.

pquentin (changes requested):
Thank you for your hard work! This is looking great. I like that thanks to the collaboration here we ended up with something much smaller.

This is looking great, I just have one comment and one nit.

Also, it's too bad that we don't measure the coverage of our test code, otherwise it would have allowed us to encode that OpenSSL 1.1.1 and 3.0.0 should be tested.

.github/workflows/ci.yml
illia-v: Done

pquentin (approved):
Thanks for iterating! This now looks good to me.

Since this changes the jobs that run in CI, I can't merge this. @sethmlarson will have to update the required jobs. And then we'll have to update existing PRs to include this change.

sethmlarson (approved):
Looks great, thanks for the review @pquentin! Let's get this merged.

sethmlarson:
@illia-v If you're eligible to be paid for your contribution you can submit an expense to [the urllib3 Open Collective](https://opencollective.com/urllib3). Let us know if you have questions about this process. Thanks again!

illia-v:
@pquentin @sethmlarson thanks for reviews!

> @illia-v If you're eligible to be paid for your contribution you can submit an expense to [the urllib3 Open Collective](https://opencollective.com/urllib3)

Thanks, I'll try.

PR #2627, opened 2022-06-09
Consistently wrap errors in load_cert_chain

sfreilich:
This wraps OpenSSL.SSL.Error with ssl.SSLError in
PyOpenSSLContext.load_cert_chain, similar to the error handling in
other methods of PyOpenSSLContext.

### DIFF
--- a/changelog/2628.bugfix.rst
+++ b/changelog/2628.bugfix.rst
@@ -1,0 +1,1 @@
+Wrap OpenSSL.SSL.Error with ssl.SSLError in PyOpenSSLContext.load_cert_chain.
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -486,12 +486,15 @@ class PyOpenSSLContext:
         keyfile: Optional[str] = None,
         password: Optional[str] = None,
     ) -> None:
-        self._ctx.use_certificate_chain_file(certfile)
-        if password is not None:
-            if not isinstance(password, bytes):
-                password = password.encode("utf-8")  # type: ignore[assignment]
-            self._ctx.set_passwd_cb(lambda *_: password)
-        self._ctx.use_privatekey_file(keyfile or certfile)
+        try:
+            self._ctx.use_certificate_chain_file(certfile)
+            if password is not None:
+                if not isinstance(password, bytes):
+                    password = password.encode("utf-8")  # type: ignore[assignment]
+                self._ctx.set_passwd_cb(lambda *_: password)
+            self._ctx.use_privatekey_file(keyfile or certfile)
+        except OpenSSL.SSL.Error as e:
+            raise ssl.SSLError(f"Unable to load certificate chain: {e!r}") from e
 
     def set_alpn_protocols(self, protocols: List[Union[bytes, str]]) -> None:
         protocols = [util.util.to_bytes(p, "ascii") for p in protocols]
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -352,21 +352,21 @@ class TestClientCerts(SocketDummyServerTestCase):
     def test_load_keyfile_with_invalid_password(self) -> None:
         assert ssl_.SSLContext is not None
         context = ssl_.SSLContext(ssl_.PROTOCOL_SSLv23)
-
-        # Different error is raised depending on context.
-        if ssl_.IS_PYOPENSSL:
-            from OpenSSL.SSL import Error  # type: ignore[import]
-
-            expected_error = Error
-        else:
-            expected_error = ssl.SSLError
-
-        with pytest.raises(expected_error):
+        with pytest.raises(ssl.SSLError):
             context.load_cert_chain(
                 certfile=self.cert_path,
                 keyfile=self.password_key_path,
                 password=b"letmei",
             )
+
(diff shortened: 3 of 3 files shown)


sfreilich:
Sorry, I thought I'd fixed that typo before, and not sure how it got back into my local checkout before push.

pquentin (commented):
Thanks! This looks good. Can you please add a changelog entry? If you need that soon, you'll also have to backport to the 1.26.x branch.

nateprewitt:
It looks like we're consistently failing to raise an error with the new test on macOS. That may need a quick check on root cause. If we're not concerned, we may skip that test for the darwin platform.

sfreilich:
Filed issue #2628. (But I shouldn't just used the PR number? Sorry just a bit tired and getting the instructions mixed up.)

Noticed there's a typo in my change description, too.

sfreilich:
@nateprewitt Interesting, is there a more consistent way to get the underlying error on Mac? Otherwise, how should I skip the test on that platform?

nateprewitt:
It looks like macOS may not be raising an error at all:

```python
_________________ TestClientCerts.test_load_invalid_cert_file __________________
  File "/Users/runner/work/urllib3/urllib3/test/with_dummyserver/test_socketlevel.py", line 367, in test_load_invalid_cert_file
    certfile=self.password_key_path,
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-7/lib/python3.7/site-packages/_pytest/python_api.py", line 952, in __exit__
    fail(self.message)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-7/lib/python3.7/site-packages/_pytest/outcomes.py", line 1[96](https://github.com/urllib3/urllib3/runs/6821420445?check_suite_focus=true#step:7:97), in fail
    raise Failed(msg=reason, pytrace=pytrace)
Failed: DID NOT RAISE <class 'ssl.SSLError'>
```

We'd need to do some quick testing to confirm if the test doesn't raise an error or if it raises a different error. I'm assuming it's the former but it would be good to check. Once we have that answer, the next question is should `load_cert_chain` be raising an error? I don't immediately know the answer for that without some digging.

If we do find this is expected behavior, we can add a skip similar to [this](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1) based on the host platform.

i.e.
```python
@pytest.mark.skipif(sys.platform == 'darwin', reason='No error raised on macOS')
```

sethmlarson:
@sfreilich Thanks for creating this. I've asked our team for anyone with macOS to give the test suite a spin to see if they can see the error since this is a SecureTransport test.

sethmlarson:
Oh right, our SecureTransport `load_cert_chain` call is lazy and waits until we're doing a handshake. Let's skip the test on macOS if @nateprewitt and @pquentin are okay with this approach?

sethmlarson:
@sfreilich Your latest commit skips the test on macOS completely, however we actually only want to skip when SecureTransport is in use as the test should run and pass on macOS w/ Python's ssl and pyOpenSSL. We have the [`notSecureTransport` pytest annotation](https://github.com/urllib3/urllib3/blob/main/test/__init__.py#L158) for this purpose, could you change to using this instead?

sfreilich:
Updated my patch. I think skipping the test in Mac sounds reasonable if the underlying library is just deferring the validation that would raise an exception in that spot. The eventual call to cnx.do_handshake() is already covered by similar error-wrapping logic.

sfreilich:
@sethmlarson Missed your comment when I wrote my last. Done.

nateprewitt (dismissed):
Great, thanks @sfreilich! This looks good from my concern.

nateprewitt:
Looks like the PR may need another quick rebase after our merges this morning.

sethmlarson:
@nateprewitt You can squash this once it's passing CI, but remove my name from the commit please :) Didn't realize GitHub's rebase mode would add the rebase button pusher to every commit.

nateprewitt:
Hmm, it's not clear squashing will remove you from the attribution, @sethmlarson. It doesn't have the normal co-author reference in the extended comment. I can merge as is and we hope, or we'll need @sfreilich to rebase themselves.

sfreilich:
I tried, but it doesn't seem to have done anything, probably because this branch already has those commits. Not sure what I can do to fix that.

nateprewitt:
You can try squashing them yourself and force pushing if you're comfortable doing that. Otherwise, we can just merge at this point.

sfreilich:
@nateprewitt Done. Needs reapproval.

nateprewitt (approved):
Perfect, thanks again @sfreilich!

sethmlarson:
Thanks @nateprewitt and @sfreilich for fighting through the mess I made ;)

pquentin:
Does someone understand why a change in the pyOpenSSL backend affected the SecureTransport tests? This still baffles me.

sfreilich:
The code change doesn't effect the SecureTransport tests. I added a new test that asserts that errors in load_cert_chain are consistently raised. (And also modified an existing test that had been working around the inconsistency to expect consistent behavior, as it should have.) In my first draft, the new test was applied to all the SSL backends. But it failed for SecureTransport because the underlying library defers some validation, so it doesn't raise an exception in the situation the new test asserts about. So the new test was added just for the other backends.

pquentin:
Ok I get it now, thanks for the explanation!

PR #2642, opened 2022-06-17
Fix `test_default_ssl_context_ssl_min_max_versions` on Fedora 36

illia-v:
This fixes an OS-specific error found in https://github.com/urllib3/urllib3/issues/2477#issuecomment-1151381159 and explained in https://github.com/urllib3/urllib3/issues/2477#issuecomment-1151452150.


sethmlarson (approved):
LGTM, thanks for the follow-up on this! :rocket:

Issue #2513, opened 2022-01-02
OverflowError: signed integer is greater than maximum


sethmlarson:
Both of the below calls result in an `OverflowError` being raised because the `Content-Length` is longer than 3GB. OpenSSL only supports reading (2**31)-1 bytes in a single read call.

```python
url = "https://dailymed-data.nlm.nih.gov/public-release-files/dm_spl_release_human_rx_part1.zip"

resp = http.request("GET", url)
resp.data

resp = http.request("GET", url, preload_content=False)
list(resp.stream(None))
```

```
Traceback (most recent call last):
  File "/home/sethmlarson/urllib3/testme.py", line 5, in <module>
    resp = http.request("GET", "https://dailymed-data.nlm.nih.gov/public-release-files/dm_spl_release_human_rx_part1.zip")
  File "/home/sethmlarson/urllib3/src/urllib3/__init__.py", line 112, in request
    return _DEFAULT_POOL.request(
  File "/home/sethmlarson/urllib3/src/urllib3/_request_methods.py", line 109, in request
    return self.request_encode_url(
  File "/home/sethmlarson/urllib3/src/urllib3/_request_methods.py", line 142, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "/home/sethmlarson/urllib3/src/urllib3/poolmanager.py", line 443, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/home/sethmlarson/urllib3/src/urllib3/connectionpool.py", line 744, in urlopen
    response = self.ResponseCls.from_httplib(
  File "/home/sethmlarson/urllib3/src/urllib3/response.py", line 748, in from_httplib
    resp = ResponseCls(
  File "/home/sethmlarson/urllib3/src/urllib3/response.py", line 467, in __init__
    self._body = self.read(decode_content=decode_content)
  File "/home/sethmlarson/urllib3/src/urllib3/response.py", line 664, in read
    data = self._fp.read() if not fp_closed else b""
  File "/home/sethmlarson/.pyenv/versions/3.8.6/lib/python3.8/http/client.py", line 471, in read
    s = self._safe_read(self.length)
  File "/home/sethmlarson/.pyenv/versions/3.8.6/lib/python3.8/http/client.py", line 612, in _safe_read
    data = self.fp.read(amt)
  File "/home/sethmlarson/.pyenv/versions/3.8.6/lib/python3.8/socket.py", line 669, in readinto
[...]

PR #2647, opened 2022-06-20
Test for OpenSSL read OverflowError

justlund:
Test case for https://github.com/urllib3/urllib3/issues/2513

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1496,7 +1496,7 @@ class TestSSL(SocketDummyServerTestCase):
         """
         Ensure OpenSSL read doesn't fail with OverflowError when reading > (2**31)-1 bytes.
         """
-        close_event = Event()
+        done_event = Event()
 
         def socket_handler(listener: socket.socket) -> None:
             sock = listener.accept()[0]
@@ -1514,9 +1514,9 @@ class TestSSL(SocketDummyServerTestCase):
                 b"Content-Length: 2147483648\r\n"
                 b"\r\n"
             )
+            done_event.wait(1)
             ssl_sock.close()
             sock.close()
-            close_event.set()
 
         self._start_server(socket_handler)
         with HTTPSConnectionPool(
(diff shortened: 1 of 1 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1491,6 +1492,44 @@ def socket_handler(listener: socket.socket) -> None:
...
+                ca_certs=DEFAULT_CA,
+            )
+
+            sock.send(
+                b"HTTP/1.1 200 OK\r\n"
+                b"Content-type: text/plain\r\n"
+                b"Content-Length: 2147483648\r\n"
+                b"\r\n"
+            )
+            sock.close()
```
sethmlarson, line 1516: This test might be failing due to CI running at different speeds than locally. Might require waiting for an `Event` object to run `sock.close()` as is being done in other places.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1491,6 +1492,49 @@ def socket_handler(listener: socket.socket) -> None:
...
+            cert_reqs="REQUIRED",
+            ca_certs=DEFAULT_CA,
+        ) as pool:
+            with pytest.raises(urllib3.exceptions.ProtocolError) as e:
+                pool.request("GET", "/", retries=False)
+
+            assert (
+                str(e.value)
+                == "('Connection broken: IncompleteRead(0 bytes read, 2147483648 more expected)', IncompleteRead(0 bytes read, 2147483648 more expected))"
+            )
```
illia-v, line 1534:
```suggestion
            assert isinstance(e.value.args[1], ConnectionResetError)
```

illia-v (commented):
Your test should fail only on CPython 3.8 and 3.9.2 workflow runs because those are affected CPython versions (https://github.com/urllib3/urllib3/issues/2513#issuecomment-1152559900.)

The suggestion seems to fix the test on other versions and even on 3.8 and 3.9.2 when this branch is merged with #2657.

PR #2649, opened 2022-06-21
Refactor internal APIs to use our own HTTPResponse

### LINKED ISSUE #2648: HTTPConnection should return urllib3.HTTPResponse instead of http.client.HTTPResponse
Part of https://github.com/urllib3/urllib3/issues/1985

Currently our `urllib3.connection.HTTPConnection` classes return directly a `http.client.HTTPResponse`. This ties our APIs directly to `http.client` which isn't desirable for future additions of new HTTP protocol implementations.

Instead our `HTTPConnection` APIs should return an instance of `urllib3.response.HTTPResponse` directly and should be forwarded upwards instead of doing the wrapping inside of `HTTPConnectionPool.urlopen`.

#### Minimum requirements

- [ ] Research and understand which exceptions can be raised at which stage from `HTTPConnection.urlopen` and `HTTPConnection.getresponse` in both urllib3 and http.client. This research should be posted into this issue.
- [ ] Move the wrapping of `http.client.HTTPResponse` from `HTTPConnectionPool` to `HTTPConnection`
- [ ] Update type hints for `HTTPConnection`
- [ ] Add test cases for `HTTPConnection` directly returning a `urllib3.response.HTTPResponse`.
- [ ] Change existing test cases for the new logic.

shadycuz:
Fixes #2648 

Minimum requirements
- [x] Move the wrapping of http.client.HTTPResponse from HTTPConnectionPool to HTTPConnection
- [x] Update type hints for HTTPConnection
- [x] Add test cases for HTTPConnection directly returning a urllib3.response.HTTPResponse.
- [x] Change existing test cases for the new logic.


src/urllib3/connection.py
```diff
@@ -369,6 +370,22 @@ def request_chunked(
...
+        httplib_response = super().getresponse()
+
+        # Wrap http.client.HTTPResponse as a urllib3.response.HTTPResponse
+        response = HTTPResponse.from_httplib(
+            httplib_response,
+            pool=self,
+            connection=response_conn,
+            retries=retries,
+            **response_kw,
+        ) # Post of these parameters will need to change since we are not calling from inside of urllib3 ConnectionPool anymore
```
shadycuz, line 385: Most of these parameters from from ConnectionPool.urlopen:
https://github.com/urllib3/urllib3/blob/f9c7434090ce96ca93180137356531a9cfcce822/src/urllib3/connectionpool.py#L747-L753

When we move the wrapping to `HTTPConnection.getresponse()` we lose access to variables like `retries` and `pool`. 

It seems like `retries` might be useful as an instance variable? but I'm not sure about `pool`? Maybe it could be an optional argument like `get_response(pool=None)`.

I have not read the code for `from_httplib` but that is my next stop.

src/urllib3/connection.py
```diff
@@ -29,6 +29,7 @@
 
 from .util.timeout import _DEFAULT_TIMEOUT, _TYPE_TIMEOUT, Timeout
 from .util.util import to_str
+from .response import HTTPResponse
```
shadycuz, line 32: This will need to be fixed as well, it caused a circular import. 

```
ImportError: cannot import name '_TYPE_BODY' from partially initialized module 'flit_core.dummy.import1.connection' (most likely due to a circular import) (/home/runner/work/urllib3/urllib3/src/urllib3/connection.py)
```

shadycuz: This has been harder to fix than expected, lots of imports circular imports between connections.py and respone.py

sethmlarson: You'll need to use relative imports for urllib3, so it'd be `from .connection import X`. Reason we do this is some projects vendor us so we need to work in those circumstances too.

shadycuz: okay, but the only way forward I can see right now, is to combine `connection.py` and `response.py`. Since they import so many things from each other.

shadycuz: I found a way but its messy, lets come back to this.

src/urllib3/connectionpool.py
```diff
@@ -475,12 +479,12 @@ def _make_request(
             url,
             # HTTP version
             conn._http_vsn_str,  # type: ignore[attr-defined]
-            httplib_response.status,
-            httplib_response.length,
+            response.status,
+            response.length,
```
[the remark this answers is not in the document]
sethmlarson: Yeah having a `length` property sounds like a good idea.

src/urllib3/connection.py
```diff
@@ -369,6 +372,19 @@ def request_chunked(
         """
         self.request(method, url, body=body, headers=headers, chunked=True)
 
+    def getresponse(self, **response_kw: Any):
+
+        # Get the response from http.client.HTTPConnection
+        httplib_response = super().getresponse()
+        from .response import HTTPResponse
```
shadycuz, line 380: This was the only way I could get this to work without circular import

I'm going to put it at the top of the funciton with a comment to explain why its here.

src/urllib3/response.py
```diff
@@ -452,6 +453,7 @@ def __init__(
 
         self._pool = pool
         self._connection = connection
+        self.length = length
```
shadycuz, line 456: I feel like this length should interact someway with the init_length thing, but I'm not sure how.

shadycuz:
Tests are mostly passing, just a few more I need to fix up. 

```console
===== 6 failed, 1654 passed, 540 skipped, 52 warnings in 132.60s (0:02:12) =====
```

test/tz_stub.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 
 try:
-    import zoneinfo  # type: ignore[import]
+    import zoneinfo
```
shadycuz, line 10: Need to add this back, it complained on my machine that its an unused ignore, but fails in GitHub actions now that its removed.

shadycuz:
I have tried for several hours but unfortunately I'm stuck mocking this test

```python
    @patch.object(_HTTPConnection, "getresponse")
    def test_custom_http_response_class(self, super_getresponse: Mock) -> None:
        class CustomHTTPResponse(HTTPResponse):
            pass

        class CustomConnectionPool(HTTPConnectionPool):
            ResponseCls = CustomHTTPResponse

        # httplib_response = httplib.HTTPResponse(MockSock)
        # httplib_response.fp = MockChunkedEncodingResponse([b"f", b"o", b"o"])  # type: ignore[assignment]
        # httplib_response.headers = httplib_response.msg = httplib.HTTPMessage()  # type: ignore[assignment]

        super_getresponse.return_value = "Testing"

        with CustomConnectionPool(host="localhost", maxsize=1, block=True) as pool:
            conn = Mock()
            response = pool._make_request(
                conn, "GET", "/", retries=False, chunked=True, preload_content=False
            )

            assert isinstance(response, CustomHTTPResponse)
```

I'm trying to patch the `getresponse` method on the HTTPConnection class from httplib, which is the parent class of the urllib4 HTTPConnection. I've tried several things and have only been able to mock `getresponse` on the child class, which is not going to work.

@sethmlarson

src/urllib3/response.py
```diff
@@ -742,16 +752,31 @@ def from_httplib(
         """
         headers = r.msg
 
+        try:
+            assert_header_parsing(headers)
+        except (HeaderParsingError, TypeError) as hpe:
+            log.warning(
+                "Failed to parse headers (url=%s): %s",
+                _absolute_url(**response_kw),
```
shadycuz, line 760: So `_absolute_url` needs variables that are mostly found in `HTTPConnectionPool`.  It seems like `pool` is always passed in `response_kw`? But is it really always passed? 

I'm not familiar with how the internals of urlib3 work. Will `HTTPConnection` 's always belong to a connection pool? 

It didn't break any tests, so I assumed it was safe to move this function here and count on `pool` always being in `response_kw` kwargs. But I was also wondering why `host`, `scheme` and `port` were not available directly from `HTTPConnection`. Also shouldn't `HTTPConnection` have an attribute/instance variable linking it to the `HTTPConnectionPool` that created it? Something like `conn.pool`?

Just some of my thoughts while working my way through an unfamiliar codebase.

shadycuz (commented):
@sethmlarson I have left some details on my implementation and why I implemented things the way I did. I also found room for possible improvements either now or in the future. It would be great if you or some other maintainers could give a code review now that I have a working POC. Once I get feedback I can implement any changes, write some more tests and have this PR ready =).

src/urllib3/response.py
sethmlarson: `length` is defined in http.client as "Number of bytes left in the response" so... looks like they're the same value? Let's stick with our existing property then and remove `length`.

src/urllib3/response.py
```diff
@@ -742,16 +752,31 @@ def from_httplib(
...
+                hpe,
+                exc_info=True,
+            )
+
         if not isinstance(headers, HTTPHeaderDict):
             headers = HTTPHeaderDict(headers.items())  # type: ignore[assignment]
```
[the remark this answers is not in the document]
sethmlarson: This check should probably be inside of `getresponse()` so it can be called on the http.client response.

src/urllib3/response.py
```diff
@@ -423,6 +427,7 @@ def __init__(
         request_method: Optional[str] = None,
         request_url: Optional[str] = None,
         auto_close: bool = True,
+        length: Optional[int] = None,
```
sethmlarson, line 430: This can be removed.

src/urllib3/connectionpool.py
```diff
@@ -391,6 +400,10 @@ def _make_request(
             :class:`urllib3.util.Timeout`, which gives you more fine-grained
             control over your timeouts.
         """
+
+        # Set the ConnectionPools response class on the connection.
+        conn.ResponseClass_overide = self.ResponseCls
```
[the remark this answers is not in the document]
sethmlarson: I don't think we'll need this.

src/urllib3/connectionpool.py
```diff
@@ -461,7 +474,9 @@ def _make_request(
 
         # Receive the response from the server
         try:
-            httplib_response = conn.getresponse()
+            if response_kw is None:
+                response_kw = {}
+            response = conn.getresponse(retries, **response_kw)
```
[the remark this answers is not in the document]
sethmlarson: Instead of using `response_kw` we should explicitly define which parameters are being forwarded to `getresponse()` so we can make them a part of the `BaseHTTPConnection` API in the future. You can start with:

- retries
- connection
- pool
- request_method
- request_url

If there are extras that fail we can add them too.

src/urllib3/connectionpool.py
```diff
@@ -475,24 +490,11 @@ def _make_request(
             url,
             # HTTP version
             conn._http_vsn_str,  # type: ignore[attr-defined]
-            httplib_response.status,
-            httplib_response.length,
+            response.status,
+            response.length,
```
sethmlarson, line 494: Change to `response.length_remaining`.

src/urllib3/response.py
```diff
@@ -731,7 +737,11 @@ def stream(
 
     @classmethod
     def from_httplib(
```
sethmlarson, line 739: We should be able to remove `from_httplib` in this PR.

src/urllib3/response.py
sethmlarson: You can reconstruct the URL using the connection's host, port, scheme, and the request target (`request_url` in this instance)

src/urllib3/response.py
```diff
@@ -945,3 +970,10 @@ def __iter__(self) -> Iterator[bytes]:
                 buffer.append(chunk)
         if buffer:
             yield b"".join(buffer)
+
+
+# Mypy is upset incompatible type "HTTPConnectionPool"; expected "Type[HTTPConnectionPool]"
+def _absolute_url(
+    request_url: str, pool: Type["HTTPConnectionPool"], **_: Dict[str, Any]
+) -> str:
+    return Url(scheme=pool.scheme, host=pool.host, port=pool.port, path=request_url).url
```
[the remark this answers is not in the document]
sethmlarson: Should be `HTTPConnectionPool`, not `Type[...]`. We're not passing a class instance into the `pool` variable.

src/urllib3/util/typing.py
```diff
@@ -0,0 +1,3 @@
+from typing import IO, Any, Iterable, Union
+
+_TYPE_BODY = Union[bytes, IO[Any], Iterable[bytes], str]
```
sethmlarson, line 3: Where are you seeing a circular reference that requires this change? If it's due to type hints you can do the `if typing.TYPE_CHECKING:` conditional import to break the chain.

sethmlarson (changes requested):
Thanks for opening this. Here's my first collection of comments:

src/urllib3/connectionpool.py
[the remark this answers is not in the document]
shadycuz: You don't think we need the ability to return a custom HTTResponse class?

src/urllib3/response.py
shadycuz: Really? It seems like a handy class method to have. I would have to take all the logic from `from_httplib` and stick it someone else like `getresponse`? Maybe thats desirable for future plans, but it seems like it would also hurt the ability to have a custom reponse class.

src/urllib3/connectionpool.py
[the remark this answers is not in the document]
shadycuz: removed

src/urllib3/connectionpool.py
```diff
@@ -375,8 +382,10 @@ def _make_request(
         url: str,
         timeout: _TYPE_TIMEOUT = _DEFAULT_TIMEOUT,
         chunked: bool = False,
+        response_kw: Optional[Dict[str, Any]] = None,
+        retries: Optional[Retry] = None,
         **httplib_request_kw: Any,
```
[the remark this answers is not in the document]
shadycuz: fixedish

src/urllib3/connectionpool.py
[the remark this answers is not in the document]
shadycuz: This is fixed

src/urllib3/response.py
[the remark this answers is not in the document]
shadycuz: Fixed

src/urllib3/response.py
shadycuz: I did some more weird workaround stuff 😊, but I will resolve this as its been moved to another file

src/urllib3/response.py
shadycuz: Done

shadycuz:
@sethmlarson I have made the changes you requested. I will leave a more detailed review tomorrow with my thoughts or perhaps the day after. I'm a bit 😴. All the tests and stuff are still passing 🚀.

src/urllib3/connectionpool.py
```diff
@@ -391,6 +390,19 @@ def _make_request(
...
+        if "request_url" in httplib_request_kw:
+            del httplib_request_kw["request_url"]
+
+        # The key above and all the following keys need removed because conn.request() will complain of
+        # unexpected keyword arugments
+        preload_content: bool = httplib_request_kw.pop("preload_content", True)
+        decode_content: bool = httplib_request_kw.pop("decode_content", True)
+        enforce_content_length: bool = httplib_request_kw.pop(
+            "enforce_content_length", True
+        )
```
shadycuz, line 404: These values all had to be removed from `httplib_request` because of unexpected kwarg errors later when they were passed to `conn.request()`. I also needed to pass them on to `getresponse()`.

src/urllib3/connectionpool.py
```diff
@@ -461,7 +473,16 @@ def _make_request(
...
+            response = conn.getresponse(
+                url,
+                method,
+                self,
+                retries,
+                preload_content,
+                decode_content,
+                response_conn,
+                enforce_content_length,
+            )
```
shadycuz, line 485: pretty large breaking change in the API. I think I want to move `response_conn` up before `self` and maybe do `conn=response_conn` so it's clear for the next person what is going on.

src/urllib3/poolmanager.py
```diff
@@ -471,7 +471,7 @@ def urlopen(  # type: ignore[override]
             kw["headers"] = new_headers
 
         try:
-            retries = retries.increment(method, url, response=response, _pool=conn)  # type: ignore[arg-type]
+            retries = retries.increment(method, url, response=response, _pool=conn)
```
shadycuz, line 474: I guess `increment()` took a `HTTPResponse` from `urllib3` and now it's finally getting one? All I know is `mypy` complained about it.

test/test_connectionpool.py
```diff
@@ -506,17 +507,43 @@ class _raise_once_make_request_function:
...
+                    version=httplib_response.version,
+                    reason=httplib_response.reason,
+                    original_response=httplib_response,
+                    retries=retries,
+                    request_method=method,
+                    request_url=url,
+                    preload_content=False,
+                    connection=conn,
+                    pool=self._pool,
+                )
```
shadycuz, line 546: I think I messed up here, I should be getting `response_conn` from `kwargs` and using that to construct my `HTTPResponse` object. Since that's how its currently done in `_make_request()` and `getresponse()`.

test/test_connectionpool.py
```diff
@@ -557,22 +586,3 @@ def test_read_timeout_0_does_not_raise_bad_status_line_error(self) -> None:
...
-                httplib_response = httplib.HTTPResponse(MockSock)  # type: ignore[arg-type]
-                httplib_response.fp = MockChunkedEncodingResponse([b"f", b"o", b"o"])  # type: ignore[assignment]
-                httplib_response.headers = httplib_response.msg = httplib.HTTPMessage()
-                return httplib_response
-
-        with CustomConnectionPool(host="localhost", maxsize=1, block=True) as pool:
-            response = pool.request(
-                "GET", "/", retries=False, chunked=True, preload_content=False
-            )
-            assert isinstance(response, CustomHTTPResponse)
```
shadycuz, line 578: The most recent changes that were asked for removed this functionality. So I removed this failing test.

shadycuz (commented):
@sethmlarson Here are my thoughts on how I got to this point. I also leave my ideas for some other small changes. If you agree with some of my ideas then reply or give me a 👍.

src/urllib3/response.py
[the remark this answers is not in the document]
shadycuz: done

src/urllib3/response.py
shadycuz: done

src/urllib3/response.py
shadycuz: and here I was thinking "total" content-length... 😊 

done

src/urllib3/util/typing.py
```diff
@@ -0,0 +1,4 @@
+
+from typing import Union, IO, Any, Iterable
+
+_TYPE_BODY = Union[bytes, IO[Any], Iterable[bytes], str]
```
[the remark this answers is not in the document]
shadycuz: removed

src/urllib3/connection.py
```diff
@@ -369,6 +372,26 @@ def request_chunked(
...
+
+        # Wrap http.client.HTTPResponse as a urllib3.response.HTTPResponse
+        response = HTTPResponse.from_httplib(
+            httplib_response,
+            retries,
+            response_class=self.ResponseClass_overide,
+            **response_kw,
+        )
+
+        return response
```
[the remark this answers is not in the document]
shadycuz: Did you see this @sethmlarson or are you thinking it over 🙃

src/urllib3/connection.py
```diff
@@ -369,6 +377,58 @@ def request_chunked(
...
+        self,
+        request_url: str,
+        request_method: str,
+        pool: "HTTPConnectionPool",
+        retries: Optional["Retry"],
+        preload_content: bool,
+        decode_content: bool,
+        response_conn: Optional["HTTPConnection"],
+        enforce_content_length: bool,
+    ) -> "HTTPResponse":
```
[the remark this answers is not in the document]
sethmlarson: Interesting discovery on `conn` and `response_conn`, that seems like a bug or at least an oversight? Can we only pass one to `getresponse`?

src/urllib3/connection.py
```diff
@@ -369,6 +377,58 @@ def request_chunked(
...
+        except (HeaderParsingError, TypeError) as hpe:
+            log.warning(
+                "Failed to parse headers (url=%s): %s",
+                _absolute_url(self, pool.scheme, request_url),
+                hpe,
+                exc_info=True,
+            )
+
+        if not isinstance(headers, HTTPHeaderDict):
+            headers = HTTPHeaderDict(headers.items())  # type: ignore[assignment]
```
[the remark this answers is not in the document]
sethmlarson: Yeah let's unconditionally fold `.msg` into an `HTTPHeaderDict`.

src/urllib3/connection.py
```diff
@@ -369,6 +377,58 @@ def request_chunked(
...
+            original_response=httplib_response,
+            retries=retries,
+            request_method=request_method,
+            request_url=request_url,
+            preload_content=preload_content,
+            decode_content=decode_content,
+            connection=response_conn,
+            pool=pool,
+            enforce_content_length=enforce_content_length,
+        )
```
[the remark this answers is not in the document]
sethmlarson: That seems like a good idea!

src/urllib3/connection.py
```diff
@@ -743,3 +803,9 @@ class DummyConnection:
 
 
 VerifiedHTTPSConnection = HTTPSConnection
+
+
+def _absolute_url(
+    conn: Union[HTTPConnection, "HTTPConnectionPool"], scheme: str, request_url: str
+) -> str:
+    return Url(scheme=scheme, host=conn.host, port=conn.port, path=request_url).url
```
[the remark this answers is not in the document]
sethmlarson: Yeah this function can have any names it wants, it's internal. So let's name `request_url` to be `target` or something like that.

src/urllib3/connectionpool.py
```diff
@@ -375,8 +372,10 @@ def _make_request(
         url: str,
         timeout: _TYPE_TIMEOUT = _DEFAULT_TIMEOUT,
         chunked: bool = False,
+        retries: Optional[Retry] = None,
+        response_conn: Optional[HTTPConnection] = None,
         **httplib_request_kw: Any,
-    ) -> _HttplibHTTPResponse:
+    ) -> HTTPResponse:
```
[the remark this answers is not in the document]
sethmlarson: Can we ditch `response_conn` somehow? Does anything depend on it that can't use `conn`?

src/urllib3/connectionpool.py
```diff
@@ -391,6 +390,19 @@ def _make_request(
             :class:`urllib3.util.Timeout`, which gives you more fine-grained
             control over your timeouts.
         """
+
+        # This value should already be assigned to the `url` parameter
+        if "request_url" in httplib_request_kw:
```
sethmlarson, line 395: We shouldn't need to modify `httplib_request_kw` at this point, can we make all parameters explicit and avoid a `**` at all in our method signature?

src/urllib3/connectionpool.py
```diff
@@ -461,7 +473,16 @@ def _make_request(
 
         # Receive the response from the server
         try:
-            httplib_response = conn.getresponse()
+            response = conn.getresponse(
```
sethmlarson, line 476: Let's use keyword arguments here.

src/urllib3/connection.py
```diff
@@ -369,6 +377,58 @@ def request_chunked(
         """
         self.request(method, url, body=body, headers=headers, chunked=True)
 
+    def getresponse(  # type: ignore[override]
+        self,
+        request_url: str,
```
sethmlarson, line 382: Let's make this method only accept keyword arguments (ie `f(*, a, b, c)`) so that position doesn't matter.

sethmlarson (changes requested):
This is looking really great! Sorry for the delay, summer is a busy time :) Here's a handful of more comments:

src/urllib3/connectionpool.py
sethmlarson: As above, we can enumerate the parameters explicitly and remove any need for `**`

src/urllib3/connection.py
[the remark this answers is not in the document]
shadycuz: @sethmlarson We are only passing one to `getresponse`. Do you mean to `_make_request`? or do you mean to not pass `response_con` to `getresponse`?

To be honest I'm not sure, here is the original code:
https://github.com/urllib3/urllib3/blob/43c372f6eb9f9f94848c7b7645ad19ebf6c047c5/src/urllib3/connectionpool.py#L726-L753

I cant figure out how to handle `reponse_conn` with this refactor. I will have to double-check locally, but I'm pretty sure when I didn't pass `response_conn` to the urllib3 `HTTPResponse` constructor, then stuff broke.

src/urllib3/connection.py
shadycuz: The `*` is new to me, just googled it, coolio

sethmlarson: Wrote a [whole article on it](https://sethmlarson.dev/blog/strict-python-function-parameters) if you're interested :P

src/urllib3/connection.py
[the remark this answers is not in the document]
shadycuz: Yeah I checked locally @sethmlarson and this breaks stuff, especially around releasing connections. 

```
FAILED test/with_dummyserver/test_connectionpool.py::TestConnectionPool::test_for_double_release - assert 5 == (5 - 1)
```

I feel like this would take a separate effort move the connection releasing logic inside the `HTTPConnection`.

src/urllib3/connection.py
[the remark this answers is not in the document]
shadycuz: done

src/urllib3/connection.py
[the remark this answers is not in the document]
shadycuz: @sethmlarson Done. 2

We are passing all of the parameters to `HTTPResponse` except for two. 

`msg: Optional[_HttplibHTTPMessage] = None` but when I looked, it doesn't appear the `HTTPResponse` is even using `msg`? It only sets it in `__init__`. It's not being passed to `super` and I dont see it being used in the base class? 

The other thing we are not passing is `# auto_close: bool = True`, which kinda explains why when I remove `response_conn` we have so many issues with tests breaking.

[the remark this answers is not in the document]
shadycuz: I tried setting `auto_close` to `False` and removing the `response_conn` but I still see the double release test breaking.

src/urllib3/connectionpool.py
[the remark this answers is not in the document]
shadycuz: It breaks stuff, mostly around connection releasing in the HTTPConnectionPool

src/urllib3/connection.py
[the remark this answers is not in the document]
shadycuz: I did stuff, I hope you like 🙈

src/urllib3/connectionpool.py
shadycuz: Done

test/test_connectionpool.py
shadycuz: done

src/urllib3/connectionpool.py
shadycuz: done

src/urllib3/connection.py
shadycuz: done

src/urllib3/connectionpool.py
shadycuz: done

src/urllib3/connectionpool.py
```diff
@@ -428,9 +432,9 @@ def _make_request(
         # urllib3.request. It also calls makefile (recv) on the socket.
         try:
             if chunked:
-                conn.request_chunked(method, url, **httplib_request_kw)
+                conn.request_chunked(method, url, body, headers)
```
sethmlarson, line 435: Let's use keyword arguments for `body` and `headers`

src/urllib3/connectionpool.py
```diff
@@ -391,6 +394,7 @@ def _make_request(
             :class:`urllib3.util.Timeout`, which gives you more fine-grained
             control over your timeouts.
         """
+
```
sethmlarson, line 397: nit: Remove this added newline

src/urllib3/connectionpool.py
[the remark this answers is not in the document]
sethmlarson: Okay, maybe we can work on that in a separate PR since this one is looking very close to done.

src/urllib3/connection.py
```diff
@@ -743,3 +801,13 @@ class DummyConnection:
 
 
 VerifiedHTTPSConnection = HTTPSConnection
+
+
+def url_from_connection(
```
sethmlarson, line 806: Since the `url_from_connection` and `url_from_pool` functions aren't meant to be a part of our public API we should make them private via an underscore prefix.

src/urllib3/connection.py
```diff
@@ -24,12 +24,20 @@
...
     from .util.ssl_ import _TYPE_PEER_CERT_RET_DICT
     from .util.ssltransport import SSLTransport
+    from .util.retry import Retry
 
+from ._collections import HTTPHeaderDict
+from .util.response import assert_header_parsing
 from .util.timeout import _DEFAULT_TIMEOUT, _TYPE_TIMEOUT, Timeout
 from .util.util import to_str
 
+# Needed to move this far below to avoid circular import issues
```
sethmlarson, line 38: Remove this comment block.

sethmlarson (commented):
This is looking really great, nearly there! Here's a few more comments for you:

src/urllib3/connection.py
shadycuz: It was a recommended best practice from StackOverflow, it's supposed to be a hint that the thing is imported. But I can remove it.

src/urllib3/connectionpool.py
shadycuz: I can remove, but I feel like the added white space makes the code more readable.

shadycuz:
I will need to tidy up any docstrings where I have modified the signature. I need to write a test or two for things I have changed. Like `getresponse()` returning a urllib3 HTTPResponse. I'm not actually sure I need to add many tests, since their are already tests that `urlopen()` already returns an HTTPResponse. 

I also need to do a changelog entry.

src/urllib3/connectionpool.py
shadycuz: done

src/urllib3/connection.py
shadycuz: done

src/urllib3/connectionpool.py
shadycuz: done

src/urllib3/connection.py
shadycuz: done

changelog/2648.feature.rst
```diff
@@ -0,0 +1 @@
+Modified ``urllib3.HTTPConnection.getresponse`` to return an instance of ``urllib3.HTTPResponse`` instead of ``http.client.HTTPResponse``.
```
shadycuz, line 1: If you dont like this... it's your own words 😝

src/urllib3/connection.py
```diff
@@ -369,6 +374,103 @@ def request_chunked(
...
+        *,
+        method: str,
+        url: str,
+        pool: "HTTPConnectionPool",
+        retries: Optional["Retry"],
+        response_conn: Optional["HTTPConnection"],
+        preload_content: bool,
+        decode_content: bool,
+        enforce_content_length: bool,
+    ) -> "HTTPResponse":
```
shadycuz, line 388: post of these are not optional 🤔 which seems wrong?

test/with_dummyserver/test_connection.py
```diff
@@ -0,0 +1,96 @@
...
+    response: HTTPResponse = conn.getresponse(
+        method=method,
+        url=path,
+        pool=pool,
+        retries=None,
+        decode_content=False,
+        enforce_content_length=False,
+        preload_content=False,
+        response_conn=None,
+    )
```
shadycuz, line 46: Does this look right to you? 

I call `conn.request()` and I pass in the `method` and `url`. Then I call `conn.getresponse()` but again I have to pass in `method` and `url`? 

Does the connection not really know what response I want? Is it making the request again???

It seems to me like this could be better. Like what if:
```python
response = conn.request(method, path)
``` 
or what if returned some kinda pending response?
```python
pending_r = conn.request(method, path)
response = pending_r.getresponse()
```
___
Why does `HTTPResponse` need to know about retries? Isn't that handled by the `ConnectionPool`? Specifically by `urlopen`?
___
Why do we need so many options for reading the response? It seems to me would replace all these with somekinda config object.

```python
r_config = HTTPResponseConfig() This would have same defaults we use today

conn.getresponse(r_config)
```

You could then customise `HTTPResponseConfig` in the constructor or the java way lol
```python
response = r_config.with_conn(conn).with_decode().with_enforce_content_length().getresponse()
```
___

NONE of the above is for this PR <3, just food for thought on potential future enhancements.

shadycuz (commented):
@sethmlarson This is the hardest I have ever worked for a 150 LOC change. 

Everything is passing except: `Downstream / integration` tests and I'm not sure what to do about it, if anything.

test/with_dummyserver/test_connection.py
```diff
@@ -0,0 +1,96 @@
...
+        preload_content=True,
+        response_conn=None,
+    )
+
+    response.close()
+
+    assert pool.pool.qsize() == 0  # type: ignore[union-attr]
+
+
+def test_releases_conn(pool: HTTPConnectionPool) -> None:
```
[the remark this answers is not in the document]
sethmlarson: Agree that this testing style is preferable. We'll get there one day for the rest of the test suite! :)

sethmlarson:
@shadycuz I admire your bravery, this issue is one of the big ones for v2.0 and is a tough one to tackle as your first PR to this repository. I've thought a good deal about something similar to what you suggested above in having a "handle", or having `HTTPConnection.request` return something that you call `.getresponse()` on to receive the response back.

I'm putting together a PR pointed at this PR that might get us more towards what we're both thinking AND making integration tests pass. Then we can work on that idea more in a separate PR!

sethmlarson:
PR is here: https://github.com/shadycuz/urllib3/pull/1

shadycuz:
We are missing a tiny tiny test for:

```
# Raise the same error as http.client.HTTPConnection
        if self._response_options is None:
            raise ResponseNotReady()
```

but I can push a test for this.

sethmlarson:
@shadycuz Sounds great!

shadycuz:
@sethmlarson Coverage is back to 100%

On a random note, if pypy is supposed to be 7x faster than python, why is it consistently the slowest job to finish?

sethmlarson:
@shadycuz  Great, thanks! Pypy is "slow" due to us running coverage during trst execution. Pypy has poor performance when using coverage.

sethmlarson:
@shadycuz Wanted to follow up and let you know that this PR is 100% complete and is only waiting on external changes to happen to Requests. Thanks again for your patience!

shadycuz:
@sethmlarson do they want me to make the changes? I probably have the most context.

I'm going to work on the exception part that is mentioned in the issue.

sethmlarson:
@shadycuz That'd be great, going to tag @nateprewitt as he'd be the reviewer in that case.

shadycuz:
Coolio, I'll dig into it on the weekend.

shadycuz:
@sethmlarson @nateprewitt 

Note: TLDR is at the bottom.

So this morning I started digging into making changes in the `requests` package  to make it compatible with the upcoming breaking changes in `urllib3`. 

The only current issue this PR introduces is around `chunked` payloads because `requests` is not using the `chunked` capability of `urllib3` it insteads uses the low level API calls to do it manually. Previous discussion with more details is [here](https://github.com/shadycuz/urllib3/pull/1#issuecomment-1182666577).

This morning I started by reading the code and discovered that `requests` has originally used the `urllib3` ability to `chunk` payloads. It was introduced [here](https://github.com/psf/requests/commit/2da7fe08586aeaff614c017a9a834a22c937384a). It was then reverted in this [PR](https://github.com/psf/requests/pull/5353) because...

> The reason the first PR was reverted was a lack of review, and failing to meet the original criteria we'd set to make this change.

The change to using urlib3 for `chunking` was then re-submitted in this https://github.com/psf/requests/pull/5664 and it has been stuck there ever since. 

I took https://github.com/psf/requests/pull/5664 and brought it back up to date with the `main` branch of `requests` and then tested it against the latest changes in this PR. 

```
requests.exceptions.SSLError: HTTPSConnectionPool(host='localhost', port=42037): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for 'localhost'. (_ssl.c:1129)")))

requests/adapters.py:519: SSLError
============================================================================== short test summary info ===============================================================================
FAILED tests/test_requests.py::TestRequests::test_https_warnings - requests.exceptions.SSLError: HTTPSConnectionPool(host='localhost', port=42037): Max retries exceeded with url: ...
========================================================== 1 failed, 589 passed, 13 skipped, 1 xfailed in 79.23s (0:01:19) ===========================================================
vscode@bac018748aaf:/workspaces/requests$ git log
commit e7773451a23a1e020d14aa90cc58612cc9b06ce6 (HEAD -> native_chunked, upmerge)
Author: Leon Verrall <LVerrall@slb.com>
Date:   Tue Jul 2 10:17:33 2019 +0100

    Use urllib for chunked requests AGAIN

commit 352dbc3856d47fe2f4f64a07d44998e23e285e89
Author: James Pickering <james.pickering@xml-solutions.com>
Date:   Tue Jun 27 16:15:05 2017 +0100

    Fix #3844

commit 177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0 (origin/native_chunked, origin/main, origin/HEAD, main)
Author: David Cain <davidjosephcain@gmail.com>
Date:   Wed Jul 27 10:22:21 2022 -0700

    Remove Python 2 mention on `chardet` behavior (#6204)
```

All tests are passing except for the one because of the SSL error. I'm not sure if it's a local issue with my machine that wont show up in the `requests` CI/CD pipeline or maybe it's because of this commit:

```
commit 352dbc3856d47fe2f4f64a07d44998e23e285e89
Author: James Pickering <james.pickering@xml-solutions.com>
Date:   Tue Jun 27 16:15:05 2017 +0100

    Fix #3844
```
Which was some kinda fix for SSL and proxy errors? 

Here is the question I have for @sethmlarson and @nateprewitt... 

## TLDR

Since `urllib3` is about to have breaking API changes. Do we use that as an opportunity to introduce breaking API changes in `requests`? If yes, then I can fix up and resubmit https://github.com/psf/requests/pull/5664. 

Or... are we going to try and shield `requests` from breaking API changes? If so I could try and get the manual low-level chunking up to date with the changes we made in this PR. I don't know if it's even possible and if it is then it will continue to make that section of code undesirable.

Please get together to decide on a path forward. I think an `HFC` (high fidelity conversation) would really help here. Perhaps you two could have a 15 min video call?

shadycuz:
@sethmlarson @nateprewitt Just checking up on you. Did you have a chance to read the above ^ and meet + discuss it?

shadycuz:
@sethmlarson please provide an update

sethmlarson:
@shadycuz Sorry for not providing you an update on the last ping. The plan iirc between @nateprewitt and I is to move Requests to not use the low-level HTTPConnection APIs that are causing the problems we're seeing during integration testing. If you could create a separate PR that recreates https://github.com/psf/requests/pull/5664 that would likely be welcome, although I'm not sure what additional testing @nateprewitt has in mind.

shadycuz:
Thanks @sethmlarson

I will reopen an updated version of https://github.com/psf/requests/pull/5664

sethmlarson:
Since this PR is a dependency on some other work I'd like to complete this week I've pushed a patch to skip all the "chunked" tests in our Requests integration testing. These will be re-enabled once https://github.com/psf/requests/pull/6226 is merged.

sethmlarson (approved):

sethmlarson:
@shadycuz Sorry this languished for a while, you're welcome to submit to OpenCollective to collect the $300 bounty for closing https://github.com/urllib3/urllib3/issues/2648. Thanks again!

PR #2651, opened 2022-06-22
Use proxy_ssl_context as SSL ctx when forwarding

### LINKED ISSUE #2577: Proxies should use proxy_ssl_context when connecting via ProxyConfig.use_forwarding_with_https = True
This was discovered in https://github.com/urllib3/urllib3/pull/2558 and discussed with @jalopezsilva  on Discord. The gist is it appears that we're using `HTTPSConnection.ssl_context` to connect to HTTPS proxies when using `use_forwarding_with_https = True` mode. This is likely caused by us treating the proxy like it's the origin when we have proxies in "forwarding" mode and previously we didn't have a forwarding mode for HTTPS.

avi364:
Fixes #2577 by setting ssl_context to be the provided proxy_ssl_context when ```use_forwarding_for_https``` is set to True.

I enabled the failing test, but because the test doesn't really make sense (I think) for when a forwarding proxy is used, I removed that case from the test. However, I added additional tests to cover the different cases of valid ```proxy_ssl_context``` and ```ssl_context``` being either provided or not provided (with the existing ```test_https_proxy_tls_error``` test covering the case when a valid ```proxy_ssl_context``` is not provided).

The issue described that providing ```ssl_context``` should only be a DeprecationWarning for version 1.26.x. Should that change be done as a separate PR against that branch?

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -424,26 +424,16 @@ class HTTPSConnection(HTTPConnection):
 
         if proxy_config and proxy_config.use_forwarding_for_https:
             if proxy_config.ssl_context is not None and ssl_context is not None:
-                warnings.warn(
-                    "The 'ssl_context' parameter is not used to configure the connection to the proxy when the 'use_forwarding_for_https' parameter is set to True "
-                    "and should not be set. The 'proxy_ssl_context' parameter is used to configure the connection. "
-                    "In a later urllib3 v2.x release, setting ssl_context will result in an error",
-                    DeprecationWarning,
-                    stacklevel=2,
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True"
                 )
             elif proxy_config.ssl_context is None and ssl_context is not None:
-                warnings.warn(
-                    "The 'ssl_context' parameter should not be used to configure the connection to the proxy when the 'use_forwarding_for_https' parameter is set to True, "
-                    "Use the 'proxy_ssl_context' parameter to customize the connection. "
-                    "In a later urllib3 v2.x release, setting ssl_context will result in an error",
-                    DeprecationWarning,
-                    stacklevel=2,
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True. Use the 'proxy_ssl_context' parameter to customize the connection to the proxy"
                 )
-                self.proxy_config = ProxyConfig(
-                    ssl_context, proxy_config.use_forwarding_for_https
-                )
+            else:
 
-            self.ssl_context = proxy_config.ssl_context
+                self.ssl_context = proxy_config.ssl_context
         else:
             self.ssl_context = ssl_context
 
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -421,7 +421,32 @@ class HTTPSConnection(HTTPConnection):
         self.key_file = key_file
         self.cert_file = cert_file
         self.key_password = key_password
-        self.ssl_context = ssl_context
+
+        if proxy_config and proxy_config.use_forwarding_for_https:
+            if proxy_config.ssl_context is not None and ssl_context is not None:
+                warnings.warn(
+                    "The 'ssl_context' parameter is not used to configure the connection to the proxy when the 'use_forwarding_for_https' parameter is set to True "
+                    "and should not be set. The 'proxy_ssl_context' parameter is used to configure the connection. "
+                    "In a later urllib3 v2.x release, setting ssl_context will result in an error",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+            elif proxy_config.ssl_context is None and ssl_context is not None:
+                warnings.warn(
+                    "The 'ssl_context' parameter should not be used to configure the connection to the proxy when the 'use_forwarding_for_https' parameter is set to True, "
+                    "Use the 'proxy_ssl_context' parameter to customize the connection. "
+                    "In a later urllib3 v2.x release, setting ssl_context will result in an error",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+                self.proxy_config = ProxyConfig(
+                    ssl_context, proxy_config.use_forwarding_for_https
+                )
+
+            self.ssl_context = proxy_config.ssl_context
+        else:
+            self.ssl_context = ssl_context
+
         self.server_hostname = server_hostname
         self.ssl_version = None
         self.ssl_minimum_version = None
@@ -500,6 +525,11 @@ class HTTPSConnection(HTTPConnection):
                 SystemTimeWarning,
             )
 
+        if self._connecting_to_proxy and self.proxy_config:
+            ssl_context = self.proxy_config.ssl_context
+        else:
+            ssl_context = self.ssl_context
+
         sock_and_verified = _ssl_wrap_socket_and_match_hostname(
             sock=sock,
             cert_reqs=self.cert_reqs,
@@ -513,7 +543,7 @@ class HTTPSConnection(HTTPConnection):
             key_file=self.key_file,
             key_password=self.key_password,
             server_hostname=server_hostname,
-            ssl_context=self.ssl_context,
+            ssl_context=ssl_context,
             tls_in_tls=tls_in_tls,
             assert_hostname=self.assert_hostname,
             assert_fingerprint=self.assert_fingerprint,
--- a/test/with_dummyserver/test_proxy_poolmanager.py
+++ b/test/with_dummyserver/test_proxy_poolmanager.py
@@ -78,14 +78,25 @@ class TestHTTPProxyManager(HTTPDummyProxyTestCase):
     def test_https_proxy_with_proxy_ssl_context(self) -> None:
         proxy_ssl_context = create_urllib3_context()
         proxy_ssl_context.load_verify_locations(DEFAULT_CA)
+        ctx = ssl.create_default_context()
+        ctx.load_verify_locations(DEFAULT_CA)
         with proxy_from_url(
             self.https_proxy_url,
             proxy_ssl_context=proxy_ssl_context,
-            ca_certs=DEFAULT_CA,
+            ssl_context=ctx,
         ) as https:
             r = https.request("GET", f"{self.https_url}/")
             assert r.status == 200
 
+            r = https.request("GET", f"{self.http_url}/")
+            assert r.status == 200
+
+    def test_https_proxy_with_proxy_ssl_context_to_http_target(self) -> None:
+        proxy_ssl_context = create_urllib3_context()
+        proxy_ssl_context.load_verify_locations(DEFAULT_CA)
+        with proxy_from_url(
+            self.https_proxy_url, proxy_ssl_context=proxy_ssl_context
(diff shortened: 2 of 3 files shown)


matttpt:
(Chiming in here since I've also been thinking about this issue today. Please forgive me if I'm stepping on your toes at all!)

I think there's another case when `proxy_ssl_context` must be used that this PR should address: requesting an HTTP destination forwarded through an HTTPS proxy. It looks like `ssl_context` is currently being used in this case also. (The existing test suite does not catch it: in particular, [`test_https_proxy_with_proxy_ssl_context`](https://github.com/urllib3/urllib3/blob/main/test/with_dummyserver/test_proxy_poolmanager.py#L78) sets `ca_certs=DEFAULT_CA` when it creates the proxy, which hides the fact that `proxy_ssl_context` is ignored.) To see this, you can try adding a test like this to `test/with_dummyserver/test_proxy_poolmanager.py` (will fail):

```python
proxy_ssl_context.load_verify_locations(DEFAULT_CA)
with proxy_from_url(
    self.https_proxy_url,
) as proxy:
    proxy.request("GET", f"{self.http_url}/")
```

Hope this is useful.

avi364:
@matttpt I think I better understand what's going and I think the only change required is to the documentation rather than the behavior, but please let me know if you agree.

https://urllib3.readthedocs.io/en/stable/user-guide.html#certificate-verification describes how the ```cert_reqs``` and ```ca_certs``` parameters for ```PoolManager``` can be used to globally configure certificate validation. And while not documented, ```ca_cert_dir``` and ```ca_cert_data``` are also parameters that can be set to control what certificates are used for validation.

The key word is **globally**, meaning that it is intended that the above parameters will overwrite the parameters set in ```ssl_context```. The reason I think this is intended is because ```test_context_combines_with_ca_certs``` in ```test_https.py``` explicitly tests this case.

As ```ProxyManager``` is a subclass of ```PoolManager```, the same rules from above apply, with the addition that  ```cert_reqs```, ```ca_certs```, ```ca_cert_dir``` and ```ca_cert_data``` will overwrite the parameters set in ```proxy_ssl_context``` as well.

In your test case, ```proxy_ssl_context``` was set up so that the client to proxy connection could be successfully validated, but the validation of the client to target connection failed because certificates to validate the connection were not provided, either through a provided ```ssl_context``` or through ```ca_certs```.

However, somehow this test case is passing when using ```HTTPSConnectionPool``` directly, meaning that the passed ```ca_certs``` is not overwriting ```ssl_context```, so there still is something that I'm missing.

```
ctx = util.ssl_.create_urllib3_context()
ctx.load_verify_locations(DEFAULT_CA)
with HTTPSConnectionPool(
    self.host, self.port, cert_reqs=ssl.CERT_REQUIRED, ca_certs=self.bad_ca_path, ssl_context=ctx
) as https_pool:
    https_pool.request("GET", "/")
```

avi364:
@matttpt I was completely off-base with my understanding and my previous comments were completely irrelevant. You were right that proxy_ssl_context was being ignored for HTTP destinations. I have corrected that as well and add your test case.

pquentin:
Hey @avi364 and @matttpt thanks for the great work here! We're not forgetting you but answering your questions requires some digging and we've had lot of activity recently

avi364:
@matttpt thanks for confirming.

@pquentin no problem and sorry for the noise, I really confused myself for a while, so most of my questions are no longer necessary to answer.

My only remaining question is about emitting a DeprecationWarning in 1.26.X and a ValueError in 2.x. I assume main will eventually be the branch used to cut 2.x. To make it easier to cherry pick this change onto the 1.26 branch, I broke the fix into 2 commits, where the first commit emits the warning. Is this the right approach?

src/urllib3/connection.py
```diff
@@ -421,7 +421,22 @@ def __init__(
...
-        self.ssl_context = ssl_context
+
+        if proxy_config and proxy_config.use_forwarding_for_https:
+            if proxy_config.ssl_context is not None and ssl_context is not None:
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True"
+                )
+            elif proxy_config.ssl_context is None and ssl_context is not None:
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True. Use the 'proxy_ssl_context' parameter to customizet the connection to the proxy"
```
jalopezsilva, line 432: nit: You have a typo, cutomizet

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -571,19 +619,13 @@ def test_https_proxy_tls_error(
...
+        ["proxy_scheme"],
         [
-            ("http", False),
-            ("https", False),
-            ("https", True),
+            ("http",),
+            ("https",),
         ],
     )
-    def test_proxy_https_target_tls_error(
```
jalopezsilva, line 581: Can you expand on why you disabled this test?

> I enabled the failing test, but because the test doesn't really make sense (I think) for when a forwarding proxy is used, I removed that case from the test. 

Shouldn't we verify that setting ssl_context will raise in the ``https`` proxy scheme case and that the context is not expected on ``http`` proxy scheme case.

jalopezsilva (commented):
I'm very excited to see progress to see this issue! This is great work and I'm looking forward to merging it.

I have you a quick review so far, but I'll take another pass.

jalopezsilva:
@SethMichaelLarson, how do we want to handle the following:

> My only remaining question is about emitting a DeprecationWarning in 1.26.X and a ValueError in 2.x. I assume main will eventually be the branch used to cut 2.x. To make it easier to cherry pick this change onto the 1.26 branch, I broke the fix into 2 commits, where the first commit emits the warning. Is this the right approach?

Were we planning on doing another release with 1.26.X? It would be ideal to push the DeprecationWarning first.

test/with_dummyserver/test_proxy_poolmanager.py
avi364: My thought was that instead of adding a condition in the test itself for (https proxy, use forwarding), I instead created ```test_forwarding_for_https_error_with_ssl_ctx``` and ```test_forwarding_for_https_error_with_only_ssl_ctx``` to test for ssl_context being set, as the scenario is unique for these proxy settings. However, I can add it back and handle the expected exception in the test itself if that would be better.

For the http proxy case, currently passing a ```proxy_ssl_context``` does not result in any behavior change or warning. Should a warning be added, recommending that ssl_context be set instead?

avi364:
@jalopezsilva is there anything else you would like me to do?

test/with_dummyserver/test_proxy_poolmanager.py
jalopezsilva: Reviewing this more carefully, the changes you did are correct. I don't see much value in keeping the ``use_forwarding_for_https`` flag if it's being skipped. 

> For the http proxy case, currently passing a proxy_ssl_context does not result in any behavior change or warning. Should a warning be added, recommending that ssl_context be set instead?

That's expected IMO and raising a warning might be confusing. Happy to discuss this more.

jalopezsilva:
> @jalopezsilva is there anything else you would like me to do?

Not at the moment. I want to hear @sethmlarson's opinion on the following though. 

> My only remaining question is about emitting a DeprecationWarning in 1.26.X and a ValueError in 2.x. I assume main will eventually be the branch used to cut 2.x. To make it easier to cherry pick this change onto the 1.26 branch, I broke the fix into 2 commits, where the first commit emits the warning. Is this the right approach?

Should we release the warning with 1.26.X and deprecate it fully with 2.x?

sethmlarson:
Sorry for taking so long on this @avi364 and @jalopezsilva. Let's emit a `DeprecatingWarning` on 1.26.x and a `ValueError` on `main`. Thanks for your work here!

avi364:
Thanks! From my understanding of how Github pull requests work, the 2 commits can be merged as-is into main, and then a committer can manually cherry-pick commit 929fba8df2e to 1.26.x to obtained the desired end-state on both branches.

jalopezsilva (approved):
Looks good to me! We probably need a review from @sethmlarson as well :)

sethmlarson (commented):
@jalopezsilva can you confirm my thinking on these comments?

src/urllib3/connection.py
```diff
@@ -421,7 +421,22 @@ def __init__(
...
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True"
+                )
+            elif proxy_config.ssl_context is None and ssl_context is not None:
+                raise ValueError(
+                    "The 'ssl_context' parameter cannot be set when use_forwarding_for_https is True. Use the 'proxy_ssl_context' parameter to customize the connection to the proxy"
+                )
+            else:
+
+                self.ssl_context = proxy_config.ssl_context
```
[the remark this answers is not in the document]
jalopezsilva: I agree with @sethmlarson here. You're already modifying the selection logic within ``connect()``. Keep this simple?

src/urllib3/connection.py
```diff
@@ -500,6 +515,11 @@ def connect(self) -> None:
                 SystemTimeWarning,
             )
 
+        if self._connecting_to_proxy and self.proxy_config:
```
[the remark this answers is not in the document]
jalopezsilva: @sethmlarson this might not work.

``self._connecting_to_proxy`` will remain True when we're using the absolute URI forwarding mode after the block in L486. I suspect that's why @avi364 is relying on it. I don't like relying on it though, seems brittle.

Could we check ``not self._is_using_tunnel() and self.proxy_config`` instead? That will default to the absolute UI forwarding mode. It might be better to check if ``proxy_ssl_context`` is set as well.

jalopezsilva (changes requested):

src/urllib3/connection.py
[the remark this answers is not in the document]
sethmlarson: Makes complete sense to me! @avi364 could you implement these changes and then we can merge?

skydiver1980 (approved):

abebeos:
@avi364 do you plan on continuing this?

abebeos:
> I do not, feel free to take it

Ok, thank you.

@sethmlarson , as mentioned within https://github.com/urllib3/urllib3/issues/2577#issuecomment-1891555808, i' pick only the testcases (if you'd say they are ok), and start the implementation from scratch.

(Though i am a little worried about the quit low review frequency. Thought it was the end year holidays, but it continues very slow)

abebeos:
@sethmlarson , seems this can be closed, too (see https://github.com/urllib3/urllib3/pull/2651#issuecomment-1893669034)

abebeos:
@sethmlarson , I just notice your https://github.com/urllib3/urllib3/pull/2651#discussion_r959957099

> Makes complete sense to me! avi364 could you implement these changes and then we can merge?

Is this PR otherwise ok? And thus...

- [ ] implementing the changes discussed
- [ ] rebasing

...would make it mergable?

abebeos:
Follow-up PR:
- https://github.com/urllib3/urllib3/pull/3295

ecerulm:
> I do not, feel free to take it

I think , I there is some missing comment in the issue. 

@avi364, do you mean that you are not going to pursue this PR further?

PR #2652, opened 2022-06-22
Allow for requests to send repeat headers through HTTPHeaderDict

### LINKED ISSUE #2242: Support repeated request headers using HTTPHeaderDict

rtpg:
This is an attempt to resolve #2242, and builds off of #2572

 This adds a `combine` parameter to `HTTPHeaderDict.add`, which 
 lets you say whether multiple values should be combined (via `,`), or 
 whether we should instead duplicate the header.

 This doesn't allow for both to happen on the same header, and will
 only work if a "behaviour change bit" is set on the header.

 The relevant change is in how `items` works.
 ```
        >>> headers = HTTPHeaderDict(foo='bar', password='hunter')
        >>> headers.respect_duplicates_in_items = True
        >>> headers.add('Foo', 'baz', combine=False)
        >>> headers.add('PASSWORD', '2', combine=True)
        >>> list(headers.items())
        [('foo', 'bar'), ('foo', 'baz'), ('password', 'hunter, 2')]
 ```

 This tries to maintain backwards compatibility by adding an opt-in
 parameter which "fixes" the HTTPHeaderDict to offer iteration that
 matches what http.client expects.

 While working on this, it became a bit unclear whether, when passing
 headers into requests, we should treat the dictionaries as mutable or
 not. I tried to go for "assume immutable" and opting for correctness
 over performance when trying to manage this.

 If we are OK totally breaking backwards compatibility on this class
 then a cleaner way of handling this would be to change how the items
 method works to respect these options. That would definitely lead to
 cleaner code.

 Another option: leave this dictionary, add another implementation with
 the incompatibilities but in a public namespace, and tell people to
 use that one instead when having issues.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -141,6 +141,14 @@ class HTTPHeaderDict(MutableMapping):
     def __init__(self, headers=None, **kwargs):
         super(HTTPHeaderDict, self).__init__()
         self._container = OrderedDict()
+        # XXX this is tracking our header duplication choices
+        # present => duplicate header
+        # not preset => merge with commas
+        self._duplicate_headers = set()
+        # XXX backwards compatibility bit regarding items
+        # used for making iteration work according to what http.client
+        # expects
+        self.respect_duplicates_in_items = False
         if headers is not None:
             if isinstance(headers, HTTPHeaderDict):
                 self._copy_from(headers)
@@ -212,7 +220,7 @@ class HTTPHeaderDict(MutableMapping):
         except KeyError:
             pass
 
-    def add(self, key, val):
+    def add(self, key, val, combine=None):
         """Adds a (name, value) pair, doesn't overwrite the value if it already
         exists.
 
@@ -220,6 +228,22 @@ class HTTPHeaderDict(MutableMapping):
         >>> headers.add('Foo', 'baz')
         >>> headers['foo']
         'bar, baz'
+
+        If you have respect_duplicates_in_items set:
+
+        - if you add an element with combine=True, then iterating with headers.items()
+          will merge the elements into one comma-separated item
+
+        - if you add an element with combine=False, then iterating with headers.items()
+          will iterate over each value as its own item
+
+        >>> headers = HTTPHeaderDict(foo='bar', password='hunter')
+        >>> headers.respect_duplicates_in_items = True
+        >>> headers.add('Foo', 'baz', combine=False)
+        >>> headers.add('PASSWORD', '2', combine=True)
+        >>> list(headers.items())
+        [('foo', 'bar'), ('foo', 'baz'), ('password', 'hunter, 2')]
+
         """
         key_lower = key.lower()
         new_vals = [key, val]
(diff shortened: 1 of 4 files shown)


sethmlarson:
@rtpg Thanks for opening this.

So I took another look and I believe the "solution" to [the issue](https://github.com/urllib3/urllib3/issues/2242) is to support sending headers via `HTTPHeaderDict` using all HTTP methods and with and without a body. As we saw in https://github.com/urllib3/urllib3/pull/2572 using `GET` already sends the header multiple times when we use `HTTPHeaderDict.add()`.

So the solution here isn't to necessarily add more parameters or complexity to `HTTPHeaderDict` but instead to use the `headers` parameters in a way so they aren't collapsed to a `dict` no matter what conditional branches get taken.

To test for this we should parameterize the test you've added to include the HTTP method `['GET', 'POST']` and a body `[None, b'x' * 10, (b'x' for _ in range(10))]`.

rtpg:
@sethmlarson right, I'll try this again. I had pulled in the test in #2572, and had noticed the need to parametrize the methods and had added that, but will add the body thing as well.

Regarding `GET`: we do support sending multiple names through `HTTPHeaderDict` of course, but we also have the existing behavior that collapses multiple values into comma separation. I believe that maintaining backwards compatibility here requires us to support both behaviors.

That is to say, we want to both support:
```
C: a, b, c
```
_and_
```
C: a
C: b
C: c
```
style header sending.

Currently that behavior differs between `GET` and `POST`. Will post a little table regarding this behavior so that I can properly document what change I think will be required here.

sethmlarson:
@rtpg I think we can leave the behavior on 1.26.x as-is for now and focus on the `main` branch. I've [updated the issue](https://github.com/urllib3/urllib3/issues/2242) with a list of minimum requirements. I would create a new PR and use https://github.com/urllib3/urllib3/pull/2572 only as a base for test cases.

rtpg:
Gotcha! Will instead use the `main` branch and set things up to do the "right thing"

rtpg:
Given I ended up on a totally alternative approach and am targetting a different base, I opened up an alternate branch over at #2669

sethmlarson:
Closing in favor of https://github.com/urllib3/urllib3/pull/2669

PR #2655, opened 2022-06-25
Close #2654, fix type hint of geturl

### LINKED ISSUE #2654: Problem with urllib3.HTTPResponse.geturl() type hint
### Subject

Why the return type hint for `urllib3.response.BaseHTTPResponse.geturl()` is `Optional[Union[str, "Literal[False]"]]` but not `Optional[str]` ?

suzaku:
`geturl` simply returns `self.url`, and `self.url` is a property with type hint `def url(self) -> Optional[str]:`, I think we can simply write `Optional[str]` here.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -411,7 +411,7 @@ class BaseHTTPResponse(io.IOBase):
     def info(self) -> HTTPHeaderDict:
         return self.headers
 
-    def geturl(self) -> Optional[Union[str, "Literal[False]"]]:
+    def geturl(self) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


hramezani:
I will check it. Probably tomorrow

hramezani (approved):

hramezani:
Thanks @suzaku for finding and fixing this.

@sethmlarson You mentioned this on the [original PR](https://github.com/urllib3/urllib3/pull/2255#discussion_r647083295) and It seems that I forgot to fix it :man_facepalming:

PR #2657, opened 2022-06-25
Fix `OverflowError` when SSL is used on some Python versions

### LINKED ISSUE #2513: OverflowError: signed integer is greater than maximum

illia-v:
Fixes #2513.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -45,6 +45,7 @@ try:
 except (AttributeError, ImportError, ValueError):  # Defensive:
     zstd = None
 
+from . import util
 from ._collections import HTTPHeaderDict
 from .connection import _TYPE_BODY, BaseSSLError, HTTPConnection, HTTPException
 from .exceptions import (
@@ -669,9 +670,10 @@ class HTTPResponse(BaseHTTPResponse):
     def _fp_read(self, amt: Optional[int] = None) -> bytes:
         """
         Read a response with the thought that reading more than 2 GiB
-        (`int` max value in C) at a time via SSL on some CPython 3.8 and
-        3.9 versions leads to an overflow error that has to be prevented
-        if `amt` or `self.length_remaining` indicate that it may happen.
+        (`int` max value in C) at a time via SSL when pyOpenSSL is used
+        or in any case on some CPython 3.8 and 3.9 versions leads to an
+        overflow error that has to be prevented if `amt` or
+        `self.length_remaining` indicate that it may happen.
         """
         assert self._fp
         c_int_max = (2**31) - 1
@@ -679,7 +681,7 @@ class HTTPResponse(BaseHTTPResponse):
         if (
             expected_length
             and expected_length > c_int_max
-            and (3, 8) <= sys.version_info < (3, 9, 7)
+            and (util.IS_PYOPENSSL or (3, 8) <= sys.version_info < (3, 9, 7))
         ):
             buffer = io.BytesIO()
             while amt is None or amt != 0:
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -685,16 +685,12 @@ class HTTPResponse(BaseHTTPResponse):
         c_int_max = 2**31 - 1
         if (
             (
-                amt
-                and amt > c_int_max
-                and (util.IS_PYOPENSSL or sys.version_info < (3, 10))
+                (amt and amt > c_int_max)
+                or (self.length_remaining and self.length_remaining > c_int_max)
             )
-            or (
-                self.length_remaining
-                and self.length_remaining > c_int_max
-                and (util.IS_PYOPENSSL or (3, 8) <= sys.version_info < (3, 9, 7))
-            )
-        ) and not util.IS_SECURETRANSPORT:
+            and not util.IS_SECURETRANSPORT
+            and (util.IS_PYOPENSSL or sys.version_info < (3, 10))
+        ):
             buffer = io.BytesIO()
             # Besides `max_chunk_amt` being a maximum chunk size, it
             # affects memory overhead of reading a response by this
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -666,6 +666,37 @@ class HTTPResponse(BaseHTTPResponse):
             if self._original_response and self._original_response.isclosed():
                 self.release_conn()
 
+    def _fp_read(self, amt: Optional[int] = None) -> bytes:
+        """
+        Read a response with the thought that reading more than 2 GiB
+        (`int` max value in C) at a time via SSL on some CPython 3.8 and
+        3.9 versions leads to an overflow error that has to be prevented
+        if `amt` or `self.length_remaining` indicate that it may happen.
+        """
+        if self._fp is None:
+            return b""
+        c_int_max = (2**31) - 1
+        expected_length = amt or self.length_remaining
+        if (
+            expected_length
+            and expected_length > c_int_max
+            and (3, 8) <= sys.version_info < (3, 9, 7)
+        ):
+            buffer = io.BytesIO()
+            while amt is None or amt != 0:
+                if amt is not None:
+                    chunk_amt = min(amt, c_int_max)
+                    amt -= chunk_amt
+                else:
+                    chunk_amt = c_int_max
+                data = self._fp.read(chunk_amt)
+                if not data:
+                    break
+                buffer.write(data)
+            return buffer.getvalue()
+        else:
+            return self._fp.read(amt)
+
     def read(
         self,
         amt: Optional[int] = None,
@@ -702,43 +733,14 @@ class HTTPResponse(BaseHTTPResponse):
         flush_decoder = False
         fp_closed = getattr(self._fp, "closed", False)
 
-        # Reading more than 2 GiB (`int` max value in C) via SSL on some
-        # CPython 3.8 and 3.9 versions leads to an overflow error
-        # that has to be prevented if `amt` or `self.length_remaining`
-        # indicate that it may happen.
-        c_int_max = (2**31) - 1
-        if (
-            (amt and amt > c_int_max)
-            or (self.length_remaining and self.length_remaining > c_int_max)
-        ) and (3, 8) <= sys.version_info < (3, 9, 7):
-
-            def read(amt: Optional[int] = None) -> bytes:
-                if self._fp is None:
-                    return b""
-                buffer = io.BytesIO()
-                while amt is None or amt != 0:
-                    if amt is not None:
-                        chunk_amt = min(amt, c_int_max)
-                        amt -= chunk_amt
-                    else:
-                        chunk_amt = c_int_max
-                    data = self._fp.read(chunk_amt)
-                    if not data:
-                        break
-                    buffer.write(data)
-                return buffer.getvalue()
-
-        else:
-            read = self._fp.read
-
         with self._error_catcher():
             if amt is None:
                 # StringIO doesn't like amt=None
-                data = read() if not fp_closed else b""
+                data = self._fp_read() if not fp_closed else b""
                 flush_decoder = True
             else:
                 cache_content = False
-                data = read(amt) if not fp_closed else b""
+                data = self._fp_read(amt) if not fp_closed else b""
                 if (
                     amt != 0 and not data
                 ):  # Platform-specific: Buggy versions of Python.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -703,17 +703,14 @@ class HTTPResponse(BaseHTTPResponse):
         fp_closed = getattr(self._fp, "closed", False)
 
         # Reading more than 2 GiB (`int` max value in C) via SSL on some
-        # CPython 3.8, 3.9, and 3.10 versions leads to an overflow error
+        # CPython 3.8 and 3.9 versions leads to an overflow error
         # that has to be prevented if `amt` or `self.length_remaining`
         # indicate that it may happen.
         c_int_max = (2**31) - 1
         if (
             (amt and amt > c_int_max)
             or (self.length_remaining and self.length_remaining > c_int_max)
-        ) and (
-            (3, 8, 0, "a", 4) <= sys.version_info < (3, 9, 7)
-            or (3, 10, 0, "b", 1) <= sys.version_info < (3, 10, 0, "rc", 1)
-        ):
+        ) and (3, 8) <= sys.version_info < (3, 9, 7):
 
             def read(amt: Optional[int] = None) -> bytes:
                 if self._fp is None:
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -733,7 +733,7 @@ class HTTPResponse(BaseHTTPResponse):
 
         with self._error_catcher():
             if amt is None:
-                # cStringIO doesn't like amt=None
+                # StringIO doesn't like amt=None
                 data = read() if not fp_closed else b""
                 flush_decoder = True
             else:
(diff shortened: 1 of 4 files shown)


src/urllib3/response.py
```diff
@@ -701,14 +702,46 @@ def read(
...
+        # CPython 3.8, 3.9, and 3.10 versions leads to an overflow error
+        # that has to be prevented if `amt` or `self.length_remaining`
+        # indicate that it may happen.
+        c_int_max = (2**31) - 1
+        if (
+            (amt and amt > c_int_max)
+            or (self.length_remaining and self.length_remaining > c_int_max)
+        ) and (
+            (3, 8, 0, "a", 4) <= sys.version_info < (3, 9, 7)
+            or (3, 10, 0, "b", 1) <= sys.version_info < (3, 10, 0, "rc", 1)
```
pquentin, line 715: Let's not consider unreleased versions of Python please :)

src/urllib3/response.py
```diff
@@ -701,14 +702,46 @@ def read(
...
+                        break
+                    buffer.write(data)
+                return buffer.getvalue()
+
+        else:
+            read = self._fp.read
+
         with self._error_catcher():
             if amt is None:
                 # cStringIO doesn't like amt=None
```
pquentin, line 739: Can we please update this comment? `cStringIO` was removed in Python 3. (Can be done in a future PR)

src/urllib3/response.py
```diff
@@ -701,14 +702,46 @@ def read(
...
+            or (self.length_remaining and self.length_remaining > c_int_max)
+        ) and (
+            (3, 8, 0, "a", 4) <= sys.version_info < (3, 9, 7)
+            or (3, 10, 0, "b", 1) <= sys.version_info < (3, 10, 0, "rc", 1)
+        ):
+
+            def read(amt: Optional[int] = None) -> bytes:
+                if self._fp is None:
+                    return b""
+                buffer = io.BytesIO()
```
pquentin, line 721: Since we only use `buffer.write()` and `buffer.getvalue()`, why not use a simple `bytearray`?

src/urllib3/response.py
```diff
@@ -701,14 +702,46 @@ def read(
...
+        c_int_max = (2**31) - 1
+        if (
+            (amt and amt > c_int_max)
+            or (self.length_remaining and self.length_remaining > c_int_max)
+        ) and (
+            (3, 8, 0, "a", 4) <= sys.version_info < (3, 9, 7)
+            or (3, 10, 0, "b", 1) <= sys.version_info < (3, 10, 0, "rc", 1)
+        ):
+
+            def read(amt: Optional[int] = None) -> bytes:
```
pquentin, line 718: While nested functions are nice, they're slower than normal functions since the function object gets built for every invocation and can be confusing. (Generally speaking, pretending Python is a functional language never ends well. Tail recursion is the other good example.)

Can you please:
 * place this nested function at the same level as the existing `read()` function? Maybe call it `_read_aux` or `_read_chunks`
 * place the `if` *inside* this new function?

src/urllib3/response.py
```diff
@@ -701,14 +702,46 @@ def read(
...
+                    if amt is not None:
+                        chunk_amt = min(amt, c_int_max)
+                        amt -= chunk_amt
+                    else:
+                        chunk_amt = c_int_max
+                    data = self._fp.read(chunk_amt)
+                    if not data:
+                        break
+                    buffer.write(data)
+                return buffer.getvalue()
```
pquentin, line 732: OK so we're talking about gigabytes of memory here. For a 3GB read, what is the peak memory usage? Is there a way to ensure we never use more than 3GB of memory?

pquentin (changes requested):
Thanks! The approach looks good.

src/urllib3/response.py
jayaddison: Relocating a comment from https://github.com/illia-v/urllib3/commit/2d6cf6c6f2714e5ad6455c1e665af83e9af9293f#r76646438 (thank you @illia-v for redirecting here):

Would it be worth considering using [`util.ssl_._is_ge_openssl_v1_1_1`](https://github.com/urllib3/urllib3/blob/f9c7434090ce96ca93180137356531a9cfcce822/src/urllib3/util/ssl_.py#L28-L38) -- or something like it -- to check whether to enable this logic?

One of my thoughts was that that could allow people who compile `cpython` themselves (for example, with different SSL libraries/version) and who use other variants of Python to benefit from the fix.

pquentin: Sure, if this fix depends on the OpenSSL version, we should definitely check that instead. I haven't investigated enough to tell.

illia-v: @jayaddison I can reproduce the overflow on CPython 3.8 compiled against OpenSSL 1.0.2 and 1.1.1. Does it really depend on the OpenSSL version?

sethmlarson: I was also under the impression that this was an issue in Python, not with OpenSSL.

src/urllib3/response.py
sethmlarson: Let's remove now since we're touching this area and the comment is out of date.

src/urllib3/response.py
jayaddison: Hmm, yep, I think I misunderstood something.  Re-constructing the situation:

The underlying 2GB limitation derives from OpenSSL <1.1.1, which used signed 32-bit ints ([bpo-42853 note](https://bugs.python.org/issue42853#msg384566)) in the relevant library function (`SSL_read`).

However, Python (cpython?) from 3.10 onwards calls the [more-recent](https://github.com/openssl/openssl/commit/4ee7d3f94590d3766c28eed03b969f338ddf96dc) [`SSL_read_ex`](https://github.com/python/cpython/blob/3.10/Modules/_ssl.c#L2513) function in OpenSSL that _does_ provide >2GB support.

So: an old Python compiled with OpenSSL >=1.1.1 would still have the limitation/bug, and Python with OpenSSL <1.1.1 would fail to compile in the first place (because the `SSL_read_ex` function wouldn't exist to link to).

Please double-check my logic there - but yep, I think I was mistaken, sorry for the distraction.

sethmlarson: @jayaddison No problem, thanks for participating in the discovery. It's good to know the innards of the space we're working on! :)

src/urllib3/response.py
sethmlarson: It looks like `BytesIO` is consistently faster than an unbounded `bytearray`?

```
$ python -m timeit -s 'import io; x = io.BytesIO()' 'for _ in range(10000):
    x.write(b"x" * 10000)

x.getvalue()
'
5 loops, best of 5: 45.9 msec per loop

$ python -m timeit -s 'import io; x = bytearray()' 'for _ in range(10000):
    x += b"x" * 10000

bytes(x)
'
2 loops, best of 5: 121 msec per loop
```

If that's the case we should use `io.BytesIO`, I wouldn't have guessed that!

src/urllib3/response.py
sethmlarson: If the content is 3GB and the user calls `resp.read()` we have to load the whole 3GB into memory, there's no way around that unfortunately?

pquentin: Sure but it's easy to accidentally use say 6GB instead.

sethmlarson: Ah now I understand what you meant, wasn't thinking about overhead!

src/urllib3/response.py
illia-v: Done in 890474a

src/urllib3/response.py
illia-v: > Let's not consider unreleased versions of Python please :)

OK, I updated the condition in 49ed9a1

src/urllib3/response.py
illia-v: I updated "cStringIO" to "StringIO" in 8fff25d, I was not sure if it is good to remove the comment

src/urllib3/response.py
illia-v: BTW, if you know how to check whether a response is read via SSL, please let me know.

src/urllib3/response.py
pquentin: Is this a good reason to use smaller chunks?

illia-v: Good idea! The smaller the chunk, the smaller the overhead should be.

I would not make it too small though. Maybe 512 MiB?

Python used to read in 1 MiB-chunks. Obviously, there were complaints https://github.com/python/cpython/issues/80231.

sethmlarson:
Huh, so SecureTransport only supports 16K reads at a time? Is that something wrong with our implementation or is that a ST quirk?

illia-v:
`SSL_read` called by CPython and PyPy seems to return at most 16 KiB too. But when `http.client.HTTPResponse.read` is called with a bigger `amt`, they seem to combine the 16 KiB chunks so that the method returns the requested `amt` of bytes. `SecureTransport` seems to read only one 16 KiB chunk when `amt` is bigger than 16 KiB.

illia-v:
PyPy tests are timing out because downloads with PyPy are very slow and memory-inefficient. One `test_requesting_large_resources_via_ssl` runs for 4.5 minutes locally, there are three variants in each of `test.with_dummyserver.test_socketlevel.TestSSL` and `test.contrib.test_pyopenssl.TestSocketSSL`, so the test is too slow for CI.

BTW, PyPy is prone to the overflow error too, the test results in `OverflowError: integer 2147483648 does not fit 'int'` in the main branch of urllib3. The change in this branch fixes the error and the tests pass.

sethmlarson:
@illia-v Interesting! The SecureTransport issue seems like a separate issue and could impact the 1.26.x branch, could we try creating a separate PR which adds a test case that calls `resp.read(16K+1)` inside of `test/with_dummyserver/test_https.py` and asserts that 16K+1 bytes are returned for each TLS backend?

If we can get that issue fixed that would help us along with this PR as well.

src/urllib3/response.py
```diff
@@ -665,6 +667,58 @@ def _error_catcher(self) -> Generator[None, None, None]:
...
+
+        The known cases:
+          * 3.8 <= CPython < 3.9.7 because of a bug
+            https://github.com/urllib3/urllib3/issues/2513#issuecomment-1152559900.
+          * urllib3 injected with pyOpenSSL-backed SSL-support.
+          * CPython < 3.10 only when `amt` does not fit 32-bit int.
+        """
+        assert self._fp
+        c_int_max = 2**31 - 1
+        if (
```
sethmlarson, line 686: This statement is really complex and I don't like having the `version_info` patch version determining behavior, what are your thoughts on simplifying it to this (since only 3.7 and late 3.9 benefit from our escape hatch):

```python
        if (
            (
                (amt and amt > c_int_max)
                or (self.length_remaining and self.length_remaining > c_int_max)
            )
            and not util.IS_SECURETRANSPORT
            and (util.IS_PYOPENSSL or sys.version_info < (3, 10))
        ):
```

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1491,6 +1492,63 @@ def socket_handler(listener: socket.socket) -> None:
...
+    # https://github.com/urllib3/urllib3/pull/2674
+    @notSecureTransport()
+    @pytest.mark.skipif(
+        os.environ.get("CI") == "true" and sys.implementation.name == "pypy",
+        reason="too slow to run in CI",
+    )
+    @pytest.mark.parametrize(
+        "preload_content,read_amt", [(True, None), (False, None), (False, 2**31)]
+    )
+    def test_requesting_large_resources_via_ssl(
```
sethmlarson, line 1505: Can we try to manually run PyPy on a `sys.version_info` version less than 3.10 to see that our implementation isn't super slow like it is in the test suite? Would be good to have a comparison between using urllib3+PyPy on a response that's size 2**31-1 and 2**31

sethmlarson (commented):
Left two comments, otherwise I think this PR is ready to ship :tada: Once we merge we can backport to 1.26.x

test/with_dummyserver/test_socketlevel.py
illia-v: It took 5 minutes 27 seconds for PyPy 3.9 to run the three tests. When I changed `2**31` to `2**31 - 100` (this plus the headers length end ups being near the limit) in the test, it took 5 minutes and 16 seconds.

src/urllib3/response.py
illia-v: I do not mind 🙂. I did this in 9a19fc0.

test/with_dummyserver/test_socketlevel.py
sethmlarson: Great, thank you!

sethmlarson (approved):
This looks great, thanks @illia-v! Will merge once CI passes.

sethmlarson:
Awesome, thanks @illia-v! Once you backport this feature+test to the 1.26.x branch you can collect the bounty.

PR #2666, opened 2022-07-01
Create a workflow for publishing to PyPI

sethmlarson:
Security measures in place:
- Added the `.github/workflows` directory to `CODEOWNERS`
- Tags for the `*` pattern are protected to maintain/admin privileges
- The `publish` environment requires a review from @shazow, @pquentin, or I to execute
- The `publish` environment has a `PYPI_TOKEN` configured for only the urllib3 package
- Used commits instead of tags for GitHub actions we're using within the workflow

I've added @pquentin to our `CODEOWNERS` reviewers because there isn't any additional risk as you're already a release manager now with reviewer privileges of the `publish` environment and you can help with the reviewing load to `CODEOWNERS` protected files.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,9 @@
 # Restrict all files related to deploying to
 # require lead maintainer approval.
 
-.github/CODEOWNERS          @sethmlarson @shazow
-src/urllib3/_version.py     @sethmlarson @shazow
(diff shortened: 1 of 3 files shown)


shazow (approved):
I didn't test the workflow, but the rest looks good!

pquentin (approved):
Thanks for testing this! Looks good to me.

PR #2669, opened 2022-07-06
Add support for configuring header merging when using HTTPHeaderDict

### LINKED ISSUE #2242: Support repeated request headers using HTTPHeaderDict

rtpg:
This is a second attempt to resolve #2242, as an alternative to #2652.

 When passing in `combine=True` during calls to `add`, iteration over the `HTTPHeaderDict` will merge duplicate header values into one with a comma.

 This also resolves an issue where GET requests and POST requests using the same `HTTPHeaderDict` would exhibit different header merging behavior. Previously GET requests would repeat header values over multiple lines, while POST requests were instead merging them with a comma. Now both request methods match the same behavior.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -174,14 +174,17 @@ class HTTPHeaderDictItemView(Set[Tuple[str, str]]):
     >>> d['X-Header-Name']
     'Value1, Value2, Value3'
 
-    However, if we iterate over an HTTPHeaderDict's items, we want to get a
-    distinct item for every different value of a header:
+    However, if we iterate over an HTTPHeaderDict's items, we will optionally combine
+    these values based on whether combine=True was called when building up the dictionary
 
+    >>> d = HTTPHeaderDict({"A": "1", "B": "foo"})
+    >>> d.add("A", "2", combine=True)
+    >>> d.add("B", "bar")
     >>> list(d.items())
     [
-        ('X-Header-Name', 'Value1')
-        ('X-Header-Name', 'Value2')
-        ('X-Header-Name', 'Value3')
+        ('A', '1, 2'),
+        ('B', 'foo'),
+        ('B', 'bar'),
     ]
 
     This class conforms to the interface required by the MutableMapping ABC while
@@ -273,6 +276,13 @@ class HTTPHeaderDict(MutableMapping[str, str]):
             return key.lower() in self._container
         return False
 
+    def setdefault(self, key: str, default: str = "") -> str:
+        if key in self:
+            return self[key]
+        else:
+            self[key] = default
+            return default
+
     def __eq__(self, other: object) -> bool:
         maybe_constructable = ensure_can_construct_http_header_dict(other)
         if maybe_constructable is None:
@@ -301,14 +311,24 @@ class HTTPHeaderDict(MutableMapping[str, str]):
         except KeyError:
             pass
 
-    def add(self, key: str, val: str) -> None:
+    def add(self, key: str, val: str, *, combine: bool = False) -> None:
         """Adds a (name, value) pair, doesn't overwrite the value if it already
         exists.
+
+        If this is called with combine=True, instead of adding a new header value
+        as a distinct item during iteration, this will instead append the value to
+        any existing header value with a comma. If no existing header value exists
+        for the key, then the value will simply be added, ignoring the combine parameter.
 
         >>> headers = HTTPHeaderDict(foo='bar')
         >>> headers.add('Foo', 'baz')
         >>> headers['foo']
         'bar, baz'
+        >>> list(headers.items())
+        [('foo', 'bar'), ('foo', 'baz')]
+        >>> headers.add('foo', 'quz', combine=True)
+        >>> list(headers.items())
+        [('foo', 'bar, baz, quz')]
         """
         # avoid a bytes/str comparison by decoding before httplib
         if isinstance(key, bytes):
@@ -318,7 +338,13 @@ class HTTPHeaderDict(MutableMapping[str, str]):
         # Keep the common case aka no item present as fast as possible
         vals = self._container.setdefault(key_lower, new_vals)
         if new_vals is not vals:
-            vals.append(val)
+            # if there are values here, then there is at least the initial
+            # key/value pair
+            assert len(vals) >= 2
+            if combine:
+                vals[-1] = vals[-1] + ", " + val
+            else:
+                vals.append(val)
 
     def extend(self, *args: ValidHTTPHeaderSource, **kwargs: str) -> None:
         """Generic import function for any type of header-like object.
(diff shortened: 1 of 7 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1643,6 +1643,62 @@ def socket_handler(listener: socket.socket) -> None:
...
+        (when combine=False)
+        """
+        data_sent: List[bytes] = []
+
+        def socket_handler(listener: socket.socket) -> None:
+            sock = listener.accept()[0]
+            sock.settimeout(0)
+            buf = b""
+            start_time = time.time()
+            while time.time() - start_time < (LONG_TIMEOUT / 2):
```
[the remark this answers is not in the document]
rtpg: I was hitting stability issues so decided to add a helper function that does basic parsing of bodies to try and handle `Content-Length` cleanly

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1643,6 +1643,120 @@ def socket_handler(listener: socket.socket) -> None:
             ]
             assert expected_response_headers == actual_response_headers
 
+    def _simple_read_http_request(self, sock: socket.socket) -> Tuple[(bytes, bytes)]:
```
rtpg, line 1646: This is kind of messy, but honestly I find this a more satisfying solution than sprinkling timeouts into the socket handlers to try and hope that the body doesn't get in the way.

test/with_dummyserver/test_socketlevel.py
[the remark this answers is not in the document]
rtpg: Full disclosure: my "stability issues" turned out to partly be a race condition (sending replies before sending responses), but I still think that my request parser is a nice thing to have in general for this test file in particular.

src/urllib3/_collections.py
```diff
@@ -174,14 +174,17 @@ class HTTPHeaderDictItemView(Set[Tuple[str, str]]):
     >>> d['X-Header-Name']
     'Value1, Value2, Value3'
 
-    However, if we iterate over an HTTPHeaderDict's items, we want to get a
-    distinct item for every different value of a header:
+    However, if we iterate over an HTTPHeaderDict's items, we will sometimes combine
```
sethmlarson, line 177: Instead of "sometimes" which makes it seem like behavior is undefined let's use "optionally".

src/urllib3/_collections.py
```diff
@@ -301,14 +311,23 @@ def discard(self, key: str) -> None:
         except KeyError:
             pass
 
-    def add(self, key: str, val: str) -> None:
+    def add(self, key: str, val: str, *, combine: Optional[bool] = None) -> None:
```
sethmlarson, line 314: `combine` can have a default value of `False` since that matches existing behavior. Kudos for making `combine` keyword-only! :clap:

src/urllib3/_collections.py
```diff
@@ -413,6 +440,15 @@ def itermerged(self) -> Iterator[Tuple[str, str]]:
             val = self._container[key.lower()]
             yield val[0], ", ".join(val[1:])
 
+    def iter_raw(self) -> Iterator[Tuple[str, bool, List[str]]]:
```
sethmlarson, line 443: I don't think we need this method if we choose to simplify the implementation as given above.

rtpg:
Understood RE the suggested implementation change. I think that makes good sense, is practical, and cuts down on the amount of code/explanation needed. I'll do that tomorrow (JST)

rtpg:
OK I sent in the changes, there seems to have been a test timeout on the Ubuntu 20.04/3.7 branch but it looks transitory

sethmlarson (commented):
This is so close to complete! Two comments for you, and to ensure we don't regress on past `Content-Type` behavior can you add this test case under `test/with_dummyserver/test_poolmanager::TestPoolManager`? (There's a test called `test_headers_http_header_dict` that you can place it near)

```python
    def test_headers_http_multi_header_multipart(self) -> None:
        headers = HTTPHeaderDict()
        headers.add("Multi", "1")
        headers.add("Multi", "2")
        old_headers = headers.copy()

        with PoolManager(headers=headers) as http:
            r = http.request(
                "POST",
                f"{self.base_url}/multi_headers",
                fields={"k": "v"},
                multipart_boundary="b",
                encode_multipart=True,
            )
            returned_headers = r.json()["headers"]
            assert returned_headers[4:] == [
                ["Multi", "1"],
                ["Multi", "2"],
                ["Content-Type", "multipart/form-data; boundary=b"],
            ]
            # Assert that the previous headers weren't modified.
            assert headers == old_headers

            # Set a default value for the Content-Type
            headers["Content-Type"] = "multipart/form-data; boundary=b; field=value"
            r = http.request(
                "POST",
                f"{self.base_url}/multi_headers",
                fields={"k": "v"},
                multipart_boundary="b",
                encode_multipart=True,
            )
            returned_headers = r.json()["headers"]
            assert returned_headers[4:] == [
                ["Multi", "1"],
                ["Multi", "2"],
                # Uses the set value, not the one that would be generated.
                ["Content-Type", "multipart/form-data; boundary=b; field=value"],
            ]

```

src/urllib3/_request_methods.py
```diff
@@ -208,9 +209,8 @@ def request_encode_body(
                 )
 
             extra_kw["body"] = body
-            extra_kw["headers"] = {"Content-Type": content_type}
+            extra_kw["headers"].add("Content-Type", content_type)
```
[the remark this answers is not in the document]
rtpg: I added an implementation of `setdefault` to the `HTTPHeaderDict`, though in order to keep the stuff a bit straightforward I made the default to `''` instead of `None` (including a default value was needed to match the `MutableMapping` signature for `setdefault`)

rtpg:
OK sent in a fix, there is a test failure in some unrelated code inside this connection thing, should we maybe bump up the flakiness runs for some of these tests?

Honestly I'm more of a fan of digging into these tests and removing flakiness when possible but I don't have a good understanding of what is (for example) a reasonable timeout for some of these operations. Like is this CI hiccuping, or is there an actual race condition we are hitting causing this flakiness? Or is the time given simply too short?

PR #2670, opened 2022-07-06
Add sigstore on publish for signing releases using GitHub OIDC

sethmlarson:
- Adds sigstore signing step to the `publish` workflow. This was implemented and tested here: https://github.com/sethmlarson/secure-python-package-template/commit/3e634f05e7ef0e8c5cdcc1c876a34e976f58f773.
- Adds the step of downloading sigstore cert+signature files from the GitHub Action artifacts and attach to the GitHub release.
- Fixes a typo in the name of the `publish` workflow (was `deploy`)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - "*"
 
+permissions:
+  # Needed to access the workflow's OIDC identity.
+  id-token: "write"
+
 jobs:
   publish:
     name: "Publish to PyPI"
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks!

.github/workflows/publish.yml
```diff
@@ -20,13 +24,41 @@ jobs:
...
+              --signature "sigstore-artifacts/${dist_name}.sig" \
+              --cert-oidc-issuer https://token.actions.githubusercontent.com
+
+          done
+
+      - uses: "actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8"
+        with:
+          name: "sigstore-artifacts"
+          path: "sigstore-artifacts/*"
+          if-no-files-found: "error"
```
nateprewitt, line 60: Not an issue just a thought, do we want this to be a standalone script that's invoked by the workflow? That allows this to be dry-run tested locally, doesn't couple the process to GHA.

sethmlarson: Unfortunately the process is coupled to GHA because we rely on GitHub's OpenID Connector being available ambiently for sigstore to sign the distributables with. I do see value in lifting this bash script into a separate file though for easier editing.

.github/workflows/publish.yml
```diff
@@ -20,13 +24,41 @@ jobs:
           python-version: "3.x"
 
       - name: "Install dependencies"
-        run: python -m pip install build
+        run: python -m pip install build==0.8.0 sigstore==0.6.2
```
bluetech, line 27: Since this is about supply chain security, given PyPI lackluster guarantees, it makes sense to use `pip install --require-hashes` for these? (Using `build-requirements.txt` or such)

PR #2671, opened 2022-07-07
Reduce flakiness of dummyserver tests

rtpg:
(Includes #2669)

This replaces a handful of timeout-based dummyserver socket handlers with something that uses a simple HTTP-aware handler that looks for content lengths or chunked encoding.

Locally, at least, this caused otherwise flaky tests in this file to consistently pass.

I don't feel strongly about this being the right way forward for tests, but I had a hard time finding a way to reuse existing code to do this, and the flakiness locally was a bit frustrating.

### DIFF
--- a/changelog/2242.feature.rst
+++ b/changelog/2242.feature.rst
@@ -1,0 +1,6 @@
+Added support for configuring header merging behavior with HTTPHeaderDict
+
+When using a ``HTTPHeaderDict`` to provide headers for a request, by default duplicate
+header values will be repeated. But if ``combine=True`` is passed in a call to
+``HTTPHeaderDict.add``, then the added header value will be merged in with an existing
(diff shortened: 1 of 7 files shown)


sethmlarson:
Thanks for opening this. My initial thought seeing the function is that it seems very complex, so much so that it would require tests itself. Why is this function less flaky than our existing setups for tests?

I'm also worried that because this function assumes that the code under test is perfectly following the HTTP protocol that we'd miss out on the times where we're not doing so versus what is typically done when we read the socket until data stops coming in.

sethmlarson:
After taking a second look at this, I think we're going to decline this PR. I'm interested in encapsulating some of our super-commonly used constructions (like `while buf.endswith(b"\r\n\r\n"): ...`) I don't think we want a complete HTTP parser in our test suite for socket level tests.

Thanks for interest in our test suite!

PR #2674, opened 2022-07-07
Fix reading pieces of data larger than 16 KiB on macOS

illia-v:
While I was working on #2657, I found that urllib3 injected with the `SecureTransport` integration fails to read the full response when the content length is more than 16 KiB.

A new test uncovers the problem too, the test has three variants.
All the variants failed on the CPython 3.8 and 3.10 runs, and only one variant failed on 3.7 and 3.9 runs. This happened because CPython 3.7 and 3.9 read desired amount of data in a loop. 3.8 and 3.10 did not do that. https://github.com/python/cpython/commit/153365d864c411f6fb523efa752ccb3497d815ca illustrates the difference in their code.

When I debugged the code, I found that `Security.SSLRead` reads only 16384 bytes even though `nbytes` and `buffer` are larger.
https://github.com/urllib3/urllib3/blob/72446d421c3b4faeaba392a3e2775c41dd25b7eb/src/urllib3/contrib/securetransport.py#L577-L579
Also, I checked `_read_callback` that is called by `Security.SSLRead`. The method read as much bytes from a socket as it was instructed, so I do not think it should be blamed.
I noticed that the socket was in the non-blocking mode in the callback, but changing the mode to blocking in `WrappedSocket` did not fix the problem.
https://github.com/urllib3/urllib3/blob/72446d421c3b4faeaba392a3e2775c41dd25b7eb/src/urllib3/contrib/securetransport.py#L204-L210

---

Calling `Security.SSLRead` in a loop until enough bytes are read looks to be a working solution.

### DIFF
--- a/changelog/2674.bugfix.rst
+++ b/changelog/2674.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed reading pieces of data larger than 16 KiB on macOS.
+Fixed incomplete reads of data by SecureTransport on macOS.
--- a/changelog/2674.bugfix.rst
+++ b/changelog/2674.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed reading pieces of data larger than 16 KiB on macOS.
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -573,7 +573,7 @@ class WrappedSocket:
         processed_bytes = ctypes.c_size_t(0)
         processed_bytes_total = 0
 
-        while nbytes:
+        while nbytes > 0:
             buffer = (ctypes.c_char * nbytes).from_buffer(buffer, processed_bytes.value)
 
             with self._raise_on_error():
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -570,40 +570,49 @@ class WrappedSocket:
         if nbytes is None:
             nbytes = len(buffer)
 
-        buffer = (ctypes.c_char * nbytes).from_buffer(buffer)
         processed_bytes = ctypes.c_size_t(0)
+        processed_bytes_total = 0
 
-        with self._raise_on_error():
-            result = Security.SSLRead(
-                self.context, buffer, nbytes, ctypes.byref(processed_bytes)
-            )
+        while nbytes:
+            buffer = (ctypes.c_char * nbytes).from_buffer(buffer, processed_bytes.value)
 
-        # There are some result codes that we want to treat as "not always
-        # errors". Specifically, those are errSSLWouldBlock,
-        # errSSLClosedGraceful, and errSSLClosedNoNotify.
-        if result == SecurityConst.errSSLWouldBlock:
-            # If we didn't process any bytes, then this was just a time out.
-            # However, we can get errSSLWouldBlock in situations when we *did*
-            # read some data, and in those cases we should just read "short"
-            # and return.
-            if processed_bytes.value == 0:
-                # Timed out, no data read.
-                raise socket.timeout("recv timed out")
-        elif result in (
-            SecurityConst.errSSLClosedGraceful,
-            SecurityConst.errSSLClosedNoNotify,
-        ):
-            # The remote peer has closed this connection. We should do so as
-            # well. Note that we don't actually return here because in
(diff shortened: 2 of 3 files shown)


src/urllib3/contrib/securetransport.py
```diff
@@ -570,40 +570,49 @@ def recv_into(
...
+
+            # There are some result codes that we want to treat as "not always
+            # errors". Specifically, those are errSSLWouldBlock,
+            # errSSLClosedGraceful, and errSSLClosedNoNotify.
+            if result == SecurityConst.errSSLWouldBlock:
+                # If we didn't process any bytes, then this was just a time out.
+                # However, we can get errSSLWouldBlock in situations when we *did*
+                # read some data, and in those cases we should just read "short"
+                # and return.
+                if processed_bytes.value == 0:
```
sethmlarson, line 595: Should this be `if processed_bytes_total == 0`? If we issue two reads, one that returns data and one that doesn't return data w/ errSSLWouldBlock then we want to return the already read data instead of erroring with a timeout.

src/urllib3/contrib/securetransport.py
```diff
@@ -570,40 +570,49 @@ def recv_into(
...
 
-        buffer = (ctypes.c_char * nbytes).from_buffer(buffer)
         processed_bytes = ctypes.c_size_t(0)
+        processed_bytes_total = 0
 
-        with self._raise_on_error():
-            result = Security.SSLRead(
-                self.context, buffer, nbytes, ctypes.byref(processed_bytes)
-            )
+        while nbytes:
```
sethmlarson, line 576: To avoid overrunning into always-truthy negative numbers can we use `while processed_bytes_total < nbytes:`

sethmlarson (changes requested):
Thanks, Illia! Two thoughts then this is ready to merge! :rocket:

changelog/2674.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed reading pieces of data larger than 16 KiB on macOS.
```
sethmlarson, line 1: Can we be specific and say this issue affects SecureTransport on macOS.

illia-v: Yes, I updated the entry. Also, I removed the exact amount because I remembered that I saw SecureTransport reading 8 and 32 KiB at a time from real resources while I was debugging the problem.

src/urllib3/contrib/securetransport.py
illia-v: I updated the condition to avoid the problem, but it is `while nbytes < 0:` now because both `nbytes` and `processed_bytes_total` change every successful iteration and `processed_bytes_total < nbytes` may become true too early.

src/urllib3/contrib/securetransport.py
illia-v: I am not sure about this.

CPython 3.7 and 3.9 read data in a loop as I wrote in the description. If the second or subsequent iteration of the loop calls `urllib3.contrib.securetransport.WrappedSocket.recv_into`, gets to the check and finds it true, it will raise the timeout error.

My code moves the loop inside of `recv_into`. So I assume that erroring with a timeout  `if processed_bytes.value == 0:` is consistent with what CPython 3.7 and 3.9 do.

---

I am not sure if returning read data knowing that it is not complete is the right way because the user may not even know that the response is incomplete.

illia-v: Yes, there is no loop in the method, but `BufferedReader` is used when `SecureTransport` is not used: https://github.com/urllib3/urllib3/pull/2674#issuecomment-1183657472.

> You can see that the `socket.recv` implementation restricts the amount of data you can ask for in one read on Windows to 16K.

Isn't that 2**32 - 1 bytes?

sethmlarson:
@illia-v I've been thinking about this PR a bit, do we need to fix this issue given we're deprecating SecureTransport? Don't want to do work that will end up being removed eventually if things are working good enough currently. Let me know your thoughts.

illia-v:
@sethmlarson I think we can close it since there are no other bug reports 🙂

sethmlarson:
Sounds good, thank you for your investigative work on this issue. It was top-notch, hoping to learn even more about how our low-level buffering and I/O works :)

PR #2678, opened 2022-07-14
Pin and move all Python dependencies to 'requirements' directory

sethmlarson:
Part of OpenSSF scorecard is pinning all dependencies to a specific version, optimally with hashes. These lock files are kept up-to-date with Dependabot which understands pip-compile in order to recompile lock files.

Follow-up PR will be done to lock GitHub Actions versions to git hashes as well.


sethmlarson:
Unfortunately because we're testing multiple operating systems this will be tough to fulfill without better support on pip-tools for environment conditionals in lockfiles. Going to close for now.

PR #2688, opened 2022-07-23
Fix type annotation for `pyopenssl.WrappedSocket.getpeercert`

sobolevn:
Hi! This return annotation seems incorrect, because you have [`no_implicit_optional = True`](https://github.com/urllib3/urllib3/blob/3951d3cf8f8c8119c455ce50c171dc0ca50cb130/setup.cfg#L22) and `get_peer_certificate()` [can return `None`](https://github.com/pyca/pyopenssl/blob/21.0.0/src/OpenSSL/SSL.py#L2185-L2194).

Refs https://github.com/python/typeshed/pull/8374

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -402,7 +402,7 @@ class WrappedSocket:
         else:
             self._io_refs -= 1
 
-    def getpeercert(self, binary_form: bool = False) -> Dict[str, List[Any]]:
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! You're right. Here's the relevant pyOpenSSL code:

https://github.com/pyca/pyopenssl/blob/38f9b4e524ac6479d57021bba2270df84d85b672/src/OpenSSL/SSL.py#L2352-L2361

PR #2693, opened 2022-07-28
Pin all workflows to commit shas

sethmlarson:
Part of the OpenSSF scorecard for pinning dependencies is pinning all GitHub Actions to commit SHAs.

```
$ scorecard-linux-amd64 --local urllib3 --checks Pinned-Dependencies

Starting [Pinned-Dependencies]
Finished [Pinned-Dependencies]

RESULTS
-------
Aggregate score: 10.0 / 10

Check scores:
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
|  SCORE  |        NAME         |           REASON            |                                             DOCUMENTATION/REMEDIATION                                              |
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
| 10 / 10 | Pinned-Dependencies | all dependencies are pinned | https://github.com/ossf/scorecard/blob/e42af756609b2cde6d757fd45ea05ddf0016ff62/docs/checks.md#pinned-dependencies |
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
```

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -12,13 +12,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - name: "Checkout repository"
+        uses: "actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b"
         with:
           # `towncrier check` runs `git diff --name-only origin/main...`, which
           # needs a non-shallow clone.
           fetch-depth: 0
 
-      - name: Check changelog
+      - name: "Check changelog"
         if: "!contains(github.event.pull_request.labels.*.name, 'Skip Changelog')"
         run: |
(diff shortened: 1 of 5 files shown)


pquentin (commented):
Thanks, the pin themselves look good. Why did we have to remove the names though? The resulting logs are harder to read:

image

sethmlarson:
@pquentin Hmm! I thought that the actions would use some sort of default name that would keep them readable. I'll add back the names.

sethmlarson:
I quoted all the names and I liked having the conditional always on top so I moved the `if` section to the top of each job step. This makes me want there to be a GitHub Action auto-formatter like Black.

illia-v (approved):

pquentin (approved):
Thanks! LGTM.

Issue #2692, opened 2022-07-28
Add deprecation warnings for urllib3.contrib.securetransport


sethmlarson:
#### Minimum requirements

- [ ] Create a newsfragment about the deprecation
- [ ] Add deprecation warning section to the SecureTransport docs
- [ ] Emit a `DeprecationWarning` for urllib3's SecureTransport support when importing the `urllib3.contrib.securetransport` module. The deprecation should point to the new docs.

Issue #2681, opened 2022-07-15
Deprecate the SecureTransport TLS implementation


sethmlarson:
The SecureTransport implementation was added to support OpenSSL versions that didn't support TLS 1.2 and later on macOS. Python on macOS has been improved greatly over time that we're now in a place to force OpenSSL 1.1.1+ in Python and urllib3.

Now that there's little upside to using SecureTransport it's hard to justify the maintenance burden.

My proposal for now is:

- Create documentation for SecureTransport's deprecation in the docs. We'll be targetting a "future 2.x release" with the actual removal.
- Begin unconditionally emitting a `DeprecationWarning` when `urllib3.contrib.securetransport.inject_into_urllib3` is called and point to the documentation.
- Find projects using `urllib3.contrib.securetransport` and help them remove support.

In a future 2.x version:

- Make the `inject_into_urllib3` function raise an exception with a URL to the documentation.

In a version after that:

- Remove the `inject_into_urllib3` function and `urllib3.contrib.securetransport` contrib module.

umarfarouk98:
Hi @sethmlarson i've added the deprecation warning to the inject_into_urllib3() method in the securetransport module.
Just a quick questions:

1. Hope it's fine to reference the warning to the https://urllib3.readthedocs.io/en/stable/reference/contrib/securetransport.html section of the doc as the deprecation notice will also be in this section of the doc.

```
def inject_into_urllib3() -> None:
    
    warnings.warn(
        "'urllib3.contrib.securetransport.inject_into_urllib3()' is deprecated and will be removed "
        "in a future release of urllib3 2.x. Read more in the doc: "
        "https://urllib3.readthedocs.io/en/stable/reference/contrib/securetransport.html",
        category=DeprecationWarning,
        stacklevel=2,
    )
    .......
    .......
    .......
```

pquentin:
Yes, links to the docs are one honking great idea -- let's do more of those!

umarfarouk98:
OK nice @pquentin I'll try to work on the remaining task.

PR #2695, opened 2022-07-31
Added deprecation warning for urllib3.contrib.securetransport

IvanLauLinTiong:
This PR is to address issue #2692

### DIFF
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -93,10 +93,10 @@ from ._securetransport.low_level import (
 )
 
 warnings.warn(
-    "'securetransport' module is deprecated and will be removed "
-    "in a future release of urllib3 2.x. Read more in doc: "
-    "https://urllib3.readthedocs.io/en/stable/reference/contrib/securetransport.html",
-    DeprecationWarning,
+    "'urllib3.contrib.securetransport' module is deprecated and will be removed "
+    "in a future release of urllib3 2.x. Read more in this issue: "
+    "https://github.com/urllib3/urllib3/issues/2681",
+    category=DeprecationWarning,
     stacklevel=2,
 )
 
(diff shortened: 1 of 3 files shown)


src/urllib3/contrib/securetransport.py
```diff
@@ -91,6 +92,14 @@
     _temporary_keychain,
 )
 
+warnings.warn(
+    "'securetransport' module is deprecated and will be removed "
```
sethmlarson, line 96: Let's use the full name of the module: `urllib3.contrib.securetransport`.

src/urllib3/contrib/securetransport.py
```diff
@@ -91,6 +92,14 @@
...
+    "in a future release of urllib3 2.x. Read more in doc: "
+    "https://urllib3.readthedocs.io/en/stable/reference/contrib/securetransport.html",
+    DeprecationWarning,
```
sethmlarson, line 99: Let's use the `category=DeprecationWarning` keyword argument here.

src/urllib3/contrib/securetransport.py
```diff
@@ -91,6 +92,14 @@
     _temporary_keychain,
 )
 
+warnings.warn(
+    "'securetransport' module is deprecated and will be removed "
+    "in a future release of urllib3 2.x. Read more in doc: "
```
sethmlarson, line 97: "Read more in this issue: ..."

sethmlarson (changes requested):
Few comments:

- Let's point people to this issue: https://github.com/urllib3/urllib3/issues/2681 in both the documentation and the warning message. That way we can provide updates if necessary on the situation.

Also note I rejected your Open Collective expense as the work hadn't been completed yet, hold off on creating one until the work is complete. Thanks!

IvanLauLinTiong:
Hi @sethmlarson ,  I have addressed all your comments in latest commit.

sethmlarson (approved):
LGTM. Thanks for this!

PR #2698, opened 2022-08-01
Update lint dependencies

hramezani:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.34.0
+    rev: v2.37.3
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -17,7 +17,7 @@ repos:
       - id: isort
 
   - repo: https://gitlab.com/pycqa/flake8
-    rev: 4.0.1
+    rev: 3.9.2
```
sethmlarson, line 20: Isn't this a downgrade?

hramezani: I did with `pre-commit autoupdate`. it seems we were pointing to the wrong repo.

sethmlarson (approved):
LGTM, thanks!!

PR #2699, opened 2022-08-02
Add Scorecard GitHub Action

sethmlarson:
Followed the steps here: https://github.com/ossf/scorecard-action

Created a Personal Access Token for my account according to their specifications and added it under `SCORECARD_TOKEN` so the complete set of Scorecard checks can run on the project.

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -1,0 +1,33 @@
+name: "Scorecard"
+on:
+  branch_protection_rule:
+  schedule:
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM. Too bad that you need to use a personal access token.

sethmlarson:
@pquentin Yeah, maybe one day GitHub Actions will be able to set repository permissions of the `GITHUB_TOKEN`. The only additional permission is being able to read Webhooks.

Issue #2691, opened 2022-07-28
Add deprecation warnings for urllib3.contrib.pyopenssl


sethmlarson:
#### Minimum requirements

- [x] Create a newsfragment about the deprecation
- [x] Add deprecation warning section to the pyOpenSSL docs
- [x] Emit a `DeprecationWarning` for urllib3's pyOpenSSL support when importing the `urllib3.contrib.pyopenssl` module. The deprecation should point to the new docs.
- [x] Add the `urllib3-secure-extra` dependency to the `secure` extra
- [x] Emit a `DeprecationWarning` for `urllib3[secure]` when it's detected that the `urllib3-secure-extra` package is importable. The deprecation should point to the new docs.
- [x] Find all issues in the issue tracker about pyOpenSSL and put them in this issue to be closed.

Part of: https://github.com/urllib3/urllib3/issues/2680

IvanLauLinTiong:
Hi @sethmlarson,  quick question for task 5: 
``Emit a DeprecationWarning for urllib3[secure] when it's detected that the urllib3-secure-extra package is importable. The deprecation should point to the new docs.``

Perhaps I miss out something, but I don't see the `urllib3-secure-extra` package resides here. Thus how do I add the `warnings.warn()`  method on the file ?

sethmlarson:
@IvanLauLinTiong I uploaded the package `urllib3-secure-extra` to PyPI not long ago: https://pypi.org/project/urllib3-secure-extra

You'd add this package to our `[secure]` extra and then if it's importable (via `import urllib3_secure_extra`) then you know that the `[secure]` extra was used somewhere.

IvanLauLinTiong:
> @IvanLauLinTiong I uploaded the package `urllib3-secure-extra` to PyPI not long ago: https://pypi.org/project/urllib3-secure-extra
> 
> You'd add this package to our `[secure]` extra and then if it's importable (via `import urllib3_secure_extra`) then you know that the `[secure]` extra was used somewhere.

Hi @sethmlarson,  I managed to add `urllib3-secure-extra` to `secure` extra under `pyproject.toml`. However, I don't understand, how could we yank `urllib3-secure-extra` with the `DeprecationWarning` message if we  `import urllib3_secure_extra` ?  That's because from file `urllib3_secure_extra.py` it does not have `warnings.warn()` implemented or does it?:
[...]

PR #2701, opened 2022-08-06
Add deprecation warnings for pyopenssl

IvanLauLinTiong:
This PR is to address issue #2691

### DIFF
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -21,7 +21,7 @@ from .util.retry import Retry
 from .util.timeout import Timeout
 
 try:
-    import urllib3_secure_extra
+    import urllib3_secure_extra  # noqa: F401
 except ImportError:
     pass
 else:
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -20,7 +20,10 @@ from .util.request import make_headers
 from .util.retry import Retry
 from .util.timeout import Timeout
 
-# note to vendors and OS distributors delete this block
+# === NOTE TO REPACKAGERS AND VENDORS ===
+# Please delete this block, this logic is only
+# for urllib3 being distributed via PyPI.
+# See: https://github.com/urllib3/urllib3/issues/2680
 try:
     import urllib3_secure_extra  # type: ignore # noqa: F401
 except ModuleNotFoundError:
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -20,9 +20,10 @@ from .util.request import make_headers
 from .util.retry import Retry
 from .util.timeout import Timeout
 
+# note to vendors and OS distributors delete this block
 try:
     import urllib3_secure_extra  # type: ignore # noqa: F401
-except ImportError:
+except ModuleNotFoundError:
     pass
 else:
     warnings.warn(
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -21,7 +21,7 @@ from .util.retry import Retry
 from .util.timeout import Timeout
 
 try:
-    import urllib3_secure_extra  # noqa: F401
+    import urllib3_secure_extra  # type: ignore # noqa: F401
 except ImportError:
     pass
 else:
--- a/src/urllib3/__init__.py
+++ b/src/urllib3/__init__.py
@@ -20,6 +20,19 @@ from .util.request import make_headers
 from .util.retry import Retry
 from .util.timeout import Timeout
 
+try:
+    import urllib3_secure_extra
+except ImportError:
+    pass
+else:
+    warnings.warn(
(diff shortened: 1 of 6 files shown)


sethmlarson:
@IvanLauLinTiong Looks like the "lint" job is failing, can you add a `# noqa: F401` comment in front of the import and run `nox -rs format` locally?

IvanLauLinTiong:
> @IvanLauLinTiong Looks like the "lint" job is failing, can you add a `# noqa: F401` comment in front of the import and run `nox -rs format` locally?

Hi @sethmlarson,  It seems the `format' job fails to run when I running `nox -rs format` on my Window machine. I opened an issue here #2709 .

src/urllib3/__init__.py
```diff
@@ -20,6 +20,19 @@
 from .util.retry import Retry
 from .util.timeout import Timeout
 
+try:
+    import urllib3_secure_extra  # type: ignore # noqa: F401
+except ImportError:
```
graingert, line 25: I think in 3.7+?

```suggestion
except ModuleNotFoundError:
```

src/urllib3/__init__.py
```diff
@@ -20,6 +20,19 @@
 from .util.retry import Retry
 from .util.timeout import Timeout
 
+try:
```
graingert, line 23:
```suggestion
# note to vendors and OS distributors delete this block
try: 
```

src/urllib3/contrib/pyopenssl.py
```diff
@@ -53,13 +53,22 @@ class UnsupportedExtension(Exception):  # type: ignore[no-redef]
...
 import ssl
+import warnings
 from io import BytesIO
 from socket import socket as socket_cls
 from socket import timeout
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 from .. import util
 
+warnings.warn(
```
graingert, line 64: You might want this warning only when you construct PyOpenSSLContext or call `inject_into...`

src/urllib3/__init__.py
IvanLauLinTiong: According to Python [doc](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError), `ModuleNotFoundError` is a subclass of `ImportError`.  Ya I agree with you,  `ModuleNotFoundError` seems more intuitive and serves right purpose here though `except ImportError` still able to catch the exception.

graingert: The purpose of ModuleNotFoundError is to distinguish cases where something breaks in a significant way from the module simply not existing.

If an ImportError that's not a ModuleNotFoundError shows up here then users will want to know about it

src/urllib3/__init__.py
IvanLauLinTiong: @graingert  I don't understand your question. What does it mean `we don't want pip itself to hit this`? 
What this block of code doing here is to try importing the marker package `urllib3-secure-extra` (which indicate `secure` extra has installed), if import success, we raise the `DeprecationWarning` message.

graingert: so urllib3 is pasted wholesale into `pip._vendor` and we don't want installing a package in someone's virtual environment to start causing pip issuing warnings

Additionally some distributors like RedHat/fedora/canonical/debian will redistribute urllib3 and they might not want it issuing warnings to their users. We certainly don't want them to attempt to distribute the marker package.

Doubtless they'll make the right choice here even if we don't add a note here

sethmlarson: Excellent idea, thanks @graingert!

sethmlarson (dismissed):
LGTM, now let's backport this to 1.26.x branch. We'll have to use `ImportError` there to support Python versions earlier than 3.7.

graingert:
```
if sys.version_info < (3, 6):
    ModuleNotFoundError = ImportError
```

IvanLauLinTiong:
> [benjaminp/six#201](https://github.com/benjaminp/six/issues/201)
> 
> ```
> if sys.version_info < (3, 6):
>     ModuleNotFoundError = ImportError
> ```

Not bad! we can add this checking

sethmlarson (approved):

sethmlarson:
There was a coverage issue when we switched ModuleNotFoundError -> ImportError but I've fixed it on your branch. Going to merge once we get a green light :)

sethmlarson (approved):

sethmlarson:
Thanks @IvanLauLinTiong, can you backport this PR to 1.26.x and then you can claim your bounty :)

IvanLauLinTiong:
> Thanks @IvanLauLinTiong, can you backport this PR to 1.26.x and then you can claim your bounty :)

Hi @sethmlarson. I already created the  [PR](https://github.com/urllib3/urllib3/pull/2701) to backport the changes to 1.26.x.

PR #2702, opened 2022-08-06
Add proxy_assert_(hostname|fingerprint) to ProxyManager

### LINKED ISSUE #2409: Add support for assert_proxy_cert_fingerprint and assert_proxy_hostname
For v2, it would be nice if we expand the ways we can match an HTTPS proxy certificate. After #2407  is merged, we'll only have hostname verification but it would be ideal to be consistent in what we support for HTTPS destinations.

### Minimum requirements

:moneybag: **You can get paid to complete this issue! [Please read the docs for more information](https://urllib3.readthedocs.io/en/latest/contributing.html#getting-paid-for-your-contributions).**

* [ ] Add new parameter ``assert_proxy_cert_fingerprint``
* [ ] Add new parameter ``assert_proxy_hostname``
* [ ] Add test cases for both parameters with similar rigor to the parameters `assert_cert_fingerprint` and `assert_hostname` tests.

jschneier:
This is good for review.

Fixes #2409

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -159,6 +159,21 @@ def no_san_server_with_different_commmon_name(
 
 
 @pytest.fixture
+def san_proxy_with_server(
+    tmp_path_factory: pytest.TempPathFactory,
+) -> Generator[Tuple[ServerConfig, ServerConfig], None, None]:
+    tmpdir = tmp_path_factory.mktemp("certs")
+    ca = trustme.CA()
+    proxy_cert = ca.issue_cert("localhost")
+    server_cert = ca.issue_cert("localhost")
+
+    with run_server_and_proxy_in_thread(
+        "https", "localhost", tmpdir, ca, proxy_cert, server_cert
+    ) as cfg:
+        yield cfg
+
+
+@pytest.fixture
 def no_san_proxy_with_server(
     tmp_path_factory: pytest.TempPathFactory,
 ) -> Generator[Tuple[ServerConfig, ServerConfig], None, None]:
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -3,7 +3,6 @@ import platform
 import socket
 import ssl
 import sys
-import threading
 from pathlib import Path
 from typing import AbstractSet, Any, Dict, Generator, NamedTuple, Optional, Tuple
 
@@ -13,7 +12,7 @@ from tornado import ioloop, web
 
 from dummyserver.handlers import TestingApp
 from dummyserver.proxy import ProxyHandler
-from dummyserver.server import HAS_IPV6, run_tornado_app
+from dummyserver.server import HAS_IPV6, run_loop_in_thread, run_tornado_app
 from dummyserver.testcase import HTTPSDummyServerTestCase
 from urllib3.util import ssl_
 
@@ -65,8 +64,7 @@ def run_server_in_thread(
     io_loop = ioloop.IOLoop.current()
     app = web.Application([(r".*", TestingApp)])
     server, port = run_tornado_app(app, io_loop, server_certs, scheme, host)
-    server_thread = threading.Thread(target=io_loop.start)
-    server_thread.start()
+    server_thread = run_loop_in_thread(io_loop)
 
     yield ServerConfig("https", host, port, ca_cert_path)
 
@@ -101,8 +99,7 @@ def run_server_and_proxy_in_thread(
     )
     proxy_config = ServerConfig(proxy_scheme, proxy_host, proxy_port, ca_cert_path)
 
-    server_thread = threading.Thread(target=io_loop.start)
-    server_thread.start()
+    loop_thread = run_loop_in_thread(io_loop)
 
     yield (proxy_config, server_config)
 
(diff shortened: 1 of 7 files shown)


test/conftest.py
```diff
@@ -162,6 +158,21 @@ def no_san_server_with_different_commmon_name(
         yield cfg
 
 
+@pytest.fixture
+def san_proxy_with_server(
+    tmp_path_factory: pytest.TempPathFactory,
```
sethmlarson, line 163: Let's add `loopback_host` fixture here so we can get tests for IPv4 and IPv6 automatically as well. Check out the `san_server` fixture for an example.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -651,6 +654,79 @@ def test_basic_ipv6_proxy(self) -> None:
...
+            proxy_url,
+            ca_certs=proxy.ca_certs,
+            proxy_assert_fingerprint=proxy_fingerprint,
+        ) as https:
+            with pytest.raises(MaxRetryError) as e:
+                https.request("GET", destination_url)
+
+            assert "Fingerprints did not match" in str(e)
+
+    def test_https_proxy_assert_hostname(
```
sethmlarson, line 701: Once we add the `loopback_host` fixture to `san_proxy_with_server` we can change the `proxy_assert_hostname` to be `proxy.host`.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -651,6 +654,79 @@ def test_basic_ipv6_proxy(self) -> None:
...
+        destination_url = f"https://{server.host}:{server.port}"
+
+        proxy_hostname = "example.com"
+        with proxy_from_url(
+            proxy_url, ca_certs=proxy.ca_certs, proxy_assert_hostname=proxy_hostname
+        ) as https:
+            with pytest.raises(MaxRetryError) as e:
+                https.request("GET", destination_url)
+
+            msg = f"hostname \\'{proxy_hostname}\\' doesn\\'t match \\'{proxy.host}\\'"
```
sethmlarson, line 727: Are the double `\\` necessary here if the outer quotes are `"`?

sethmlarson (changes requested):
This is excellent, nice work! I have a few comments and I'm going to request a quick review from @jalopezsilva but otherwise I think this looks great.

test/with_dummyserver/test_proxy_poolmanager.py
jschneier: I did try that first, I think there must be some nested stringifying or similar; I will double check though.

jschneier: Confirmed they are needed.

jschneier:
@sethmlarson addressed your comments.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -651,6 +655,93 @@ def test_basic_ipv6_proxy(self) -> None:
...
+        try:
+            addr = ipaddress.ip_address(host)
+        except ValueError:
+            return host
+
+        if addr.version != 6:
+            return host
+
+        # Transform ipv6 like '::1' to 0:0:0:0:0:0:0:1 via '0000:0000:0000:0000:0000:0000:0000:0001'
+        return addr.exploded.replace("0000", "0").replace("000", "")
```
jalopezsilva, line 677: Doesn't ``addr.compressed`` achieve this? What am I missing?

jalopezsilva (approved):
Great work! I had one small comment.  Once addressed, we'll be all set.

test/with_dummyserver/test_proxy_poolmanager.py
jschneier: The loopback address that shows up in the certificate is `0:0:0:0:0:0:0:1` but Python only gives me the options via method to get `::1` or `0000:0000:0000:0000:0000:0000:0000:0001'` so this logic is in here to get the matching certificate one.

Not sure if `0:0:0:0:0:0:0:1` format has a name.

jschneier: This is a decent feature request for Python I think.

jschneier: Alternatively I could just write the assert as `hostname \\'{proxy_hostname}\\' doesn\\'t match` and ignore the certificate side.

It doesn't really matter any which way.

jschneier:
Hi @sethmlarson & @jalopezsilva. I've addressed both of your comments which all focus around how I need to format the assertion string to check that the hostname is actually present in the error. Unfortunately the code I wrote is required, I tried all the suggestions in addition to a fair number of other things.

Do you all want me to change the code to do a simpler assertion or are we good to merge as-is?

sethmlarson (approved):
I think we're all good, thanks for working on this and getting it past the finish line @jschneier! :clap:

sethmlarson:
You can claim the $300 bounty for https://github.com/urllib3/urllib3/issues/2409 on Open Collective now, let me know if you need help with this.

Issue #2168, opened 2021-03-08
Drop support for OpenSSL<1.1.1


pquentin:
### Context

The TLS situation in Python has considerably improved since the early years of urllib3, thanks to the hard work of persons like Christian Heimes and Cory Benfield. urllib3 took advantage of new features even when only a subset of users could use it, and still accepts OpenSSL versions that don't have SNI, for example.

Here's what OpenSSL currently supports:

 * Version 1.1.1 will be supported until 2023-09-11 (LTS).
 * Version 1.0.2 is no longer supported. Extended support for 1.0.2 to gain access to security fixes for that version is available.
 * Versions 1.1.0, 1.0.1, 1.0.0 and 0.9.8 are no longer supported.

[RHEL 6 supports 1.0.1e+](https://access.redhat.com/articles/1462223) and [RHEL 7 only supports 1.0.2k and beyond](https://access.redhat.com/articles/1462343).

We also know that Python 3.10+ will require OpenSSL 1.1.1+ thanks to [PEP 644](https://www.python.org/dev/peps/pep-0644/).

Given this the only operating systems that would be in a tough spot if we decide to drop support for OpenSSL <1.1.1 are OSes who:

- Still support OpenSSL <1.1.1 as their default OpenSSL
- Use Python >=3.7 but <3.10 for their default Python
- Are likely to upgrade their system package for urllib3 to v2.0

The combination of the above three is very unlikely. We've identified a few OSes we'd like to evaluate to make sure before we release v2.0:

- Amazon Linux 2
- Gentoo

### Minimum requirements

:moneybag: **You can get paid to complete this issue! [Please read the docs for more information](https://urllib3.readthedocs.io/en/latest/contributing.html#getting-paid-for-your-contributions).**

- [x] Evaluate the above OSes to see if they'd be impacted by dropping support for OpenSSL <1.1.1. Leave this in a comment in this issue.
- [ ] Raise an `ImportError` if not OpenSSL or `ssl.OPENSSL_VERSION < (1, 1, 1)` with a message about urllib3 v2.0 requiring OpenSSL 1.1.1+
- [ ] Remove work-arounds for conditional features around the `ssl` module that are due to OpenSSL <1.1.1 (`minimum_version`, `HAS_SNI`, `_is_openssl_gt_v1_1_1`, [more examples below](https://github.com/urllib3/urllib3/issues/2168#issuecomment-794915225))
- [ ] Add documentation for urllib3 requiring OpenSSL 1.1.1+
[...]

PR #2705, opened 2022-08-09
Drop support for OpenSSL <1.1.1

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/2168

From the original issue, decided to not to implement these:

- Didn't remove detection for SSLContext.minimum_version because it requires Python 3.8
- Didn't remove SSLKEYLOGFILE detection because it requires Python 3.8
- Didn't remove post_handshake_auth detection because it requires Python 3.8

### DIFF
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -42,19 +42,12 @@ class TestSSL:
     def test_is_ipaddress_false(self, addr: Union[bytes, str]) -> None:
         assert not ssl_.is_ipaddress(addr)
 
-    @pytest.mark.parametrize(
-        ["ciphers", "expected_ciphers"],
-        [
-            ("ECDH+AESGCM:ECDH+CHACHA20", "ECDH+AESGCM:ECDH+CHACHA20"),
-        ],
-    )
     def test_create_urllib3_context_set_ciphers(
         self,
         monkeypatch: pytest.MonkeyPatch,
-        ciphers: Optional[str],
-        expected_ciphers: str,
     ) -> None:
 
+        ciphers = "ECDH+AESGCM:ECDH+CHACHA20"
         context = mock.create_autospec(ssl_.SSLContext)
         context.set_ciphers = mock.Mock()
         context.options = 0
@@ -63,7 +56,7 @@ class TestSSL:
         assert ssl_.create_urllib3_context(ciphers=ciphers) is context
 
         assert context.set_ciphers.call_count == 1
-        assert context.set_ciphers.call_args == mock.call(expected_ciphers)
+        assert context.set_ciphers.call_args == mock.call(ciphers)
 
     def test_create_urllib3_no_context(self) -> None:
         with mock.patch("urllib3.util.ssl_.SSLContext", None):
@@ -134,8 +127,7 @@ class TestSSL:
         assert context.post_handshake_auth == expected_pha
 
     def test_create_urllib3_context_default_ciphers(
-        self,
-        monkeypatch: pytest.MonkeyPatch,
+        self, monkeypatch: pytest.MonkeyPatch
     ) -> None:
         context = mock.create_autospec(ssl_.SSLContext)
         context.set_ciphers = mock.Mock()
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -43,8 +43,7 @@ class TestSSL:
         assert not ssl_.is_ipaddress(addr)
 
     def test_create_urllib3_context_set_ciphers(
-        self,
-        monkeypatch: pytest.MonkeyPatch,
+        self, monkeypatch: pytest.MonkeyPatch
     ) -> None:
 
         ciphers = "ECDH+AESGCM:ECDH+CHACHA20"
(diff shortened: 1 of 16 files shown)


pquentin:
OK the CI passes now! I had to make changes for docs, requests, botocore and coverage. I believe the main unresolved question is now LibreSSL in the issue.

sethmlarson:
I wonder about boto and `DEFAULT_CIPHERS`. Instead of removing it should we carry this list around as vestigal with a `DeprecationWarning` on the module level access of it?

nateprewitt:
We could potentially internalize the cipher list in Botocore if you don't want to maintain this into 2.0. I think the reason they were imported currently was the assumption that urllib3 would do a better job of making sure they were up to date.

pquentin:
My thinking was that if botocore upgrades to urllib3 2.0 and thus enforces OpenSSL 1.1.1+ it would make sense to trust the default OpenSSL cipher list in botocore too. (urllib3 1.26.x currently uses the existing OpenSSL 1.1.1 cipher list on all OpenSSL versions.) This is of course up to you @nateprewitt. You don't really have to decide now, I think either way the value of the botocore downstream tests is the same?

sethmlarson:
I'm defending the usage of `fmt: off` here as this is what Black produced by default:

```python
    if not ssl.OPENSSL_VERSION.startswith("OpenSSL ") or ssl.OPENSSL_VERSION_INFO < (
        1,
        1,
        1,
    ):  # Defensive:
        raise ImportError(
            "urllib3 v2.0 only supports OpenSSL 1.1.1+, currently "
            f"the 'ssl' module is compiled with {ssl.OPENSSL_VERSION}. "
            "See: https://github.com/urllib3/urllib3/issues/2168"
        )
```

test/test_ssl.py
```diff
@@ -172,24 +133,18 @@ def test_create_urllib3_context_pha(
 
         assert context.post_handshake_auth == expected_pha
 
-    @pytest.mark.parametrize("use_default_sslcontext_ciphers", [True, False])
     def test_create_urllib3_context_default_ciphers(
-        self, monkeypatch: pytest.MonkeyPatch, use_default_sslcontext_ciphers: bool
+        self,
+        monkeypatch: pytest.MonkeyPatch,
```
pquentin, line 138: nit: removing the trailing comma will use one less line

pquentin (changes requested):
Thanks! We're nearly there. I left a bunch of nits. I also think we can remove two more things:

* According to https://docs.python.org/3/library/ssl.html#ssl.SSLContext.maximum_version `minimum_version` and `maximum_version` are Python 3.7+, so any detection can be removed? (Not sure where it's done)
* We can also remove `hasattr(context, "load_default_certs")` in two places as `load_default_certs` is available in Python 3.4+: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs

sethmlarson:
@pquentin Thanks! Addressed your review comments.

For the `load_default_certs` we need to continue checking for availability as both the pyOpenSSL and SecureTransport SSLContext implementations don't support the method.

We don't do any detection for `minimum_version` and `maximum_version`, original implementation wouldn't attempt to use the feature unless the value was specified.

test/test_ssl.py
```diff
@@ -42,67 +42,21 @@ def test_is_ipaddress_true(self, addr: Union[bytes, str]) -> None:
...
-    @pytest.mark.parametrize(
-        ["ciphers", "expected_ciphers"],
-        [
-            (None, ssl_.DEFAULT_CIPHERS),
-            ("ECDH+AESGCM:ECDH+CHACHA20", "ECDH+AESGCM:ECDH+CHACHA20"),
-        ],
-    )
     def test_create_urllib3_context_set_ciphers(
         self,
         monkeypatch: pytest.MonkeyPatch,
```
pquentin, line 47: nit: new trailing comma :)

pquentin (approved):
Thanks, this looks great to me!

> For the `load_default_certs` we need to continue checking for availability as both the pyOpenSSL and SecureTransport SSLContext implementations don't support the method.

Oh I see. Maybe we should mention that to know when to remove it?

pquentin (approved):
LGTM! Thanks for bearing with me 😅

martinwunderlich:
Just FYI, I have run into some issues related to this removal of DEFAULT_CIPHERS. I am trying to run a python lambda function on AWS with some dependencies that in turn depend on package request 2.30.0 which depens on urllib3 2.0.2. 
I guess I'll somehow have to downgrade urllib or update botocore. 
There is a related issue here in the requests project: https://github.com/psf/requests/issues/6438

```
{
    "errorMessage": "Unable to import module 'mycode': cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_' (/opt/python/urllib3/util/ssl_.py)",
    "errorType": "Runtime.ImportModuleError",
    "requestId": "id",
    "stackTrace": []
}
```

martinwunderlich:
@pquentin Thanks for the feedback! I've downgraded urllib to <2 (which resolved to 1.26.15) and now it works.

PR #2706, opened 2022-08-09
Allow reuse of SSL sessions

macro1:
As discussed in #590, #604, #1886, and #1898. This PR borrows heavily from @PleasantMachine9's pr #1886.

The value seems to be primarily when connecting to a service that aggressively closes connections and also supports the abbreviated SSL handshake that's possible by reusing the SSL session.

~~I still need to add some restrictions to when urllib3 will allow the functionality to be enabled per the discussion in #1886~~ and I need to copy over the documentation.

Testing with standard integration tests that demonstrate the session reuse seems unlikely, so currently this is only confirming the debug log is hit. But I did confirm functionality against an HTTPS server that supports session resuming.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,7 @@ max-line-length=99
 [tool:pytest]
 xfail_strict = true
 python_classes = Test *TestCase
+log_level = DEBUG
 
 [isort]
 profile=black
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -387,6 +387,7 @@ class HTTPSConnection(HTTPConnection):
     ssl_maximum_version: Optional[int] = None
     assert_fingerprint: Optional[str] = None
     tls_in_tls_required: bool = False
+    ssl_session: Optional["ssl.SSLSession"] = None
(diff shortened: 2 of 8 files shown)


sethmlarson:
My hangup on this feature is that it's only for TLS 1.2, we know that [most TLS handshakes are TLS 1.3 these days](https://www.f5.com/labs/articles/threat-intelligence/the-2021-tls-telemetry-report). Should we wait for TLS 1.3 0-RTT / session resumption to be implemented in Python's ssl module first before we start trying to implement this?

macro1:
> My hangup on this feature is that it's only for TLS 1.2, we know that [most TLS handshakes are TLS 1.3 these days](https://www.f5.com/labs/articles/threat-intelligence/the-2021-tls-telemetry-report). Should we wait for TLS 1.3 0-RTT / session resumption to be implemented in Python's ssl module first before we start trying to implement this?

No worries. If this isn't a contribution you want, I'll close it. From the other conversations it looked like this hadn't moved forward due to a lack of work, not a lack of interest.

PR #2712, opened 2022-08-14
Standardize HTTPResponse.read(X) behavior regardless of compression

### LINKED ISSUE #2128: Standardize HTTPResponse.read(X) behavior regardless of compression
Currently calls to `HTTPResponse.read()` with a specified amount can vary in size depending on HTTP compression. This makes using `HTTPResponse` with other `io` primitives like `io.BufferedReader()` difficult as they assume `.read(X)` won't return more bytes than `X`.

Related issue: https://github.com/urllib3/urllib3/issues/709

### Minimum requirements

- [ ] Calling `HTTPResponse.read(x)` should result in the same behavior regardless of whether the response is compressed or not.
- [ ] Test cases for `decode_content=True/False`, `MultiDecoder` (more than one `Content-Encoding`), all the different Content-Encodings and `identity`. Test when `.stream()`-ing from the `HTTPResponse` or in manual `.read()`.
- [ ] May need to buffer output across calls to `.read()` and `decompress` until there's at least `n` bytes available to be read.

franekmagiera:
Closes #2128

Using a `bytearray` as a buffer for decoded data. Tested manually with httpbin so far, need to add unit tests.


sethmlarson:
Wanted to drop in and say thank you for picking up this issue @franekmagiera! Happy to see your name again :) Would you like me to assign you to https://github.com/urllib3/urllib3/issues/2128 so you can "claim" the issue despite the draft state of this PR?

franekmagiera:
Thanks @sethmlarson! I self-assigned, will try to create a first "reviewable" version soon.

test/test_response.py
```diff
@@ -499,13 +511,16 @@ def test_gzipped_streaming(self) -> None:
...
-            fp, headers={"content-encoding": "gzip"}, preload_content=False
+            fp,
+            headers={"content-encoding": "gzip", "content-length": length},
+            preload_content=False,
         )
         stream = resp.stream(2)
 
-        assert next(stream) == b"f"
-        assert next(stream) == b"oo"
+        assert next(stream) == b"fo"
```
sethmlarson, line 522: 🤩 How succinctly the impact change is captured by this little diff!

franekmagiera:
I think this should be ready for a first look. I see that Windows pipelines are failing, will check it out.

changelog/2128.bugfix.rst
```diff
@@ -0,0 +1 @@
+Standarized the ``HTTPResponse.read(n)`` behavior regardeless of compression.
```
sethmlarson, line 1: nit: "regardless"

pquentin:
How big can the buffer get? Is it linear in n? It's not obvious to me.

franekmagiera:
How large the buffer gets depends on the message in the body. If the message is very compressible, the buffer can get quite large even though we're asking for a small amount of bytes. For example, running the following script:

```python
from urllib3.response import DeflateDecoder
import zlib

message = b"abcdefghij" + b"foo" * 20 + b"klmnop"

print(f"Original message length: {len(message)}")

compressed = zlib.compress(message)

print(f"Compressed message length: {len(compressed)}")

decoder = DeflateDecoder()

pos = 3
while pos <= len(compressed):
    data = decoder.decompress(compressed[pos-3:pos])
    print(f"pos = {pos}; data length = {len(data)}; data: {data}")
    pos += 3
```

gives a following result:

```
Original message length: 76
Compressed message length: 29
pos = 3; data length = 0; data: b''
pos = 6; data length = 3; data: b'abc'
pos = 9; data length = 3; data: b'def'
pos = 12; data length = 3; data: b'ghi'
pos = 15; data length = 3; data: b'jfo'
pos = 18; data length = 58; data: b'ofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoofoo'
pos = 21; data length = 3; data: b'klm'
pos = 24; data length = 3; data: b'nop'
pos = 27; data length = 0; data: b''
```

It's a pretty tough question, it seems to me it's impossible to give a more precise answer without making any assumptions about the payload message itself and even then it feels like it'd be quite tricky to come up with a reasonable model.

franekmagiera:
Got some time today to take a look at failing Windows pipelines - there were `MemoryError`s caused by introducing the buffer. I put a bandaid on that by setting `decode_content=False` in failing tests (this circumvents writing to the buffer), couldn't see a better solution at the moment, will spend some time later to think about this. Also coverage is failing, will write more tests to cover missing 2 lines.

franekmagiera:
Regarding my last comment - I didn't have much time to think about anything smarter than setting `decode_content=False` to fix the tests that were failing on Windows due to `MemoryError`. The errors were caused by copying large amounts of data to the buffer in the default case. I don't think it's a significant regression though, let me know if you disagree. 

Also added the tests to bring back 100% coverage.

I think it is ready for a first look again, let me know what you think about it.

BTW, I don't mean to hurry you up by this comment - I understand it's not always easy to find time and I don't mind waiting for the review. I just realized that judging by my previous comment it is not clear whether I intended to make any more changes or not.

src/urllib3/response.py
```diff
@@ -785,12 +766,104 @@ def read(
...
+        if (
+            amt is not None
+            and self._decoded_bytes_in_buffer() < amt
+            and self.length_remaining is not None
+            and self.length_remaining > 0
+        ):
+            while self.length_remaining > 0 and self._decoded_bytes_in_buffer() < amt:
+                data = self._raw_read(amt)
+                decoded_data = self._decode(data, decode_content, flush_decoder)
+                self._decoded_bytes.extend(decoded_data)
```
pquentin, line 858: https://docs.python.org/3/library/io.html#io.BufferedIOBase clearly says that return `b''` indicates EOF and that there is no more data to read. The point of this pull request is to returns bytes when there are bytes to read and allow using `io.BufferedIOBase` freely. So I don't understand why we only issue raw_read calls when Content-Length is set. How about this instead?

```suggestion
        if amt is not None:                                                                                                                                    
            while self._decoded_bytes_in_buffer() < amt and data:              
                data = self._raw_read(amt)                                                                                                                     
                decoded_data = self._decode(data, decode_content, flush_decoder)                                                                               
                self._decoded_bytes.extend(decoded_data)
```

pquentin (changes requested):
Thanks! The base functionality works, but I think the reliance on content-length makes the pull request more complicated than needed. What am I missing? Do we really need to use it?

> Regarding my last comment - I didn't have much time to think about anything smarter than setting decode_content=False to fix the tests that were failing on Windows due to MemoryError. The errors were caused by copying large amounts of data to the buffer in the default case. I don't think it's a significant regression though, let me know if you disagree.

I don't know, it seems significant. It means this pull request interacts badly with https://github.com/urllib3/urllib3/pull/2657, at least on Windows.

src/urllib3/response.py
```diff
@@ -392,7 +395,7 @@ def _flush_decoder(self) -> bytes:
 
     # Compatibility methods for `io` module
     def readable(self) -> bool:
-        return True
+        return False
```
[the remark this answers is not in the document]
franekmagiera: Right, I think it is unrelated - iirc this was just something I noticed that it should return `False` because the `read` methods are not implemented

src/urllib3/response.py
franekmagiera: Right, this makes sense, I think it was just me overcomplicating because I didn't see the simpler solution.

src/urllib3/response.py
```diff
@@ -507,8 +510,11 @@ def __init__(
         # Determine length of response
         self.length_remaining = self._init_length(request_method)
 
+        # A buffer for decoded bytes that are read in the `read` method.
+        self._decoded_bytes = bytearray()
+
         # If requested, preload the body.
-        if preload_content and not self._body:
+        if preload_content and not self._body and self._fp:
```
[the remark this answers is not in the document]
franekmagiera: I think it would make sense (as well as the suggestion from below to use `is_readable`)

src/urllib3/response.py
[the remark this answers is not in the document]
pquentin: Removed in [`42a7fa1` (#2712)](https://github.com/urllib3/urllib3/pull/2712/commits/42a7fa1772ca62003c8fb96ca278d915b07386fc) and opened https://github.com/urllib3/urllib3/issues/2765 to discuss this before the 2.0 release.

pquentin:
I'm happy with the code now. I just need to:
 * figure out any test failures (but what I've seen until now also failed on main) and
 * investigate the `MemoryError`s on Windows.

@franekmagiera Can you please take another look? Some of my commits have message to explain my reasoning.

franekmagiera:
Went through the changes - I think those are great improvements. Thanks Quentin!

changelog/2128.removal.rst
```diff
@@ -0,0 +1,7 @@
+Standardized :meth:`~urllib3.response.HTTPResponse.read` to respect the semantics of BufferedIOBase regardless of compression. Specifically, this method:
+
+ * only returns an empty bytes object to indicate EOF (that is, the response has been fully consumed),
+ * never returns more bytes than requested,
+ * can issue multiple system calls.
+
+If you want to directly read the bytes as sent over the wire without all that logic, use :meth:`~urllib3.response.HTTPResponse.raw_read` instead.
```
sethmlarson, line 7: What's the use-case for the `raw_read` function on `HTTPResponse`? My thinking was the previous behavior of returning data while not following `io.BufferedIOBase.read` semantics was a bug.

Thinking on extensibility to HTTP/2: there's a chance that data is available in a buffer somewhere w/o needing a system call due to stream multiplexing, our `read` (and `readX`) functions should be documented to use either 0-1 (for `read1`, etc) or 0-N system calls instead of "1 or more".

src/urllib3/response.py
```diff
@@ -785,12 +763,86 @@ def read(
...
+
+        if amt is not None and self._decoded_bytes_in_buffer() >= amt:
+            return self._popleft_decoded_bytes(amt)
+
+        data = self.raw_read(amt)
+
+        if not data and self._decoded_bytes_in_buffer() == 0:
+            return data
+
+        self._init_decoder()
```
sethmlarson, line 827: Would splitting all the logic for amt is None and amt is not None here into their own blocks simplify both implementations?

test/test_response.py
```diff
@@ -65,10 +65,21 @@ def test_getheader(self) -> None:
 
 class TestResponse:
     def test_cache_content(self) -> None:
-        r = HTTPResponse(b"foo")
+        r = HTTPResponse(b"foo", decode_content=False)
```
sethmlarson, line 68: Is this a behavior change?

sethmlarson (commented):
This is looking superb, nice work all. Was just thinking this would be an interesting place to add hypothesis based tests for different combinations of compression and reads and asserting tells/eof, etc.

Also a general ask for a few more comments since these functions are quite complex, otherwise besides the questions below LGTM!

changelog/2128.removal.rst
pquentin: Thanks. I've opened https://github.com/urllib3/urllib3/issues/2769 to discuss this separately from this pull request. `raw_read` is now private again: [`341dca9` (#2712)](https://github.com/urllib3/urllib3/pull/2712/commits/341dca97ea666fb035017307979d781f4ccdbd0a)

src/urllib3/response.py
```diff
@@ -785,12 +763,86 @@ def read(
...
+            'content-encoding' header.
+
+        :param cache_content:
+            If True, will save the returned data such that the same result is
+            returned despite of the state of the underlying file object. This
+            is useful if you want the ``.data`` property to continue working
+            after having ``.read()`` the file object. (Overridden if ``amt`` is
+            set.)
+        """
+        if amt is not None:
```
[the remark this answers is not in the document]
pquentin: This method is definitely complex. However, I don't understand, what are you suggesting? Should we switch the `if cache_content` conditions to `if cache_content and amt is None`?

src/urllib3/response.py
```diff
@@ -785,12 +763,86 @@ def read(
...
             if cache_content:
                 self._body = data
+            return data
+
+        if amt is not None and self._decoded_bytes_in_buffer() >= amt:
+            return self._popleft_decoded_bytes(amt)
+
+        data = self.raw_read(amt)
+
+        if not data and self._decoded_bytes_in_buffer() == 0:
```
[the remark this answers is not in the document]
pquentin: Great question. I think raw_read() can only return 0 bytes when we reach EOF OR when amt is set to 0. My understanding is that in both cases it's OK return `b''` as well. Should we add a comment?

src/urllib3/response.py
pquentin: Definitely: [`57c5778` (#2712)](https://github.com/urllib3/urllib3/pull/2712/commits/57c5778712209566b8743167520510b61a28f848)

test/test_response.py
pquentin: No, this is irrelevant because this tests the cache set at `__init__` time. Removed: [`86b6b53` (#2712)](https://github.com/urllib3/urllib3/pull/2712/commits/86b6b5378e132f86aed9cd242ea86e10acd04879)

src/urllib3/response.py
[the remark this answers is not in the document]
pquentin: How would you do that?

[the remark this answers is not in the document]
sethmlarson: Since `raw_read` doesn't modify the buffer, can we assume that `self._decoded_bytes_in_buffer()` will return the same value before and after the `raw_read()`? If so we can cache the value from before the `raw_read` call. Since this is getting into super-nit territory I'll approve the PR though :)

sethmlarson (approved):
:shipit:

src/urllib3/response.py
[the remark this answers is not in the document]
pquentin: No problem regarding nits, I still have the memory issue to investigate anyways.

But it's just a len() call and a function call so I'm not sure that there's a lot to gain with caching. Maybe I should just replace this function with `len(self._decoded_bytes)`?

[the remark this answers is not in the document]
sethmlarson: Yeah let's do that!

pquentin:
Regarding the MemoryError issue, Windows is only the messenger here. There's apparently a [known issue with Windows in GitHub Actions](https://github.com/actions/runner-images/issues/2642) where memory available is lower but you can increase the pagefile (swap) to get jobs to run. However, it's not because Linux and macOS don't complain that this is fine. On my Linux laptop, I ran test_requesting_large_resources_via_ssl with `decode_content` enabled and disabled under `mprof run`:

### decode_content=False

decode_content=False

### decode_content=True

decode_content=True

This patch triples the memory usage! I need to look into using a single buffer here, but it's not trivial.

pquentin:
So this highlighted something useless we were doing when amt was None (the default): put the bytes in a buffer to get them back immediately. (This was much easier to see thanks to Seth's suggestion was above to have a specific block for that case.)

```
diff --git a/src/urllib3/response.py b/src/urllib3/response.py
index 956153ad..1c9b74b3 100644
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -827,9 +827,7 @@ class HTTPResponse(BaseHTTPResponse):
         self._init_decoder()
 
         if amt is None:
-            decoded_data = self._decode(data, decode_content, flush_decoder=True)
-            self._decoded_bytes.extend(decoded_data)
-            data = self._popall_decoded_bytes()
+            data = self._decode(data, decode_content, flush_decoder=True)
 
             if cache_content:
                 self._body = data
```

amt_none

In the last two remaining spikes, amt is 2GiB.

pquentin:
Using memray I understand clearly now why we're seeing four copies in the data:

image

 * The data read from the wire (self._raw_read)
 * The copy added to self._decoded_bytes
 * And two more copies made in popleft:
   * self._decoded_bytes[:n] (a new bytearray)
   * bytes(self._decoded_bytes[:n]) (a bytes object)

franekmagiera:
Closing this one in favor of #2798 
Impressive work @pquentin! Sorry for wasting some time with this one.

pquentin:
You did not waste any time, it was great work! Thankfully Windows told us something was amiss memory wise. You can also submit the 500$ bounty on Open Collective btw.

franekmagiera:
Yup, just feel bad that my instinct was to ignore it :/ Anyway, thanks, learnt a lot with this one. I will skip on the bounty though, wouldn't feel right for me to take it.

pquentin:
Your work was great, you deserve the bounty, no worries.

PR #2716, opened 2022-08-19
Add SLSA generic generator to publish workflow

sethmlarson:
Switches from Sigstore to SLSA for provenance attestation using GitHub OIDC. Allows users to verify that wheel and source distribution were built from our repository, workflow, and tag.

SLSA is also the new standard for a 10/10 on Scorecard: https://github.com/ossf/scorecard/pull/2144

Once the 1.26.12 release is live I'll do an E2E test, if that passes I will write documentation on how consumers can verify the wheels to close 

Tested that this setup works here:
- https://github.com/sethmlarson/python-slsa-release-test/releases/tag/0.1.14
- https://github.com/sethmlarson/python-slsa-release-test/blob/0.1.14/.github/workflows/release.yml

(Ignore new failures on this repository, was testing more things)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/release.md
+++ b/.github/PULL_REQUEST_TEMPLATE/release.md
@@ -8,14 +8,12 @@
     git push <REMOTE> --tags
     ```
 * [ ]  Execute the `publish` GitHub workflow. This requires a review from a maintainer.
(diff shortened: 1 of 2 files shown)


pquentin (commented):
Thanks! LGTM.

pquentin (approved):

PR #2719, opened 2022-08-20
add retry_after_replace_zero param to util.retry

### LINKED ISSUE #2718: Handle '0' Value in Retry-After Header
### Context

What are you trying to do?
_Optionally specify an overriding duration in the instance that the Retry-After header is '0' in `utils.Retry`._

How do you expect to do it?
_Add an optional parameter called `retry_after_replace_zero` to the class and update the `Retry.parse_retry_after` method to handle if this parameter is set._

_I don't think so._

_No_

### Alternatives

_Not currently._

### Duplicate

_I don't think so._

### Contribution

Yes

moseswynn:
closes #2718

### DIFF
--- a/changelog/2718.feature.rst
+++ b/changelog/2718.feature.rst
@@ -1,0 +1,3 @@
+Added optional parameter ``retry_after_replace_zero`` to ``urllib3.util.retry`` 
+module to allow overriding the wait time in the event that the Retry-After
+header on the response is '0'.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -174,6 +174,11 @@ class Retry:
         Whether to respect Retry-After header on status codes defined as
         :attr:`Retry.RETRY_AFTER_STATUS_CODES` or not.
 
+    :param int retry_after_replace_zero:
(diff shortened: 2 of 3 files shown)


sethmlarson:
This sounds like an issue with the server, not something we should handle in the client? Have you tried notifying the server maintainers about this issue?

moseswynn:
You're right that it is a server issue. I'm not sure it's high on the the server maintainers' radar in the most recent instance I've encountered it, but I've seen it in the past as well when interacting with APIs that have rate limits. Depending on how the delay-seconds value is calculated, if the value is <.5 or <1, then 0 will be returned in some cases. Causing the retry request to be sent before the limit timer is up and therefore extending it.

This change wouldn't be modifying any default behavior, but rather give developers the option to easily override the value if they encounter this issue, but I understand if it doesn't belong here.

sethmlarson:
I think we will decline this PR because it sounds like a specific bug with the server in question rather than an issue affecting many users. Instead I suggest you look into sub-classing `Retry` and overriding `parse_retry_after()` to have a minimum value for the server you're contacting. You might also try to get in contact with the owners of the server so they're aware of the bad behavior.

PR #2723, opened 2022-08-22
Add outputs.hashes to build action

sethmlarson:
Fix for failure here: https://github.com/urllib3/urllib3/runs/7951585781?check_suite_focus=true


pquentin (approved):
I have no ides what that does but you know what you're doing :)

sethmlarson:
@pquentin It forwards the output of the sha256sum function used within to the job. I forgot to copy-paste that section over from my SLSA playground.

PR #2727, opened 2022-08-25
Add max_retry_wait_lenght to Retry

### LINKED ISSUE #1338: Retry.get_retry_after does't respect time-out

sobelek:
This PR Fixes #1338
I have reused some of the initial implementation from #1587

### DIFF
--- a/test/tz_stub.py
+++ b/test/tz_stub.py
@@ -7,7 +7,7 @@ from typing import Generator, Optional
 import pytest
 
 try:
-    import zoneinfo  # type: ignore[import]
+    import zoneinfo
 except ImportError:
     # Python < 3.9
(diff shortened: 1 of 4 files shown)


test/tz_stub.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 
 try:
-    import zoneinfo  # type: ignore[import]
```
sobelek, line 10: Had to made this change for mypy to run.

sobelek: Or mb nervermind. It might be caused by my local mypy config. Reverted and rerunning CI again

sobelek:
@sethmlarson 
Hey :)
Any chance for review?

PR #2732, opened 2022-08-29
Restrict sdist files in Hatch

sethmlarson:
Hatch doesn't use MANIFEST.in so we should specify which files we want included in the sdist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,3 +64,15 @@ socks = [
 
 [tool.hatch.version]
 path = "src/urllib3/_version.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
+  "/docs",
+  "/dummyserver",
+  "/src",
+  "/test",
+  "/dev-requirements.txt",
+  "/CHANGES.rst",
+  "/README.md",
+  "/LICENSE.txt",
+]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,0 @@
-include README.md CHANGES.rst LICENSE.txt dev-requirements.txt Makefile src/urllib3/py.typed
-recursive-include dummyserver *
-recursive-include test *
-recursive-include docs *
-recursive-exclude docs/_build *


pyproject.toml
```diff
@@ -64,3 +64,18 @@ socks = [
...
+  "/dummyserver",
+  "/src",
+  "/test",
+  "/dev-requirements.txt",
+  "/CHANGES.rst",
+  "/README.md",
+  "/LICENSE.txt",
+]
+exclude = [
+  "/docs/_build"
```
ofek, line 80: No need since it's in https://github.com/urllib3/urllib3/blob/main/.gitignore

ofek (changes requested):

pyproject.toml
sethmlarson: Oooo, Hatch also reads .gitignore, interesting!

ofek: Yup!

- https://hatch.pypa.io/latest/config/build/#vcs
- https://hatch.pypa.io/latest/config/build/#default-file-selection

pquentin (commented):
So while removing MANIFEST.in is obviously good, I'm wondering about the the allowlist because maintaining it adds a small cost (eg. if we add a file we need to remember about it).

Is it only useful for local builds where you could have extra file, or does it also make a difference when building on CI?

pquentin:
This makes sense, thanks. Two questions about the sdist:

 * Since we ship the test directory, should we ship noxfile.py, dev-requirements.txt and possibly mypy-requirements.txt too?
 * Should we include setup.py so that users get a meaningful error when trying to use setup.py install?

sethmlarson:
> Since we ship the test directory, should we ship noxfile.py, dev-requirements.txt and possibly mypy-requirements.txt too?

I was wondering if we should cut everything except src/, README, and LICENSE? But that would only break redistributors so figured just keep exactly what we have currently to not change anything. 

> Should we include setup.py so that users get a meaningful error when trying to use setup.py install?

I don't think so, we're providing a `setup.py` here mostly for people cloning the repo according to the old instructions in the README and old pip installs of github.

pquentin (approved):
This makes sense, the current dist also does not have noxfile.py. I guess leaving the other files out is better, but this PR strictly improves things already, so I'm approving it.

Sorry for the long reviewing delays.

sethmlarson:
@pquentin No worries at all, I've been swamped too, as you know :)

PR #2735, opened 2022-09-13
Fix mypy error in test.conftest

hramezani:


pquentin (approved):
Thanks! LGTM. Do we understand why this only appears now? Is it because we're not pinning the types packages?

hramezani:
> Thanks! LGTM. Do we understand why this only appears now? Is it because we're not pinning the types packages?

I think yes!

PR #2737, opened 2022-09-17
fix type annotation in contrib/pyopenssl

alex:


alex:
I think this should have a "Skip Changelog" label.

hramezani (approved):
LGTM Thanks 👍

PR #2739, opened 2022-09-18
Remove workaround for old Python versions

pquentin:
test_buggy_incomplete_read is still here to attest that the behavior did not change.


sethmlarson (approved):
Nice! Thanks Quentin :)

sethmlarson:
@pquentin Hmm, the test suite is timing out? Any ideas why?

PR #2740, opened 2022-09-19
Set correct workflow permissions

sashashura:
This fixes the Publish workflow failure.


pquentin:
This appears to make sense, but I'll let @sethmlarson review as they've been working on this. We don't want a changelog entry as it's not a user-facing change, so I've added the "Skip Changelog" label.

pquentin (changes requested):
see my comment above

sethmlarson (approved):

PR #2752, opened 2022-10-23
Provisional support for Python 3.12 with SSL/TLS breaking changes

### LINKED ISSUE #2686: Ensure urllib3 can operate without ssl features removed in Python 3.12
Python 3.12 is removing many deprecated features of the `ssl` module:

- `PROTOCOL_TLSv*` (TLSv1, TLSv1_1, etc)
- `PROTOCOL_TLS`
- `SSLVersion.SSLv3`
- `SSLVersion.TLSv1`
- `SSLVersion.TLSv1_1`

See: https://github.com/python/cpython/pull/94599 cc @tiran

Ousret:
Fixes #2686 
First attempt to mitigate the planned changes in cpython:main.

**Immediately related:**
- https://github.com/python/cpython/pull/94599
- https://github.com/python/cpython/issues/94598
- https://github.com/python/cpython/issues/94199
- https://github.com/python/cpython/pull/94203

**Notes:**
The tracking issue was briefly attached to the v2 milestone and then removed. My understanding is that this patch is meant for v1.26.x branch.
The v2.x branch can benefit from a similar but highly simplified version of this patch. Thanks to the more lax constraint >= 3.7

**Main problems:**
- ssl.PROTOCOL_TLS is in fact deprecated but not yet removed. It is now worth `NotImplemented`.

```
DeprecationWarning: ssl.PROTOCOL_TLS is no longer supported. The constant will be removed in 3.13. Use ssl.PROTOCOL_TLS_CLIENT or ssl.PROTOCOL_TLS_SERVER instead.
```

urllib3 maintainers already took the time to support ssl.PROTOCOL_TLS_CLIENT in urllib3.utils._ssl, 
To avoid propagating PROTOCOL_TLS=NotImplemented, I made it so that it mirrors PROTOCOL_TLS_CLIENT instead in that precise case. (backward-compatibility purposes)

```
>>> from ssl import PROTOCOL_TLS, PROTOCOL_TLS_CLIENT
>>> PROTOCOL_TLS
<_SSLMethod.PROTOCOL_TLS: 2>
>>> PROTOCOL_TLS_CLIENT
<_SSLMethod.PROTOCOL_TLS_CLIENT: 16>
```

- In fact, we can no longer set protocols using their respective constants.

```python
ctx = ssl.SSLContext(ssl_version)
ctx.load_cert_chain(certfile, keyfile)
ctx.verify_mode = cert_reqs
ctx.maximum_version = max_ssl_version
```

Where `max_ssl_version` is one of `ssl.TLSVersion` IntEnum and `ssl_version` is either `ssl.PROTOCOL_TLS_CLIENT` or `ssl.PROTOCOL_TLS_SERVER`.

- gh-94598 is still a draft but I am confident that nothing will preclude this from working after further changes in the linked PR.

- 'urllib3.contrib.pyopenssl' module is deprecated but still there until v2, so we have to _slightly_ patch it in order to avoid a crash.

My understanding is that any protocol higher than TLSv1_2 is mapped to SSLv23_METHOD, so I acted accordingly.

- It is okay to use and map IntEnum protocol version when available? 

Like so `_openssl_versions[ssl.TLSVersion.TLSv1_2] = OpenSSL.SSL.TLSv1_2_METHOD`.

- SSLContext no longer accept default `protocol=None`, they decided to enforce the parameter rightfully.

I had to remove the positional parameter. Cf. gh-94598.

**Environment:**

I had no choice but to clone https://github.com/tiran/cpython/tree/py312_ssl_removal and build it locally then use it against the test suite.

There is no suitable python-version action available at the moment. cf.  gh-94598.

**Opening topics:**

- tornado dummy server hangs when `PROTOCOL_TLS_CLIENT` is passed

Make sense, so we fall back to PROTOCOL_TLS_SERVER instead, and everything goes well.

- I feel like we introduce some ambiguities about ssl_version and max_ssl_version.

ssl_version tends to not do what it seems like it is supposed to. Maybe a topic of discussion there.

Local run: https://github.com/Ousret/urllib3-experimental

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,16 +68,8 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v2
 
-      - if: ${{ !endsWith(matrix.python-version, '-dev') }}
-        name: "Setup Python ${{ matrix.python-version }}"
+      - name: "Setup Python ${{ matrix.python-version }}"
         uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      # Use deadsnakes for development versions
-      - if: endsWith(matrix.python-version, '-dev')
-        name: "Setup Python ${{ matrix.python-version }}"
-        uses: deadsnakes/action@v2.1.1
         with:
           python-version: ${{ matrix.python-version }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
         os: [macos-latest, windows-latest, ubuntu-latest]
         experimental: [false]
         nox-session: ['']
@@ -56,10 +56,10 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: app_engine
-          - python-version: 3.11-dev
+          - python-version: "3.12-dev"
             os: ubuntu-latest
             experimental: true
-            nox-session: test-3.11
+            nox-session: test-3.12
 
     runs-on: ${{ matrix.os }}
     name: ${{ fromJson('{"macos-latest":"macOS","windows-latest":"Windows","ubuntu-latest":"Ubuntu"}')[matrix.os] }} ${{ matrix.python-version }} ${{ matrix.nox-session}}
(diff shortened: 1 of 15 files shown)


pquentin:
Thanks, this is quite proactive! Unfortunately, we can't merge this until 1/ https://github.com/python/cpython/pull/94599 is merged (as its contents could change) and 2/ we have a way to test this in CI (to notice any regressions).

That said, can you please look into the remaining failures?

Ousret:
Indeed, it would be wiser to wait for the merge. As for the CI, I expect that a "3.12-dev" should be available soon. According to PEP 693, alpha1 is scheduled for today. I will wait until the GHA python-version is available to include it.

For the remaining failures:
- Coverage
  - Because we don't have the 3.12-session and miss a line in utils.\_\_init\_\_.py, I can confirm that I hit the line properly.
- Hang in some cases
  - I will try to look into it when I can.
- readthedocs
  - Crash that does seems related to urllib3

sethmlarson:
Can be rebased once https://github.com/urllib3/urllib3/pull/2760 is merged.

Ousret:
The core maintainer (cpython) isn't able to allocate time for that change, so it seems. There are two sides to this news, first, there is no need for that patch until a future version, and it is likely that urllib3 will evolve until then. Secondly, sadly it still ships with deprecated and potentially unusable constants.

I knew that was a potential risk opening that PR without a prior merge in cpython:main, but this was an excellent opportunity to learn new things. This (PR) will become obsolete quickly as things move on.
So I am closing this, in favor of not having many PRs opened, and queued.

PR #2763, opened 2022-11-06
Do not mention deprecated urllib3[secure] as an option how to install certifi

mila:
The example can be understood as recommending using urllib3[secure] to install certifi. For most cases, `python -m pip install certifi` should be OK.

Users can still learn about urllib3[secure] from the warning that immediately follows the removed example.


sethmlarson (approved):
Nice catch, thanks!

PR #2768, opened 2022-11-08
Create BaseHTTPConnection APIs

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/1985

Started by taking apart and documenting the HTTP connection (+pool) lifecycle and what would be necessary at every stage. Decided it might be good to document these in notes that we can keep up to date once we settle on an API.

In general I'd like to fix up a bit of our super complicated proxy situation (cc @jalopezsilva) since we technically "own" the HTTPConnection API. Timeouts are similarly complicated but done in multiple ways.

There's some more features I want to dig into and document like `HTTPConnection.auto_open` which appears to be to do with whether a connection is reusable.

I have some local work creating a `BaseHTTPConnection` protocol and trying it with type hints but that's not quite ready yet.


notes/connection-lifecycle.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Current implementation
+
+`HTTPConnection` should be instantiated with `host` and `port` of the
+**first origin being connected to** to reach the target origin. This either means
+the target origin itself or the proxy origin if one is desired.
+
+If the connection is an `HTTPSConnection`, then `HTTPSConnection.set_cert()` should
+be called once with information about how to verify certificates. (What is stopping us
+from putting this information in `HTTPSConnection.__init__()`?)
```
pquentin, line 11: Good question, I don't see any reason to do that. `set_cert` does not require an active connection and moving it into `__init__` would simplify the current code and let us get rid of `_prepare_conn`.

notes/connection-lifecycle.md
```diff
@@ -0,0 +1,77 @@
...
+
+If using CONNECT tunneling with the proxy, call `HTTPConnection.set_tunnel()`
+with the tunneled host, port, and headers. This will mark the `HTTPConnection`
+as not-reusable.
+
+(If CONNECT tunneling and TLS-in-TLS is required, the HTTPSConnectionPool also sets `HTTPConnection.tls_in_tls_required` to true, is this required? Can the connection detect this from `proxy` and `proxy_config` reliably?)
```
pquentin, line 17: Great point, the connection also has access to `proxy` as `self.proxy`, so we can add the exact same condition in the connection itself.

notes/connection-lifecycle.md
```diff
@@ -0,0 +1,77 @@
...
+### Downsides
+
+- Lots of internal property usage by connection pools/response.
+  - Checking whether a connection is established
+  - Accessing `.sock` property
+- Proxies seem super clunky to setup even though we're giving all the information to the connection pool/connection?
+- Timeouts are getting set throughout an HTTPConnection's lifecycle by the connection pool
+- Does `HTTPConnection.set_tunnel()` give us anything? Remove? No-op?
+
+## How could things be better?
```
pquentin, line 65: This is most probably irrelevant, but I wanted to mention that having a pool-of-pools (ie. having both PoolManager and HTTPConnectionPool) is weird.

pquentin (approved):
This is great, and I love how this highlights improvements that we could carry on today.

The internal property usage is a big problem.

Have you studied HTTP1Connection from the v2 branch and hip? https://github.com/python-trio/hip/blob/e198149c677edbeec023aeb934758c9195a4d2e3/src/ahip/connection.py#L317-L565? I think there are valuable lessons there.

Finally, I think those notes stand on their own, and I would happy to merge them as-is.

notes/connection-lifecycle.md
sethmlarson: Totally weird, we need to create a better recipe for a single connection pool of mixed-config connections.

sethmlarson:
Noting this down because I have to leave: default `port` when specifying a `scheme` in `HTTPConnection.set_tunnel()`.

changelog/1985.removal.rst
```diff
@@ -0,0 +1,17 @@
+Added the ``scheme`` parameter to ``HTTPConnection.set_tunnel`` to configure the scheme of the origin being tunnelled to.
```
pquentin, line 1: Should we mention that all those changes are there to distance ourselves from http.client a bit and allow future HTTP implementations in the future such as `h11`?

.coveragerc
```diff
@@ -25,5 +25,5 @@ exclude_lines =
     .*:.* # Python \d.*
     .* # Abstract
     .* # Defensive:
-    if TYPE_CHECKING:
+    if (?:typing.)?TYPE_CHECKING:
```
pquentin, line 28: nit: I would prefer using the same form throughout the project

ci/0005-botocore-fakesocket-settimeout.patch
```diff
@@ -0,0 +1,14 @@
...
+index 22bd9a7..c9d4060 100644
+--- a/tests/unit/test_awsrequest.py
++++ b/tests/unit/test_awsrequest.py
+@@ -58,6 +59,9 @@ class FakeSocket:
+     def close(self):
+         pass
+ 
++    def settimeout(self, value):
++        pass
++
```
pquentin, line 11: Should we report this upstream?

notes/connection-lifecycle.md
```diff
@@ -0,0 +1,83 @@
+# Connection lifecycle
+
+## Current implementation
```
pquentin, line 3: Should we update those notes so that they remain true after we merge the pull request? They still mention `set_cert`, `auto_open`, etc.

src/urllib3/connection.py
```diff
@@ -548,25 +595,24 @@ def set_cert(
...
+            if self._tunnel_scheme == "https":
                 self.sock = sock = self._connect_tls_proxy(self.host, sock)
                 tls_in_tls = True
 
-            self._connecting_to_proxy = False
+            # If we're tunneling it means we're connected to our proxy.
+            self._has_connected_to_proxy = True
 
-            self._tunnel()
+            self._tunnel()  # type: ignore[attr-defined]
```
pquentin, line 613: Why is mypy complaining here?

pquentin (commented):
Awesome work. :clap:

ci/0005-botocore-fakesocket-settimeout.patch
sethmlarson: Yes I'll create a PR right after we merge.

.coveragerc
sethmlarson: Agreed, I've created this issue: https://github.com/urllib3/urllib3/issues/2789

src/urllib3/connection.py
sethmlarson: `_tunnel` isn't a part of the [typeshed definition for `http.client.HTTPConnection`](https://github.com/python/typeshed/blob/805519717b7213ed232593931df815fd6c839ff5/stdlib/http/client.pyi#L139).

notes/connection-lifecycle.md
sethmlarson: Done in https://github.com/urllib3/urllib3/pull/2768/commits/f25fec70eb7f4661ad8f4692fc6d8997413c122d

changelog/1985.removal.rst
sethmlarson: Done in https://github.com/urllib3/urllib3/pull/2768/commits/afc660e87256768c13d0f0c0330b94bf6c0bb3c7

notes/connection-lifecycle.md
```diff
@@ -6,78 +6,125 @@
...
 
-Set the connection's `timeout` property to the "connect timeout" value.
+conn.set_tunnel("example.com", scheme="http", headers={"Proxy-Header": "value"})
+```
 
 Connect to the first origin by calling the `HTTPConnection.connect()` method.
 If an error occurs here you can check whether the error occurred during the
-connection to the proxy if `HTTPConnection._connecting_to_proxy` is true.
+connection to the proxy if `HTTPConnection.has_connected_to_proxy` is true.
 If the value is false then the error didn't occur while connecting to a proxy.
```
pquentin, line 43: So if `HTTPConnection.has_connected_to_proxy` is `False`, you haven't connected to the proxy, and the error occurred while connecting to it? (It seems that `_connecting_to_proxy` and `has_connected_to_proxy` are opposites so the wording should change.)

notes/connection-lifecycle.md
```diff
@@ -0,0 +1,130 @@
...
+If the value is false then the error didn't occur while connecting to a proxy.
+
+```python
+# Explicitly connect to the origin. This isn't
+# required as sending the first request will
+# automatically connect if not done explicitly.
+conn.connect()
+```
+
+After connecting to the origin, the connection can be checked to see if `is_verified` is set to true. If not the `HTTPConnectionPool` would emits a warning. The warning only matters for when verification is disabled, because otherwise an error is raised on unverified TLS handshake.
```
pquentin, line 52:
```suggestion
After connecting to the origin, the connection can be checked to see if `is_verified` is set to true. If not the `HTTPConnectionPool` would emit a warning. The warning only matters for when verification is disabled, because otherwise an error is raised on unverified TLS handshake.
```

notes/connection-lifecycle.md
```diff
@@ -6,78 +6,125 @@
...
+If the read timeout is different from the connect timeout then the
+`HTTPConnection.timeout` property can be changed at this point.
 
-Then response headers (and other info) are read from the connection via `HTTPConnection.getresponse()` and returned as a `urllib3.HTTPResponse`. The `HTTPResponse` instance carries a reference to the `HTTPConnection` instance. (Within `HTTPConnectionPool` the `HTTPResponse._connection` is set to `response_conn`, is that needed or can `HTTPConnection.getresponse()` handle this?)
+```python
+conn.timeout = 5.0  # (read timeout)
+```
 
-If pooling is in use set `_connection` and `_pool` on the `HTTPResponse` instance. If retries are in use set `retries` on the `HTTPResponse` instance.
+Then the HTTP request can be sent with `HTTPConnection.request()`. If a `BrokenPipeError` (should we include `SSLEOFError`?) is raised while sending the request body it can be swallowed as a response can still be received from the origin even when the request isn't completely sent.
```
pquentin, line 68: Regarding your question, I'm not clear on whether you can read anything after `SSLEOFError` was raised. If we can read, then it's worth swallowing? Or is it something actionable where the urllib3 user should change something and retry?

pquentin (approved):
This looks good to me, thank you! I've left a few comments, but this does not need another round of review. :tada:

vEpiphyte:
For reference this did not document the removal of the VerifiedHTTPSConnection class ( despite its rename in #1805 )

pquentin:
@vEpiphyte `VerifiedHTTPSConnection` is still here in urllib3 2.0, we just stopped importing it from connectionpool.py. If this is where you were importing it from, your code was only working by accident.

You can use `from urllib3.connection import VerifiedHTTPSConnection` instead but of course at this point it would be better to use the new name with `from urllib3.connection import HTTPSConnection`.

Why did you need this?

vEpiphyte:
@pquentin Looking into this issue in vcrpy ( which you commented on, thank you very much! ) https://github.com/kevin1024/vcrpy/issues/688 since that has caused an immediate outage for myself and my downstream users.

Almost more so a note for myself, since git bisect & github's compare feature didn't quickly tell me where the removal of the import came from.

PR #2777, opened 2022-11-09
Fix Tornado event_loop warning on Python 3.10 and later

### LINKED ISSUE #2772: Fix "DeprecationWarning: There is no current event loop" in test suite
Starting with Python 3.10, using `asyncio.get_event_loop()` in a synchronous context is [deprecated in favor of `asyncio.run`](https://bugs.python.org/issue39529) and will stop working eventually. This is why we get a warning when calling `tornado.ioloop.IOLoop.current()` in our test suite: it calls `asyncio.get_event_loop()` under the hood.

Tornado has always encouraged users to do what we do: setup the server synchronously, and then start it in a background thread. This was quite cool as we did not have to worry about async at all. To address the Python 3.10 deprecation, [Tornado 6.2](https://www.tornadoweb.org/en/stable/releases/v6.2.0.html) now reluctantly encourages calling `asyncio.run()` and doing the setup in async code. However, just making `run_tornado_app` async won't be enough in our case because our synchronous test code needs to know the port number of the server. And retrieving that value from the background thread is likely to be annoying.

What we can do instead is partially setup the server synchronously (and thus get the port number) and then finalize setup in async. It does require some synchronization but I now have a working proof-of-concept that passes even with `-Werror`:

```python3
import asyncio
import threading

import tornado.web

import urllib3


class MainHandler(tornado.web.RequestHandler):
    def get(self):
        self.write("Hello, world")


class TornadoServerThread(threading.Thread):
    def __init__(self):
        super().__init__()
        app = tornado.web.Application([(r"/", MainHandler)])

        self.server = tornado.httpserver.HTTPServer(app)
        self.sockets = tornado.netutil.bind_sockets(None, address="localhost")
        self.port = self.sockets[0].getsockname()[1]

        self.started = threading.Event()
        self.stopped = asyncio.Event()

    async def main(self):
        self._loop = asyncio.get_event_loop()
        self.server.add_sockets(self.sockets)
        self.started.set()
        await self.stopped.wait()

    def run(self):
        print("run")
        asyncio.run(self.main())

    def stop(self):
        print("stop")
        self._loop.call_soon_threadsafe(self.stopped.set)


t = TornadoServerThread()
t.start()
t.started.wait()
print(t.port)
print(urllib3.request("GET", f"http://localhost:{t.port}").data)
t.stop()
```

(I initially tried to use `IOLoop(make_current=False)` as suggested in https://github.com/tornadoweb/tornado/issues/3156 but did not manage to make it work without warnings.)

This issue is about taking this proof-of-concept (or coming with a better one) and use it in urllib3's test suite. We should then verify that running with Python 3.10 and Tornado 6.2 does not raise any deprecation warning.

graingert:
Fixes #2772

### DIFF
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -4,6 +4,11 @@
 Dummy server used for unit testing.
 """
 
+from __future__ import annotations
+
+import asyncio
+import concurrent.futures
+import contextlib
 import logging
 import os
 import socket
@@ -11,8 +16,9 @@ import ssl
 import sys
 import threading
 import warnings
+from collections.abc import Coroutine, Generator
 from datetime import datetime
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, TypeVar
 
 import tornado.httpserver
 import tornado.ioloop
@@ -25,10 +31,15 @@ from cryptography.hazmat.primitives import serialization
 from urllib3.exceptions import HTTPWarning
 from urllib3.util import ALPN_PROTOCOLS, resolve_cert_reqs, resolve_ssl_version
 
+if TYPE_CHECKING:
+    from typing_extensions import ParamSpec
+
+    P = ParamSpec("P")
+
 log = logging.getLogger(__name__)
 
 CERTS_PATH = os.path.join(os.path.dirname(__file__), "certs")
-DEFAULT_CERTS: Dict[str, Any] = {
+DEFAULT_CERTS: dict[str, Any] = {
     "certfile": os.path.join(CERTS_PATH, "server.crt"),
     "keyfile": os.path.join(CERTS_PATH, "server.key"),
     "cert_reqs": ssl.CERT_OPTIONAL,
@@ -106,7 +117,7 @@ class SocketServerThread(threading.Thread):
         self,
         socket_handler: Callable[[socket.socket], None],
         host: str = "localhost",
-        ready_event: Optional[threading.Event] = None,
+        ready_event: threading.Event | None = None,
     ) -> None:
         super().__init__()
         self.daemon = True
@@ -144,7 +155,7 @@ def ssl_options_to_context(  # type: ignore[no-untyped-def]
     certfile=None,
     server_side=None,
     cert_reqs=None,
-    ssl_version: Optional[Union[str, int]] = None,
+    ssl_version: str | int | None = None,
     ca_certs=None,
     do_handshake_on_connect=None,
     suppress_ragged_eofs=None,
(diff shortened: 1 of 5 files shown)


dummyserver/server.py
```diff
@@ -232,3 +219,77 @@ def encrypt_key_pem(private_key_pem: trustme.Blob, password: bytes) -> trustme.B
...
+
+        print(f"Listening on http://{host}:{port}")
+        await asyncio.Event().wait()
+        return 0
+
+    return asyncio.run(amain())
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
graingert, line 295: using `sys.exit(main())` is idiomatic and means the code gets type checked

dummyserver/testcase.py
```diff
@@ -160,21 +161,25 @@ class HTTPDummyServerTestCase:
...
     @classmethod
     def _start_server(cls) -> None:
-        cls.io_loop = ioloop.IOLoop.current()
-        app = web.Application([(r".*", TestingApp)])
-        cls.server, cls.port = run_tornado_app(
-            app, cls.io_loop, cls.certs, cls.scheme, cls.host
-        )
-        cls.server_thread = run_loop_in_thread(cls.io_loop)
+        with contextlib.ExitStack() as stack:
+            io_loop = stack.enter_context(run_loop_in_thread())
```
graingert, line 169: ExitStack is used so we can use `with ` for the `run_loop_in_thread()` context manager and close it if anything fails during setup then `cls._stack = stack.pop_all()` is used to close the context manager later

pquentin: So if we had a pytest fixture here instead of setup_class/teardown_class, could we use `run_loop_in_thread` as a normal context manager? If yes, can we please do that?

graingert: yes, can I leave that for another PR?

pquentin (changes requested):
Thanks. OK for removing ExitStack in another pull request and for leaving `from __future__ import annotations`. Can you please:

 * Add `error:There is no current event loop:DeprecationWarning:tornado.ioloop` in setup.cfg as `filterwarnings`
 * Upgrade to Tornado 6.2
 * Restore the scheme parameter (this will get rid of the confusing ternary expressions)

graingert:
> Thanks. OK for removing ExitStack in another pull request and for leaving `from __future__ import annotations`. Can you please:
> 
>     * Add `error:There is no current event loop:DeprecationWarning:tornado.ioloop` in setup.cfg as `filterwarnings`
> 
>     * Upgrade to Tornado 6.2
> 
>     * Restore the scheme parameter (this will get rid of the confusing ternary expressions)

done!

pquentin (approved):
Thanks you! Looks good to me.

PR #2779, opened 2022-11-09
add from __future__ import annotations with isort

graingert:


sethmlarson (commented):
Would appreciate not merging this until after Quentin and I's hackathon week as there will likely be merge conflicts.

I don't know much about the `annotations` future, does this impact the Python versions we're able to support in any way? v2.0 supports Python 3.7 and later.

setup.py
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 # This file is protected via CODEOWNERS
 
+from __future__ import annotations
+
```
pquentin, line 5: We need to exclude `setup.py` because it needs to suport Python 2.7. This is why the `unsupported_setup_py` check is failing. Skipping the whole file seems to be the easiest way: https://pycqa.github.io/isort/docs/configuration/action_comments.html#isort-skip_file.

pquentin (changes requested):
Yes, let's wait until the week is over to avoid conflicts.

So the good part is that it does work in Python 3.7.0 and above! However, we don't yet know if it will be mandatory at some point. https://docs.python.org/3/library/__future__.html reads:

> `from __future__ import annotations` was previously scheduled to become mandatory in Python 3.10, but the Python Steering Council twice decided to delay the change ([announcement for Python 3.10](https://mail.python.org/archives/list/python-dev@python.org/message/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/); [announcement for Python 3.11](https://mail.python.org/archives/list/python-dev@python.org/message/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/)). No final decision has been made yet. See also [PEP 563](https://peps.python.org/pep-0563/) and [PEP 649](https://peps.python.org/pep-0649/).

Also, what tool did you use to make the conversion?

pquentin:
Also, will this affect other tool like editors?

graingert:
> Also, will this affect other tool like editors?

How do you mean?

setup.py
graingert: Yep I'll skip the whole file

pquentin:
> > Also, will this affect other tool like editors?
> 
> How do you mean?

Are there tools that don't support PEP 563? I would not want things like code completion to break or be less useful because of this.

graingert:
> > > Also, will this affect other tool like editors?
> > 
> > 
> > How do you mean?
> 
> Are there tools that don't support PEP 563? I would not want things like code completion to break or be less useful because of this.

pycharm used to lag behind but it's supported now

graingert:
> Also, what tool did you use to make the conversion?

Just your existing pre-commit config with pyupgrade

sethmlarson:
@graingert Definitely like how this looks, and if it's automated with pyupgrade then even better. I've merged the major changes for the week now so if you wanted to pull the latest and run pyupgrade fresh we can get this merged.

pquentin:
Last minute change: I discussed with Seth the results of https://github.com/urllib3/urllib3/issues/2787 and I'm going to try to get #2712 out of the door today so let's wait until next week before merging this `from __future__ import annotations` pull request.

venthur:
The PR minus the conflicts looks good to me.

pquentin:
@graingert OK, now that #2798 is in, we can resume the work here.

pquentin (changes requested):
Can you please look into the docs build failing?

venthur:
I fixed the docs issue and solved the remaining merge conflicts. Sorry again @graingert for accidentally pushing to the wrong upstream.

graingert:
How do you want me to take this PR forward?

venthur:
https://github.com/urllib3/urllib3/pull/2813 has been merged, which contains all your changes plus the ones I added on top. So we can close this one.

PR #2780, opened 2022-11-09
use asyncio.run for tornado

graingert:

### DIFF
--- a/dummyserver/server.py
+++ b/dummyserver/server.py
@@ -4,6 +4,9 @@
 Dummy server used for unit testing.
 """
 
+import asyncio
+import concurrent.futures
+import contextlib
 import logging
 import os
 import socket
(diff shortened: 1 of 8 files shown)


dummyserver/server.py
```diff
@@ -232,3 +215,70 @@ def encrypt_key_pem(private_key_pem: trustme.Blob, password: bytes) -> trustme.B
...
+            elif f is ran:
+                # if this is the first iteration the loop failed to start
+                # if it's the second iteration the loop has finished or
+                # the loop failed to close and we need to raise the exception
+                ran.result()
+                return
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
pquentin, line 284: What is the point of using `sys.exit()` here? Can't we just inline main?

dummyserver/server.py
```diff
@@ -232,3 +215,70 @@ def encrypt_key_pem(private_key_pem: trustme.Blob, password: bytes) -> trustme.B
...
+def main():
+    # For debugging dummyserver itself - python -m dummyserver.server
+    from .handlers import TestingApp
+
+    host = "127.0.0.1"
+
+    async def main():
+        io_loop = tornado.ioloop.IOLoop.current()
+        app = tornado.web.Application([(r".*", TestingApp)])
+        server, port = run_tornado_app(app, io_loop, None, "http", host)
```
pquentin, line 229: Would you say that assert io_loop == tornado.ioloop.IOLoop.current() is still useful in run_tornado_app?

dummyserver/testcase.py
```diff
@@ -163,18 +164,21 @@ class HTTPDummyServerTestCase:
...
     @classmethod
     def _start_server(cls) -> None:
-        cls.io_loop = ioloop.IOLoop.current()
-        app = web.Application([(r".*", TestingApp)])
-        cls.server, cls.port = run_tornado_app(
-            app, cls.io_loop, cls.certs, cls.scheme, cls.host
-        )
-        cls.server_thread = run_loop_in_thread(cls.io_loop)
+        with contextlib.ExitStack() as stack:
+            io_loop = stack.enter_context(run_loop_in_thread())
```
pquentin, line 168: What is the point of an `ExitStack` if you'll only have one `enter_context` call?

pquentin (changes requested):
Thank you! This looks great, in particular the error handling is much better. I'm wondering if this will help with threads that used to print exception without pytest seeing them.

I've made a few changes - we don't want to hide all warnings, but fix them, like we're doing for this one.

I left a few comments. Can you please also fix the mypy errors? `nox -rs mypy` should be clean.

PR #2783, opened 2022-11-10
Stop considering LibreSSL for HAS_NEVER_CHECK_COMMON_NAME

pquentin:
We don't support LibreSSL at all since https://github.com/urllib3/urllib3/pull/2705.


sethmlarson (approved):
LGTM, should we be doing something to warn users about using Python compiled with non-OpenSSL? Warning or error?

sethmlarson:
I think since we're documenting this in the changelog we can just merge this.

pquentin:
We actually raise an ImportError in `urllib3/__init__.py` so this was unreachable code

PR #2784, opened 2022-11-10
add vscode devcontainer and vscode specific settings

### LINKED ISSUE #2782: Enhance the urllib3 developer experience
### Context

I would like to make it easier for people to contribute to `urlib3`. 

I would like to start with the following:
- Add a contributors guide to the repository
 - Add a [vscode devcontainer](https://code.visualstudio.com/docs/devcontainers/containers) to be able to run most of the checks that we do in GitHub actions. That means being able to do the following:
   - Install the correct python versions
   - Install and setup pre-commit
   - Install and configure nox
- Add vscode settings to the repository. This will allow vscode users to have code formatting, linting and import sorting while they are writing code. This is a nicer experience than writing a bunch of code, trying to commit and having pre-commit reject it.

I already have completed some of this work for when I submitted #2649, I just need to clean it up and submit a PR.

I found getting started with contributing to `urllib3` was harder than my usual experience with open-source projects and I think lowering the barrier to entry would be impactful for this project.

I know lots of these features are specific to vscode, but they would not hinder non-vscode users from contributing. I think it would still help non-vscode users because we can point them to the vscode settings in this repo that show what tooling is being used and with what settings. They can use that as a reference to set up their preferred editor/IDE. I will also make sure to have something in the contributing guide for non-vscode users.

### Alternatives

I think we can do nothing but I know `urllib3` is trying to attract contributors.

### Contribution

I have everything ready to go. Just need to clean up some stuff.

shadycuz:
Closes #2782 

With these changes, anyone using vscode and the [devcontainer](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension. Will be able to do the following with 0 setup or effort. 

- Run `pre-commt -a` (just like in actions)
- pre-commit will also now run on every commit
  > image
- run `pytest`
- run nox against a single python version `nox -p 3.9`
- run nox against 3.7, 3.8, 3.9 and 3.10 `nox -r -s test` (tests only)
- run tests from inside vscode in the side bar
  > image
- run tests from inside the editor while writing tests
  > image
- automatic `black` formatting when saving python files
- automatic `isort` import sorting when saving python files
- automatic `flake8` checking while writing code
  > image
- automatic `mypy` type checking while writing code
  > image
- [GItLens](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens) which has many great features, but a favorite of mine is git history in the editor
  > image
- possibly others that I am forgetting about? 

Cons:
- Since we are building 4 pythons from source, it takes a long time the first time you start the devcontainer. But it's docker so it's a one-time thing, subsequent starts will be nearly instant.
- Some of the tests are not working in the docker container and I have not investigated why
  > image

Todo:
- [ ] Create a contributing guide
- [ ] Update README to point to the contributing guide
- [ ] Speed up devcontainer creation by moving to a pre-built image with python versions pre-installed


sethmlarson:
I appreciate the work you've put into this, however I don't see devcontainers any different than I see things like editor configurations files. IMO they shouldn't be committed to the repository. This especially seems like a ton more stuff that we'd have to maintain and keep up to date as things change, and that's already something we have enough of already.

Thanks again, but I believe we'll be declining this PR.

PR #2790, opened 2022-11-10
configure filterwarnings=error

graingert:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,6 +101,7 @@ filterwarnings = [
     '''default:ssl\.PROTOCOL_TLSv1_2 is deprecated:DeprecationWarning''',
     '''default:unclosed .*:ResourceWarning''',
     '''default:ssl\.wrap_socket\(\) is deprecated, use SSLContext\.wrap_socket\(\):DeprecationWarning''',
+    '''default:ssl NPN is deprecated, use ALPN instead:DeprecationWarning''',
 ]
 
 [tool.isort]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,7 +94,6 @@ filterwarnings = [
     '''default:No IPv6 support. skipping:urllib3.exceptions.HTTPWarning''',
     '''default:ssl\.TLSVersion\.TLSv1 is deprecated:DeprecationWarning''',
     '''default:ssl\.PROTOCOL_TLS is deprecated:DeprecationWarning''',
-    '''default:ssl\.SSLContext\(\) without protocol argument is deprecated.:DeprecationWarning''',
     '''default:ssl\.PROTOCOL_TLSv1 is deprecated:DeprecationWarning''',
     '''default:ssl\.TLSVersion\.TLSv1_1 is deprecated:DeprecationWarning''',
     '''default:ssl\.PROTOCOL_TLSv1_1 is deprecated:DeprecationWarning''',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,8 +80,27 @@ include = [
 [tool.pytest.ini_options]
 xfail_strict = true
 python_classes = ["Test", "*TestCase"]
+addopts = ["--strict-markers", "--strict-config"]
 filterwarnings = [
-    "error:There is no current event loop:DeprecationWarning:tornado.ioloop",
+    "error",
+    '''default:'urllib3\[secure\]' extra is deprecated and will be removed in urllib3 v2\.1\.0.*:DeprecationWarning''',
+    '''default:'urllib3\.contrib\.pyopenssl' module is deprecated and will be removed in urllib3 v2\.1\.0.*:DeprecationWarning''',
+    '''default:'urllib3\.contrib\.securetransport' module is deprecated and will be removed in urllib3 v2\.1\.0.*:DeprecationWarning''',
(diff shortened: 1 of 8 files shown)


pquentin:
We have this mysterious coverage issue which does not make any sense to me:

```
src/urllib3/util/ssltransport.py           146     24    84%   86, 89, 92, 100, 156, 164-166, 169, 173-174, 177-179, 182, 201, 204, 207, 210, 213, 216, 222, 234, 293
```

pyproject.toml
```diff
@@ -76,3 +76,28 @@ include = [
   "/README.md",
   "/LICENSE.txt",
 ]
+
+[tool.pytest.ini_options]
+addopts = ["--strict-markers", "--strict-config"]
```
graingert, line 81:
```suggestion
addopts = ["--strict-markers", "--strict-config"]
python_classes = ["Test", "*TestCase"]
```

graingert:
@pquentin ok I fixed that weird coverage issue by rebasing on your pyproject.toml pytest config

test/test_ssltransport.py
```diff
@@ -116,10 +116,13 @@ def start_dummy_server(
...
-                request = consume_socket(ssock)
-                validate_request(request)
-                ssock.send(sample_response())
+            try:
+                with self.server_context.wrap_socket(sock, server_side=True) as ssock:
+                    request = consume_socket(ssock)
+                    validate_request(request)
+                    ssock.send(sample_response())
+            except (ConnectionAbortedError, ConnectionResetError):
+                return
```
pquentin, line 125: Should we use `contextlib.suppress` here and below?

Also, in what context is it useful? It's not intuitive that this could raise an exception and not actually start the server

test/with_dummyserver/test_https.py
```diff
@@ -244,7 +244,7 @@ def test_verified(self) -> None:
                 r = https_pool.request("GET", "/")
                 assert r.status == 200
 
-            assert w == []
+            assert [str(wm) for wm in w] == []
```
pquentin, line 247: Oh, did you use that to help you debug? Because the two statements are equivalent otherwise

test/with_dummyserver/test_socketlevel.py
```diff
@@ -25,6 +25,7 @@
     resolvesLocalhostFQDN,
 )
 from threading import Event
+from typing import TYPE_CHECKING
```
pquentin, line 28: Can you please import typing instead and use `typing.TYPE_CHECKING`?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -137,6 +143,38 @@ def socket_handler(listener: socket.socket) -> None:
                 ), "missing ALPN protocol in SSL handshake"
 
 
+def original_ssl_wrap_socket(
```
pquentin, line 146: Why call it original? Is the idea that we continue using the original interface of `ssl_wrap_socket`?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -137,6 +143,38 @@ def socket_handler(listener: socket.socket) -> None:
...
+    server_side: bool = False,
+    cert_reqs: ssl.VerifyMode = ssl.CERT_NONE,
+    ssl_version: int = ssl.PROTOCOL_TLS,
+    ca_certs: str | None = None,
+    do_handshake_on_connect: bool = True,
+    suppress_ragged_eofs: bool = True,
+    ciphers: str | None = None,
+) -> ssl.SSLSocket:
+    if server_side and not certfile:
+        raise ValueError("certfile must be specified for server-side " "operations")
```
pquentin, line 159:
```suggestion
        raise ValueError("certfile must be specified for server-side operations")
```

pyproject.toml
pquentin: Can we use noxfile.py for this? Then have a pull request that puts everything into addopts maybe?

pquentin (changes requested):
Thanks, excellent work!

test/test_ssltransport.py
graingert: You don't use `contextlib.suppress` anywhere else and I find it to be confusing compared to the regular try/catch

test/with_dummyserver/test_socketlevel.py
graingert: You already have a function called ssl_wrap_socket but it's part of the system under test. Happy to take suggestions on the name

test/test_ssltransport.py
pquentin: OK, but in what context is it useful? It's not intuitive that this could raise an exception and not actually start the server

pquentin (changes requested):
only one question left!

test/test_ssltransport.py
graingert: previously it was as if there was a `try: .... except BaseException: return` there, this is actually just limiting the exceptions to just those. This is because effectively all exceptions in a thread are ignored by default and the filterwarings=error collects those exceptions so we need to do something about the ones that do happen

rather than add PytestUnhandledThreadExceptionWarning to the filterwarnings, as a TODO, I think it's better to do the narrower workaround here

pquentin (approved):
Thanks! LGTM.

Issue #2645, opened 2022-06-19
Connect timeout does not apply to underlying socket when reusing a connection from the pool


Ryangwaite:
### Subject

When making an HTTP(s) request to a host for which a connection has already been established, this connection is reused from the pool however the timeout provided is only applied to the HTTPSConnection/HTTPConnection object but not it's underlying socket (the original requests timeout remains). Therefore, if the initial request has a short timeout and a subsequent request is initiated that happens to take more time than the initial, the following error will be raised in accordance with the original requests timeout:

```
urllib3.exceptions.ProtocolError: ('Connection aborted.', TimeoutError('The write operation timed out'))
```
Note this bug applies to HTTPConnections and HTTPSConnections using the stdlib SSLContext backend but not HTTPSConnections using the PyOpenSSLContext backend.

### Environment
OS: OS Linux-5.4.0-120-generic-x86_64-with-glibc2.31
Python: 3.10.5
urllib3: main (f9c7434)

The docker-compose environment to reliably reproduce this in can be found [here](https://github.com/Ryangwaite/urllib3-docker-dev-env).

### Steps to Reproduce

```python
import time
import urllib3
from urllib3.util import Timeout

TOKEN = "uploadtoken"
BASE_URL = "http://upload-server"
# BASE_URL = "https://upload-server"  # Fails for HTTPS too when using stdlib SSLContext but not PyOpenSSLContext

http = urllib3.PoolManager(cert_reqs='CERT_NONE')

# 1. Seed the timeout on the conn
r = http.request("GET", f"{BASE_URL}/files/test.txt?token={TOKEN}", timeout=Timeout(connect=5, read=6))

# 2. Upload file that takes >5 seconds to complete
# 500MB file at 500mbps bandwidth uploads in ~10.5s with the following code
with open("/var/data-for-upload/500MB.bin") as fp:
    file_data = fp.read()
upload_start = time.time()
try:
    r = http.request("POST", f"{BASE_URL}/upload?token={TOKEN}", fields={"file": ("500MB.bin", file_data)}, timeout=15)
finally:
    print(f"Upload request finished in {time.time() - upload_start:.3f} seconds")

```

### Expected Behavior

For the scenario above, the second request should succeed and be bounded by its 15s timeout.

And the behaviour of the stdlib SSLContext and PyOpenSSLContext backends should match.

### Actual Behavior

The second request fails with:
```
Upload request finished in 6.317 seconds
Traceback (most recent call last):
[...]

PR #2792, opened 2022-11-10
Fix socket timeout value when HTTPConnection is reused

sethmlarson:
Closes https://github.com/urllib3/urllib3/issues/2645

Found this while looking at https://github.com/urllib3/urllib3/issues/2645. Figured out the `timeout.connect_timeout` was getting set to `HTTPConnection.timeout` which could potentially contain the `_DEFAULT_TIMEOUT` sentinel which blows up when passed to `settimeout`. So this PR also fixes a type error that was trying to tell us something was amiss. Another win for types!


pquentin (commented):
Thanks! This looks good. However, for new connections, where do we set the read timeout? Should we continue doing it now that we have this new settimeout call?

sethmlarson:
@pquentin The read `sock.settimeout` call occurs within `HTTPConnection.getresponse()`: https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L452

pquentin (dismissed):
Oh, this is about the connect timeout. This makes sense, thanks.

PR #2794, opened 2022-11-11
Temp fix for SLSA generator

laurentsimon:
Sigstore made a breaking change as part of their recent GA announcement. We need a temporary fix to avoid builder failure (see https://github.com/slsa-framework/slsa-github-generator/issues/1163)

/cc @asraa @ianlewis


sethmlarson:
Nice timing, we were planning on making a release in the next few days. Much appreciated!

sethmlarson (approved):

Issue #2787, opened 2022-11-10
How can we minimize memory consumption when decoding data?


pquentin:
For the needs of https://github.com/urllib3/urllib3/pull/2712, I'm looking for a queue-like data structure that:

 * holds bytes
 * supports put()ing an arbitrary number of bytes
 * supports get()ing an arbitrary number of bytes
 * consumes O(size()) memory
 * is less complex than a [rope](https://en.wikipedia.org/wiki/Rope_(data_structure))!

pquentin:
The current solution is a bytearray, which is pretty bad because it consumes three times the amount of bytes for get() operations:
 * the data still in the buffer
 * the byterarray slice
 * the bytes conversion of that slice


Here's a self-contained illustration of that problem:

```python3
from urllib3.response import MultiDecoder


class ByteArrayDecoder:
    def __init__(self):
        self.buffer = bytearray()
        self.decoder = MultiDecoder("gzip")

    def put(self, compressed_data: bytes):
        decoded = self.decoder.decompress(compressed_data)
        self.buffer.extend(decoded)

    def get(self, n: int):
        decoded_data = bytes(self.buffer[:n])
        del self.buffer[:n]
        return decoded_data
```

I'm using this code on the following test data (2GiB of zeroes, this compresses to 2MiB of data):

```python3
import gzip

with open("zeroes.gc", "w") as f:
    f.write(bytes(2**31))
```

And then I run the following code using memray:

```python3
with open("zeroes.gz", "rb") as f:
    compressed = f.read()

decoder = ByteArrayDecoder()
decoder.put(compressed)
decoder.get(2**31)
```


We can use a memoryview to avoid one of the extra copies:

```python3
class MemoryViewDecoder:
    def __init__(self):
        self.buffer = bytearray()
        self.decoder = MultiDecoder("gzip")

    def put(self, compressed_data: bytes):
        decoded = self.decoder.decompress(compressed_data)
        self.buffer.extend(decoded)

    def get(self, n: int):
        buffer_view = memoryview(self.buffer)
        decoded_data = bytes(buffer_view[:n])
        del buffer_view
        del self.buffer[:n]
        return decoded_data
```

Now both `put` and `get` do an extra copy, so we're looking at 4GiB of memory use. Hmm. put() too!? Let's confirm that:

```python3
from urllib3.response import MultiDecoder

with open("zeroes.gz", "rb") as f:
    compressed = f.read()
decoder = MultiDecoder("gzip")
[...]

Issue #2769, opened 2022-11-08
Should HTTPResponse support decompression with raw reads?


pquentin:
### Context

https://github.com/urllib3/urllib3/pull/2712 is about to close #2128 and #709. However, one piece is missing. In https://github.com/urllib3/urllib3/issues/709#issuecomment-144356586, @\Lukasa mentioned 7 years ago that if we were to implement #2712, we should also make sure we don't take away any important capability from urllib3. Here's the full comment:

> If we want to satisfy all use cases, we're in a really tricky place. The space of options and whether they're currently supported is:
> 
> |  | Decompress | Don't Decompress |
> | --- | :-: | :-: |
> | **Bytes read from the wire** | ✓ | ✓ |
> | **Bytes received in the application** | ✗ | ✓ |
> 
> Put another way, if you are about the number of bytes you read _from the wire_, you can control that perfectly with urllib3 today, regardless of `decode_content`. If you care about the number of bytes your application receives, you can only control that if you set `decode_content` to `False` and handle the decoding yourself.
> 
> I have no objection to supporting that final box, but I think we should consider really carefully whether we want to do it by adding a cross mark in an already supported box. The proposal in this issue (to change the semantic of `read`) would do that, _unless_ it was accompanied by either another parameter or a new function that preserves the current logic.
> 
> I am +1 on supporting _all_ of the above. I am -1 on supporting a different configuration of the above (i.e. moving the cross to another box). I am +0.1 on doing nothing (because the status quo has some value). I am open to relegating the current 'default' logic to a backup method (`raw_read` or some such thing), and moving @gilesbrown's suggestion to the `read` method, though it will break a surprising amount of code.

_Originally posted by @\Lukasa in https://github.com/urllib3/urllib3/issues/709#issuecomment-144356586_
      
This was unclear for me for a while, so I'll put it yet another way.
 1. Most users care about the bytes they receive in their application. And in the face of decompression, then having `read()` returning a random number of bytes, including sometimes 0, is a bug.
 2. Some users are instead interested in controlling precisely the bytes they read from the wire: each `read()` call issues a single system call and that's it. And maybe they want decompression as well, and don't mind if `read()` returns 0 bytes or too many bytes because the important part is that `read(n)` reads n bytes.

What #2712 does when fixing that bug is breaking the second set of users that want 1/ read() to issue single sytem calls and 2/ decompression.

[...]

PR #2798, opened 2022-11-11
Standardize HTTPResponse.read(X) behavior regardless of compression 

### LINKED ISSUE #709: .read should only return empty (false) when there is nothing left to read (EOS)
I have some code that does a partial read of a "raw" response in requests to look for "magic bytes".  Digging into the urllib3 code I find that the size you pass in is used as the pre-decode amount which means that the decompressor doesn't have enough bytes to give you anything.  This confused my code into thinking it had read all of the response, which it hadn't.

Here is a simple example:

In [1]: import urllib3
In [2]: http = urllib3.PoolManager()
In [3]: resp = http.request('GET', 'http://httpbin.org/gzip', preload_content=False)
In [4]: resp.read(8)
Out[4]: b''

I would argue that the amount passed into .read (8 in this case) should be the amount _after_ decoding.

[EDIT]
I originally did not include the `preload_content=False`.  This is necessary.  Without it the content is (correctly) loaded into `.data`.
##

### LINKED ISSUE #2128: Standardize HTTPResponse.read(X) behavior regardless of compression

pquentin:
Using the findings from https://github.com/urllib3/urllib3/issues/2787, I reimplemented #2712 to only do one copy at worst, using 4GiB of memory for a 2GiB read. This is much better than #2712 which was doing three copies at worst, and two copies when using memory view.

More importantly, this is equivalent to the current solution where we also have one copy at worst when calling `self._decode()`! So the blockers from #2718 are gone.

Closes #709 (given the decision made in #2769), Closes #2712, Closes #2128

Open questions:

- [x] https://github.com/urllib3/urllib3/issues/2799
- [x] https://github.com/urllib3/urllib3/issues/2800
- [x] should we avoid using a buffer when the content is not encoded?
- [ ] should we try to do bigger read() at the beginning if amt is really small? or save it for later as this isn't a regression


src/urllib3/response.py
```diff
@@ -223,6 +225,62 @@ def _get_decoder(mode: str) -> ContentDecoder:
...
+            raise ValueError("buffer is empty")
+        elif n < 0:
+            raise ValueError("n should be > 0")
+
+        fetched = 0
+        ret = io.BytesIO()
+        while fetched < n:
+            remaining = n - fetched
+            chunk = self.buffer.popleft()
+            if remaining < len(chunk):
```
sethmlarson, line 267: len(chunk) is calculated 3 times, we can cache this value.

src/urllib3/response.py
```diff
@@ -223,6 +225,62 @@ def _get_decoder(mode: str) -> ContentDecoder:
...
+                left_chunk, right_chunk = chunk[:remaining], chunk[remaining:]
+                ret.write(left_chunk)
+                self.buffer.appendleft(right_chunk)
+                self._size -= len(left_chunk)
```
sethmlarson, line 271: If I'm reading the logic right, len(left_chunk) will always be "remaining", right?

src/urllib3/response.py
```diff
@@ -720,6 +781,48 @@ def _fp_read(self, amt: Optional[int] = None) -> bytes:
...
 
+    def _raw_read(
+        self,
+        amt: Optional[int] = None,
+    ) -> bytes:
+        """
+        Reads `amt` of bytes from the socket.
+        """
+        if self._fp is None:
+            return None  # type: ignore[return-value]
```
sethmlarson, line 792: Should this return `b""`? Otherwise we should type annotate this function properly.

src/urllib3/response.py
```diff
@@ -223,6 +225,62 @@ def _get_decoder(mode: str) -> ContentDecoder:
...
+    def __len__(self) -> int:
+        return self._size
+
+    def put(self, data: bytes) -> None:
+        self.buffer.append(data)
+        self._size += len(data)
+
+    def get(self, n: int) -> bytes:
+        if not self.buffer:
+            raise ValueError("buffer is empty")
```
sethmlarson, line 258: nit: Should this be a `RuntimeError` instead of `ValueError`?

src/urllib3/response.py
```diff
@@ -750,51 +853,40 @@ def read(
...
-            if self.length_remaining is not None:
-                self.length_remaining -= len(data)
+        flush_decoder = False
+        if amt is None:
+            flush_decoder = True
+        elif amt != 0 and not data:
+            flush_decoder = True
 
-            data = self._decode(data, decode_content, flush_decoder)
+        if not data and len(self._decoded_buffer) == 0:
```
sethmlarson, line 870: This is one of the locations you're referencing for https://github.com/urllib3/urllib3/issues/2799 ?

test/test_response.py
```diff
@@ -24,13 +25,64 @@
...
+    @pytest.mark.skipif(
+        sys.version_info < (3, 8), reason="pytest-memray requires Python 3.8+"
+    )
+    @pytest.mark.limit_memory("12.5 MB")  # assert that we're not doubling memory usage
+    def test_memory_usage(self) -> None:
+        # Allocate 10 1MiB chunks
+        buffer = BytesQueueBuffer()
+        for i in range(10):
+            # This allocates 2MiB, putting the max at around 12MiB. Not sure why.
+            buffer.put(bytes(2**20))
```
sethmlarson, line 81: Doesn't this allocate 1MB (2**20 = 1MB) making the max of 12MB make sense?

sethmlarson (dismissed):
This looks great, only had a few nits and two question then we can ship! :shipit:

src/urllib3/response.py
pquentin: Yes. More precisely, #2799 is an issue since urllib3 1.11, and this is a new place that is wrong and will need to be updated if we decide to fix #2799.

test/test_response.py
pquentin: memray tells me that it allocates 2MiB, and the whole test uses 12MiB. I would have expected 11MiB instead.

src/urllib3/response.py
pquentin: Addressed in [`b24d1da` (#2798)](https://github.com/urllib3/urllib3/pull/2798/commits/b24d1da111f4a5b0f24ff361ec980c616ba864d9)

src/urllib3/response.py
pquentin: I usually don't add micro optimizations without benchmarking, but this time I was lazy: [`b24d1da` (#2798)](https://github.com/urllib3/urllib3/pull/2798/commits/b24d1da111f4a5b0f24ff361ec980c616ba864d9)

src/urllib3/response.py
pquentin: Addressed in [`0dc26a3` (#2798)](https://github.com/urllib3/urllib3/pull/2798/commits/0dc26a352d3178f8d66784443f79ed65d18b60f6).

pquentin:
My implementation of #2800 breaks the requests test suite, not sure why yet.

sethmlarson:
@pquentin I wonder if we need to change the call to read() with decode_content=False inside of the "drain" method on HTTPResponse? I know Requests allows for raw data access.

src/urllib3/response.py
pquentin: This pull request does not change anything here, we should address this independently imo.

sethmlarson (dismissed):
This looks excellent, let's merge!

sethmlarson (approved):

PR #2801, opened 2022-11-12
3.11 is final

graingert:


graingert:
oh it's already done here https://github.com/urllib3/urllib3/pull/2760/files

Issue #2802, opened 2022-11-13
`util.parse_url` allows invalid characters in reg-names


kenballus:
### `util.parse_url` allows invalid characters in reg-names.

RFC3986 defines reg-name as follows:
`reg-name = *( unreserved / pct-encoded / sub-delims )`
`unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"`
`sub-delims = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="`
`pct-encoded = "%" HEXDIG HEXDIG`

Currently, `urllib3.util.parse_url` accepts reg-names containing characters outside those enumerated above.
The incorrectly accepted ASCII characters are `[\x00-\x20]`, `"`, `<`, `>`, `^`, `` ` ``, `{`, `|`, `}`, and `\x7f`.

### Environment
```python3
>>> print("OS", platform.platform())
OS Linux-6.0.7-arch1-1-x86_64-with-glibc2.36
>>> print("Python", platform.python_version())
urllib3 2.0.0.dev0
>>> print("urllib3", urllib3.__version__)
Python 3.10.8
```

### Steps to Reproduce

```python3
>>> import urllib3
>>> urllib3.util.parse_url("http://|") # Whatever happens, "|" should not end up in a reg-name.
Url(scheme='http', auth=None, host='|', port=None, path=None, query=None, fragment=None)
>>> # But it does!
```

sigmavirus24:
3986 isn't the end of the line. WHAT WG supports Unicode ignoring 3987

kenballus:
WHATWG also disallows many of these characters. Quoting the WHATWG spec:

"If input contains a [forbidden host code point](https://url.spec.whatwg.org/#forbidden-host-code-point), [validation error](https://url.spec.whatwg.org/#validation-error), return failure."
"A forbidden host code point is U+0000 NULL, U+0009 TAB, U+000A LF, U+000D CR, U+0020 SPACE, U+0023 (#), U+002F (/), U+003A (:), U+003C (<), U+003E (>), U+003F (?), U+0040 (@), U+005B ([), U+005C (\), U+005D (]), U+005E (^), or U+007C (|)."
(If you see any of these characters, the parse should fail)

"If input contains a [forbidden host code point](https://url.spec.whatwg.org/#forbidden-host-code-point), [validation error](https://url.spec.whatwg.org/#validation-error), return failure."
[...]

PR #2803, opened 2022-11-13
Change reg-name pattern to match the RFC

kenballus:
(Fixes [#2802](https://github.com/urllib3/urllib3/issues/2802))
This patch makes it so that _REG_NAME_PAT matches RFC3986.
This fixes a bug, but it also introduces incompatibilities with Unicode.
I'd be happy to make this patch work with Unicode, but I wasn't sure if that's within the scope of this module.

Please let me know if I should add Unicode support, and then I'll add tests, autoformatting, etc.


sethmlarson:
@kenballus We do accept Unicode in URLs and encode it properly for users.

Issue #2804, opened 2022-11-13
util.parse_url improperly rejects port numbers with more than 5 digits


kenballus:
### `util.parse_url` improperly rejects port numbers with more than 5 digits
The RFC 3986 and 3987 definition of a port number:
`port = *DIGIT`

The WHATWG definition of a port number:
"""
A URL-port string must be one of the following:
- the empty string
- one or more [ASCII digits](https://infra.spec.whatwg.org/#ascii-digit) representing a decimal number no greater than 2^16 − 1.

"""

`urllib3`'s URL parser improperly rejects port numbers that are within the unsigned 16-bit integer range, but consist of more than 5 digits because of leading `0` characters.

### Environment
```python
>>> print("OS", platform.platform())
OS Linux-6.0.7-arch1-1-x86_64-with-glibc2.36
>>> print("Python", platform.python_version())
Python 3.10.8
>>> print("urllib3", urllib3.__version__)
urllib3 2.0.0.dev0
```

### Steps to Reproduce
```python3
>>> import urllib3
>>> urllib3.util.parse_url("http://example.com:000001") # Shouldn't be a problem.
Traceback (most recent call last):
  File "<REDACTED>/urllib3/util/url.py", line 425, in parse_url
    host, port = _HOST_PORT_RE.match(host_port).groups()  # type: ignore[union-attr]
AttributeError: 'NoneType' object has no attribute 'groups'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<REDACTED>/urllib3/util/url.py", line 451, in parse_url
    raise LocationParseError(source_url) from e
urllib3.exceptions.LocationParseError: Failed to parse: http://example.com:000001
>>> # But it was!
```

I've tested out 7 or 8 other URL parsers, and none of them have this behavior.

PR #2806, opened 2022-11-14
Strip leading zeros from ports

venthur:
fixed parsing for ports with more than 5 ascii digits (i.e. lots of leading zeros). I've changed the regexp to strip leading zeros from the port in a non-capturing group.

Also added a test.

closes: #2804


src/urllib3/util/url.py
```diff
@@ -61,7 +61,7 @@
 _BRACELESS_IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT[2:-2] + "$")
 _ZONE_ID_RE = re.compile("(" + _ZONE_ID_PAT + r")\]$")
 
-_HOST_PORT_PAT = ("^(%s|%s|%s)(?::([0-9]{0,5}))?$") % (
+_HOST_PORT_PAT = ("^(%s|%s|%s)(?::([0-9]*))?$") % (
```
Ousret, line 64: Bit of a warning about this. We have to restrict the size.

```
int("0"*10000 + "1")
ValueError: Exceeds the limit (4300) for integer string conversion: value has 10001 digits
```

Ousret: Maybe simply removing leading zeros would do it. Nonetheless, I am not sure about what is reasonable, len-wise. <4300 for sure.

venthur: I see, I think we can fix the regex to catch any leading zeros in the port in a non-capturing group, discarding them.

sethmlarson (approved):
LGTM

pquentin:
This pull requested broke the test suite of elastic-transport-python: https://github.com/elastic/elastic-transport-python/pull/97.

Old behavior:

```
>>> from urllib3.util import parse_url
>>> parse_url("https://[::1]:0")
Url(scheme='https', auth=None, host='[::1]', port=0, path=None, query=None, fragment=None)
```

New behavior:

```
>>> from urllib3.util import parse_url
>>> parse_url("https://[::1]:0")
Url(scheme='https', auth=None, host='[::1]', port=None, path=None, query=None, fragment=None)
>>> 
```

The port 0 is reserved but valid, I don't think we should return None? If we consider the interpretation from #2806 valid, we should make that zero part of the group. The `rfc3986` library does returns None, but `urllib.parse.urlparse` from the standard library returns 0, which seems more correct.

@sigmavirus24 Any opinions here?

venthur:
Interesting problem! The way I read the RFC, leading zeros are permitted in the port part, and there's the "normalization" i.e. the port can be removed when empty (or in sync with the scheme) -- that seems aligned with our current implementation.

On the other hand: I cannot find a mention of a single zero in the port in the RFC. @pquentin you mention it is reserved, can you point me to the part where that is mentioned?

pquentin:
It's not in RFC 3986, but here: https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml. Which I guess is my point, it should not be treated specially.

sethmlarson:
@venthur The wording of "can be removed when empty" means if there's no data within the port section (ie `http://localhost:`) not a zero value. I believe the new behavior of setting the `port` value to `None` is a bug.

venthur:
I'm trying to find some clever regex that removes all zeroes but one, but i stand defeated!

sethmlarson:
@venthur  How about something like: `0*?(|0|[1-9][0-9]{0,4})`

venthur:
thanks @sethmlarson that seems to work, I've created a pr: https://github.com/urllib3/urllib3/pull/2849

PR #2810, opened 2022-11-15
Run tests in verbose mode

pquentin:
We have some hanging tests, the verbose mode will help identifying them.


pquentin:
So it's indeed test_requesting_large_resources_via_ssl that hangs. It has been very useful for #2798, but now that we have memray, I think I'm going to turn it into a unit test. The value it brings is not enough compared to its slowness and flakiness.

sethmlarson (approved):

pquentin:
One problem here is that we have some tests with really verbose parametrizations that we [cannot disable](https://github.com/pytest-dev/pytest/issues/5285):

image

I still think verbosity is a net win but am starting to wonder if it should be CI only.

sethmlarson:
@pquentin Ah yeah, that's a good point. We should probably make this CI-only.

bluetech:
You can set short IDs for these parametrizations: https://docs.pytest.org/en/7.2.x/example/parametrize.html#different-options-for-test-ids

PR #2811, opened 2022-11-15
Only test _fp_read once

pquentin:
This is a slow integration test taking 10 seconds and we were running it six times, for no additional value.


illia-v:
> This is a slow integration test taking 10 seconds and we were running it six times, for no additional value.

There was value because some of httplib, pyOpenSSL, SecureTransport worked differently when `amt` was `None` and non-`None`.

PR #2814, opened 2022-11-16
Restore but deprecate getheaders() and getheader() calls

pquentin:
I tried using urllib3 2.0.0a1 with Rally and this is the first thing that failed.


sethmlarson (approved):
This looks good to me, should we also start raising this warning in 1.26.x and make a release there too then?

pquentin:
Yes, definitely!

PR #2816, opened 2022-11-16
Make downstream integration tests re-usable

### LINKED ISSUE #2778: Make our downstream integration tests re-runnable
Currently our downstream tests both clone a git repository into the venv created by nox. Due to how the setup is written the nox session will fail when run a second time due to the repository already existing. It would be nice to be able to run:

`$ nox -rs downstream_requests` over and over without needing to reclone the repository or manually configure the test setup.

I think checking that the repository exists and if so not applying the git patches and continuing as normal would work okay for this case?

Ousret:
Close #2778

Quote
> I think checking that the repository exists and if so not applying the git patches and continuing as normal would work okay for this case?

I choose to reset, pull and apply the patches every time. If there is an update in those repositories, we want to check that the patch still applies.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -100,7 +100,7 @@ def git_clone(session: nox.Session, git_url: str) -> None:
     if not os.path.isdir(expected_directory):
         session.run("git", "clone", "--depth", "1", git_url, external=True)
     else:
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
LGTM, thank you!

PR #2817, opened 2022-11-17
Prevent issue in HTTPResponse().read() when decoded_content is True, then False

### LINKED ISSUE #2800: Should we prevent `read(decode_content=True)` followed by `read(decode_content=False)`?
`BaseHTTPResponse.read()` has a `decode_content` parameter to let the user choose between... decoding the content or not. (I'm not sure why the word is "decode" because what we're really doing is decompressing. Anyway.)

Until #2712 and #2798, issuing read() calls with differing values of decode_content would be really weird but you would not loose any data. Now that we have a buffer, issuing `decode_content=False` after having issued `decode_content=True` will cause bugs:

 * if you have enough decoded data in the buffer for a read(amt, decode_content=False) you will get decoded data
 * if you don't have enough data, the existing data in the buffer will be silently ignored and lost

I don't think there's an use case here to support, but should we actively prevent this footgun?

### Minimum requirements

- [ ] Calling `HTTPResponse.read(decode_content=True)` **or** `HTTPResponse.read(decode_content=False)` should continue working as before.
- [ ] Calling `HTTPResponse.read(decode_content=True)` **followed by** `HTTPResponse.read(decode_content=False)` should raise a RuntimeError. The other way is fine.
- [ ] The requests and botocore integration tests should continue to pass.

Ousret:
Provided it has initialized an eligible decoder(decompressor) and did decode once.
Close #2800


test/test_response.py
```diff
@@ -564,6 +564,50 @@ def test_io_not_autoclose_textiowrapper(self) -> None:
         with pytest.raises(ValueError, match="I/O operation on closed file.?"):
             next(reader)
 
+    def test_read_with_illegal_mix_decode_toggle(self) -> None:
+        compress = zlib.compressobj(6, zlib.DEFLATED, -zlib.MAX_WBITS)
+        data = compress.compress(b"foo")
+        data += compress.flush()
```
pquentin, line 570: Please use `data = zlib.compress(b"foo")` instead in this test and the one below.

changelog/2800.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a potential data loss scenario when using ``HTTPResponse.read()`` with ``decode_content=True``
+and then subsequently calling the method with ``decode_content=False``.
```
pquentin, line 2: @sethmlarson nit: should we replace "Fixed" with "Prevented"? It's slightly more precise but maybe you only want to use a few verbs in the changelog.

pquentin (approved):
Thanks! This looks great. I commented on details.

test/test_response.py
Ousret: Done. I adapted expectations (read size) to match the initial raw payload struct.

changelog/2800.bugfix.rst
sethmlarson: My convention of only using a few verbs comes from keepachangelog.com. I agree that this wording isn't right: it's a "change" not a fix.

Maybe something like:

```
Changed ``HTTPResponse.read()`` to raise an error when calling with ``decode_content=False`` after using ``decode_content=True`` to prevent data loss.
```

pquentin: Sounds great!

pquentin: And now it looks great:

Capture d’écran 2022-11-20 à 21 43 26

pquentin (approved):

sethmlarson (approved):

sethmlarson:
@Ousret  There is a bounty of the linked issue, you can claim it from our OpenCollective.

Issue #2815, opened 2022-11-16
Tests randomly fail in pipeline due to timeout


venthur:
### Subject

The heavy lifting of the tests is currently limited to a 20mins timeout. (https://github.com/urllib3/urllib3/blob/main/.github/workflows/ci.yml#L88) -- however, the actual execution time on the CI pipeline for the **same** testsuite (i.e. os-python-combo), can vary from 5mins to >20mins. E.g. often, one of the tests in the test matrix fails due to the timeout and restarting the same test (sometimes several times) solves the problem. I found that the same test that hit the 20mins timeout can finish in ~5mins on the next run.

### Environment

gh-actions

### Steps to Reproduce

pull request :)

I propose to either increase the timeout to 30mins or finding the individual tests that take so long and fix them somehow.

PR #2818, opened 2022-11-17
increased test-timeout for the heavy gh-actions from 20 to 30mins.

venthur:
closes: #2815


pquentin (changes requested):
So I see how this makes sense and could make our tests pass. But I'd like to only do this as a temporary measure until we turn `test_requesting_large_resources_via_ssl` into an unit test. We currently spend 6 minutes there.

Can we at least print the 10 slowest tests? https://docs.pytest.org/en/6.2.x/usage.html#profiling-test-execution-duration

venthur:
> Can we at least print the 10 slowest tests? https://docs.pytest.org/en/6.2.x/usage.html#profiling-test-execution-duration

Good point! I wasn't even sure which of them suddenly took so long. I've added the appropriate option.

pquentin:
Thanks! So maybe we should use adopts more but right now all our options are in noxfile.py, can you please use that instead?

pquentin (approved):

PR #2822, opened 2022-11-20
Stop showing thousands of @ in test output

pquentin:
As pointed out by @bluetech in https://github.com/urllib3/urllib3/pull/2810#issuecomment-1320334670.


sethmlarson (approved):
Nice! I always forget that this is an option, thanks Quentin.

PR #2823, opened 2022-11-22
[1.26] Strip leading zeros from ports

sethmlarson:
Backports https://github.com/urllib3/urllib3/pull/2806

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -347,6 +347,10 @@ class TestUtil(object):
         with pytest.raises(LocationParseError):
             parse_url("https://www.google.com:-80/")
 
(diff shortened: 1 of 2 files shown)


test/test_util.py
```diff
@@ -347,6 +347,10 @@ def test_parse_url_negative_port(self):
         with pytest.raises(LocationParseError):
             parse_url("https://www.google.com:-80/")
 
+    def test_parse_url_remove_leading_zeros(self) -> None:
```
hramezani, line 350:
```suggestion
    def test_parse_url_remove_leading_zeros(self):
```

hramezani (approved):

PR #2829, opened 2022-11-23
Require pyOpenSSL 17.1.0+ to support cryptography 39+

pquentin:
Looking at the build in https://github.com/urllib3/urllib3/pull/2827, I noticed the following warning:

> /home/docs/checkouts/readthedocs.org/user_builds/urllib3/envs/2827/lib/python3.7/site-packages/cryptography/hazmat/backends/openssl/x509.py:17: CryptographyDeprecationWarning: This version of cryptography contains a temporary pyOpenSSL fallback path. Upgrade pyOpenSSL now.
  utils.DeprecatedIn35,

This is because our pyOpenSSL backend does `from cryptography.hazmat.backends.openssl.x509 import _Certificate`. It will be gone in the next version of cryptography. And people could blame pyOpenSSL, when we are the one to blame. Since this was only in a fallback, I bumped the minimum pyOpenSSL version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ zstd = [
 ]
 secure = [
   "pyOpenSSL>=17.1.0",
-  "cryptography>=1.3.4",
+  "cryptography>=1.9",
   "idna>=2.0.0",
   "certifi",
   "urllib3-secure-extra",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ zstd = [
   "zstandard>=0.18.0",
 ]
 secure = [
-  "pyOpenSSL>=0.14",
+  "pyOpenSSL>=17.1.0",
   "cryptography>=1.3.4",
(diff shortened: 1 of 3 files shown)


sethmlarson (dismissed):

pyproject.toml
```diff
@@ -46,7 +46,7 @@ zstd = [
   "zstandard>=0.18.0",
 ]
 secure = [
-  "pyOpenSSL>=0.14",
+  "pyOpenSSL>=17.1.0",
   "cryptography>=1.3.4",
```
illia-v, line 50: What do you think about making this in line with what [pyOpenSSL requires](https://github.com/pyca/pyopenssl/blob/17.1.0/setup.py#L99) to prevent any resolving issues? 

```suggestion
  "cryptography>=1.9",
```

pquentin: Thanks, good idea. We will have to update the changelog too

sethmlarson (approved):

PR #2836, opened 2022-11-28
Add 1.26.13 release to CHANGES.rst on main

sethmlarson:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -117,6 +117,15 @@ Read the `v2.0 migration guide <https://urllib3.readthedocs.io/en/latest/v2-migr
 * Fixed a reference cycle bug in ``urllib3.util.connection.create_connection()`` (`#2277 <https://github.com/urllib3/urllib3/issues/2277>`__).
 * Fixed a socket leak if ``HTTPConnection.connect()`` fails (`#2571 <https://github.com/urllib3/urllib3/pull/2571>`__).
 
+1.26.13 (2022-11-23)
+--------------------
+
+* Deprecated the ``HTTPResponse.getheaders()`` and ``HTTPResponse.getheader()`` methods.
+* Fixed an issue where parsing a URL with leading zeroes in the port would be rejected
+  even when the port number after removing the zeroes was valid.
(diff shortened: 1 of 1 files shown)


hramezani (approved):

CHANGES.rst
```diff
@@ -117,6 +117,15 @@ Read the `v2.0 migration guide <https://urllib3.readthedocs.io/en/latest/v2-migr
 * Fixed a reference cycle bug in ``urllib3.util.connection.create_connection()`` (`#2277 <https://github.com/urllib3/urllib3/issues/2277>`__).
 * Fixed a socket leak if ``HTTPConnection.connect()`` fails (`#2571 <https://github.com/urllib3/urllib3/pull/2571>`__).
 
+1.26.13 (2022-11-23)
+--------------------
+
+* Deprecated the ``HTTPResponse.getheaders()`` and ``HTTPResponse.getheader()`` methods.
```
pquentin, line 123: We already had a changelog entry for this, should we remove it? Also this entry does not mention that 2.1.0 will remove this.

CHANGES.rst
```diff
@@ -117,6 +117,15 @@ Read the `v2.0 migration guide <https://urllib3.readthedocs.io/en/latest/v2-migr
...
+* Fixed an issue where parsing a URL with leading zeroes in the port would be rejected
+  even when the port number after removing the zeroes was valid.
+* Fixed a deprecation warning when using cryptography v39.0.0.
```
pquentin, line 126: This was fixed differently in main, with its own changelog entry. So in that case I think we should delete this line.

CHANGES.rst
illia-v: I think we should keep porting 1.26 changelogs as they are, so that anyone can track changes in stable versions from the main branch.

Having duplicates in 1.26.x and non-final 2.0 sections should be fine as long as we remember to remove them from the final 2.0.0 section when it is released.

pquentin: Yeah I was talking about removing the changelog entry in the 2.0 section.

PR #2837, opened 2022-11-28
Fix typos in warnings

kurtmckee:
Found while reviewing CI errors in another project.

I see that the changelog check is failing, please let me know if this is significant enough to warrant a changelog entry.


sethmlarson (approved):
Thanks for this! This change will need to be backported to the 1.26.x branch as well, can you open a PR there as well?

kurtmckee:
Yep, can do!

PR #2841, opened 2022-11-30
disable Nagle's algorithm in test_ssl_read_timeout

graingert:
it's possible that the `ssl_sock.send(` doesn't send enough data to satisfy Nagle's algorithm causing the kernel to wait over `LONG_TIMEOUT = 0.01` seconds before sending any data which causes read_headers to fail with a read timeout

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1267,6 +1267,8 @@ class TestSSL(SocketDummyServerTestCase):
 
         def socket_handler(listener: socket.socket) -> None:
             sock = listener.accept()[0]
+            # disable Nagle's algorithm so there's no delay in sending a partial body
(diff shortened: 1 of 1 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1267,6 +1267,7 @@ def test_ssl_read_timeout(self) -> None:
 
         def socket_handler(listener: socket.socket) -> None:
             sock = listener.accept()[0]
+            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, True)
```
graingert, line 1270:
```suggestion
            # disable Nagle's algorithm so there's no delay in sending a partial body 
            sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, True)
```

Issue #2757, opened 2022-10-28
urllib3 does not send http.client.connect audit events


apmorton:
### Subject

Despite being implemented in terms of `http.client`, urllib3 doesn't cause `http.client.connect` audit events to be emitted

### Steps to Reproduce

```python
import http.client
import sys

import urllib3


def _hook(event: str, args: tuple):
    if event == "http.client.connect":
        print(event, args)
    if event == 'socket.connect':
        print(event, args)


sys.addaudithook(_hook)

print('using http.client')
http.client.HTTPConnection('www.python.org').connect()
print()
print('using urllib3')
http = urllib3.PoolManager()
resp = http.request("GET", "https://www.python.org")
```

```
using http.client
http.client.connect (<http.client.HTTPConnection object at 0x7f1303983fd0>, 'www.python.org', 80)
socket.connect (<socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 0)>, ('199.232.36.223', 80))

using urllib3
socket.connect (<socket.socket fd=4, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 0)>, ('199.232.36.223', 443))
```

This is because the urllib3 `HTTPConnection` object overrides `connect`, which is where the audit events are sent from.
https://github.com/python/cpython/blob/e87ada48a9e5d9d03f9759138869216df0d7383a/Lib/http/client.py#L954

### Expected Behavior

`http.client.connect` audit events should be emitted

### Actual Behavior

They aren't

sigmavirus24:
If we don't use http.client but try to provide a similar API not a fully complete implementation, why must we do what only standard libraries do?

apmorton:
My expectation that urllib3 would send this audit event is this sentence in the API docs:

> Based on http.client.HTTPConnection

To me, this implies that it is a superset of the functionality present in http.client.HTTPConnection.

Regardless of implementation details, it would be useful for urllib3 to send `http.client.connect` audit events - or less preferably, a different audit event.

sigmavirus24:
> My expectation that urllib3 would send this audit event is this sentence in the API docs:
> 
> > Based on http.client.HTTPConnection
> 
> To me, this implies that it is a superset of the functionality present in http.client.HTTPConnection.
> 
[...]

PR #2847, opened 2022-12-05
Added support for sending http.client.connect audit events

ravi-kale:
Fixes issue --> https://github.com/urllib3/urllib3/issues/2757

PR #2848, opened 2022-12-06
added http client connect audit event support

ravi-kale:
Fixes --> https://github.com/urllib3/urllib3/issues/2757

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -5,6 +5,7 @@ import logging
 import os
 import re
 import socket
+import sys
 import warnings
 from socket import error as SocketError
 from socket import timeout as SocketTimeout
@@ -169,6 +170,9 @@ class HTTPConnection(_HTTPConnection, object):
 
         if self.socket_options:
             extra_kw["socket_options"] = self.socket_options
+
+        if hasattr(sys, "audit"):
+            sys.audit("http.client.connect", self, self.host, self.port)
 
         try:
             conn = connection.create_connection(


src/urllib3/connection.py
```diff
@@ -170,6 +171,9 @@ def _new_conn(self):
         if self.socket_options:
             extra_kw["socket_options"] = self.socket_options
 
+        if hasattr(sys, "audit"):
```
sethmlarson, line 174: This `hasattr` check gets called per connection, can we detect the feature once in the global scope and cache whether `sys.audit` is available? Let's also add a comment that mentions sys.audit being Python 3.8+

sethmlarson (changes requested):
Thanks for this. Can we also add a test case that verifies the names/values of all the `sys.audit` calls for the `http` namespace?

We also want this implemented in the `main` branch, the `1.26.x` branch is no longer receiving new features.

src/urllib3/connection.py
ravi-kale: Yeah, thanks for your feedback.
Will continue working on https://github.com/urllib3/urllib3/pull/2847. Will close this Pull Request.

ravi-kale:
> Thanks for this. Can we also add a test case that verifies the names/values of all the `sys.audit` calls for the `http` namespace?
> 
> We also want this implemented in the `main` branch, the `1.26.x` branch is no longer receiving new features.

Hey @sethmlarson, If the `1.26.x` branch is no longer receiving a new feature, I will continue to work on my previous pull request (https://github.com/urllib3/urllib3/pull/2847)  which was closed by me.

PR #2849, opened 2022-12-09
fixed issue with port "0" returning None

venthur:
since merging https://github.com/urllib3/urllib3/pull/2806 -- we report port "0" as None, although it should be "0". 

this pr fixes it and adds two tests for this case. thanks @sethmlarson for the magic regex :)

closes: #2850

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -63,7 +63,7 @@ _IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT + "$")
 _BRACELESS_IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT[2:-2] + "$")
 _ZONE_ID_RE = re.compile("(" + _ZONE_ID_PAT + r")\]$")
 
-_HOST_PORT_PAT = ("^(%s|%s|%s)(?::0*?(|0|[1-9][0-9]{0,5}))?$") % (
+_HOST_PORT_PAT = ("^(%s|%s|%s)(?::0*?(|0|[1-9][0-9]{0,4}))?$") % (
     _REG_NAME_PAT,
     _IPV4_PAT,
     _IPV6_ADDRZ_PAT,
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -63,7 +63,7 @@ _IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT + "$")
 _BRACELESS_IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT[2:-2] + "$")
 _ZONE_ID_RE = re.compile("(" + _ZONE_ID_PAT + r")\]$")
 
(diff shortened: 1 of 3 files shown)


pquentin:
Could this lead to some mild catastrophic backtracking as we have various ways to interpret `00000`?

venthur:
@pquentin the PR adds two tests, covering a single 0 and several ones:

1. xxx:0 -- the boring case -> 0
2. xxx:00000000 -- your case? -> 0

is that what you meant?

pquentin:
No, I know the code is correct. I just want to make sure that it's not vulnerable to https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS

sethmlarson:
I tried https://devina.io/redos-checker and the regex was classified as linear so I think we're good?

pquentin (approved):
Thanks! Looks good to me

src/urllib3/util/url.py
```diff
@@ -63,7 +63,7 @@
 _BRACELESS_IPV6_ADDRZ_RE = re.compile("^" + _IPV6_ADDRZ_PAT[2:-2] + "$")
 _ZONE_ID_RE = re.compile("(" + _ZONE_ID_PAT + r")\]$")
 
-_HOST_PORT_PAT = ("^(%s|%s|%s)(?::0*([0-9]{0,5}))?$") % (
+_HOST_PORT_PAT = ("^(%s|%s|%s)(?::0*?(|0|[1-9][0-9]{0,4}))?$") % (
```
jayaddison, line 66: Sorry @sethmlarson @venthur, this is a nitpick question, but: would it make sense to refactor this to:

```python
_HOST_PORT_PAT = ("^(%s|%s|%s)(?::0*?([0-9]{0,5}))?$") % (
```

(it feels like the port match section of the regex in particular is carefully avoiding a few non-obvious pitfalls -- and any changes that improve readability and distance-from-future-mistakes could be worthwhile, I think)

venthur: I think this was also one of my proposals in the past and we used the more complicated version we have now to cover the case where the port is a single zero.

jayaddison: It does (surprisingly) pass the unit tests (including the single-zero case).

The match behaviour of `0*?` seems to be the reason.  I'm guessing somewhat, but it looks like the `?` causes the repeat-greedy-zero-match to be skipped during the first string  examination pass -- allowing the max-five-digits to match the [single-zero-port case](https://github.com/urllib3/urllib3/blob/dacd84d01f3037d30adf387e2e1bfc95f55dedfe/test/test_util.py#L356) -- but it is permitted during backtracking -- allowing matches for the [more-than-five-zeros-port case](https://github.com/urllib3/urllib3/blob/dacd84d01f3037d30adf387e2e1bfc95f55dedfe/test/test_util.py#L359).

Maybe the more complicated (but more explicitly-defined) pattern is better, because that behaviour is non-obvious.

Something about the empty-string case match (first matching choice in `(|0|[1-9][0-9]{0,4})`) seems non-ideal to me, but I can't formulate a good reason why yet.

PR #2859, opened 2022-12-23
Audit http.client.connect events

### LINKED ISSUE #2757: urllib3 does not send http.client.connect audit events

CCLDArjun:
Trying to resolve #2757


src/urllib3/connection.py
```diff
@@ -76,6 +77,11 @@ class BaseSSLError(BaseException):  # type: ignore[no-redef]
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 
+_SYSAUDIT = False
```
sethmlarson, line 80: Simplify this to `_HAS_SYS_AUDIT = hasattr(sys, "audit")`

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
     assert isinstance(response, HTTPResponse)
 
 
+hook_reported = False
+
+
+def test_audit_event(pool: HTTPConnectionPool) -> None:
+    # Audit hooks are only available in versions >= 3.8
+    if not hasattr(sys, "audit"):
```
sethmlarson, line 42: We should be skipping this test if there's no `sys.audit`, you can use `pytest.mark.skipif` for this.

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+        assert True
+        return
+
+    def _hook(event: str, _: tuple[typing.Any]) -> None:
```
sethmlarson, line 46: nit: This doesn't need to be private.

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+        global hook_reported
```
sethmlarson, line 47: This should be `nonlocal` instead of `global`

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+        if event == "http.client.connect":
+            hook_reported = True
```
sethmlarson, line 49: Let's capture the event data and make an assertation on the values below as well.

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+
+    sys.addaudithook(_hook)  # type: ignore
+
+    conn = pool._get_conn()
+
+    method = "GET"
```
sethmlarson, line 55: nit: No need to store `method` and `path`, these can be inlined into `conn.request(...)`.

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+    path = "/"
+
+    conn.request(method, path)
+
+    assert hook_reported
```
sethmlarson, line 60: Let's augment this assertation to also assert the event data `(conn, host, port)`

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,32 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
     assert isinstance(response, HTTPResponse)
 
 
+hook_reported = False
```
sethmlarson, line 37: This shouldn't be a global, this variable can be a local within the test case.

changelog/2757.bugfix.rst
```diff
@@ -0,0 +1 @@
+Send http.client.connect audit events
```
sethmlarson, line 1: We use keepachangelog.com style changelogs, so let's start this newsfragment with "Fixed", ie:

> Fixed `urllib3.connection.HTTPConnection` to raise the `http.client.connect` audit event to have the same behavior as the standard library HTTP client".

sethmlarson (changes requested):
Thanks for this, I have some comments for you:

ravi-kale:
Duplicate of https://github.com/urllib3/urllib3/issues/2757

ravi-kale:
Duplicate of https://github.com/urllib3/urllib3/pull/2847

sethmlarson (dismissed):
Thanks for this!

test/with_dummyserver/test_connection.py
```diff
@@ -33,6 +34,29 @@ def test_returns_urllib3_HTTPResponse(pool: HTTPConnectionPool) -> None:
...
+    sys.addaudithook(hook)  # type: ignore
+
+    conn = pool._get_conn()
+    conn.request("GET", "/")
+
+    assert audit_event == "http.client.connect"
+    assert len(audit_args) > 2
+    assert audit_args[0] == conn
+    assert audit_args[1] == conn.host
+    assert audit_args[2] == conn.port
```
illia-v, line 57: It's a pity that the audit hook cannot be temporary in this case. Changing the test to check that the `audit` function is called instead of adding a hook accessing nonlocal variables from all next tests seems to fix the segmentation faults.

```suggestion
@pytest.mark.skipif(not hasattr(sys, "audit"), reason="requires python 3.8")
@patch("urllib3.connection.sys.audit")
def test_audit_event(audit_mock: Mock, pool: HTTPConnectionPool) -> None:
    conn = pool._get_conn()
    conn.request("GET", "/")
    audit_mock.assert_any_call("http.client.connect", conn, conn.host, conn.port)
    # Ensure the event is raised only once.
    connect_events = [
        call for call in audit_mock.mock_calls if call.args[0] == "http.client.connect"
    ]
    assert len(connect_events) == 1
```

illia-v:
@CCLDArjun please add `from unittest.mock import Mock, patch` to the test file

illia-v (dismissed):

src/urllib3/connection.py
```diff
@@ -216,6 +219,10 @@ def _new_conn(self) -> socket.socket:
                 self, f"Failed to establish a new connection: {e}"
             ) from e
 
+        # Audit hooks are only available in versions >= 3.8
```
pquentin, line 222: Please replace "versions >= 3.8" with "Python 3.8+", that way when we drop support for 3.8 it will be easy to grep.

changelog/2757.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed urllib3.connection.HTTPConnection to raise the http.client.connect audit event to have the same behavior as the standard library HTTP client
```
pquentin, line 1: Can you please add single backticks around urllib3.connection.HTTPConnection and http.client.connect? I think it should make links to the relevant documentation

pquentin (approved):

illia-v:
I've pushed the suggested changes so that this can be included in #3084 quicker 🙂

illia-v (approved):

changelog/2757.bugfix.rst
pquentin: Sorry I meant something like this as I had missed that `http.client.connect` was not an Intersphinx target:

```reStructuredText
Fixed :class:`~urllib3.connection.HTTPConnection` to raise the ``http.client.connect`` audit event to have the same behavior as the standard library HTTP client (`#2757 <https://github.com/urllib3/urllib3/issues/2757>`__)
```

But the version that got merged is fine too.

PR #2860, opened 2022-12-23
Send a clearer error message if response is truncated before a chunk

rnewson:
If a chunked response is truncated exactly before sending a new chunk
(that is, before the first byte of the chunk size), urllib3 throws
`InvalidChunkLength(got length b'\\n', 0 bytes read)`. This is
confusing as no chunk length was sent (a valid chunk length line would
also be \r\n-terminated and we didn't see those either).

If the chunk length is the empty string, send a different error message.


skydiver1980 (approved):

changelog/2860.bugfix.rst
```diff
@@ -0,0 +1 @@
+Don't claim 'invalid chunk length' if response terminates before the chunk length is sent.
```
ecerulm, line 1:
```suggestion
Use ProtocolError instead of InvalidChunkLength if response terminates before the chunk length is sent.
```

ecerulm (dismissed):
It LGTM, @pquentin , @SethMichaelLarson  any reason not to merge this?

pquentin:
No specific reason, this sounds sensible indeed.

sethmlarson (approved):

sparrowt:
Thanks for this! I was also looking into this exception and thinking it kind of obscures the real problem.

I noticed however that the final changelog seems to have inverted the explanation and made it wrong - it now says this:
>  Changed ``ProtocolError`` to ``InvalidChunkLength`` ...

whereas `changelog/2860.bugfix.rst` added in this PR correctly said:
> Use ProtocolError instead of InvalidChunkLength

Should I raise a new PR to fix the changelog? Presumably _that_ PR shouldn't be in the changelog to avoid some kind of inception which would cause GitHub to set on fire 😉

sethmlarson:
@sparrowt Sounds good, please submit a PR.

beszabo1:
Could this change be added to the 1.26.X version as well?

PR #2865, opened 2023-01-01
Fix the type hint of PoolKey.key_retries

zhanpon:
Thanks for maintaining this great project! Here is a simple patch to fix some type hints.

`key_retries` also accepts `bool` values: https://github.com/urllib3/urllib3/blob/6cfdca5825e3557f66fd00d0c2f4407c3e1e0d1e/src/urllib3/util/retry.py#L261

### DIFF
--- a/changelog/2865.bugfix.rst
+++ b/changelog/2865.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the type hint of ``PoolKey.key_retries`` by adding ``bool`` to the union.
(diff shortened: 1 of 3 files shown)


sethmlarson (approved):
Thanks for fixing this issue! LGTM.

If you're interested and have more time, we have a set of "[Contributor Friendly](https://github.com/urllib3/urllib3/issues?q=is%3Aissue+is%3Aopen+label%3A%22Contributor+Friendly+%E2%99%A5%22)" issues to also consider contributing to.

PR #2866, opened 2023-01-04
Explicitly chain exceptions in HTTPConnectionPool._make_request

### LINKED ISSUE #2313: [v2] Chaining exceptions

zhanpon:
Fix #2313. I didn't search other places for the issue, so there might be other `raise` statements that can be improved as well.

I skipped a changelog entry to be consistent with #2340.


skydiver1980 (approved):

src/urllib3/connectionpool.py
```diff
@@ -489,7 +489,7 @@ def _make_request(
                 new_e, (OSError, NewConnectionError, TimeoutError, SSLError)
             ) and (conn and conn.proxy and not conn.has_connected_to_proxy):
                 new_e = _wrap_proxy_error(new_e, conn.proxy.scheme)
-            raise new_e
+            raise new_e from e
```
ecerulm, line 492: `_wrap_proxy_error` does already explicitly set the `__cause__` and `raise new_e from e`  will overwrite that `__cause__`

so in the case of e being a `CertificateError` the current code can produce ProxyError -> SSLError -> CertificateError but with this change it will produce ProxyError -> CertificateError instead. 

So I think in this case if we really would like to get the "The above exception was the direct cause of the following exception:" instead of "During handling of the above exception, another exception occurred:"  would be better to set `new_e.__cause__` explicitly after the `SSLError`.


Reference https://docs.python.org/3/library/exceptions.html#exception-context

ecerulm (changes requested):
`raise new_e from e` is not exactly equivalent to the current `raise new_e` when e is a `BaseSSLError` or a `CertificateError`.

ecerulm:
@zhanpon, I'm closing this issue as per my other comment, feel free to reopen it you disagree.

PR #2869, opened 2023-01-06
[1.26] Fix socket timeout value when HTTPConnection is reused

nickwilliams-zaxiom:
This is a backport of PR #2792, for the purposes of applying the fix for issue #2645 to the 1.26.x release stream.

### DIFF
--- a/changelog/2645.bugfix.rst
+++ b/changelog/2645.bugfix.rst
@@ -1,3 +1,0 @@
-Fixed an issue where an ``HTTPConnection`` instance would erroneously reuse the socket
-read timeout value from reading the previous response instead of a newly configured connect timeout.
-Instead now if ``HTTPConnection.timeout`` is updated before sending the next request the new timeout value will be used.
--- a/changelog/2645.bugfix.rst
+++ b/changelog/2645.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed an issue where an ``HTTPConnection`` instance would erroneously reuse the socket
+read timeout value from reading the previous response instead of a newly configured connect timeout.
+Instead now if ``HTTPConnection.timeout`` is updated before sending the next request the new timeout value will be used.
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -229,6 +229,11 @@ class HTTPConnection(_HTTPConnection, object):
             )
 
     def request(self, method, url, body=None, headers=None):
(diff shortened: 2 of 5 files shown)


pquentin:
I merged from master to include #2883 which will fix most CI failures. Sorry for the noise. Not sure about the app engine failure, but note that we dropped app engine in the `main` branch, which explains why it does not fail there.

nickwilliams-zaxiom:
Thanks, @pquentin. That did, indeed, fix everything except `app_engine`. If we no longer support app_engine in `main`, I suggest we ignore that result here. But if I need to accommodate the `app_engine` failure, I think the easiest solution is to add `self.sock = None` to the constructor to mirror the CPython parent class constructor.

src/urllib3/connection.py
```diff
@@ -229,6 +229,11 @@ def putheader(self, header, *values):
             )
 
     def request(self, method, url, body=None, headers=None):
+        # Update the inner socket's timeout value to send the request.
+        # This only triggers if the connection is re-used.
+        if self.sock is not None:
```
sethmlarson, line 234: To support AppEngine and other HTTPConnection implementations we can add `if getattr(self, "sock", None) is not None:`

changelog/2645.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed an issue where an ``HTTPConnection`` instance would erroneously reuse the socket
```
sethmlarson, line 1: We should delete this file since there's no `changelog/` directory on the 1.26.x branch.

sethmlarson (changes requested):
Thanks for opening this, we do need to support App Engine in the 1.26.x stream. See below for my suggestions:

src/urllib3/connection.py
nickwilliams-zaxiom: Done

changelog/2645.bugfix.rst
nickwilliams-zaxiom: Done

nickwilliams-zaxiom:
Hopefully everything passes this time. 🤞

nickwilliams-zaxiom:
Would you look at that. All checks have passed. :-) Just need a 👍  from someone.

sethmlarson (approved):
Thanks much! :shipit: If you're interested in contributing more we've got lots of open issues, take a peek and see if you see one you're interested in :)

PR #2881, opened 2023-01-10
Speed up Python 3.11 CI

pquentin:
This will use a wheel for coverage and include a few performance coverage improvements that help on PyPy and CPython 3.11.

coverage 7.0.0 improved coverage combine but broke the .coveragerc configuration, let's see if I can figure out a config that works.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,16 +3,10 @@ source =
     urllib3
 
 omit =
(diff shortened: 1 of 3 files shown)


sethmlarson (approved):
LGTM, thanks!

PR #2882, opened 2023-01-10
Test brotli with Python 3.10, not 3.11

pquentin:
Python 3.11 runs test extremely slowly for now.

This requires changing the required checks again sorry.


pquentin:
Closing in favor of #2881 where I made Python 3.11 fast, including the brotli tests that were running on 3.11.

PR #2887, opened 2023-01-13
[1.26] Stop relying on distutils from standard library

pquentin:
Python 3.12 no longer includes distutils and we have a recent enough setuptools for this to work, even on Python 2.7.


sethmlarson (approved):
Thanks Quentin!

PR #2889, opened 2023-01-13
Fix reference to private urllib3.util.ssl_ module in custom context example

JustAnotherArchivist:
#2446 moved the `create_urllib3_context` function from the private `urllib3.util.ssl_` to the public `urllib3.util` module, but the code example (added by #2357) was not adjusted to no longer point at the private module.

### DIFF
--- a/docs/advanced-usage.rst
+++ b/docs/advanced-usage.rst
@@ -619,7 +619,7 @@ flag that isn't set by default, and then makes a HTTPS request:
     import ssl
 
     from urllib3 import PoolManager
-    from urllib3.util.ssl_ import create_urllib3_context
+    from urllib3.util import create_urllib3_context
 
     ctx = create_urllib3_context()
     ctx.load_default_certs()


JustAnotherArchivist:
This is a trivial docs change, so I assume a changelog entry isn't necessary.

sigmavirus24:
It's not a private module. It would start with an underscore if it were meant to be private, not end

JustAnotherArchivist:
@sigmavirus24 That's what I thought originally, and the trailing underscore confused me at the time (no need for it since `urllib3.util.ssl` would be perfectly fine without colliding with `ssl`). But then I found several comments about it being considered private anyway, e.g. <https://github.com/urllib3/urllib3/pull/2357#pullrequestreview-757553185> and <https://github.com/urllib3/urllib3/pull/2446#discussion_r725133453>. It's also undocumented, which is practically the same as private API from the user perspective.

pquentin (approved):
Thanks! I like your attention to detail :) Looks good to me.

PR #2892, opened 2023-01-17
Ignore Dependabot updates for patches

sethmlarson:
Dependabot is fairly disruptive to my notifications, so let's ignore patch releases since they're pretty useless to us. Security fixes are always opened by Dependabot so we shouldn't be worried about missing those.


pquentin (approved):
Thanks! LGTM.

Issue #2680, opened 2022-07-15
Deprecate [secure] extra


sethmlarson:
## Are you seeing a DeprecationWarning?

Hello! If you're seeing one of the below messages in `DeprecationWarning` you're in the right place:

> 'urllib3[secure]' extra is deprecated and will be removed in a future release of urllib3 2.x.

> 'urllib3.contrib.pyopenssl' module is deprecated and will be removed in a future release of urllib3 2.x.

These warnings mean you're using deprecated code. But no worries! You can stop the deprecation warnings by following these steps:

- **If you're using Python 2.7.10 or later and OpenSSL 1.1.1+** (this is almost everyone) then you won't need the `urllib3[secure]` extra or `urllib3.contrib.pyopenssl` module. Both were necessary a long time ago, but aren't necessary any more! You can read more context below if you're interested.

- **If you're using an old version of Requests** (ie <2.26.0) then you should upgrade to a newer version. [Requests v2.26.0](https://github.com/psf/requests/blob/main/HISTORY.md#2260-2021-07-13) stopped using `urllib3[secure]` and `urllib3.contrib.pyopenssl` and didn't encounter any issues for users upgrading.

- **If you're using an old version of Selenium** (ie <4.4.3) then you should upgrade to a newer version. [Selenium v4.4.1](https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES) stopped using urllib3[secure], but only in v4.4.3 was an appropriate dependency on certifi added.

- Ensure that you're not installing `urllib3[secure]` anywhere, either in a `requirements.txt` or directly. You should remove all instances of `[secure]` and instead only install `urllib3`.

- ~~Ensure that your code isn't using the `urllib3.contrib.pyopenssl` module.~~ We've decided to undeprecate this module in https://github.com/urllib3/urllib3/issues/3126.

- Ensure that none of your dependencies are installing the `urllib3[secure]` extra. We have [created a list](https://github.com/urllib3/urllib3/issues/2700) of projects using the `urllib3[secure]` extra which we have notified and created a PR for each but not all have responded. If you find additional packages then notifying the package about this issue is appreciated.

- ~~Ensure that none of your dependencies are using the `urllib3.contrib.pyopenssl` module. If you find a package that is using this module then I suggest notifying the package by pointing to this issue.~~ We've decided to undeprecate this module in https://github.com/urllib3/urllib3/issues/3126.

[...]

PR #2896, opened 2023-01-25
More Delegation in PyOpenSSL Contrib Module

brocktimus:
Hey All,

I was trying to use the PyOpenSSL module externally of urllib3 because it's a fantastic shim to make it act more like a stdlib ssl context. I ran into issues because WrappedSocket doesn't expose the "full" socket interface such as settimeout and do_handshake. This was an easy fix to just delegate to the PyOpenSSL Connection, which in turn delegates to the underlying socket.

Totally understand if this is rejected because its not needed by urllib3 and also because the contrib module seems soft deprecated  :-).

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -295,6 +295,9 @@ class WrappedSocket:
         self._io_refs = 0
         self._closed = False
 
+    def __getattr__(self, name: str) -> typing.Any:
+        return getattr(self.connection, name)
+
     def fileno(self) -> int:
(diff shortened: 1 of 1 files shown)


sethmlarson:
Hello @brocktimus, thanks for taking the time to open this PR. Unfortunately we've deprecated and plan to remove the `urllib3.contrib.pyopenssl` module in v2.1.0 as it's no more secure now than the standard library and is a maintenance burden whenever implementing new TLS features. See here for more info: https://github.com/urllib3/urllib3/issues/2680

Due to this, I'm going to decline this PR. Thanks again!

PR #2898, opened 2023-01-25
Add '.' to acceptable characters in a URL scheme

kenballus:
Addresses #2897

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,12 +2,6 @@
 source =
     urllib3
 
-omit =
-    *urllib3/packages/*
-    *urllib3/contrib/pyopenssl.py
-    *urllib3/contrib/securetransport.py
-    *urllib3/contrib/_securetransport/*
-
 [paths]
 source =
     src/urllib3
@@ -15,6 +9,11 @@ source =
     *\urllib3
 
 [report]
+omit =
+    src/urllib3/contrib/pyopenssl.py
+    src/urllib3/contrib/securetransport.py
+    src/urllib3/contrib/_securetransport/*
+
 exclude_lines =
     except ModuleNotFoundError:
     except ImportError:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      # Ignore all patch releases as we can manually
+      # upgrade if we run into a bug and need a fix.
+      - update-types: ["version-update:semver-patch"]
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: "actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8"
+        uses: "actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b"
         with:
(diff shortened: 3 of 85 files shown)


sethmlarson (changes requested):
Unfortunately we can't add support for this because we support schemeless URLs ie "example.com". You'll have to work around that requirement.

kenballus:
Can you point me to a URL that should work with this but doesn't? The simple schemeless URLs I've tested (like example.com) work as expected even with this patch.

kenballus:
Oh! I can see from the CI results that schemeless URLs with ports fail. That's probably a more common use case than schemes with '.' (although RFC2056 defines two URL schemes that use '.')
I'll see if there's a way I can find a happy medium on this.

sigmavirus24:
2065 predates 3986 and the more recent non-RFC URL standards which don't allow for a '.' in the scheme

kenballus:
> the more recent non-RFC URL standards which don't allow for a '.' in the scheme

Which standards are you talking about, specifically? Both RFC3986 and WHATWG state that '.' should be allowed in a scheme.

WHATWG:
"A URL-scheme string must be one [ASCII alpha](https://infra.spec.whatwg.org/#ascii-alpha), followed by zero or more of [ASCII alphanumeric](https://infra.spec.whatwg.org/#ascii-alphanumeric), U+002B (+), U+002D (-), and U+002E (.)."

3986:
`scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )`

> 2065 predates 3986
I don't think this is relevant. RFC2056 is not a URL standard; it just defines some special use cases for URLs. It is not obsoleted by 3986, and I think URLs of those forms see use inside libraries.
If you'd like other examples of schemes with '.', take a look  in [the registry](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml).
They define the following dotted schemes:
```
iris.beep
iris.lwz
iris.xpc
iris.xpcs
microsoft.windows.camera
microsoft.windows.camera.multipicker
microsoft.windows.camera.picker
soap.beep
soap.beeps
xmlrpc.beep
xmlrpc.beeps
z39.50
z39.50r
z39.50s
```

sethmlarson:
@kenballus Yes dots are allowed in schemes in both RFC 3986 and WHATWG URL. Any progress on this PR btw? I know I've looked in the past and it didn't seem like there was much that could be done, unfortunately.

kenballus:
I think something like the following could work:
Let `t` be the contents of the field in question (either a scheme or a host, but not sure which because it contains a `'.'`)
If `t` is followed by a colon, and the characters after the colon do not match `r"^[0-9]+(\?|#|/)"`, then `t` is a scheme.
Otherwise, `t` is a host.

This allows for both dotted schemes and schemeless URLs.
This leaves two ambiguous situations that I can think of:
 - Dotted scheme URLs with relative paths that begin with a numeric segment (i.e. `'soap.beep:123/'`) will be misinterpreted as schemeless URLs with port numbers. (`soap.beep` is a real scheme!)
 - Schemeless URLs that have a superfluous `':'` (i.e. `'example.com:/'`) will be misinterpreted as URLs with dotted schemes.

This seems like a reasonable middle ground; I don't imagine there are many people doing either of those things, but I'm not sure.

sethmlarson:
I think we should close this PR in favor of https://github.com/urllib3/urllib3/issues/2920 to start and then eventually remove support for scheme-less URLs and thus allowing dots in schemes. What do you think about this @kenballus?

kenballus:
@sethmlarson  Sounds good to me.

Issue #2899, opened 2023-01-26
_UNRESERVED_PAT shouldn't contain '!'


kenballus:
### _UNRESERVED_PAT shouldn't contain '!'

From src/urllib3/util/url.py:
```python3
_UNRESERVED_PAT = (
    r"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._!\-~"
)
```

The RFC defines unreserved like this:
`unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"`

kenballus:
Somehow, and I don't know how this slipped past me, but this character class contains '.' when it should contain '\.'. This would theoretically allow for any character other than '\n' to be in a zoneid, which is definitely not good.

PR #2900, opened 2023-01-26
Fix UNRESERVED regex.

kenballus:
Addresses #2899


sethmlarson (approved):
Great eye, thank you for this. Would you also open a PR to the `1.26.x` branch with this fix?

Issue #2901, opened 2023-01-26
_idna_encode checks for non-ASCII characters incorrectly


kenballus:
### _idna_encode checks for non-ASCII characters incorrectly
From `src/urllib3/util/url.py`:
```python3
if name and any([ord(x) > 128 for x in name]):
    try:
        import idna
```
Since the ASCII range ends at `chr(127)`, the comparison should use `>=` rather than `>`.

PR #2902, opened 2023-01-26
Fix `_idna_encode`'s handling of `'\x80'`

kenballus:
### Fix `_idna_encode`'s handling of `'\x80'`
Addresses #2901.

Also removes explicity code point comparisons in favor of `str.isascii`.


sethmlarson (approved):
Thank you, if you could open a PR to backport this to `1.26.x` that'd be appreciated as well.

PR #2905, opened 2023-01-29
Adjust _REG_NAME_PAT to reject invalid characters

kenballus:
Addresses #2802.


kenballus:
It might be preferable to percent-encode invalid characters instead of rejecting them. Feedback welcome!

jayaddison:
If you have time @kenballus, there's another potential regex simplification that I've been wondering about within the same Python module.  Perhaps we could trade each other some code review/inspection time?

kenballus:
> If you have time @kenballus, there's another potential regex simplification that I've been wondering about within the same Python module. Perhaps we could trade each other some code review/inspection time?

I am definitely up for that. Shoot me an email and we can set up a time.
(My email is listed at the bottom of [this page](http://kallus.org))

jayaddison:
Despite my initial enthusiasm I haven't found the time to arrange a time yet, my apologies.  A couple of questions that did occur to me in the meantime:

- Could/should the [`rfc3986` library](https://github.com/python-hyper/rfc3986.git/) provide much of the validation here (perhaps even instead of some/all of the regexes currently in use)?
- Is support for WHATWG URLs required here?

kenballus:
> Could/should the [`rfc3986` library](https://github.com/python-hyper/rfc3986.git/) provide much of the validation here (perhaps even instead of some/all of the regexes currently in use)?

Unfortunately, that library has URL parsing/validating issues as well. See [this issue](https://github.com/python-hyper/rfc3986/issues/30).

> Is support for WHATWG URLs required here?

Some parts of WHATWG are supported, like '\\' to '/' translation, but my understanding is that the RFCs are prioritized over WHATWG in this library.

jayaddison:
Thanks @kenballus, and sorry for the delays responding.  I think I'm beyond my level of understanding on this one, so I'm going to step aside.

sethmlarson:
Closing this with justification here: https://github.com/urllib3/urllib3/issues/2802#issuecomment-1519130036

PR #2919, opened 2023-02-14
Revert "Workaround GitHub not supporting Flit metadata"

pquentin:
This reverts commit 954ff5989246e45ef2490d04d63d43c94f6d9bc2. GitHub's dependency graph now supports PEP 621.

https://github.blog/changelog/2023-02-13-dependency-graph-supports-the-python-pep-621-standard/

Thanks @ofek for the tip.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,6 @@
 
 import sys
 
-from setuptools import setup
-
 sys.stderr.write(
     """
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
🎉  So excited to get rid of this hack!

PR #2923, opened 2023-03-04
Fix `test_default_ssl_context_ssl_min_max_versions` for pyOpenSSL

illia-v:
I noticed the following errors on Fedora:
```
FAILED test/contrib/test_pyopenssl.py::TestHTTPS::test_default_ssl_context_ssl_min_max_versions - AssertionError: assert <TLSVersion.MAXIMUM_SUPPORTED: -1> == <TLSVersion.TLSv1_3: 772>
FAILED test/contrib/test_pyopenssl.py::TestHTTPS_TLSv1_2::test_default_ssl_context_ssl_min_max_versions - AssertionError: assert <TLSVersion.MAXIMUM_SUPPORTED: -1> == <TLSVersion.TLSv1_3: 772>
FAILED test/contrib/test_pyopenssl.py::TestHTTPS_TLSv1_3::test_default_ssl_context_ssl_min_max_versions - AssertionError: assert <TLSVersion.MAXIMUM_SUPPORTED: -1> == <TLSVersion.TLSv1_3: 772>
```

#2642 fixed `test_default_ssl_context_ssl_min_max_versions` for cases when `ssl.SSLContext` was used.
But the test function is reused for cases when pyOpenSSL- or SecureTransport-backed SSL support is injected and urllib3 does set the default maximum version.
https://github.com/urllib3/urllib3/blob/8966d780237c639f0c31ce81d0b414adb05eeccb/src/urllib3/contrib/pyopenssl.py#L428
https://github.com/urllib3/urllib3/blob/8966d780237c639f0c31ce81d0b414adb05eeccb/src/urllib3/contrib/securetransport.py#L756

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -960,15 +960,20 @@ class TestHTTPS(HTTPSDummyServerTestCase):
     def test_default_ssl_context_ssl_min_max_versions(self) -> None:
         ctx = urllib3.util.ssl_.create_urllib3_context()
         assert ctx.minimum_version == ssl.TLSVersion.TLSv1_2
-        # urllib3 is not expected to change the maximum version, so the
-        # version should be the same as in a pure context. It will be
-        # either the `ssl.TLSVersion.MAXIMUM_SUPPORTED` magic constant
-        # or one of the exact versions if a system defines it.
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Since we're hopefully removing these long term this is probably the best way to solve this. Thanks Illia!

PR #2925, opened 2023-03-06
Use the default attestation name

sethmlarson:
Instead of providing a name we should use the default `multiple.intoto.jsonl`. After this is approved I'll rename all the existing released filenames to match so we can document a strategy for verifying provenance across all releases.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -58,7 +58,6 @@ jobs:
     uses: "slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.4.0"
     with:
       base64-subjects: "${{ needs.build.outputs.hashes }}"
-      attestation-name: "urllib3.intoto.jsonl"
       upload-assets: true
       compile-generator: true # Workaround for https://github.com/slsa-framework/slsa-github-generator/issues/1163
 


pquentin (approved):

illia-v:
When I saw this type of files in releases, I thought it would be more convenient to have versions in addition to the package name in their file names.

It may be useful if you have to download and store bundles for multiple packages or versions of one package. [The spec](https://github.com/in-toto/attestation/blob/main/spec/v1.0-draft/bundle.md#file-naming-convention) does not forbid this.

If we generated attestations only for wheels, the file for the latest release would be named `urllib3-1.26.14-py2.py3-none-any.whl.intoto.jsonl` based on [this release info](https://github.com/slsa-framework/slsa-github-generator/releases/tag/v1.2.1#:~:text=%3Cartifact%20filename%3E.intoto.jsonl).
Since a bundle includes attestations for both a wheel and a source code archive, what do you think about naming it `urllib3-1.26.14.intoto.jsonl`?

sethmlarson:
Hmm, I'm not sure what is best or recommended here. Users could choose which name they want via `curl --output <name> <url>`. Going to ping the SLSA folks for an opinion.

sethmlarson:
Response from SLSA folks was that it doesn't matter what you go with as long as it's consistent. I'm going to move forward with `multiple.intoto.jsonl` since I believe this will be commonly done as it's the default value.

PR #2943, opened 2023-03-23
Added support for union operators to ``HTTPHeaderDict``

wimglenn:
This allows a simple way to merge headers provided in a request with the ``PoolManager`` headers, rather than replace them.

Related issue: https://github.com/urllib3/urllib3/issues/2254

I was inspired by Ran's example there:

```
pm.request('POST', 'https://foo.bar', headers={
    **pm.headers,
    'Content-Type': 'application/json',
}, ...)
```

The basic usage was also tested [here](https://github.com/urllib3/urllib3/pull/2274/files). However this approach has limitations - the return type is always a dict, and that happens at the Python level so there isn't much flexibility using that operation if you want to merge a `HTTPHeaderDict` instance either from the pool manager or from the request. The standard dict must have unique keys, so no merging of existing keys is possible, only overwriting.

```
>>> pm = urllib3.PoolManager(headers=HTTPHeaderDict({"k1": "v1"}))
>>> pm.headers
HTTPHeaderDict({'k1': 'v1'})
>>> {**pm.headers, "k2": "v2"}
{'k1': 'v1', 'k2': 'v2'}
>>> {**pm.headers, "k1": "v2"}  # uh-oh, we lost "v1"
{'k1': 'v2'}
```

My proposal would allow:

```
pm.request('POST', 'https://foo.bar', headers=pm.headers | more_headers, ...)
```

Which can merge, on either side, as expected:

```
>>> pm.headers | {"k1": "v2"}
HTTPHeaderDict({'k1': 'v1, v2'})
>>> {"k1": "v2", "k3": "v3"} | pm.headers
HTTPHeaderDict({'k1': 'v2, v1', 'k3': 'v3'})
```

Standard dicts support unions since Python 3.9 (see [_PEP 584 – Add Union Operators To dict_](https://peps.python.org/pep-0584/)) so I've tried to make the `HTTPHeaderDict` quack in that way.

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -432,3 +432,24 @@ class HTTPHeaderDict(typing.MutableMapping[str, str]):
         if header_name in self:
             return potential_value in self._container[header_name.lower()][1:]
         return False
+
+    def __ior__(self, other: object) -> HTTPHeaderDict:
+        # Supports extending a header dict in-place using operator |=
+        # combining items with add instead of __setitem__
+        maybe_constructable = ensure_can_construct_http_header_dict(other)
+        if maybe_constructable is None:
+            return NotImplemented
+        other_as_http_header_dict = type(self)(maybe_constructable)
+        self.extend(other_as_http_header_dict)
+        return self
+
+    def __or__(self, other: object) -> HTTPHeaderDict:
+        # Supports merging header dicts using operator |
+        # combining items with add instead of __setitem__
+        maybe_constructable = ensure_can_construct_http_header_dict(other)
+        if maybe_constructable is None:
+            return NotImplemented
+        result = self.copy()
+        other_as_http_header_dict = type(self)(maybe_constructable)
+        result.extend(other_as_http_header_dict)
+        return result
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -453,3 +453,13 @@ class HTTPHeaderDict(typing.MutableMapping[str, str]):
         other_as_http_header_dict = type(self)(maybe_constructable)
         result.extend(other_as_http_header_dict)
         return result
+
+    def __ror__(self, other: object) -> HTTPHeaderDict:
+        # Supports merging header dicts using operator | when other is on left side
+        # combining items with add instead of __setitem__
+        maybe_constructable = ensure_can_construct_http_header_dict(other)
+        if maybe_constructable is None:
+            return NotImplemented
+        result = type(self)(maybe_constructable)
+        result.extend(self)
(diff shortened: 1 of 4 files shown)


wimglenn:
@sethmlarson I've addressed the coverage miss.  `nox -s format` didn't seem to want to change anything. Not sure what mypy is unhappy about - it seems to think that the `HTTPHeaderDict` is not a `typing.Mapping[str, str]` or something? I'm not very familiar with mypy.

wimglenn:
@sethmlarson  ping?

illia-v:
Other options can be `headers=http.headers | HTTPHeaderDict(added_headers)` or using `headers` directly `headers=headers | added_headers`, but the latter is less likely to be used in reality probably.

I think requiring the union operator support in the annotation in `PoolManager` is not an option until urllib3 supports Python versions which have dicts without the support.
Something that may be worth checking is what can happen if headers are always converted to `HTTPHeaderDict` on the following line:
https://github.com/urllib3/urllib3/blob/10ceef1c8944a55b51ad53651998c8aa9caedacb/src/urllib3/_request_methods.py#L52

src/urllib3/_collections.py
```diff
@@ -432,3 +432,34 @@ def _has_value_for_header(self, header_name: str, potential_value: str) -> bool:
...
             return potential_value in self._container[header_name.lower()][1:]
         return False
+
+    def __ior__(self, other: object) -> HTTPHeaderDict:
+        # Supports extending a header dict in-place using operator |=
+        # combining items with add instead of __setitem__
+        maybe_constructable = ensure_can_construct_http_header_dict(other)
+        if maybe_constructable is None:
+            return NotImplemented
+        other_as_http_header_dict = type(self)(maybe_constructable)
```
illia-v, line 442: Are there any advantages of the conversion in `__ior__` and `__or__`?

wimglenn: IIRC mypy was unhappy about that, because `__ior__` and `__or__` must accept any object as "other" (and will opt-out of the comparison by returning `NotImplemented`), but `HTTPHeaderDict.extend` explicitly wants a `ValidHTTPHeaderSource`.

illia-v: This creates new objects and looks to have little functional benefit because `extend` will do similar kind of validation. Can `typing.cast` be used instead?

wimglenn: Yes, likely. I'll try it out when I have the spare time and report back.

illia-v: Nice, thanks!

illia-v (approved):

Issue #2941, opened 2023-03-22
Investigate zstandard DecodeError coverage issue


sethmlarson:
We lapsed on 100% coverage, specifically something to do with a `DecoderError` case in `ZstdDecoder`. I'll note that zstandard 0.20.0 was released in February so that may have something to do with this lapse in coverage?

```
Name                                     Stmts   Miss  Cover   Missing
----------------------------------------------------------------------
src/urllib3/__init__.py                     24      0   100%
src/urllib3/_base_connection.py             12      0   100%
src/urllib3/_collections.py                186      0   100%
src/urllib3/_request_methods.py             43      0   100%
src/urllib3/_version.py                      1      0   100%
src/urllib3/connection.py                  272      0   100%
src/urllib3/connectionpool.py              303      0   100%
src/urllib3/contrib/__init__.py              0      0   100%
src/urllib3/contrib/socks.py                55      0   100%
src/urllib3/exceptions.py                  103      0   100%
src/urllib3/fields.py                       83      0   100%
src/urllib3/filepost.py                     30      0   100%
src/urllib3/poolmanager.py                 205      0   100%
src/urllib3/response.py                    510      1    99%   177
src/urllib3/util/__init__.py                 1      0   100%
src/urllib3/util/connection.py              56      0   100%
src/urllib3/util/proxy.py                    8      0   100%
src/urllib3/util/request.py                 86      0   100%
src/urllib3/util/response.py                25      0   100%
src/urllib3/util/retry.py                  156      0   100%
src/urllib3/util/ssl_.py                   130      0   100%
src/urllib3/util/ssl_match_hostname.py      58      0   100%
src/urllib3/util/ssltransport.py           146      0   100%
src/urllib3/util/timeout.py                 63      0   100%
src/urllib3/util/url.py                    188      0   100%
src/urllib3/util/util.py                    23      0   100%
src/urllib3/util/wait.py                    45      0   100%
----------------------------------------------------------------------
TOTAL                                     2812      1    99%
```

The investigation should:
- Determine the root cause of the change in behavior/coverage
- Get us back to 100% coverage
- Potentially report back to the zstandard package if necessary

sethmlarson:
Closed in https://github.com/urllib3/urllib3/pull/2949 @Ousret you can submit a bounty claim on OpenCollective to collect the $100. Thanks again!

sethmlarson:
Thanks @Ousret! I've approved your expense in OpenCollective for $100

PR #2949, opened 2023-03-29
Fixed coverage regression when zstd >= 0.20 is installed

Ousret:
Hello,

Fix the coverage regression seen recently and reported by #2941
I confirm that this _regression_ appears once python-zstd 0.20 is installed no matter the interpreter version.

The unit test that is responsible for the missing coverage is:

```python
@onlyZstd()
@pytest.mark.parametrize("data", [b"foo", b"x" * 100])
def test_decode_zstd_error(self, data: bytes) -> None:
    ...
```

The dependency (not the Python interface but rather the native decoder) recently brought changes that make the decoding stricter. And that is good news. 

See attached traceback from before (<0.20)

```
  File "/home/ahmed/PycharmProjects/urllib3/test/test_response.py", line 354, in test_decode_zstd_error
    HTTPResponse(fp, headers={"content-encoding": "zstd"}, preload_content=False).read()
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 884, in read
    data = self._decode(data, decode_content, flush_decoder)
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 448, in _decode
    data += self._flush_decoder()
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 458, in _flush_decoder
    return self._decoder.decompress(b"") + self._decoder.flush()
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 177, in flush
    raise DecodeError("Zstandard data is incomplete")
urllib3.exceptions.DecodeError: Zstandard data is incomplete
```

and after (>=0.20)

```
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 438, in _decode
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 172, in decompress
    return self._obj.decompress(data)  # type: ignore[no-any-return]
zstd.ZstdError: zstd decompressor error: Unknown frame descriptor

  File "/home/ahmed/PycharmProjects/urllib3/test/test_response.py", line 354, in test_decode_zstd_error
    HTTPResponse(fp, headers={"content-encoding": "zstd"}, preload_content=False).read()
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 884, in read
    data = self._decode(data, decode_content, flush_decoder)
  File "/home/ahmed/PycharmProjects/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/response.py", line 442, in _decode
    raise DecodeError(
urllib3.exceptions.DecodeError: ('Received response with content-encoding: zstd, but failed to decode it.', ZstdError('zstd decompressor error: Unknown frame descriptor'))
```

The test actually passed because we are looking for `DecodeError` and not for `ZstdError` raised in `.../urllib3/response.py", line 172`.

So now without `preload_content=False` in the Response instance, the decoding fail early in the process, thus the missing coverage line.

I looked at other (_similar_) PR and I purposely did not add a news fragment.

Finally, regarding the last point in the issue:

> Potentially report back to the zstandard package if necessary

I could be mistaken but, I think that is not necessary as it makes more sense for it to fail there instead of later. It feels safer that way.

I hope I was clear enough in this investigation.
Regards,

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -349,13 +349,11 @@ class TestResponse:
     @onlyZstd()
     @pytest.mark.parametrize("data", [b"foo", b"x" * 100])
     def test_decode_zstd_error(self, data: bytes) -> None:
-        fp = BytesIO(data)
+        data = zstd.compress(data)
+        fp = BytesIO(data[:-1])  # purposely use incomplete frame
 
         with pytest.raises(DecodeError):
-            r = HTTPResponse(
-                fp, headers={"content-encoding": "zstd"}, preload_content=False
-            )
-            r.read()
+            HTTPResponse(fp, headers={"content-encoding": "zstd"})
 
     def test_multi_decoding_deflate_deflate(self) -> None:
         data = zlib.compress(zlib.compress(b"foo"))
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -352,7 +352,10 @@ class TestResponse:
         fp = BytesIO(data)
 
         with pytest.raises(DecodeError):
-            HTTPResponse(fp, headers={"content-encoding": "zstd"})
+            r = HTTPResponse(
+                fp, headers={"content-encoding": "zstd"}, preload_content=False
+            )
+            r.read()
 
     def test_multi_decoding_deflate_deflate(self) -> None:
         data = zlib.compress(zlib.compress(b"foo"))
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -349,8 +349,16 @@ class TestResponse:
     @onlyZstd()
     @pytest.mark.parametrize("data", [b"foo", b"x" * 100])
     def test_decode_zstd_error(self, data: bytes) -> None:
+        fp = BytesIO(data)
+
+        with pytest.raises(DecodeError):
+            HTTPResponse(fp, headers={"content-encoding": "zstd"})
+
+    @onlyZstd()
+    @pytest.mark.parametrize("data", [b"foo", b"x" * 100])
+    def test_decode_zstd_incomplete(self, data: bytes) -> None:
         data = zstd.compress(data)
-        fp = BytesIO(data[:-1])  # purposely use incomplete frame
+        fp = BytesIO(data[:-1])
 
         with pytest.raises(DecodeError):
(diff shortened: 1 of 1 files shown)


pquentin:
Also note that with your current pull request a new line is uncovered:

https://github.com/urllib3/urllib3/blob/969fdb9545caeb4e8141ea939231cf03a2db606f/src/urllib3/response.py#L177

sethmlarson:
Can we keep the existing test case so we continue to cover the "completely invalid" input case in addition to this new "invalid frame" case?

Ousret:
> What do you think about catching ZstdError as we catch zlib.error and reraising DecodeError? We want all invalid inputs here to raise DecodeError

Absolutely, I agree. Otherwise, I think that should be done in another PR, unless you want me to do it there, your call.
While we're at it, maybe brotly decoding should have the same treatment.

> Also note that with your current pull request a new line is uncovered:

Was a tiny mistake from my end, it is corrected now.

> Can we keep the existing test case so we continue to cover the "completely invalid" input case in addition to this new "invalid frame" case?

Yes, I was going to do it. It is a good idea for sure.

sethmlarson:
> > What do you think about catching ZstdError as we catch zlib.error and reraising DecodeError? We want all invalid inputs here to raise DecodeError
> 
> Absolutely, I agree. Otherwise, I think that should be done in another PR, unless you want me to do it there, your call. While we're at it, maybe brotly decoding should have the same treatment.

Could you either open an issue for this or create a PR with this fix as well? Thanks much.

Ousret:
> Could you either open an issue for this or create a PR with this fix as well? Thanks much.

Will do.

Issue #2920, opened 2023-02-14
Deprecate schemeless URLs?


kenballus:
### Context

Schemeless URLs can cause a lot of parsing confusion because they aren't standardized in the RFC. Examples like the following are especially weird, and could cause problems when interoperability with other parsers matters:
```
"evil.com://good.com"
```

Here are some popular URL parsers' interpretations:
urllib3:
```
Scheme:   (nil)
Userinfo: (nil)
Host:     evil.com
Port:     (nil)
Path:     //good.com
Query:    (nil)
Fragment: (nil)
```

cpython urllib:
```
Scheme:   evil.com
Userinfo: (nil)
Host:     good.com
Port:     (nil)
Path:     (nil)
Query:    (nil)
Fragment: (nil)
```

furl:
```
Scheme:   evil.com
Userinfo: (nil)
Host:     good.com
Port:     (nil)
Path:     (nil)
Query:    (nil)
Fragment: (nil)
```

hyperlink:
```
Scheme:   evil.com
Userinfo: (nil)
Host:     good.com
Port:     (nil)
Path:     /
Query:    (nil)
Fragment: (nil)
```

rfc3986:
```
Scheme:   evil.com
Userinfo: (nil)
Host:     good.com
Port:     (nil)
Path:     (nil)
Query:    (nil)
Fragment: (nil)
```

yarl:
```
Scheme:   evil.com
Userinfo: (nil)
Host:     good.com
Port:     (nil)
Path:     /
Query:    (nil)
Fragment: (nil)
```

As you can see, we're the outlier here.

In my opinion, this is something worth fixing, but I imagine that schemeless URLs are in pretty widespread use with urllib3. Thus, we might consider adding a DeprecationWarning that encourages people to explicitly state their schemes.

sethmlarson:
Totally agree that this is something worth fixing, thank you for doing all this research and opening this. My hope is that schemeless URLs aren't as widespread as we think and that we've been supporting this mostly for backwards compatibility and not because a large swath of the ecosystem relies on it.

I think adding a `DeprecationWarning` as you mentioned that links to this issue (we can edit the issue body to provide actionable feedback for users) and says that a removal will occur in a future version. We can then get that released in 1.26.x and 2.x to start warning users and gathering feedback if necessary.

DarkPhily:
Hi,
I'm interested in contributing.
I would like to give this issue a try, but I'm not sure what should be accomplished at the end?
Raise an DeprecationWarning?
[...]

PR #2950, opened 2023-03-29
Deprecate schemeless url, warn that the protocol must be explicitly given

Ousret:
?answer to #2920 

I drafted this PR for an initial deprecation when URLs are missing a scheme (post parsing).
It is probably wrong to say "unsupported in v2.1.0" as the original issue did not mention that was the target. Should I be vague instead?

As for the news fragment, by looking at another similar PR, it did not seem to be required. Nonetheless, I added a line in the v2 migration guide that can be reformulated.


src/urllib3/poolmanager.py
```diff
@@ -311,6 +313,15 @@ def connection_from_context(
         ``request_context`` must at least contain the ``scheme`` key and its
         value must be a key in ``key_fn_by_scheme`` instance variable.
         """
+        if "implicit_scheme" in request_context:
+            warnings.warn(
+                "Schemeless URL are deprecated and will be unsupported "
```
sethmlarson, line 318: To avoid the use of "Schemeless" (which might be tough to parse for some folks) let's opt for something like:

"URLs without a scheme (ie 'https://') are deprecated and begin raising an error in a future version of urllib3. To avoid this DeprecationWarning ensure all URLs start with 'https://' or 'http://'. Read more in this issue: ..."

src/urllib3/poolmanager.py
```diff
@@ -294,6 +294,8 @@ def connection_from_host(
             raise LocationValueError("No host specified.")
 
         request_context = self._merge_pool_kwargs(pool_kwargs)
+        if scheme is None:
```
sethmlarson, line 297: Almost no one uses urllib3's `connection_from_context()` method directly, so I think we're safe to hoist the warning to come from `urlopen` so that the `stacklevel` isn't a strange looking function.

test/test_poolmanager.py
```diff
@@ -259,6 +259,22 @@ def test_http_connection_from_context_case_insensitive(self) -> None:
...
+            p.connection_from_url("evil.com://good.com")
+
+        msg = (
+            "Schemeless URL are deprecated and will be unsupported "
+            "in urllib3 v2.1.0. Read more in this issue: "
+            "https://github.com/urllib3/urllib3/issues/2920"
+        )
+
+        assert len(records) == 1
+        assert isinstance(records[0].message, Warning)
```
sethmlarson, line 275: `isinstance(..., DeprecationWarning)`

test/test_poolmanager.py
```diff
@@ -259,6 +259,22 @@ def test_http_connection_from_context_case_insensitive(self) -> None:
         assert pool is other_pool
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
 
+    def test_deprecated_no_scheme(self) -> None:
+        p = PoolManager()
+
+        with pytest.warns(DeprecationWarning) as records:
+            p.connection_from_url("evil.com://good.com")
```
sethmlarson, line 266: Let's use `.request()` here as our entry point

sethmlarson (changes requested):
Thanks for this! I have a few comments for you:

- Let's remove the "removed in urllib3 v2.1.0" since it likely won't be removed for some time.
- Add a test case that a warning isn't raised when redirecting to a non-absolute URL
- Add a test case for connecting to an schemeless URL in all proxy flavors (HTTP->HTTP, HTTPS->HTTP, HTTP->HTTPS, HTTPS->HTTPS) look at the other tests for a nice way to do this with the test fixtures.

src/urllib3/poolmanager.py
Ousret: Done ✔️

src/urllib3/poolmanager.py
Ousret: Done ✔️

test/test_poolmanager.py
Ousret: Done ✔️

test/test_poolmanager.py
Ousret: Changed, yeah, 🤦 
tx.

Ousret:
Thank you for your review.
Applied all of your guidance.

Except for the last one.

> Add a test case for connecting to an schemeless URL in all proxy flavors (HTTP->HTTP, HTTPS->HTTP, HTTP->HTTPS, HTTPS->HTTPS) look at the other tests for a nice way to do this with the test fixtures.

I tried to draft a test but faced several problems. Most likely due to misunderstandings on my end.
When you said `HTTP->HTTP, HTTPS->HTTP`, you meant `ProxyScheme->TargetScheme` I suppose.

I wanted to use the existing test case `class TestHTTPProxyManager(HTTPDummyProxyTestCase)` which has the required props.
But when requesting a schemeless URL having a port specified, we *hugh ironically* face an error on URL parsing like already discussed.
And we can't infer HTTPS from a schemeless URL, so testing that case seems a bit of a stretch. Should we _just_ do `HTTP->HTTP, HTTPS->HTTP`?

WDYT?

Ousret:
ping @sethmlarson

src/urllib3/poolmanager.py
```diff
@@ -419,6 +419,16 @@ def urlopen(  # type: ignore[override]
         """
         u = parse_url(url)
 
+        if u.scheme is None:
+            warnings.warn(
+                "URLs without a scheme (ie 'https://') are deprecated and begin raising an error "
```
sethmlarson, line 424: nit: `...are deprecated and will raise an error in a future version of urllib3`

test/test_poolmanager.py
```diff
@@ -259,6 +259,23 @@ def test_http_connection_from_context_case_insensitive(self) -> None:
...
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
 
+    def test_deprecated_no_scheme(self) -> None:
+        p = PoolManager()
+
+        with pytest.warns(DeprecationWarning) as records:
+            p.request(method="GET", url="evil.com://good.com")
+
+        msg = (
+            "URLs without a scheme (ie 'https://') are deprecated and begin raising an error "
```
sethmlarson, line 269: Message will have to be updated here too

sethmlarson (changes requested):
Sorry for delay, here's the last few changes:

test/with_dummyserver/test_connectionpool.py
```diff
@@ -465,6 +465,12 @@ def test_tunnel(self) -> None:
             finally:
                 conn.close()
 
+    def test_redirect_relative_url_no_deprecation(self) -> None:
+        with HTTPConnectionPool(self.host, self.port) as pool:
+            with warnings.catch_warnings():
```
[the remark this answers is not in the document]
Ousret: Done ✔️

src/urllib3/poolmanager.py
Ousret: Done ✔️

test/test_poolmanager.py
Ousret: Done ✔️

Ousret:
Thank you for the review.
Everything should be in order 👌

sethmlarson (approved):
LGTM, thank you!

Ousret:
Sorry for the extra ping, does this effectively fulfill the bounty requirements? or there's something more? I am unsure.

pquentin:
For context if anyone stumbles upon this, yes we approved the expense: https://github.com/urllib3/urllib3/issues/2920#issuecomment-1566177949

PR #2952, opened 2023-03-30
Add backoff jitter to Retry

tmct:
A nice-to-have, to enable the approach recommended in e.g. https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/


illia-v (commented):
I think this is a good idea. Thanks @tmct!

What do you think about adding a separate `:param float backoff_jitter:` entry to the changed docstring and making the parameter default to 0 to be similar to `backoff_factor`?

tmct:
Hi Illia - no problem! I have seen people installing Tenacity dependencies because this feature didn't exist, so I'd be happy to see it available in urllib3 so that they don't necessarily need to do that.

So yes, that's what I was originally going to do - to add a new `:param` entry to the docstring - but I noticed something funny about the existing docstring. The existing parameter "backoff_max" is documented only as part of the "backoff_factor" docstring, and is described only in the context of the latter... Maybe that makes sense, or maybe it could have its own section, but it feels like we should do something consistent with all the "backoff_" params - either to list each separately or together in one description for "backoff_factor" (which is required for them to make any sense!) - what do you think?

As for the default value of backoff_jitter: yes, you've convinced me that `0` is the equivalent to the no-jitter case and so should be the new default. I'll apply that change shortly.

Thanks!

src/urllib3/util/retry.py
```diff
@@ -289,6 +297,8 @@ def get_backoff_time(self) -> float:
             return 0
 
         backoff_value = self.backoff_factor * (2 ** (consecutive_errors_len - 1))
+        if self.backoff_jitter > 0.0:
```
illia-v, line 300: Tenacity seems to allow a negative jitter implicitly. Is there a reason to disallow it here?
https://github.com/jd/tenacity/blob/433324956abb028f6d993195d31e4dd8308115c3/tenacity/wait.py#L222
```
>>> import random
>>> random.uniform(0, -1)
-0.45467571802820805
```

illia-v:
> So yes, that's what I was originally going to do - to add a new `:param` entry to the docstring - but I noticed something funny about the existing docstring. The existing parameter "backoff_max" is documented only as part of the "backoff_factor" docstring, and is described only in the context of the latter...

Fair enough 🙂

src/urllib3/util/retry.py
tmct: Interesting. We can leave that in here too, sure, but I'll continue to leave out the random.random call in the most common (default) 0.0 case?

tmct: I've added that in with a minimum of zero

illia-v: > but I'll continue to leave out the random.random call in the most common (default) 0.0 case?

Good, this is an efficient way

illia-v (approved):
@tmct thanks for adding this nice feature. Would you like to backport it to the 1.26.x branch?

tmct:
Ah sorry, I didn't see your edit! Just finished the backport: #2988

tmct:
Thanks for reviewing

illia-v:
> Ah sorry, I didn't see your edit! Just finished the backport: #2988

I'm sorry for editing it quietly
Would you mind if we don't include it in 1.26.x? There is an intent to have only critical stuff backported at this point

PR #2661, opened 2022-06-29
Make poolmanger threadsafe.

stephan-hof:
This pull request is a follow up on 

* https://github.com/urllib3/urllib3/issues/1252#issuecomment-954958872
* https://github.com/urllib3/urllib3/issues/1252#issuecomment-957450016

As mentioned in the comment the fix is to close the pool when all references to it are gone.

I need a bit help with the changelog. Should I create a file called `changelog/1252.bugfix.rst` ?

PR #2954, opened 2023-04-07
Make poolmanger threadsafe in 1.26.x

thomasmaclean:
Backport of #2661, which addresses a thread-safety issue where accessing a `PoolManager` with many distinct origins would cause connection pools to be closed while requests are in progress. #1252 causes issues in multiprocessing workflows for dependent libraries like `requests` etc.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -1115,7 +1115,7 @@ def _normalize_host(host, scheme):
     return host
 
 
-def _close_pool_connections(pool: "queue.LifoQueue[Any]") -> None:
+def _close_pool_connections(pool):
     """Drains a queue of connections and closes each one."""
     try:
         while True:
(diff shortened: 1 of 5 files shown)


sethmlarson:
@thomasmaclean Thanks for opening this PR, unfortunately this release stream still supports Python 2.7 so you'll need to find an alternative to `weakref.finalize`, maybe there's an existing work-around for this?

thomasmaclean:
> @thomasmaclean Thanks for opening this PR, unfortunately this release stream still supports Python 2.7 so you'll need to find an alternative to `weakref.finalize`, maybe there's an existing work-around for this?

I'm copying a backporting the CPython version of finalize as it's pretty portable the way it's written. I tried to copy the pattern used for `backports_makefile` but please let me know if there's a better way to do what I did.

sethmlarson (approved):

PR #2958, opened 2023-04-12
avoid distutils Version classes are deprecated warning by dropping pytest-freezegun

graingert:
it looks like pytest-freezegun is unmaintained and it's only used in one test where it's cleaner to use freezegun directly

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,9 +1,9 @@
 coverage==7.0.4
+freezegun==1.2.2
 tornado==6.2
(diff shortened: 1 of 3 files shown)


graingert:
@sethmlarson actually it was found by @pquentin back in 2021 https://github.com/ktosiek/pytest-freezegun/issues/35

pquentin:
Still, thanks for doing this!

It's only used in one test, so this PR is fine, but note that pytest-freezer is now a thing too.

PR #2959, opened 2023-04-12
fix test_ssl_object_attributes

graingert:
https://github.com/python/cpython/issues/96931 fixes `SSLObject.shared_ciphers()` to always return None for client side sockets this is released in 3.10.11, 3.11.3 and 3.12.0a7

### DIFF
--- a/test/test_ssltransport.py
+++ b/test/test_ssltransport.py
@@ -4,6 +4,7 @@ import platform
 import select
 import socket
 import ssl
+import sys
 import typing
 from unittest import mock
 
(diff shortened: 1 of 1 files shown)


test/test_ssltransport.py
```diff
@@ -229,8 +230,16 @@ def test_ssl_object_attributes(self) -> None:
             assert ssock.selected_npn_protocol() is None
 
             shared_ciphers = ssock.shared_ciphers()
-            assert type(shared_ciphers) == list
-            assert len(shared_ciphers) > 0
+            # https://github.com/python/cpython/issues/96931
```
sethmlarson, line 233: Since we don't *really* think this test will ever fail, just to cover the lines where we're forwarding attributes, can we simplify this test case to not have all the versions? ie `assert shared_ciphers is None or (type(shared_ciphers) is list and len(shared_ciphers) > 0)`?

sethmlarson (commented):
Thanks! We will need this in the 1.26.x branch as well. One thought to potentially simplify the assert:

test/test_ssltransport.py
sethmlarson: Hmm, I was thinking we'd want to implement whatever the behavior is for that Python version, even if that behavior isn't correct by the docs?

graingert: Yeah there's a lot of options here, can I hand this PR over to you?

sethmlarson (approved):

PR #2964, opened 2023-04-15
Change dependency warning link for PySocks

dslackw:


pquentin:
Thanks! Did you mean to raise this against the 1.26 branch?

dslackw:
Yes, if you think is the right branch.

src/urllib3/contrib/socks.py
```diff
@@ -51,7 +51,7 @@
         (
             "SOCKS support in urllib3 requires the installation of optional "
             "dependencies: specifically, PySocks.  For more information, see "
-            "https://urllib3.readthedocs.io/en/latest/contrib.html#socks-proxies"
+            "https://urllib3.readthedocs.io/en/stable/advanced-usage.html#socks-proxies"
```
pquentin, line 54: I had missed from my phone that contrib was changed to advanced-usage, sorry. But this is the main branch, so we should stick with `latest` please:

```suggestion
            "https://urllib3.readthedocs.io/en/latest/advanced-usage.html#socks-proxies"
```

pquentin (changes requested):
I had missed that you change

sethmlarson (approved):
LGTM, thank you!

illia-v (approved):

PR #2966, opened 2023-04-17
feat: Remove id-token permission during build

laurentsimon:
Remove `id-token: write` permission for the build part. It is only needed for provenance signing.


sethmlarson (approved):
Thank you!

Issue #2972, opened 2023-04-18
`test/with_dummyserver/test_no_ssl.py::TestHTTPSWithoutSSL::test_simple` is invalid


graingert:
### Subject

there's supposed to be an SSLError raised but it never gets raised:

https://github.com/urllib3/urllib3/blob/f8fada5ed479e0e6b8440c819632b3a8c040cf65/test/with_dummyserver/test_no_ssl.py#L26-L29

the test should use `with pytest.raises(`

graingert:
note that the actual code raises `ImportError` when the ssl module cannot be found:

```
>>> import sys
>>> sys.modules["ssl"] = None
>>> import urllib3
>>> pool = urllib3.HTTPSConnectionPool("google.com", 443, cert_reqs="NONE")
>>> pool.request("GET", "/")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/graingert/projects/urllib3/src/urllib3/_request_methods.py", line 110, in request
    return self.request_encode_url(
  File "/home/graingert/projects/urllib3/src/urllib3/_request_methods.py", line 143, in request_encode_url
    return self.urlopen(method, url, **extra_kw)
  File "/home/graingert/projects/urllib3/src/urllib3/connectionpool.py", line 769, in urlopen
    conn = self._get_conn(timeout=pool_timeout)
  File "/home/graingert/projects/urllib3/src/urllib3/connectionpool.py", line 295, in _get_conn
    return conn or self._new_conn()
  File "/home/graingert/projects/urllib3/src/urllib3/connectionpool.py", line 1056, in _new_conn
    raise ImportError(
ImportError: Can't connect to HTTPS URL because the SSL module is not available.
```

graingert:
this test was invalidated by [`2ec3018` (#1102)](https://github.com/urllib3/urllib3/pull/1102/commits/2ec3018926427c96914373947d06cf32ec7d61fa#r97308148)

PR #2973, opened 2023-04-19
fix test_no_ssl by using subprocess to disable ssl import

graingert:
fixes https://github.com/urllib3/urllib3/issues/2972


graingert:
annoyingly this fails coverage with `src/urllib3/connection.py                  272      4    99%   32-35, 878`

pquentin:
There are ways to fix this: https://coverage.readthedocs.io/en/stable/subprocess.html. Do they apply here?

sethmlarson (approved):
LGTM, thanks!

sethmlarson:
Ah, only looked at the PR and not the coverage status, I hope you can figure out the subprocess coverage?

Issue #2970, opened 2023-04-18
pyopenssl.WrappedSocket and secureransport.WrappedSocket close implementation broken


graingert:
### Subject

calling `ssl_sock.close()` with a makefile open doesn't close the socket when the makefile closes, resulting in a ResourceWarning:

```python
ssl_sock = context.wrap_socket(sock)
f = ssl_sock.makefile()
ssl_sock.close()
f.close() # ssl_sock is still open
del f
del ssl_sock  # ResourceWarning raised here
```

additionally calling `ssl_sock.close()` twice with a makefile open causes the socket to close early

```python
with context.wrap_socket(sock) as ssl_sock:
    with ssl_sock.makefile() as f:
        ssl_sock.close()
        ssl_sock.close()  # socket closed here incorrectly
```

graingert:
the issue is caused by the close methods incorrectly dec-refing _io_refs

https://github.com/urllib3/urllib3/blob/f8fada5ed479e0e6b8440c819632b3a8c040cf65/src/urllib3/contrib/pyopenssl.py#L382-L390
https://github.com/urllib3/urllib3/blob/f8fada5ed479e0e6b8440c819632b3a8c040cf65/src/urllib3/contrib/securetransport.py#L630-L647

the implementation should be as in cpython:

https://github.com/python/cpython/blob/4e04393b2f64f9e44aefa436cbdac42db98d4ef0/Lib/socket.py#L496-L504
```python
def _real_close(self, _ss=_socket.socket):
        # This function should not reference any globals. See issue https://github.com/python/cpython/issues/808164.
        _ss.close(self)

    def close(self):
        # This function should not reference any globals. See issue https://github.com/python/cpython/issues/808164.
        self._closed = True
        if self._io_refs <= 0:
            self._real_close()
```

graingert:
I have a fix in #2842 to fix the ResourceWarning

https://github.com/urllib3/urllib3/blob/45cd5c88677f93cfdb2da5ccd9d631c004661fb1/src/urllib3/contrib/securetransport.py#L631-L650
https://github.com/urllib3/urllib3/blob/45cd5c88677f93cfdb2da5ccd9d631c004661fb1/src/urllib3/contrib/pyopenssl.py#L382-L391

but it should probably be backported to v1

PR #2842, opened 2022-11-30
raise ResourceWarnings as errors 

graingert:
fixes https://github.com/urllib3/urllib3/issues/2970

PR #2974, opened 2023-04-20
fix pyopenssl.WrappedSocket and securetransport.WrappedSocket close

graingert:
fixes https://github.com/urllib3/urllib3/issues/2970

this is just the changes extracted from #2842 that need a backport to v1


src/urllib3/contrib/securetransport.py
```diff
@@ -628,23 +629,28 @@ def shutdown(self) -> None:
...
+            self.context = None
+        if self._client_cert_chain:
+            CoreFoundation.CFRelease(self._client_cert_chain)
+            self._client_cert_chain = None
+        if self._keychain:
+            Security.SecKeychainDelete(self._keychain)
+            CoreFoundation.CFRelease(self._keychain)
+            shutil.rmtree(self._keychain_dir)
+            self._keychain = self._keychain_dir = None
+        return self.socket.close()
```
sethmlarson, line 653: Is there a reason we're returning `self.socket.close()` here?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -966,6 +967,63 @@ def socket_handler(listener: socket.socket) -> None:
...
+
+        self._start_server(consume_ssl_socket)
+        with socket.create_connection(
+            (self.host, self.port)
+        ) as sock, contextlib.closing(
+            ssl_wrap_socket(sock, server_hostname=self.host, ca_certs=DEFAULT_CA)
+        ) as ssl_sock, ssl_sock.makefile(
+            "rb"
+        ):
+            ssl_sock.close()
```
sethmlarson, line 1022: Why are we calling `close()` twice here, for coverage? Can we pull this into a separate test if so?

src/urllib3/contrib/pyopenssl.py
```diff
@@ -380,14 +380,15 @@ def shutdown(self) -> None:
...
-                return self.connection.close()  # type: ignore[no-any-return]
-            except OpenSSL.SSL.Error:
-                return
-        else:
-            self._io_refs -= 1
+        self._closed = True
+        if self._io_refs <= 0:
+            self._real_close()
+
+    def _real_close(self) -> None:
```
sethmlarson, line 387: If `real_close` is only used once, do we need to separate it into its own function? Also `_io_refs` is no longer decremented, is that expected?

sethmlarson (commented):
Thanks for this, had a few questions:

test/with_dummyserver/test_socketlevel.py
graingert: It's a regression test, the previous code closed the socket when .close() is called twice even if a makefile is open

src/urllib3/contrib/pyopenssl.py
graingert: I'm copying the pattern from the stdlib socket

src/urllib3/contrib/securetransport.py
graingert: That's just from the original code it's not really needed

test/with_dummyserver/test_socketlevel.py
graingert: > Can we pull this into a separate test if so?

I'm not sure what you mean by this

sethmlarson: Was suggesting moving the double close to a separate test if it was only for coverage, since this is required to cause the bug we should keep it as-is, thanks for the explanation.

sethmlarson (approved):
Thanks @graingert!

PR #2976, opened 2023-04-20
Remove dummy setup.py file

graingert:


pquentin:
This is quite tempting. For context, GitHub Actions will remove Python 2.7 from the last images that had it, leaving us with no easy way to test this: https://github.com/actions/runner-images/issues/7401

Python 2.7 currently accounts for 4% of our download stats, and probably some of them use pip 9+ so maybe it's insignificant enough to do just what this pull request does.

@sethmlarson @illia-v what do you think?

.github/workflows/ci.yml
```diff
@@ -107,14 +95,6 @@ jobs:
...
           NOX_SESSION: ${{ matrix.nox-session }}
 
-      - if: ${{ matrix.nox-session != 'unsupported_setup_py' }}
-        name: "Upload artifact"
-        uses: "actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce"
-        with:
-          name: coverage-data
-          path: ".coverage.*"
-          if-no-files-found: error
-
```
illia-v, line 117: Only the `if: ${{ matrix.nox-session != 'unsupported_setup_py' }}`  line has to be removed 🙂

pquentin:
As mentioned on Discord, the changelog entry check is failing because the entry should be named `2975.removal.rst` instead of `removal.2975.rst`.

graingert:
Oops

pquentin (approved):
Thanks! Looks good to me.

I'm OK with the changelog entry because we pushed alpha versions of urllib3 2.0, but for the final release it would be best to remove it from CHANGES.rst as no final version of urllib3 was released with that dummy setup.py file.

PR #2979, opened 2023-04-21
[1.26] Fix compatible Python versions in v2-roadmap.rst

### LINKED ISSUE #2978: Inconsistent document about v2 supported CPython version
https://urllib3.readthedocs.io/en/stable/v2-roadmap.html#optimized-for-python-3-6

> In v2.0 we’ll be specifically be targeting CPython 3.6+ and PyPy 7.0+ (compatible with CPython 3.6) and dropping support Python versions 2.7 and 3.5.

https://urllib3.readthedocs.io/en/latest/v2-migration-guide.html#what-are-the-important-changes

> Python version must be 3.7 or later (previously supported Python 2.7, 3.5, and 3.6).

The aforementioned documents conflict on whether CPython 3.6 is supported or not.

zhangyoufu:
Fix #2978

### DIFF
--- a/docs/v2-roadmap.rst
+++ b/docs/v2-roadmap.rst
@@ -80,15 +80,15 @@ verification in Python or OpenSSL will be immediately
 available.
 
 
-**✨ Optimized for Python 3.6+**
+**✨ Optimized for Python 3.7+**
 --------------------------------
 
 In v2.0 we'll be specifically be targeting
-CPython 3.6+ and PyPy 7.0+ (compatible with CPython 3.6)
-and dropping support Python versions 2.7 and 3.5.
+CPython 3.7+ and PyPy 7.3.2+ (compatible with CPython 3.7)
+and dropping support Python versions 2.7 and 3.6.
 
 By dropping end-of-life Python versions we're able to optimize
-the codebase for Python 3.6+ by using new features to improve
+the codebase for Python 3.7+ by using new features to improve
 performance and reduce the amount of code that needs to be executed
 in order to support legacy versions.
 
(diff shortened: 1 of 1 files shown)


zhangyoufu:
@pquentin I'm not experienced with PyPy. I found PyPy version number from https://github.com/jwodder/pyversion-info-data. PTAL, thanks.

docs/v2-roadmap.rst
```diff
@@ -80,15 +80,15 @@ verification in Python or OpenSSL will be immediately
...
 
-**✨ Optimized for Python 3.6+**
+**✨ Optimized for Python 3.7+**
 --------------------------------
 
 In v2.0 we'll be specifically be targeting
-CPython 3.6+ and PyPy 7.0+ (compatible with CPython 3.6)
-and dropping support Python versions 2.7 and 3.5.
+CPython 3.7+ and PyPy 7.3.2+ (compatible with CPython 3.7)
+and dropping support Python versions 2.7 and 3.6.
```
pquentin, line 88:
```suggestion
CPython 3.7+ and PyPy 7.3.10+ (compatible with CPython 3.8)
and dropping support Python versions 2.7 and 3.6.
```

Since PyPy stopped supporting CPython 3.7, I think it's OK for us to do the same, even if we do test PyPy with CPython 3.7 support right now.

pquentin (approved):
#2981 will fix the build failures, but the docs here are OK: https://urllib3--2979.org.readthedocs.build/en/2979/v2-roadmap.html#optimized-for-python-3-7

pquentin:
Thank you!

PR #2984, opened 2023-04-24
[1.26] fix pyopenssl.WrappedSocket and securetransport.WrappedSocket close

graingert:
…pedSocket.close()

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -372,14 +372,15 @@ class WrappedSocket(object):
         self.connection.shutdown()
 
     def close(self):
-        if self._makefile_refs < 1:
-            try:
-                self._closed = True
-                return self.connection.close()
-            except OpenSSL.SSL.Error:
-                return
-        else:
-            self._makefile_refs -= 1
+        self._closed = True
+        if self._makefile_refs <= 0:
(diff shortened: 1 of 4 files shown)


src/urllib3/contrib/securetransport.py
```diff
@@ -323,7 +323,7 @@ def _write_callback(connection_id, data_buffer, data_length_pointer):
 _write_callback_pointer = Security.SSLWriteFunc(_write_callback)
 
 
-class WrappedSocket(object):
+class WrappedSocket:
```
graingert, line 326:
```suggestion
class WrappedSocket(object):
```

src/urllib3/contrib/pyopenssl.py
```diff
@@ -372,14 +372,15 @@ def shutdown(self):
...
 
     def close(self):
-        if self._makefile_refs < 1:
-            try:
-                self._closed = True
-                return self.connection.close()
-            except OpenSSL.SSL.Error:
-                return
-        else:
-            self._makefile_refs -= 1
```
graingert, line 382: ah this is a pain, pypy2.7 socket.close is broken in the same way as this so this will need to be conditional based on pypy2.7

graingert:
I'm going to stop working on this as it's too big a change

PR #2992, opened 2023-04-25
ensure ssl_sock is closed if assert_fingerprint or match_hostname fail

### LINKED ISSUE #2991: ResourceWarning due to _ssl_wrap_socket_and_match_hostname assert_fingerprint or assert_hostname failures not closing socket
### Subject

ResourceWarning due to _ssl_wrap_socket_and_match_hostname assert_fingerprint or assert_hostname failures not closing socket
discovered using https://github.com/urllib3/urllib3/pull/2842

### Environment

Describe your environment.

### Steps to Reproduce
```python
import sys
import urllib3
import pytest
import gc


def main():
    with urllib3.HTTPSConnectionPool(
        "google.com",
        443,
        assert_fingerprint=(
            "AA:AA:AA:AA:AA:AAAA:AA:AAAA:AA:AA:AA:AA:AA:AA:AA:AA:AA:AA"
        ),
    ) as https_pool:
        with pytest.raises(urllib3.exceptions.MaxRetryError):
            https_pool.request("GET", "/", retries=0)


if __name__ == "__main__":
    sys.exit(main())
```

### Expected Behavior

prints nothing

### Actual Behavior
prints:

```
Exception ignored in: <ssl.SSLSocket fd=3, family=10, type=1, proto=6, laddr=('redacted', 51286, 0, 0), raddr=('2a00:1450:4009:820::200e', 443, 0, 0)>
ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=10, type=1, proto=6, laddr=('redacted', 51286, 0, 0), raddr=('2a00:1450:4009:820::200e', 443, 0, 0)>
```

### Steps to Reproduce
```python
import sys
import urllib3
import pytest
import gc


def main():
    with urllib3.HTTPSConnectionPool(
        "google.com",
        443,
        assert_hostname="example.com",
    ) as https_pool:
        with pytest.raises(urllib3.exceptions.MaxRetryError):
            https_pool.request("GET", "/", retries=0)


if __name__ == "__main__":
    sys.exit(main())
```

### Expected Behavior

prints nothing

### Actual Behavior
prints:

```
Exception ignored in: <ssl.SSLSocket fd=3, family=10, type=1, proto=6, laddr=('redacted', 51286, 0, 0), raddr=('2a00:1450:4009:820::200e', 443, 0, 0)>
ResourceWarning: unclosed <ssl.SSLSocket fd=3, family=10, type=1, proto=6, laddr=('redacted', 51286, 0, 0), raddr=('2a00:1450:4009:820::200e', 443, 0, 0)>
```

### LINKED ISSUE #2999: Potential regression in 2.0.0 (commit ea96fde9450b2d0d233f0414c7be0a75148fae8b)
Previously the test `test_ssl_cert_pinning_fails` would succeed in 1.26.15, but fails by hanging in 2.0.0:

Ref: https://github.com/elastic/apm-agent-python/blob/main/tests/transports/test_urllib3.py#L256

From bisecting, the offending commit is: https://github.com/urllib3/urllib3/commit/ea96fde9450b2d0d233f0414c7be0a75148fae8b

graingert:
Closes #2991
Closes #2999 

extracted from https://github.com/urllib3/urllib3/pull/2842

this doesn't need a backport to v1.26 because the issue was introduced in https://github.com/urllib3/urllib3/commit/ea96fde9450b2d0d233f0414c7be0a75148fae8b

### DIFF
--- a/changelog/2991.bugfix.rst
+++ b/changelog/2991.bugfix.rst
@@ -1,0 +1,1 @@
+avoid a ``ResourceWarning`` if ``assert_fingerprint`` or ``match_hostname`` fail
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -784,36 +784,42 @@ def _ssl_wrap_socket_and_match_hostname(
         tls_in_tls=tls_in_tls,
     )
 
-    if assert_fingerprint:
-        _assert_fingerprint(ssl_sock.getpeercert(binary_form=True), assert_fingerprint)
-    elif (
-        context.verify_mode != ssl.CERT_NONE
-        and not context.check_hostname
-        and assert_hostname is not False
-    ):
-        cert: _TYPE_PEER_CERT_RET_DICT = ssl_sock.getpeercert()  # type: ignore[assignment]
+    try:
+        if assert_fingerprint:
+            _assert_fingerprint(
+                ssl_sock.getpeercert(binary_form=True), assert_fingerprint
+            )
+        elif (
+            context.verify_mode != ssl.CERT_NONE
+            and not context.check_hostname
+            and assert_hostname is not False
+        ):
+            cert: _TYPE_PEER_CERT_RET_DICT = ssl_sock.getpeercert()  # type: ignore[assignment]
 
-        # Need to signal to our match_hostname whether to use 'commonName' or not.
-        # If we're using our own constructed SSLContext we explicitly set 'False'
-        # because PyPy hard-codes 'True' from SSLContext.hostname_checks_common_name.
-        if default_ssl_context:
-            hostname_checks_common_name = False
-        else:
-            hostname_checks_common_name = (
-                getattr(context, "hostname_checks_common_name", False) or False
+            # Need to signal to our match_hostname whether to use 'commonName' or not.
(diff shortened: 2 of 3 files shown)


sethmlarson (approved):
LGTM, thanks @graingert for finding and fixing this!

pquentin:
This does fix the error nicely, and I did not initially understand why. But @graingert explained on Discord that "The problem is they are using a single threaded server and the socket is being kept open by a reference in the traceback". That was still cryptic to me, so here's an attempt on a more detailed explanation.

 * When `_assert_fingerprint` or `_match_hostname` fail, they raise an exception. A traceback is attached to that exception, and the traceback keeps a reference to the socket.
 * urllib3 keeps a reference to that exception to include it in the `MaxRetryError` exception it will eventually raise. This means the garbage collector will not close the socket on its own, which means the connection will be kept open.
 * pytest-httpserver and pytest-localserver are single threaded and can only accept one request at a time. So when urllib3 retries, it won't be able to open a connection, because the server is waiting for the previous try to finish. That's a deadlock!

When we disable retries, the situation is similar. The test passes, but I think pytest somehow keeps a reference to the exception, which means that the server cannot shutdown, so its fixture never finishes, so the test suite hangs too.

PR #2995, opened 2023-04-26
fix flaky test_ssl_failure_midway_through_conn

graingert:
test_ssl_failure_midway_through_conn was failing intermittently on windows in the handshake with a ConnectionResetError rather than an ssl.SSLError
eg https://github.com/urllib3/urllib3/actions/runs/4805604022/jobs/8552185456#step:5:2826 however it should never have failed in the handshake! It was failing because of a missing ca_certs kwarg and so the test was invalid


sethmlarson (approved):
LGTM, thanks!

PR #2998, opened 2023-04-28
Fix HTTPResponse.read(0) as first read call

pquentin:
HTTPResponse uses a bytes buffer since https://github.com/urllib3/urllib3/pull/2798. This buffer fails if we call get(n) and the buffer is empty, but that means that if the first call to HTTPResponse.read(n) is done with n = 0, the read() call fails. I don't think there is a good reason to do that in practice, but requests-toolbelt has a unit test that does it, which currently fails with urllib3 2.0: https://github.com/requests/toolbelt/actions/runs/4834111890/jobs/8614945073?pr=356

### DIFF
--- a/changelog/2998.bugfix.rst
+++ b/changelog/2998.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``HTTPResponse.read(0)`` call when underlying buffer is empty.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -244,7 +244,9 @@ class BytesQueueBuffer:
         self._size += len(data)
 
     def get(self, n: int) -> bytes:
-        if not self.buffer:
+        if n == 0:
+            return b""
+        elif not self.buffer:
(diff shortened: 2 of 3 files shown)


sethmlarson (dismissed):
Great find, thanks for this!

graingert:
I think the failure in

```
FAILED test/contrib/test_pyopenssl.py::TestSocketSSL::test_ssl_custom_validation_failure_terminates - AssertionError: The socket was not terminated
assert False
 +  where False = <bound method Event.wait of <threading.Event object at 0x00000255E46E6108>>(0.5)
 +    where <bound method Event.wait of <threading.Event object at 0x00000255E46E6108>> = <threading.Event object at 0x00000255E46E6108>.wait
```
is due to https://github.com/urllib3/urllib3/issues/2991

sethmlarson (approved):
LGTM, thanks Quentin!

pquentin:
The failures are unrelated, merging.

PR #3004, opened 2023-04-30
Replace flake8, isort, and pyupgrade with ruff

cclauss:
[Ruff](https://beta.ruff.rs/) supports [over 500 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,10 +100,6 @@ filterwarnings = [
     '''default:ssl NPN is deprecated, use ALPN instead:DeprecationWarning''',
 ]
 
-[tool.isort]
-profile = "black"
-add_imports = "from __future__ import annotations"
-
 [tool.mypy]
 mypy_path = "src"
 check_untyped_defs = true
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -100,10 +100,6 @@ filterwarnings = [
     '''default:ssl NPN is deprecated, use ALPN instead:DeprecationWarning''',
 ]
 
-[tool.isort]
-profile = "black"
-add_imports = "from __future__ import annotations"
```
cclauss, line 105: This is the one line that I could not figure out how to replicate with ruff.
* https://beta.ruff.rs/docs/settings/#isort

graingert: https://github.com/charliermarsh/ruff/pull/1762

cclauss: Thanks!! Fixed.

pquentin:
Hello, thanks for the pull request. We discussed this between maintainers and we have mixed feelings about this. On the plus side, the speed is nice (but linting never was a bottleneck and black is slower anyway), it's nice to get rid of setup.cfg and the amount of changes to the code base is minimal. However, this is yet another tool/[standard](https://xkcd.com/927/), and it adds a lot of configuration to maintain. Black has spoiled us here, we would like less configuration, not more :)

We'll close this for now but will reevaluate in the future.

PR #3018, opened 2023-05-05
Document common 2.0 upgrading issues

### LINKED ISSUE #3016: Add guidance for Amazon Linux, RedHat Enterprise Linux, wrt OpenSSL versions
This error has been cropping up in a few GitHub issues/comments:

```
ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2k-fips  26 Jan 2017. See: https://github.com/urllib3/urllib3/issues/2168
```

There's likely some additional guidance we can give folks that are receiving this error in our v2 migration guide:

- Create a separate section for this issue and how to mitigate it
- Have this error message verbatim in the migration guide under the section
- Mention specifically Amazon Linux and RHEL, but also other platforms we're aware of
- Recommend to users to figure out if they can upgrade their OpenSSL version or platform to get onto an OpenSSL release which is receiving security releases
- If they can't, then they need to pin `urllib3<2` until they do.

pquentin:
Closes #3016 

You can review the result here: https://urllib3--3018.org.readthedocs.build/en/3018/v2-migration-guide.html#common-upgrading-issues

### DIFF
--- a/docs/v2-migration-guide.rst
+++ b/docs/v2-migration-guide.rst
@@ -190,10 +190,10 @@ ssl module is compiled with OpenSSL 1.0.2.k-fips
   ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2k-fips  26 Jan 2017.
   See: https://github.com/urllib3/urllib3/issues/2168
 
-Remediation depends on your system.
+Remediation depends on your system:
 
-- **AWS Lambda**: Upgrade to the Python3.10 runtime as it uses OpenSSL 1.1.1.
-  Alternatively, you can use a `custom Docker image
+- **AWS Lambda**: Upgrade to the Python3.10 runtime as it uses OpenSSL 1.1.1. Alternatively, you can
+  use a `custom Docker image
   <https://aws.amazon.com/blogs/aws/new-for-aws-lambda-container-image-support/>`_ and ensure you
   use a Python build that uses OpenSSL 1.1.1 or later.
 - **Amazon Linux 2**: Upgrade to `Amazon Linux 2023
@@ -202,8 +202,8 @@ Remediation depends on your system.
   tool like pyenv.
 - **Red Hat Enterpritse Linux 7 (RHEL 7)**: Upgrade to RHEL 8 or RHEL 9.
 - **Read the Docs**: Upgrade your `configuration file to use Ubuntu 22.04
-  <https://docs.readthedocs.io/en/stable/config-file/v2.html>`_ by using `os: ubuntu-22.04` in the
-  `build` section. Feel free to use the `urllib3 configuration
+  <https://docs.readthedocs.io/en/stable/config-file/v2.html>`_ by using ``os: ubuntu-22.04`` in the
+  ``build`` section. Feel free to use the `urllib3 configuration
   <https://github.com/urllib3/urllib3/blob/2.0.0/.readthedocs.yml>`_ as an inspiration.
 
 docker.errors.dockerexception: error while fetching server api version: request() got an unexpected keyword argument 'chunked'
@@ -215,20 +215,21 @@ ImportError: cannot import name 'gaecontrib' from 'requests_toolbelt._compat'
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To be compatible with urllib3 2.0, Requests Toolbelt released version 1.0.0 without Google App
-Engine Standard Python 2.7 support. Most users that reported this issue were using the Pyrebase
-library that provides an API for the Firebase API. This library is unmaintained, but `replacements
-exist <https://github.com/thisbejim/Pyrebase/issues/435>`_.
+Engine Standard Python 2.7 support. Most users that reported this issue were using the `Pyrebase
+<https://github.com/thisbejim/Pyrebase>`_ library that provides an API for the Firebase API. This
+library is unmaintained, but `replacements exist
+<https://github.com/thisbejim/Pyrebase/issues/435>`_.
 
 ``ImportError: cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_'``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This likely happens because you're using botocore which `does not support urllib3 2.0 yet
 <https://github.com/boto/botocore/issues/2921>`_. The good news is that botocore explicitly declares
-in its dependencies that it only supports urllib3<2. Make sure to use a recent pip. That way, pip
+in its dependencies that it only supports ``urllib3<2``. Make sure to use a recent pip. That way, pip
 will install urllib3 1.26.x until botocore starts supporting urllib3 2.0.
 
 If you're deploying to an AWS environment such as Lambda or a host using Amazon Linux 2,
-you'll need to explicitly pin to `urllib3<2` in your project to ensure urllib3 2.0 isn't
+you'll need to explicitly pin to ``urllib3<2`` in your project to ensure urllib3 2.0 isn't
 brought into your environment. Otherwise, this may result in unintended side effects with
 the default boto3 installation.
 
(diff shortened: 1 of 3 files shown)


docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: text
+
+  ImportError: urllib3 v2.0 only supports OpenSSL 1.1.1+, currently the 'ssl' module is compiled with OpenSSL 1.0.2k-fips  26 Jan 2017.
+  See: https://github.com/urllib3/urllib3/issues/2168
+
+Remediation depends on your system.
+
+- **AWS Lambda**: Upgrade to the Python 3.9 or 3.10 runtimes that come OpenSSL 1.1.1.
```
sethmlarson, line 201: Is this accurate for our current understanding? Perhaps we should recommend folks use Python 3.10?

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+  Alternatively, you can use a `custom Docker image
+  <https://aws.amazon.com/blogs/aws/new-for-aws-lambda-container-image-support/>`_ and ensure you
+  use a Python build that uses OpenSSL 1.1.1 or later.
+- **Amazon Linux 2**: Upgrade to `Amazon Linux 2023
+  <https://aws.amazon.com/linux/amazon-linux-2023/>`_. Alternatively, you can install OpenSSL 1.1.1
+  on Amazon Linux 2 using ``yum install openssl11 openssl11-devel`` and then install Python with a
+  tool like pyenv.
+- **Red Hat Enterpritse Linux 7 (RHEL 7)**: Upgrade to RHEL 8 or RHEL 9.
+- **Read the Docs**: Upgrade your `configuration file to use Ubuntu 22.04
```
sethmlarson, line 210: Let's give a concrete example with the config file here (ie `build.os: ubuntu-22.04`)

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+  <https://docs.readthedocs.io/en/stable/config-file/v2.html>`_. Feel free to use the `urllib3
+  configuration <https://github.com/urllib3/urllib3/blob/2.0.0/.readthedocs.yml>`_ as an
+  inspiration.
+
+docker.errors.dockerexception: error while fetching server api version: request() got an unexpected keyword argument 'chunked'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+docker-py swaps some of the urllib3 internals in an unsupported way. The following pull request
```
sethmlarson, line 218: This guidance can be changed to "upgrade to `docker==6.1.0`" :partying_face:

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+solves the issue: https://github.com/docker/docker-py/pull/3116. Until it gets reviewed, merged and
+released, your only option is to pin urllib3 to 1.26.x using ``urllib3<2.0``.
+
+ImportError: cannot import name 'gaecontrib' from 'requests_toolbelt._compat'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To be compatible with urllib3 2.0, Requests Toolbelt released version 1.0.0 without Google App
+Engine Standard Python 2.7 support. Most users that reported this issue were using the Pyrebase
+library that provides an API for the Firebase API. This library is unmaintained, see
```
sethmlarson, line 227: nit: Let's use an named link instead of a raw URL

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+AttributeError: module 'urllib3.connectionpool' has no attribute 'VerifiedHTTPSConnection'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``VerifiedHTTPSConnection`` class has always been documented to be in the
+:mod:`~urllib3.connection` module. It used to be possible to import it from
+:mod:`~urllib3.connectionpool` but that was acccidental and is no longer possible due to a
+refactoring in urllib3 2.0.
+
+Note that the new name of this class is :class:`~urllib3.connection.HTTPSConnection`. It can be used
+starting from urllib3 1.25.9.
```
sethmlarson, line 247: Let's give a concrete example of how to use `urllib3.connection.HTTPSConnection`.

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+
+AttributeError: 'HTTPResponse' object has no attribute 'strict'
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The strict parameter is unneeded with Python 3 and should be removed. We are considering adding it
+back to help users of the unmaintained cachecontrol library, see https://github.com/urllib3/urllib3/issues/3010.
```
sethmlarson, line 253: Let's remove the note about adding it back since it'll get removed again regardless of us adding it back or not, instead we can link to the issue and if folks are interested they can read more.

In order for people to support both we can provide a code sample which does `getattr(resp, "strict", True)` noting that default behavior is now `strict=True`.

sethmlarson (commented):
Thanks for this Quentin! Here's some thoughts for you:

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
...
+library that provides an API for the Firebase API. This library is unmaintained, see
+https://github.com/thisbejim/Pyrebase/issues/435 for possible replacements.
+
+``ImportError: cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_'``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This likely happens because you're using botocore which does not support urllib3 2.0 yet:
+https://github.com/boto/botocore/issues/2921. The good news is that botocore explcitly declares in
+its dependencies that it only supports urllib3<1.27. Make sure to use a recent pip. That way, pip
+will install urllib3 1.26.x until botocore starts supporting urllib3 2.0.
```
nateprewitt, line 236: We may try adding some more info to fully scope the issue.

```suggestion
This likely happens because you're using botocore which does not support urllib3 2.0 yet:
https://github.com/boto/botocore/issues/2921. The good news is that botocore explicitly declares in
its dependencies that it only supports urllib3<1.27. Make sure to use a recent pip. That way, pip
will install urllib3 1.26.x until botocore starts supporting urllib3 2.0.

If you're deploying to an AWS environment such as Lambda or a host using Amazon Linux 2,
you'll need to explicitly pin to `urllib3<1.27` in your project to ensure urllib3 2.0 isn't
brought into your environment. Otherwise, this may result in unintended side effects with
the default boto3 installation.
```

docs/v2-migration-guide.rst
```diff
@@ -179,6 +179,78 @@ bug fixes will be shipped to the 1.26.x release stream**.
 If your organization relies on urllib3 and is interested in continuing support you can learn
 more about the `Tidelift Subscription for Enterprise <https://tidelift.com/subscription/pkg/pypi-urllib3?utm_source=pypi-urllib3&utm_medium=referral&utm_campaign=docs>`_.
 
+**🤔 Common upgrading issues**
+-------------------------------
+
+The following sections explain common issues and their remediation. If all else fails, you should
```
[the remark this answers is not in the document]
pquentin: Done in [`179acab` (#3018)](https://github.com/urllib3/urllib3/pull/3018/commits/179acab09c5e2d562de14d786a7acade5e5eef5b)

docs/v2-migration-guide.rst
pquentin: Yes, it's accurate, but I changed this to only recommend 3.10 for simplicity: [`8f87259` (#3018)](https://github.com/urllib3/urllib3/pull/3018/commits/8f87259b949028efd65ccc10b5260101921167a8)

docs/v2-migration-guide.rst
pquentin: Done in [`008c89e` (#3018)](https://github.com/urllib3/urllib3/pull/3018/commits/008c89eee4e2dfc55a2e3e729372cc7c3509fd9b)

docs/v2-migration-guide.rst
pquentin: [`df8daa1` (#3018)](https://github.com/urllib3/urllib3/pull/3018/commits/df8daa1b21fd1911b3e6f0297f95066a3d662bba)

docs/v2-migration-guide.rst
pquentin: This seems out of scope, also note that there's a link to the docs.

docs/v2-migration-guide.rst
pquentin: I removed it in [`c9839ed` (#3018)](https://github.com/urllib3/urllib3/pull/3018/commits/c9839ed4ab4d3aace2e16d19acb2e9ebab5db446).

There's no point to use `getattr` here: this parameter is no longer needed with Python 3. So if you're in the position of changing your code, you can just remove strict.

pquentin:
In addition to the review response, I have updated the advice to pin <1.27 to <2 and made a second pass, fixing typos/formatting issues. PTAL

sethmlarson (approved):
LGTM

alex0112:
> Closes #3016
> 
> You can review the result here: https://urllib3--3018.org.readthedocs.build/en/3018/v2-migration-guide.html#common-upgrading-issues

Hello! Dead link here. Should it instead go to https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html ?

Leaving here in case it's of further use to anyone else.

Issue #3008, opened 2023-05-02
Partial responses when using chunked encoding + zstd compression


grossag:
### Subject

I am testing using zstd compression with [bazel-remote](https://github.com/buchgr/bazel-remote/). As background, bazel-remote provides content-addressible storage under the /cas/ folder and storage of encoded build action data under the /ac/ folder. For the purposes of this bug report, I am only doing /cas/ downloads. So bazel-remote is used for basic PUT and GET of binary data.

Without providing the server headers to say that I support zstd compression, I don't receive chunked encoding and response.data contains the full binary data that I would expect. When I provide the server headers to say that I support zstd compression, the response from the server indicates chunked encoding and len(response.data) is always set to 1048576 bytes for files that are larger than that.

### Environment

OS Windows-10-10.0.22621-SP0
Python 3.10.11
urllib3 2.0.1

pip freeze results:

appdirs==1.4.4
certifi==2022.9.24
charset-normalizer==2.1.1
colorama==0.4.6
flake8==5.0.4
idna==3.4
mccabe==0.7.0
numpy==1.23.4
pycodestyle==2.9.1
pydiffx==1.1
pyflakes==2.5.0
python-gitlab==3.12.0
RBTools==4.0
requests==2.28.1
requests-toolbelt==0.10.1
six==1.16.0
texttable==1.6.7
tqdm==4.64.1
typing_extensions==4.4.0
urllib3==2.0.1
zstandard==0.21.0

### Steps to Reproduce

1. I deployed bazel-remote and did a PUT of a file larger than 1048576 bytes into /cas/<sha_256_hash_of_the_file_contents>
2. I ran the following code:
```
import urllib3

server_url = 'http://<server_name_redacted>:8000'
cas_url = '/cas/d1dcece4b328f7a12d86dd77cb0373496da2dd3971db2f3e3a29612f831a9738'
headers = urllib3.make_headers(accept_encoding=True)

response = urllib3.request('GET', server_url + cas_url, headers=headers)

print('Status: %d' % response.status)
print('Info: %s' % response.info())
print('Got %d back' % len(response.data))
```
### Expected Behavior

Script should print: Got 1308160 back.

### Actual Behavior

Script prints: Got 1048576 back

### Notes

If I set headers to None, the script outputs the following:
```
Status: 200
[...]

PR #3021, opened 2023-05-08
[WiP] Rewrite zstd decoder to use an API that supports multiple frames (fix issue #3008)

grossag:
Compressed zstd data is composed of one or more frames. python-zstandard has several APIs that only support decoding one frame. This is not useful for urllib3 because it will result in decompressed data that is truncated to 1048576 bytes.

This patch changes the zstd decoder to use an API that is known to be able to decompress multiple frames at a time. It uses ZstdDecompressionReader, which takes any object that fulfills the interface of io.RawBase. The approach here is to have ZstdDecoder create an io.BytesIo object, feed data to it, and have python-zstandard consume data when it can.

With this change, two zstd tests are broken:
- `test_decode_zstd_incomplete` seems to be an incorrect test because it assumes that slicing the last element will cause zstd decompression to fail, but the current compressed package doesn't behave that way. TODO: I need to change the test to create data with checksums and slice in a way that breaks the checksum validation.
- `test_chunked_decoding_zstd` is expecting an exception to be thrown but it isn't. This seems to be because the `HTTPResponse.read()` is misbehaving. The issue is that there is a loop `while len(self._decoded_buffer) < amt and data:` which drains enough data to grab headers. But this test is draining the full response buffer and thus causing `_fp` to close, so `ZstdDecoder.flush()` is never called. TODO: figure out how to fix `HTTPResponse.read()` so `ZstdDecoder.flush()` is called in this case.

This change also adds a new test `test_decode_zstd_multiple_frames` which is a work-in-progress. I'd like to find a way to generate this data programmatically instead of adding the file `text.txt.zstd`, but I haven't been able to do it so far.


grossag:
I'm going to close this because https://github.com/urllib3/urllib3/pull/3022 is a better solution than mine.

PR #3022, opened 2023-05-08
Fix multi-frame Zstandard response decoding

### LINKED ISSUE #3008: Partial responses when using chunked encoding + zstd compression

Rogdham:
Fixes #3008.

## Background on Zstandard

Data compressed with Zstandard is a sequence of frames. A frame can be of two kinds:
- Zstandard frame, containing compressed data
- Skippable frame, to be ignored by the decompressor

When decompressing some data, data from each frame should be concatenated.

[Link to the Zstandard specification](https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md)

## Multi-frame with python-zstandard

The [python-zstandard](https://python-zstandard.readthedocs.org) library used to decompress the data currently stops after the first frame. As a result, if several frames are chained in the response, the decompressed response will be partial.

This seems to be a temporary state of affairs as the author's library wants to introduce a `read_across_frames` parameter to their decompress APIs, which would default to `False` for some time before defaulting to `True`.

With the current version, I relied of the `unused_data` attribute of the `ZstdDecompressionObj` when `eof` is reached to continue with the next frame. This behaviour is specified [in the documentation](https://python-zstandard.readthedocs.io/en/latest/decompressor.html#zstddecompressionobj).

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -169,10 +169,15 @@ if zstd is not None:
         def decompress(self, data: bytes) -> bytes:
             if not data:
                 return b""
-            return self._obj.decompress(data)  # type: ignore[no-any-return]
+            data_parts = [self._obj.decompress(data)]
+            while self._obj.eof and self._obj.unused_data:
+                unused_data = self._obj.unused_data
+                self._obj = zstd.ZstdDecompressor().decompressobj()
+                data_parts.append(self._obj.decompress(unused_data))
+            return b"".join(data_parts)
 
         def flush(self) -> bytes:
-            ret = self._obj.flush()
+            ret = self._obj.flush()  # note: this is a no-op
             if not self._obj.eof:
                 raise DecodeError("Zstandard data is incomplete")
             return ret  # type: ignore[no-any-return]
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -333,6 +333,25 @@ class TestResponse:
         assert r.data == b"foo"
 
     @onlyZstd()
+    def test_decode_multiframe_zstd(self) -> None:
+        data = (
+            # Zstandard frame
+            zstd.compress(b"foo")
+            # skippable frame (must be ignored)
+            + bytes.fromhex(
+                "50 2A 4D 18"  # Magic_Number (little-endian)
+                "07 00 00 00"  # Frame_Size (little-endian)
+                "00 00 00 00 00 00 00"  # User_Data
+            )
+            # Zstandard frame
+            + zstd.compress(b"bar")
+        )
+
+        fp = BytesIO(data)
+        r = HTTPResponse(fp, headers={"content-encoding": "zstd"})
+        assert r.data == b"foobar"
+
+    @onlyZstd()
     def test_chunked_decoding_zstd(self) -> None:
         data = zstd.compress(b"foobarbaz")
 
--- a/changelog/3008.bugfix.rst
+++ b/changelog/3008.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed response decoding with Zstandard when compressed data is made of several frames.


grossag:
This solution is better than my https://github.com/urllib3/urllib3/pull/3021 . I like how you were able to create one `ZstdDecompressionObj` per frame. That's a clever way to work around the API's single frame limitation.

src/urllib3/response.py
```diff
@@ -169,10 +169,15 @@ def __init__(self) -> None:
         def decompress(self, data: bytes) -> bytes:
             if not data:
                 return b""
-            return self._obj.decompress(data)  # type: ignore[no-any-return]
+            data_parts = [self._obj.decompress(data)]
+            while self._obj.eof and self._obj.unused_data:
```
grossag, line 173: Is there any risk of an infinite loop if there is less data here than a frame's worth? Basically, is there a case where `decompress(unused_data)` results in more unused data because there is not enough data for zstd to do anything with?

src/urllib3/response.py
```diff
@@ -169,10 +169,15 @@ def __init__(self) -> None:
         def decompress(self, data: bytes) -> bytes:
             if not data:
                 return b""
-            return self._obj.decompress(data)  # type: ignore[no-any-return]
+            data_parts = [self._obj.decompress(data)]
```
grossag, line 172: I'm slightly concerned about decompress performance in this case. The current design of urllib3 seems to be to just push any data it has right to zstd and ignore `zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE`, but won't that hurt performance when downloading larger files? It's nothing new to your change, but I am wondering if there is a solution that could respect `zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE` to not call `decompress()` so often.

grossag (commented):
Clever solution! I like it, but have a few smaller questions.

src/urllib3/response.py
Rogdham: My understanding is that in that case `eof` will be `False`

Rogdham: This is confirmed with some testing:

```py
>>> compressed = zstd.compress(b'foo') + zstd.compress(b'bar')
>>> obj = zstd.ZstdDecompressor().decompressobj()

>>> obj.decompress(compressed[:10])
b'f'
>>> obj.eof
False
>>> obj.unused_data
b''

>>> obj.decompress(compressed[10:])
b'oo'
>>> obj.eof
True
>>> obj.unused_data
b'(\xb5/\xfd \x03\x19\x00\x00bar'
```

Rogdham:
Hello @grossag thank you for your review!

I did not know you were working on it since you were not assigned to the ticket and your PR was not published when I started working on it. I'm glad you like my implementation though!

src/urllib3/response.py
Rogdham: To test this I have created a 1GB file containing fully-random bytes (so that output is >1GB) and compressed it:
```
$ zstd -l data.zst
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
     1      0     954 MiB                       XXH64  data.zs
```

Then I run the following code with `pytest-benchmark`:

<details>
  <summary>Benchmark code</summary>
  
```python
import zstandard as zstd

with open("data.zst", "rb") as f:
    DATA = f.read()


def oneshot():
    obj = zstd.ZstdDecompressor().decompressobj()
    return len(obj.decompress(DATA))


def test_oneshot(benchmark):
    size = benchmark(oneshot)
    assert size == 1_000_000_000


def incremental():
    obj = zstd.ZstdDecompressor().decompressobj()
    size = 0
    for i in range(0, len(DATA), zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE):
        size += len(
            obj.decompress(DATA[i : i + zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE])
        )
    return size


def test_incremental(benchmark):
    size = benchmark(incremental)
    assert size == 1_000_000_000
```

</details>

The results show that decompressing one 1G string is ~2.5 time slower than splitting it and sending the data incrementally:

```
---------------------------------------- benchmark: 2 tests ----------------------------------------
Name (time in ms)          Min                 Max                Mean              Median          
----------------------------------------------------------------------------------------------------
test_incremental      317.5479 (1.0)      384.3817 (1.0)      358.2648 (1.0)      359.8864 (1.0)    
test_oneshot          921.7404 (2.90)     968.6641 (2.52)     949.1417 (2.65)     964.4232 (2.68)   
----------------------------------------------------------------------------------------------------
```

---

Now I leave urllib3's maintainer decide if they want me to make a quick change in the code to do so or if that's ok as it is.

One could also argue that python-zstandard should do it automatically.

Rogdham: Also @grossag I did not find much documentation about `DECOMPRESSION_RECOMMENDED_INPUT_SIZE` (nor  `ZSTD_DStreamInSize()` for that matter), feel free to share if you have some poniters!

sethmlarson:
Thank you @grossag and @Rogdham  for all your work on this issue, I'd like to invite you both to submit an Open Collective invoice of $300 upon completion if you're able to! 🙏 

I'm taking a look at this PR approach tonight.

src/urllib3/response.py
grossag: `zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE` defaults to the result of a call into zstd's function `ZSTD_DStreamInSize()`. After reading through some zstd source and also https://github.com/facebook/zstd/issues/340 , I think this value may represent the chunk size that zstd performs best with.

That said, optimizing for network download + decompress is much different than disk read + decompress. Also, the variance will be much higher in terms of network download speed. So it's hard to know how much to tune this.

The downside with delaying decompression until you have a buffer of size `DECOMPRESSION_RECOMMENDED_INPUT_SIZE` is that it defers more work into `flush()`, which as shown in https://github.com/urllib3/urllib3/pull/3021 can break tests. If we delay decompression, the normal case for smaller chunked transfers is that `flush()` will actually do the decompress operation.

So I am fine either way and will defer to @sethmlarson on whether more work is needed to delay decompression. Either way, I still think we should go with your approach. Mine has the benefit of using the only API that is known to be able to handle multiple frames now, which could slightly improve performance in that case. But I think the performance gain is lost by my approach having to use an intermediate `io.BytesIO()` which would result in an additional copy. I like how your approach doesn't need extra copies.

src/urllib3/response.py
grossag: Actually, sorry, one more question here. What if there is truncated data? Is there any case where there would repeatedly be unused data if the network connection was closed early without receiving everything? Or would it throw an exception in that case?

Rogdham: If the `eof` attribute is True and there is some `unused_data`, then it will go into the next loop. At that point a new object is instantiated, and the `unused_data` from the previous one is fed for decompression. And so on and so forth.

If you are concerned about possible infinite loop, then it will not happen because the `unused_data` part between each iteration of the loop is guarantied to strictly decrease in size with each iteration.

Now, if the data is truncated, then nothing special will happen. The part of the received data that can be decompressed will be returned, and that will be it. No exception is raised in that case, at least due to the code about Zstandard decompression; I would guess that the code from HTTP handling will detect that packets are lost or something like this and react accordingly.

Not sure if that answers your question, please feel free to clarify otherwise!

src/urllib3/response.py
Rogdham: Thank you for the pointers.

I am confused though. I though you asked about what happens when we feed **more** data than `zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE` to the decompressor, but then in your last message you speak about what happens when we feed **less** data. So here are my thoughts:

- Feeding more data: we see in the benchmark that there seems to be a penalty cost to do so (600ms per GB). I'm open to changing the code to consider that but would argue that the python-zstandard's lib should do it itself.
- Feeding less data: this case allows for the consumer of urllib3 to get the data as soon as it is ready, which would be helpful when streaming HTTP responses. I'm not sure how often this scenario would happen in real life though.

Also, you spoke about calling `flush()`, but in the case of this PR `ZstdDecompressionObj.flush` does nothing at all. I put a comment in the code for clarity, but did not remove the code calling it in case python-zstandard's API changes in the future.

grossag: Sorry! I am wondering specifically about feeding less data than `zstd.DECOMPRESSION_RECOMMENDED_INPUT_SIZE`. But as discussed in Discord, zstd frames can be any size so my concern isn't important.

My biggest concern was: what happens when we receive half of a frame's worth of data? python-zstandard's underlying call to zstd's `ZSTD_decompressStream` will return 3, meaning that there is more data to decode. But there isn't enough data to decode so we basically need to wait for more data.

I wrote some Python code where I read the full contents of a .zstd file and sliced it in half at different points. I then ran your code over it. It worked perfectly for various ways that I sliced the buffer, so I think your approach is great.

The part I didn't realize is that zstd internally remembers where it is at during a decompress operation. If there is not enough data, it will do as much as it can and return. Then your code is doing the right thing in reusing `self._obj` if it doesn't have enough data and it's mid-frame (`self._obj.eof` being False if we are mid-frame).

src/urllib3/response.py
grossag: Yes, that answers my question, thanks!

src/urllib3/response.py
Rogdham: Great! Thank you for taking the time to confirming that everything works as expected! :+1:

grossag:
This change looks good to me. I tested thousands of zstd-compressed GETs with your topic branch and confirmed that they all worked correctly.

I am not a maintainer of urllib3 so I am not sure if I should hit the Approve button. I'll defer to the maintainers on approving it. But I think it's a great improvement that will help a lot, thanks for your work on this!

pquentin (approved):
This looks great, thanks! Did you check that the test was failing without your change?

I left two small questions.

test/test_response.py
```diff
@@ -332,6 +332,21 @@ def test_decode_zstd(self) -> None:
         r = HTTPResponse(fp, headers={"content-encoding": "zstd"})
         assert r.data == b"foo"
 
+    @onlyZstd()
+    def test_decode_multiframe_zstd(self) -> None:
+        data = (
+            # Zstandard frame
+            zstd.compress(b"foo")
+            # skippable frame (must be ignored)
+            + bytes.fromhex("502A4D1810000000726f676468616d207761732068657265")
```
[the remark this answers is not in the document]
Rogdham: The magic number is in little-endian, and the last half-byte (noted `?` in the spec) can be anything, I chose `0` here but it is arbitrary. So 0x184D2A50, which encoded little-endian is `502A4D18`.

Then followed by the frame size, I chose 16, so `10000000`.

Then the user data part, I used 16 arbitrary bytes.

Maybe it would be clearer if I put some spaces in there, like one of the following:
```python
bytes.fromhex("502A4D18 10000000 726f676468616d207761732068657265")
bytes.fromhex("50 2A 4D 18    10 00 00 00    72 6f 67 64 68 61 6d 20 77 61 73 20 68 65 72 65")
```

Tell me if that makes more sense or if you want me to change it to something else.

Rogdham:
> Did you check that the test was failing without your change?

The failing test is a timeout in one of the environments only because it reached 30min; I don't think it's related to my change.

grossag:
> > Did you check that the test was failing without your change?
> 
> The failing test is a timeout in one of the environments only because it reached 30min; I don't think it's related to my change.

I think the question was whether your new zstd test fails without your current change. And I did confirm previously when reviewing your change that the new zstd test fails when run on current main top of tree. So you should be all set for that.

Rogdham:
> I think the question was whether your new zstd test fails without your current change.

Oh my bad I completely misread the question. Thank you for clarifying (and having tested it!), I have also confirmed it on my side.

src/urllib3/response.py
```diff
@@ -169,10 +169,15 @@ def __init__(self) -> None:
...
+            while self._obj.eof and self._obj.unused_data:
+                unused_data = self._obj.unused_data
+                self._obj = zstd.ZstdDecompressor().decompressobj()
+                data_parts.append(self._obj.decompress(unused_data))
+            return b"".join(data_parts)
 
         def flush(self) -> bytes:
-            ret = self._obj.flush()
+            ret = self._obj.flush()  # note: this is a no-op
```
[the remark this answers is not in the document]
pquentin: OK, the first reason makes sense, thanks.

test/test_response.py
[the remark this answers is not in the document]
pquentin: I like option 3! Can we mention additionally that magic number and frame size are little-endian? And maybe use only zeroes for user data? Otherwise it sounds like the randomness is significant.

pquentin (changes requested):
Thanks, I should have been clearer regarding the failing test. The only thing remaining is commenting the skippable frame per https://github.com/urllib3/urllib3/pull/3022#discussion_r1191039333

test/test_response.py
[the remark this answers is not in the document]
sethmlarson: I agree, let's add comments to the parts of the frame and use zeroes to denote insignificant data.

sethmlarson (changes requested):
This looks great, thanks for working on this! I had two comments:

test/test_response.py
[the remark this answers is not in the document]
Rogdham: So to test I wrote a quick TCP server to send the same exact data as request body.

<details><summary>HTTP server code</summary>
<p>

```python
import socketserver
import zstandard as zstd


data = (
    # Zstandard frame
    zstd.compress(b"foo")
    # skippable frame (must be ignored)
    + bytes.fromhex(
        "50 2A 4D 18"  # Magic_Number (little-endian)
        "07 00 00 00"  # Frame_Size (little-endian)
        "00 00 00 00 00 00 00 "  # User_Data
    )
    # Zstandard frame
    + zstd.compress(b"bar")
)


class MyTCPHandler(socketserver.BaseRequestHandler):
    def handle(self):
        f = self.request.makefile("rb")
        while f.readline().strip():
            pass  # consume request
        self.request.sendall(
            (
                b"HTTP/1.1 200 OK\r\n"
                b"Content-Encoding: zstd\r\n"
                b"Content-Type: text/plain\r\n"
                b"Content-Length: %d\r\n"
                b"\r\n"
                b"%s"
            )
            % (len(data), data)
        )


if __name__ == "__main__":
    HOST, PORT = "localhost", 9999
    with socketserver.TCPServer((HOST, PORT), MyTCPHandler) as server:
        server.serve_forever()
```

</p>
</details> 

And then queried it with curl:
```sh
$ curl --compressed http://localhost:9999/
foobar
```

However both Firefox & Chromium directly download the compressed file. It's not clear to me but it seems that [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1301878) and [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1246971) don't support zstd compression yet.

And confirmed the fix of this PR:
- with `urllib3==2.0.2`:
```python
>>> urllib3.request('GET', 'http://localhost:9999').data
b'foo'
```
- with the PR:
```python
>>> urllib3.request('GET', 'http://localhost:9999').data
b'foobar'
```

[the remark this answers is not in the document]
Rogdham: Done! I have changed the frame size to 7 so that it is clearer that it does not need to be a round number.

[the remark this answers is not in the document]
sethmlarson: curl doing the right thing is enough for me, thank you!

sethmlarson (dismissed):
LGTM, thanks @Rogdham and @grossag!

test/test_response.py
```diff
@@ -332,6 +332,25 @@ def test_decode_zstd(self) -> None:
...
+    @onlyZstd()
+    def test_decode_multiframe_zstd(self) -> None:
+        data = (
+            # Zstandard frame
+            zstd.compress(b"foo")
+            # skippable frame (must be ignored)
+            + bytes.fromhex(
+                "50 2A 4D 18"  # Magic_Number (little-endian)
+                "07 00 00 00"  # Frame_Size (little-endian)
+                "00 00 00 00 00 00 00 "  # User_Data
```
pquentin, line 344: nit: Can you please remove the trailing space after the last 00?

pquentin (dismissed):
LGTM! Thanks for iterating. And thanks everyone for the careful reviews.

test/test_response.py
Rogdham: Sorry for that, fixed!

pquentin (approved):

sethmlarson (approved):

sethmlarson:
Thanks @Rogdham and @grossag! You are both invited to apply for the $300 OpenCollective bounty. Please submit an expense to [our OpenCollective](https://opencollective.com/urllib3) if you're eligible to do so and let us know if there's anything we can do to help with that process :)

grossag:
> Thanks @Rogdham and @grossag! You are both invited to apply for the $300 OpenCollective bounty. Please submit an expense to [our OpenCollective](https://opencollective.com/urllib3) if you're eligible to do so and let us know if there's anything we can do to help with that process :)

Thanks @sethmlarson ! I'm not eligible but appreciate the offer. The fix is going to be very useful for my work anyway. Thanks @Rogdham for making it.

Rogdham:
Thank you for merging the PR, and for the bounty offer! I have sent [an expense](https://opencollective.com/urllib3/expenses/138985) on OpenCollective. Feel free to contact me if you have any questions with this piece of code in the future. Many thanks!

pquentin:
Thanks! Approved. ✅

PR #3025, opened 2023-05-11
Update mypy to 1.3.0

hramezani:


pquentin (approved):
Thanks! LGTM. I think this is the first time that we get no changes at all! Interesting.

Issue #3026, opened 2023-05-11
CI: Use "trusted publishing" to publish to PyPI


woodruffw:
(This is a CI request, but I've tried to shoehorn it into the general issue template below).

### Context

PyPI now supports "trusted publishing" with GitHub: rather than maintaining a long-lived API token, projects can generate ephemeral publishing tokens based on the OIDC credentials present in GitHub Action workflows.

`urllib3` already uses a GitHub Action for publishing to PyPI, meaning that switching to trusted publishing should be trivial: the old manual secret can be removed, and adding the `id-token: write` permission will tell the `pypa/gh-action-pypi-publish` action to use the ambient OIDC identity instead.

Relevant links:

* Blog post: https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
* Documentation: https://docs.pypi.org/trusted-publishers/

### Alternatives

Status quo! `urllib3` already uses PyPI API tokens for authentication.

### Duplicate

I don't believe so; couldn't find it in a quick closed issue search.

### Contribution

Happy to! I'll need to coordinate with a maintainer about removing the repository secret, as well as best practices for securing the publishing workflow.

sethmlarson:
Sounds good to me, I configured a trusted publisher for the urllib3 package. Can you make the change to both the `main` and `1.26.x` branches?

woodruffw:
> Sounds good to me, I configured a trusted publisher for the urllib3 package. Can you make the change to both the `main` and `1.26.x` branches?

Sure thing, can do tomorrow!

sethmlarson:
This work is complete! Thank you, and no worries on the changelog entries you're correct in your assumption :)

PR #3028, opened 2023-05-14
workflows/publish: use trusted publishing for PyPI

woodruffw:
I'll do one of these for the `1.26.x` branch as well.

See #3026.


sethmlarson (approved):
Thanks @woodruffw!

PR #3029, opened 2023-05-14
workflows/publish: use trusted publishing for PyPI

woodruffw:
Like #3028, but for the 1.26.x series.

See #3026.


sethmlarson (approved):
Thanks @woodruffw!

PR #3034, opened 2023-05-18
Added OpenGraph information to the documentation configuration

### LINKED ISSUE #1983: Add OpenGraph information to our documentation
Would be good to fill in some open graph fields for our documentation so when we link to things they show up with rich output / images rather than as a standard link. Really works nicely for Social Media, Slack, etc.

More information: https://ogp.me/

Mr-Sunglasses:
Fix #1983 Added the OpenGraph information metadata to the documentation configuration so that it looks nice when shared on the platforms supporting The Open Graph protocol.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,10 +40,10 @@ extensions = [
 
 # Open Graph metadata
 ogp_title = "urllib3 documentation"
-ogp_site_url = "https://urllib3.readthedocs.io/en/stable/"
+ogp_site_url = "https://urllib3.readthedocs.io"
 ogp_type = "website"
 ogp_image = "https://github.com/urllib3/urllib3/raw/main/docs/_static/banner_github.svg"
-ogp_description = "Python HTTP library with thread-safe connection pooling, file post support, user friendly, and more."
+ogp_description = "urllib3 is a user-friendly HTTP client library for Python."
 
 # Test code blocks only when explicitly specified
 doctest_test_doctest_blocks = ""
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,7 +35,15 @@ extensions = [
     "sphinx_copybutton",
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
+    "sphinxext.opengraph",
 ]
+
+# Open Graph metadata
+ogp_title = "urllib3 documentation"
+ogp_site_url = "https://urllib3.readthedocs.io/en/stable/"
+ogp_type = "website"
+ogp_image = "https://github.com/urllib3/urllib3/raw/main/docs/_static/banner_github.svg"
+ogp_description = "Python HTTP library with thread-safe connection pooling, file post support, user friendly, and more."
 
 # Test code blocks only when explicitly specified
 doctest_test_doctest_blocks = ""
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,3 +3,4 @@ sphinx>3.0.0
 requests
 furo
 sphinx-copybutton
+sphinxext-opengraph


sethmlarson:
@Mr-Sunglasses Thanks for this! It looks like the docs build is failing though, can you take a look?

Mr-Sunglasses:
I have Updated the `requirements.txt` Now It'll work fine.

sethmlarson (approved):

docs/conf.py
```diff
@@ -35,8 +35,16 @@
...
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
+    "sphinxext.opengraph",
 ]
 
+# Open Graph metadata
+ogp_title = "urllib3 documentation"
+ogp_site_url = "https://urllib3.readthedocs.io"
+ogp_type = "website"
+ogp_image = "https://github.com/urllib3/urllib3/raw/main/docs/_static/banner_github.svg"
```
hugovk, line 45: OpenGraph does not support SVG, use a PNG instead:

https://stackoverflow.com/a/26876671/724176

Compare https://github.com/python-pillow/Pillow/pull/5498

docs/conf.py
```diff
@@ -35,8 +35,16 @@
     "sphinx_copybutton",
     "sphinx.ext.doctest",
     "sphinx.ext.intersphinx",
+    "sphinxext.opengraph",
 ]
 
+# Open Graph metadata
+ogp_title = "urllib3 documentation"
+ogp_site_url = "https://urllib3.readthedocs.io"
```
hugovk, line 43: For some reason, the Facebook sharing debugger gets 404s for `/latest` URLs:
https://urllib3.readthedocs.io/en/latest/advanced-usage.html

https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Furllib3.readthedocs.io%2Fen%2Flatest%2Fadvanced-usage.html

I wonder if it's this line causing mixup with the canonical link. 

We don't use it for Pillow (see above) and https://github.com/wpilibsuite/sphinxext-opengraph#options says:
> Users hosting documentation on Read The Docs *do not* need to set any of the following unless custom configuration is wanted.

hugovk:
And as a bonus there's also a "theme colour" you can set: https://github.com/python/devguide/pull/953#issuecomment-1304274808

For example:

* https://github.com/python/devguide/pull/953
* https://github.com/python/cpython/pull/99931

Mr-Sunglasses:
@hugovk will fix it as soon as possible 😊

sethmlarson:
@Mr-Sunglasses Were you still up for fixing these issues? Otherwise someone else may tackle them in the next few days to get a release out.

illia-v:
I can see we refer to a `READTHEDOCS` env variable in the config file. I am not sure if it's available at the needed time, but there is [`READTHEDOCS_CANONICAL_URL`](https://docs.readthedocs.io/en/stable/reference/environment-variables.html#envvar-READTHEDOCS_CANONICAL_URL) too which may be used as `ogp_site_url = os.getenv("READTHEDOCS_CANONICAL_URL")`.

hugovk:
We can omit `ogp_site_url` entirely for RtD: https://github.com/wpilibsuite/sphinxext-opengraph#options

illia-v:
> We can omit `ogp_site_url` entirely for RtD: https://github.com/wpilibsuite/sphinxext-opengraph#options

This is nice, thanks for the info!

docs/conf.py
illia-v: I found that with the following change, it generates social cards for every page
```diff
diff --git a/docs/conf.py b/docs/conf.py
index 69f5c4dd..a2a93fdb 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,7 +42,7 @@ extensions = [
 ogp_title = "urllib3 documentation"
 ogp_site_url = "https://urllib3.readthedocs.io"
 ogp_type = "website"
-ogp_image = "https://github.com/urllib3/urllib3/raw/main/docs/_static/banner_github.svg"
+ogp_social_cards = {"image": "images/logo.png", "line_color": "#F09837"}
 ogp_description = "urllib3 is a user-friendly HTTP client library for Python."
 
 # Test code blocks only when explicitly specified
```



@sethmlarson @pquentin what do you this about using them instead of a static image?

FYI, the little icon is Sphinx's

pquentin: Sounds good to me, the logo is enough here, `urllib3` is already mentioned multiple times.

PR #3044, opened 2023-05-23
Add 1.26.16 release to CHANGES.rst on main

sethmlarson:


pquentin (commented):
Should we remove the entry from the 2.0.0 changelog then?

illia-v:
> Should we remove the entry from the 2.0.0 changelog then?

I'd apply a rule that for a release we should list changes from the closest lower version released on a date before the new release's date 🤔
So, in this case I wouldn't remove the entry.

pquentin:
So you're saying we should list it twice? That works for me

illia-v:
> So you're saying we should list it twice? That works for me

Yes, I'd list them twice 🙂

illia-v (approved):

Issue #3049, opened 2023-05-28
Python 3.12 Support


Ousret:
### Subject

Python 3.12b1 has recently been released. Few issues appeared.

### Deprecations

They are ALL indirect deprecations. that can be ignored in the meantime using:

```
'''default:ast\.([a-zA-Z]+) is deprecated and will be removed:DeprecationWarning''',
'''default:Attribute s is deprecated and will be removed in:DeprecationWarning''',
'''default:datetime\.utcfromtimestamp\(\) is deprecated and scheduled for removal in:DeprecationWarning''',
```

The first two are triggered in **pytest** and the last one in python-dateutil.

### References

- https://github.com/pytest-dev/pytest/issues/10977
- https://github.com/dateutil/dateutil/issues/1284

### Errors

Three tests fail consistently.

```
FAILED test/with_dummyserver/test_proxy_poolmanager.py::TestHTTPProxyManager::test_redirect - assert 500 == 303
 +  where 500 = <urllib3.response.HTTPResponse object at 0x7f9a271f3d30>.status
FAILED test/with_dummyserver/test_proxy_poolmanager.py::TestHTTPProxyManager::test_cross_host_redirect - Failed: DID NOT RAISE <class 'urllib3.exceptions.MaxRetryError'>
FAILED test/with_dummyserver/test_proxy_poolmanager.py::TestHTTPProxyManager::test_cross_protocol_redirect - Failed: DID NOT RAISE <class 'urllib3.exceptions.MaxRetryError'>
```

```
Traceback (most recent call last):
  File "/home/runner/work/urllib3/urllib3/test/with_dummyserver/test_proxy_poolmanager.py", line 222, in test_redirect
    assert r.status == 303
AssertionError: assert 500 == 303
 +  where 500 = <urllib3.response.HTTPResponse object at 0x7f9a271f3d30>.status
```

I suspect that the dev server crashed. I'm afraid I don't have any more details for now.

Ousret:
Mmm...

The tests passed locally. Cannot reproduce, this is up to a bad start.

```
platform linux -- Python 3.12.0b1, pytest-7.2.0, pluggy-1.0.0
test/with_dummyserver/test_proxy_poolmanager.py::TestHTTPProxyManager::test_redirect PASSED
...
```

Ubuntu 22.04 lts.

graingert:
[...]

PR #3052, opened 2023-05-30
Ignore Python 3.12.0b1 warnings and work-around gen.throw() cpython bug

graingert:
Fixes https://github.com/urllib3/urllib3/issues/3049

### DIFF
--- a/dummyserver/proxy.py
+++ b/dummyserver/proxy.py
@@ -45,29 +45,6 @@ class ProxyHandler(tornado.web.RequestHandler):
     SUPPORTED_METHODS = ["GET", "POST", "CONNECT"]  # type: ignore[assignment]
 
     async def get(self) -> None:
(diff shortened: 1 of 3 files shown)


dummyserver/proxy.py
```diff
@@ -45,29 +45,6 @@ class ProxyHandler(tornado.web.RequestHandler):
     SUPPORTED_METHODS = ["GET", "POST", "CONNECT"]  # type: ignore[assignment]
 
     async def get(self) -> None:
-        async def handle_response(response: tornado.httpclient.HTTPResponse) -> None:
-            if response.error and not isinstance(
-                response.error, tornado.httpclient.HTTPError
-            ):
-                self.set_status(500)
```
graingert, line 52: tornado already handles exceptions as 500 errors so this is redundant

pquentin (approved):
Thanks! LGTM. I like how clean this is.

Issue #3051, opened 2023-05-30
v2.x : Breaks `assert_hostname = False`


Matthew1471:
Hi,

If I upgrade to urllib3 v2.x then my disabling of hostname checking code (based upon `assert_hostname = False` as per https://github.com/urllib3/urllib3/pull/194 and https://github.com/psf/requests/pull/1405) no longer works in v2.x.

I have read that v2.x now does SAN instead of CN checking (https://urllib3.readthedocs.io/en/stable/v2-migration-guide.html#stop-verifying-commonname-in-certificates) but I don't believe the `assert_hostname = False` functionality should have also been removed / silently ignored as part of it (that param looks like it could now work with SAN hostnames instead?).

I am creating a HTTPAdapter via Requests:

```python
# Requests references passed from calling code.
from requests.adapters import HTTPAdapter, DEFAULT_POOLBLOCK

class IgnoreHostnameAdapter(HTTPAdapter):

    def init_poolmanager(self, connections, maxsize, block=DEFAULT_POOLBLOCK, **pool_kwargs):
        pool_kwargs['assert_hostname'] = False
        super(IgnoreHostnameAdapter, self).init_poolmanager(connections=connections, maxsize=maxsize, block=block, **pool_kwargs)
```

which is mounted to the session elsehwere:

```python
# Make the session verify all HTTPS requests with trust for this certficiate.
self.session.verify = 'configuration/gateway.cer'

# Requests to this host will ignore the hostname in the certificate being incorrect.
self.session.mount(self.host, IgnoreHostnameAdapter())
```

This works in v1.x (`pip install urllib3==1.26.16`) but breaks in v2.x releases (`pip install urllib3 --upgrade`).

Is there an equivalent "`subjectAltName` hostname check" disable feature in v2.x?

This is for connecting to an embedded device where the certificate details are self-signed with an incorrect hostname and I would like to pin the cert rather than disable certificate verification entirely.

Using requests' `verify=False` to disable all certificate verification of course works but I would like to continue to use requests' `session.verify` to pin the certificate (just ignore the mismatching hostname).

Thank you for your time.

pquentin:
Hello, `assert_hostname=False` is definitely supposed to work: we even have an unit test for it. I'll try to reproduce locally.

sg3-141-592:
I've been having a look this morning. I've got a clean recreation adding this additional test to `test_https.py` which is passing with `urllib3==1.26.16` and failing on `urllib3==2.0.2`. 

[...]

PR #3055, opened 2023-06-02
Fix assert_hostname=False

sg3-141-592:
Initial attempt at fixing #3051 

The issue seems to be that when `assert_hostname` is disabled, the `SSLContext` is being created with `check_hostname: True` in urllib 2.x and throwing exception. In urllib1.26.16 it's set as `False`.

This change adds two new unit tests to check that disabling assert hostname works with invalid SAN and invalid CA.

And adds a check for `assert_hostname` before calling `ssl_wrap_socket()`.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -741,6 +741,8 @@ def _ssl_wrap_socket_and_match_hostname(
         # `ssl` can't verify fingerprints or alternate hostnames
         assert_fingerprint
         or assert_hostname
+        # assert_hostname can be set to False to disable hostname checking
+        or assert_hostname is False
         # We still support OpenSSL 1.0.2, which prevents us from verifying
         # hostnames easily: https://github.com/pyca/pyopenssl/pull/933
         or ssl_.IS_PYOPENSSL
@@ -770,9 +772,6 @@ def _ssl_wrap_socket_and_match_hostname(
             normalized = normalized[: normalized.rfind("%")]
         if is_ipaddress(normalized):
             server_hostname = normalized
-
-    if not assert_hostname:
-        context.check_hostname = False
 
     ssl_sock = ssl_wrap_socket(
         sock=sock,
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -771,6 +771,9 @@ def _ssl_wrap_socket_and_match_hostname(
         if is_ipaddress(normalized):
             server_hostname = normalized
 
+    if not assert_hostname:
+        context.check_hostname = False
+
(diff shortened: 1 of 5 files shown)


pquentin (approved):
Great find, thanks! In urllib3 2.0 we rely on the standard library to check the server hostname when possible, so `context.check_hostname` now defaults to `True` instead of `False` and we did not have a test for `assert_hostname=False`, only `assert_hostname=None`.

In the interested of speed, I've made a few cosmetic changes, fixed the lint and added a changelog entry. I hope that's OK.

src/urllib3/connection.py
```diff
@@ -741,6 +741,8 @@ def _ssl_wrap_socket_and_match_hostname(
         # `ssl` can't verify fingerprints or alternate hostnames
         assert_fingerprint
         or assert_hostname
+        # assert_hostname can be set to False to disable hostname checking
+        or assert_hostname is False
```
sg3-141-592, line 745: Hey @pquentin , I think this check is redundant, did you mean to drop line :743 ?

``` python 
        or assert_hostname
        # assert_hostname can be set to False to disable hostname checking
        or assert_hostname is False
```

Also I tried this change originally to try and fix this issue and it added new failing tests

```
test_https_proxy_assert_hostname_non_matching
test_ssl_read_timeout
test_assert_hostname_closes_socket
```

I've not had a chance to review what the test failures are, as I agree where you're proposing this change is a more natural place for it.

sg3-141-592:
Hey @pquentin , sure go for it. I've left a comment on some failing tests that I can see when I've tried this change yesterday.

I'll try and have a look later to see what the difference is between setting the `check_hostname` on line :744 and just before `ssl_wrap_socket` that's causing the tests to fail.

src/urllib3/connection.py
pquentin: I don't think it's redundant:

 * `or assert_hostname` is for the case where assert_hostname is an actual domain
 * `or assert_hostname is False` specifically allows the False case.

We could change this to `assert_hostname is not None` but that felt less clear.

Regarding the failing tests, I'm not sure what happened, but some tests are unreliable, and they certainly seem to pass according to CI.

sg3-141-592: Ah, I'd missed that `assert_hostname` is defaulting to `None`. Sure, agree it's more readable. And I can't reliably re-create the failing tests. Looks good 👍

pquentin:
Thank you for your contribution! Really appreciate the help here.

sethmlarson:
Thanks @sg3-141-592  for the quick turnaround on this fix, it's really appreciated! 🚀

PR #3056, opened 2023-06-02
avoid dev dep on freezegun and python-dateutil

graingert:
freezegun depends on python-dateutil for date parsing, python-dateutil hasn't has a release since Jul 14, 2021, and it looks like it's simpler to patch time.time to test Retry instead

### DIFF
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 from test import DUMMY_POOL
 from unittest import mock
 
-import freezegun  # type: ignore[import]
 import pytest
 
 from urllib3.exceptions import (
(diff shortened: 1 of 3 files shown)


graingert:
it's important that Retry uses `time.time()` specifically rather than any of the other timers that freezegun also patches, eg `time.monotonic()`, because it's supposed to be parsing http Retry-After headers based on the current wall time

test/test_retry.py
```diff
@@ -400,7 +399,7 @@ def test_respect_retry_after_header_sleep(
     ) -> None:
         retry = Retry(respect_retry_after_header=respect_retry_after_header)
 
-        with freezegun.freeze_time("2019-06-03 11:00:00", tz_offset=0), mock.patch(
+        with mock.patch("time.time", return_value=1559559600.0), mock.patch(
```
pquentin, line 402: Is there a way to start from something more readable than 1559559600? What do you think about `datetime.fromisoformat("2019-06-03 11:00:00Z").timestamp()`?

graingert: how about constructing a datetime.datetime directly?

pquentin (approved):
Thanks! LGTM.

pquentin:
Wait no the failures are real. Can you please fix the tests?

sethmlarson (approved):

pquentin (approved):

PR #3063, opened 2023-06-07
Encode string bodies to Latin-1 isntead of UTF-8

pquentin:
Relates #3053 by fixing the immediate regression. We may introduce warnings in the future to help users not rely on this behavior. The changes are split in 3 commits (but squashing is fine if we merge this):

1. I've changed the default encoding of simple bodies from UTF-8 to Latin-1, and made the code more explicit and tested
2. Even though the formal name of the encoding is ISO-8859-1, we used `latin-1` in enough places that I preferred to be consistent and updated the remaining occurrences of `iso-8859-1` to `latin-1`.
3. Chunked encoding already used Latin-1, but I've changed the test to ensure it stays that way in the future.

### DIFF
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -213,7 +213,7 @@ def body_to_chunks(
 
     # Bytes or strings become bytes
     elif isinstance(body, (str, bytes)):
-        chunks = (to_bytes(body),)
+        chunks = (to_bytes(body, "latin-1"),)
         content_length = len(chunks[0])
 
     # File-like object, TODO: use seek() and tell() for length?
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -227,7 +227,7 @@ def body_to_chunks(
                 if not datablock:
                     break
                 if encode:
-                    datablock = datablock.encode("iso-8859-1")
+                    datablock = datablock.encode("latin-1")
                 yield datablock
 
         chunks = chunk_readable()
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -1089,7 +1089,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             assert request_headers["User-Agent"] == "test header"
 
     @pytest.mark.parametrize(
-        "user_agent", ["Schönefeld/1.18.0", "Schönefeld/1.18.0".encode("iso-8859-1")]
+        "user_agent", ["Schönefeld/1.18.0", "Schönefeld/1.18.0".encode("latin-1")]
     )
     def test_user_agent_non_ascii_user_agent(self, user_agent: str) -> None:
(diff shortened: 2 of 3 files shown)


test/with_dummyserver/test_chunked_transfer.py
```diff
@@ -65,7 +65,7 @@ def _test_body(self, data: bytes | str | None) -> None:
 
             assert b"Transfer-Encoding: chunked" in header.split(b"\r\n")
             if data:
-                bdata = data if isinstance(data, bytes) else data.encode("utf-8")
+                bdata = data if isinstance(data, bytes) else data.encode("latin-1")
```
pquentin, line 68: This is actually suspicious because 1.26.x used utf-8 here:

https://github.com/urllib3/urllib3/blob/3c0148048a523325819377b23fc67f8d46afc3aa/test/with_dummyserver/test_chunked_transfer.py#L53-L75

sethmlarson: Weird, but there was no test case for non-ASCII characters?

illia-v: There are a few non-ASCII characters on line 75

illia-v: And changing `data.encode("utf-8")` to `data.encode("latin-1")` breaks the test in 1.26.x
```
test_unicode_body failed; it passed 0 out of the required 1 times.
        <class 'AssertionError'>
        assert ((b'\r\n' + b'thisshouldbeonechunk\r\n\xe4\xf6\xfc\xdf') + b'\r\n') in b'1e\r\nthisshouldbeonechunk\r\n\xc3\xa4\xc3\xb6\xc3\xbc\xc3\x9f\r\n0\r\n\r\n'
```

illia-v: Related encoding happens on this line in 1.26.x
https://github.com/urllib3/urllib3/blob/57181d6ea910ac7cb2ff83345d9e5e0eb816a0d0/src/urllib3/connection.py#L274

pquentin: Thanks, good find! So we were encoding chunks with UTF-8 and other bodies with Latin-1, and 2.0 reversed that unintentionally. I'll fix that too.

sethmlarson:
Adding my two cents, if it's encoding as UTF-8 and the world isn't broken we may want to call this a "feature" and encode every body as UTF-8. That's what the world is using nowadays anyways, so we're likely to be causing more issues by using latin-1 instead?

pquentin:
I like the idea, but note that chunks are encoded as UTF-8 in 1.26.x but encoded as Latin-1 in 2.x. So we should only fix that part, and encoding chunks as UTF-8?

sethmlarson:
@pquentin Oh gotcha, yeah let's UTF-8 all the things!

skydiver1980 (approved):

skydiver1980 (approved):

PR #3064, opened 2023-06-08
Fix Open Graph metadata in docs

illia-v:
#3034 introduced the metadata but `og:url` was not matching real versioned ones.
This PR deletes the `ogp_site_url` config as it is not needed on RtD (thanks @hugovk for [the tip](https://github.com/urllib3/urllib3/pull/3034#issuecomment-1579937636).)

Also, the SVG image is not supported for `og:image`, I am replacing it with social cards.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,6 +72,9 @@ exclude_patterns = ["_build"]
 # The name of the Pygments (syntax highlighting) style to use.
 pygments_style = "friendly"
 
+# The base URL with a proper language and version.
+html_baseurl = os.environ.get("READTHEDOCS_CANONICAL_URL", "/")
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 html_theme = "furo"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,7 +41,7 @@ extensions = [
 # Open Graph metadata
 ogp_title = "urllib3 documentation"
 ogp_type = "website"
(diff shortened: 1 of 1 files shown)


illia-v:
https://urllib3--3064.org.readthedocs.build/en/3064/ – `og:image` leads to a nonexistent image.

---

It turns up sphinxext-opengraph uses canonical URLs for `og:url` and `og:image` which always begin with `https://urllib3.readthedocs.io/en/stable/` for urllib3 [according to RtD](https://docs.readthedocs.io/en/stable/canonical-urls.html#implementation) and in reality.

This may conform to the Open Graph protocol, but brings some uncertainty for situations when a page is not available in the stable version yet.

Also, the generated social cards have a hash in filenames, so filenames in other versions may be different from ones in "stable" and `og:image` will lead to a nonexistent file.

illia-v:
There is a slightly different but similar existing issue
https://urllib3.readthedocs.io/en/1.26.16/v2-roadmap.html provides `<link rel="canonical" href="https://urllib3.readthedocs.io/en/stable/v2-roadmap.html">`, but https://urllib3.readthedocs.io/en/stable/v2-roadmap.html does not exist

illia-v:
I found a solution in https://github.com/readthedocs/readthedocs.org/pull/10224, the links are correct now, check https://www.opengraph.xyz/url/https%3A%2F%2Furllib3--3064.org.readthedocs.build%2Fen%2F3064%2Fv2-migration-guide.html

sethmlarson (approved):
LGTM, thanks @illia-v!

PR #3067, opened 2023-06-11
Fix JSON example in user guide

pamelafox:
The current example in the docs does not work, since it uses 'body' not 'json'. After a close reading, I tried using 'json' instead, and it worked :)
My PR rewords the text, fixes the example, and provides a fuller example of the output (though still compressed slightly).


pamelafox:
I do not think this warrants a changelog entry, but I don't think I can add the skip changelog label myself.

PR #3070, opened 2023-06-13
Emit insecure proxy event when using unverified proxies

### LINKED ISSUE #2791: Emit a warning when proxy_is_verified is False
Currently we have `InsecureRequestWarning` which is emitted when an HTTPSConnection isn't verified (for example, when verification is disabled). Now that we have `HTTPConnection.proxy_is_verified` we can detect when the `proxy.scheme` is `https` but the proxy connection isn't verified.

Perhaps the warning should be something like `InsecureProxyWarning`?

tushar5526:
Fixes #2791


illia-v:
@tushar5526 thanks for the change! Could you please add a test similar to `test.with_dummyserver.test_no_ssl.TestHTTPSWithoutSSL` and a changelog entry?

src/urllib3/connectionpool.py
```diff
@@ -1091,7 +1092,22 @@ def _validate_conn(self, conn: BaseHTTPConnection) -> None:
...
+            conn.proxy
+            and conn.proxy.scheme == "https"
+            and conn.proxy_is_verified is False
+        ):
+            warnings.warn(
+                (
+                    f"Unverified HTTPS request is being made using proxy host '{conn.host}'. "
+                    "Adding certificate verification is strongly advised. See: "
+                    "https://urllib3.readthedocs.io/en/latest/advanced-usage.html"
+                    "#tls-warnings"
```
illia-v, line 1105: Shouldn't it contain `conn.proxy.host` instead of `conn.host`?

It looks like a docs section at that link needs to mention `InsecureProxyWarning` as it does for `InsecureRequestWarning`.

tushar5526: Yes, thanks for the catch. I will update the docs.

src/urllib3/connectionpool.py
```diff
@@ -1091,7 +1092,22 @@ def _validate_conn(self, conn: BaseHTTPConnection) -> None:
...
+                (
+                    f"Unverified HTTPS request is being made using proxy host '{conn.proxy.host}'. "
+                    "Adding certificate verification is strongly advised. See: "
+                    "https://urllib3.readthedocs.io/en/latest/advanced-usage.html"
+                    "#tls-warnings"
+                ),
+                InsecureProxyWarning,
+            )
+
+        elif not conn.is_verified:
```
sethmlarson, line 1110: Since both can be true, can we change this `elif` to an `if` and add a testcase for both warnings being emitted in one connection?

test/with_dummyserver/test_no_ssl.py
```diff
@@ -21,13 +21,22 @@ def test_simple(self) -> None:
...
                     pool.request("GET", "/")
                 except urllib3.exceptions.SSLError as e:
                     assert "SSL module is not available" in str(e)
+
+    def test_simple_proxy(self) -> None:
+        https_proxy_url = f"https://{self.proxy_host}:{int(self.https_proxy_port)}"
+        with urllib3.ProxyManager(https_proxy_url, cert_reqs="NONE") as pool:
+            with pytest.warns(InsecureProxyWarning):
+                try:
+                    pool.request("GET", "/")
```
sethmlarson, line 40: `/` isn't a valid URL for `ProxyManager`, we should use an absolute URL instead.

test/with_dummyserver/test_no_ssl.py
```diff
@@ -21,13 +21,22 @@ def test_simple(self) -> None:
...
             with pytest.warns(InsecureRequestWarning):
                 try:
                     pool.request("GET", "/")
                 except urllib3.exceptions.SSLError as e:
                     assert "SSL module is not available" in str(e)
+
+    def test_simple_proxy(self) -> None:
+        https_proxy_url = f"https://{self.proxy_host}:{int(self.https_proxy_port)}"
+        with urllib3.ProxyManager(https_proxy_url, cert_reqs="NONE") as pool:
+            with pytest.warns(InsecureProxyWarning):
```
sethmlarson, line 38: Can we make an assertion about the message of `InsecureProxyWarning` (including it containing the right host) as well?

test/with_dummyserver/test_no_ssl.py
```diff
@@ -21,13 +21,22 @@ def test_simple(self) -> None:
...
-            self.host, self.port, cert_reqs="NONE"
+            self.https_host, self.https_port, cert_reqs="NONE"
         ) as pool:
             with pytest.warns(InsecureRequestWarning):
                 try:
                     pool.request("GET", "/")
                 except urllib3.exceptions.SSLError as e:
                     assert "SSL module is not available" in str(e)
+
+    def test_simple_proxy(self) -> None:
```
sethmlarson, line 35: It would be good to test both HTTPS->HTTP and HTTPS->HTTPS proxy scenarios.

test/with_dummyserver/test_no_ssl.py
```diff
@@ -21,13 +21,22 @@ def test_simple(self) -> None:
             assert r.status == 200, r.data
 
 
-class TestHTTPSWithoutSSL(HTTPSDummyServerTestCase, TestWithoutSSL):
+class TestHTTPSWithoutSSL(HTTPDummyProxyTestCase, TestWithoutSSL):
```
sethmlarson, line 24: You might want to add some assertions to existing HTTPS proxy tests as well, specifically it'd be good to assert that the `InsecureProxyWarning` isn't raised when using the proper configuration for secure proxies.

sethmlarson (changes requested):
This is an excellent start, thank you! :rocket: Here's a few suggestions I can think of:

src/urllib3/connectionpool.py
tushar5526: There seems to be something off when I am using a proxy. The following snippet will raise both `InsecureRequestWarning` and `InsecureProxyWarning` after I change the *elif to if*. 

```
import urllib3
proxy = urrlib3.ProxyManager('https://localhost:8443', cert_reqs='NONE')
res = proxy.request('GET', 'https://www.google.com')
```
Outputs the following. 

```
/home/tushar55/projects/urllib3/src/urllib3/connectionpool.py:1100: InsecureProxyWarning: Unverified HTTPS request is being made using proxy host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(
/home/tushar55/projects/urllib3/src/urllib3/connectionpool.py:1111: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(

```

Both the conn.host and conn.proxy.host is set to the same value.  `conn._tunnel_host` is set to `www.google.com` 

I suspect this line is resetting the https://github.com/urllib3/urllib3/blob/main/src/urllib3/connectionpool.py#L1062 updates the host and port when proxy is being used. What is the intended behaviour here ?

tushar5526: A few more examples that I tested out locally after setting up a simple python http server on port 8000 using `python -m http.server`

```
import urllib3
proxy = urllib3.ProxyManager('https://localhost:8443', ca_certs='/home/tushar55/projects/urllib3/dummyserver/certs/cacert.pem', cert_reqs='NONE')
res = proxy.request('GET', 'http://127.0.0.1:8000')
print(res.status, res.data)
```

This raises the following error - 
```
/home/tushar55/projects/urllib3/src/urllib3/connectionpool.py:1146: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhos
t'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#tls-warnings
  warnings.warn(
```

sethmlarson: This is definitely confusing, but in the case of an HTTPS origin domain we use CONNECT tunneling so the actual destination host we want to connect to is `HTTPConnection._tunnel_host`. There might need to be some modifications to a warning message in order to pick the right host to use. Hopefully this helps?

tushar5526: Yeah I think it's clear now. So, if certificate verification is disabled for a proxy manager then it should throw both the `InsecureProxy` and `InsecureRequest` warnings, right?

```
proxy = urllib3.ProxyManager('https://localhost:8443', cert_reqs='NONE')
res = proxy.request('GET', 'https://localhost:8000')
```

and for normal PoolManger, it should just throw `InsecureRequest` warning. 

```
pool = urllib3.PoolManager(cert_reqs='NONE')
res = pool.request('GET', 'https://localhost:8000')
```

I will update the `InsecureRequest` warning's message to `host` or `_tunnel_host` depending upon whether we are using a proxy or not.

test/with_dummyserver/test_no_ssl.py
tushar5526: What is the expected HTTPS -> HTTP proxy scenario? Currently, it throws an InsecureRequestWarning as `conn.is_verified` is set to `False` while using HTTPs->HTTP connection. Should it only throw `InsecureProxyWarning` ?

illia-v: If I understand correctly, this Seth's suggestion is related to https://github.com/urllib3/urllib3/pull/3070#discussion_r1232951507.
You change the message for `InsecureRequestWarning`, it would be good to test what host is included in the message when a proxy is used as well as changing `test_simple` to check what host is used there.

test/with_dummyserver/test_no_ssl.py
tushar5526: Using https://urllib3.readthedocs.io now for tests.

test/with_dummyserver/test_no_ssl.py
tushar5526: Added checks in the recent commits.

test/with_dummyserver/test_no_ssl.py
tushar5526: Thanks, I have added them.

tushar5526:
Hi @sethmlarson , it seems that `proxy_is_verified` is never set for `HTTPs to HTTP` proxies.  `proxy_is_verified` is set only at https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L694 and the function `_connect_tls_proxy` containing this line is called only when we are creating a HTTP tunnel https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L672. We would have to set the value for `proxy_is_verified` for scenarios when we are not using a `TLS-in-TLS tunnel` and instead setting up a TLS connection for `HTTPs to HTTP` case.

illia-v (commented):
@tushar5526 interesting, it does look like a bug that HTTPS proxy with a verified certificate + HTTP destination sets `is_verified=True` and `proxy_is_verified=None`. I'd expect `is_verified=False` and `proxy_is_verified=True` according to https://github.com/urllib3/urllib3/blob/main/notes/connection-lifecycle.md

tushar5526:
> @tushar5526 interesting, it does look like a bug that HTTPS proxy with a verified certificate + HTTP destination sets `is_verified=True` and `proxy_is_verified=None`. I'd expect `is_verified=False` and `proxy_is_verified=True` according to https://github.com/urllib3/urllib3/blob/main/notes/connection-lifecycle.md

Yes, thanks for confirming. This has been bugging me for a long time. Should I open a new issue? 

The tests for this PR is dependent on the correct behaviour of `proxy_is_verifed` and `is_verified`

illia-v:
@tushar5526 yes, an issue for this would be really appreciated

skydiver1980 (approved):

skydiver1980 (approved):

PR #3083, opened 2023-06-30
Update mypy to 1.4.1

hramezani:


pquentin (approved):
Another `type: ignore` bites the dust!

PR #3094, opened 2023-07-24
Update dev/mypy dependencies monthly

pquentin:


pquentin:
One blocker here is that pip-compile does not have great cross-version support, so it can't tell that the memray dependencies such as rich or markdown-it-py are Python 3.8+. There's not great solution here I think apart from dropping Python 3.7 support ourselves.

dev-requirements.txt
```diff
@@ -1,13 +1,74 @@
...
+    # via
+    #   click-default-group
+    #   towncrier
+click-default-group==1.2.2
+    # via towncrier
 coverage==7.0.4
-tornado==6.2
-PySocks==1.7.1
+    # via -r dev-requirements.in
+cryptography==41.0.2 ; implementation_name != "pypy" or implementation_version >= "7.3.10"
```
sethmlarson, line 17: Looks like cryptography can't be split into separate versions with pip-compile :/

PR #3096, opened 2023-07-26
Debug logs for _new_conn has wrong type format

aDrongo:
Fixed bug in _new_conn where String Formatting was calling string instead of digit for port.

### DIFF
--- a/changelog/3097.bugfix.rst
+++ b/changelog/3097.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed bug in _new_conn where String Formatting was calling string instead of digit for port.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -242,10 +242,10 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         """
         self.num_connections += 1
         log.debug(
-            "Starting new HTTP connection (%d): %s:%s",
+            "Starting new HTTP connection (%d): %s:%d",
             self.num_connections,
             self.host,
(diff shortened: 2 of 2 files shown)


illia-v (commented):
The case is interesting, it looks like there are a few more cases in the file where the port is logged as `%s`.

However, the solution may be not so easy, it should be backward compatible for users who provide a port as a string. For example, the following code works fine
```python
import urllib3

with urllib3.HTTPSConnectionPool("example.com", "443") as pool:
    response = pool.request("GET", "/")
    print(response.status)
```

And ideally, the PR should include unit tests 🙂

aDrongo:
Thanks for the review, I'll work on some unit tests that would catch this.

We could instead convert the argument to string, that should be safer then converting to int/digit
```
        log.debug(
            "Starting new HTTP connection (%d): %s:%s",
            self.num_connections,
            self.host,
            str(self.port) or "80",
        )
```

sigmavirus24:
I would suggest something that handles this for us. For example, if we're logging the port like this in multiple places, we may also need to handle the case where someone provides the port as bytes because `str(b'80')` will give you a very wrong log string.

I'd argue for a private property that takes `self.port` and converts it to a string. Then replace usage of `self.port` in the logging with the private property

aDrongo:
I'm actually going to close this, after more testing it seems that using record.msg = record.getMessage() is not compatible with this tuple style string formatting, it is not to do with anything different urllib3 is doing, the debug formating urllib3 uses works fine. Thank you.

PR #3098, opened 2023-07-28
Remove Botocore patch for urllib3 2.x

nateprewitt:
Hey there 👋 

This PR will remove the last Botocore patch for urllib3 2.x now that we've modified the 100-continue behavior in the latest release. I wasn't sure if there was a preference on leaving the scaffolding code for future patches, so I made the minimal change. If you'd like me to cleanup lines 101-102 as well, I'm happy to do that.

Version bump timeline for Botocore's urllib3 dependency is still TBD, but this will unblock Linux distros and users who want to bypass pip's warnings for moving to urllib3 2.x.


nateprewitt:
It looks like mypy isn't happy about the lack of specificity with the empty list so it's likely cleaner to just remove the two lines in the noxfile.

pquentin (approved):
Thank you for this! LGTM.

PR #3060, opened 2023-06-03
fixing [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE]

elminson:
Related to this issue 
https://github.com/urllib3/urllib3/issues/3059

PR #3107, opened 2023-08-09
Fix the use of default OpenSSL ciphers

### LINKED ISSUE #3100: Behavior change: v2.x raises SSLEOFError with Python3.10 where v2.x does not with Python3.8
### Subject

Setup: Using requests library to issue GET requests to a non-public web server over https. The web server is third-party.
The GET requests work fine using Python3.8, openssl [1.1.1f|3.0.2] and urllib3 2.x
The GET requests, however, raise following error when using the same context with only Python changed to 3.10:

`SSLError: HTTPSConnectionPool(host='10.1.1.32', port=443): Max retries exceeded with url: /netio.json (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:1007)')))`

### Environment

Reproduced on Ubuntu 20.04 LTS and 22.04 LTS (openssl 1.1.1f and openssl 3.0.2)

Combination of libs that raise the error (Ubuntu 22.04 LTS):
```
OS Linux-5.4.0-131-generic-x86_64-with-glibc2.35
Python 3.10.12
OpenSSL 3.0.2 15 Mar 2022
urllib3 2.0.4
```
Combination of libs that do **not** raise the error (Ubuntu 22.04 LTS):
```
OS Linux-5.4.0-131-generic-x86_64-with-glibc2.35
Python 3.10.12
OpenSSL 3.0.2 15 Mar 2022
urllib3 1.26.15
```
Difference between the two combinations:
`pip install --upgrade --force-reinstall urllib3==1.26.15`

Combination of libs that do **not** raise the error as well (Ubuntu 20.04 LTS):
```
OS Linux-5.4.0-131-generic-x86_64-with-glibc2.29
Python 3.8.10
OpenSSL 1.1.1f  31 Mar 2020
urllib3 2.0.4
```

### Steps to Reproduce

It's hard to reproduce publicly because the web server is not public.
But my findings lead to the SSL_OP_IGNORE_UNEXPECTED_EOF flag.

### Expected Behavior

v2.x shall behave the same way with Python 3.10 as it does with Python 3.8 regarding the error described above (i. e. not raising any SSLEOFError)

### Actual Behavior

v2.x in combination with Python 3.10 raises SSLEOFError
v2.x in combination with Python 3.8 does not raise SSLEOFError

pquentin:
Closes #3100

### DIFF
--- a/changelog/3100.bugfix.rst
+++ b/changelog/3100.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the use of default OpenSSL ciphers as we were relying on CPython defaults by accident.
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -822,7 +822,8 @@ class SecureTransportContext:
         return self.set_default_verify_paths()
 
     def set_ciphers(self, ciphers: typing.Any) -> None:
-        raise ValueError("SecureTransport doesn't support custom cipher strings")
+        if ciphers != "DEFAULT":
+            raise ValueError("SecureTransport doesn't support custom cipher strings")
 
     def load_verify_locations(
         self,
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -294,10 +294,13 @@ def create_urllib3_context(
     if ssl_maximum_version is not None:
         context.maximum_version = ssl_maximum_version
 
-    # Unless we're given ciphers defer to either system ciphers in
-    # the case of OpenSSL 1.1.1+ or use our own secure default ciphers.
+    # If given explicit ciphers, respect that. Otherwise, defer to the system default
+    # ciphers as they can be controlled by users. We ask for `DEFAULT` explicitly to
+    # avoid using CPython's default cipher string
     if ciphers:
         context.set_ciphers(ciphers)
+    else:
+        context.set_ciphers("DEFAULT")
 
     # Setting the default here, as we may have no ssl module on import
     cert_reqs = ssl.CERT_REQUIRED if cert_reqs is None else cert_reqs
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -136,7 +136,7 @@ class TestSSL:
 
         ssl_.create_urllib3_context()
 
-        context.set_ciphers.assert_not_called()
+        context.set_ciphers.assert_called_once_with("DEFAULT")
 
     @pytest.mark.parametrize(
         "kwargs",


sigmavirus24 (approved):
Thanks for debugging this @pquentin

sethmlarson (approved):
Thanks @pquentin and @illia-v!

pquentin:
Thanks for the reviews. So the reason I have not merged yet is that I first thought that `DEFAULT` would be the system default, to allow users to control the ciphers, but I have not managed to do that on Fedora or Ubuntu. So this still needs some investigation.

sshishov:
Am I missing something here? This MR is almost completely the same as https://github.com/urllib3/urllib3/pull/3060
And where @pquentin rejected the change. Did we discover something new? Like defaults from Cython?

sethmlarson:
@sshishov My understanding is "DEFAULT" is different than the system configured default, which is what we want. However, CPython has started setting a default which means that we're unable to revert to the true system default.

elminson:
> Am I missing something here? This MR is almost completely the same as #3060 And where @pquentin rejected the change. Did we discover something new? Like defaults from Cython?
🤷‍♂️ @sshishov had same question :)

skydiver1980 (approved):

pquentin:
Sorry for not updating this earlier. The reason using "DEFAULT" does not work is that [CPython sets its own ciphers](https://github.com/python/cpython/blob/786cac0c64dc156dfee817e87f15ae56b7e3ed00/Modules/_ssl.c#L164-L181) in a way that user code cannot control, using Hynek's list (the one we use in 1.26.x):

```C
#elif PY_SSL_DEFAULT_CIPHERS == 1
/* Python custom selection of sensible cipher suites
 * @SECLEVEL=2: security level 2 with 112 bits minimum security (e.g. 2048 bits RSA key)
 * ECDH+*: enable ephemeral elliptic curve Diffie-Hellman
 * DHE+*: fallback to ephemeral finite field Diffie-Hellman
 * encryption order: AES AEAD (GCM), ChaCha AEAD, AES CBC
 * !aNULL:!eNULL: really no NULL ciphers
 * !aDSS: no authentication with discrete logarithm DSA algorithm
 * !SHA1: no weak SHA1 MAC
 * !AESCCM: no CCM mode, it's uncommon and slow
 *
 * Based on Hynek's excellent blog post (update 2021-02-11)
 * https://hynek.me/articles/hardening-your-web-servers-ssl-ciphers/
 */
  #define PY_SSL_DEFAULT_CIPHER_STRING "@SECLEVEL=2:ECDH+AESGCM:ECDH+CHACHA20:ECDH+AES:DHE+AES:!aNULL:!eNULL:!aDSS:!SHA1:!AESCCM"
  #ifndef PY_SSL_MIN_PROTOCOL
    #define PY_SSL_MIN_PROTOCOL TLS1_2_VERSION
  #endif
```

Ideally it should stop doing that, but before asking we should probably look at actual defaults in recent versions to see how good they are. In the meantime, https://quentin.pradet.me/blog/fixing-sslv3_alert_handshake_failure-with-urllib3-20.html continues to be the official guidance to fix handshake failures in urllib3. Closing.

PR #3112, opened 2023-08-27
Attempting Happy Eyes Support

sg3-141-592:
Very WIP, not ready for review


pquentin:
Did you mean Happy Eyeballs instead of Happy Eyes? #797 

I realize you say it's not ready for review and I appreciate the work you're putting into this, but are you confident that this is possible to do correctly without concurrency? https://www.youtube.com/watch?v=oLkfnc_UMcE also convinced me that concurrency was not enough for this, we need structured concurrency.

I also realize there's a bounty on it, but as you can see many approaches have been tried over the year without being merged, so I'm trying to make sure that you don't waste too much of your time. Would you mind *describing* your approach first? Also @sethmlarson probably has a lot of thoughts given https://github.com/sethmlarson/rfc6555.

sg3-141-592:
Hey there @pquentin , yes Happy Eyeballs. Typo by me.

Thanks for having a look. I would say I'm not trying anything new that hasn't been discussed on #797 . I've been walking through other peoples attempts, recreating and fixing unit tests.

That Trio presentation is a really impressive, it's the perfect framework to meet all these subtle Happy Eyeballs requirements like the timeouts and ordering issues. And it looks a lot more readable than asyncio.

I did manage to get the unit tests passing in Ubuntu with a non-blocking implementation, but I'm currently stuck on lots of subtle differences in Windows and Linux socket handling behaviour. Things like you can use `getpeername()` in Ubuntu to test if a socket is connected but that doesn't work consistently in Windows.

I think I agree that Happy Eyeballs in urllib3 can't be done well without concurrency after spending some time on it. The standard assumes you have async or threads available. For example on DNS resolution sticks out [RFC8305 - Happy Eyeballs Version 2](https://datatracker.ietf.org/doc/html/rfc8305)

> 
> Implementations SHOULD NOT wait for both families of answers to
   return before attempting connection establishment.
>
> ... the client SHOULD treat
   DNS resolution as asynchronous.  Note that if the platform does not
   offer an asynchronous DNS API, this behavior can be simulated by
   making two separate synchronous queries on different threads, one per
   address family.

My other worry is that this methodology may not deliver a performance improvement with the pseudo async we're doing with the non-blocking sockets and polling them. So it'd need some performance tests that show the new approach is significantly faster.

I think I'll stop work on this soon. I've learned a bunch about sockets in the meanwhile which is useful for some other projects I have so not wasted effort 🙂

pquentin:
Thank you @sg3-141-592! Closing then.

PR #3115, opened 2023-08-29
Configure dependabot to set the "Skip Changelog" label to PRs

illia-v:


pquentin (dismissed):
Thanks! LGTM.

sethmlarson (dismissed):
Great idea, thanks Illia!

illia-v:
I found out that the defined labels override the default ones, so I added the latter to the list too

sethmlarson:
Tbh I'm actually okay if we don't have those extra labels, I haven't used them for anything productive. Does anyone else find them useful?

illia-v:
> Tbh I'm actually okay if we don't have those extra labels, I haven't used them for anything productive. Does anyone else find them useful?

Not much, but color of the `dependencies` label helps me distinguish dependabot's PRs from others

sethmlarson (approved):

Issue #3117, opened 2023-09-05
Redundant conditional in response module


Rodrigo-Novas:
### Context

Idk if it fits as a feature but in the response module there is a [conditional](https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L882C1-L885C33) that evaluates the value of the amt attribute, this conditional has an elif that is redundant and could be reduced to a single if.

Change:

``` python
if amt is None:
    flush_decoder = True
elif amt != 0 and not data:
    flush_decoder = True
```

to

``` python
if amt is None or (amt != 0 and not data):
    flush_decoder = True
```

PR #3118, opened 2023-09-05
Refactor double conditional into one

Rodrigo-Novas:
Related to: https://github.com/urllib3/urllib3/issues/3117

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -878,9 +878,7 @@ class HTTPResponse(BaseHTTPResponse):
 
         data = self._raw_read(amt)
 
-        flush_decoder = False
-        if amt is None or (amt != 0 and not data):
-            flush_decoder = True
+        flush_decoder = True if amt is None or (amt != 0 and not data) else False
 
         if not data and len(self._decoded_buffer) == 0:
             return data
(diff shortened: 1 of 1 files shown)


Rodrigo-Novas:
I don't know if this small change warrants a changelog entry.

pquentin:
Well we could go one step further and write `flush_decoder = amt is None or (amt != 0 and not data)`. I'll let someone else weigh in as I wrote this code initially and am not sure if any of this makes a big difference.

Rodrigo-Novas:
> Well we could go one step further and write `flush_decoder = amt is None or (amt != 0 and not data)`. I'll let someone else weigh in as I wrote this code initially and am not sure if any of this makes a big difference.

Yes good catch!. I think it is not a core change but having the code without **elifs** makes it more readable and without so much redundancy.

pquentin:
Closing given the lack of interest, sorry.

src/urllib3/response.py
```diff
@@ -878,11 +878,7 @@ def read(
 
         data = self._raw_read(amt)
 
-        flush_decoder = False
-        if amt is None:
-            flush_decoder = True
-        elif amt != 0 and not data:
-            flush_decoder = True
+        flush_decoder = True if amt is None or (amt != 0 and not data) else False
```
illia-v, line 881: I like the change, `flush_decoder = amt is None or (amt != 0 and not data)` should be enough here

illia-v (approved):

PR #3119, opened 2023-09-05
Keep GitHub Action versions as comments

pquentin:
Dependabot will keep the comments, which will make updates to SHAs easier to reason about for mere humans.

See https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/ for more details. The idea is to merge this before #3116 in order to test it.


illia-v (approved):
Thanks!

PR #3123, opened 2023-09-13
Fix docs build

pquentin:
See https://github.com/sphinx-doc/sphinx/issues/11662 for details

Here's an example failing build: https://readthedocs.org/projects/urllib3/builds/21856472/

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,6 @@
 -r ../dev-requirements.txt
-sphinx>3.0.0,!=7.2.5
+# https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1713887182
+sphinx>3.0.0,<7.2.5
 requests
 furo
 sphinx-copybutton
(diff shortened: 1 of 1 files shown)


docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 -r ../dev-requirements.txt
-sphinx>3.0.0
+sphinx>3.0.0,!=7.2.5
```
illia-v, line 2: It looks like something that [is going to be fixed in PyO3](https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1713887182). Shouldn't the condition be `<7.2.5` here until then so that the new version of Sphinx does not bring the problem back if PyO3 is still the same version?

pquentin: Sure, good idea!

```suggestion
# https://github.com/sphinx-doc/sphinx/issues/11662#issuecomment-1713887182
sphinx>3.0.0,<7.2.5
```

illia-v (approved):
Thanks!

PR #3125, opened 2023-09-13
Fix a name of an attribute in docs

illia-v:
Docs suggested using `conn.is_proxy_verified` to check whether TLS connection to a proxy origin is verified, but the actual attribute name is `proxy_is_verified`.


pquentin (approved):
Thanks! LGTM.

PR #3124, opened 2023-09-13
Refer to current internet standards in docs on proxies

illia-v:
Docs section on proxies refers to two proposed standards obsoleted by RFC 9110 and RFC 9112. I'm changing the links to refer to the current standards.

PR #3128, opened 2023-09-19
Fix a link to "absolute URI" definition

illia-v:
I made a mistake changing one of the links in https://github.com/urllib3/urllib3/pull/3124


sethmlarson (approved):

sethmlarson:
Nice catch!

PR #3134, opened 2023-10-01
Typechecking fixes

dimbleby:
a couple of typechecking fixes, derived by enabling mypy's `ignore-without-code` and paying attention to the places that it flagged

the first commit is the most interesting one: a supposedly immutable dictionary of headers was being updated.  This went in at #2250, so far as I can see without comment, I guess this is simply not intended?

the second commit ever so slighly rearranges some similar code in a way that satisfies the typechecker, and the third enables `ignore-without-code`.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -231,7 +231,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             "upload_filename": "lolcat.txt",
             "filefield": ("lolcat.txt", data),
         }
-        fields["upload_size"] = len(data)  # type: ignore
+        fields["upload_size"] = len(data)  # type: ignore[assignment]
 
         with HTTPConnectionPool(self.host, self.port) as pool:
             r = pool.request("POST", "/upload", fields=fields)
@@ -270,7 +270,7 @@ class TestConnectionPool(HTTPDummyServerTestCase):
             "upload_filename": filename,
             fieldname: (filename, data),
         }
-        fields["upload_size"] = size  # type: ignore
+        fields["upload_size"] = size  # type: ignore[assignment]
         with HTTPConnectionPool(self.host, self.port) as pool:
             r = pool.request("POST", "/upload", fields=fields)
             assert r.status == 200, r.data
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,3 +116,6 @@ warn_redundant_casts = true
 warn_return_any = true
 warn_unused_configs = true
 warn_unused_ignores = true
+enable_error_code = [
+  "ignore-without-code",
+]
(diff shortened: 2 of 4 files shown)


dimbleby:
(rebased in the hope that y'all would have fixed your CI: but it seems to have gotten worse!  for some reason flake8 in pre-commit is linting strings as though they were code.  It'll be embarrassing when this is somehow my fault: but I can't reproduce locally...)

pquentin:
#3169 fixes the lint check, sorry about that.

test/with_dummyserver/test_poolmanager.py
```diff
@@ -614,18 +614,19 @@ def test_top_level_request_with_timeout(self) -> None:
...
         assert r.status == 200
         assert r.json() == body
-        if headers is not None and "application/json" not in headers.values():
-            assert "text/plain" in r.headers["Content-Type"].replace(" ", "").split(",")
-        else:
-            assert "application/json" in r.headers["Content-Type"].replace(
-                " ", ""
-            ).split(",")
+        content_type = None if headers is None else headers.get("content-Type")
+        content_type = "application/json" if content_type is None else content_type
```
illia-v, line 625: These lines are a bit hard to read. What do you think about changing this to one of the following variants?

```suggestion
        if not headers:
            content_type = "application/json"
        else:
            content_type = headers.get("content-Type", "application/json")
```
```suggestion
        content_type = (headers or {}).get("content-Type", "application/json")
```

test/with_dummyserver/test_connectionpool.py
```diff
@@ -231,7 +231,7 @@ def test_upload(self) -> None:
             "upload_filename": "lolcat.txt",
             "filefield": ("lolcat.txt", data),
         }
-        fields["upload_size"] = len(data)  # type: ignore
+        fields["upload_size"] = len(data)  # type: ignore[assignment]
```
illia-v, line 234: I didn't check if such a change to this line affects test coverage, but it looks to fix the ignored mypy error

```suggestion
        fields["upload_size"] = str(len(data))
```

illia-v:
> the first commit is the most interesting one: a supposedly immutable dictionary of headers was being updated. This went in at #2250, so far as I can see without comment, I guess this is simply not intended?

Nice catch!

Could you please split this PR into two: one for fixing the bug, another for `ignore-without-code`?

test/with_dummyserver/test_poolmanager.py
dimbleby: if you really mean "what do you think?": then I find the way that I wrote it more readable.

if you really mean "please change it" then sure, I dont care: please say that!

test/with_dummyserver/test_connectionpool.py
dimbleby: Sorry: but I don't care to change this

dimbleby:
> Could you please split this PR into two: one for fixing the bug, another for `ignore-without-code`?

#3203 is just the first commit

test/with_dummyserver/test_poolmanager.py
illia-v: I changed these lines in #3203, thanks for taking the bugfix into a separate PR!

illia-v (approved):
Thanks!

PR #3136, opened 2023-10-02
[1.26] Fix `brotli` extra for Python 2.7

illia-v:
The extra has been requiring `brotli>=1.0.9`. Brotli 1.1.0 dropped support for Python 2.7, but pip still installs it for 2.7.
https://github.com/google/brotli/issues/1074


illia-v:
Actually, current brotli still has classifiers for Python 2, so it could be just a mistake in 1.1.0 making the version faulty for Python 2.7

sethmlarson (approved):
LGTM, thanks!

aksh1493:
Which version of brotli we can use now for python 2.7?

PR #3137, opened 2023-10-02
[1.26] Fix running CPython 2.7 tests in CI

### LINKED ISSUE #3077: [1.26] GHA Python 2 support ended on June 17th
Warning: The support for python 2.7 will be removed on June 19. Related issue: https://github.com/actions/setup-python/issues/672

illia-v:
Fixes #3077

actions/setup-python dropped support for CPython 2.7.
Since we still need to run tests using the version, CI starts to use Ubuntu's CPython 2.7 version.
Unfortunately, testing using CPython 2.7 on macOS and Windows is dropped by this PR.

This PR is based on #3136 to display the tests passing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,14 +29,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         os: [macos-11, windows-latest, ubuntu-latest]
         experimental: [false]
         nox-session: ['']
         exclude:
-          # GitHub Actions dropped support for Python 2.7 and 3.7 in Ubuntu 22.04
-          - python-version: "2.7"
-            os: ubuntu-latest
+          # actions/setup-python dropped support for 3.6 in Ubuntu 22.04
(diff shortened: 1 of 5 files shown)


sethmlarson:
Let's test all the changes we implemented to 1.26.x against this Python 2.7 install, thanks @illia-v!

illia-v:
The coverage problem in Ubuntu 22.04 is related to not reaching [this warning](https://github.com/urllib3/urllib3/blob/1.26.x/src/urllib3/connection.py#L441-L447) because Ubuntu disabled TLS 1.0 & 1.1.

I am not sure what's the problem with the two tests related to IP versions in Ubuntu 20.04, but it looks like fixing them won't fix the coverage issue because Ubuntu 20.04 disables TLS 1.0 & 1.1 too (there is `SKIPPED [84] /home/runner/work/urllib3/urllib3/test/conftest.py:264: Test requires TLSv1.1` in [the log](https://github.com/urllib3/urllib3/actions/runs/6393273395/job/17352184396).)

I tried possible fixes like using Python 2.7 in a container in CI or tweaking OpenSSL config files, but it didn't work and I cannot invest more time in this for now. What do you think about ignoring the coverage miss for now to have tests passing?

sethmlarson:
@illia-v Let's ignore the coverage miss for the 1.26.x branch and get this merged if that's all that's needed.

PR #3143, opened 2023-10-05
Drop support for Python 3.7

pquentin:


sethmlarson (dismissed):
This is awesome to see, look at all those removals! :rocket: Let's add a changelog entry saying we removed 3.7.

sethmlarson (approved):

PR #3144, opened 2023-10-05
Fix docs build on Python 3.12

pquentin:
`ssl.wrap_socket` is gone in Python 3.12, let's link to `SSLContext.wrap_socket` instead: https://docs.python.org/3.12/whatsnew/3.12.html#ssl


sethmlarson (approved):
LGTM! :shipit:

PR #3145, opened 2023-10-06
Remove requires_ssl_context_keyfile_password()

pquentin:
SSLContext.load_cert_chain() supports passwords since Python 3.3. Note that this test was only executed on macOS, as requires_ssl_context_keyfile_password was actually only requiring the presence of Secure Transport.


pquentin:
Closing this one in favor of #3146  because this was not a mistake, and it's indeed SecureTransport that does not support this.

PR #3150, opened 2023-10-08
Added ca cert data key support in ssl proxy content

### LINKED ISSUE #3065: Cannot pass ssl kwarg cert_data in proxy context
Discovered that one while writting tls-in-tls http/2 tests in #3030 

While we can do

```python
with proxy_from_url("...", ca_certs=DEFAULT_CA) as http:
    ...
```

we cannot 

```python
with proxy_from_url("...", cert_data=open(DEFAULT_CA, "rb").read()) as http:
    ...
```

output `unexpected kw key_cert_data for pool conn`

I suspect that we missed propagating required kw.

It is an easy fix.

should patch `SSL_KEYWORDS = (....)` and `class PoolKey(typing.NamedTuple):`

ravi-kale:
Fixes #3065


test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -78,6 +78,16 @@ def test_https_proxy(self) -> None:
             r = https.request("GET", f"{self.http_url}/")
             assert r.status == 200
 
+    def test_http_and_https_kwarg_ca_cert_data_proxy(self) -> None:
+        with proxy_from_url(
+            self.https_proxy_url, ca_cert_data=open(DEFAULT_CA, "rb").read()
```
illia-v, line 83: I guess the test fails in CI because `DEFAULT_CA` contains a PEM-encoded certificate and it is not provided as a string.

https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain
> The cadata object, if present, is either an ASCII string of one or more PEM-encoded certificates or a [bytes-like object](https://docs.python.org/3/glossary.html#term-bytes-like-object) of DER-encoded certificates. Like with capath extra lines around PEM-encoded certificates are ignored but at least one certificate must be present.

illia-v: Please remember to close the file properly, use `with` for opening it

.gitignore
```diff
@@ -8,3 +8,4 @@
 /build
 /docs/_build
 coverage.xml
+myenv
```
illia-v, line 11:
```suggestion
```

changelog/3065.bugfix.rst
illia-v, line 1: Mention `ca_cert_data` instead of `cert_data`

illia-v (changes requested):
Thanks for the fix, please check my comments

changelog/3065.bugfix.rst
ravi-kale: updated the changelog

test/with_dummyserver/test_proxy_poolmanager.py
ravi-kale: Yes, Thanks for the review. Made changes

illia-v (approved):

illia-v:
Thanks @ravi-kale!
#3065 is covered by a $100 bounty, please submit an expense to [Open Collective](https://opencollective.com/urllib3) if you’re able to receive funds for working on OSS

ravi-kale:
That's great, Thank You @illia-v

PR #3153, opened 2023-10-09
[1.26] Use vendored six in urllib3.contrib.securetransport

hroncok:
Otherwise, I get:

    ModuleNotFoundError: No module named 'six'

Note that this is otherwise untested.

I've seen this in Fedora when I wanted to smoke test importing all submodules.
This module is only for macOS, so I do not have a way to check anything further than that.
I realize the 1.26.x series is probably security-only, feel free to close this if you are not interested.
Also, feel free to amend this in any way you wish.

### DIFF
--- a/src/urllib3/contrib/securetransport.py
+++ b/src/urllib3/contrib/securetransport.py
@@ -64,7 +64,7 @@ import struct
 import threading
 import weakref
 
-import six
+from ..packages import six
 
 from .. import util
 from ..util.ssl_ import PROTOCOL_TLS_CLIENT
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Wow, I am surprised that this erroneous import path wasn't discovered for [over 3 years](https://github.com/urllib3/urllib3/commit/688584d692f3d77f5d6beca6bb97faef15f0977b). Just goes to show how ubiquitous the six package is. Thanks for noticing this and sending a PR.

PR #3156, opened 2023-10-17
Add CPython 3.13 to CI

illia-v:
Unfortunately, compiling some extensions against CPython 3.13 fails at the moment


skydiver1980 (approved):

illia-v:
`test/test_response.py::TestResponse::test_base_io` started failing with `PytestUnraisableExceptionWarning` because of https://github.com/python/cpython/issues/62948 included in CPython 3.13.

illia-v:
Tests display that everything except one test is compatible with CPython 3.13. I am not sure if we should merge this at the moment with the workarounds though.

pquentin:
I'm not worried by the workarounds as long as they're clearly labeled as such so that we can remove them at some point.

pquentin:
The Trio pull request for Python 3.13 support has been merged, by the way.

illia-v:
> The Trio pull request for Python 3.13 support has been merged, by the way.

Yes, that's great! I updated the pin to a commit from Trio's main branch

illia-v:
> `test/test_response.py::TestResponse::test_base_io` started failing with `PytestUnraisableExceptionWarning` because of [python/cpython#62948](https://github.com/python/cpython/issues/62948) included in CPython 3.13.
> 
> 

It looks like there are a few new errors similar to this one because of the new HTTP/2 code

pquentin (approved):
Thanks! LGTM. Can you please open an issue to fix the new Python 3.13 errors?

illia-v:
> Can you please open an issue to fix the new Python 3.13 errors?

I fixed the new ones by adding `HTTP2Response.close` to prevent raising `NotImplementedError`.

pquentin:
RIght! I was more concerned about https://github.com/urllib3/urllib3/pull/3156/files#diff-f853b537efd0c07ee4e2affffe963ac50d718f5894aa4e6f4e57c5f98ad98599R562 but looking more closely I understand now that the `filterwarnings` is an OK fix, since indeed `BaseHTTPResponse` does not implement close(). Not sure why we have such a test though! Maybe for coverage.

illia-v:
> but looking more closely I understand now that the `filterwarnings` is an OK fix

Yes, it is a fix rather than a workaround.

BTW, it's the only test where `BaseHTTPResponse` is initialized.

PR #3162, opened 2023-10-21
Tests: skip the requires_network() when heavily firewalled

stefanor:
requires_network() attempts to detect Internet access by probing non-local IPs. A restricted environment may appear to have a routes to the Internet, but block all outbound traffic.

Treat that as not having network.

The use-case here is Debian package builds without Intenet. Ideally we'd have a way to explicitly disable tests that require Internet.


sigmavirus24:
> Ideally we'd have a way to explicitly disable tests that require Internet.

Pytest has a way of defining a custom option in the conftest file that could be used as part of a decorator that defaults to true to enable tests like this by default but allow you to disable them

stefanor:
> Pytest has a way of defining a custom option in the conftest file that could be used as part of a decorator that defaults to true to enable tests like this by default but allow you to disable them

Yes, if you're happy I could implement that. From this method existing, I got the feeling that something automatic was preferred.

sigmavirus24:
No, it is much better to be obvious why distros aren't going to ever get 100% test coverage or full confidence in the way they distribute things.

stefanor:
Even with #3166, there's still an issue here. The problem is that `skipif()`s condition is evaluated during test collection, before mark selection is taken into account.

So, having a ConnectionRefusedError be raised will still crash out the test-suite during collection.

illia-v:
@stefanor regarding 3166, was it possible to have `skipif` wrapped by `pytest.mark.requires_network`?

stefanor:
> @stefanor regarding 3166, was it possible to have skipif wrapped by pytest.mark.requires_network?

That's how it's structured. But the whole thing gets evaluated during collection. I assume collection happens before selection.

stefanor:
If your concern is the heuristic, we could always have it return True, like a timeout, instead of False.

sethmlarson (approved):
I think this looks fine to me, @illia-v can you take another look too?

illia-v:
If I skip tests with the new `requires_network` pytest marker, I'd expect that the tests don't reach to external hosts. But `_has_route()` still tries to create a connection.

This patch silences the error for the Debian's case, but somebody else may face a different behavior in the future and we may need to add handling for other errors.

I'd rather attempt to prevent `_has_route()` evaluation when the tests are skipped. It looks like it's possible to do based on examples from [this pytest docs section](https://docs.pytest.org/en/7.1.x/example/markers.html#marking-platform-specific-tests-with-pytest): wrap tests by `@pytest.mark.requires_network` instead of `@requires_network`, then check `_has_route()` for marked tests in `pytest_runtest_setup`. `_has_route` can be wrapped by `@functools.lru_cache`.

stefanor:
> I'd rather attempt to prevent _has_route() evaluation when the tests are skipped.

Filed #3187 to do that, within the current structure.

PR #3163, opened 2023-10-22
Don't actually attempt to connect to evil.com in test_deprecated_no_scheme

stefanor:
We're testing that a DeprecationWarning is raised, not trying to talk to `evil.com`

### DIFF
--- a/test/test_poolmanager.py
+++ b/test/test_poolmanager.py
@@ -259,11 +259,15 @@ class TestPoolManager:
         assert pool is other_pool
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
 
-    def test_deprecated_no_scheme(self) -> None:
+    @patch("urllib3.poolmanager.PoolManager.connection_from_host")
+    def test_deprecated_no_scheme(self, connection_from_host: mock.MagicMock) -> None:
+        # Don't actually make a network connection, just verify the DeprecationWarning
+        connection_from_host.side_effect = ConnectionError("Not attempting connection")
         p = PoolManager()
 
         with pytest.warns(DeprecationWarning) as records:
-            p.request(method="GET", url="evil.com://good.com")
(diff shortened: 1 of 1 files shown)


test/test_poolmanager.py
```diff
@@ -259,11 +259,14 @@ def test_http_connection_from_context_case_insensitive(self) -> None:
...
 
-    def test_deprecated_no_scheme(self) -> None:
+    @patch("urllib3.poolmanager.PoolManager.connection_from_host")
+    def test_deprecated_no_scheme(self, connection_from_host: mock.MagicMock) -> None:
+        connection_from_host.side_effect = ConnectionError("Not attempting connection")
         p = PoolManager()
 
         with pytest.warns(DeprecationWarning) as records:
-            p.request(method="GET", url="evil.com://good.com")
+            with pytest.raises(ConnectionError):
```
sigmavirus24, line 268: Everywhere else this will **fail** because you're introducing Debian specific behaviour here that we cannot and will not replicate elsewhere

test/test_poolmanager.py
```diff
@@ -259,11 +259,14 @@ def test_http_connection_from_context_case_insensitive(self) -> None:
         assert pool is other_pool
         assert all(isinstance(key, PoolKey) for key in p.pools.keys())
 
-    def test_deprecated_no_scheme(self) -> None:
+    @patch("urllib3.poolmanager.PoolManager.connection_from_host")
+    def test_deprecated_no_scheme(self, connection_from_host: mock.MagicMock) -> None:
+        connection_from_host.side_effect = ConnectionError("Not attempting connection")
```
illia-v, line 264: Please add a comment describing that this side effect exists only to allow the test to pass without internet access

sethmlarson (approved):
Thanks @stefanor and @illia-v for the review!

Issue #3164, opened 2023-10-22
test_socketlevel has some flaky tests


stefanor:
### Subject

I see repeated unrelability in some test_socketlevel tests, in particular:
`test_chunked_specified`, `test_chunked_not_specified`, `test_framing_set_via_headers`.

Sometimes they pass, but not always.

### Environment

Debian unstable. But the issue is reproducible inside nox virtualenvs too.

At least, paste here the output of:

```
>>> print("OS", platform.platform())
OS Linux-6.5.0-2-amd64-x86_64-with-glibc2.37
>>> print("Python", platform.python_version())
Python 3.11.6
>>> print(ssl.OPENSSL_VERSION)
OpenSSL 3.0.11 19 Sep 2023
>>> print("urllib3", urllib3.__version__)
urllib3 1.26.17 <- but I'm filing this bug against current main (4e98d57809dacab1cbe625fddeec1a290c478ea9)
```

### Steps to Reproduce

```
nox -e test-3.11 -- -k 'test_chunked_specified or test_chunked_not_specified or test_framing_set_via_headers'
```

### Expected Behavior

Tests should pass, reliably.

### Actual Behavior

Tests fail sporadically, e.g. These are collated from several runs:

```
__________________ TestContentFraming.test_chunked_specified[file-PUT-True] __________________
Traceback (most recent call last):
  File "/tmp/urllib3/.nox/test-3-11/lib/python3.11/site-packages/_pytest/runner.py", line 341, in from_call
    result: Optional[TResult] = func()
                                ^^^^^^
  File "/tmp/urllib3/.nox/test-3-11/lib/python3.11/site-packages/_pytest/runner.py", line 262, in <lambda>
    lambda: ihook(item=item, **kwds), when=when, reraise=reraise
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/urllib3/.nox/test-3-11/lib/python3.11/site-packages/pluggy/_hooks.py", line 493, in __call__
    return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/urllib3/.nox/test-3-11/lib/python3.11/site-packages/pluggy/_manager.py", line 115, in _hookexec
    return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
[...]

PR #3165, opened 2023-10-22
Mark tests known to be flaky as flaky

stefanor:
They really should be investigated and fixed, but this at least makes the test suite usable.

See: #3164

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 coverage==7.3.2
+flaky==3.7.0
 tornado==6.3.3
 PySocks==1.7.1
 pytest==7.4.2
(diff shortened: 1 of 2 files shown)


sigmavirus24:
I believe these are only flakey in your debian testing, right? That doesn't mean these are flakey everywhere

stefanor:
> I believe these are only flakey in your debian testing, right? That doesn't mean these are flakey everywhere

I assuming this flakyness wasn't Debian-specific. But if you've never observed them to fail before, maybe there is something to that?

pquentin:
Yes, those tests work fine in CI. However, I noticed they use `LONG_TIMEOUT`, which is something we bump in CI based on the presence of environment variables:

https://github.com/urllib3/urllib3/blob/bba0b2b7773075297972e578e9cad500ea5f65d2/test/__init__.py#L76C1-L79

Is it feasible to set the CI enviromnent variable in your testing?

PR #3166, opened 2023-10-23
Tests: Use a mark for requires_network

stefanor:
As an alternative to #3162, we can add a `requires_network` pytest mark, rather than relying on automatic network detection.

The use-case here is Debian package builds without Intenet access. We have environments where Internet access is available, and environments where it isn't. We'd like to be able to specify our expectation to the test-suite.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,9 @@ include = [
 [tool.pytest.ini_options]
 xfail_strict = true
 python_classes = ["Test", "*TestCase"]
-markers = ["limit_memory"]
+markers = [
+    "limit_memory: Limit memory with memray",
+]
 log_level = "DEBUG"
 filterwarnings = [
     "error",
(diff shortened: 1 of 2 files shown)


illia-v (commented):
I am not sure about dropping `requires_network`, but the explicit marker looks to be a nice addition. 

What do you think about decorating `requires_network` itself by `pytest.mark.requires_network`? This way it should keep the existing behavior while adding a possibility to deselect the tests with `@requires_network()` explicitly.

stefanor:
> What do you think about decorating requires_network itself by pytest.mark.requires_network? This way it should keep the existing behavior while adding a possibility to deselect the tests with @requires_network() explicitly.

Done, that is much simpler.

sethmlarson (approved):
Thanks @stefanor and @illia-v for the review!

PR #3169, opened 2023-10-31
Fix lint on Python 3.12

pquentin:
GitHub Actions now defaults to Python 3.12, which made flake8 6.0 report E231 errors because it was looking at f-strings incorrectly. Upgrading to 6.1 fixed the issue, but introduced E721 that I fixed too.

This commit also changes `nox -s lint` and `nox -s mypy` to use Python 3.12, allowing to remove one type ignore.

This was noticed by @dimbleby in #3134. Thanks!

### DIFF
--- a/test/test_ssltransport.py
+++ b/test/test_ssltransport.py
@@ -91,11 +91,11 @@ def validate_response(
 
 def validate_peercert(ssl_socket: SSLTransport) -> None:
     binary_cert = ssl_socket.getpeercert(binary_form=True)
-    assert type(binary_cert) == bytes
+    assert isinstance(binary_cert, bytes)
     assert len(binary_cert) > 0
 
     cert = ssl_socket.getpeercert()
-    assert type(cert) == dict
+    assert isinstance(cert, dict)
     assert "serialNumber" in cert
     assert cert["serialNumber"] != ""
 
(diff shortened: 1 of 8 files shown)


illia-v (approved):
Thanks @pquentin!

test/test_ssltransport.py
```diff
@@ -91,11 +91,11 @@ def validate_response(
 
 def validate_peercert(ssl_socket: SSLTransport) -> None:
     binary_cert = ssl_socket.getpeercert(binary_form=True)
-    assert type(binary_cert) == bytes
+    assert isinstance(binary_cert, bytes)
```
sethmlarson, line 94: Doesn't this change the meaning of these asserts? Now subclasses would pass, too.

illia-v: It does, is this an issue?

sethmlarson: I'd prefer keeping as-is, does the tool say anything about why they're upgrading this? Very strange given that its a change in behavior.

illia-v: pycodestyle (flake8's dependency) had [a related change](https://pycodestyle.pycqa.org/en/latest/developer.html#id3:~:text=E721%3A%20adjust%20handling%20of%20type%20comparison.%20Allowed%20forms%20are%20now%20isinstance(x%2C%20t)%20or%20type(x)%20is%20t.%20PR%20%231086%2C%20%231167.) recently

The exact error is "E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`", so `assert type(binary_cert) is bytes` is allowed, this is a more correct way to check the type than with `==`

pquentin: Our tests are littered with isinstance() checks that could be more precise, but I revert the ones in this pull request (and a few more) in https://github.com/urllib3/urllib3/pull/3169/commits/b3fa87a944024023ae6f54f51b315fdca6ad19c9.

sethmlarson (approved):
LGTM

Issue #3160, opened 2023-10-17
`set-output` will be disabled soon in GitHub Actions


sethmlarson:
We're seeing this warning especially in our release GHA configuration, we should fix it up so that our releases continue to work: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #3175, opened 2023-11-03
ci: replace deprecated set-output

FineFindus:
Closes https://github.com/urllib3/urllib3/issues/3160 by replacing the deprecated `set-output` with the newer `$GITHUB_OUTPUT` environment files, as described in [GitHub's Blog](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). 
I [did a test action run](https://github.com/FineFindus/urllib3/actions/runs/6749422237), and it did not get any warnings, so it should be fixed.
Action run without deprecation warning

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -37,7 +37,7 @@ jobs:
       - name: "Generate hashes"
         id: hash
         run: |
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Thanks much! And this issue is covered by a $100 bounty, so if you want to submit that to our Open Collective (with a link to this PR) that would be great.

PR #3176, opened 2023-11-03
Treat x-gzip content encoding as gzip

### LINKED ISSUE #3174: Support `Content-Encoding: x-gzip`
### Context

I'm using an API that returns a gzipped response, but the reported `Content-Encoding` is `x-gzip` rather than `gzip`, so urllib3 fails to decompress it.

According to [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.4.1.3):

>  A recipient SHOULD consider "x-gzip" to be equivalent to "gzip".

Therefore, I would expect e.g. `Response.read(decode_content=True)` to automatically decompress such responses.

Unfortunately, I can't find a public server that would produce such responses, so I can't provide a simple reproducer.

### Alternatives

I can set `decode_content=False` and gunzip the content manually.

### Duplicate

> Has the feature been requested before?

Not that I could find.

### Contribution

Probably not.

jeremycline:
According to RFC 9110, the "x-gzip" content coding should be treated as "gzip".

Fixes #3174

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -229,14 +229,15 @@ class TestResponse:
         assert r.read() == b""
         assert r.read() == b""
 
-    def test_chunked_decoding_gzip(self) -> None:
+    @pytest.mark.parametrize("content_encoding", ["gzip", "x-gzip"])
+    def test_chunked_decoding_gzip(self, content_encoding: str) -> None:
         compress = zlib.compressobj(6, zlib.DEFLATED, 16 + zlib.MAX_WBITS)
         data = compress.compress(b"foo")
         data += compress.flush()
 
         fp = BytesIO(data)
         r = HTTPResponse(
-            fp, headers={"content-encoding": "gzip"}, preload_content=False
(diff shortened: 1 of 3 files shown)


test/test_response.py
```diff
@@ -260,6 +260,16 @@ def test_decode_gzip_error(self) -> None:
         with pytest.raises(DecodeError):
             HTTPResponse(fp, headers={"content-encoding": "gzip"})
 
+    def test_decode_x_gzip(self) -> None:
```
sethmlarson, line 263: Since this test is identical to the above test could you instrad parametrize it?

sethmlarson (commented):
One comment:

test/test_response.py
jeremycline: Sure thing, I parametrized `test_chunked_decoding_gzip()` and rebased.

sethmlarson (approved):
LGTM, thank you!

sethmlarson:
This PR is eligible for a $100 bounty, submit an expense to our Open Collective and link back to this PR if you are eligible to!

PR #3181, opened 2023-11-05
Run test_requesting_large_resources_via_ssl separately

pquentin:
We have many timeouts on GitHub Actions due to those tests that can take minutes to run on overloaded CI runners. This is compounded by the fact they run six times (3 parametrizations * 2 SSL configuration - pyopenssl and standard library SSL). Running on multiple platforms isn't useful. We found a "Windows-specific" bug once but only because the Windows worker had less memory, and we have `limit_memory` tests now for the memory aspects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,10 +43,22 @@ jobs:
         nox-session: ['']
         include:
           - experimental: false
+          # integration
+          # 3.8 and 3.9 have a known issue with large SSL requests that we work around:
+          # https://github.com/urllib3/urllib3/pull/3181#issuecomment-1794830698
+          - python-version: "3.8"
+            os: ubuntu-latest
+            experimental: false
+            nox-session: test_integration
+          - python-version: "3.9"
+            os: ubuntu-latest
+            experimental: false
+            nox-session: test_integration
           - python-version: "3.12"
             os: ubuntu-latest
             experimental: false
             nox-session: test_integration
+          # pypy
           - python-version: "pypy-3.8"
             os: ubuntu-latest
             experimental: false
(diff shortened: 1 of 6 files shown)


sethmlarson (dismissed):
This is excellent and I love the result for CI times! Now I'm inspired to look into the macOS times :rocket:

illia-v:
Thanks @pquentin, this is a great idea!

Based on the following docstring, the library code relevant to the test may work differently in CPython 3.8, 3.9, and in any version with the pyOpenSSL injection.
https://github.com/urllib3/urllib3/blob/f7cd7f3f632cf5224f627536f02c2abf7e4146d1/src/urllib3/response.py#L763-L767

What do you think about running the test in CPython 3.8, 3.9, and the latest one to make sure that no changes to `_fp_read` break the method for users of old CPython versions?

pquentin:
Excellent point, thank you. I'm wondering if I should not instead try to monkeypatch the calls that CPython is making to avoid decrypting multiple gigabytes of actual SSL data. If it proves difficult, then I'll run the integration tests on 3.8, 3.9 and latest. How does that sound?

illia-v:
@pquentin I'm curious to see what can be patched to avoid that.

IIRC there are three approaches CPython uses in different versions:
* using `SSL_read_ex` call in a loop in CPython 3.10+
* using `SSL_read` call in a loop with the https://github.com/python/cpython/commit/153365d864c411f6fb523efa752ccb3497d815ca patch
* using `SSL_read` call in a loop without the patch

They have different consequences for urllib3.

Also, when pyOpenSSL is used, the library is responsible for the decryption.

sethmlarson:
I'm liking the idea to run it on 3.8, 3.9, and latest if only to speed up every other PRs CI time for now? Quite the improvement, would like to take advantage :eyes:

pquentin:
I still think the mocking idea could work, but I'd prefer focusing my energy on more useful things! Added 3.8 and 3.9.

illia-v (approved):
Thanks!

PR #3186, opened 2023-11-08
Support read1() method in HTTPResponse

### LINKED ISSUE #2123: When iterating over a `stream`ed response with amt=None, entire response is returned instead of streaming

smason:
Fix suggested in #2125 but I actually wanted to get #2123 fixed and this seems like a reasonable starting point.

Presumably needs more tests as I'm not doing much on that front at the moment.  But didn't want to do too much before getting some feedback.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -280,6 +280,19 @@ class BytesQueueBuffer:
 
         return ret.getvalue()
 
+    def get_all(self) -> bytes:
+        buffer = self.buffer
+        if not buffer:
+            assert self._size == 0
+            return b""
+        if len(buffer) == 1:
+            result = buffer.pop()
+        else:
+            result = b"".join(buffer)
+            buffer.clear()
+        self._size = 0
+        return result
+
 
 class BaseHTTPResponse(io.IOBase):
     CONTENT_DECODERS = ["gzip", "x-gzip", "deflate"]
@@ -390,6 +403,13 @@ class BaseHTTPResponse(io.IOBase):
         amt: int | None = None,
         decode_content: bool | None = None,
         cache_content: bool = False,
+    ) -> bytes:
+        raise NotImplementedError()
+
+    def read1(
+        self,
+        amt: int | None = None,
+        decode_content: bool | None = None,
     ) -> bytes:
         raise NotImplementedError()
 
@@ -752,7 +772,12 @@ class HTTPResponse(BaseHTTPResponse):
             if self._original_response and self._original_response.isclosed():
                 self.release_conn()
 
-    def _fp_read(self, amt: int | None = None) -> bytes:
+    def _fp_read(
+        self,
+        amt: int | None = None,
+        *,
+        read1: bool = False,
+    ) -> bytes:
         """
         Read a response with the thought that reading the number of bytes
         larger than can fit in a 32-bit int at a time via SSL in some
@@ -770,8 +795,14 @@ class HTTPResponse(BaseHTTPResponse):
         c_int_max = 2**31 - 1
         if (
             (amt and amt > c_int_max)
-            or (self.length_remaining and self.length_remaining > c_int_max)
+            or (
+                amt is None
+                and self.length_remaining
+                and self.length_remaining > c_int_max
+            )
         ) and (util.IS_PYOPENSSL or sys.version_info < (3, 10)):
+            if read1:
+                return self._fp.read1(c_int_max)
             buffer = io.BytesIO()
             # Besides `max_chunk_amt` being a maximum chunk size, it
             # affects memory overhead of reading a response by this
@@ -792,6 +823,8 @@ class HTTPResponse(BaseHTTPResponse):
                 buffer.write(data)
                 del data  # to reduce peak memory usage by `max_chunk_amt`.
             return buffer.getvalue()
+        elif read1:
+            return self._fp.read1(amt) if amt is not None else self._fp.read1()
         else:
             # StringIO doesn't like amt=None
             return self._fp.read(amt) if amt is not None else self._fp.read()
@@ -799,6 +832,8 @@ class HTTPResponse(BaseHTTPResponse):
     def _raw_read(
         self,
         amt: int | None = None,
+        *,
+        read1: bool = False,
     ) -> bytes:
         """
         Reads `amt` of bytes from the socket.
@@ -809,7 +844,7 @@ class HTTPResponse(BaseHTTPResponse):
         fp_closed = getattr(self._fp, "closed", False)
 
         with self._error_catcher():
-            data = self._fp_read(amt) if not fp_closed else b""
+            data = self._fp_read(amt, read1=read1) if not fp_closed else b""
             if amt is not None and amt != 0 and not data:
                 # Platform-specific: Buggy versions of Python.
                 # Close the connection when no data is returned
@@ -908,6 +943,57 @@ class HTTPResponse(BaseHTTPResponse):
             data = self._decoded_buffer.get(amt)
 
         return data
+
+    def read1(
+        self,
+        amt: int | None = None,
+        decode_content: bool | None = None,
+    ) -> bytes:
+        """
+        Similar to ``http.client.HTTPResponse.read1`` and documented
+        in :meth:`io.BufferedReader.read1`, but with an additional parameter:
+        ``decode_content``.
+
+        :param amt:
+            How much of the content to read.
+
+        :param decode_content:
+            If True, will attempt to decode the body based on the
+            'content-encoding' header.
+        """
+        if decode_content is None:
+            decode_content = self.decode_content
+        # try and respond without going to the network
+        if self._has_decoded_content:
+            if not decode_content:
+                raise RuntimeError(
+                    "Calling read1(decode_content=False) is not supported after "
+                    "read1(decode_content=True) was called."
+                )
+            if len(self._decoded_buffer) > 0:
+                if amt is None:
+                    return self._decoded_buffer.get_all()
+                return self._decoded_buffer.get(amt)
+        if amt == 0:
+            return b""
+
+        # FIXME, this method's type doesn't say returning None is possible
+        data = self._raw_read(amt, read1=True)
+        if not decode_content or data is None:
+            return data
+
+        self._init_decoder()
+        while True:
+            flush_decoder = not data
+            decoded_data = self._decode(data, decode_content, flush_decoder)
+            self._decoded_buffer.put(decoded_data)
+            if decoded_data or flush_decoder:
+                break
+            data = self._raw_read(8192, read1=True)
+
+        if amt is None:
+            return self._decoded_buffer.get_all()
+        return self._decoded_buffer.get(amt)
 
     def stream(
         self, amt: int | None = 2**16, decode_content: bool | None = None
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -73,6 +73,23 @@ class TestBytesQueueBuffer:
         assert buffer.get(4) == b"rbaz"
         assert len(buffer) == 0
 
+    def test_get_all_empty(self) -> None:
+        q = BytesQueueBuffer()
+        assert q.get_all() == b""
+
+    def test_get_all_single(self) -> None:
+        q = BytesQueueBuffer()
+        q.put(b"a")
+        assert q.get_all() == b"a"
+
+    def test_get_all_many(self) -> None:
+        q = BytesQueueBuffer()
+        q.put(b"a")
+        q.put(b"b")
+        q.put(b"c")
+        assert q.get_all() == b"abc"
+        assert len(q) == 0
+
     @pytest.mark.limit_memory("12.5 MB")  # assert that we're not doubling memory usage
     def test_memory_usage(self) -> None:
         # Allocate 10 1MiB chunks
@@ -184,6 +201,39 @@ class TestResponse:
         assert r.read(2) == b"oo"
         assert r.read() == b""
         assert r.read() == b""
+
+    def test_reference_read1(self) -> None:
+        fp = BytesIO(b"foobar")
+        r = HTTPResponse(fp, preload_content=False)
+
+        assert r.read1(0) == b""
+        assert r.read1(1) == b"f"
+        assert r.read1(2) == b"oo"
+        assert r.read1() == b"bar"
+        assert r.read1() == b""
+
+    def test_reference_read1_nodecode(self) -> None:
+        fp = BytesIO(b"foobar")
+        r = HTTPResponse(fp, preload_content=False, decode_content=False)
+
+        assert r.read1(0) == b""
+        assert r.read1(1) == b"f"
+        assert r.read1(2) == b"oo"
+        assert r.read1() == b"bar"
+        assert r.read1() == b""
+
+    def test_decoding_read1(self) -> None:
+        data = zlib.compress(b"foobar")
+
+        fp = BytesIO(data)
+        r = HTTPResponse(
+            fp, headers={"content-encoding": "deflate"}, preload_content=False
+        )
+
+        assert r.read1(1) == b"f"
+        assert r.read1(2) == b"oo"
+        assert r.read1() == b"bar"
+        assert r.read1() == b""
 
     def test_decode_deflate(self) -> None:
         data = zlib.compress(b"foo")
@@ -624,6 +674,35 @@ class TestResponse:
         ):
             resp.read(decode_content=False)
 
+    def test_read1_with_illegal_mix_decode_toggle(self) -> None:
+        data = zlib.compress(b"foo")
+
+        fp = BytesIO(data)
+
+        resp = HTTPResponse(
+            fp, headers={"content-encoding": "deflate"}, preload_content=False
+        )
+
+        assert resp.read1(1) == b"f"
+
+        with pytest.raises(
+            RuntimeError,
+            match=(
+                r"Calling read1\(decode_content=False\) is not supported after "
+                r"read1\(decode_content=True\) was called"
+            ),
+        ):
+            resp.read1(1, decode_content=False)
+
+        with pytest.raises(
+            RuntimeError,
+            match=(
+                r"Calling read1\(decode_content=False\) is not supported after "
+                r"read1\(decode_content=True\) was called"
+            ),
+        ):
+            resp.read1(decode_content=False)
+
     def test_read_with_mix_decode_toggle(self) -> None:
         data = zlib.compress(b"foo")
 
@@ -725,6 +804,9 @@ class TestResponse:
                 if len(self.payloads) > 0:
                     return self.payloads.pop(0)
                 return b""
+
+            def read1(self, amt: int) -> bytes:  # type: ignore[override]
+                return self.read(amt)
 
         uncompressed_data = zlib.decompress(ZLIB_PAYLOAD)
 
@@ -921,6 +1003,9 @@ class TestResponse:
                     self.fp = None
 
                 return data
+
+            def read1(self, amt: int) -> bytes:
+                return self.read(1)
 
             def close(self) -> None:
                 self.fp = None
@@ -1292,6 +1377,9 @@ class MockChunkedEncodingResponse:
         return self.pop_current_chunk(till_crlf=True)
 
     def read(self, amt: int = -1) -> bytes:
+        return self.pop_current_chunk(amt)
+
+    def read1(self, amt: int = -1) -> bytes:
         return self.pop_current_chunk(amt)
 
     def flush(self) -> None:
--- a/changelog/3186.feature.rst
+++ b/changelog/3186.feature.rst
@@ -1,0 +1,1 @@
+Added support for HTTPResponse.read1() method.
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -1583,12 +1583,47 @@ class TestSSL(SocketDummyServerTestCase):
                 pool.request("GET", "/", retries=False, timeout=LONG_TIMEOUT)
         assert server_closed.wait(LONG_TIMEOUT), "The socket was not terminated"
 
+    def _run_preload(self, pool: HTTPSConnectionPool, content_length: int) -> None:
+        response = pool.request("GET", "/")
+        assert len(response.data) == content_length
+
+    def _run_read_None(self, pool: HTTPSConnectionPool, content_length: int) -> None:
+        response = pool.request("GET", "/", preload_content=False)
+        assert len(response.read(None)) == content_length
+        assert response.read(None) == b""
+
+    def _run_read_amt(self, pool: HTTPSConnectionPool, content_length: int) -> None:
+        response = pool.request("GET", "/", preload_content=False)
+        assert len(response.read(content_length)) == content_length
+        assert response.read(5) == b""
+
+    def _run_read1_None(self, pool: HTTPSConnectionPool, content_length: int) -> None:
+        response = pool.request("GET", "/", preload_content=False)
+        remaining = content_length
+        while True:
+            chunk = response.read1(None)
+            if not chunk:
+                break
+            remaining -= len(chunk)
+        assert remaining == 0
+
+    def _run_read1_amt(self, pool: HTTPSConnectionPool, content_length: int) -> None:
+        response = pool.request("GET", "/", preload_content=False)
+        remaining = content_length
+        while True:
+            chunk = response.read1(content_length)
+            if not chunk:
+                break
+            remaining -= len(chunk)
+        assert remaining == 0
+
     @pytest.mark.integration
     @pytest.mark.parametrize(
-        "preload_content,read_amt", [(True, None), (False, None), (False, 2**31)]
+        "method",
+        [_run_preload, _run_read_None, _run_read_amt, _run_read1_None, _run_read1_amt],
     )
     def test_requesting_large_resources_via_ssl(
-        self, preload_content: bool, read_amt: int | None
+        self, method: typing.Callable[[typing.Any, HTTPSConnectionPool, int], None]
     ) -> None:
         """
         Ensure that it is possible to read 2 GiB or more via an SSL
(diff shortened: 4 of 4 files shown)


src/urllib3/response.py
```diff
@@ -876,39 +898,72 @@ def read(
...
+        self._init_decoder()
+        if decode_content is None:
+            decode_content = self.decode_content
+
+        # seems unlikely that 32MiB will ever be available in the OS buffer
+        if amt is None:
+            amt = 2**25
+
+        if len(self._decoded_buffer) >= amt:
+            return self._decoded_buffer.get(amt)
```
illia-v, line 940: One `SSL_read` call [can return 16kB](https://www.openssl.org/docs/man3.2/man3/SSL_read.html#:~:text=The%20data%20are%20received%20in%20records%20(with%20a%20maximum%20record%20size%20of%2016kB).) of data; a small experiment shows that this amount may be decompressed to much more than 32MiB.
```sh
$ dd if=/dev/zero bs=64MiB count=1 | tr '\000' '1'  > 64MiB.txt
1+0 records in
1+0 records out
67108864 bytes (67 MB, 64 MiB) copied, 0.138876 s, 483 MB/s
$ du -sh 64MiB.txt   
64M	64MiB.txt
$ brotli 64MiB.txt
$ du -sh 64MiB.txt.br 
4.0K	64MiB.txt.br
```

Is there a reason not to return everything from `self._decoded_buffer` and make no `self._raw_read` calls when `self._decoded_buffer` is not empty?

src/urllib3/response.py
```diff
@@ -767,6 +779,10 @@ def _fp_read(self, amt: int | None = None) -> bytes:
           * CPython < 3.10 only when `amt` does not fit 32-bit int.
         """
         assert self._fp
+        max_chunk_amt = 2**28
+        if one_read:
+            amt = max_chunk_amt if amt is None else min(amt, max_chunk_amt)
+            return self._fp.read1(amt)
```
illia-v, line 785: It's not obvious why you have to choose a default amount and limit it to `2**28`.

I'd expect:
* something like this to be used normally:
https://github.com/urllib3/urllib3/blob/db7d96b7cc2ddc184e459301956ff203ccd91ec6/src/urllib3/response.py#L811-L812

* and only if this condition is true, the amount to be limited to `c_int_max`
https://github.com/urllib3/urllib3/blob/db7d96b7cc2ddc184e459301956ff203ccd91ec6/src/urllib3/response.py#L787-L790

src/urllib3/response.py
```diff
@@ -876,39 +898,72 @@ def read(
...
+        if not data and len(self._decoded_buffer) == 0:
+            return data
+
+        # do not waste memory on buffer when not decoding
+        if not decode_content:
+            if self._has_decoded_content:
+                raise RuntimeError(
+                    "Calling read(decode_content=False) is not supported after "
+                    "read(decode_content=True) was called."
+                )
```
illia-v, line 952: The method name in the error will need to be changed

illia-v (commented):
Thanks for taking this on

src/urllib3/response.py
smason: That's an impressive expansion factor!  My use case didn't want enormous blocks of memory being pushed around unnecessarily, but don't think I thought it through.  They'd still exist in this case, just buffered (i.e. in `_decoded_buffer`), and hence it makes sense to just return them here when a caller asks for them.

Will update the code to actually return everything when it was requested!

Am still a bit worried about running out of memory (i.e. attempting to allocate very large `bytes` buffers for returning things like this), are there other bits of code you know of that deal with this that you know of?

src/urllib3/response.py
smason: I just wanted to do something relatively quick to code initially because I wasn't sure how likely this change was to be accepted...

The reference to `StringIO` seems weird to me as well; the types indicate that the method returns bytes, but that's not what you'd get from a `StringIO`.  I'm also wondering whether that check should be moved out into a top-method somewhere so the check wouldn't need to be done on every read.

smason:
@illia-v thanks for the review!

As per the initial comment I wanted to get `stream(None)` returning sensible sized blocks, I presume I should do that in a separate pull-request when this one gets accepted?

smason:
I've noticed that [`BufferedIOBase.read1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) (and `read`) are specified to use `size=-1` to indicate that "an arbitrary number of bytes are returned" whereas urllib3 uses `None` as the default value.

Is it worth changing this around at some point? that value gets propagated to the underlying stream and it feels correct to use the documented value for getting an arbitrary number of bytes.

src/urllib3/response.py
```diff
@@ -811,7 +811,9 @@ def _fp_read(
                 return self._fp.read1(size)
         elif (
             (size > c_int_max)
-            or (self.length_remaining and self.length_remaining > c_int_max)
+            or (
+                size < 0 and self.length_remaining and self.length_remaining > c_int_max
+            )
```
smason, line 816: Interested to know opinions about this deliberate change in semantics.

I feel as though the code should change to be doing the same chunked reading workaround when `length_remaining is None`.

smason:
@illia-v I think I've addressed everything you noted, please let me know if I've missed anything!

sethmlarson:
Thanks @smason! There's a lot happening in this one, would like to get some eyes on the one too.

smason:
@sethmlarson when I started I thought it would be an easy change but with all the fiddly edge cases it became easier to do a bit of a refactor

let me know if it would help to split things out somehow, e.g. rework the commits into more logical chunks, or maybe a few smaller pull-requests?

test/test_response.py
```diff
@@ -35,6 +35,8 @@
 from urllib3.util.response import is_fp_closed
 from urllib3.util.retry import RequestHistory, Retry
 
+OptInt: typing.TypeAlias = int | None
```
illia-v, line 38: Unlike other typing stuff not evaluated because of `from __future__ import annotations`, `int | None` here is so tests in Python 3.8 and 3.9 fail

illia-v (changes requested):
> I've noticed that [`BufferedIOBase.read1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) (and `read`) are specified to use `size=-1` to indicate that "an arbitrary number of bytes are returned" whereas urllib3 uses `None` as the default value.
> 
> Is it worth changing this around at some point? that value gets propagated to the underlying stream and it feels correct to use the documented value for getting an arbitrary number of bytes.

It makes sense to treat -1 the same as `None` in `read` and `read1`. If this is done, setting `-1` as a default shouldn't be a breaking change and may be accepted.
I can see you made changes toward this in `_fp_read`, but `read` still treats -1 and `None` differently when it decides whether to cache content, for example.
I'd suggest keeping the previous behavior of `_fp_read` when `amt` is -1, and moving the changes to a separate future PR.

> As per the initial comment I wanted to get `stream(None)` returning sensible sized blocks, I presume I should do that in a separate pull-request when this one gets accepted?

Let's discuss possible options in a separate issue.

> let me know if it would help to split things out somehow, e.g. rework the commits into more logical chunks, or maybe a few smaller pull-requests?

Please split changes unnecessary for adding `read1` (like refactoring of `_fp_read`, changes to `read`, and type hints in existing tests) into separate small PRs, it looks like some of them are more than refactoring so they may need more attention and separate changelog entries.

illia-v:
> As per the initial comment I wanted to get `stream(None)` returning sensible sized blocks, I presume I should do that in a separate pull-request when this one gets accepted?

Yes, please submit the change to `stream` separately.

If we use `read1` during streaming, won't it just work to pass a sensible size of a block instead of `None` as `amt`?
BTW, `amt` of [`HTTPResponse.stream`](https://urllib3.readthedocs.io/en/2.1.0/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream) defaults to 65536.

smason:
> > As per the initial comment I wanted to get `stream(None)` returning sensible sized blocks
> 
> If we use `read1` during streaming, won't it just work to pass a sensible size of a block instead of `None` as `amt`? BTW, `amt` of [`HTTPResponse.stream`](https://urllib3.readthedocs.io/en/2.1.0/reference/urllib3.response.html#urllib3.response.HTTPResponse.stream) defaults to 65536.

Yup, using `read1` is what I wanted to do.  Notably it would be able to yield byte sized chunks from endpoints like: https://httpbin.org/drip?duration=2&numbytes=8 which currently block until the stream closes before returning anything.

Having it default to 64KiB chunks would likely remain a good default; it's possible to get slightly higher performance with larger chunks but it's often not much beyond that point.

smason:
> > I've noticed that [`BufferedIOBase.read1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) (and `read`) are specified to use `size=-1` to indicate that "an arbitrary number of bytes are returned" whereas urllib3 uses `None` as the default value.
> > Is it worth changing this around at some point? that value gets propagated to the underlying stream and it feels correct to use the documented value for getting an arbitrary number of bytes.
> 
> It makes sense to treat -1 the same as `None` in `read` and `read1`. If this is done, setting `-1` as a default shouldn't be a breaking change and may be accepted. I can see you made changes toward this in `_fp_read`, but `read` still treats -1 and `None` differently when it decides whether to cache content, for example. I'd suggest keeping the previous behavior of `_fp_read` when `amt` is -1, and moving the changes to a separate future PR.

The conditionals were getting too complicated for me to have confidence in with the prior code.  But thinking about it again it is a breaking change and hence it makes sense to keep this separate.

illia-v:
Also, I forget to mention that I noticed that `read1` makes a `_raw_read` call even if `self._decoded_buffer` contains some bytes.
`BufferedIOBase.read1` [documents](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) that it reads with at most one call, so making 0 calls is allowed. What do you think about skipping `_raw_read` if `len(self._decoded_buffer) >= amt` or even always when the buffer is not empty?

smason:
> The conditionals were getting too complicated for me to have confidence in with the prior code. But thinking about it again it is a breaking change and hence it makes sense to keep this separate.

Been trying to do this sensibly, but `_fp_read` in HEAD:

https://github.com/urllib3/urllib3/blob/4dd4a824e60ba3bb81f319cf2b8e320aa8a87dc6/src/urllib3/response.py#L783-L793

looks broken for `amt=-1`. It'll just keep trying to read the entire stream ignoring the `max_chunk_amt` constraint.

Is that actually correct?  maybe it would make sense to fix that as an initial pull-request?  The aim would be to tidy up the handling of negative vs `None` `amt`s.

illia-v:
> Been trying to do this sensibly, but `_fp_read` in HEAD looks broken for `amt=-1`

It just doesn't expect that `-1` can be passed and neither `read` does. The response class implements `io.IOBase` which doesn't define a `read` method, so adding support for `-1` is a new feature not a bug fix 😁

If possible, let's finish this firstly keeping the current `read` logic when `-1` is passed and using the same for `read1`.

illia-v:
> It'll just keep trying to read the entire stream ignoring the `max_chunk_amt` constraint.

You can ignore `max_chunk_amt` for `read1`, but conditions related to `c_int_max` have to be taken into account.

illia-v:
Actually, looking at [the source code](https://github.com/python/cpython/blob/b11c443bb2ebfdd009e43ff208fa6324b658d15d/Lib/http/client.py#L456-L494) of CPython's `http.client.HTTPResponse.read`, `-1` is not expected to be passed there too.

`self._fp` is usually an instance of `http.client.HTTPResponse`.

smason:
> > Been trying to do this sensibly, but `_fp_read` in HEAD looks broken for `amt=-1`
> 
> It just doesn't expect that `-1` can be passed and neither `read` does. The response class implements `io.IOBase` which doesn't define a `read` method, so adding support for `-1` is a new feature not a bug fix 😁

Huh, good point, that makes things easier!  I will ignore handling negative lengths for now as they seem broken in `http.client`.

> Actually, looking at [the source code](https://github.com/python/cpython/blob/b11c443bb2ebfdd009e43ff208fa6324b658d15d/Lib/http/client.py#L456-L494) of CPython's `http.client.HTTPResponse.read`, `-1` is not expected to be passed there too.

Yup, passing `-1` will read until the end of the stream, bypassing the [check](https://github.com/python/cpython/blob/b11c443bb2ebfdd009e43ff208fa6324b658d15d/Lib/http/client.py#L469-L471) that `amt <= self.length`.  That's assuming that `http.client.HTTPResponse`'s `fp` is the expected `BufferedReader` that comes from a socket's `makefile`.

Can't find much in CPython's Github issues, about this.  There are some mentions of the io API and negative values in https://github.com/python/cpython/issues/67403 but nothing much closer.  The following MWE just hangs for me:
```python
con = http.client.HTTPConnection("httpbin.org")
con.request("GET", "/")
resp = con.getresponse()
assert resp.headers["connection"] == "keep-alive"
print(resp.read(-1))
```
which feels wrong, or am I missing something?  Is it worth raising an issue about this?

smason:
> > which feels wrong, or am I missing something?
> 
> The example prints HTML for me but with a significant delay, this feels wrong to me too. `http.client.HTTPResponse` [is said](https://docs.python.org/3/library/http.client.html#httpresponse-objects) to implement `io.BufferedIOBase` since Python 3.5, so I'd expect it to treat `None` and a negative number as the `read` argument the same.

Glad it's not just me misinterpreting APIs!  Have raised https://github.com/python/cpython/issues/112064 about this.

src/urllib3/response.py
```diff
@@ -768,7 +796,14 @@ def _fp_read(self, amt: int | None = None) -> bytes:
         """
         assert self._fp
         c_int_max = 2**31 - 1
-        if (
+        if one_read:
+            # StringIO doesn't like amt=None
+            if amt is None:
+                return self._fp.read1()
+            if amt > c_int_max and (util.IS_PYOPENSSL or sys.version_info < (3, 10)):
+                return self._fp.read1(c_int_max)
```
illia-v, line 804: I believe there is a possibility of an overflow error here because the `(self.length_remaining and self.length_remaining > c_int_max)` part of the following condition is skipped for `one_read`.
https://github.com/urllib3/urllib3/blob/4dd4a824e60ba3bb81f319cf2b8e320aa8a87dc6/src/urllib3/response.py#L771-L774

When `amt` is `None` and the content length is known, `http.client.HTTPResponse.read1` will [try to read](https://github.com/python/cpython/blob/9302f05f9af07332c414b3c19003efd1b1763cf3/Lib/http/client.py#L654-L655) the length amount of bytes. This can lead to an overflow error as described in the docstring of `_fp_read`.

Ideally (and probably required for 100% coverage), a test similar to existing `test_requesting_large_resources_via_ssl` has to check possible cases for `read1`. The problem with the existing test is that it causes timeouts in CI. Adding a similar one will worsen the situation. Although, a test for `read1` should be quicker because it will not read all 2 GiB.

src/urllib3/response.py
```diff
@@ -909,6 +946,41 @@ def read(
...
+        if decode_content is None:
+            decode_content = self.decode_content
+        if not decode_content and self._has_decoded_content:
+            raise RuntimeError(
+                "Calling read1(decode_content=False) is not supported after "
+                "read1(decode_content=True) was called."
+            )
+
+        if amt is not None and amt < len(self._decoded_buffer):
+            return self._decoded_buffer.get(amt)
```
illia-v, line 964: > Read with at most one underlying system call.  If at least one byte is buffered, return that instead.

This is from the [docstring](https://github.com/python/cpython/blob/9302f05f9af07332c414b3c19003efd1b1763cf3/Lib/http/client.py#L647-L648) of `http.client.HTTPResponse.read1`. I would follow the last sentence in urllib3 too.

So, if `self._decoded_buffer` is not empty, make no `self._raw_read` calls and:
* when `amt` is a number, return up to `amt` bytes from the buffer;
* when `amt` is `None`, return everything from the buffer.

@sethmlarson @pquentin what do you think about this?

illia-v (commented):
> I've squashed all my WIP changes into a single one that's just about adding read1 support.

Thanks, this is really helpful!

src/urllib3/response.py
smason: Avoiding calling into the kernel makes sense, I just copied the semantics of `read` which obviously don't apply here.  Have changed that condition to return early when possible.

Be interesting to see what other people think!

illia-v (commented):
This looks pretty good with the latest changes!

I merged the main branch with some updates fixing CI timeouts. Could you please add more tests to fix [the coverage misses](https://github.com/urllib3/urllib3/actions/runs/6876950948/job/18705065000)?
Also, please add a test similar to `test_requesting_large_resources_via_ssl` for `read1`,  and check the docs building failure in CI.

src/urllib3/response.py
```diff
@@ -909,6 +948,57 @@ def read(
...
+                )
+            if len(self._decoded_buffer) > 0:
+                if amt is None:
+                    return self._decoded_buffer.get_all()
+                return self._decoded_buffer.get(amt)
+        if amt == 0:
+            return b""
+
+        data = self._raw_read(amt, one_read=True)
+        if (not decode_content) or (not data and len(self._decoded_buffer) == 0):
```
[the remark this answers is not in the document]
smason: The decoder still needs to be flushed in this case doesn't it?  i.e. not sure if it's valid to bypass the `decode_content` check

smason:
@illia-v Have rebased and applied those two fixes

Will look at writing some tests to get the coverage back up

src/urllib3/response.py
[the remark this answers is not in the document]
illia-v: It looks like `read` doesn't get to flushing the decoder when there is no data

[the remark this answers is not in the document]
smason: > It looks like `read` doesn't get to flushing the decoder when there is no data

The code you linked to is in `read1`, are you referring to the original `read` method here?

I'm asking because I don't see any issue with my logic in `read1`.  I've just noticed it still contains some redundancies due to me basing it on the `read` implemention (two very similar looking calls to `self._decode()` and `self._decoded_buffer.put()`), but I don't think that affects correctness.

smason:
@illia-v coverage seems to be back up to 100% now

not sure why the docs build is giving:

> WARNING: py:meth reference target not found: http.client.HTTPResponse.read1

I can't find any special handling for `http.client.HTTPResponse.read` and yet it finds that method fine

illia-v:
> @illia-v coverage seems to be back up to 100% now

Awesome ✨

> not sure why the docs build is giving:
> 
> > WARNING: py:meth reference target not found: http.client.HTTPResponse.read1
> 
> I can't find any special handling for `http.client.HTTPResponse.read` and yet it finds that method fine

I am not sure why too, one assumption is that it happens because `http.client.HTTPResponse.read1` is not documented by CPython https://docs.python.org/3/library/http.client.html#httpresponse-objects

src/urllib3/response.py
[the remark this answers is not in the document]
illia-v: > > It looks like `read` doesn't get to flushing the decoder when there is no data
> 
> The code you linked to is in `read1`, are you referring to the original `read` method here?

Sorry for confusing you. I meant that the `read` method on the following lines returns an empty bytes object regardless of `decode_content`.
https://github.com/urllib3/urllib3/blob/bc41d52707c0f7717364bd571adc2d3f353ad60e/src/urllib3/response.py#L919-L920
It's like an early exit. So I assumed an `if not data: return data` condition will copy the behavior from `read` to `read1`. Do you see any issue in doing so?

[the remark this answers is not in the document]
smason: it looks like a bug to me...

`git blame` doesn't suggest anything obvious, https://github.com/urllib3/urllib3/commit/c35033f6cc54106ca66ef2d48a9e3564d4fb0e07 is when `read` was the last changed and was quite a large change.

maybe there's a test for that specific behavior you know of?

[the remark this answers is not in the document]
illia-v: I commented out the mentioned `read` lines, and many tests failed because `self._fp` was `None`, so `self._raw_read` returned `None`, and then joining `None` with some bytes was attempted

It's possible to reproduce a similar error by running `HTTPResponse().read1()`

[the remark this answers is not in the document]
smason: Why does `_raw_read` return `None`?! the return type is just `bytes`!

Maybe it should raise an exception when the stream has been closed?  Presume this is due to the stream getting implicitly closed when reading, but that feels as wrong.  Probably a another separate issue.

I also note that the whole content decoding feels backwards to me, why does `_init_decoder` keep getting called on every read which then pulls the header string out every time.  Surely that can be done once and then remembered — the headers shouldn't change should they?

smason:
It occurred to me that `read` should be using `read1` when a decoder is in use, i.e. doing something like `_raw_read(65536, one_read=True)` rather than just passing `amt` down the chain.

The user might have requested a specific number of bytes, but because of compression it likely won't take a read of that size from the socket to fulfill the request.

src/urllib3/response.py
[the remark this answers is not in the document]
illia-v: I agree that there is a room for tidying up this part of urllib3
Regarding `_raw_read`: returning `None` is a pity, maybe that's a consequence of `self._fp` defaulting to `None` in some cases like in tests, so that should be changed

illia-v:
> It occurred to me that `read` should be using `read1` when a decoder is in use, i.e. doing something like `_raw_read(65536, one_read=True)` rather than just passing `amt` down the chain.
> 
> The user might have requested a specific number of bytes, but because of compression it likely won't take a read of that size from the socket to fulfill the request.

I'd expect a user to choose `amt` based on `Content-Length` that provides a size of an encoded (possibly compressed) body

src/urllib3/response.py
```diff
@@ -280,6 +280,22 @@ def get(self, n: int) -> bytes:
...
+        if not buffer:
+            assert self._size == 0
+            return b""
+        if len(buffer) == 1:
+            result = buffer.pop()
+        else:
+            # BytesIO.writelines says "newlines are not added"
+            with io.BytesIO() as fd:
+                fd.writelines(buffer)
+                result = fd.getvalue()
```
illia-v, line 294: Can this be simplified?
```suggestion
            result = b"".join(buffer)
```

illia-v (dismissed):
LGTM, just one small comment

@sethmlarson @pquentin do you want to take a look at the PR before we merge it?

pquentin:
@illia-v I fully trust your judgment. Thanks!

sethmlarson (commented):
High-level question, how is `read1()` different than `read()`'s implementation except for passing `read1=True` into `_raw_read()`? If that is a potential implementation path (even if it's not as efficient due to not using `get_all()`) can that be considered?

I ask this because our HTTPResponse data reading is already so complicated, having two parallel and slightly different implementations makes me worried for future maintenance.

src/urllib3/response.py
```diff
@@ -770,8 +798,14 @@ def _fp_read(self, amt: int | None = None) -> bytes:
         c_int_max = 2**31 - 1
         if (
             (amt and amt > c_int_max)
-            or (self.length_remaining and self.length_remaining > c_int_max)
+            or (
+                amt is None
```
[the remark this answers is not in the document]
smason: @sethmlarson `read1` is a standard method provided by `io.BufferedReader`s which `http.client.HTTPResponse` also provides

this PR is mostly about getting `read1` supported, there are various cleanups that make sense once it's in there.  e.g. with compressed streams it doesn't make sense to me that `amt` is passed to `read` on the stream.  the decompressed data might be much larger and hence using `read1` would mean that the caller might get a response much earlier

> I'm not sure why `amt is None` was added here? The new branch `if one_read` doesn't use `amt`.

without it this "slow path" would get used for large responses even if `amt` is small

src/urllib3/response.py
smason: not sure why that didn't occur to me! think I was basing it in `get` and didn't notice that's all it was doing in the end

smason:
> > It occurred to me that `read` should be using `read1` when a decoder is in use, i.e. doing something like `_raw_read(65536, one_read=True)` rather than just passing `amt` down the chain.
> > The user might have requested a specific number of bytes, but because of compression it likely won't take a read of that size from the socket to fulfill the request.
> 
> I'd expect a user to choose `amt` based on `Content-Length` that provides a size of an encoded (possibly compressed) body

Speaking as a user of `requests` I don't think I've ever done that!  `HTTPResponse.stream` certainly doesn't do anything like that, so why would calling code (that presumably knows even less details about what's going on at the HTTP level) know what makes sense to pass here?

smason:
I've squashed and rebased again, hopefully that will lead to a cleaner merge

src/urllib3/response.py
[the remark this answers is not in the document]
illia-v: @sethmlarson in case urllib3 or http.client is going to `read1` 2 GiB or more and get the overflow error, the code under this if block prevents this by passing the maximum safe amount to `read1`

The condition would be more correct with the new `amt is None` check even without the `read1` addition. Not having it had no adverse effect previously, but now it's either `amt is None` has to be added here or code on lines 804–805 has to become more complex.

illia-v:
> High-level question, how is `read1()` different than `read()`'s implementation except for passing `read1=True` into `_raw_read()`?

I can see two changes besides calling `_raw_read`:
* if there is at least one byte in `self._decoded_buffer`, `read1` unlike `read` will always return the buffered content and make no `_raw_read` calls;
* `read1` doesn't try to call `_raw_read` until `amt` is reached.

pquentin:
> I can see two changes besides calling `_raw_read`

Should this be in a code comment?

smason:
> > I can see two changes besides calling `_raw_read`
> 
> Should this be in a code comment?

The docstring of `read1` points to [`io.BufferedReader.read1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) as guiding the semantics, i.e. performing at most one IO, and there's a comment near the top of `read1` saying "try and respond without going to the network".  I've tried to keep the code readable and saying much more would just seem to duplicate the code in comments, which is often counter productive in my experience.

I've tried to stay away from changing `read` as the changes were quickly getting out of hand.

illia-v (approved):

smason:
thanks for merging!  :tada:

PR #3187, opened 2023-11-09
Check _has_route *within* the test function

stefanor:
Test selection occurs after collection, so if we're in an environment where we don't want to make outbound network connections, defer the check to after selection.

This was the aim of #3166, but it wasn't tested in an environment where ConnectionError is raised.


illia-v (approved):
LGTM, thanks

Issue #3038, opened 2023-05-19
Add an HTTP/2-capable handler to the test suite


sethmlarson:
Currently our test suite is written with HTTP/1.1 in mind. We need to be able to receive and respond with HTTP/2 in order to implement the protocol.

## Requirements

- [ ] Add an HTTP/2 capable handler to the test suite (Hypercorn?)
- [ ] Have one test case that uses the new handler with HTTP/1.1

Mr-Sunglasses:
@sethmlarson Is this Issue open for work ?

Ousret:
@sethmlarson Are you open to switching tornado for another (mature) solution? They don't support h2 as you may already know and it may not be appropriate to make unstable arrangements for it.

sethmlarson:
@Ousret  Yes! @pquentin  suggested Hypercorn+Quart as a potential solution.

pquentin:
> @Ousret Yes! @pquentin suggested Hypercorn+Quart as a potential solution.

Indeed, if Tornado supported HTTP/2, all the `with_dummyserver` tests could be migrated essentially for free. But it doesn't, so we need an alternative.

It seems to me there are two general directions we could take:
 * Use a popular server like nginx, and either:
   1. use it as a reverse proxy to terminate HTTP/2 and speak to Tornado using HTTP/1.1
   2. use it as an actual HTTP/2 server and reimplement our [TestingApp](https://github.com/urllib3/urllib3/blob/4e9060bfa03af9d6057423b3a54f67e0cd2e7892/dummyserver/handlers.py#L69) using nginx.
 * Use a Python HTTP/2 server, which is simpler to implement and gives us more flexibility as we can use an API instead of launching a new process, and looking at the code and patching it is very useful: [we have done this in the past with Tornado](https://github.com/urllib3/urllib3/pull/1728).

The only disadvantage I can see with using a Python HTTP/2 server is that we would also use h2, so we won't discover interoperability problems as we would have with a different stack. But the benefits (flexibility, maintainability, speed of conversion) sound more important to me. (I keep mentioning a "Python HTTP/2 server", but I only know of one: the Hypercorn+Quart combo nicely maintained by @pgjones.)

What features would be ideally supported by this server?
 * Since we [generate many different certificates](https://github.com/urllib3/urllib3/blob/main/test/conftest.py), the server should start quickly while accepting arbitrary certificates on launch: https://pgjones.gitlab.io/hypercorn/how_to_guides/configuring.html#configuration-options
[...]

PR #3190, opened 2023-11-15
Run one test using Hypercorn

pquentin:
Closes https://github.com/urllib3/urllib3/issues/3038

### DIFF
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -13,6 +13,7 @@ from http.client import responses
 from io import BytesIO
 from urllib.parse import urlsplit
 
+from quart_trio import QuartTrio
 from tornado import httputil
 from tornado.web import RequestHandler
 
@@ -362,3 +363,11 @@ class TestingApp(RequestHandler):
 
     def shutdown(self, request: httputil.HTTPServerRequest) -> typing.NoReturn:
         sys.exit()
+
+
+hypercorn_app = QuartTrio(__name__)
+
+
+@hypercorn_app.route("/")
+async def hello():
+    return "Dummy Hypercorn server!"
--- a/dummyserver/handlers.py
+++ b/dummyserver/handlers.py
@@ -369,5 +369,5 @@ hypercorn_app = QuartTrio(__name__)
 
 
 @hypercorn_app.route("/")
-async def hello():
+async def hello() -> str:
     return "Dummy Hypercorn server!"
--- a/dummyserver/hypercornserver.py
+++ b/dummyserver/hypercornserver.py
@@ -58,6 +58,8 @@ def run_hypercorn_in_thread(
             shutdown_event,
         )
         ready_event.wait(5)
+        if not ready_event.is_set():
+            raise Exception("most likely failed to start server")
 
         yield
 
--- a/dummyserver/hypercornserver.py
+++ b/dummyserver/hypercornserver.py
@@ -28,7 +28,7 @@ async def _start_server(
     shutdown_event: threading.Event,
 ) -> None:
     async with trio.open_nursery() as nursery:
-        await nursery.start(
+        config.bind = await nursery.start(
             functools.partial(
                 hypercorn.trio.serve,
                 app,
@@ -70,7 +70,7 @@ def main() -> int:
     from .handlers import hypercorn_app
 
     config = hypercorn.Config()
-    config.bind = ["localhost:8080"]
+    config.bind = ["localhost:0"]
     ready_event = threading.Event()
     shutdown_event = threading.Event()
     trio.run(_start_server, config, hypercorn_app, ready_event, shutdown_event)
--- a/dummyserver/hypercornserver.py
+++ b/dummyserver/hypercornserver.py
@@ -5,6 +5,7 @@ import contextlib
 import functools
 import sys
 import threading
+from typing import Generator
 
 import hypercorn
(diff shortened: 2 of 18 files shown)


test/with_dummyserver/test_poolmanager.py
```diff
@@ -7,15 +7,32 @@
...
+)
+from dummyserver.tornadoserver import HAS_IPV6
 from urllib3 import HTTPHeaderDict, HTTPResponse, request
 from urllib3.connectionpool import port_by_scheme
 from urllib3.exceptions import MaxRetryError, URLSchemeUnknown
 from urllib3.poolmanager import PoolManager
 from urllib3.util.retry import Retry
 
 
+class TestHypercornPoolManager(HypercornDummyServerTestCase):
```
pquentin, line 23: The idea is to make `TestPoolManager` a subclass of `HypercornDummyServerTestCase`, but starting with a separate class was easier as my focus for this pull request was to make `HypercornDummyServerTestCase` work.

dummyserver/testcase.py
```diff
@@ -292,6 +294,26 @@ class IPv6HTTPDummyProxyTestCase(HTTPDummyProxyTestCase):
...
+class HypercornDummyServerTestCase:
+    host: str = "localhost"
+    port: str = "8080"
+    base_url: typing.ClassVar[str]
+
+    _stack: contextlib.ExitStack
+
+    @classmethod
+    def setup_class(cls) -> None:
+        with contextlib.ExitStack() as stack:
```
pquentin, line 306: It would have been easy to use a pytest fixture with the `run_hypercorn_in_thread` context manager, but I wanted to mirror the existing `TestCase` convention which makes it very easy to attach class variables like `base_url`.

dummyserver/testcase.py
```diff
@@ -292,6 +294,26 @@ class IPv6HTTPDummyProxyTestCase(HTTPDummyProxyTestCase):
     proxy_host_alt = "127.0.0.1"
 
 
+class HypercornDummyServerTestCase:
+    host: str = "localhost"
+    port: str = "8080"
```
pquentin, line 299: With Tornado, we use the port 0 (None in Python), but that does not seem to be possible with Hypercorn.

dummyserver/handlers.py
```diff
@@ -362,3 +363,11 @@ def redirect_after(self, request: httputil.HTTPServerRequest) -> Response:
 
     def shutdown(self, request: httputil.HTTPServerRequest) -> typing.NoReturn:
         sys.exit()
+
+
+hypercorn_app = QuartTrio(__name__)
```
pquentin, line 368: I think this should be in its own file.

dummyserver/hypercornserver.py
```diff
@@ -0,0 +1,81 @@
...
+    ) as executor:
+        future = executor.submit(
+            trio.run,
+            _start_server,  # type: ignore[arg-type]
+            config,
+            app,
+            ready_event,
+            shutdown_event,
+        )
+        ready_event.wait(5)
```
sethmlarson, line 60: Should we error out if the `ready_event` doesn't trigger in the timeout time?

dummyserver/testcase.py
sethmlarson: That's a shame, can we open an issue on Hypercorn for this feature and link back to this?

noxfile.py
```diff
@@ -10,7 +10,8 @@
 def tests_impl(
     session: nox.Session,
     extras: str = "socks,brotli,zstd",
-    byte_string_comparisons: bool = True,
+    # hypercorn dependency h2 compares bytes and strings
```
sethmlarson, line 13: Can we create an issue on h2 for this? Would be good to keep this one enabled.

test/with_dummyserver/test_poolmanager.py
sethmlarson: ++ I'm happy with this for the first case!

sethmlarson (commented):
This is awesome, I don't have any blocking comments only some suggestions. Thanks so much @pquentin!

dummyserver/testcase.py
pgjones: It should be, https://github.com/pgjones/hypercorn/issues/19

noxfile.py
pquentin: Thanks, I added a link to https://github.com/python-hyper/h2/issues/1236 in https://github.com/urllib3/urllib3/pull/3190/commits/be8f4e786ef67ac3c77e122dd9479da947901b79. Seems difficult to fix at first glance.

dummyserver/hypercornserver.py
pquentin: Thanks, addressed in https://github.com/urllib3/urllib3/pull/3190/commits/956aa8fbd4a9d7a9df740f0ff6cc647f43e8c653

dummyserver/testcase.py
pquentin: Oh, right, thanks for the tip! I've addressed in https://github.com/urllib3/urllib3/pull/3190/commits/931c749e74ee68f670b11f80c8232e39a036dd2c by sneakily abusying `task_status.started` to overwrite `config.bind`.

sethmlarson (approved):
LGTM, thanks Quentin!

PR #3191, opened 2023-11-15
Clarify Tornado involvement in dummyserver names

pquentin:
This is part of https://github.com/urllib3/urllib3/pull/3190, but I'd like to change the names in a separate commit to get a cleaner history. This is also the right pull request to bikeshed about the names.


sethmlarson (approved):
LGTM!

PR #3192, opened 2023-11-16
Use curl to test our Hypercorn server can speak HTTP/2

sethmlarson:
Talk about bootstrapping, we need a way to make sure Hypercorn can actually speak HTTP/2 before urllib3 is able to. Created this simple test using `curl` which easily handles that task for us.

### DIFF
--- a/test/with_dummyserver/test_http2.py
+++ b/test/with_dummyserver/test_http2.py
@@ -1,0 +1,25 @@
+from __future__ import annotations
+
+import subprocess
+
+from dummyserver.testcase import HypercornDummyServerTestCase
+
+
+class TestHypercornDummyServerTestCase(HypercornDummyServerTestCase):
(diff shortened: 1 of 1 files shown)


pquentin (approved):
LGTM if CI passes!

I suspect Windows will fail as curl isn't installed there: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

sethmlarson:
@pquentin Appears curl might be on Windows anyways? The tests are failing with failing to match something on the output (might be due to different backends w/ curl?)

pquentin:
The curl available on Windows VMs does not speak HTTP/2, which explains the failure.

sethmlarson:
Going to merge since none of the failures are related to this PR. We're passing on all the OSes.

PR #3195, opened 2023-11-16
Feature: Native Emscripten support

### LINKED ISSUE #2951: Native support for Pyodide, PyScript (emscripten?)
When PyScript was announced at PyCon US 2022 I immediately tried to add support for synchronous HTTP requests w/ Pyodide/PyScript into urllib3 but was unfortunately unsuccessful at the time. I can see that support for this has grown since that time :rocket:

There's a lot to look at and learn on this topic, this issue is currently a stub and needs more filling in.

General thoughts right now from me are having a new module `urllib3.contrib.fetch` which exposes an `HTTP[S]Connection` implementation that uses Pyodide's pyfetch (and the equivalent for emscripten) under the hood. Wouldn't need to support the full range of options available to urllib3 (things like custom certificates, etc) but I suspect we'd get a lot from just method, url, headers, body, and some other basic features.

The tougher parts of this would be testing and documentation, we can learn from the below projects potentially?

## Requirements

- [ ] Figure out how we can test urllib3 consistently in a browser
- [ ] Add functionality which uses the JavaScript fetch API instead of httplib in a third-party module (ie `urllib3.contrib.fetch`)
- [ ] Document the new feature and restrictions
- [ ] Verify that Requests works with urllib3 in the browser

## Related projects / issues

- https://github.com/koenvo/pyodide-http
- https://github.com/emscripten-forge/requests-wasm-polyfill
- https://github.com/koenvo/pyodide-http/issues/28
- https://github.com/koenvo/pyodide-http/issues/6
- https://github.com/pyodide/pyodide/issues/3160

cc @koenvo

joemarshall:
Fixes #2951 

This is a non-trivial PR, and I've included a bunch of tests due to that - it does the following:
1) Overrides `HTTPConnection` and `HTTPSConnection` with `EmscriptenHTTP(s)Connection` if you are running on Emscripten
2) Calls through to javascript XMLHTTPRequest for non-streaming requests and converts error codes & responses as best as possible back to urllib3.
3) In the case that a streaming request is made (i.e. without prefetch), it will if possible do a streaming request on a separate web-worker thread and using javascript fetch calls on that thread. For this to be possible, many stars need to be aligned (see below). In the case that the gods are not smiling upon you, it falls back to the non-streaming mode and uses synchronous XMLHTTPRequest. 

For streaming to work, you need:
a) To be in a web-worker
b) The browser page you are running on must be cross-origin isolated
c) The background worker thread to have launched. This only happens if you return control to javascript *after* you have called `import urllib3`. This requires you to be running async python code and to `await` a javascript promise that waits for long enough for the browser to create the worker. There is a function included to do this waiting like this:
`await urllib3.contrib.emscripten.wait_for_streaming_ready()`. Until this is called (or something else that awaits back into javascript for long enough), requests will fall back to synchronous XMLHTTPRequest.
d) Thanks to weirdnesses of the XMLHttpRequest specification, timeouts only work if you are in a web-worker. This is not fixable; if you're in the browser thread, you are at the mercy of the browser as to what the timeout is, and it may well be quite long. This sucks, but it is there in the W3C spec and browsers implement it like this.

You can see in the tests some example code that is run on a worker thread to successfully do a streaming request, 
```
import urllib3.contrib.emscripten.fetch
await urllib3.contrib.emscripten.fetch.wait_for_streaming_ready()
from urllib3.response import HTTPResponse
from urllib3.connection import HTTPConnection

conn = HTTPConnection(host,port)
method = "GET"
url = "/bigfile"
conn.request(method, url,preload_content=False)
response = conn.getresponse()
assert isinstance(response, HTTPResponse)
assert urllib3.contrib.emscripten.fetch._SHOWN_STREAMING_WARNING==False
data=response.data.decode('utf-8')
```

the key thing there is to do:
```
import urllib3.contrib.emscripten.fetch
await urllib3.contrib.emscripten.fetch.wait_for_streaming_ready()
```
and then everything should work fine and dandy.


joemarshall:
Incidentally, right now I think it doesn't work on node, but that could probably be fixed relatively easily if required.

.github/workflows/ci.yml
```diff
@@ -104,6 +108,9 @@ jobs:
       - name: "Install dependencies"
         run: python -m pip install --upgrade pip setuptools nox
 
+      - name: "Install chrome"
```
sethmlarson, line 111: nit: Capitalize Chrome

.github/workflows/ci.yml
```diff
@@ -104,6 +108,9 @@ jobs:
...
+        uses: browser-actions/setup-chrome@v1
```
sethmlarson, line 112: Use the commit sha instead of tag, Dependabot will keep it up-to-date.

noxfile.py
```diff
@@ -152,6 +155,72 @@ def lint(session: nox.Session) -> None:
     mypy(session)
 
 
+# TODO: node support is not tested yet - it should work if you require('xmlhttprequest') before
+# loading pyodide, but there is currently no nice way to do this with pytest-pyodide
+# because you can't override the test runner properties -
+# https://github.com/pyodide/pytest-pyodide/issues/118
+@nox.session(python="3.11")
```
sethmlarson, line 162: Do we have to restrict this to Python 3.11? Or will any Python 3 do? If there's a range that we can't allow can we filter that out within the task itself?

.github/workflows/ci.yml
```diff
@@ -81,6 +81,10 @@ jobs:
             os: ubuntu-20.04  # CPython 3.9.2 is not available for ubuntu-22.04.
             experimental: false
             nox-session: test-3.9
+          - python-version: "3.11"
```
sethmlarson, line 84: More a wondering, don't feel the need to implement this in this PR: are we going to need to test multiple browsers? How much does emscripten support differ across browsers, do you have insights into this @joemarshall?

noxfile.py
```diff
@@ -152,6 +155,72 @@ def lint(session: nox.Session) -> None:
...
+@nox.parametrize("runner", ["chrome"])
+def emscripten(session: nox.Session, runner: str) -> None:
+    """Test on Emscripten with Pyodide & Chrome"""
+    session.install("build","pyodide_build")
```
sethmlarson, line 166: We'll want to pin these dependencies to specific versions. Perhaps we can capture this in a separate `emscripten-requirements.txt` file?

noxfile.py
```diff
@@ -152,6 +155,72 @@ def lint(session: nox.Session) -> None:
...
+            session,
+            "emscripten-test",
+            pytest_extra_args=[
+                "--rt",
+                "chrome-no-host",
+                "--dist-dir",
+                str(dist_dir),
+                "test",
+            ],
+        )
```
sethmlarson, line 221: We should fail if we receive an unknown runner (ie `else: raise ValueError(...)`).

pyproject.toml
```diff
@@ -49,6 +49,11 @@ zstd = [
 socks = [
   "PySocks>=1.5.6,<2.0,!=1.5.7",
 ]
+emscripten_test = [
```
sethmlarson, line 52: Let's move these into a `emscripten-requirements.txt`

src/urllib3/contrib/emscripten/__init__.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+
+import urllib3.connection
+
+from ...connectionpool import HTTPConnectionPool, HTTPSConnectionPool
+from .connection import EmscriptenHTTPConnection, EmscriptenHTTPSConnection
+
+
+def _override_connections_for_emscripten() -> None:
```
sethmlarson, line 9: To match the terminology we previously used, let's name this `inject_into_urllib3()`.

test/contrib/emscripten/conftest.py
```diff
@@ -0,0 +1,180 @@
...
+        self.port = port
+        self.selenium = selenium
+
+    def run_webworker(self, code: str) -> Any:
+        if isinstance(code, str) and code.startswith("\n"):
+            # we have a multiline string, fix indentation
+            code = textwrap.dedent(code)
+
+        return self.selenium.run_js(
+            """
```
sethmlarson, line 48: Since we're using .format() already, can we change this into an f-string?

test/contrib/emscripten/conftest.py
```diff
@@ -0,0 +1,180 @@
...
+class ServerRunnerInfo:
+    def __init__(self, host: str, port: int, selenium: Any) -> None:
+        self.host = host
+        self.port = port
+        self.selenium = selenium
+
+    def run_webworker(self, code: str) -> Any:
+        if isinstance(code, str) and code.startswith("\n"):
+            # we have a multiline string, fix indentation
+            code = textwrap.dedent(code)
```
sethmlarson, line 45: Is `dedent` safe to call always?

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+    copy_files_to_pyodide,
+)
+
+from .conftest import PyodideServerInfo, ServerRunnerInfo  # noqa: E402
+
+# make our ssl certificates work in chrome
+pytest_pyodide.runner.CHROME_FLAGS.append("ignore-certificate-errors")
+
+
+@copy_files_to_pyodide(file_list=[("dist/*.whl", "/tmp")], install_wheels=True)  # type: ignore[misc]
```
sethmlarson, line 24: This decorator is repeated a bunch, probably worth making its own? Also `/tmp` is Linux-specific, might need to use `tempfile`?

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+def test_index(selenium: typing.Any, testserver_http: PyodideServerInfo) -> None:
+    @run_in_pyodide  # type: ignore[misc]
```
sethmlarson, line 26: Is it possible to turn `@run_in_pyodide` into a pytest fixture that executes the pytest function itself (and injects selenium, testserver, etc?)

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+        try:
+            conn.request(method, url)
+            _ = conn.getresponse()
+            pytest.fail("No response, should throw exception.")
+        except BaseException as ex:
+            assert isinstance(ex, urllib3.exceptions.ResponseError)
+
+    import socket
+
+    sock = socket.socket()
```
sethmlarson, line 93: We have a helper for this `test/port_helper` iirc?

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+
+@copy_files_to_pyodide(file_list=[("dist/*.whl", "/tmp")], install_wheels=True)  # type: ignore[misc]
+def test_timeout_in_worker(
+    selenium: typing.Any,
+    testserver_http: PyodideServerInfo,
+    run_from_server: ServerRunnerInfo,
+) -> None:
+    worker_code = f"""
+        import micropip
+        await micropip.install('http://{testserver_http.http_host}:{testserver_http.http_port}/wheel/urllib3-2.0.7-py3-none-typing.Any.whl',deps=False)
```
sethmlarson, line 151: Why is `.Any.whl` used as the suffix?

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+
+@copy_files_to_pyodide(file_list=[("dist/*.whl", "/tmp")], install_wheels=True)  # type: ignore[misc]
+def test_index_https(selenium: typing.Any, testserver_http: PyodideServerInfo) -> None:
+    @run_in_pyodide  # type: ignore[misc]
+    def pyodide_test(selenium, host: str, port: int) -> None:  # type: ignore[no-untyped-def]
+        from urllib3.connection import HTTPSConnection
+        from urllib3.response import HTTPResponse
+
+        conn = HTTPSConnection(host, port)
+        method = "GET"
```
sethmlarson, line 183: This applies elsewhere too, but it's okay to inline the method, path, url, etc into `conn.request()` directly.

test/contrib/emscripten/test_emscripten.py
```diff
@@ -0,0 +1,381 @@
...
+        path = "/"
+        url = f"https://{host}:{port}{path}"
+        conn.request(method, url, preload_content=True)
+        response = conn.getresponse()
+        assert isinstance(response, HTTPResponse)
+        data = response.data
+        assert data.decode("utf-8") == "Dummy server!"
+        # no console warnings because we didn't ask it to stream the response
+        assert not urllib3.contrib.emscripten.fetch._SHOWN_STREAMING_WARNING
```
sethmlarson, line 215: Is there a way to capture and assert on `console.log` calls from pyodide?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+    _fetcher = _StreamingFetcher()
+else:
+    _fetcher = None
+
+
+def send_streaming_request(request: EmscriptenRequest) -> EmscriptenResponse | None:
+    if _fetcher and streaming_ready():
+        return _fetcher.send(request)
+    else:
+        _show_streaming_warning()
```
sethmlarson, line 387: Instead of warning about falling back on non-streaming, what do you think about erroring out with a link to documentation about how to properly do streaming w/ web workers? Then users will always get the behavior they are expecting.

The same might be applicable to timeouts? Although we might need people to opt-in to "browser controlled timeouts" via `urllib3.contrib.emscripten.BROWSER_TIMEOUT` sentinel or the like?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+
+class _StreamingError(_RequestError):
+    pass
+
+
+class _TimeoutError(_RequestError):
+    pass
+
+
+_STREAMING_WORKER_CODE = """
```
sethmlarson, line 73: This code might be better off in a `.js` file so we get syntax highlighting and then load the file via importlib resources? Also are there lightweight JS formatters we can use from Python-land?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+import js  # type: ignore[import]
+from pyodide.ffi import JsArray, JsException, JsProxy, to_js  # type: ignore[import]
+
+if TYPE_CHECKING:
+    from typing_extensions import Buffer
+
+from .request import EmscriptenRequest
+from .response import EmscriptenResponse
+
+"""
```
sethmlarson, line 39: Instead of always ignoring `User-Agent`, we should document that we don't send our default `User-Agent` when using emscripten however we still want to respect users preferences if /they/ want to send a `User-Agent`, in that case we document the potential downsides of doing so.

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+        self.read_pos += ret_length
+        return ret_length
+
+
+class _StreamingFetcher:
+    def __init__(self) -> None:
+        # make web-worker and data buffer on startup
+        self.streaming_ready = False
+
+        dataBlob = js.Blob.new(
```
sethmlarson, line 256: I don't know if there are conventions for this already, but it might be nice to have JS magic objects named with a `js_` prefix or the like? So we know when we're passing things back and forth between Python and JS-land.

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+            [_STREAMING_WORKER_CODE], _obj_from_dict({"type": "application/javascript"})
+        )
+
+        def promise_resolver(res: JsProxy, rej: JsProxy) -> None:
```
sethmlarson, line 260: Let's not shorten these names, we're already unfamiliar with Promises as Python devs :P

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+                {
+                    "buffer": shared_buffer,
+                    "url": absolute_url,
+                    "fetchParams": fetch_data,
+                }
+            )
+        )
+        # wait for the worker to send something
+        js.Atomics.wait(int_buffer, 0, 0, timeout)
+        if int_buffer[0] == 0:
```
sethmlarson, line 302: Is this zero a magic value too? Should it be stored in a similar way to the others?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+            raise _TimeoutError(
+                "Timeout connecting to streaming request",
+                request=request,
+                response=None,
+            )
+        if int_buffer[0] == SUCCESS_HEADER:
```
sethmlarson, line 308: nit: `elif`

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,477 @@
...
+                headers=response_obj["headers"],
+                body=io.BufferedReader(
+                    _ReadStream(
+                        int_buffer,
+                        byte_buffer,
+                        request.timeout,
+                        self._worker,
+                        response_obj["connectionID"],
+                    ),
+                    buffer_size=1048576,
```
sethmlarson, line 331: Where does this value come from?

sethmlarson (commented):
All I can say is wow! This is incredible, thank you @joemarshall for this work you've done. I'm super excited to see this land, this has been something I've wanted since I first saw pyodide and pyscript and all of that hype happening.

I gave this as quick an initial review as I could muster since it is quite large (but tbh, I don't think you can cut this particular PR down any!)

A high level comment, this might benefit a ton from having a PR adding the documentation concurrent to this PR since this is all new to us and we want to make sure our users get all the details and caveats while we're putting this together.

I am certain there will be more reviews so I hope you are okay with that, but I want to be cognizant of this effort not stalling out too. Please let us know if how we're receiving this PR is getting in the way of landing this.

Thanks again! :rocket:

src/urllib3/contrib/emscripten/fetch.py
joemarshall: It's an arbitrary value for thread buffer transfer that is a decently big buffer that many files will fit in, but not so big that it takes up tons of memory. Could be set as a constant somewhere though?

src/urllib3/contrib/emscripten/fetch.py
joemarshall: I've moved it out to .js and added eslint and prettier to the pre-commit for .js files only.

src/urllib3/contrib/emscripten/fetch.py
joemarshall: I think this is a bad idea - it breaks firefox if anyone uses a library that sets its own user agent. `requests` being the obvious one. It means that that library will just flat-out not work on e.g. github pages.

src/urllib3/contrib/emscripten/fetch.py
joemarshall: I think the risk here is that if people use library code that itself uses urllib3 and sets it it to stream by default, they won't be happy if their use case fails in e.g. pyscript or pyodide without web-workers even when it doesn't absolutely need streaming to function. Streaming with web-workers requires cross-site isolation which is a bit fiddly, especially for non-trivial websites, and also doesn't work easily on some big hosting providers including github pages.

I think this is all a transitional thing that will keep things running until the webassembly javascript promise integration (JSPI) proposal comes to all browsers, which allows webassembly code to suspend waiting on a javascript promise, which fixes all this stuff, but until then, I'd rather do whatever breaks least existing code.

( JSPI = https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md )

test/contrib/emscripten/conftest.py
joemarshall: yes - it's needed because otherwise you have to write any code strings on the first column and it messes up the indentation of your program.

.github/workflows/ci.yml
joemarshall: It should test on firefox, but there's an outstanding issue in pytest-pyodide that means you can't turn off certificate verification in firefox easily.

I started with 3.11 because the stable version of pyodide I'm testing against uses it, and I find everything is easier if you don't mix and match host and pyodide versions of python too much.

test/contrib/emscripten/test_emscripten.py
joemarshall: have done. /tmp is the destination on the emscripten side, so is right.

test/contrib/emscripten/test_emscripten.py
joemarshall: You can often use it like that, but I couldn't get it to work here - it requires all the arguments to be simple and pickleable, and there are other type constraints. I haven't delved into it as @run_in_pyodide is non-trivial and I don't want to mess with it when using an inner function with basic types is functional.

test/contrib/emscripten/test_emscripten.py
joemarshall: oops, copy and paste error, fixed now

test/contrib/emscripten/test_emscripten.py
joemarshall: done

test/contrib/emscripten/test_emscripten.py
joemarshall: yes. Can override it. It's a pain because it has to go inside each test for boring reasons, but I've put it in.

noxfile.py
joemarshall: It needs a python supported by `pytest-pyodide` which right now is 3.11

src/urllib3/contrib/emscripten/fetch.py
joemarshall: good idea, is done.

src/urllib3/contrib/emscripten/fetch.py
joemarshall: Maybe an emscripten specific setting api to enable user-agent changes or something.

joemarshall:
@sethmlarson Fixed most of those things above - a couple I've left open with comments on, as I think discussion might be needed as to what to do in those cases.

I put in firefox tests because that was trivial btw, as it appears headless firefox by default ignores certificates in selenium environments or something, so https does work there without needing messing around.

.github/workflows/ci.yml
```diff
@@ -104,6 +108,12 @@ jobs:
       - name: "Install dependencies"
         run: python -m pip install --upgrade pip setuptools nox
 
+      - name: "Install Chrome"
+        uses: browser-actions/setup-chrome@11cef13cde73820422f9263a707fb8029808e191
```
pquentin, line 112: Can you please also add the version as a comment, as in the "Setup Python" and "Checkout repository" steps above? This will also be updated by dependabot, and allows humans to know what version we're using.

emscripten-requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest-pyodide>=0.54.0
+pyodide-build>=0.24.0
+webdriver-manager>=4.0.0
```
pquentin, line 3: We tend to pin instead of using lower bounds. This sounds especially important as the pyiodide dependencies are still in the 0.x range. What do you think?

noxfile.py
```diff
@@ -54,6 +57,7 @@ def tests_impl(
         "--strict-config",
         "--strict-markers",
         *(session.posargs or ("test/",)),
+        *pytest_extra_args,
```
pquentin, line 60: While a good idea, I'd like to be consistent: either we migrate integration and byte_string_comparisons to pytest_extra_args, or we name this emscripten_extra_args or pyiodide_extra_args.

Also it should be *before* `session.posargs` if possible to allow manual override of options using `nox -s  session -- posarg1 posarg2`.

src/urllib3/__init__.py
```diff
@@ -202,3 +203,9 @@ def request(
         timeout=timeout,
         json=json,
     )
+
+
+if sys.platform == "emscripten":
+    from .contrib.emscripten import inject_into_urllib3  # noqa: 401
+
+    inject_into_urllib3()
```
pquentin, line 211: Have we considered *not* doing this by default?

src/urllib3/contrib/emscripten/connection.py
```diff
@@ -0,0 +1,216 @@
...
+from ...response import BaseHTTPResponse
+from ...util.connection import _TYPE_SOCKET_OPTIONS
+from ...util.timeout import _DEFAULT_TIMEOUT, _TYPE_TIMEOUT
+from ...util.url import Url
+from .fetch import _RequestError, _TimeoutError, send_request, send_streaming_request
+from .request import EmscriptenRequest
+from .response import EmscriptenHttpResponseWrapper
+
+
+class EmscriptenHTTPConnection(HTTPConnection):
```
pquentin, line 19: Would it be possible to inherit from `BaseHTTPConnection` instead?

pquentin (commented):
Here's a few general comments about the integration in the codebase. (I haven't looked at the actual important code, sorry.)

src/urllib3/__init__.py
sethmlarson: Yeah that seems correct, we can document that you need to call this for Emscripten support.

src/urllib3/contrib/emscripten/fetch.py
sethmlarson: We can always go that route later if we decide it's too much of a hassle, when you set a default behavior it's forever :) From my POV, almost all packages that use urllib3 don't use the streaming functionality and the ones that do either let you control that (like Requests) or are fetching gigantic files likely to be written to the disk (something you wouldn't do in a browser).

If we wait for JSPI we can then offer streaming functionality safely without this behavior. What do you think about this approach?

src/urllib3/__init__.py
joemarshall: Personally I really think it makes more sense to make it work out of the box on emscripten. Otherwise you end up with the situation that someone downloads a library which depends on requests or urllib3 and it inexplicably just doesn't work at all, and the only way they can make it work is to know that they need to read the dependency list, find urllib3 on it, and then read the urllib3 documentation to find that while urllib3 does support emscripten, you need to call a special function before urllib3 works on emscripten, and without it you just get a range of obscure not implemented errors inside socket We're already seeing similar with the current pyodide-http monkeypatch module, where you end up putting in code downstream for emscripten support that basically just looks for sys.platform and calls pyodide-http to monkeypatch everything.

emscripten-requirements.txt
joemarshall: Yeah, pinning should be fine, given urllib3 its a pure python module it's only a testing thing, and I don't think pyodide is likely to break any of this in the short term at least.

src/urllib3/contrib/emscripten/connection.py
joemarshall: Can do. I only did it this way because it reduced the amount of messing with the existing codebase. No problems switching it.

.github/workflows/ci.yml
joemarshall: Think I did this right now

src/urllib3/__init__.py
joemarshall: One possible thing to do would be to add a check for whether sockets are implemented into this code just in case something changes in the range of environments targeted by emscripten. Realistically though, sockets support is really unlikely in browser webasembly, so it is likely we'll remain in the situation where only http based protocols are available for at least the foreseeable future.

noxfile.py
pquentin: (I will do this later, let's focus on the important things)

src/urllib3/contrib/emscripten/connection.py
pquentin: Can you clarify what amount of messing with the code base we're talking about?

src/urllib3/__init__.py
pquentin: I don't have a strong opinion, just thought it should be considered.

What are the arguments in favor of opting in? Explicit is better than implicit, mostly. Having urllib3 work magically could give the impression that sockets *are* supported and set unrealistic expectations. More hypothetically, having an explicit API gives us more options.

Your arguments against opting in do make sense though!

src/urllib3/contrib/emscripten/connection.py
joemarshall: I did it, and actually it seems I only had to change one bit in the end - in BaseHttpResponse, where something expects an HTTPConnection, not a BaseHTTPConnection

pquentin:
Make sure you have https://github.com/urllib3/urllib3/pull/3204 in your branch otherwise lots of tests will fail.

joemarshall:
@pquentin just merged again, tests okay here, cross fingers it is happy in CI now.

joemarshall:
I added a test for requests working in pyodide. It relies on internet access, but there's a similar one in the main CI. Hope that's okay.

joemarshall:
I put coverage support for the pyodide tests in, coverage is not 100% yet, but much better than it was.

src/urllib3/__init__.py
sethmlarson: We can stick to this for now, I want to poke around for a bit once this is merged too.

sethmlarson:
@joemarshall Is there a way we can ignore coverage for this module for now? I would prefer to work on it over time than try to get it all fixed in one PR.

I'm pretty happy to merge this as-is and iterate (hopefully you are still interested in helping with that @joemarshall ?) We'll still need docs and more tweaking before we can make this as stable. Until then it'll be "experimental with no guarantees". Very exciting stuff! 🚀

skydiver1980 (approved):

joemarshall:
@sethmlarson I'm happy for it to merge as it is now assuming the bugfixes I pushed today haven't broken any non-emscripten stuff (don't think they will, but tests are still running) - the coverage is alright now anyway, and I found and fixed a number of subtle little bugs where things didn't behave quite right on emscripten compared to standard urllib3.

sethmlarson (approved):
Awesome, I think I'm happy enough to hit the merge button on this one. I'm going to hit approve and then submit some follow-up PRs to tidy things up and start working on documentation. Thanks @joemarshall!

sethmlarson:
@joemarshall Also https://github.com/urllib3/urllib3/issues/2951 is a bounty issue for $500, if you are eligible to receive a bounty you can submit an expense to our Open Collective. Please include a link to this PR and the issue in your expense :) Thanks again!

psymbio:
@joemarshall 

How do I make this work for Jupyter Lite, currently I'm trying to run the following in a cell:

```python
import micropip
await micropip.install("urllib3", verbose=True)
import urllib

import urllib3.contrib.emscripten.fetch
await urllib3.contrib.emscripten.fetch.wait_for_streaming_ready()
from urllib3.response import HTTPResponse
from urllib3.connection import HTTPConnection

conn = HTTPConnection(host,port)
method = "GET"
url = "/bigfile"
conn.request(method, url,preload_content=False)
response = conn.getresponse()
assert isinstance(response, HTTPResponse)
assert urllib3.contrib.emscripten.fetch._SHOWN_STREAMING_WARNING==False
data=response.data.decode('utf-8')
```

But I get: `ModuleNotFoundError: No module named 'urllib3.contrib.emscripten'` is there a wheel I need to install?

pquentin:
@psymbio This work was not released yet. Usually this means you should install urlllib3 from git, which is easy with pip, but micropip cannot support that: https://github.com/pyodide/micropip/issues/77. That said I would be glad if we could get some testing before the release, so hopefully @joemarshall can provide a way for you to test this anyway.

joemarshall:
@psymbio try installing this wheel in jupyterlite, with any luck it should work...

[urllib3-2.1.0-py3-none-any.zip](https://github.com/urllib3/urllib3/files/13591691/urllib3-2.1.0-py3-none-any.zip)

psymbio:
Okay, so here are some testing results:

Testing Environment: Jupyter Lite (https://jupyter.org/try-jupyter/lab/)

Test Case 1:
```python
import micropip
await micropip.install('urllib3', keep_going=True)
import urllib3

await micropip.install("ssl")
import ssl
http = urllib3.PoolManager()
url = "https://swapi.dev/api/people/1"
response = http.request('GET', url)
print("Status code:", response.status)
print("Response data:", response.data.decode('utf-8'))
```
Fails with: `MaxRetryError: HTTPSConnectionPool(host='swapi.dev', port=443): Max retries exceeded with url: /api/people/1 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x219da68>: Failed to establish a new connection: [Errno 50] Protocol not available'))`

Test Case 2:
```python
import micropip
await micropip.install('https://raw.githubusercontent.com/psymbio/tiktoken_rust_wasm/main/packages/urllib3/urllib3-2.1.0-py3-none-any.whl', keep_going=True)
import urllib3

urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
http = urllib3.PoolManager()
url = "https://swapi.dev/api/people/1"
response = http.request('GET', url)
assert response.status == 200
print("Response data:", response.data.decode('utf-8'))
```
This passes.

However, I think I'm still stuck on how to test this for streaming requests/responses and this test case stated here: https://github.com/urllib3/urllib3/pull/3195#issue-1997548433 > @joemarshall can this be re-written to be able to test?

joemarshall:
The test code at the bottom will stream if possible (if you're serving jupyterlite with cross origin isolation (i.e. serving everything with headers set on the server as I do in the first snippet below:
```
self.send_header('Cross-Origin-Opener-Policy','same-origin')
self.send_header('Cross-Origin-Embedder-Policy','require-corp')
```

Test code:
```
URLLIB3_HOST="localhost"
URLLIB3_PORT=8000

URLLIB3_WHEEL=f"http://{URLLIB3_HOST}:{URLLIB3_PORT}/urllib3-2.1.0-py3-none-any.whl"

import micropip
await micropip.install(URLLIB3_WHEEL)

import urllib3.contrib.emscripten.fetch
await urllib3.contrib.emscripten.fetch.wait_for_streaming_ready()

from urllib3.connection import HTTPConnection
# streaming request - get the wheel because we know it exists!
conn=HTTPConnection(URLLIB3_HOST,URLLIB3_PORT)
conn.request("GET", URLLIB3_WHEEL,preload_content=False)
response = conn.getresponse()
data=response.data
# check we have a wheel zip
assert(data[0:2]==b'PK')

if urllib3.contrib.emscripten.fetch._SHOWN_STREAMING_WARNING==False:
    print("Streamed")
else:
    print("didn't stream, is page cross-origin isolated?")
```

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,413 @@
...
+                js_resolve_fn(e)
+
+            def onErr(e: JsProxy) -> None:
+                js_reject_fn(e)  # Defensive: never happens in ci
+
+            self.js_worker.onmessage = onMsg
+            self.js_worker.onerror = onErr
+
+        js_data_url = js.URL.createObjectURL(js_data_blob)
+        self.js_worker = js.globalThis.Worker.new(js_data_url)
```
hoodmane, line 210: Is there a reason you say `js.globalThis` here? I would think that's the same as `js`...

pquentin: @joemarshall any idea?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -0,0 +1,413 @@
...
+from __future__ import annotations
+
+import io
+import json
+from email.parser import Parser
+from importlib.resources import files
+from typing import TYPE_CHECKING, Any
+
+import js  # type: ignore[import]
+from pyodide.ffi import JsArray, JsException, JsProxy, to_js  # type: ignore[import]
```
pmp-p, line 35: This is not "Native Emscripten support" this is pyodide support. You could have fenced it.

Issue #3197, opened 2023-11-17
Migrate run_server_in_thread and run_server_and_proxy_in_thread to Hypercorn


pquentin:
The first blocker to add HTTP/2 support to urllib3 is to have a test server that supports HTTP/2. Since https://github.com/tornadoweb/tornado/issues/1438, we're migrating to Hypercorn. This issue is about migrating all uses of `run_tornado_loop_in_thread` in test/conftest.py.

- [x] #3198 
- [ ] #3222

sethmlarson:
From our discussion it seems like proxies might be much harder than a regular server. Let's split the proxy issue off?

PR #3198, opened 2023-11-17
Migrate run_server_in_thread to Hypercorn

pquentin:
Depends on #3193, the only commit to review is https://github.com/urllib3/urllib3/commit/7c8c664b1e2afded8b63052da766f71d28869a74 (it was surprisingly easy).

Relates #3197 (but does not close it, as proxy tests still rely on Tornado).

### DIFF
--- a/dummyserver/app.py
+++ b/dummyserver/app.py
@@ -1,0 +1,98 @@
+from __future__ import annotations
+
+import contextlib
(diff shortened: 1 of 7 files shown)


sethmlarson (approved):
This is awesome! Thanks Quentin!

Issue #3196, opened 2023-11-17
Migrate HTTPDummyServerTestCase to Hypercorn


pquentin:
The first blocker to add HTTP/2 support to urllib3 is to have a test server that supports HTTP/2. Since [Tornado does not have plans to support HTTP/2](https://github.com/tornadoweb/tornado/issues/1438), we're migrating to Hypercorn. This issue is about migrating all child classes of HTTPDummyServerTestCase.

## HTTPDummyServerTestCase

- [x] #3193
- [x] #3209
- [x] #3205 
- [x] #3208
- [x] #3210
- [x] #3212
- [x] #3213

## HTTPSDummyServerTestCase

- [x] #3214 
- [x] #3217 
- [x] #3211

## IPv6HTTPDummyServerTestCase

- [x] #3215

## HTTPDummyProxyTestCase and IPv6HTTPDummyProxyTestCase

- [x] #3226

## PyodideDummyServerTestCase

- [x] #3230

sethmlarson:
Congrats @pquentin, with [this PR now merged](https://github.com/urllib3/urllib3/pull/3253) can this issue be closed? :partying_face:

pquentin:
Yes!

PR #3202, opened 2023-11-20
Move test_skip_header to test/test_connection.py

pquentin:
It relies on sending requests that violate the HTTP spec, and that are refused by h11 as used by Hypercorn, our new test server. Thankfully, we don't need the return of the server, looking at what urllib3 would have sent is enough.

Relates #3196


sethmlarson (approved):
LGTM

sethmlarson:
Failures are unrelated.

PR #3203, opened 2023-11-20
don't make changes to immutable headers mapping

dimbleby:
just the first commit from #3134, per https://github.com/urllib3/urllib3/pull/3134#issuecomment-1819787067


illia-v (approved):
Thanks @dimbleby!

PR #3206, opened 2023-11-22
Fix flaky time-based socketlevel tests

pquentin:
Instead, rely on the expected bytes to know when to stop calling recv().

Closes https://github.com/urllib3/urllib3/issues/3164


illia-v (approved):
Thanks for reducing the number of failing tests!

PR #3213, opened 2023-11-26
Migrate final connection pool tests to Hypercorn

pquentin:


sethmlarson (approved):
:shipit:

pquentin:
Will need to investigate this:

```
FAILED test/with_dummyserver/test_connectionpool.py::TestFileBodiesOnRetryOrRedirect::test_redirect_put_file - urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=39429): Max retries exceeded with url: /redirect?target=/echo&status=307 (Caused by ResponseError('too many redirects'))
```

pquentin:
OK I've fixed a timeout issue where we were hardcoding `0.1` instead of using LONG_TIMEOUT (see https://github.com/urllib3/urllib3/issues/1706 for more details). The async generator warning is quite mysterious to me and has not reoccurred. I would suggest merging like this, I'll keep an eye on those tests.

Please take another look.

sethmlarson (approved):

PR #3214, opened 2023-11-27
Migrate supported_tls_versions to Hypercorn

pquentin:


sethmlarson (approved):
Nice! Happy to see an HTTPS test case, we'll be using these plenty for HTTP/2 👀

PR #3216, opened 2023-11-27
make HTTPResponse.stream use read1 when amt=None

smason:
Following up on #3186 (and suggested in #2125) this uses the new `read1` method when streaming a non-chunked response and `amt=None`.

This would mean that https://github.com/psf/requests/issues/5536 could also be closed.


illia-v:
@saschpe thanks! The test failures look to be a result of this PR.

The case is pretty similar to hanging you described in https://github.com/python/cpython/issues/112064.
I guess this while loop iterates many times needlessly because `self._fp` is not closed when all data is read:
https://github.com/urllib3/urllib3/blob/4ae4b7168bf7271cbf8dd56e689d75c34e3fe50d/src/urllib3/response.py#L1022

src/urllib3/response.py
```diff
@@ -827,6 +827,8 @@ def _fp_read(
                 del data  # to reduce peak memory usage by `max_chunk_amt`.
             return buffer.getvalue()
         elif read1:
+            # work around: https://github.com/python/cpython/issues/112064
+            amt = amt or self.length_remaining
             return self._fp.read1(amt) if amt is not None else self._fp.read1()
```
illia-v, line 832: I don't see why this is needed. The mentioned issue is related to passing `-1` to `HTTPResponse.read`, this part of code doesn't use the method.

src/urllib3/response.py
```diff
@@ -1020,10 +1022,13 @@ def stream(
...
-                data = self.read(amt=amt, decode_content=decode_content)
-
-                if data:
-                    yield data
+                if amt is None:
+                    data = self.read1(amt=amt, decode_content=decode_content)
+                else:
+                    data = self.read(amt=amt, decode_content=decode_content)
+                if not data:
+                    break
```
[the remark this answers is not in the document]
smason: > When the loop is broken here after `read1` was used, `self._fp` remains not closed even though all data has been read. 

Hadn't realised the semantics around what gets closed and when, have done a bit more reading now.  Have had a bit of a refactor of the change.

> Closing it may be a better way to break the loop.  What do you think about checking `self.length_remaining` after calling `read1` and, if it is 0 and `not self.closed`, calling `self._fp.close()`?

It feels as though it could raise an `IncompleteRead` if it finishes and `length_remaining` indicates there's missing data (as `_raw_read` does), any preferences on this / other changes

src/urllib3/response.py
smason: Not sure what I was thinking there, sorry, reverted!

smason:
rebased due to all the bumping in dependencies and test changes

src/urllib3/response.py
```diff
@@ -1020,10 +1020,18 @@ def stream(
...
+                else:
+                    data = self.read(amt=amt, decode_content=decode_content)
                 if data:
                     yield data
+                else:
+                    if self._fp:
+                        self._fp.close()
+                    break
+        if self.length_remaining:
+            raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
```
illia-v, line 1034: This line misses test coverage

src/urllib3/response.py
```diff
@@ -1020,10 +1020,18 @@ def stream(
...
-
+                if amt is None:
+                    data = self.read1(amt=amt, decode_content=decode_content)
+                else:
+                    data = self.read(amt=amt, decode_content=decode_content)
                 if data:
                     yield data
+                else:
+                    if self._fp:
+                        self._fp.close()
```
illia-v, line 1031: Please add a comment explaining in which situation `self._fp` is not closed by `http.client`

src/urllib3/response.py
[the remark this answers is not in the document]
illia-v: > It feels as though it could raise an `IncompleteRead` if it finishes and `length_remaining` indicates there's missing data (as `_raw_read` does), any preferences on this / other changes

Good point!
Looking at [CPython's code](https://github.com/python/cpython/blob/2d76be251d0aee89f76e6fa5a63fa1ad3f2b76cf/Lib/http/client.py#L657-L658), it's possible to get in the situation at least if `self._fp` is closed between `read1` calls. However, you have to respect `self.enforce_content_length` as  `_raw_read` does.

src/urllib3/response.py
smason: I don't have a good feeling for when it isn't closed, but have commented on the case you pointed out

src/urllib3/response.py
```diff
@@ -1020,10 +1020,23 @@ def stream(
...
-                data = self.read(amt=amt, decode_content=decode_content)
-
+                if amt is None:
+                    data = self.read1(amt=amt, decode_content=decode_content)
+                else:
+                    data = self.read(amt=amt, decode_content=decode_content)
                 if data:
                     yield data
+                else:
+                    # http.client.HTTPResponse will often close this for us and
```
sethmlarson, line 1030: This isn't the right place for this to go. This comment makes me very apprehensive about merging this and also makes me worry about the correctness of the `read1` PR.

src/urllib3/response.py
sethmlarson: This comment makes me very apprehensive about merging this and also makes me worry about the correctness of the `read1` PR.

illia-v:
Since we discovered some problems with `http.client.HTTPResponse.read1` while working on this PR, more confidence needs to be gained to make the change to `HTTPResponse.stream`.

In the meanwhile, I created #3235 and #3236 to fix two blocking issues.

nateprewitt:
I wanted to check in on this PR. It looks like the Python landscape and urllib3 have changed a bit since there was last discussion. I'm seeing the previously outstanding concerns as:

**Raising IncompleteRead correctly**: Handled in this PR
**Closing connection when read1 used**: #3235
**Memory efficiency concerns**: #3236

Is there anything else outstanding I'm missing?

illia-v:
@nateprewitt IIRC this was blocked by https://github.com/python/cpython/pull/115998

PR #3217, opened 2023-11-28
Migrate TestHTTPS to Hypercorn

pquentin:
Note: I'm planning to clean up the Tornado tests after finishing the migration, instead of removing the exact bits that are unused, which would take time.

I'm using a fork of Hypercorn here. I'm planning to merge everything back upstream as soon as possible, but doing so requires implementing the full ASGI specification, which is more work.

This migration uncovered two bugs in the test suite that was not considering that asking for "TLS 1.2 or above" could result in "TLS 1.3". Apparently this is something that Tornado was doing wrong?

### DIFF
--- a/test/with_dummyserver/test_https.py
+++ b/test/with_dummyserver/test_https.py
@@ -901,7 +901,7 @@ class TestHTTPS(HTTPSHypercornDummyServerTestCase):
                 conn = https_pool._get_conn()
                 try:
                     conn.connect()
-                    if maximum_version != self.tls_version():
+                    if maximum_version == TLSVersion.MAXIMUM_SUPPORTED:
                         # A higher protocol than tls_protocol_name could be negotiated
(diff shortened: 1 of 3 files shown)


test/with_dummyserver/test_https.py
```diff
@@ -898,7 +901,11 @@ def test_tls_version_maximum_and_minimum(self) -> None:
                 conn = https_pool._get_conn()
                 try:
                     conn.connect()
-                    assert conn.sock.version() == self.tls_protocol_name  # type: ignore[attr-defined]
+                    if maximum_version != self.tls_version():
```
sethmlarson, line 904: Should we make this comparison more explicit with: `if maximum_version == TLSVersion.MAXIMUM_SUPPORTED`?

sethmlarson (approved):
This looks good to me, only one comment.

PR #3218, opened 2023-11-28
Fix two more LONG_TIMEOUT tests

pquentin:
Noticed here: https://github.com/urllib3/urllib3/actions/runs/7017580922/job/19091154584?pr=3217. We don't have any occurrence of `timeout=0.1` in the code base now.

Ultimately, timeouts are more annoying than useful, and I'm considering removing many of them or using much larger values, as what's worse than a timeout is a test hanging.


pquentin:
(Closed by mistake, so I reopened)

sethmlarson (approved):

PR #3222, opened 2023-12-05
Migrate run_server_and_proxy_in_thread to Hypercorn

pquentin:
- [ ] mypy
- [ ] TODO: use ExitStack?

### DIFF
--- a/dummyserver/asgi_proxy.py
+++ b/dummyserver/asgi_proxy.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import httpx
+import trio
+from starlette.requests import Request
+from starlette.responses import Response
+
+
+async def absolute_uri(request, scope, receive, send):
+    async with httpx.AsyncClient() as client:
+        response = await client.request(
+            method=request.method,
+            url=scope["path"],
+            headers=request.headers,
+            data=await request.body(),
+        )
+    headers = {}
+    for header in (
+        "Date",
+        "Cache-Control",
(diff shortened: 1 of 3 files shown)


dummyserver/asgi_proxy.py
```diff
@@ -0,0 +1,72 @@
...
+                break
+            if not data:
+                break
+            await writer.send_all(data)
+        await writer.aclose()
+
+    upstream = await trio.open_tcp_stream(host, int(port))
+    client: trio.SocketStream = scope["extensions"]["_transport"]
+
+    async with trio.open_nursery(strict_exception_groups=True) as nursery:
```
sethmlarson, line 59: I love the conciseness of Trio, well done! :)

sethmlarson (commented):
This is quite exciting! You mentioned in Discord that by the end of this PR we can remove Tornado? :rocket:

dummyserver/asgi_proxy.py
pquentin: Thanks! To be fair it's just like Tornado (https://github.com/urllib3/urllib3/blob/661072c4f805fc448bfe52c04fc6e26ef454cdaa/dummyserver/proxy.py#L86-L112) who has `tornado.gen.multi` but I guess that does not give the same guarantees.

sethmlarson (approved):
This is awesome! I'm happy we're using our own tiny ASGI implementation :shipit:

PR #3224, opened 2023-12-06
PoolManager: Fix incorrect retry configuration in case of redirection

RemiLacroix-IDRIS:
Hi,

When handling a redirection, PoolManager will not use the wanted retry configuration but use the default retry configuration instead.

Best,
Rémi


RemiLacroix-IDRIS:
@pquentin : Sorry to bug you but could you help review this?

sethmlarson (changes requested):
@RemiLacroix-IDRIS Thanks for this. Can you add a [changelog entry](https://github.com/urllib3/urllib3/blob/main/changelog/README.rst) and test case?

PR #3225, opened 2023-12-07
Turn proxy_app into a class to allow configuring it

pquentin:
This pull request makes no difference, but prepares the switch of `HTTPDummyProxyTestCase` to `Hypercorn`.


sethmlarson (commented):
LGTM, but found a print:

PR #3230, opened 2023-12-11
Migrate emscripten test server to Hypercorn

pquentin:
While this should be relatively straightforward, various tests don't work, including some that just hang, and I have no idea how to debug this. I made some progress by looking at the server logs and noticed a few 404 errors, but everything looks good from the server point of view now. And the client errors are cryptic like "No module named js" (and that's after I changed two raise statements to stop throwing away exceptions).

@joemarshall Would you mind taking a look? This is the last test using Tornado, which I am in the process in removing, in order to have a test server that supports HTTP/2.

### DIFF
--- a/dummyserver/app.py
+++ b/dummyserver/app.py
@@ -5,7 +5,6 @@ import contextlib
 import datetime
 import email.utils
 import gzip
-import json
 import mimetypes
 import zlib
 from io import BytesIO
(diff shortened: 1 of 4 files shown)


joemarshall:
Here it is:

https://github.com/joemarshall/urllib3/tree/hypercorn-emscripten

joemarshall:
(you just need to pull that last commit across to your branch)

pquentin:
> (you just need to pull that last commit across to your branch)

Thank you! I'd appreciate learning how to fish, but thanks a lot for the fish! Various streaming tests don't pass at all on my Fedora 39 laptop, and I have no idea why.

sethmlarson (approved):
Looks great! Merge when ready.

PR #3233, opened 2023-12-14
Add support for having a maximum retry after wait length.

alexjircan:
Introducing a new feature that allows you to set a maximum duration for retry wait length using the `max_retry_wait_length` argument. If a response contains a `Retry-After` header with value greater than the maximum set and `respect_retry_after_header` is set to `True`, we will raise a `MaxRetryWaitExceededError`.

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -82,6 +82,13 @@ ConnectionError = ProtocolError
 # Leaf Exceptions
 
 
+class MaxRetryWaitExceededError(HTTPError):
+    def __init__(self, retry_after_header_value: float) -> None:
+        self.retry_after_header_value = retry_after_header_value
+
+        super().__init__()
+
+
 class MaxRetryError(RequestError):
     """Raised when the maximum number of retries is exceeded.
 
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -13,6 +13,7 @@ from ..exceptions import (
     ConnectTimeoutError,
     InvalidHeader,
     MaxRetryError,
+    MaxRetryWaitExceededError,
     ProtocolError,
     ProxyError,
     ReadTimeoutError,
@@ -215,6 +216,7 @@ class Retry:
             str
         ] = DEFAULT_REMOVE_HEADERS_ON_REDIRECT,
         backoff_jitter: float = 0.0,
+        max_retry_wait_length: float | None = None,
     ) -> None:
         self.total = total
         self.connect = connect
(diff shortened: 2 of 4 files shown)


src/urllib3/util/retry.py
```diff
@@ -335,6 +338,8 @@ def get_retry_after(self, response: BaseHTTPResponse) -> float | None:
     def sleep_for_retry(self, response: BaseHTTPResponse) -> bool:
         retry_after = self.get_retry_after(response)
         if retry_after:
+            if self.max_retry_wait_length and retry_after > self.max_retry_wait_length:
+                raise MaxRetryWaitExceededError(retry_after_header_value=retry_after)
```
D-stefaang, line 342: The `is_retry` method probably should be updated too to respect the `max_retry_wait_length`.
e.g. the [responses library relies on it.](https://github.com/getsentry/responses/blob/489b84768af3be97a2f6df6fabcc0dd8d749e2a9/responses/__init__.py#L1154)

src/urllib3/exceptions.py
```diff
@@ -80,6 +80,13 @@ class ProtocolError(HTTPError):
 # Leaf Exceptions
 
 
+class MaxRetryWaitExceededError(HTTPError):
+    def __init__(self, retry_after_header_value: float) -> None:
+        self.retry_after_header_value = retry_after_header_value
+
+        super().__init__()
```
D-stefaang, line 87:
```suggestion
        message = f"Retry-After header value of {retry_after_header_value} exceeds the maximum allowed wait time"
        super().__init__(message)
```

D-stefaang: Without a message, this error returns an empty string for the `__str__` method.

src/urllib3/util/retry.py
alexjircan: Why? A response can still be a retry response even if its `Retry-After` header exceeds the limit, right? I don't see why we should raise an error just for checking this.

alexjircan: A better method name for what you want might be something like `is_retriable()`.

rtpg:
```
        """Is this method/status code retryable? (Based on allowlists and control
        variables such as the number of total retries to allow, whether to
        respect the Retry-After header, whether this header is present, and
        whether the returned status code is on the list of status codes to
        be retried upon on the presence of the aforementioned header)
        """
```
the docstring of `is_retry`. It's not about whether a retry is a retry (after all, `is_retry` is a method on `Retry`! it's a `Retry` object!). In essence, `Retry.is_retry` _is_ `Retry.is_retriable`

PR #3234, opened 2023-12-15
Factorize hypercon.Config() creation

pquentin:
&nbsp;


sethmlarson (dismissed):
LGTM, thanks Quentin!

sethmlarson:
@pquentin Looks like potential failures on the certificate tests, might be related?

pquentin:
Definitely related, thanks.

dummyserver/hypercornserver.py
```diff
@@ -62,7 +76,7 @@ def run_hypercorn_in_thread(
         if not ready_event.is_set():
             raise Exception("most likely failed to start server")
 
-        yield
+        yield typing.cast(int, parse_url(config.bind[0]).port)
```
graingert, line 79:
```suggestion
        port = parse_url(config.bind[0]).port
        assert port is not None
        yield port
```

dummyserver/hypercornserver.py
```diff
@@ -42,8 +44,20 @@ async def _start_server(
 
 @contextlib.contextmanager
 def run_hypercorn_in_thread(
-    config: hypercorn.Config, app: hypercorn.typing.ASGIFramework
-) -> Generator[None, None, None]:
+    host: str, certs: dict[str, typing.Any] | None, app: hypercorn.typing.ASGIFramework
+) -> typing.Iterator[int]:
```
graingert, line 48: Iterator is technically the wrong type, because contextmanager incorrectly accepts an object without `.throw`

```suggestion
) -> Generator[int, None, None]
```

dummyserver/hypercornserver.py
pquentin: When can this happen?

dummyserver/hypercornserver.py
```diff
@@ -42,8 +44,20 @@ async def _start_server(
...
+    config = hypercorn.Config()
+    if certs:
+        config.certfile = certs["certfile"]
+        config.keyfile = certs["keyfile"]
+        if "cert_reqs" in certs:
+            config.verify_mode = certs["cert_reqs"]
+        if "cacerts" in certs:
```
pquentin, line 55:
```suggestion
        if "ca_certs" in certs:
```

dummyserver/hypercornserver.py
pquentin: OK I understand now that this is to avoid the cast.

dummyserver/hypercornserver.py
pquentin: Since this is accepted by mypy, could be fixed in typeshed, is only in our tests and is only "technically wrong", I'd rather keep it as it's easier to understand that way.

dummyserver/hypercornserver.py
pquentin: Done in [`54f28c9` (#3234)](https://github.com/urllib3/urllib3/pull/3234/commits/54f28c9b38d10c656a35d9093dc6d15fee9a95f4)

sethmlarson (approved):
LGTM! :rocket:

PR #3235, opened 2023-12-15
Make `HTTPResponse.read1` close response when all data is read

illia-v:
In #3216, we discovered that CPython's `http.client.HTTPResponse.read1` never closes the response after all data has been read and content length was known.

This PR adds a fix for the issue to `_raw_read` near a similar old fix that closes a response under a different condition.
I'll create a PR to fix this in CPython too most likely.


src/urllib3/response.py
```diff
@@ -869,6 +869,11 @@ def _raw_read(
                     # raised during streaming, so all calls with incorrect
                     # Content-Length are caught.
                     raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
+            elif read1 and (not data or self.length_remaining == len(data)):
+                # All data is read, but CPython's `self._fp.read1` doesn't
+                # always close `http.client.HTTPResponse`, so we close it
+                # here.
+                self._fp.close()
```
pquentin, line 876: Can you please add a link to the CPython issue and mention that this branch can be removed when we drop support for Python 3.12?

test/test_response.py
```diff
@@ -560,6 +560,41 @@ def test_io_closed_consistently(self, sock: socket.socket) -> None:
...
+                    assert not resp._fp.isclosed()
+                    assert not is_fp_closed(resp._fp)
+                    assert not resp.isclosed()
+                    resp.read1(read_amt)
+                assert resp.closed
+                assert resp._fp.isclosed()
+                assert is_fp_closed(resp._fp)
+                assert resp.isclosed()
+        finally:
+            hlr.close()
```
pquentin, line 596: Note that you can also use a context manager here:

```python3
with contextlib.closing(httplib.HTTPResponse(sock)) as hlr:
   hlr.fp = ...
```

But nore sure if clearer or not.

pquentin (approved):
Looks great, thanks! Congrats in getting your CPython fix in.

test/test_response.py
illia-v: Thanks for pointing out. Here I just used the style from the method above, but `HTTPResponse` itself can be used as a context manager as anything implementing `io.IOBase`

src/urllib3/response.py
illia-v: Great idea, added

src/urllib3/response.py
```diff
@@ -869,6 +869,12 @@ def _raw_read(
                     # raised during streaming, so all calls with incorrect
                     # Content-Length are caught.
                     raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
+            elif read1 and (not data or self.length_remaining == len(data)):
```
sethmlarson, line 872: `read1` can return zero bytes without the stream being finished, the semantics of `read1` are different than `read`: `read1(n)` returns up to N bytes, `read(n)` returns either exactly N bytes or if the stream is complete less than N bytes.

For this reason we can't have the `not data` case, otherwise we're at risk of closing our HTTP response early.

sethmlarson (changes requested):

src/urllib3/response.py
illia-v: Docs [state](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1) "If size is -1 (the default), an arbitrary number of bytes are returned (**more than zero unless EOF is reached**)." It looks like [CPython's source](https://github.com/python/cpython/blob/6a1d5a4c0f209d51ab33d6529935d643bcdb3ba2/Lib/http/client.py#L664-L665) expects the same "more than zero" for sizes greater than 0 too. Am I missing something?

Your comment made me think about the case when user calls `.read1(0)`, we shouldn't close the IO in such a case.

sethmlarson: Hmm, maybe I misunderstood the contract of `read1`, so it's identical to `read` only without the chaining, cool. Thanks Illia!

test/test_response.py
```diff
@@ -560,6 +560,38 @@ def test_io_closed_consistently(self, sock: socket.socket) -> None:
         finally:
             hlr.close()
 
+    @pytest.mark.parametrize("read_amt", (None, 3))
```
sethmlarson, line 563: Can we add a case too for when `read_amt` is less than the total length (thus requiring more than one `read1` call?) and have assertions on the in-between state and final state.

sethmlarson (commented):
Thanks for clarifying the behavior of `read1()` to me! I only had one review comment then this LGTM:

test/test_response.py
illia-v: I added two more cases: one suggested by you and one reading 0 bytes

illia-v:
@sethmlarson @pquentin can we merge this?

test/test_response.py
```diff
@@ -576,6 +576,99 @@ def test_io_closed_consistently(self, sock: socket.socket) -> None:
...
+
+    @pytest.mark.parametrize("length_known", (True, False))
+    def test_io_not_closed_after_requesting_0_bytes(
+        self, sock: socket.socket, length_known: bool
+    ) -> None:
+        with httplib.HTTPResponse(sock) as hlr:
+            hlr.fp = BytesIO(b"foo")  # type: ignore[assignment]
+            hlr.chunked = 0  # type: ignore[assignment]
+            length_remaining = 3
+            hlr.length = length_remaining if length_known else None
```
sethmlarson, line 655: This `length` versus `length_remaining` difference is annoying, and obviously has an effect on our public API? Can we create an issue to deprecate one of these properties?

test/test_response.py
```diff
@@ -576,6 +576,99 @@ def test_io_closed_consistently(self, sock: socket.socket) -> None:
...
+                    assert not is_fp_closed(resp._fp)
+                    assert not resp.isclosed()
+                    resp.read1(read_amt)
+                assert resp.closed
+                assert resp._fp.isclosed()
+                assert is_fp_closed(resp._fp)
+                assert resp.isclosed()
+
+    @pytest.mark.parametrize("length_known", (True, False))
+    def test_io_not_closed_untill_all_data_is_read(
```
sethmlarson, line 612: nit: typo "untill"

sethmlarson (approved):
Seems fine to me, I'm sad we're having to get so specific for `http.client` but since we're working around a bug in that implementation it makes sense. Thanks Illia!

test/test_response.py
illia-v: Good catch, thanks!

test/test_response.py
sethmlarson: Gotcha, thanks for clarifying! Too many `HTTPResponse` names flying around :zany_face:

sethmlarson (approved):

PR #3236, opened 2023-12-15
Make `BytesQueueBuffer.get_all` more memory efficient

illia-v:
This PR adds memray tests for functionality added in #3186 and improves memory efficiency of `BytesQueueBuffer.get_all`.

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -291,8 +291,9 @@ class BytesQueueBuffer:
         if len(buffer) == 1:
             result = buffer.pop()
         else:
-            result = b"".join(buffer)
-            buffer.clear()
+            ret = io.BytesIO()
+            ret.writelines(buffer.popleft() for _ in range(len(buffer)))
+            result = ret.getvalue()
         self._size = 0
         return result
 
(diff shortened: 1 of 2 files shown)


src/urllib3/response.py
```diff
@@ -291,8 +291,9 @@ def get_all(self) -> bytes:
         if len(buffer) == 1:
             result = buffer.pop()
         else:
-            result = b"".join(buffer)
-            buffer.clear()
+            ret = io.BytesIO()
+            ret.writelines(buffer.popleft() for _ in range(len(buffer)))
+            result = ret.getvalue()
```
pquentin, line 296: Were you able to measure the memory gain? I would expect `str.join` to do exactly that under the hood. If it made a difference, the surely we could lower some of the `pytest.mark.limit_memory` decorators?

pquentin (commented):
The tests are great, thanks! I would be interested in seeing proof that the `get_all` change makes a difference.

src/urllib3/response.py
illia-v: With `b"".join(buffer)` there was a moment when `result` took memory for all the data, but `ret` has not been cleared yet and it took the same amount of memory additionally. So `test_memory_usage` reported around 20 MB instead of around 12 MB (5 * 2 MiB chunks + 2 MiB for overhead from the last chunk.)

pquentin (approved):
Thanks! LGTM.

illia-v:
@pquentin thanks for your review!

PR #3237, opened 2023-12-18
CI: Only run cron jobs for upstream

hugovk:
These two crons have been running on my fork:

image

I don't think we need them for forks.

This config doesn't stop the workflows starting on schedule, but it immediately skips them to save lots of CI time.

More info:

https://dev.to/hugovk/til-how-to-disable-cron-for-github-forks-2d0l


pquentin (approved):
Thanks! LGTM.

PR #3244, opened 2023-12-19
Fix coverage

pquentin:
We recently had coverage issues, where `coverage combine` complained about missing coverage in a way that made no sense, see https://github.com/urllib3/urllib3/actions/runs/7214376906/job/19669120651:

```
Name                                           Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------
...
src/urllib3/_base_connection.py                   38     26    32%   68-143
```

However, this lines up with the release of [coverage.py 7.3.3](https://github.com/nedbat/coveragepy/releases/tag/7.3.3). Since we are not pinning the version of coverage in the action but pin when running tests, this can lead to issues. Let's see if that solves this particular one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
           python-version: "3.x"
 
       - name: "Install coverage"
-        run: "python -m pip install --upgrade coverage"
+        run: "python -m pip install -r dev-requirements.txt"
 
       - name: "Download artifact"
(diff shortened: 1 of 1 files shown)


pquentin:
Merging without a review as it's a one line fix, easy to revert and will unblock several pull requests. Feel free to review after the fact to suggest a different approach.

illia-v:
@pquentin I have no objection to merging such PRs without reviews. One thing I noticed though is doing so can affect OpenSSF [score](https://deps.dev/pypi/urllib3) 😕

illia-v:
Thanks for fixing the coverage issue!

Issue #2770, opened 2022-11-08
TestBrokenPipe.test_ignore_broken_pipe_errors fails on macOS 12 (Monterey)


pquentin:
We're using `macos-latest` in CI, and about two weeks ago GitHub Actions upgraded from macOS 11.7 (Big Sur) to macOS 12.6 (Monterey). Since then, `test_ignore_broken_pipe_errors` fails reliably like this:

```
Traceback (most recent call last):
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/connectionpool.py", line 732, in urlopen
    httplib_response = self._make_request(
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/connectionpool.py", line 444, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/connection.py", line 355, in request
    self.send(chunk)
  File "/Users/runner/hostedtoolcache/Python/3.8.14/x64/lib/python3.8/http/client.py", line 972, in send
    self.sock.sendall(data)
ConnectionResetError: [Errno 54] Connection reset by peer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/runner/work/urllib3/urllib3/test/with_dummyserver/test_socketlevel.py", line 2023, in test_ignore_broken_pipe_errors
    r = pool.request("POST", "/", body=buf)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/_request_methods.py", line 117, in request
    return self.request_encode_body(
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/_request_methods.py", line 216, in request_encode_body
    return self.urlopen(method, url, **extra_kw)
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/connectionpool.py", line 799, in urlopen
    retries = retries.increment(
  File "/Users/runner/work/urllib3/urllib3/.nox/test-3-8/lib/python3.8/site-packages/urllib3/util/retry.py", line 467, in increment
    raise reraise(type(error), error, _stacktrace)
[...]

PR #3246, opened 2023-12-26
Fix TestBrokenPipe on macOS

zawan-ila:
Fixes [2770](https://github.com/urllib3/urllib3/issues/2770)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         os:
-          - macos-11
+          - macos-latest
           - windows-latest
           - ubuntu-20.04  # OpenSSL 1.1.1
           - ubuntu-22.04  # OpenSSL 3.0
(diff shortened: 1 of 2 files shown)


sethmlarson:
@zawan-ila This doesn't have an explanation for why this PR is correct, I don't think we should be silencing exceptions we don't understand. Do you know why this issue is occurring?

zawan-ila:
@sethmlarson The reason that this is occuring is essentially the same as the one that we "catch" `EPROTOTYPE` here for. Remembrances from the past [here](https://github.com/urllib3/urllib3/pull/1524/commits/567a8f4b265b216fc78f92d2af9dc3caf5f4b878) and [here](https://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/)

What's happened since then is that Apple changed some of their code and some of the code paths that resulted in the `EPROTOTYPE` mentioned above now result in the setting of `ECONNRESET`. In particular, note the difference in the if block between the [old version](https://github.com/apple-oss-distributions/xnu/blob/rel/xnu-7195/bsd/netinet/tcp_usrreq.c#L1020C1-L1024) and the [new one](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/netinet/tcp_usrreq.c#L1164-L1168). In fact, I don't think we need the special "catch" for `EPROTOTYPE` anymore.

sethmlarson (approved):
Thanks for the explanation! LGTM

pquentin:
Since we closed https://github.com/urllib3/urllib3/issues/2770, can we bump CI to macos-12?

sethmlarson:
@pquentin I believe we can! :)

PR #3253, opened 2024-01-01
Remove Tornado

pquentin:
This also removes all direct imports of asyncio as we're currently using Trio for our test server.


pquentin:
As far as I can tell the only failure is https://github.com/urllib3/urllib3/pull/3254.

sethmlarson (approved):
Leaving Tornado in 2023, love it! :wave:

PR #3255, opened 2024-01-01
Use coverage sys.monitoring for a speed-boost

sethmlarson:
New in coverage 7.4.0! See: https://nedbatchelder.com/blog/202312/coveragepy_with_sysmonitoring.html


pquentin:
Nice! According to https://mastodon.social/@hugovk/111657679446754746 this option only works on Python 3.12 for now.

sethmlarson:
@pquentin Yeah I forgot to upgrade our coverage dependency, resolving that issue now locally.

noxfile.py
```diff
@@ -36,6 +36,20 @@ def tests_impl(
...
 
+    # Environment variables being passed to the pytest run.
+    pytest_session_envvars = {
+        "PYTHONWARNINGS": "always::DeprecationWarning",
+    }
+
+    # In coverage 7.4.0 we can only set the setting for Python 3.12+
+    # Future versions of coverage will use sys.monitoring based on availability.
+    if (
+        isinstance(session.python, str)
```
pquentin, line 47: In what case is this not a string?

pquentin (approved):
Thanks! LGTM if CI passes.

noxfile.py
sethmlarson: This is mostly to make type hints happy, apparently the type of `session.python` is `True | None | str`.

sethmlarson:
All of the 3.12 jobs passed, merging!

Issue #3250, opened 2023-12-29
void


abebeos:

sethmlarson:
I don't think we should be configuring this at the level of an HTTP response, the response is an IO object so its behavior should be dictated by the consuming user by calling either `.read()` or `.read1()`.

abebeos:
> so its behavior should be dictated by the consuming user by calling either `.read()` or `.read1()`.

That can't be done when preloading content, it is hardcoded to `.read()` 

https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L620

(alternative would be to introduce a `preload_via_read1=False` parameter, so the scope is limited to only preloading)

sethmlarson:
Does calling `read` versus `read1` matter for preloaded content? We have to block until everything is downloaded anyways, there's no "opportunistic" reads/streaming we can do in that case.

abebeos:
I do not know. Me processing this quite abstractly.

`read1` parameter was introduced for `_raw_read()` and `_fp_read()`, but this is not enough to allow for flexibility in testing.

(maybe await some feedback from the author or read1, @smason)

abebeos:
Ok, update: it looks like it is not necessary to theorize about this.

Work on https://github.com/urllib3/urllib3/issues/3249 will demonstrate the necessity clearly.

(I've fixed the test-failures 'quick&dirty', needs more time for production quality)

sethmlarson:
Brought this up in Discord too, copying here. I think `read1`'s API contract is incompatible with `decode_content` since we can't uphold the IO stream contract of always returning some bytes if `amt > 0` and the stream isn't EOL. For `read()` we can uphold the contract because we're allowed to issue multiple underlying `read()` calls, `read1()` requires us to only issue one syscall though so we can't guarantee that we'll receive enough data in order to emit decompressed data.

abebeos:
In this issue here I pick up the existent code, add test-cases (https://github.com/urllib3/urllib3/pull/3248/commits/3496f0f86a50a64ada0d6b5b87c537955e4a11a4, that fail initially), fix test-cases etc. - I am assuming that the basic read1 design is ok.

But: if read1 is already flawed-by-design (as your comment implies), then this should be processed in a new issue, like **"Validate read1()'s usage of 'decode_content' via test-cases"** - something like that.

(personally I think that it's fine to have `decode_content` within read1() - only `cache_content` is obsolete (as its a single read)).

[...]

PR #3258, opened 2024-01-02
add 'incomplete-read/read1' tests

abebeos:
- re #3250

PR https://github.com/urllib3/urllib3/pull/3186 introduced support for the read1() method in class HTTPResponse.

The added tests showcase that the current implementation of `read1()` is inconsistent.

Depending on data, `DecodeError` is either raised or not. It should be raised always (and all tests should pass, as with `read()`)

```sh
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[f] PASSED                                 [ 25%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[fo] FAILED                                [ 25%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[foo] FAILED                               [ 26%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[foofoofoofoofoo] PASSED                   [ 26%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[123456789012345] FAILED                   [ 26%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[foo4] FAILED                              [ 26%]
test/test_response.py::TestResponse::test_decode_zstd_incomplete_read1[xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx] PASSED
```

(please focus within this PR only on the tests themselves. If they are correct, then they should be merged, possibly with an 'ignore/skip' marker for the failing test)

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -448,6 +448,45 @@ class TestResponse:
         with pytest.raises(DecodeError):
             HTTPResponse(fp, headers={"content-encoding": "zstd"})
 
+    decode_param_set = [
+        b"f",
+        b"fo",
+        b"foo",
+        b"foo" * 5,
+        b"123456789012345",
+        b"foo4",
+        b"x" * 100,
+    ]
+
+    @onlyZstd()
+    @pytest.mark.parametrize("data", decode_param_set)
+    def test_decode_zstd_incomplete_read(self, data: bytes) -> None:
+        data = zstd.compress(data)
+        fp = BytesIO(data[:-1])  # shorten the data to trigger DecodeError
+
+        # create response object without(!) reading/decoding the content
+        r = HTTPResponse(
+            fp, headers={"content-encoding": "zstd"}, preload_content=False
+        )
+
+        # read/decode, expecting DecodeError
+        with pytest.raises(DecodeError):
+            r.read(decode_content=True)
+
+    @onlyZstd()
+    @pytest.mark.parametrize("data", decode_param_set)
+    def test_decode_zstd_incomplete_read1(self, data: bytes) -> None:
+        data = zstd.compress(data)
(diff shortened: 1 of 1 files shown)


test/test_response.py
```diff
@@ -448,6 +448,45 @@ def test_decode_zstd_incomplete(self, data: bytes) -> None:
...
+        data = zstd.compress(data)
+        fp = BytesIO(data[:-1])
+
+        r = HTTPResponse(
+            fp, headers={"content-encoding": "zstd"}, preload_content=False
+        )
+
+        # read/decode via read1(!), expecting DecodeError
+        with pytest.raises(DecodeError):
+            r.read1(decode_content=True)
```
illia-v, line 488: Thanks for checking the new `read1` functionality!

In this case, I am not convinced there is a bug, the tests look incorrect. You can get this test to stop failing if you add one more `r.read1(decode_content=True)` call. Similarly, you can make `test_decode_zstd_incomplete_read` fail if you call `r.read(amt=len(data), decode_content=True)`.

`read1` is not guaranteed to read all the data. If it reads a piece of the compressed data and successfully decodes the data, it returns it without flushing the decoder. This is just like `read` when passed `amt` is less than the total content length.

In this case, `r.read1(decode_content=True)` returns `data[:-1]`.

abebeos: I think i've understood.

abebeos:
@illia-v, could you please reopen so i can try to adjust the tests?

abebeos:
follow-up: https://github.com/urllib3/urllib3/pull/3265

PR #3260, opened 2024-01-02
Fix timeout tests on Windows

pquentin:
`TestConnectionPoolTimeouts::test_timeout` sometimes fails with `assert 0.4999573230743408 >= 0.5`. The default clock on Windows has an average
granularity of 15.625ms which maybe means that our code sometimes
thinks that we waited less than we actually did? Switching to
perf_counter() should fix this issue.

https://randomascii.wordpress.com/2020/10/04/windows-timer-resolution-the-great-rule-change/ contains more details.

Noticed in https://github.com/urllib3/urllib3/actions/runs/7384069987/job/20086320256?pr=3256 (and earlier).

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -98,10 +98,10 @@ class TestConnectionPoolTimeouts(SocketDummyServerTestCase):
             self.host, self.port, timeout=short_timeout, retries=False
         ) as pool:
             wait_for_socket(ready_event)
-            now = time.time()
+            now = time.perf_counter()
             with pytest.raises(ReadTimeoutError):
                 pool.request("GET", "/", timeout=LONG_TIMEOUT)
-            delta = time.time() - now
+            delta = time.perf_counter() - now
 
             message = "timeout was pool-level SHORT_TIMEOUT rather than request-level LONG_TIMEOUT"
             assert delta >= LONG_TIMEOUT, message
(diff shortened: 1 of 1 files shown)


ecerulm:
I believe this was taken care of in https://github.com/urllib3/urllib3/pull/3276

sethmlarson:
Thanks @ecerulm!

pquentin:
Using a more precise clock would have been cleaner, but the important part is to fix the issue. Thanks!

ecerulm:
I don't have anything against a higher resolution clock, but I think `time.perf_counter` would not have really help in this case. 

` 0.5 - 0.4999978542327881` is `2.1457672119140625e-06`, that is 0.002 milliseconds or 2 microseconds. 

I believe this test case failure is not about the clock resolution.  , but float point arithmetic error , we are substracting two float numbers to get the elapsed time and it's very counter intuitive but in python (and any IEEE 754 for that matter) `0.3 - 0.1 != 0.2` . See [Python 3 documentation > 15. Floating Point Arithmetic: Issues and Limitations
][floatingpointissues]

```
python -c 'print(0.3-0.1 == 0.2)'
False
```

Anytime you are using floating point arithmetic (`time.perf_counter()` still returns a `float`)  you need to account for tolerances. 

[floatingpointissues]: https://docs.python.org/3/tutorial/floatingpoint.html

pquentin:
Right, but the clock resolution is around 15 milliseconds on Windows when using `time.time()`. This is why this test was only failing on Windows: with a more precise clock, the floating point errors don't matter.

Issue #2799, opened 2022-11-11
Do we flush the decoder when reaching EOF in partial reads?


pquentin:
I started writing a very detailed bug report and then convinced me that it was not real. Sorry, this will be the short version now. Looks at our flush_decoder logic:

https://github.com/urllib3/urllib3/blob/938c0a96bddc469f10358ccb35e86deee48ede3c/src/urllib3/response.py#L756-L794

We should only flush when there is no data left. In the streaming case, it would be because we reached EOF, and data is `b''`. Which is done correctly line 776. But then we don't call `self._decode` because `if data` will prevent us from doing it. So we don't flush().

This bug was introduced in urllib3 1.11, released in July 2015: https://github.com/urllib3/urllib3/commit/29e144ff88a32a263121fb6eda59a767702e63d4. Surely in 7 years someone would have noticed if they had missing data.

So surely the flush() call is not actually doing anything, whatever the reason is.

abebeos:
Hi there!

I've started this yesterday, mostly by looking at https://github.com/urllib3/urllib3/commit/29e144ff88a32a263121fb6eda59a767702e63d4 to understand what happened back then.

I should have a documented answer within one or two days, ongoing work documented within:

- https://github.com/abebeos/urllib3/issues/1

sethmlarson:
@abebeos  Thanks for digging deep into the issue, can't wait to hear your conclusions! Looks like you're on a good path so far.

abebeos:
@sethmlarson , this should be the result - need to clean it up and post it here as a comment, but will be nearly the same :

- https://github.com/abebeos/urllib3/issues/1

abebeos:
## Answer

Yes.

Whenever "no further decoding" is expected, the decoder is to be flushed.

`decoder.flush()` is essentially `finalize decoding`.

### Documentation
* [x] https://docs.python.org/3/library/zlib.html?highlight=flush#zlib.Decompress.flush
  * >All pending input is processed, and a bytes object containing the remaining uncompressed output is returned. After calling [flush()](https://docs.python.org/3/library/zlib.html?highlight=flush#zlib.Decompress.flush), the [decompress()](https://docs.python.org/3/library/zlib.html?highlight=flush#zlib.decompress) method cannot be called again; the only realistic action is to delete the object.
    * => `flush` is essentially a `finalize_decoding`

## Root Cause

[...]

PR #3262, opened 2024-01-02
[DEMO] 2799 flush decoder clarify

abebeos:
re ##2799

This should clarify several issues around `flush_decoder` and related stuff.

~(summary diff is confusing, better go commit by commit)~

~summary diff is now readable, individual diffs allow for better validation of refactoring steps.~

squashed

- restored commits

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -907,15 +907,6 @@ class HTTPResponse(BaseHTTPResponse):
         if decode_content is None:
             decode_content = self.decode_content
 
-        if amt is not None:
-            cache_content = False
-
-            if len(self._decoded_buffer) >= amt:
-                return self._decoded_buffer.get(amt)
-
-            if amt == 0:
-                return b""
-
         if amt is None:
             data = self._raw_read()
             if not data and len(self._decoded_buffer) == 0:
@@ -927,7 +918,15 @@ class HTTPResponse(BaseHTTPResponse):
             if cache_content:
                 self._body = data
 
-        else:
+        else:  # amt is not None
+            cache_content = False
+
+            if len(self._decoded_buffer) >= amt:
+                return self._decoded_buffer.get(amt)
+
+            if amt == 0:
+                return b""
+
             data = self._raw_read(amt)
 
             # do not waste memory on buffer when not decoding
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -907,18 +907,7 @@ class HTTPResponse(BaseHTTPResponse):
         if decode_content is None:
             decode_content = self.decode_content
 
-        if amt is None:
-            data = self._raw_read()
-            if not data and len(self._decoded_buffer) == 0:
-                return data
-
-            # here, 'data' contains all content, thus setting 'flush_decoder=True'
-            # to finalize decoding, as this will be the only/last _decode call
-            data = self._decode(data, decode_content, flush_decoder=True)
-            if cache_content:
-                self._body = data
-
-        else:  # amt is not None
+        if amt is not None:
             cache_content = False
 
             if len(self._decoded_buffer) >= amt:
@@ -946,6 +935,17 @@ class HTTPResponse(BaseHTTPResponse):
                     break
                 data = self._raw_read(amt)
             data = self._decoded_buffer.get(amt)
+
+        if amt is None:
+            data = self._raw_read()
+            if not data and len(self._decoded_buffer) == 0:
+                return data
+
+            # here, 'data' contains all content, thus setting 'flush_decoder=True'
+            # to finalize decoding, as this will be the only/last _decode call
+            data = self._decode(data, decode_content, flush_decoder=True)
+            if cache_content:
+                self._body = data
 
         return data
 
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -930,10 +930,8 @@ class HTTPResponse(BaseHTTPResponse):
                     )
                 return data
 
-            # restoring old behaviour to see if tests fail
-            flush_decoder = not data
             while True:
-                # flush_decoder = not data
+                flush_decoder = not data
                 decoded_data = self._decode(data, decode_content, flush_decoder)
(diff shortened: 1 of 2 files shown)


sethmlarson:
Looks like there's a code coverage issue in CI: https://github.com/urllib3/urllib3/actions/runs/7469435382/job/20327201070?pr=3262

abebeos:
@sethmlarson , re

> [move amt==0 up to become reachable code](https://github.com/urllib3/urllib3/pull/3262/commits/b059dba2bae20af521c4144f2b699c733081525c)

```py
            if amt == 0:
                return b""

            if len(self._decoded_buffer) >= amt:
                return self._decoded_buffer.get(amt)
```

the `amt=0` check is redundant, if placed after(!) the `if len(self._decoded_buffer) >= amt:`

but for the sake of code-clarity, I've kept it and moved it up.

abebeos:
This looks like it does not need a changelog

src/urllib3/response.py
```diff
@@ -910,21 +910,17 @@ def read(
...
 
+            if amt == 0:
+                return b""
+
             if len(self._decoded_buffer) >= amt:
                 return self._decoded_buffer.get(amt)
 
-        data = self._raw_read(amt)
-
-        flush_decoder = amt is None or (amt != 0 and not data)
```
illia-v, line 918: What do you think about keeping the current code structure for now, but defining `flush_decoder` directly in the `self._decode` calls? I.e., calling `self._decode(data, decode_content, flush_decoder=True)` when `amt is None`  and `self._decode(data, decode_content, flush_decoder=not data)` otherwise.

As per https://github.com/urllib3/urllib3/issues/2799#issuecomment-1901102673, this can be a user-facing change, so it has to be accompanied by new tests and a news entry.

abebeos: There should(!) be no functional change, but naturally I may have overseen something 

I'll try to recreate the squashed commits from a backup (thought this https://github.com/urllib3/urllib3/pull/3262#issuecomment-1884214847 was final, so i squashed), and answer you then.

src/urllib3/response.py
```diff
@@ -938,23 +938,13 @@ def read(
...
-            while len(self._decoded_buffer) < amt and data:
-                # TODO make sure to initially read enough data to get past the headers
-                # For example, the GZ file header takes 10 bytes, we don't want to read
-                # it one byte at a time
-                data = self._raw_read(amt)
-
-                # TODO verify if flush_decoder is redundant here, as it is not updated
-                #      anyways within the loop. Passing 'False' directly does not lead
-                #      to test-failures. Should be 'True' only(!) for last time called
+            while True:
```
abebeos, line 933: This introduces a functional change, which fixes the logical bug of the original code (flush_decoder was not updated within loop).

src/urllib3/response.py
abebeos: here we go: https://github.com/urllib3/urllib3/pull/3262#discussion_r1460238172

abebeos: > What do you think about keeping the current code structure for now
@illia-v , it is fixed with step https://github.com/urllib3/urllib3/pull/3262#discussion_r1460238172, all other steps should be functionally neutral or minor.

I'm fine with however you proceed, i can try to send a separate PR in, but if you prefer, you can author a PR (and if you worry about crediting, just add something like "based on digging by abebeos within https://github.com/urllib3/urllib3/issues/2799#issuecomment-1872448595" to the commit comment).

You (lead maintainers) must(!) however realize that the cause of all this #2799 (and other issues, incl. security related) is(!) the current code-structure.

src/urllib3/response.py
```diff
@@ -944,6 +948,10 @@ def read(
                 # For example, the GZ file header takes 10 bytes, we don't want to read
                 # it one byte at a time
                 data = self._raw_read(amt)
+
+                # TODO verify if flush_decoder is redundant here, as it is not updated
+                #      anyways within the loop. Passing 'False' directly does not lead
+                #      to test-failures. Should be 'True' only(!) for last time called
```
abebeos: I identified a potential bug correctly here (fixed later on within https://github.com/urllib3/urllib3/pull/3262/files#r1460238172), but failed to update my report (https://github.com/urllib3/urllib3/issues/2799#issuecomment-1872448595, section `2023-12-27: Todays Code`)

src/urllib3/response.py
```diff
@@ -912,6 +912,9 @@ def read(
 
             if len(self._decoded_buffer) >= amt:
                 return self._decoded_buffer.get(amt)
+            
+            if amt == 0:
+                return b""
```
abebeos, line 915: This introduces a functional change, needs to be reversed.

src/urllib3/response.py
```diff
@@ -934,6 +938,9 @@ def read(
                     )
                 return data
 
+            flush_decoder = not data
```
abebeos: potential functional change in combination with https://github.com/urllib3/urllib3/pull/3262/files/c49227eb983300c512098809105c1050fea1f25d#r1460291175

(possibly should become `(amt != 0 and not data)` )

abebeos:
Some time has passed, so I was able to self-review and detect one mistake I made, see https://github.com/urllib3/urllib3/pull/3262#discussion_r1460291175.

As things stand, this PR has served its purpose, and can be closed unmerged (I would anyways proceed with a new PR, if the decision is taken to revise the code for clarity purposes).

(I do some experiments and then close it)

src/urllib3/response.py
abebeos: > What do you think about keeping the current code structure for now

yes, see https://github.com/urllib3/urllib3/pull/3262#issuecomment-1901960676

abebeos:
- follow-up: https://github.com/urllib3/urllib3/pull/3292

src/urllib3/response.py
abebeos: applied the minimum steps

- => https://github.com/urllib3/urllib3/pull/3292

PR #3264, opened 2024-01-03
remove "watcher" thread suggestion from docs

### LINKED ISSUE #1741: How to implement a "watcher" thread as suggested by the documentation?
The documentation of the [`timeout`](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.timeout) module says:

> If your goal is to cut off any request after a set amount of wall clock time, consider having a second “watcher” thread to cut off a slow request.

How would that work?

It seems like it is [strongly discouraged or even impossible](https://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread) to kill a thread in Python, so what would that watcher thread do?

If it is not possible to write a watcher thread in Python, the documentation shouldn't suggest to do it.

abebeos:
closes #1741 (see https://github.com/urllib3/urllib3/issues/1741#issuecomment-551669274)

### DIFF
--- a/src/urllib3/util/timeout.py
+++ b/src/urllib3/util/timeout.py
@@ -101,10 +101,6 @@ class Timeout:
         the case; if a server streams one byte every fifteen seconds, a timeout
         of 20 seconds will not trigger, even though the request will take
         several minutes to complete.
-
-        If your goal is to cut off any request after a set amount of wall clock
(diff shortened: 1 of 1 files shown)


abebeos:
@pquentin , a trivial one

pquentin (approved):
Thanks! LGTM.

PR #3265, opened 2024-01-04
add read/read1 incomplete tests (V2)

abebeos:
~re https://github.com/urllib3/urllib3/issues/3250~

Add additional tests for both, HTTPResponse.read() and HTTPResponse.read1()

(first attempt within #3258 was wrong, this PR provides corrected tests, which pass)

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -431,8 +431,13 @@ class TestResponse:
                 break
         assert ret == b"foobarbaz"
 
+    decode_param_set = [
+        b"foo",
+        b"x" * 100,
+    ]
+
     @onlyZstd()
-    @pytest.mark.parametrize("data", [b"foo", b"x" * 100])
+    @pytest.mark.parametrize("data", decode_param_set)
     def test_decode_zstd_error(self, data: bytes) -> None:
         fp = BytesIO(data)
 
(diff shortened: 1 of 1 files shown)


test/test_response.py
```diff
@@ -448,6 +448,49 @@ def test_decode_zstd_incomplete(self, data: bytes) -> None:
...
+            fp, headers={"content-encoding": "zstd"}, preload_content=False
+        )
+
+        # read/decode via read1(!), expecting DecodeError
+        with pytest.raises(DecodeError):
+            amt_decoded = 0
+            # loop, as read1() may return just partial data
+            while amt_decoded < len(data):
+                part = r.read1(decode_content=True)
+                amt_decoded += len(part)
```
illia-v, line 492: What do you think about making the behavior check more explicit by checking that the first `read1` returns the truncated data and the second one leads to `DecodeError`? Reading all data from a not changing `BytesIO` object always requires just one `read1()` unless the amount is specified.

test/test_response.py
```diff
@@ -448,6 +448,49 @@ def test_decode_zstd_incomplete(self, data: bytes) -> None:
...
 
+    decode_param_set = [
+        b"f",
+        b"fo",
+        b"foo",
+        b"foo" * 5,
+        b"123456789012345",
+        b"foo4",
+        b"x" * 100,
+    ]
```
illia-v, line 459: It doesn't look like the bigger number of cases is more useful, let's use the ones from `test_decode_zstd_incomplete`

illia-v (commented):
Thanks for adding more tests

test/test_response.py
abebeos: me bit dizzy from the other issues to understand. If the test-case is valid, it should go in, can be extended/refined later.

(note that this is in context of https://github.com/urllib3/urllib3/issues/3250, so while I get a better understanding, the teest-cases will be refined anyways)

test/test_response.py
abebeos:
```
        b"foo" * 5,
        b"123456789012345",
```

behaved differently (even if it was wrong code).

And I believe that testing for a single char should be included (b"f")

So, possibly go with this (at least for now)

```
    decode_param_set = [
        b"f",
        b"foo",
        b"foo" * 5,
        b"123456789012345",
        b"x" * 100,
    ]
```

Note that if I am to work on the read1 related issue, I would want to have a broader param_set anyways. Can be always reduced down again.

I can of course change this PR to the original set of 2, and use an extended set locally. No problem, so, if you still prefere, I can restore the set here.

abebeos: => reduced to original set

test/test_response.py
abebeos: => after a bit thought: keep test focused on DecodeError

abebeos:
@illia-v , should be ok now. Naturally, please feel free to extend/revise pre/post merge.

illia-v (approved):

PR #3270, opened 2024-01-08
Add explicit_package_bases to pyproject.toml

ecerulm:
Currently the mypy step of `nox -rs lint` can fail if for example you have a  `__init__.py` file in you `~`/`$HOME` directory

```
nox -rs lint
nox > Running session lint
nox > Re-using existing virtual environment at .nox/lint.
nox > python -m pip install pre-commit
nox > pre-commit run --all-files
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
prettier.................................................................Passed
eslint...................................................................Passed
nox > python -m pip install -r mypy-requirements.txt
nox > mypy --version
mypy 1.5.1 (compiled: yes)
nox > mypy dummyserver noxfile.py src/urllib3 test
/Users/rubelagu/__init__.py: error: Source file found twice under different module names: "rubelagu.git.urllib3.src.urllib3.exceptions" and "urllib3.exceptions"
/Users/rubelagu/__init__.py: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules for more info
/Users/rubelagu/__init__.py: note: Common resolutions include: a) adding `__init__.py` somewhere, b) using `--explicit-package-bases` or adjusting MYPYPATH
Found 1 error in 1 file (errors prevented further checking)
nox > Command mypy dummyserver noxfile.py src/urllib3 test failed with exit code 2
nox > Session lint failed.
``` 

by adding  the [explicit_package_bases = true](https://mypy.readthedocs.io/en/stable/config_file.html#confval-explicit_package_bases) as option to mypy then that kind of problem is prevented, making it more reliable.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ add_imports = "from __future__ import annotations"
 
 [tool.mypy]
 mypy_path = "src"
+explicit_package_bases = true
 check_untyped_defs = true
 disallow_any_generics = true
 disallow_incomplete_defs = true


sethmlarson:
@ecerulm Would this also be solved by listing `dummyserver noxfile.py src/urllib3 test` for `mypy_path` instead of only `src`?

pquentin:
One concern I have is that this option is not documented to fix issues like this.

ecerulm:
@sethmlarson , no , the problem is that mypy will look in all parent directories beside the paths in `mypy_path`. That's what you avoid with `explicit_package_bases = true` , then mypy will only search in the paths that you explicitly gave. 

For example if you by mistake create a file `$HOME/__init__.py`  mypy will find that although is not in the `mypy_path` and assume that the whole $HOME is a source directory. 

Pasted_Image_2024-01-09__08_11

Using `mypy_path = "dummyserver,noxfile.py,src/urllib3,test"`  does not help. I tried with it. And it will still fail if you `touch ~/__init__.py`. 

In all fairness, this probably does not affect many people, I only realized because `nox -rs lint` was failing for me and then I realized that I had by chance a `~/__init__.py` in my home directory that I copied there by mistake many months ago. But I think `explicit_package_bases=true` is probably more correct. 

But the behavior is described https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules

> mypy will allow for the possibility that directories without __init__.py[i] are packages. Specifically, **mypy will look at all parent directories of the file and use the location of the highest __init__.py[i] in the directory tree** to determine the top-level package.

In any case I think `explicit_package_bases=true` is probably desirable, since I don't think you want mypy ever to scan out of the source directory. 

@pquentin , this option **is documented to fix issues like this**, the error message mentions `--explicit-package-bases` directly:

>> note: Common resolutions include: a) adding `__init__.py` somewhere, **b) using `--explicit-package-bases`** or adjusting MYPYPATH


Anyway you can try in your own environment

```
nox -rs lint
touch ~/__init__.py
nox -rs lint
```

pquentin:
Thanks for the detailed explanation. Regarding documentation, I was referring to https://mypy.readthedocs.io/en/stable/config_file.html#confval-explicit_package_bases which states:

> This option is only useful in the absence of `__init__.py`.

Which is what makes me -0.5 on this.

ecerulm:
If you don't like `--explicit-package-bases` then the other solution is to use `mypy -p dummyserver -m noxfile -p urlib3 -p test` instead of `mypy dummyserver noxfile.py src/urllib3 test`. I tested that one , and it works even when there is a `~/__init__.py` file present:

```
nox -rs lint                                                   
nox > Running session lint
nox > Re-using existing virtual environment at .nox/lint.
nox > python -m pip install pre-commit
nox > pre-commit run --all-files
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
prettier.................................................................Passed
eslint...................................................................Passed
nox > python -m pip install -r mypy-requirements.txt
nox > mypy --version
mypy 1.5.1 (compiled: yes)
nox > mypy -p dummyserver -p urllib3 -p test -m noxfile
Success: no issues found in 79 source files
nox > Session lint was successful.
```

If you are ok with using `-p` and `-m` in mypy, then I can create a PR. 

Or alternatively, close this ticket.  To be honest is probably enough to have this PR searchable on the internet. If anybody happens to make the same mistake as me and create a `$HOME/__init__.py` file it will probably find this PR and realize how to fix it. 

I leave it up to you.

ecerulm:
If you don't like `--explicit-package-bases` then the other solution is to use `mypy -p dummyserver -m noxfile -p urlib3 -p test` instead of `mypy dummyserver noxfile.py src/urllib3 test`. I tested that one , and it works even when there is a `~/__init__.py` file present:

```
nox -rs lint                                                   
nox > Running session lint
nox > Re-using existing virtual environment at .nox/lint.
nox > python -m pip install pre-commit
nox > pre-commit run --all-files
pyupgrade................................................................Passed
black....................................................................Passed
isort....................................................................Passed
flake8...................................................................Passed
prettier.................................................................Passed
eslint...................................................................Passed
nox > python -m pip install -r mypy-requirements.txt
nox > mypy --version
mypy 1.5.1 (compiled: yes)
nox > mypy -p dummyserver -p urllib3 -p test -m noxfile
Success: no issues found in 79 source files
nox > Session lint was successful.
```

If you are ok with using `-p` and `-m` in mypy, then I can create a PR. 

Or alternatively, close this ticket.  To be honest is probably enough to have this PR searchable on the internet. If anybody happens to make the same mistake as me and create a `$HOME/__init__.py` file it will probably find this PR and realize how to fix it. 

I leave it up to you.

pquentin:
I do like this better, thanks!

PR #3278, opened 2024-01-14
ensure ProxyManager honors default_socket_options

### LINKED ISSUE #2936: ProxyManager doesn't honor HTTPConnection.default_socket_options
ProxyManager doesn't honor HTTPConnection.default_socket_options. I'm attempting to enable TCP Keep Alive which I monitor via "netstat -tnope" on Linux. Only the third block below works when using a Proxy. 

**Without a proxy, default_socket_options enables TCP Keep Alive successfully**
`from urllib3.connection import HTTPConnection
import urllib3
]
http = urllib3.PoolManager()
http.request('GET', <url>)
`

**With a proxy, default_socket_options does not enable TCP Keep Alive (or any other socket option)**
`import urllib3
]
http = urllib3.ProxyManager(<proxy url>)
http.request('GET', <url>)
`

**With a proxy, ProxyManager constructor param "socket_options" does enable TCP Keep Alive**
`from urllib3.connection import HTTPConnection
import urllib3
http = urllib3.ProxyManager(<proxy url>, socket_options=HTTPConnection.default_socket_options + [
    (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1)
])
http.request('GET', <url>)
`

This was observed while using the "requests" module to connect to a URL via a proxy that has a short idle timeout, and the common "HTTPConnection.default_socket_options" workaround did nothing. I couldn't find a way to pass "socket_options" to ProxyManager through "requests" methods.

abebeos:
closes #2936

### Spawned Issue

- https://github.com/urllib3/urllib3/issues/3280

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -99,11 +99,14 @@ class HTTPConnection(_HTTPConnection, object):
       Or you may want to disable the defaults by passing an empty list (e.g., ``[]``).
     """
 
+    _SOCKET_OPTIONS_NAGLE_ENABLE = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 0)]
+    _SOCKET_OPTIONS_NAGLE_DISABLE = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
+
     default_port = port_by_scheme["http"]
 
     #: Disable Nagle's algorithm by default.
     #: ``[(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]``
-    default_socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
+    default_socket_options = _SOCKET_OPTIONS_NAGLE_DISABLE
 
     #: Whether this connection verifies the host's certificate.
     is_verified = False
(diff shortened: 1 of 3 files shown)


abebeos:
Strange, it works within my IDE (after some struggle).

is the usage of `HTTPConnection.default_socket_options` thread-safe?

src/urllib3/connection.py
```diff
@@ -101,9 +101,7 @@ class HTTPConnection(_HTTPConnection, object):
 
     default_port = port_by_scheme["http"]
 
-    #: Disable Nagle's algorithm by default.
-    #: ``[(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]``
-    default_socket_options = [(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)]
+    default_socket_options = []
```
sigmavirus24, line 104: This is backwards incompatible. There is certainly code today that clones this list and includes it when passing socket options so that they get the good and reasonable defaults from the library and whatever extra they wish to use.

src/urllib3/connectionpool.py
```diff
@@ -220,11 +220,6 @@ def __init__(
         self.conn_kw = conn_kw
 
         if self.proxy:
-            # Enable Nagle's algorithm for proxies, to avoid packet fragmentation.
```
sigmavirus24, line 223: This comment would indicate that the intent here was to do something different than what the code did. I suspect something got lost before we had sufficient code coverage

sigmavirus24 (changes requested):

src/urllib3/connectionpool.py
abebeos: Possibly author just forgot to make the correct/final implementation.

src/urllib3/connection.py
abebeos: > This is backwards incompatible.

Clearly!

src/urllib3/connectionpool.py
abebeos: This is the bug location, the author used the 'user provided defaults" to delete the options, instead of implementing another mechanism. I just realized that I bugged now the `HTTPConnectionPool` (dizzy last night), I'll go on to see to provide a final fix that keeps `HTTPConnectionPool` intact.

src/urllib3/connection.py
abebeos: => restored

src/urllib3/connectionpool.py
abebeos: => not actionable

sethmlarson:
@abebeos It looks like this PR is targeting the 1.26.x branch, is that intentional?

abebeos:
> this PR is targeting the 1.26.x branch, is that intentional?

Yes, see https://github.com/urllib3/urllib3/issues/2936#issuecomment-1469910082

Issue #3072, opened 2023-06-17
HttpHeaderDict does not support bytes object


tushar5526:
### Subject

Passing bytes as the value in `HttpHeaderDict` is not handled completely. 

### Environment

Describe your environment.
At least, paste here the output of:

```python
import platform
import ssl
import urllib3

print("OS", platform.platform())
print("Python", platform.python_version())
print(ssl.OPENSSL_VERSION)
print("urllib3", urllib3.__version__)
```
```
OS Linux-5.19.0-43-generic-x86_64-with-glibc2.35
Python 3.10.6
OpenSSL 3.0.2 15 Mar 2022
```

### Steps to Reproduce

```
import urllib3

headers = urllib3.HTTPHeaderDict()
encoded_data = "Schönefeld/1.18.0".encode('latin-1')

headers['user-agent'] = encoded_data
headers['user-agent']
print(headers['user-agent'])
```

### Expected Behavior

You can access bytes objects passed in headers. 

### Actual Behavior
Errors out with this
```
Traceback (most recent call last):
  File "/home/tushar55/test.py", line 8, in <module>
    headers['user-agent']
  File "/home/tushar55/projects/urllib3/src/urllib3/_collections.py", line 260, in __getitem__
    return ", ".join(val[1:])
TypeError: sequence item 0: expected str instance, bytes found

```

Bytes are not handled - https://github.com/urllib3/urllib3/blob/main/src/urllib3/_collections.py#L257 , https://github.com/urllib3/urllib3/blob/main/src/urllib3/_collections.py#L426 

Using bytes as keys has some level of handling - https://github.com/urllib3/urllib3/blob/main/src/urllib3/_collections.py#L251

Found this issue while going through #3047

panzer:
@tushar5526 Are you looking for someone to contribute a fix? If so, I can look into it.

tushar5526:
Hi @ panzer, this issue is still not confirmed. The core maintainers have not triaged this one yet. Let's wait for them :)

panzer:
@tushar5526 I'm looking through some old issues and pull requests. It seems the maintainers do not do much triage. I'm willing to look work on this issue. If we want to wait for a maintainer to review, let's get in touch directly with that person

sethmlarson:
@tushar5526 @panzer Thanks for the patience, it would be great to have support for bytes as values in `HTTPHeaderDict`. I'm not sure how exactly that would work internally, maybe you have some ideas :)

[...]

PR #3279, opened 2024-01-14
allow byte values in HTTPHeaderDict

zawan-ila:
Closes [3072](https://github.com/urllib3/urllib3/issues/3072)

## Overview
I've followed the same path that supports byte keys. Open to any other suggestions.


abebeos:
Looks straightforward (but me unexperienced in codebase)

- Tiny lint problem: https://github.com/urllib3/urllib3/actions/runs/7522156998/job/20473813933?pr=3279#step:4:136
- and add changelog 3072.bugfix.rst, like e.g. In https://github.com/urllib3/urllib3/blob/main/changelog/3268.bugfix.rst

pquentin (commented):
While I realize we have a bounty for this issue, I'm not sure that blindly assuming `latin-1` in 2024 is the right thing to do. In RFC 9110 (started in 2018 and published in 2022 and which applies to all HTTP versions), [Section 5.5. Field Values](https://www.rfc-editor.org/rfc/rfc9110#section-5.5) states (emphasis mine):

> Field values are usually constrained to the range of US-ASCII characters [[USASCII](https://www.rfc-editor.org/rfc/rfc9110#USASCII)]. Fields needing a greater range of characters can use an encoding, such as the one defined in [[RFC8187](https://www.rfc-editor.org/rfc/rfc9110#RFC8187)]. Historically, HTTP allowed field content with text in the ISO-8859-1 charset [[ISO-8859-1](https://www.rfc-editor.org/rfc/rfc9110#ISO-8859-1)], supporting other charsets only through use of [[RFC2047](https://www.rfc-editor.org/rfc/rfc9110#RFC2047)] encoding. **Specifications for newly defined fields SHOULD limit their values to visible US-ASCII octets (VCHAR), SP, and HTAB.** A recipient SHOULD treat other allowed octets in field content (i.e., [obs-text](https://www.rfc-editor.org/rfc/rfc9110#fields.values)) as opaque data.

What we could do, if the user insists in sending bytes, is to join values using `b", "`, which I guess could work? Or we could refuse bytes altogether. (The current exception is not helpful and I'd like to fix that somehow.)

Also, would you mind checking what encoding is used when "Schönefeld/1.18.0" is passed as a string?

zawan-ila:
> Also, would you mind checking what encoding is used when "Schönefeld/1.18.0" is passed as a string?

Do you mean `passed as a header value`? If yes, it is encoded as `latin-1`. Apparently courtesy of [http.client](https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L1302-L1304).

> What we could do, if the user insists in sending bytes, is to join values using b", ", which I guess could work? Or we could refuse bytes altogether.

What do you suggest here? It'd be better to be able to send bytes. However, I'm curious if urllib3 can read arbitrary header values? It'd be awkward to be able to send particular header values but not receive them. Looks like probably `http.client` would restrict us here in some way?

pquentin:
Yes, `http.client` limits us for HTTP/1.1, but we try to fix bugs when it's possible. We can't fix decoding, but can fix encoding. And we can do the right thing for HTTP/2, using the `header_encoding` h2 parameter, whatever the right thing is!

As pointed out in https://github.com/urllib3/urllib3/pull/2397#issuecomment-907832691, it makes sense to be able to send headers as bytes, because the user may not want to send latin-1. As mentioned in the HTTP spec referenced above, `A recipient SHOULD treat other allowed octets in field content (i.e., [obs-text](https://www.rfc-editor.org/rfc/rfc9110#fields.values)) as opaque data.` This means that if a user wants to use UTF-8, it's perfectly OK, and the receiving server should treat this as opaque data.

So my current thinking is that we should indeed join byte values using `b", "` and forbid mixed encoding and bytes (to be able to join). And not try to decode bytes to latin-1. @sethmlarson would know what's best here.

@sethmlarson By the way, did your opinion change since 2021? Should we explicitly allow bytes in the API as was done in this pull request?

zawan-ila:
@pquentin I have updated the PR. What do you think now?

PS: There's some lint issues. Plus we need to update some type annotations. Do let me know if they are blockers for a review.

pquentin:
@zawan-ila Yes, type annotation failures usually block review, especially for a typing change and with so many errors:

```
nox > python -m pip install -r mypy-requirements.txt
nox > mypy --version
mypy 1.8.0 (compiled: yes)
nox > mypy -p dummyserver -m noxfile -p urllib3 -p test
src/urllib3/_collections.py:202: error: Argument 2 to "_has_value_for_header" of "HTTPHeaderDict" has incompatible type "str | bytes"; expected "str"  [arg-type]
src/urllib3/_collections.py:261: error: Argument 1 to "join" of "bytes" has incompatible type "list[str]"; expected "Iterable[Buffer]"  [arg-type]
src/urllib3/_collections.py:339: error: Unsupported operand types for + ("str" and "bytes")  [operator]
src/urllib3/_collections.py:454: error: Argument 1 to "join" of "bytes" has incompatible type "list[str]"; expected "Iterable[Buffer]"  [arg-type]
test/test_collections.py:221: error: Incompatible types in assignment (expression has type "bytes", target has type "str")  [assignment]
test/test_collections.py:222: error: Non-overlapping equality check (left operand type: "str", right operand type: "Literal[b'foo, bar']")  [comparison-overlap]
test/test_collections.py:295: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:296: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:337: error: Incompatible types in assignment (expression has type "bytes", target has type "str")  [assignment]
test/test_collections.py:338: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:339: error: Argument "rawcookie" to "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:340: error: Argument "rawcookie" to "NonMappingHeaderContainer" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:348: error: Incompatible types in assignment (expression has type "bytes", target has type "str")  [assignment]
test/test_collections.py:349: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:350: error: Argument "rawcookie" to "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:351: error: Argument "rawcookie" to "NonMappingHeaderContainer" has incompatible type "bytes"; expected "str"  [arg-type]
test/test_collections.py:399: error: Incompatible types in assignment (expression has type "bytes", target has type "str")  [assignment]
test/test_collections.py:400: error: Non-overlapping container check (element type: "tuple[str, bytes]", container item type: "tuple[str, str]")  [comparison-overlap]
test/test_collections.py:401: error: Non-overlapping container check (element type: "tuple[str, bytes]", container item type: "tuple[str, str]")  [comparison-overlap]
test/test_collections.py:480: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
test/with_dummyserver/test_socketlevel.py:1957: error: Argument 2 to "add" of "HTTPHeaderDict" has incompatible type "bytes"; expected "str"  [arg-type]
Found 21 errors in 3 files (checked 80 source files)
nox > Command mypy -p dummyserver -m noxfile -p urllib3 -p test failed with exit code 1
nox > Session lint failed.
```

That said, #3343 made this pull request much more important so I'll take a look anyway! Sorry for not answering before.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1926,6 +1926,41 @@ def socket_handler(listener: socket.socket) -> None:
...
+
+            while expected not in buffer:
+                with contextlib.suppress(BlockingIOError):
+                    buffer += sock.recv(65536)
+
+            request_headers = buffer.split(b"\n", 1)[1]
+            sock.sendall(
+                b"HTTP/1.1 200 OK\r\n"
+                b"Server: example.com\r\n"
+                b"Content-Length: 0\r\n" + request_headers
```
pquentin, line 1950: Please send request headers in the body to avoid the decoding that is done by httplib on responses.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1926,6 +1926,41 @@ def socket_handler(listener: socket.socket) -> None:
...
+
+        self._start_server(socket_handler)
+
+        headers = HTTPHeaderDict()
+        headers.add("rawcookie", b"\xc2\x80")
+        with HTTPConnectionPool(self.host, self.port, retries=False) as pool:
+            r = pool.request("GET", "/", headers=headers)
+            assert r.status == 200
+            # Note that b"\xc2\x80" is Â\x80 when decoded as latin-1
+            assert r.headers["rawcookie"] == "Â\x80"
```
pquentin, line 1962: With the above change, you should ultimately be able to write `assert r.data == b"rawcookie: \xc2\x80"`

pquentin (commented):
Thanks for this work!

Can you please fix the mypy errors before I can review in more detail? I suspect this will involve explicitly typing `self._container` as `Mapping[str, str | bytes]`.

You will also need to add a [changelog entry](https://github.com/urllib3/urllib3/tree/main/changelog).

And finally, we need to test with `-bb` to check for string/bytes comparisons, but the use of the h2 library makes that more difficult.

zawan-ila:
@pquentin I've updated the types. However I am not too happy with my work in this regard. There are too many ignores. The main difficulties appear to me to be the following:-

1. While request header values can be bytes, response header values are not.
2. The values in self._container are lists such that the first value is the header string and the rest are the values. Hence one is forced to type them as list[str | bytes] 
3. This is my first time using mypy.

It would be really nice if you could have a look and provide some helpful suggestions.

src/urllib3/_collections.py
```diff
@@ -155,7 +155,7 @@ def keys(self) -> set[_KT]:  # type: ignore[override]
             return set(self._container.keys())
 
 
-class HTTPHeaderDictItemView(typing.Set[typing.Tuple[str, str]]):
+class HTTPHeaderDictItemView(typing.Set[typing.Tuple[str, typing.Union[str, bytes]]]):
```
pquentin, line 158: What error do you get when using `str | bytes`? Union types are supported by Python 3.10+ (https://docs.python.org/3/library/stdtypes.html#types-union), and I can see other places (line 240) where they appear to work just fine.

test/with_dummyserver/test_poolmanager.py
```diff
@@ -629,7 +629,7 @@ def test_request_with_json(self, headers: HTTPHeaderDict) -> None:
         content_type = HTTPHeaderDict(old_headers).get(
             "Content-Type", "application/json"
         )
-        assert content_type in r.headers["Content-Type"].replace(" ", "").split(",")
+        assert content_type in r.headers["Content-Type"].replace(" ", "").split(",")  # type: ignore[arg-type]
```
pquentin, line 632: What's the mypy error here?

src/urllib3/response.py
```diff
@@ -339,7 +341,7 @@ def __init__(
         self.chunked = False
         tr_enc = self.headers.get("transfer-encoding", "").lower()
         # Don't incur the penalty of creating a list and then discarding it
-        encodings = (enc.strip() for enc in tr_enc.split(","))
+        encodings = (enc.strip() for enc in tr_enc.split(","))  # type: ignore[arg-type]
```
pquentin, line 344: What's the mypy error here?

pquentin (commented):
> While request header values can be bytes, response header values are not.

Saying that all header values can be bytes is not a problem in itself, but if it requires adding type ignores everywhere, then we could reconsider. I think this depends on what we want to do with HTTP/2 headers, which we haven't defined yet.

HTTP/2 headers passed to the h2 library have to be bytes, but we can ask [configure h2 to decode our headers](http://python-hyper.org/projects/h2/en/stable/api.html#h2.config.H2Configuration). If we do that, then we could use [generics](https://mypy.readthedocs.io/en/stable/generics.html) to have one HTTPHeaderDict for requests (support str | bytes) and one for responses (only supporting str), which would help.

@sethmlarson @illia-v What do you think?

> The values in self._container are lists such that the first value is the header string and the rest are the values. Hence one is forced to type them as list[str | bytes]

Yeah, that's a hack to keep the original case of the key. But at least it's self-contained in HTTPHeaderDictView. We could solve this by introducing a second dict, but I would need to see microbenchmarks and a memory profile to see the performance impact.

> This is my first time using mypy.

That's OK! We're here to help, and you're doing great anyway. I'm happy that you're also trying to find a clean solution. Also, sorry for the delay between reviews.

> [PS] I also now understand why type annotations ought to be a blocker for review :-)

:-)

PR #3283, opened 2024-01-17
ensure HTTPSconnection.is_verified is set properly

### LINKED ISSUE #3267: https_proxy to http_host leads to `is_verfied==True`
(followup to #3130)

### Subject

@illia-v , the test you provided within https://github.com/urllib3/urllib3/pull/3149#pullrequestreview-1683158046

```py
     def test_https_proxy(self) -> None:                                                                                                                                                                                                                                                                                                                                                                            
         with proxy_from_url(self.https_proxy_url, ca_certs=DEFAULT_CA) as https:                                                                                                                                                                                                                                                                                                                                   
[...]                                                                                                                                                                                                                                                                                                                                                                                                     
             r = https.request("GET", f"{self.http_url}/")                                                                                                                                                                                                                                                                                                                                                          
             assert r.status == 200                                                                                                                                                                                                                                                                                                                                                                                 
+            pool = list(https.pools._container.values())[-1]                                                                                                                                                                                                                                                                                                                                                       
+            connection = pool.pool.queue[-1]                                                                                                                                                                                                                                                                                                                                                                       
+            assert connection.is_verified is False                                                                                                                                                                                                                                                                                                                                                                 
+            assert connection.proxy_is_verified is True                                                                                                                                                                                                                                                                                                                                                          
                                                             
```

Thus, the 

* `proxy_is_verifed is True`, `is_verified is False`

I tend to agree with this logic, but the current code does

* `proxy_is_verifed is True`, `is_verified is True`.

This seems like an additional (besides #3130) bug (or at least flaw).

You can review the behavior within 

- https://github.com/urllib3/urllib3/pull/3266

the relevant test is `test_is_verified_https_proxy_to_http_target`

abebeos:
- closes #3267

### Followup Issue

Relevant code needs to be revised & (self)documented (to avoid such kind of bugs in future)

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -670,7 +670,10 @@ class HTTPSConnection(HTTPConnection):
         )
         self.sock = sock_and_verified.socket
 
-        # https://github.com/urllib3/urllib3/issues/3267.
+        # Forwarding proxies can never have a verified target since
+        # the proxy is the one doing the verification. Should instead
+        # use a CONNECT tunnel in order to verify the target.
+        # See: https://github.com/urllib3/urllib3/issues/3267.
         if self.proxy_is_forwarding:
             self.is_verified = False
         else:
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -265,6 +265,13 @@ class HTTPConnection(_HTTPConnection):
     def has_connected_to_proxy(self) -> bool:
         return self._has_connected_to_proxy
 
+    @property
+    def proxy_is_forwarding(self) -> bool:
+        """
+        Return True if a forwarding proxy is configured, else return False
+        """
+        return bool(self.proxy) and self._tunnel_host is None
+
     def close(self) -> None:
         try:
             super().close()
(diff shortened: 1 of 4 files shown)


src/urllib3/connection.py
```diff
@@ -663,11 +670,11 @@ def connect(self) -> None:
...
-        # TODO: Set correct `self.is_verified` in case of HTTPS proxy +
-        #       HTTP destination, see
-        #       `test_is_verified_https_proxy_to_http_target` and
-        #       https://github.com/urllib3/urllib3/issues/3267.
-        self.is_verified = sock_and_verified.is_verified
+        # https://github.com/urllib3/urllib3/issues/3267.
+        if self.proxy_is_forwarding:
+            self.is_verified = False
+        else:
+            self.is_verified = sock_and_verified.is_verified
```
sigmavirus24, line 677: Why not more simply

```
self.is_verified = sock_and_verified.is_verified and not self.proxy_is_forwarding
```

sigmavirus24 (changes requested):

src/urllib3/connection.py
abebeos: Personal preference.

src/urllib3/connection.py
```diff
@@ -265,6 +265,13 @@ def is_connected(self) -> bool:
     def has_connected_to_proxy(self) -> bool:
         return self._has_connected_to_proxy
 
+    @property
+    def proxy_is_forwarding(self) -> bool:
```
sethmlarson, line 269: Can you open an issue for creating a property for `proxy_is_tunnel`?

sethmlarson (approved):
LGTM

PR #3292, opened 2024-01-20
flush decoder when reaching EOF in partial reads

### LINKED ISSUE #2799: Do we flush the decoder when reaching EOF in partial reads?

abebeos:
(closes #2799)

This is the minimun patch extracted from the work within

- https://github.com/urllib3/urllib3/pull/3262

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -915,13 +915,12 @@ class HTTPResponse(BaseHTTPResponse):
 
         data = self._raw_read(amt)
 
-        flush_decoder = amt is None or (amt != 0 and not data)
-
         if not data and len(self._decoded_buffer) == 0:
             return data
 
-        if amt is None:
-            data = self._decode(data, decode_content, flush_decoder)
+        if amt is None:  # reads all content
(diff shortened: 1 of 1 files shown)


abebeos:
@illia-v , as for a test, please see https://github.com/urllib3/urllib3/issues/2799#issuecomment-1902058961

PR #3304, opened 2024-01-22
Add test-pypy3.8 test-pypy3.9 test-pypy3.10

ecerulm:
Currently there are nox session for `test-3.8`, `test-3.9`, `test-3.10`, `test-3.11` and `test-pypy`. 

Then `test-pypy` is should in my opinion be split into `test-pypy3.8`, `test-pypy3.9`, and `test-pypy3.10` since that is already done in CI. It easier to run multiple pypy locally in the laptop if those nox sessions are explicit. 

With this PR is possible to run the nox sessions for all pypy versions like so

```zsh
brew install pyenv 
brew install nox

pyenv install pypy3.8
pyenv install pypy3.9
pyenv install pypy3.10
pyenv install 3.8
pyenv install 3.9
pyenv install 3.10
pyenv install 3.11


#zsh
path=(
$(pyenv prefix 3.8)/bin
$(pyenv prefix 3.9)/bin
$(pyenv prefix 3.10)/bin
$(pyenv prefix 3.11)/bin
$(pyenv prefix pypy3.8)/bin
$(pyenv prefix pypy3.9)/bin
$(pyenv prefix pypy3.10)/bin
$path
)

nox -rs format lint mypy test-3.8 test-pypy3.8 test-3.9 test-pypy3.9 test-3.10 test-pypy3.10 

```

The rational behind this PR is to be able to run the pypy tests locally in my computer for pypy3.8, pypy3.9 and pypy3.10 without having to delete the `.nox/pypy` in between. Also I think for consistency with the cpython tests where there is a separate nox session for each one. 

This PR has  #3286 as prerequisite though.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -13,8 +13,8 @@ If you wish to add a new feature or fix a bug:
    to start making your changes.
 #. Write a test which shows that the bug was fixed or that the feature works
    as expected.
-#. Format your changes with black using command `$ nox -rs format` and lint your
-   changes using command `nox -rs lint`.
+#. Format your changes with black using command ``nox -rs format`` and lint your
+   changes using command ``nox -rs lint``.
 #. Add a `changelog entry
    <https://github.com/urllib3/urllib3/blob/main/changelog/README.rst>`__.
 #. Send a pull request and bug the maintainer until it gets merged and published.
@@ -36,18 +36,21 @@ We use some external dependencies, multiple interpreters and code coverage
 analysis while running test suite. Our ``noxfile.py`` handles much of this for
 you::
 
-  $ nox --reuse-existing-virtualenvs --sessions test-3.8 test-3.9
+  $ nox --reuse-existing-virtualenvs --sessions test-3.12 test-pypy3.10
   [ Nox will create virtualenv if needed, install the specified dependencies, and run the commands in order.]
-  nox > Running session test-3.8
-  .......
-  .......
-  nox > Session test-3.8 was successful.
-  .......
-  .......
-  nox > Running session test-3.9
-  .......
-  .......
-  nox > Session test-3.9 was successful.
+
+
+Note that for nox to test different interpreters, the interpreters must be on the
+``PATH`` first. Check with ``which`` to see if the interpreter is on the ``PATH``
+like so::
+
+
+  $ which python3.12
+  ~/.pyenv/versions/3.12.1/bin/python3.12
+
+  $ which pypy3.10
+  ~/.pyenv/versions/pypy3.10-7.3.13/bin/pypy3.10
+
 
 There is also a nox command for running all of our tests and multiple python
 versions.::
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -8,6 +8,7 @@
 
 import nox
 
+nox.options.error_on_missing_interpreters = True
```
pquentin, line 11: Can you please remove `--error-on-missing-interpreters` in `.github/workflows/ci.yml` then?

pquentin (commented):
Thanks!

pquentin (approved):
Thanks! LGTM.

pquentin:
This would be interesting to understand why we have so many issues with PyPy. Any ideas?

ecerulm:
Well, I actually noticed that `test-pypy` fails very often under pypy3.9 on the CI, and I wanted to investigate more, but then I realized I could not test locally `pypy3.8` and `pypy3.9` hence this PR.

I don't really understand yet, why the test under pypy3.9 are sometimes very very slow (it takes more than 30 minutes to get to 50% of the test suite) , I want to check if that is something that happens only on the GItHub Actions / CI or if it is something that I can replicate with pypy3.9 locally. 

I just wrote an issue to track the the pypy on ci problems: https://github.com/urllib3/urllib3/issues/3305

ecerulm:
> This would be interesting to understand why we have so many issues with PyPy. Any ideas?

@pquentin , I created PR #3308 which I think solves (or sidesteps) the issues with pypy3.9 in particular.

sethmlarson:
@ecerulm Can you resolve the conflicts on this PR with the 3.9 one that I just merged?

docs/contributing.rst
```diff
@@ -36,7 +36,7 @@ We use some external dependencies, multiple interpreters and code coverage
 analysis while running test suite. Our ``noxfile.py`` handles much of this for
 you::
 
-  $ nox --reuse-existing-virtualenvs --sessions test-3.8 test-3.9
+  $ nox --reuse-existing-virtualenvs --sessions test-3.8 test-3.9 test-pypy3.8 test-pypy3.10
```
pquentin, line 39: Believe it or not, we make sure the test below matches the command run. Since you're changing that file:

 * Please remove the output in lines 41-50 so that we don't have to update it anymore.
 * Let's also use more recent versions, such as 3.10 for PyPy and 3.12 for CPython, to avoid having to update the file for the Python 3.8 and 3.9 removal. This applies to your note about `PATH` too.

docs/contributing.rst
```diff
@@ -49,6 +49,21 @@ you::
   .......
   nox > Session test-3.9 was successful.
 
+
+Note that for nox to test different interpreters, the interpreters must be on the
+`PATH` first. Check with `which` to see if the interpreter is on the `PATH`
```
pquentin, line 54: Those should be double backticks

PR #3310, opened 2024-01-24
Allow testing HTTP/1.1 and HTTP/2 in the same test

### LINKED ISSUE #3194: Create a mechanism to test different HTTP protocols (HTTP/1.1 and HTTP/2) with the same test case
Currently our test suite only has to worry about HTTP/1.1, but with HTTP/2 support coming we're going to need to run tests both in HTTP/1.1 and HTTP/2 modes to avoid duplicating the entire test suite for the high-level tests like for `PoolManager`.

My proposal for the mechanism would be to control Hypercorn's ALPN to only offer `http/1.1` or `h2` depending on the test suite and then either run tests twice using a fixture or allowing external control somehow (environment variable?) and then running a separate job on CI? Open to other ideas too!

pquentin:
Closes #3194

We're using a parametrized fixture that when used, will run a given test twice, with HTTP/1.1 and HTTP/2. When used with HTTP/2, it injects HTTP/2 support (which currently forces the usage of HTTP/2). The value of the fixture is "h11" or "h2", which helps to check the header return by Hypercorn.

Things I can do differently:

 * instead of h11 and h2, I could use the ALPN names http/1.1 and h2, or the full names, HTTP/1.1 and HTTP/2. Or even a more named tuple containing all the previous values.
 * instead of using a parametrized fixture, I could use an explicit `pytest.mark.parametrize` [as done by Hip](https://github.com/python-trio/hip/blob/e198149c677edbeec023aeb934758c9195a4d2e3/test/with_dummyserver/conftest.py#L3-L11). It's much more explicit (case in point, I had no idea that `loopback_host` was a parametrized fixture) but also a lot of work - AnyIO needs a lot of pytest code to make that work transparently.
 * instead of `http_version`, I could use a different name to clarify that this tests HTTP/1.1 and HTTP/2.2, but `http_version` is natural to use inside tests.

This pull request was sponsored by [Elastic](https://www.elastic.co/), my employer.


test/with_dummyserver/test_http2.py
illia-v, line 1: Should we rename this file to `test_hypercorn_server.py` now since the test uses both HTTP/1.1 and HTTP/2?

illia-v (approved):
The changes look great!

test/with_dummyserver/test_http2.py
pquentin: Or maybe it's OK to drop that file altogether now?

illia-v: Good point, I'm fine with dropping it since other tests start using HTTP/2

illia-v (approved):

PR #3314, opened 2024-01-25
Update mypy library version to 1.8.0

Nightrider0098:


src/urllib3/connection.py
```diff
@@ -307,7 +307,7 @@ def putrequest(
             method, url, skip_host=skip_host, skip_accept_encoding=skip_accept_encoding
         )
 
-    def putheader(self, header: str, *values: str) -> None:
+    def putheader(self, header: str | bytes, *values: str | bytes) -> None:
```
pquentin, line 310: Our handling of bytes in headers is not ideal (https://github.com/urllib3/urllib3/issues/3072), but I would not want to block upgrading mypy because of it.

What error do you get without this change?

src/urllib3/http2.py
```diff
@@ -72,11 +72,19 @@ def putrequest(
                 )
             )
 
-    def putheader(self, header: str, *values: str) -> None:
+    def putheader(self, header: str | bytes, *values: str | bytes) -> None:
```
pquentin, line 75: Same question, why did you need this?

pquentin (commented):
Thanks for tackling this!

src/urllib3/contrib/pyopenssl.py
```diff
@@ -412,7 +412,12 @@ class PyOpenSSLContext:
     """
 
     def __init__(self, protocol: int) -> None:
-        self.protocol = _openssl_versions[protocol]
+        ssl_method = _openssl_versions.get(OpenSSL.SSL._SSLMethod(protocol))
```
pquentin, line 415: The addition of `OpenSSL.SSL._SSLMethod` is breaking CI.

src/urllib3/connection.py
Nightrider0098: image

src/urllib3/contrib/pyopenssl.py
```diff
@@ -61,23 +63,48 @@ class UnsupportedExtension(Exception):  # type: ignore[no-redef]
...
+    PROTOCOL_TLSv1_1 = (
+        ssl.PROTOCOL_TLSv1_1 if hasattr(ssl, "PROTOCOL_TLSv1_1") else None
+    )
+    PROTOCOL_TLSv1_2 = (
+        ssl.PROTOCOL_TLSv1_2 if hasattr(ssl, "PROTOCOL_TLSv1_2") else None
+    )
+
+
 # Map from urllib3 to PyOpenSSL compatible parameter-values.
 _openssl_versions = {
```
illia-v, line 85: Can this be an alternative to the extensive changes to the file?

```suggestion
_openssl_versions: dict[int, int] = {
```

src/urllib3/connection.py
illia-v: I'd say it's better to add `# type: ignore` instead of adding a type hint for an unsupported type

src/urllib3/http2.py
```diff
@@ -72,11 +72,19 @@ def putrequest(
...
+                header_bytes = header.encode("utf-8").lower()
+            else:
+                header_bytes = header.lower()
+
+            if isinstance(value, str):
+                value_bytes = value.encode("utf-8")
+            else:
+                value_bytes = value
+
+            self._h2_headers.append((header_bytes, value_bytes))
```
illia-v, line 87: Let's move adding support for non-str values out this PR

src/urllib3/contrib/pyopenssl.py
```diff
@@ -412,7 +414,11 @@ class PyOpenSSLContext:
     """
 
     def __init__(self, protocol: int) -> None:
-        self.protocol = _openssl_versions[protocol]
+        ssl_method = _openssl_versions[protocol]
+        if ssl_method is None:
+            # Handle the case when protocol is not found
+            raise ValueError(f"Unsupported SSL/TLS protocol: {protocol}")
+        self.protocol = ssl_method
```
illia-v, line 421: Please drop the change if it isn't required for successful type checking

src/urllib3/contrib/pyopenssl.py
```diff
@@ -61,13 +61,14 @@ class UnsupportedExtension(Exception):  # type: ignore[no-redef]
 from .. import util
 
 if typing.TYPE_CHECKING:
-    from OpenSSL.crypto import X509  # type: ignore[import]
+    from OpenSSL.crypto import X509  # type: ignore[import-untyped]
 
 
 __all__ = ["inject_into_urllib3", "extract_from_urllib3"]
 
+
```
illia-v, line 69:
```suggestion
```

src/urllib3/contrib/pyopenssl.py
```diff
@@ -76,6 +77,7 @@ class UnsupportedExtension(Exception):  # type: ignore[no-redef]
 if hasattr(ssl, "PROTOCOL_TLSv1_1") and hasattr(OpenSSL.SSL, "TLSv1_1_METHOD"):
     _openssl_versions[ssl.PROTOCOL_TLSv1_1] = OpenSSL.SSL.TLSv1_1_METHOD
 
+
```
illia-v, line 80:
```suggestion
```

src/urllib3/http2.py
Nightrider0098: shall we throw error for other types? or add ignore @illia-v

illia-v: In this PR no functional changes (new errors, etc.) are wanted because we don't add support for any new types

Nightrider0098: Cool will remove the

Nightrider0098:
@pquentin Facing issue where i am getting test failure in random test. Like for the last commit it happend for 3.9 this time for macos. What could be the reason?

pquentin:
We have some flakiness issues. If 90% of tests are passing and lint is passing you are probably OK. We will double check when reviewing.

Nightrider0098:
> We have some flakiness issues. If 90% of tests are passing and lint is passing you are probably OK. We will double check when reviewing.

Have made all the required changes. Should we merge it?

src/urllib3/contrib/pyopenssl.py
```diff
@@ -412,7 +412,8 @@ class PyOpenSSLContext:
     """
 
     def __init__(self, protocol: int) -> None:
-        self.protocol = _openssl_versions[protocol]
+        ssl_method = _openssl_versions[protocol]
```
sethmlarson, line 415: Is this change needed now since `_openssl_versions` has been annotated?

src/urllib3/http2.py
```diff
@@ -97,11 +97,15 @@ def putrequest(
                 )
             )
 
-    def putheader(self, header: str, *values: str) -> None:
+    def putheader(self, header: str, *values: str) -> None:  # type: ignore[override]
         for value in values:
-            self._h2_headers.append(
-                (header.encode("utf-8").lower(), value.encode("utf-8"))
-            )
+            if isinstance(header, str):
```
sethmlarson, line 102: I'm surprising this isn't giving a type error, if `header` or `value` aren't strings then `header_bytes` and `value_bytes` aren't defined? Should we revert the change adding the `isinstance` checks?

sethmlarson (changes requested):
Thanks for this @Nightrider0098, here's my review:

src/urllib3/http2.py
Nightrider0098: Remove is instance check and add type override

illia-v (approved):
LGTM, thanks! @Nightrider0098 feel free to claim the $100 bounty from our OpenCollective.

PR #3316, opened 2024-01-25
Fix HTTP version in debug log

pquentin:
Addresses https://github.com/urllib3/urllib3/issues/3297#issuecomment-1908856383. It isn't great, but it will log HTTP/0 for emscripten, HTTP/20 for HTTP/2 and HTTP/11 for HTTP/1.1. Is it time to add an actual `http_version` or `version_string` in `BaseHTTPResponse` alongside `version`?


src/urllib3/connectionpool.py
```diff
@@ -544,17 +544,14 @@ def _make_request(
...
-            '%s://%s:%s "%s %s %s" %s %s',
+            '%s://%s:%s "%s %s HTTP/%s" %s %s',
             self.scheme,
             self.host,
             self.port,
             method,
             url,
-            # HTTP version
-            http_version,
+            response.version,
```
sethmlarson, line 554: Thoughts on using `*divmod(response.version, 10)`?

pquentin: It would give HTTP/2.0 which is less wrong but still weird? I think I would prefer storing the version string in the response? It can be done without any computation and only costs a few bytes?

sethmlarson: Yeah, I do like HTTP/2 in the logs better too.

sethmlarson: Thinking about this, we should probably keep emitting `HTTP/1.1` and `HTTP/2` since some folks might be using regular expressions on these values. Can we have an `if` block that determines the value based on the `resp.version` value?

pquentin: There's a way to avoid `if`s, what do you think of https://github.com/urllib3/urllib3/pull/3316/commits/a1f18b0166e84b83e72565d0e4f053e61fa97653?

src/urllib3/connection.py
```diff
@@ -482,6 +482,7 @@ def getresponse(  # type: ignore[override]
             headers=headers,
             status=httplib_response.status,
             version=httplib_response.version,
+            version_string=getattr(self, "_http_vsn_str", "HTTP/?"),
```
pquentin, line 485: Emscripten has its own connection class, so I don't need the getattr any more:

```suggestion
            version_string=self._http_vsn_str,
```

pquentin: But I do need to add `version_string` to https://github.com/urllib3/urllib3/blob/2aec09f6a1727dd54f65b9069ef8f168ef40a885/src/urllib3/contrib/emscripten/response.py#L47. Hopefully CI will catch this (did not run so far).

sethmlarson (approved):
Approach LGTM, looks like a linting error and Emscripten tests failing.

pquentin:
I don't think the requests failure is related but I am not sure yet.

sethmlarson (approved):
LGTM, thanks Quentin! Merge if you think the Requests issue isn't caused by this PR.

pquentin:
Confirmed the issue reproduces on main: https://github.com/urllib3/urllib3/actions/runs/9036580373/job/24833736784?pr=3391.

Merging, thanks!

src/urllib3/connectionpool.py
tcraigtyler: > Thinking about this, we should probably keep emitting `HTTP/1.1` and `HTTP/2` since some folks might be using regular expressions on these values. Can we have an `if` block that determines the value based on the `resp.version` value?

I am one who had tests that looked for the entire debug log line with the HTTP/1.1 and now they are breaking on version 2.2.2. Leaving the dot out for version 1.1 is invalid so it would be better for the long haul to bring that back.

pquentin: What are you seeing exactly? It was supposed to still be HTTP/1.1.

tcraigtyler: I'm not specifying everything in my requirements file, but when 2.2.2 became available and was installed I see.
> https://localhost:8082 "PUT /activeUser?force=true HTTP/11" 200 4

When I explicitly revert to 2.2.1 in the requirements file and rebuild the environment I see the previous behavior.
> https://localhost:8082 "PUT /activeUser?force=true HTTP/1.1" 200 4

I can open a issue.

PR #3318, opened 2024-01-25
Hide H2Connection inside _LockedObject

pquentin:
This trick was used in the hyper HTTP/2 client, and has two advantages:

 1. it is impossible to use the connection without acquiring the lock
 2. passing the connection to another object only requires one parameter

This pull request was sponsored by [Elastic](https://www.elastic.co/), my employer.


sethmlarson (approved):
LGTM!

PR #3319, opened 2024-01-25
Remove TODO about informational responses

pquentin:
Today it's not possible with HTTP/1.1 (see psf/requests#713) and even python-hyper does not handle this case.

### DIFF
--- a/src/urllib3/http2.py
+++ b/src/urllib3/http2.py
@@ -105,12 +105,7 @@ class HTTP2Connection(HTTPSConnection):
                 if received_data := self.sock.recv(65535):
                     events = h2_conn.receive_data(received_data)
                     for event in events:
-                        if isinstance(
-                            event, h2.events.InformationalResponseReceived
-                        ):  # Defensive:
-                            continue  # TODO: Does the stdlib do anything with these responses?
-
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
LGTM, is it possible to send a 1xx response from ASGI?

sigmavirus24:
I'm not sure what we might want here but I think there are use cases that make sense and we shouldn't automatically limit things because http lib sucks in this way.

See also https://github.com/python/cpython/issues/42550 and no one having fixed this yet

pquentin:
> LGTM, is it possible to send a 1xx response from ASGI?

Yes, early hints are explicitly supported: https://asgi.readthedocs.io/en/stable/extensions.html#early-hints. Good point.

> I'm not sure what we might want here but I think there are use cases that make sense and we shouldn't automatically limit things because http lib sucks in this way.

My reasoning is that there's no point in cluttering the code with something we're not going to implement right away. It's not just a comment, it's an actual condition that is evaluated. An open issue in the HTTP/2 milestone makes more sense to me. I opened https://github.com/urllib3/urllib3/issues/3336, and will merge this given there's one approval.

PR #3320, opened 2024-01-25
Move response body handling in HTTP2Response

pquentin:
Not sure what to think about this. I drew some inspiration from python-hyper.

### DIFF
--- a/src/urllib3/http2.py
+++ b/src/urllib3/http2.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import collections
 import threading
 import types
 import typing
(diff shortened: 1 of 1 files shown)


pquentin:
Closing for now as we have not yet agreed on the direction to take here. The main difficulty is that HTTP/2 uses a shared TCP connection, shared by all HTTP requests. In an async world, we would use a background task to communicate with the socket and queues to communicate with the tasks handling the actual requests, but in a sync world we cannot spawn such a thread: https://github.com/python-hyper/hyper/issues/65.

There are two main lines of thinking in the team so far.

The first option is to do what python-hyper and httpx do. Each request drives itself to completion. To do so, it has to read and write from the shared socket while holding a lock to make sure the socket state does not get corrupted. If it reads events meant for other requests, it stores them in a dictionary of events, keyed by stream id and possibly sends back a WINDOW_UPDATE frame for flow control. Before reading from the socket, it checks if there are any unprocessed events. PING frames are only answered to if there's an active connection reading data from the server. If a PING was missed, then the client may have to establish a new connection to the server.

The second option tries to simulate what an async urllib3 would do, with threads getting and sending messages from a queue, but also being able to perform the role of background thread when data should be read. The idea relies on a concurrency primitive that we aren't sure exists, which does the following:

 1. Each thread reads incoming message from their queue. and write to the socket as needed
 2. f no message is ready in any of the queues, one and only one thread is tasked to read from the socket As soon as it receives something, we go back to 1., with threads processing their message queues.

It will be easier to experiment when #3324 and an actual shared socket land.

Issue #3301, opened 2024-01-22
Allow HTTP/2 socket to be re-used for future requests


sethmlarson:
Currently connections are closed immediately after receiving their response. Instead the connection should be kept open and able to be re-used for more than on request.

illia-v:
The bounty has been removed for the reasons described in https://github.com/urllib3/urllib3/issues/3000#issuecomment-5227891874.

Please refrain from opening pull requests for this issue for the time being.

PR #3324, opened 2024-01-28
Probe and cache HTTP/2 support per-origin

sethmlarson:
Moved `_LockedObject` to `urllib3._collections`.

Adds an "HTTP/2 support" cache which does one of three things if HTTP/2 support is going to be requested from the origin:
  - Returns the cached value `True` or `False`
  - Returns `None` meaning the current connection is probing and needs to return back a result.
  - Blocks waiting for an active probe.

My thinking is that this HTTP/2 support probe cache can evolve into the HTTP/2 socket sharing mechanism. Need to implement probe cache resetting per test case, tests for the probe, and for when it's used by an HTTP/1.1 and HTTP/2 capable function.

Part of https://github.com/urllib3/urllib3/issues/3301


src/urllib3/connection.py
```diff
@@ -612,63 +663,104 @@ def set_cert(
...
-
-        is_time_off = datetime.date.today() < RECENT_DATE
-        if is_time_off:
-            warnings.warn(
-                (
-                    f"System time is way off (before {RECENT_DATE}). This will probably "
-                    "lead to SSL verification errors"
-                ),
-                SystemTimeWarning,
+        # Today we don't need to be doing this step before the /actual/ socket
```
sethmlarson, line 666: Don't love how this comment is immediately out-of-date once we proceed w/ the next step of HTTP/2 support.

src/urllib3/connection.py
```diff
@@ -612,63 +663,104 @@ def set_cert(
...
+
+        # Check if the target origin supports HTTP/2.
+        # If the value comes back as 'None' it means that the current thread
+        # is probing for HTTP/2 support. Otherwise, we're waiting for another
+        # probe to complete, or we get a value right away.
+        if "h2" in ssl_.ALPN_PROTOCOLS:
+            target_supports_http2 = _HTTP2_PROBE_CACHE.get(
+                host=probe_http2_host, port=probe_http2_port
             )
+        else:
```
sethmlarson, line 683: Since `h2` isn't in ALPN_PROTOCOLS during normal usage these new code paths aren't used for `HTTPSConnection` quite yet.

src/urllib3/connection.py
```diff
@@ -493,6 +494,56 @@ def getresponse(  # type: ignore[override]
         return response
 
 
+class _HTTP2ProbeCache:
```
sethmlarson, line 497: It might make more sense to move this whole object to the `http2` module, since it's not useful at all for non-HTTP/2 cases.

src/urllib3/connection.py
```diff
@@ -612,63 +663,104 @@ def set_cert(
...
-            warnings.warn(
-                (
-                    f"System time is way off (before {RECENT_DATE}). This will probably "
-                    "lead to SSL verification errors"
-                ),
-                SystemTimeWarning,
+        # Today we don't need to be doing this step before the /actual/ socket
+        # connection, however in the future we'll need to decide whether to
+        # create a new socket or re-use an existing "shared" socket as a part
+        # of the HTTP/2 handshake dance.
```
pquentin, line 669: So, thinking out loud, here are the cases we need to cover, bearing in mind this is the HTTPS connection.

1. If we don't support HTTP/2 (because it's not in `ALPN_PROTOCOLS`, like today), then we don't change anything. `connect()` calls should return connected sockets, without any blocking.
2. If we do support HTTP/2, there are three cases:
   1. We know if the target supports HTTP/2 or not. If it does not, it is equivalent to 1. above. If it does, we should use the existing shared socket.
   2. We don't know what the target supports, and no connection is ongoing. We should connect and prevent other connections to do so.
   3. We don't know what the target supports, and a connection is ongoing. We should wait for that connection to go through and then it becomes equivalent to 2.i.
3. With HTTP/2 prior knowledge, which is what `urllib3.http2.inject_into_urllib3` does that right now, we need to create a single connection too, like in 2.ii and 2.iii. We should do the equivalent of https://github.com/python-hyper/hyper/blob/b77e758f472f00b098481e3aa8651b0808524d84/hyper/http20/connection.py#L345-L347, even though it may be complicated as we need to support 1. and 2. too. We're not going to support this right away.

Given the above, is there a way to avoid the `None` value for the probe cache and replace by waiting on the `target_supports_http2` lock? I think this would simplify the three last lines of `_HTTP2ProbeCache.get` and avoid returning a value without holding for a lock. Maybe `_HTTP2ProbeCache.get(host, port)` could be used as a context manager even?

test/with_dummyserver/test_https.py
```diff
@@ -945,6 +947,89 @@ def test_alpn_default(self) -> None:
...
+
+        if http_version == "h2":
+            # This means the probe was successful.
+            assert http2_probe_cache.values() == {(self.host, self.port): True}
+        else:
+            # This means the probe wasn't attempted, otherwise would have a value.
+            assert http_version == "h11"
+            assert http2_probe_cache.values() == {}
+
+    @pytest.mark.xfail(reason="Hypercorn always supports both HTTP/2 and HTTP/1.1")
```
pquentin, line 975: Do we want a fixture where it only supports one?

pquentin (commented):
Thanks! Great step in the right direction.

src/urllib3/connection.py
```diff
@@ -612,63 +663,104 @@ def set_cert(
...
+        probe_http2_host = self.host if self._tunnel_host is None else self._tunnel_host
+        probe_http2_port = self.port if self._tunnel_host is None else self._tunnel_port
```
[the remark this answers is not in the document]
sethmlarson: Is this correct? `self._tunnel_port` might be `None` in situations where we don't want to use `self.port`? I'm using `if self._tunnel_host is None` to detect scenarios where we aren't tunneling.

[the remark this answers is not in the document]
sethmlarson: These lines are confusing though, I might restructure it so they don't feel like independent statements.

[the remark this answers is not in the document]
pquentin: Thanks, I missed that. Restructuring would be nice.

src/urllib3/connection.py
sethmlarson: Yes I like your suggested approach much better, thank you! Let me implement it that way.

src/urllib3/connection.py
sethmlarson: I'm not super happy with what I've done to remove the circular dependency, let me know if you have better ideas.

src/urllib3/http2.py
```diff
@@ -221,11 +191,13 @@ def inject_into_urllib3() -> None:
     urllib3.connection.HTTPSConnection = HTTP2Connection  # type: ignore[misc]
 
     # TODO: Offer 'http/1.1' as well, but for testing purposes this is handy.
+    urllib3.util.ALPN_PROTOCOLS = ["h2"]
```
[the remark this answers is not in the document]
sethmlarson: The `ALPN_PROTOCOLS` list in `ssl_` is reexported to the `urllib3.util` module even before this change, so updating that list too. We should probably remove the one in `utils`?

src/urllib3/connection.py
```diff
@@ -493,6 +494,56 @@ def getresponse(  # type: ignore[override]
...
+            except KeyError:
+                obj = _LockedObject(None)
+                self._values[(host, port)] = obj
+
+        # If the value is unknown, we acquire the lock to signal
+        # to the requesting thread that the probe is in progress
+        # or that the current thread needs to return their findings.
+        if obj._obj is None:
+            obj.lock.acquire()
+        return obj._obj
```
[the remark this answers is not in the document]
sethmlarson: Agreed, I rewrote the cache to not use `LockedObject`.

src/urllib3/connection.py
sethmlarson: This comment was from 3 weeks ago so I'm a little hazy now, I think we decided to handle the socket sharing in a separate PR but to build on the `HTTP2ProbeCache` into a proper socket-sharing solution for HTTP/2

test/with_dummyserver/test_https.py
sethmlarson: The mixed case is fine for the future, maybe this can be opened as a separate issue to allow for the server to be launched in one of two modes?

test/conftest.py
```diff
@@ -376,8 +377,17 @@ def requires_tlsv1_3(supported_tls_versions: typing.AbstractSet[str]) -> None:
...
+        yield request.param
+    finally:
+        if request.param == "h2":
+            urllib3.http2.extract_from_urllib3()
 
-    yield request.param
 
-    if request.param == "h2":
-        urllib3.http2.extract_from_urllib3()
+@pytest.fixture(autouse=True)
```
pquentin, line 387: Would it make sense to be more explicit than `autouse`, and make `reset_http2_probe_cache` a dependency of the `http_version` fixture?

This means you'll need to add it to `test_http2_probe_no_result_in_connect_error` and `test_http2_probe_no_result_in_ssl_error`, which makes sense anyway.

src/urllib3/connection.py
```diff
@@ -612,63 +613,111 @@ def set_cert(
...
+
+        # Check if the target origin supports HTTP/2.
+        # If the value comes back as 'None' it means that the current thread
+        # is probing for HTTP/2 support. Otherwise, we're waiting for another
+        # probe to complete, or we get a value right away.
+        target_supports_http2: bool | None
+        if "h2" in ssl_.ALPN_PROTOCOLS:
+            target_supports_http2 = http2_probe.get(
+                host=probe_http2_host, port=probe_http2_port
             )
```
pquentin, line 638: Should we add a comment here saying that this can block? Or rename `get` to `acquire_and_get` maybe?

src/urllib3/connection.py
```diff
@@ -612,63 +613,111 @@ def set_cert(
...
+        # of the HTTP/2 handshake dance.
+        probe_http2_port: int | None
+        if self._tunnel_host is None:
+            probe_http2_host = self.host
+            probe_http2_port = self.port
+        else:
+            probe_http2_host = self._tunnel_host
+            probe_http2_port = self._tunnel_port
+        if probe_http2_port is None:
+            probe_http2_port = 443
```
pquentin, line 628: This is clearly a nit, but I find making the first branch of the `if` related to the condition clearer. And I like to separate different conditions by a newline. Feel to ignore me!

```suggestion
        probe_http2_port: int | None
        if self._tunnel_host is not None:
            probe_http2_host = self._tunnel_host
            probe_http2_port = self._tunnel_port
        else:
            probe_http2_host = self.host
            probe_http2_port = self.port

        if probe_http2_port is None:
            probe_http2_port = 443
```

src/urllib3/http2/probe.py
```diff
@@ -0,0 +1,88 @@
...
+
+class _HTTP2ProbeCache:
+    __slots__ = (
+        "_lock",
+        "_cache_locks",
+        "_cache_values",
+    )
+
+    def __init__(self) -> None:
+        self._lock = threading.RLock()
```
pquentin, line 14: This lock does not need to be re-entrant, right? It seems to me there's readability value in only marking locks reentrant when it's needed, like it's the case for the key locks. That way I don't have to think about possible cases where the lock is acquired multiple times.

src/urllib3/http2/probe.py
```diff
@@ -0,0 +1,88 @@
...
+        # or that the current thread needs to return their findings.
+        key_lock = self._cache_locks[key]
+        key_lock.acquire()
+        try:
+            # If the by the time we get the lock the value has been
+            # updated we want to return the updated value.
+            value = self._cache_values[key]
+
+        # In case an exception like KeyboardInterrupt is raised here.
+        # KeyError shouldn't be possible.
```
pquentin, line 44: Should we assert that this isn't a `KeyError` then? They're removed with `python -O` so it's not like it's a huge performance burden.

src/urllib3/http2/probe.py
```diff
@@ -0,0 +1,88 @@
...
+    def values(self) -> dict[tuple[str, int], bool | None]:
+        """This function is for testing purposes only. Gets the current state of the probe cache"""
+        with self._lock:
+            return {k: v for k, v in self._cache_values.items()}
+
+    def reset(self) -> None:
+        """This function is for testing purposes only. Reset the cache values"""
+        with self._lock:
+            self._cache_locks = {}
+            self._cache_values = {}
```
pquentin, line 73: Should we call them `_values` and `_reset` then?

pquentin (dismissed):
This looks incredible, thanks! I really like the new direction, that I find much clearer, and that I think will make implement socket sharing easy, at least on the connection side.

LGTM if CI passes!

test/conftest.py
sethmlarson: I'm going to keep this as-is, since even in this current PR the HTTP/2 probe cache is used in tests that don't use the `http_version` fixture.

sethmlarson:
Addressed your feedback @pquentin and added another test case for blocking timings.

src/urllib3/connection.py
pquentin: Yes! This is resolved, thanks.

src/urllib3/connection.py
pquentin: What part feels wrong? I'm happy with the current solution. The probe cache is self-contained. (The way we expose methods as module functions is slightly surprising, but is still readable in http2/connection.py)

test/conftest.py
pquentin: OK, then I have a final nit:

```suggestion
@pytest.fixture(scope="function", autouse=True)
```

src/urllib3/http2.py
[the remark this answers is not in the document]
pquentin: Yes, although that can be done in a followup PR.

src/urllib3/http2/probe.py
```diff
@@ -75,14 +78,12 @@ def reset(self) -> None:
 
 _HTTP2_PROBE_CACHE = _HTTP2ProbeCache()
 
-set = _HTTP2_PROBE_CACHE.set
-get = _HTTP2_PROBE_CACHE.get
-values = _HTTP2_PROBE_CACHE.values
-reset = _HTTP2_PROBE_CACHE.reset
+set_and_release = _HTTP2_PROBE_CACHE.set_and_release
+acquire_and_get = _HTTP2_PROBE_CACHE.acquire_and_get
```
pquentin, line 82: I love the new names :star_struck:

test/with_dummyserver/test_https.py
```diff
@@ -1035,7 +1036,53 @@ def test_http2_probe_no_result_in_ssl_error(self) -> None:
...
+            max_start_time = max(start_time for start_time, _ in values)
+            assert (
+                max_start_time - min_start_time < connect_timeout
+            )  # Everything starts at a similar time.
+
+            # But all end times are sorted by completion and the differences between
+            # end times is greater or equal to the timeout value.
+            assert sorted(end_time for _, end_time in values) == [
+                end_time for _, end_time in values
+            ]
```
pquentin, line 1077: As CI shows, I'm afraid this is not that simple. Thread scheduling (especially in crowded CI environment) can be completely arbitrary.

pquentin (approved):
Thanks! LGTM. The new test may be a bit too ambitious though. Maybe there's a way to make it work by having tracing callbacks in connection.py and make sure that we never have two threads connecting at the same time? Not sure.

pquentin:
I'm still seeing failures in `test_http2_probe_blocked_per_thread`, with the times going up to 0.6 in this run: https://github.com/urllib3/urllib3/actions/runs/7984715792/job/21801981185?pr=3324

```
FAILED test/contrib/test_pyopenssl.py::TestHTTPS_TLSv1_2::test_http2_probe_blocked_per_thread - assert (1708495940.4867585 - 1708495939.9358299) < 0.5
FAILED test/contrib/test_pyopenssl.py::TestHTTPS_TLSv1_3::test_http2_probe_blocked_per_thread - assert (1708495955.7278147 - 1708495955.1609907) < 0.5
FAILED test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_http2_probe_blocked_per_thread - assert (1708496024.1679983 - 1708496022.7132747) > (0.5 * (5 - 2))
FAILED test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_3::test_http2_probe_blocked_per_thread - assert (1708496034.525668 - 1708496033.9327052) < 0.5
```

This continues showing that time-based tests are going to be fragile in CI. Is it still useful with a larger time, like say 1s instead of 0.5s?

pquentin:
As far as I can tell, the two issues left are the flaky test and the coverage. I opened https://github.com/sethmlarson/urllib3/pull/15 to address the flaky test.

src/urllib3/http2/probe.py
```diff
@@ -0,0 +1,89 @@
...
+    def set_and_release(
+        self, host: str, port: int, supports_http2: bool | None
+    ) -> None:
+        key = (host, port)
+        key_lock = self._cache_locks[key]
+        with key_lock:  # Uses an RLock, so can be locked again from same thread.
+            if supports_http2 is None and self._cache_values[key] is not None:
+                raise ValueError(
+                    "Cannot reset HTTP/2 support for origin after value has been set."
+                )
```
pquentin, line 62: This line is not covered. Is this more of a defensive thing that can't happen if the probe is used as intended?

pquentin (commented):
I merged the main branch, fixing the conflict. Still left with the two coverage issues.

src/urllib3/http2/probe.py
sethmlarson: This is defensive!

pquentin:
I opened https://github.com/sethmlarson/urllib3/pull/16 as a proposal to fix the remaining coverage issue.

pquentin (approved):
This looks good to me now! I do need a review for my own commits.

sethmlarson:
Your commits LGTM, let's merge this monster!

PR #3328, opened 2024-01-29
Ensure no remote connections during testing

miketheman:
As a way to prevent the test suite from making remote calls to anywhere unexpected, use the `pytest-socket` plugin to disable any calls to hosts other than localhost.


miketheman:
As this is part of the test infra, I think it could "Skip changelog".

sethmlarson (approved):
LGTM, thanks Mike!

illia-v (approved):

Issue #3330, opened 2024-01-30
Create a workflow (nox?) for testing Emscripten support locally


sethmlarson:
Would be great to have an easy-to-use workflow for contributors to:

* Run a single command
* Have all dependencies installed
* Opens a Pyodide console in a web browser
* Makes the local copy of urllib3 available for installation with micropip/Pyodide

This would help greatly with being able to poke around with and develop Emscripten/Pyodide support.

PR #3333, opened 2024-01-31
Fix issue where InsecureRequestWarning is emitted for HTTPS Emscripte…

### LINKED ISSUE #3331: Emscripten support emits an InsecureRequestWarning even when using HTTPS
This is a side-effect of us using JavaScript APIs instead of Python TLS and setting `is_verified` on the `EmscriptenHTTPConnection` so urllib3 is emitting an `InsecureRequestWarning` for every request, even ones that are using HTTPS.

* Set the proper value of `is_verified` depending on whether the request is HTTP or HTTPS.
* Add a test case that asserts that an `InsecureRequestWarning` is emitted for HTTP and isn't emitted for HTTPS.

sethmlarson:
…n requests

Closes #3331 I wasn't able to get tests to run locally and spent a bunch of time trying to get Jupyterlite+micropip to install a local wheel file to no avail :cry: So will be relying on CI for this one, unfortunately. We really need #3330

### DIFF
--- a/test/contrib/emscripten/test_emscripten.py
+++ b/test/contrib/emscripten/test_emscripten.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import sys
 import typing
+import warnings
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


test/contrib/emscripten/test_emscripten.py
```diff
@@ -947,3 +948,26 @@ def count_calls(self, *args, **argv):  # type: ignore[no-untyped-def]
...
+    selenium_coverage: typing.Any, testserver_http: PyodideServerInfo
+) -> None:
+    @run_in_pyodide  # type: ignore[misc]
+    def pyodide_test(selenium_coverage, host: str, port: int, https_port: int) -> None:  # type: ignore[no-untyped-def]
+        import urllib3
+        import urllib3.exceptions
+
+        http = urllib3.PoolManager()
+
+        with warnings.catch_warnings(record=True) as w:
```
pquentin, line 964: CI fails because you haven't imported `warnings`

sethmlarson: Thanks! Fixed in https://github.com/urllib3/urllib3/pull/3333/commits/2758cab1668a86f51d873ea9d6041ad558c07a09

pquentin (approved):
Thanks! LGTM.

PR #3338, opened 2024-02-01
Add nox configuration to start pyodine console

### LINKED ISSUE #3330: Create a workflow (nox?) for testing Emscripten support locally

ecerulm:
This adds a nox session called `pyodineconsole`. 

```
nox -rs pyodineconsole
nox > Running session pyodineconsole
nox > Re-using existing virtual environment at .nox/pyodineconsole.
nox > python -m pip install build
nox > python -m build
* Creating venv isolated environment...
* Installing packages in isolated environment... (hatchling>=1.6.0,<2)
* Getting build dependencies for sdist...
* Building sdist...
* Building wheel from sdist
* Creating venv isolated environment...
* Installing packages in isolated environment... (hatchling>=1.6.0,<2)
* Getting build dependencies for wheel...
* Building wheel...
Successfully built urllib3-2.2.0.tar.gz and urllib3-2.2.0-py3-none-any.whl
nox > cp templates/pyodide-console.html dist/index.html
nox > cd dist
nox > python -m http.server
Serving HTTP on :: port 8000 (http://[::]:8000/) ...
```

This nox session
* builds the wheel for `urllib` into `dist/`
* creates a `index.html` file in `dist/`, with a pyodine terminal taken 
from [pyodine src/templates/console.html](https://github.com/pyodide/pyodide/blob/main/src/templates/console.html)
* autoruns a series of commands on this terminal 
  * `import micropip`
  * `await micropip.install("http://localhost:8000/urllib3-2.2.0-py3-none-any.whl")`
  * `import urllib3`
  * `urllib3.__version__`

Then you open http://localhost:8000 and you will be presented with 

image

Closes #3330

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -200,6 +200,18 @@ def lint(session: nox.Session) -> None:
     mypy(session)
 
 
+@nox.session(python="3.11")
+def pyodineconsole(session: nox.Session) -> None:
+    # build wheel into dist folder
+    session.install("build")
+    session.run("python", "-m", "build")
+    session.run(
+        "cp", "templates/pyodide-console.html", "dist/index.html", external=True
+    )
+    session.cd("dist")
+    session.run("python", "-m", "http.server")
+
+
 # TODO: node support is not tested yet - it should work if you require('xmlhttprequest') before
 # loading pyodide, but there is currently no nice way to do this with pytest-pyodide
 # because you can't override the test runner properties easily - see
--- a/noxfile.py
+++ b/noxfile.py
@@ -201,12 +201,15 @@ def lint(session: nox.Session) -> None:
 
 
 @nox.session(python="3.11")
-def pyodineconsole(session: nox.Session) -> None:
+def pyodideconsole(session: nox.Session) -> None:
     # build wheel into dist folder
     session.install("build")
     session.run("python", "-m", "build")
     session.run(
-        "cp", "templates/pyodide-console.html", "dist/index.html", external=True
+        "cp",
+        "test/contrib/emscripten/templates/pyodide-console.html",
+        "dist/index.html",
+        external=True,
     )
     session.cd("dist")
     session.run("python", "-m", "http.server")
--- a/templates/pyodide-console.html
+++ b/templates/pyodide-console.html
@@ -1,0 +1,267 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <script src="https://cdn.jsdelivr.net/npm/jquery"></script>
+    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.35.2/js/jquery.terminal.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/jquery.terminal@2.35.2/js/unix_formatting.min.js"></script>
+    <link
+      href="https://cdn.jsdelivr.net/npm/jquery.terminal@2.35.2/css/jquery.terminal.min.css"
+      rel="stylesheet"
+    />
+    <style>
+      .terminal {
+        --size: 1.5;
+        --color: rgba(255, 255, 255, 0.8);
+      }
+      .noblink {
+        --animation: terminal-none;
+      }
+      body {
+        background-color: black;
+      }
+      #jquery-terminal-logo {
+        color: white;
+        border-color: white;
+        position: absolute;
+        top: 7px;
+        right: 18px;
+        z-index: 2;
+      }
+      #jquery-terminal-logo a {
+        color: gray;
+        text-decoration: none;
+        font-size: 0.7em;
+      }
+      #loading {
+        display: inline-block;
+        width: 50px;
+        height: 50px;
+        position: fixed;
+        top: 50%;
+        left: 50%;
+        border: 3px solid rgba(172, 237, 255, 0.5);
+        border-radius: 50%;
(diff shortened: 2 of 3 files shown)


ecerulm:
@sethmlarson , please try this pyodide console approach and tell me if that's what you were looking for when you wrote #3330

noxfile.py
```diff
@@ -200,6 +200,18 @@ def lint(session: nox.Session) -> None:
     mypy(session)
 
 
+@nox.session(python="3.11")
```
sethmlarson, line 203: Does this need to be a specific Python version, or can we use any 3.x?

noxfile.py
```diff
@@ -200,6 +200,18 @@ def lint(session: nox.Session) -> None:
...
+def pyodineconsole(session: nox.Session) -> None:
+    # build wheel into dist folder
+    session.install("build")
+    session.run("python", "-m", "build")
+    session.run(
+        "cp", "templates/pyodide-console.html", "dist/index.html", external=True
+    )
+    session.cd("dist")
+    session.run("python", "-m", "http.server")
```
sethmlarson, line 212: Let's specify a port here and then open the browser automatically, I believe there's a Python function for that?

templates/pyodide-console.html
```diff
@@ -0,0 +1,267 @@
+<!doctype html>
```
sethmlarson, line 1: Can we find a new home for this file? Maybe under `test/` for Emscripten?

Did you write this file yourself or is it from somewhere, I don't see license information.

What is the expected maintenance requirements for this file?

sethmlarson (commented):
Thanks for opening this! I put some early comments below:

noxfile.py
ecerulm: but I don't think we can let the `python -m http.server` in the background with nox. So we can open the browser before the server is up , but that would be great either.

sethmlarson: Oh right, yeah let's keep this as-is!

noxfile.py
ecerulm: If we don't specify a version then nox will run the same python interpreter used to run nox itself. 

I mean, the alternative skip nox (which is for test automation, not for this). 

I can create a python script to create virtualenv , install dependencies, etc instead. then we are not bound by the nox rules.

noxfile.py
ecerulm: I can create a python script to create virtualenv , install dependencies, etc instead. then we are not bound by the nox rules.

templates/pyodide-console.html
```diff
@@ -0,0 +1,267 @@
...
+        if (searchParams.has("noblink")) {
+          $(".cmd-cursor").addClass("noblink");
+        }
+        await term.ready;
+        await term.exec("import micropip\n");
+        await term.exec("micropip.list()\n");
+        await term.exec('await micropip.install("http://localhost:8000/urllib3-2.2.0-py3-none-any.whl")')
+        await term.exec("micropip.list()");
+        await term.exec("import urllib3");
+        await term.exec("urllib3.__version__");
```
ecerulm, line 262: @sethmlarson , this is the part that I added to the original https://github.com/pyodide/pyodide/blob/3500c17468e474902044413679b024dc7af75621/src/templates/console.html

templates/pyodide-console.html
ecerulm: > Can we find a new home for this file? Maybe under test/ for Emscripten?

Sure, I can do that. 


> Did you write this file yourself or is it from somewhere, I don't see license information.

I mentioned in the PR description 
>> creates a index.html file in dist/, with a pyodine terminal taken
from [pyodide src/templates/console.html]

The LICENSE is https://github.com/pyodide/pyodide/blob/main/LICENSE. (Mozilla Public License Version 2.0) 

I can add a comment on the top of the file with the original source url, not sure , should I add a "copyright" comment although the original does not have it? The closest copyright notice that I can find in their repo is  "2019-2022, Pyodide contributors and Mozilla", I can add that?  


> What is the expected maintenance requirements for this file?

I guess we won't maintain this file. It's the same jQuery REPL that is used in https://pyodide.org/en/stable/console.html, if it stops working we can always copy the latest from [pyodide src/templates/console.html]. And then add the lines

```
        await term.ready;
        await term.exec("import micropip\n");
        await term.exec("micropip.list()\n");
        await term.exec('await micropip.install("http://localhost:8000/urllib3-2.2.0-py3-none-any.whl")')
        await term.exec("micropip.list()");
        await term.exec("import urllib3");
        await term.exec("urllib3.__version__");
```

 [pyodide src/templates/console.html]: https://github.com/pyodide/pyodide/blob/main/src/templates/console.html

sethmlarson: Thank you, apologies for not reading the description closely enough

sethmlarson (approved):
This is awesome and perfectly recreates the issue w/ InsecureRequestWarning, thank you @ecerulm!

sethmlarson:
@ecerulm The linked issue was a bounty issue for $100, so if you're able to please submit an expense to our Open Collective linking back to this PR.

PR #3340, opened 2024-02-06
Bump cryptography to 42.0.2 and PyOpenSSL to 24.0.0

ecerulm:
cryptography 41.0.6 has a moderate security vulnerability

https://github.com/urllib3/urllib3/security/dependabot/12

> A flaw was found in the python-cryptography package. This issue may allow a remote attacker to decrypt captured messages in TLS servers that use RSA key exchanges, which may lead to exposure of confidential or sensitive data.

@dependabot has identified this in #3339 but it failed to update PyOpenSSL accordingly
hence this PR

Closes #3339

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,10 +3,10 @@ coverage==7.4.1
 PySocks==1.7.1
 pytest==7.4.4
 pytest-timeout==2.1.0
-pyOpenSSL==23.2.0
+pyOpenSSL==24.0.0
 idna==3.4
 trustme==1.1.0
-cryptography==41.0.6
+cryptography==42.0.2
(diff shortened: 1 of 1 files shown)


ecerulm:
The vulnerability seems to affect only servers so in principle I guess that it does not really matter for urllib3.

Also I wonder if we should just use 

```
cryptography>=42.0.0
```

instead of pinning a specific version like we do now.

pquentin (approved):
Thanks! LGTM. Pinning avoids surprises and means there's one less thing to worry about when we see a test failing. The downside is indeed that we need to work to upgrade pins. We unfortunately can't use pip-compile because it only supports one Python version at at a time.

PR #3342, opened 2024-02-13
Update docs for ``.json()``

kevin-paulson-mindbridge-ai:
The `json.loads` [method](https://docs.python.org/3/library/json.html#json.loads) allows for other encoding than UTF-8, but notes that the encoding should be UTF-8, UTF-16 or UTF-32. In reviewing history (https://github.com/urllib3/urllib3/pull/2250 & https://github.com/urllib3/urllib3/issues/2243) I couldn't find any particular reason why this was ensured to be parsed as UTF-8.

Let me know if I should create an issue for this or if there is anything else I can do to support this.


sethmlarson:
We're following [RFC 8529 Section 8.1](https://www.rfc-editor.org/rfc/rfc8259#section-8.1) by using only UTF-8. The CPython standard library may have been decided before UTF-8 was required for JSON in open systems.

sigmavirus24:
Besides even if we wanted to allow it, this isn't the way. An optional keyword-only parameter (defaulting to utf-8) to the method is the way.

kevin-paulson-mindbridge-ai:
OK, I'll re-work this to just be a documentation update

sethmlarson (approved):
Thanks!

PR #3344, opened 2024-02-16
Stop casting request headers to HTTPHeaderDict

### LINKED ISSUE #3343: urllib3 2.2 explicitly casts all headers to HTTPHeaderDict
### Subject

Originally reported in https://github.com/boto/botocore/issues/3111, we noticed users started experiencing TypeErrors being raised from urllib3 around the time 2.2.0 was released. After a deeper dive, it appears a change was made in https://github.com/urllib3/urllib3/commit/4ece59b2d438f605dcb9a2448e1155075cd2fe07 that started indiscriminately casting headers passed to `urlopen` as an `HTTPHeaderDict`. This has specific implications for header values that are defined as bytes, while this doesn't match the current typing information, urllib3 (and Requests) both supported this prior to the release of urllib3 2.0.

Talking to @pquentin, it sounds like this is still an intended use case, but there isn't a formal contract at this time. Below I've included a minimal example to demonstrate the error. We would like to get confirmation if this is an undocumented breakage in 2.x or if this is unintended behavior. From the perspective of `HTTPHeaderDict` in isolation, it seems odd to allow instantiation when it fails immediately on use.

I'm happy to look at supplying a patch once we determine the intended behavior.

### Environment

Reproducible on all platforms with urllib3 2.2.0.

### Steps to Reproduce

```python
from urllib3._collections import HTTPHeaderDict

hdict = HTTPHeaderDict({'x-my-header': b'my-binary-value'})
# Any access that invokes `.get()` will fail (print, get, getall, etc.)
print(hdict)
```

### Expected Behavior

`HTTPHeaderDict` either fails at initialization or supports bytes values. 

### Actual Behavior

```bash
>>> print(hdict)
  File "/private/tmp/.venv/lib/python3.11/site-packages/urllib3/_collections.py", line 423, in __repr__
    return f"{type(self).__name__}({dict(self.itermerged())})"
                                    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/.venv/lib/python3.11/site-packages/urllib3/_collections.py", line 446, in itermerged
                  ^^^^^^^^^^^^^^^^^^
```

pquentin:
While this was done to fix a mypy error, we did not notice the consequences:

 * This breaks boto3 that subclasses HTTPConnection because HTTPHeaderDict does not support bytes values yet.
 * When proxying, headers are still a dictionary by default.

We can decide to reintroduce a forced conversion to HTTPHeaderDict in urllib3 3.0 if the above issues are fixed.

Closes #3343, partly reverts #3134

### DIFF
--- a/changelog/3343.bugfix.rst
+++ b/changelog/3343.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``HTTPConnectionPool.urlopen`` to stop automatically casting non-proxy headers to ``HTTPHeaderDict``. This change was premature as it did not apply to proxy headers and ``HTTPHeaderDict`` does not handle byte header values correctly yet.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -751,8 +751,8 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # have to copy the headers dict so we can safely change it without those
         # changes being reflected in anyone else's copy.
         if not http_tunnel_required:
-            headers = HTTPHeaderDict(headers)
-            headers.update(self.proxy_headers)
+            headers = headers.copy()  # type: ignore[attr-defined]
+            headers.update(self.proxy_headers)  # type: ignore[union-attr]
 
         # Must keep the exception bound to a separate variable or else Python 3
         # complains about UnboundLocalError.


nateprewitt (approved):
🚀

PR #3345, opened 2024-02-16
Release 2.2.1

pquentin:
* [x]  See if all tests, including downstream, pass
* [x]  Get the release pull request approved by a [CODEOWNER](https://github.com/urllib3/urllib3/blob/main/.github/CODEOWNERS)
* [ ]  Squash merge the release pull request with message "`Release <VERSION>`"
* [ ]  Tag with X.Y.Z, push tag on urllib3/urllib3 (not on your fork, update `<REMOTE>` accordingly)
  * Notice that the `<VERSION>` shouldn't have a `v` prefix (Use `1.26.6` instead of `v.1.26.6`)
  * ```
    # Ensure the release commit is the latest in the main branch.
    git checkout main
    git pull origin main
    git tag -s -a '<VERSION>' -m 'Release: <VERSION>'
    git push <REMOTE> --tags
    ```
* [ ]  Execute the `publish` GitHub workflow. This requires a review from a maintainer.
* [ ]  Ensure that all expected artifacts are added to the new GitHub release. Should
       be one `.whl`, one `.tar.gz`, and one `multiple.intoto.jsonl`. Update the GitHub
       release to have the content of the release's changelog.
* [ ]  Announce on:
  * [ ]  Twitter
  * [ ]  Discord
  * [ ]  OpenCollective
* [ ]  Update Tidelift metadata
* [ ]  If this was a 1.26.x release, add changelog to the `main` branch

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,12 @@
+2.2.1 (2024-02-16)
+==================
+
+- Fixed issue where ``InsecureRequestWarning`` was emitted for HTTPS connections when using Emscripten. (`#3331 <https://github.com/urllib3/urllib3/issues/3331>`__)
+- Fixed ``HTTPConnectionPool.urlopen`` to stop automatically casting non-proxy headers to ``HTTPHeaderDict``. This change was premature as it did not apply to proxy headers and ``HTTPHeaderDict`` does not handle byte header values correctly yet. (`#3343 <https://github.com/urllib3/urllib3/issues/3343>`__)
+- Changed ``ProtocolError`` to ``InvalidChunkLength`` when response terminates before the chunk length is sent. (`#2860 <https://github.com/urllib3/urllib3/issues/2860>`__)
+- Changed ``ProtocolError`` to be more verbose on incomplete reads with excess content. (`#3261 <https://github.com/urllib3/urllib3/issues/3261>`__)
+
+
 2.2.0 (2024-01-30)
 ==================
 
--- a/changelog/2860.bugfix.rst
+++ b/changelog/2860.bugfix.rst
@@ -1,1 +1,0 @@
-Use ProtocolError instead of InvalidChunkLength if response terminates before the chunk length is sent.
--- a/changelog/3261.misc.rst
+++ b/changelog/3261.misc.rst
@@ -1,1 +1,0 @@
-Made raised ``urllib3.exceptions.ProtocolError`` more verbose when a response contains content unexpectedly.
--- a/changelog/3331.bugfix.rst
+++ b/changelog/3331.bugfix.rst
@@ -1,1 +1,0 @@
-Fixed issue where ``InsecureRequestWarning`` was emitted for HTTPS connections when using Emscripten.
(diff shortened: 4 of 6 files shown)


illia-v (approved):
Thanks @pquentin!

pquentin:
@illia-v mentioned that https://github.com/urllib3/urllib3/pull/3203 made a similar change to the one reverted in #3344, so I need to inspect it before proceeding with the release. At the very least we need to be sure that botocore will work with 2.2.1.

nateprewitt:
Confirmed current tip of `main` works in botocore as expected. Looking at #3203 it's a similar change but appears to only affect the `request` API and test infrastructure. We aren't using either so I think it's safe to say we're unaffected. The one to `request` may cause similar issues for other users though.

pquentin:
Having looked more closely at https://github.com/urllib3/urllib3/pull/3203, I see two ways to avoid using `HTTPHeaderDict`. I can ignore the type errors as done in #3344, like this:

```diff
diff --git a/src/urllib3/_request_methods.py b/src/urllib3/_request_methods.py
index 632042f0..49495c3e 100644
--- a/src/urllib3/_request_methods.py
+++ b/src/urllib3/_request_methods.py
@@ -122,8 +122,8 @@ class RequestMethods:
                 headers = self.headers
 
             if not ("content-type" in map(str.lower, headers.keys())):
-                headers = HTTPHeaderDict(headers)
-                headers["Content-Type"] = "application/json"
+                headers = headers.copy()  # type: ignore[attr-defined]
+                headers["Content-Type"] = "application/json"  # type: ignore[index]
 
             body = _json.dumps(json, separators=(",", ":"), ensure_ascii=False).encode(
                 "utf-8"
```

But I would like less ignores, not more. The other option, suggested in https://github.com/python/mypy/issues/5970, is to accept that headers can indeed be immutable given the `typing.Mapping` type, which means they have to be converted to dicts before being modified:

```diff
diff --git a/src/urllib3/_request_methods.py b/src/urllib3/_request_methods.py
index 632042f0..f08b69e8 100644
--- a/src/urllib3/_request_methods.py
+++ b/src/urllib3/_request_methods.py
@@ -122,6 +122,13 @@ class RequestMethods:
                 headers = self.headers
 
             if not ("content-type" in map(str.lower, headers.keys())):
+                # While headers is most likely dict or HTTPHeaderDict, it can
+                # also be an arbitrary immutable mapping that will stay
+                # immutable even after calling copy().
+                if isinstance(headers, HTTPHeaderDict):
+                    headers = headers.copy()
+                else:
+                    headers = dict(headers)
                 headers = HTTPHeaderDict(headers)
                 headers["Content-Type"] = "application/json"
 ```

This is more complicated code for a case that we don't care about (immutable headers object that is not dict or HTTPHeaderDict).

In summary, given that:

 * the above fixes are undiserable,
 * botocore works with #3202 and the proposed release,
 * and that unlike `urllib3/connectionpool.py`, code in `urllib3/_request_methods.py` does not have proxy headers of a different type

I'd like to release 2.2.1 as is.

sethmlarson (approved):

PR #3348, opened 2024-02-20
Fix changelog entry which was inverted

sparrowt:
The original changelog entry was correct but https://github.com/urllib3/urllib3/commit/54d6edf2a671510a5c029d3b76ffe71a5b07147a inverted its meaning.

This puts it back, see https://github.com/urllib3/urllib3/pull/2860#issuecomment-1954488635


pquentin (approved):
Thanks! And sorry about that.

PR #3350, opened 2024-02-21
Remove flask pins from requests downstream tests

nateprewitt:
This PR will remove the pins put in place with #3170 now that httpbin 0.10.2 is released.


sethmlarson (approved):
Thanks Nate!

PR #3358, opened 2024-03-04
Fix tests that leak threads (pytest 8)

ecerulm:
Bumping the pytest version to 8.0.2 revealed that some tests are leaking threads. The tests create a SocketServerThread and those threads never finish. These unterminated threads confuse pytest-memray since while running a test one of these threads can do an allocation that will be counted as part of this test. 

In particular the following tests currently do not properly terminate the threads:

* test_socket_close_socket_then_file
* test_socket_close_stays_open_with_makefile_open
* test_close_after_handshake
* test_total_timeout
* test_socket_close_socket_then_file
* test_socket_close_stays_open_with_makefile_open
* test_multipart_assert_header_parsing_no_defects
* test_chunked_specified[bytes-POST-False] (which is SKIPPED but it start the server thread, then skips itself using `pytest.skip()`)
* test_chunked_specified[bytes-PUT-False]  (which is SKIPPED but it start the server thread, then skips itself using `pytest.skip()`)
* test_chunked_specified[bytes-PATCH-False] (which is SKIPPED but it start the server thread, then skips itself using `pytest.skip()`)

This PR addresses the problem by 

* waiting after each test for the thread to finish, if it does not finish it time, generate an exception. Silently continuing only hides the problem. 
* introducing a `quit_event` , on the `teardown_method` we signal the quit_event and the socket_handlers can read it, and acts appropriately.
  * this unfortunately complicates the code since now we can't just block indefinitely, we need to timeout, to check for the quit_event and terminate in that case. 
* introduces a new pytest marker `server_threads` to allow to test funcionality. I can remove this marker after but right now you can test this with `nox -rs test-3.10 -- -m server_threads` 

Closes #3335

Alternatives 
* Ask pytest-memray to give an option to only count the allocations made on the current test thread (and not on the server threads). See https://github.com/bloomberg/pytest-memray/issues/111
* Increase the limit to ` @pytest.mark.limit_memory("10.10 MB")`. This does not fix the problem but it's way easier and less convoluted. But ideally we should not leave threads lingering from test to test.
* Make sure each server threads dies of natural causes. Most of the server thread close down gracefully since the code path servers a single request and then returns. But the tests mentioned above they are not doing that.
  * some of the tests like `test_socket_close_socket_then_file` I believe they really need a outside signal
  * some test like `* test_chunked_specified[bytes-POST-False]`, should not even create a server thread in the first place, the whole test can be SKIPPED before `self._start_server(socket_handler)`.

### DIFF
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -31,7 +31,7 @@ def consume_socket(
             break
         try:
             b = sock.recv(chunks)
-        except TimeoutError:
+        except (TimeoutError, socket.timeout):
             continue
         assert isinstance(b, bytes)
         consumed += b
@@ -104,7 +104,7 @@ class SocketDummyServerTestCase:
                     try:
                         sock = listener.accept()[0]
                         break
-                    except TimeoutError:
+                    except (TimeoutError, socket.timeout):
                         continue
                 consume_socket(sock, quit_event=quit_event)
                 if quit_event.is_set():
@@ -145,8 +145,6 @@ class SocketDummyServerTestCase:
         if hasattr(self, "server_thread"):
             if self.server_thread.quit_event:
                 self.server_thread.quit_event.set()
-            else:
-                print("Can't notify the quit_event")
             self.server_thread.join(LONG_TIMEOUT * 3)
             if self.server_thread.is_alive():
                 raise Exception("server_thread did not exit")
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -82,27 +82,44 @@ class SocketDummyServerTestCase:
         if not ready_event.is_set():
             raise Exception("most likely failed to start server")
         cls.port = cls.server_thread.port
-        print("_start_server")
 
     @classmethod
     def start_response_handler(
-        cls, response: bytes, num: int = 1, block_send: threading.Event | None = None
+        cls,
+        response: bytes,
+        num: int = 1,
+        block_send: threading.Event | None = None,
     ) -> threading.Event:
         ready_event = threading.Event()
+        quit_event = threading.Event()
 
         def socket_handler(listener: socket.socket) -> None:
             for _ in range(num):
                 ready_event.set()
 
-                sock = listener.accept()[0]
-                consume_socket(sock)
+                listener.settimeout(LONG_TIMEOUT)
+                while True:
+                    if quit_event.is_set():
+                        return
+                    try:
+                        sock = listener.accept()[0]
+                        break
+                    except TimeoutError:
+                        continue
+                consume_socket(sock, quit_event=quit_event)
+                if quit_event.is_set():
+                    sock.close()
+                    return
                 if block_send:
-                    block_send.wait()
+                    while not block_send.wait(LONG_TIMEOUT):
+                        if quit_event.is_set():
+                            sock.close()
+                            return
                     block_send.clear()
                 sock.send(response)
                 sock.close()
 
-        cls._start_server(socket_handler)
+        cls._start_server(socket_handler, quit_event=quit_event)
         return ready_event
 
     @classmethod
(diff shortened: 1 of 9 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -2341,6 +2380,7 @@ def body_generator() -> typing.Generator[bytes, None, None]:
             body.seek(0, 0)
         else:
             if chunked is False:
+                # kill the
```
pquentin, line 2383: kill the?

ecerulm:
> Thanks for this work! My main worry is the reliability of the time-based logic. You would expect `LONG_TIMEOUT * 3` to be enough, but CI always surprises me in this regard. We'll have to be ready to bump that constant if needed. =

I've increased it to `LONG_TIMEOUT *2 + 5.0` , in principle the maximum time that the thread can take to notice the `quit_event` is `LONG_TIMEOUT`  and  the thread should terminate immediately (0.1s) after that. If it takes more than 5 seconds something must be wrong. 

> Thankfully bumping it will not make the happy path slower.
yeah, the if the thread actually terminates fast the `.join()` will also return fast. It won't make the happy path slower.

ecerulm (commented):
I've removed the print statements, extraneous, comments, and the unnecessary time.sleep()

test/test_ssltransport.py
```diff
@@ -135,16 +148,18 @@ def test_start_closed_socket(self) -> None:
...
         """Socket errors should be bubbled up"""
-        self.start_dummy_server()
+        self.start_dummy_server(validate=False)
 
         sock = socket.create_connection((self.host, self.port))
         with SSLTransport(
             sock, self.client_context, server_hostname="localhost"
         ) as ssock:
             ssock.close()
+            time.sleep(1)
```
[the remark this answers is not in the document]
ecerulm: That was a leftover from some other test I was doing, I've removed it

test/with_dummyserver/test_socketlevel.py
ecerulm: removed!

illia-v (dismissed):
LGTM, thanks for this!

> > Thanks for this work! My main worry is the reliability of the time-based logic. You would expect `LONG_TIMEOUT * 3` to be enough, but CI always surprises me in this regard. We'll have to be ready to bump that constant if needed. =
> 
> I've increased it to `LONG_TIMEOUT *2 + 5.0` , in principle the maximum time that the thread can take to notice the `quit_event` is `LONG_TIMEOUT` and the thread should terminate immediately (0.1s) after that. If it takes more than 5 seconds something must be wrong.

What do you think about adding a comment to the code describing why `LONG_TIMEOUT * 2 + 5.0` was chosen?

> * introduces a new pytest marker `server_threads` to allow to test funcionality. I can remove this marker after but right now you can test this with `nox -rs test-3.10 -- -m server_threads`

I'd drop `server_threads` before the PR is merged if there is no expected use for the marker.

ecerulm:
> What do you think about adding a comment to the code describing why LONG_TIMEOUT * 2 + 5.0 was chosen?

Done, in latest version

> I'd drop server_threads before the PR is merged if there is no expected use for the marker.
Done

pyproject.toml
```diff
@@ -81,6 +81,7 @@ markers = [
     "limit_memory: Limit memory with memray",
     "requires_network: This test needs access to the Internet",
     "integration: Slow integrations tests not run by default",
+    "server_threads",
```
illia-v, line 84:
```suggestion
```

dummyserver/testcase.py
```diff
@@ -100,10 +132,25 @@ def start_basic_handler(
             block_send,
         )
 
+    @staticmethod
+    def quit_server_thread(server_thread) -> None:
```
illia-v, line 136:
```suggestion
    def quit_server_thread(server_thread: SocketServerThread) -> None:
```

illia-v:
I'll merge this to include the changes in a base for a #3366 fix

PR #3361, opened 2024-03-11
Add streaming multipart/form-data encoder

alexwlchan:
This is an attempt to get the streaming multipart/form-data encoder landed again (for #51, #624). This is something that was previously worked on in #2331 and has stalled; I'd like to have a go at getting it across the line.

So far:

* I've merged the latest main into sigmavirus24’s branch. This was mostly clean; the only issues were:

    - The project now uses `import typing; typing.X` rather than `from typing import X; X`, so I had to change all of those (but reading the comments on the original PR, I think that was the preference anyway?)
    - The definition of `_TYPE_BODY` has moved from `connection.py` to `_base_connection.py`

    Otherwise I haven't touched the original code.

* Run `nox -rs lint`, `nox -rs mypy` and `nox -rs docs` to fix various style/linting issues. Again I don't think these are especially meaningful.

* I've added some new tests to increase the coverage. I think I'm up to 100% for the decoder, but not the encoder.

* I read the review comments on the original PR. I couldn't see anything outstanding, but if I've missed something I'm happy to incorporate those changes into this new branch.

Still to do:

- [ ] Increase the coverage to 100%. I don't think I'm quite there yet, but I want to get it running in CI and hopefully find what still needs doing
- [ ] Any amendments from code review in 2024

Because the original code already had several rounds of review, I'll add some inline comments to highlight my new changes, to make it easier for reviewers.


test/multipart/test_decoder.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def test_from_response_needs_content_type(self) -> None:
+        response = mock.NonCallableMagicMock(spec=urllib3.response.HTTPResponse)
+        response.headers = {}
+        response.data = b""
+
+        with pytest.raises(
+            ValueError, match="Cannot determine content-type header from response"
+        ):
+            MultipartDecoder.from_response(response)
```
alexwlchan, line 157: This test is new, to add coverage for this error.

test/multipart/test_encoder.py
```diff
@@ -0,0 +1,287 @@
...
+        self.parts = [("field", "value"), ("other_field", "other_value")]
+        self.boundary = "this-is-a-boundary"
+        self.instance = MultipartEncoder(self.parts, boundary=self.boundary)
+
+    def test_content_type(self) -> None:
+        expected = "multipart/form-data; boundary=this-is-a-boundary"
+        assert self.instance.content_type == expected
+
+    def test_content_length(self) -> None:
+        assert self.instance.content_length == str(len(self.instance))
```
alexwlchan, line 103: This test is new, to add coverage for the `content_length` property.

test/multipart/test_encoder.py
```diff
@@ -0,0 +1,287 @@
...
+            data = m.read(blocksize)
+            if not data:
+                break
+            read_so_far += len(data)
+
+        assert read_so_far == total_size
+
+        for k, v in fields.items():
+            if k != "test" and hasattr(v[1], "close"):
+                v[1].close()
```
alexwlchan, line 193: These lines are new – pytest was raising an error about unclosed files, from the `open()` on L175.

test/multipart/test_encoder.py
```diff
@@ -0,0 +1,276 @@
...
+        m = MultipartEncoder(fields=fields)
+        assert len(m.read()) > 0
+
+    @pytest.mark.parametrize(
+        "content_type",
+        [
+            b"application/json".decode("utf-8"),
+            b"application/json",
+        ],
+    )
```
alexwlchan, line 231: This is a change for coverage/completeness – previously this test only passed a `str` as the content_type, but the code can handle `str` or `bytes`.

test/multipart/test_encoder.py
```diff
@@ -0,0 +1,287 @@
...
+                (
+                    b"filename".decode("utf-8"),
+                    b"filecontent",
+                    b"application/json",
+                ),
+            )
+        ]
+        m = MultipartEncoder(fields=fields)
+        output = m.read().decode("utf-8")
+        assert output.index("Content-Type: application/json\r\n") > 0
```
alexwlchan, line 260: This test is new, to add coverage for a couple of lines in the multipart encoder:

```python
if isinstance(file_type, bytes):
    file_type = file_type.decode("utf-8")
```

alexwlchan:
I'm getting a CI failure for Python 3.13 on Windows, but I don't think it's related to this patch?

> Failed to build cffi
> ERROR: Could not build wheels for cffi, which is required to install pyproject.toml-based projects

I can see a similar failure in [the latest build on main](https://github.com/urllib3/urllib3/actions/runs/8157102229/job/22295996636).

mypy-requirements.txt
```diff
@@ -10,5 +10,6 @@ hypercorn==0.15.0
 httpx==0.25.2
 types-backports
 types-requests
+types-mock
```
alexwlchan, line 13: I was unsure what this is for – is it still needed?

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,149 @@
...
+
+    The basic usage is::
+
+        import requests
+        from requests_toolbelt import MultipartDecoder
+
+        response = request.get(url)
+        decoder = MultipartDecoder.from_response(response)
+        for part in decoder.parts:
+            print(part.headers['content-type'])
```
alexwlchan, line 76: Note to self: should these usage examples be updated to use urllib3 imports?

test/multipart/test_encoder.py
alexwlchan: The alternative is doing this with a `pytest.mark.filterwarnings` decorator, which would be a little clearer – I might change this.

src/urllib3/filepost.py
```diff
@@ -77,6 +77,8 @@ def encode_multipart_formdata(
 
         if isinstance(data, str):
             writer(body).write(data)
+        elif isinstance(data, typing.BinaryIO):
+            body.write(data.read())
```
alexwlchan, line 81: This line isn't currently tested, and I need to think more about what the intention is here, because I haven't found a way to trigger this branch.

In particular neither an open file nor an `io.BytesIO` return True here, so I wonder if a different check is what's meant (if this branch is even needed at all?):

```python
import io
import typing

with open(__file__, 'rb') as f:
    print(isinstance(f, typing.BinaryIO))  # False

bytes_io = io.BytesIO(b'Hello, World!')
print(isinstance(bytes_io, typing.BinaryIO))  # False
```

docs/user-guide.rst
```diff
@@ -315,10 +315,34 @@ dictionary in the ``fields`` argument provided to
...
+    })
+    resp = urllib3.request(
+        "POST",
+        "https://httpbin.org/post",
+        body=encoder,
+        headers=encoder.headers,
+    )
+
+    print(resp.json()["form"])
+    # {"field": "value", "myfile": "..."}
```
alexwlchan, line 344: Another note to investigate: if I run this locally, I just get `{"field": "value"}` as the output; `filename.txt` is uploaded as a file rather than a form-encoded field. 🤔

sigmavirus24:
:heart: Thank you @alexwlchan ! I don't have the time with the kiddo and hopefully you can claim the bounty (since I couldn't even if I got the PR over the line)

alexwlchan:
(Having picked this up I've immediately stalled due to some stuff in my personal life, but I do want to continue this after this week is over!

sigmavirus24:
By the way, https://github.com/requests/toolbelt/pull/316 is a reasonable addition to the enclosed but will complicate this a smidge

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class NonMultipartContentTypeError(Exception):
+    pass
+
+
+def _header_parser(headers: bytes, encoding: str) -> typing.Sequence[tuple[str, str]]:
+    string = headers.decode(encoding)
+    items = email.parser.HeaderParser().parsestr(string).items()
+    items = typing.cast(typing.List[typing.Tuple[str, str]], items)
```
alexwlchan, line 23: I'm not sure this cast is necessary – if I try to reproduce in a minimal example, mypy doesn't complain when this line is removed.

src/urllib3/multipart/decoder.py
alexwlchan: This whole docstring might need rewriting – the phrase <code>``Response``-like ``BodyPart`` objects</code> only makes sense in a requests context, I think.

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,149 @@
...
+                and part[:4] != b"--\r\n"
+                and part != b"--"
+            )
+
+        parts = content.split(b"\r\n" + boundary)
+        self.parts = tuple(body_part(x) for x in parts if test_part(x))
+
+    @classmethod
+    def from_response(
+        cls: type[MD],
```
alexwlchan, line 141: Minor: elsewhere in the project, the `cls` parameter is left untyped and mypy infers the correct type.

e.g. https://github.com/urllib3/urllib3/blob/733f638a2faa02b4ff8a9f3b5668949d39396b8b/src/urllib3/util/retry.py#L266-L268

src/urllib3/multipart/decoder.py
```diff
@@ -0,0 +1,149 @@
...
+        ct_info = tuple(x.strip() for x in self.content_type.split(";"))
+        mimetype = ct_info[0]
+        if mimetype.split("/")[0].lower() != "multipart":
+            raise NonMultipartContentTypeError(
+                f"Unexpected mimetype in content-type: '{mimetype}'"
+            )
+        for item in ct_info[1:]:
+            attr, _, value = item.partition("=")
+            if attr.lower() == "boundary":
+                self.boundary = encode_with(value.strip('"'), self.encoding)
```
alexwlchan, line 111: I wonder what happens if you receive a value which isn't properly quoted – e.g. if `value` starts with an open quote but there's no matching close quote.

test/multipart/test_decoder.py
```diff
@@ -0,0 +1,179 @@
...
+from urllib3.multipart.decoder import (
+    BodyPart,
+    ImproperBodyPartContentError,
+    MultipartDecoder,
+    NonMultipartContentTypeError,
+)
+from urllib3.multipart.encoder import MultipartEncoder, encode_with
+
+
+class TestBodyPart(unittest.TestCase):
```
alexwlchan, line 20: This and the test case in `test_encoder.py` are the only tests in the project that subclass `unittest.TestCase`.

I'm going to leave them as-is for now, but I wonder if they should be rewritten to use pytest fixtures or similar, to match the rest of the project.

alexwlchan (commented):
I've had a more thorough read of `decoder.py` and left some comments, which I'll push a commit to address shortly.

src/urllib3/multipart/decoder.py
alexwlchan: This whole docstring might need rewriting – the phrase Response-like BodyPart objects only makes sense in a requests context, I think.

alexwlchan: Fixed!

test/multipart/test_encoder.py
alexwlchan: Done!

alexwlchan:
I've made a couple more tweaks, and I think the decoder is now working. In particular I created a small Flask server that uses requests-toolbelt to send multipart-encoded responses, so I can test the code in my urllib3 fork:

```python
from flask import Flask, Response
from requests_toolbelt import MultipartEncoder

app = Flask(__name__)

@app.route('/downloads')
def downloads():
    m = MultipartEncoder(
           fields={'field0': 'value', 'field1': 'value',
                   'field2': ('filename', open('file.py', 'rb'), 'text/plain')}
        )

    s = m.to_string()

    return Response(s, mimetype=m.content_type, headers={"content-type": m.content_type})

app.run(debug=True)
```

Next up is diving into the encoder code and seeing what else needs to be done there.

src/urllib3/filepost.py
sigmavirus24: Maybe bytearray? I can't remember honestly

alexwlchan:
I've had a look through `encoder.py`, compared it to the code in requests-toolbelt, and made some inline comments on what I think needs changing for urllib3. I'll work through those next.

sethmlarson:
Hey @alexwlchan, are you still interested in landing this pull request?

alexwlchan:
Yes! Sorry for letting it get stuck again – let me get it back up to date with `origin/main` and remind myself what still needs doing. (I remember thinking it was close, but not wanting to ask for a review until I'd ticked off a few more things.)

alexwlchan:
Ah, I remembered what issue I was running into last time – the type checker! And looking at it again, I think there may be an issue with this which is unrelated to this patch – let me open that as a separate ticket, rather than making it part of this PR.

src/urllib3/fields.py
```diff
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
 import email.utils
+import io
 import mimetypes
 import typing
 
-_TYPE_FIELD_VALUE = typing.Union[str, bytes]
+_TYPE_FIELD_VALUE = typing.Union[str, bytes, io.BytesIO, io.BufferedReader]
```
alexwlchan, line 8: Note to self: I want to pull this change back out of this file because it feels like quite an invasive change! And instead consolidate this in the multipart stuff for now.

sethmlarson:
No worries! :blush: We've been sleepy on reviewing this one too, so all good :)

sethmlarson (commented):
Leaving these early API review comments here for you!

PR #3364, opened 2024-03-17
use Self from typing_extensions

### LINKED ISSUE #3363: imprecise types on `urllib3.Retry.new` / `urllib3.Retry.increment`
### Subject

when attempting to subclass `urllib3.Retry` -- one would want to implement `def increment(...) -> Self:` as both `increment` and `new` [preserve the type](https://github.com/urllib3/urllib3/blob/733f638a2faa02b4ff8a9f3b5668949d39396b8b/src/urllib3/util/retry.py#L264) but label the return value as `Retry` (essentially breaking subclassing types): https://github.com/urllib3/urllib3/blob/733f638a2faa02b4ff8a9f3b5668949d39396b8b/src/urllib3/util/retry.py#L243

if the codebase were python 3.11+ I would use `from typing import Self` -- it appears elsewhere in the codebase one-off `SelfT` `TypeVar`s are created -- that pattern could be followed here

alternatively the `TYPE_CHECKING` trick could be used and is understood by the current major type checkers:

```
if TYPE_CHECKING:
    from typing_extensions import Self
```

### Environment

N/A -- latest primary branch source, mypy 1.9

### Steps to Reproduce

```python
from typing import Any, Self
import urllib3
class MyRetry(urllib3.Retry):
    def increment(self, *a: Any, **k: Any) -> Self:  # type error on this line!
        return super().increment(*a, **k)
```

```console
$ mypy t.py
t.py:5: error: Incompatible return value type (got "Retry", expected "Self")  [return-value]
Found 1 error in 1 file (checked 1 source file)
```

### Expected Behavior

no errors

### Actual Behavior

shown above

asottile:
resolves #3363

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -427,7 +427,7 @@ class HTTPHeaderDict(typing.MutableMapping[str, str]):
             val = other.getlist(key)
             self._container[key.lower()] = [key, *val]
 
-    def copy(self) -> HTTPHeaderDict:
+    def copy(self) -> Self:
         clone = type(self)()
         clone._copy_from(self)
         return clone
@@ -462,7 +462,7 @@ class HTTPHeaderDict(typing.MutableMapping[str, str]):
         self.extend(maybe_constructable)
         return self
 
-    def __or__(self, other: object) -> HTTPHeaderDict:
+    def __or__(self, other: object) -> Self:
         # Supports merging header dicts using operator |
         # combining items with add instead of __setitem__
         maybe_constructable = ensure_can_construct_http_header_dict(other)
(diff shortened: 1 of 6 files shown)


illia-v (approved):
Thanks @asottile!

New usage of `Self` coincides with https://typing.readthedocs.io/en/latest/spec/generics.html#valid-locations-for-self, one little nit is about the tense in the news entry.

illia-v (approved):

src/urllib3/_collections.py
```diff
@@ -462,7 +462,7 @@ def __ior__(self, other: object) -> HTTPHeaderDict:
         self.extend(maybe_constructable)
         return self
 
-    def __or__(self, other: object) -> HTTPHeaderDict:
+    def __or__(self, other: object) -> Self:
```
wimglenn, line 465: @asottile  Why not also on `__ior__` above?

PR #3371, opened 2024-03-27
Add xfails for CPython 3.13.0a5

### LINKED ISSUE #3366: Investigate CI failures with Python 3.13.0a5
Tests on all supported OSes started failing with recent Python 3.13.0a5.

Tests succeeded on Linux and macOS with Python 3.13.0a4 and on Windows with Python 3.13.0a3.

Let's investigate the errors and determine on who's side fixes are needed. If urllib3's code has to be updated to work with Python 3.13, we have to check the 1.26.x branch too.

https://github.com/urllib3/urllib3/actions/runs/8317963362/job/22759351029

illia-v:
Resolves #3366

xfails in `test/with_dummyserver/test_proxy_poolmanager.py` can be dropped when CPython 3.13.0a6 is available.


test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -77,6 +78,10 @@ def teardown_class(cls) -> None:
         super().teardown_class()
         shutil.rmtree(cls.certs_dir)
 
+    @pytest.mark.xfail(
```
sethmlarson, line 81: There's a lot of duplication, can we create a marker and reuse it?

sethmlarson (commented):
Looks good to unblock CI, do we only need this until trustme creates a new release?

illia-v:
> Looks good to unblock CI, do we only need this until trustme creates a new release?

When there is a new trustme release, we can pin the package to it instead of to the commit

The xfails can be dropped when Python 3.13 alpha 6 is available; the expected release date is [April 9th](https://peps.python.org/pep-0719/#schedule) + maybe a few days for GitHub Actions to provide the new version

test/with_dummyserver/test_proxy_poolmanager.py
illia-v: Sure

sethmlarson (approved):
LGTM

PR #3380, opened 2024-04-12
Drop xfails for CPython 3.13.0a5

illia-v:
The issue causing the expected failures has been resolved in the recent CPython 3.13.0a6


sethmlarson (approved):
LGTM!

PR #3389, opened 2024-04-25
perf: no need to decode bytes to str for `json.loads`

trim21:
In a old version of python, `json.loads` take str only. But in python >= 3.6, it can parse bytes directly.

So there is no need to decode body to str, just feed bytes to `json.loads` is fine, for python>=3.8.

https://docs.python.org/3.8/library/json.html#json.loads

### DIFF
--- a/changelog/3389.misc.rst
+++ b/changelog/3389.misc.rst
@@ -1,0 +1,1 @@
+Improve performance for ``.json()``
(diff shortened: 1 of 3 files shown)


trim21:
ci failed with a non relative tests 

FAILED test/with_dummyserver/test_connectionpool.py::TestConnectionPoolTimeouts::test_timeout - AssertionError: timeout was pool-level SHORT_TIMEOUT rather than request-level LONG_TIMEOUT

sigmavirus24:
Please provide performance benchmarking to justify this

trim21:
you are right， benchmark says no such thing

Issue #3392, opened 2024-05-10
Fix test_redirecting_to_bad_url failure in Requests


pquentin:
As part of our CI, we run the Requests test suite to make sure we don't accidentally break compatibility. The signal-to-noise ratio has been surprisingly high. That said, we are [currently seeing a failure](https://github.com/urllib3/urllib3/actions/runs/9036580373/job/24833736784?pr=3391) in `test_redirecting_to_bad_url`:

```python
_ TestPreparingURLs.test_redirecting_to_bad_url[http://localhost:-1-InvalidURL] _
self = <tests.test_requests.TestPreparingURLs object at 0x7fb1dde19be0>
httpbin = <function prepare_url.<locals>.inner at 0x7fb1dccd0ea0>
url = 'http://localhost:-1'
exception = <class 'requests.exceptions.InvalidURL'>
    @pytest.mark.parametrize("url, exception", (("http://localhost:-1", InvalidURL),))
    def test_redirecting_to_bad_url(self, httpbin, url, exception):
>       with pytest.raises(exception):
E       Failed: DID NOT RAISE <class 'requests.exceptions.InvalidURL'>
tests/test_requests.py:2726: Failed
----------------------------- Captured stderr call -----------------------------
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.3/x64/lib/python3.12/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/urllib3/urllib3/.nox/downstream_requests/lib/python3.12/site-packages/flask/app.py", line 1498, in __call__
    return self.wsgi_app(environ, start_response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/urllib3/urllib3/.nox/downstream_requests/lib/python3.12/site-packages/flask/app.py", line 1480, in wsgi_app
    return response(environ, start_response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/urllib3/urllib3/.nox/downstream_requests/lib/python3.12/site-packages/werkzeug/wrappers/response.py", line 576, in __call__
    app_iter, status, headers = self.get_wsgi_response(environ)
[...]

PR #3393, opened 2024-05-12
Mark broken Requests test as failing

pquentin:
Relates #3392 where @franekmagiera and @sigmavirus24 confirmed that the bug is not in urllib3.

We may not need it depending on the outcome of https://github.com/psf/requests/pull/6700 (that I've just seen, sorry).


pquentin:
Closing as the Requests pull request has been merged, sorry for the noise.

PR #3397, opened 2024-05-26
Fix a small typo in `retry.py`

rz-vastdata:


sigmavirus24 (changes requested):
The first sentence of the description of this explicitly says "on the second retry". By default urllib3 retries immediately after the initial failure. Thus your edit is wrong as this is not a typo.

rz-vastdata:
Sorry, my mistake 😞

PR #3408, opened 2024-06-17
Bump pypa/gh-action-pypi-publish to v1.9.0

illia-v:
This should fix publishing 2.2.x releases declaring `Metadata-Version: 2.3`


pquentin (approved):
Thanks! LGTM. It contains [gh-action-pypi-publish 1.8.13](https://github.com/pypa/gh-action-pypi-publish/releases/tag/v1.8.13) that has the fix we need.

PR #3411, opened 2024-06-17
Test publish workflow on every push

pquentin:
As documented in https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/. The `testpypi` GitHub environment is not set up yet, and nothing is tested.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,12 @@
 name: Publish to PyPI
 
-on: push
+on:
+  push:
+    branches:
+      - "1.26.x"
+      - "main"
+    tags:
+      - "*"
 
 permissions:
   contents: read
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,19 +1,14 @@
 name: Publish to PyPI
 
-on:
-  push:
-    tags:
-      - "*"
+on: push
 
 permissions:
   contents: read
 
 jobs:
   build:
-    name: "Build dists"
+    name: "Build distribution 📦"
     runs-on: "ubuntu-latest"
-    environment:
-      name: "publish"
     outputs:
       hashes: ${{ steps.hash.outputs.hashes }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,19 +1,14 @@
 name: Publish to PyPI
 
-on:
-  push:
-    tags:
-      - "*"
+on: push
```
illia-v, line 3: Do I understand correctly that it will deploy to test PyPI on a push to any branch?

pquentin: Yes, that's what it means. We could limit to to 1.26.x and main, in order to avoid the dependabot pull requests to generate deploys.

pquentin:
Blocked on #3412.

.github/workflows/publish.yml
illia-v: Then with the versioning scheme from #3412 it's possible that two releases can be created from different branches with the same number of commits after a release, right? If so, PyPI won't accept them

pquentin: I've pushed [`b6c02e1` (#3411)](https://github.com/urllib3/urllib3/pull/3411/commits/b6c02e11b3c6590e65fcffa42276b53a2ea9d91b) to only publish to Test PyPI for tags and the main/1.26.x branches. It remains to be seen if we'll adopt #3412 but doing that sounds like a good idea anyway. I also limited the GitHub Test PyPI Environment to main/1.26.x branches and tags.

illia-v: Nice, thanks!

illia-v (approved):

pquentin:
Good news, our current publishing code works on `main`. I tested it in my fork here: https://github.com/pquentin/urllib3/actions/runs/9592327092/job/26450752198. (It uploaded 2.0.2dev210 because 2.0.2 is the latest tag my fork has.)

PR #3413, opened 2024-06-18
Use more precise clock for test_timeout on Windows

pquentin:
On Windows, the clock used by `time.time()` has a resolution that is higher than 15ms, which sometimes breaks `test_timeout` that relies on a short timeout. Using `time.perf_counter()` makes this issue less likely.

You can such a failure here: https://github.com/urllib3/urllib3/actions/runs/9552707050/job/26329891735

```
   File "D:\a\urllib3\urllib3\test\with_dummyserver\test_connectionpool.py", line 107, in test_timeout
    assert delta >= (LONG_TIMEOUT - 1e-5), message
AssertionError: timeout was pool-level SHORT_TIMEOUT rather than request-level LONG_TIMEOUT
assert 0.4998745918273926 >= (0.5 - 1e-05)
```

The test fails because the timeout was 12.5ms shorter than expected.

Also, the Retry-After tests do need time.time() so I have added a comment as I've tried multiple times to switch away from it in the past.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -98,10 +98,10 @@ class TestConnectionPoolTimeouts(SocketDummyServerTestCase):
             self.host, self.port, timeout=short_timeout, retries=False
         ) as pool:
             wait_for_socket(ready_event)
-            now = time.time()
+            now = time.perf_counter()
             with pytest.raises(ReadTimeoutError):
                 pool.request("GET", "/", timeout=LONG_TIMEOUT)
-            delta = time.time() - now
+            delta = time.perf_counter() - now
 
             message = "timeout was pool-level SHORT_TIMEOUT rather than request-level LONG_TIMEOUT"
             assert delta >= (LONG_TIMEOUT - 1e-5), message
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):

tcdent:
Seeing this fail on Windows intermittently. 

Here's an run where this is the only failing test:

https://github.com/urllib3/urllib3/actions/runs/9823167408/job/27120736388?pr=3421

pquentin:
Right, the failure is still there. I still think using that theoretically more precise clock was the right thing to do, but something else is going on as we're still seeing the same differences of about 1 ms.

Ideally I would like to understand the issue, but I am also OK to allow a bigger difference (1ms instead of the current 0.01 ms) for Windows.

tcdent:
Added a conditional for Windows and gave it a bigger difference. Chose `1e-03` because it passed the cases I have seen fail.

```
>>> assert 0.49987029999999777 >= (0.5 - 1e-05)
AssertionError
>>> assert 0.49987029999999777 >= (0.5 - 1e-04)
AssertionError
>>> assert 0.49987029999999777 >= (0.5 - 1e-03)
>>> 
```

pquentin:
Thank you! That looks great. Can you please open a pull request?

Issue #3302, opened 2024-01-22
Add support for sending a request body with HTTP/2


sethmlarson:
* Requests that don't have a body should set `end_stream=True` on the Headers event.
* There should never be an empty `Data` event, some HTTP/2 implementations don't handle this properly.
* Must respect window of server. Maybe this is handled automatically?

Nightrider0098:
Hey was working on the issue. Tried to add a checks and while loop when handling body. 
https://github.com/urllib3/urllib3/pull/3322

Is it the correct way to handle the body, or i need to do something different.

tcdent:
Opened a pull request for this: https://github.com/urllib3/urllib3/pull/3421

sethmlarson:
Closed in https://github.com/urllib3/urllib3/pull/3421

PR #3421, opened 2024-07-03
Add support for sending a request body with HTTP/2 #3302

tcdent:
* Support identical method signature on `send` as built-in `http` library.
* Borrowed and re-implemented header validation patterns from built-in `http` library.
* Removed `h2_` prefix from attributes for brevity.

### DIFF
--- a/src/urllib3/http2.py
+++ b/src/urllib3/http2.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
+import re
 import threading
 import types
 import typing
@@ -10,15 +12,26 @@ import h2.events  # type: ignore[import-untyped]
 
 import urllib3.connection
 import urllib3.util.ssl_
-from urllib3.response import BaseHTTPResponse
 
+from ._base_connection import _TYPE_BODY
 from ._collections import HTTPHeaderDict
-from .connection import HTTPSConnection
+from .connection import HTTPSConnection, _get_default_user_agent
 from .connectionpool import HTTPSConnectionPool
+from .exceptions import ConnectionError
+from .response import BaseHTTPResponse
 
 orig_HTTPSConnection = HTTPSConnection
 
 T = typing.TypeVar("T")
+
+log = logging.getLogger(__name__)
+
+_is_legal_header_name = re.compile(rb"^:?[a-z0-9!#$%&\'*+\-.^_`|~]+$").match
+RE_IS_ILLEGAL_HEADER_VALUE = re.compile(rb"[\x00\x0b\x0c]|\n(?![ \t])|\r(?![ \t\n])")
+
+
+def _is_illegal_header_value(value: bytes) -> bool:
+    return bool(RE_IS_ILLEGAL_HEADER_VALUE.search(value))
 
 
 class _LockedObject(typing.Generic[T]):
@@ -52,9 +65,9 @@ class HTTP2Connection(HTTPSConnection):
     def __init__(
         self, host: str, port: int | None = None, **kwargs: typing.Any
     ) -> None:
-        self._h2_conn = self._new_h2_conn()
-        self._h2_stream: int | None = None
-        self._h2_headers: list[tuple[bytes, bytes]] = []
+        self.conn = self._new_h2_conn()
+        self.stream: int | None = None
+        self._headers: list[tuple[bytes, bytes]] = []
 
         if "proxy" in kwargs or "proxy_config" in kwargs:  # Defensive:
             raise NotImplementedError("Proxies aren't supported with HTTP/2")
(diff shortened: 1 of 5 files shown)


src/urllib3/http2.py
```diff
@@ -163,19 +235,60 @@ def getresponse(  # type: ignore[override]
...
+        if not self._has_header(b"user-agent"):
+            self.putheader(b"user-agent", _get_default_user_agent())
+
+        if body:
+            self.endheaders(message_body=body)
+            self.send(body)
+        else:
+            self.endheaders()
+
+    def _encode_request(self, request: str | bytes) -> bytes:
```
sethmlarson, line 274: Let's inline this to avoid more calls.

src/urllib3/http2.py
```diff
@@ -163,19 +235,60 @@ def getresponse(  # type: ignore[override]
...
+        *,
+        preload_content: bool = True,
+        decode_content: bool = True,
+        enforce_content_length: bool = True,
+        **kwargs: typing.Any,
+    ) -> None:
+        """Send an HTTP/2 request"""
+        if "chunked" in kwargs:
+            # TODO this is often present from upstream.
+            # raise NotImplementedError("`chunked` isn't supported with HTTP/2")
```
sethmlarson, line 253: We should maintain this error, since chunked encodings aren't valid in HTTP/2.

src/urllib3/http2.py
```diff
@@ -163,19 +235,60 @@ def getresponse(  # type: ignore[override]
...
+        if self.sock is not None:
+            self.sock.settimeout(self.timeout)
+
+        self.putrequest(method, url)
+
+        headers = headers or {}
+        for k, v in headers.items():
+            self.putheader(k, v)
+
+        if not self._has_header(b"user-agent"):
```
sethmlarson, line 265: Since we know `_headers` is all lowercase bytes headers, can we do `b"user-agent" not in self._headers`?

sethmlarson (commented):
Thanks for opening this PR! I have some high level comments for you :)

src/urllib3/http2.py
tcdent: I Left this commented out, since while the tests pass, if you run the patch methods in this module and make a call to `urllib3.request`, it will always fail. Still need to follow this up the stream.

```
from urllib3.http2 import inject_into_urllib3
inject_into_urllib3()
import urllib3
urllib3.request('GET', url="https://www.google.com/")
```

```
NotImplementedError: `chunked` isn't supported with HTTP/2
```

test/test_http2_connection.py
```diff
@@ -0,0 +1,178 @@
...
+from urllib3.connection import _get_default_user_agent
+from urllib3.http2 import (
+    HTTP2Connection,
+    _is_illegal_header_value,
+    _is_legal_header_name,
+)
+
+
+class TestHTTP2Connection:
+    def test__is_legal_header_name(self) -> None:
```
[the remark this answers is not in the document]
tcdent: This pattern comes from CPython, but the tests are based on some quick research to see what they are trying to prevent. Likely not exhaustive. 

Would prefer to defer to them, but it's a 1.1 library so it doesn't work with colon prefixes or enforce lower-case. 

https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L139

src/urllib3/http2.py
tcdent: Following CPython's lead here, again. Felt weird to leave a method behind that encoded ASCII.

It's only called by `putrequest`, so no benefit to breaking it out, either. Removed in my latest commit. 

https://github.com/python/cpython/blob/3.12/Lib/http/client.py#L1263

test/test_http2_connection.py
```diff
@@ -0,0 +1,178 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
tcdent: It feels like `PoolManager` needs to be made aware of `HTTP2Connection` before those tests can be written. Can you clarify if that's not the case?

test/test_http2_connection.py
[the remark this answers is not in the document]
tcdent: Dug a little deeper into the RFC [1] and implemented this according to the spec. Pretty confident this is complete. 

I tried to import the existing validation methods from `http` and extend them instead of creating them from scratch, but they're much more lenient than the actual spec. If we're saying we protect users from dangerous chars we should actually do that. 

[1] https://httpwg.org/specs/rfc9113.html#n-field-validity

src/urllib3/http2.py
sethmlarson: Sounds good, I created a new issue for tracking this: https://github.com/urllib3/urllib3/issues/3425

test/test_http2_connection.py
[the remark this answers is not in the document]
sethmlarson: That shouldn't be the case, the `http_version` pytest fixture injects HTTP2Connection into the library so HTTP/2 and HTTP/1.1 is tested side-by-side by running the test twice. Give it a try on one test that sends request bodies and if you run into issues we can add this in a future PR.

test/test_http2_connection.py
```diff
@@ -0,0 +1,327 @@
...
+        conn = HTTP2Connection("example.com")
+        with pytest.raises(ConnectionError):
+            conn.endheaders()
+
+    def test_send_ConnectionError(self) -> None:
+        conn = HTTP2Connection("example.com")
+        with pytest.raises(ConnectionError):
+            conn.send(b"foo")
+
+    def test_send_bytes(self) -> None:
```
sethmlarson, line 119: For these tests which mock `_h2_conn` internals, I think we should avoid doing this sort of testing since we don't "own" the internals of H2Connection. Instead I think we can get full coverage by making live requests (using the `http_version` fixture mentioned above) to be able to get test coverage without mocks.

You can keep all the tests that don't mock the h2 internals, those look great! :rocket:

sethmlarson (commented):
This is nearly ready to get merged :partying_face: Just one comment for you:

test/test_http2_connection.py
tcdent: AFAIK I'm not actually going down below the `h2` public API, just verifying that the method signatures we create are correct. `conn._h2_conn._obj` is our instance of `h2.connection.H2Connection`. These did get elaborate in the pursuit of 100% coverage. 

I get what you're saying about the `http_version` fixture now and use that in my latest commit.

sethmlarson: Oh gotcha! I forgot we maintained our own structure for that, thank you!

sethmlarson (approved):
LGTM, thank you so much for this!

sethmlarson:
So #3302 had a bounty of $500! If you're able to you can submit an expense to [our Open Collective](https://opencollective.com/urllib3) and I'll happily approve it, just be sure to include a link to this pull request. Let us know if you need help with this.

Issue #3400, opened 2024-06-05
NodeJS + pyodide support 


joemarshall:
### Context

Recently released Pyodide 0.26.0 supports webassembly Javascript Promise Integration. This means it is practical to do synchronous http requests in webassembly python using the standard (asynchronous) node-fetch api. 

This would make urllib3 work in webassembly on node.js supporting platforms such as Cloudflare workers. 

Once support for JSPI is released generally in browsers, it would be possible to unify the webassembly support to use this, which removes the need for web-workers for streaming http requests and would simplify the emscripten backend.

### Alternatives

It is possible to hack in node support using the node xmlhttprequest module, but the way that synchronous requests are handled is gross - it has to launch processes to handle them, and it is not a usable hack for server side python code.

### Duplicate

I don't think so. I think I mentioned it as future work when making the pyodide browser support.

### Contribution

I'm contracted to do this work already, and would like to upstream it once it is working.

sethmlarson:
Hey @joemarshall, nice to see you again! :) Happy to accept patches which makes urllib3 work better in those environments. If it's a big simplification (and based on your judgement) I'd be okay removing support for the more complicated backend if the simpler one is the way forwards in your mind. Let me know your thoughts on this, thanks again!

pquentin:
Done in https://github.com/urllib3/urllib3/pull/3427, closing, thank you!

PR #3427, opened 2024-07-10
Node.js / Webassembly Javascript Promise Integration support

joemarshall:
This PR adds support to the emscripten/pyodide backend for Javascript Promise Integration, a new feature in webassembly which allows webassembly code (i.e. python) to call asynchronous javascript apis like fetch. This means that on browsers where it is supported, urllib3 works in a much simpler way, directly with the browser fetch api. 

As a bonus, this means it is possible to use urllib3 in node.js, albeit with an experimental command line flag on the current stable version of node.

This is what I was talking about in #3400


sethmlarson:
@joemarshall Thank you for opening this PR, this is really exciting stuff! I'll give this a full review later. I think my high-level comment for now is to add more comments inline in the source code to explain what's happening and why. This will help us immensely for fixing issues in the future :pray:

joemarshall:
@sethmlarson I added tests + more commenting. The tests won't work within nox until I merge https://github.com/pyodide/pytest-pyodide/pull/140 but it is probably worth a code review at this point so that it can be mergable at the point that pull hits pytest-pyodide.

joemarshall:
@sethmlarson I pushed the change into pytest-pyodide and finished the ci here and it now hits 100% coverage and emscripten tests pass on node, chrome and firefox. It failed on something unrelated in Ubuntu this last run which I think might be a CI weirdness, but overall I think is ready for review now.

illia-v (commented):
@joemarshall thank you for your work on this! I'm looking forward to having this feature integrated.

I noticed that the PR includes adjustments due to urllib3 being built in Pyodide and significant changes to the testing code. Would it be feasible to split the PR into a few smaller ones to facilitate a thorough review and faster integration?

Perhaps one PR could focus on changing the documentation and tests to use the built-in urllib3, another to replace webdriver-manager with Selenium, another on installing Node.js and testing non-JSPI code failing in Node.js, and a final one for the JSPI code itself.

sethmlarson:
Thanks for the review @illia-v! @joemarshall are you still interested in landing this PR?

docs/reference/contrib/emscripten.rst
```diff
@@ -79,9 +76,14 @@ Features which don't work with Emscripten:
...
 you must be running the code within a Web Worker and set specific HTTP headers
 for the serving website to enable `Cross-Origin Isolation <https://developer.mozilla.org/en-US/docs/Web/API/crossOriginIsolated>`_.
 
 You can verify whether a given environment is cross-origin isolated by evaluating the global ``crossOriginIsolated`` JavaScript property.
+
+Node.js support
+---------------
+Node.js support uses a relatively new feature in webassembly known as Javascript Promise Integration. 
+To use urllib3 in Node.js, you are likely to need to call node.js with the ``--experimental-wasm-stack-switching`` 
+command line parameter.
```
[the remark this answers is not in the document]
hoodmane: Would also be good to use the standard capitalization of "WebAssembly".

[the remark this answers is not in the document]
hoodmane: Maybe mention here also that it requires Node >= 20.

dummyserver/app.py
```diff
@@ -353,29 +371,69 @@ async def pyodide_upload() -> ResponseReturnValue:
...
+    wheel_folder = Path(__file__).parent.parent / "dist"
+    wheels = list(wheel_folder.glob("*.whl"))
+    wheels = sorted(wheels, key=lambda w: w.stat().st_mtime)
+    if len(wheels) > 0:
+        return wheels[-1]  # newest wheel
+    else:
+        return None
+
+
+def _rewrite_pyodide_lock() -> bytes | None:
```
hoodmane, line 384: What is rewriting the lockfile needed for? In `emscripten/conftest.py`, you run `loadPackage` the `urlib3.whl` directly, which I'd think would be good enough to make sure you're getting the right version. Is there some place that installs it by name?

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -403,6 +534,145 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
...
+
+        try:
+            from pyodide.ffi import can_run_sync
+        except ImportError:
+            from pyodide_js._module import (  # type: ignore[import-not-found]
+                validSuspender,
+            )
+
+            def can_run_sync() -> bool:
+                return bool(validSuspender.value)
```
hoodmane, line 669: I think it would make sense to just drop this polyfill now and only support JSPI on Pyodide >= 0.26.2.
```suggestion
        from pyodide.ffi import can_run_sync
```

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -403,6 +534,145 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
...
+
+        return bool(can_run_sync())
+    except BaseException:
```
hoodmane, line 672: I'd think we would only want to catch `Exception` here, wouldn't want to catch a `KeyboardInterrupt` or `SystemExit`? Or maybe even just catch `ImportError`?
```suggestion
    except Exception:
```

test/contrib/emscripten/conftest.py
```diff
@@ -90,44 +141,81 @@ def _install_coverage(self: Any) -> None:
...
+            textwrap.dedent(
+                f"""
+                import pyodide_js as _pjs
+                await _pjs.loadPackage('https://{self.host}:{self.port}/dist/urllib3.whl')
+                """
+            )
+            + code
+        )
+
+        if has_jspi is False:
```
hoodmane, line 165:
```suggestion
        if not has_jspi:
```

test/contrib/emscripten/conftest.py
```diff
@@ -90,44 +141,81 @@ def _install_coverage(self: Any) -> None:
...
+            # disable jspi in this code
```
hoodmane, line 166: Why are we monkey patching `has_jspi()` here? Shouldn't `has_jspi()` correctly detect this? In real life if people are using Firefox or Safari or some other runtime without JSPI, they won't get this monkey patch.

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -322,6 +432,17 @@ def is_worker_available() -> bool:
 
 
 def send_streaming_request(request: EmscriptenRequest) -> EmscriptenResponse | None:
+    if has_jspi():
+        return send_jspi_request(request, True)
+    elif is_in_node():
+        raise _RequestError(
+            message="urllib3 only works in node.js with pyodide.runPythonAsync"
+            " and requires the flag --experimental-wasm-stack-switching in "
+            " versions of node <24",
```
hoodmane, line 441: We may want to mention here that node >= 20 is required. We're not sure that node 24 won't require the flag yet...

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -403,6 +534,145 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
...
+        from pyodide.ffi import run_sync
+
+        # run_sync here uses WebAssembly Javascript Promise Integration to
+        # suspend python until the Javascript promise resolves.
+        return run_sync(promise)
+    except JsException as err:
+        if err.name == "AbortError":
+            raise _TimeoutError(
+                message="Request timed out", request=request, response=response
+            )
```
hoodmane, line 643: Nice!

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -403,6 +534,145 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
...
+    else:
+        # get directly via arraybuffer
+        # n.b. this is another async Javascript call.
+        body = _run_sync_with_timeout(
+            response_js.arrayBuffer(),
+            timeout,
+            js_abort_controller,
+            request=request,
+            response=response,
+        ).to_py()
```
hoodmane, line 599: The `arrayBuffer()` call isn't cancelable so I think we want to just use `run_sync` directly here.
```suggestion
        body = run_sync(response_js.arrayBuffer()).to_py()
```

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -288,6 +300,104 @@ def send(self, request: EmscriptenRequest) -> EmscriptenResponse:
...
+    def close(self) -> None:
+        if not self.is_closed():
+            self.read_len = 0
+            self.read_pos = 0
+            self.js_read_stream = None
+            self._is_closed = True
+            self._is_done = True
+            self.request = None
+            self.response = None
+            super().close()
```
hoodmane, line 357: We should probably call `self.js_read_stream.cancel()` when we're done with it. Also a nit: it reads a bit nicer with an early return.
```suggestion
    def close(self) -> None:
        if self.is_closed():
            return
        self._is_closed = True
        self._is_done = True
        self.read_len = 0
        self.read_pos = 0
        self.js_read_stream.cancel()
        self.js_read_stream = None
        self.request = None
        self.response = None
        super().close()
```

src/urllib3/contrib/emscripten/fetch.py
joemarshall: https://developer.mozilla.org/en-US/docs/Web/API/Response/arrayBuffer says it can return AbortError, so I figure at least mozilla think that it can be cancelled between header and response fetch, although I haven't delved into browser code.

src/urllib3/contrib/emscripten/fetch.py
joemarshall: done

test/contrib/emscripten/conftest.py
joemarshall: It was just to save loading different browsers and force the jspi code to be executed. I simplified the code relating to this now, so it uses the non-jspi selenium instead.

joemarshall:
@sethmlarson @illia-v I fixed various issues from the reviews above and simplified this change somewhat. I'm not 100% sure how/whether to split it up into multiple PRs because everything is slightly entangled. Any comments would be welcome.

.github/workflows/ci.yml
```diff
@@ -108,10 +108,27 @@ jobs:
 
       - name: "Install Chrome"
         uses: browser-actions/setup-chrome@db1b524c26f20a8d1a10f7fc385c92387e2d0477 # v1.7.1
+        id: setup-chrome
+        if: ${{ matrix.nox-session == 'emscripten' }}
+        with:
+          chrome-version: 120
```
hoodmane, line 114: What version of chrome was in there before the version override? Presumably an old one? We might want to check if it works against chrome 126 since there are some jspi changes there.

emscripten-requirements.txt
```diff
@@ -1,3 +1,3 @@
-pytest-pyodide==0.54.0
-pyodide-build==0.24.1
-webdriver-manager==4.0.1
\ No newline at end of file
+pytest-pyodide==0.58.2
+pyodide-build==0.26.2
```
hoodmane, line 2: With the pyodide-build unvendoring, do we need a pin here @ryanking13? Should it pin latest?

src/urllib3/contrib/emscripten/fetch.py
hoodmane: Cool. Indeed if you poke around the fetch spec long enough you can find that the spec says this should happen:
> To abort a fetch() call with a promise, request, responseObject, and an error:
> ...
> 5. If response’s [body](https://fetch.spec.whatwg.org/#concept-response-body) is non-null and is [readable](https://streams.spec.whatwg.org/#readablestream-readable), then [error](https://streams.spec.whatwg.org/#readablestream-error) response’s [body](https://fetch.spec.whatwg.org/#concept-response-body) with error.


https://fetch.spec.whatwg.org/#abort-fetch

test/contrib/emscripten/conftest.py
```diff
@@ -90,44 +127,97 @@ def _install_coverage(self: Any) -> None:
...
-                )
-                + code
-            )
-            code += textwrap.dedent(
+        # replace imports of urllib3 in the code
+        # so that pyodide doesn't find them and automatically
+        # load the urllib3 module from its distribution
+        # n.b. we can't override this import otherwise because
+        # we can't run custom js in the pyodide worker test template
+        # before the python code is launched
```
hoodmane, line 145: It might be easier to replace `webworker.js`. It's only 26 lines and quite simple. But this seems fine if a bit regrettable.

test/contrib/emscripten/test_emscripten.py
```diff
@@ -18,36 +18,29 @@
...
-    ]
-):
-    return copy_files_to_pyodide(  # type: ignore[no-any-return]
-        file_list=[("dist/*.whl", "/tmp")], install_wheels=True
+if hasattr(pytest_pyodide.runner, "CHROME_FLAGS"):
+    pytest_pyodide.runner.CHROME_FLAGS.append("ignore-certificate-errors")
+else:
+    pyodide_config = pytest_pyodide.config.get_global_config()
+    pyodide_config.set_flags(
+        "chrome", ["ignore-certificate-errors"] + pyodide_config.get_flags("chrome")
```
hoodmane, line 30: Is this intended to support multiple versions of `pytest-pyodide`? Seems like you might as well pin the new version and drop the old code.

test/contrib/emscripten/test_emscripten.py
```diff
@@ -682,11 +686,10 @@ def test_streaming_not_ready_in_browser(
     )
 
 
-@install_urllib3_wheel()
 def test_requests_with_micropip(
     selenium_coverage: typing.Any, testserver_http: PyodideServerInfo
 ) -> None:
-    # this can't be @run_in_pyodide because of the async code
+    # this can't be @run_in_pyodide(packages=['urllib3']) because of the async code
```
hoodmane, line 692: Unrelated to this PR, but it should work just fine to do:
```py
@run_in_pyodide
async def test_something(...):
    await some_async_call()
```
See for instace: https://github.com/pyodide/pyodide/blob/main/src/tests/test_browser_apis.py

hoodmane (approved):
Looks good to me. I made some more very minor comments but none of them need to be addressed necessarily in my opinion.

On code size, it is true that it's almost a 1000 line diff, but I don't find it too hard to review. 80% of the changes are to `contrib/emscripten` and the ones that aren't are to ci, noxfile, dummyserver, docs stuff that doesn't go at runtime. So I don't personally see a good reason to split it up, but of course we'll have to see what the maintainers want.

Thanks @joemarshall!

test/contrib/emscripten/conftest.py
```diff
@@ -158,27 +248,110 @@ def run_webworker(self, code: str) -> Any:
...
-    # install the wheel, which is served at /wheel/*
-    selenium_coverage.run_js(
-        """
-await pyodide.loadPackage('/wheel/dist.whl')
-"""
-    )
-    selenium_coverage._install_coverage()
+    if selenium_coverage.browser != "node":
+        # on node, we don't need to be on the same origin
+        # so we can ignore all this
```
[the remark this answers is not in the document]
joemarshall: There is a way to disable it but it is deep in the firefox profile meaning it would require code in pytest-pyodide or monkeypatching; making it be on the same origin is easier.

emscripten-requirements.txt
ryanking13: You don't need to set the pyodide-build version equal to the target Pyodide version anymore, but I think it is not a bad thing to pin the version for reproducible build.

joemarshall: Actually, thinking about it, I don't need pyodide-build any more, not sure that should be in really. Will fix.

test/contrib/emscripten/conftest.py
joemarshall: yeah, I did that now.

test/contrib/emscripten/test_emscripten.py
joemarshall: yeah, I just didn't want to have to add an asyncio test plugin just for this single test

.github/workflows/ci.yml
joemarshall: It will only work against 126 once pyodide 0.27 is properly released, hence I stuck it back on 120 for testing for now, because getting the latest version of chrome breaks everything. I figure it doesn't matter, because the next version of pyodide that includes an upgraded urllib3 will also have the fix for jspi in pyodide (the feature detect thing)

hoodmane:
@sethmlarson would you mind reviewing this PR when you have a chance? Or is there another maintainer that could do the review?

.github/workflows/ci.yml
```diff
@@ -108,10 +108,32 @@ jobs:
...
+      - name: "Install node.js"
+        uses: actions/setup-node@v4
+        if: ${{ matrix.nox-session == 'emscripten' }}
+        with:
+          node-version: 22
+      - name: Cache pyodide downloads in nox cache
+        uses: actions/cache@v4
+        with:
+          path: .nox/.cache
+          key: pyodide_downloads
```
illia-v, line 136: I believe the key has to include a hash of `noxfile.py` which contains a Pyodide version.
If the key never changes, the following can happen: when Pyodide version is changed, files of the new version won't be added to the cache until the cache is evicted. And the eviction may never happen if the cache is restored frequently enough.

```suggestion
          # noxfile.py contains a Pyodide version used currently.
          key: pyodide-downloads-${{ hashFiles('noxfile.py') }}
```

docs/reference/contrib/emscripten.rst
```diff
@@ -28,16 +28,16 @@ Using urllib3 with Pyodide means you need to `get started with Pyodide first <ht
 The Pyodide project provides a `useful online REPL <https://pyodide.org/en/stable/console.html>`_ to try in your browser without
 any setup or installation to test out the code examples below.
 
+One minor note - when running pyodide code from javascript, if you use ``pyodide.runPythonAsync`` rather
```
illia-v, line 31:
```suggestion
One minor note - when running Pyodide code from JavaScript, if you use ``pyodide.runPythonAsync`` rather
```

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -403,6 +544,153 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
...
+        # suspend python until the JavaScript promise resolves.
+        return run_sync(promise)
+    except JsException as err:
+        if err.name == "AbortError":
+            raise _TimeoutError(
+                message="Request timed out", request=request, response=response
+            )
+        else:
+            raise _RequestError(message=err.message, request=request, response=response)
+        # raise (something)
```
illia-v, line 669: Is this a TODO?

illia-v (commented):
@hoodmane thank you for you review!

.github/workflows/ci.yml
illia-v: Also, this step should be limited to the `emscripten` session

.github/workflows/ci.yml
joemarshall: fixed with 0.26.3 I think

joemarshall:
@illia-v Thanks for the review, I think I've addressed all those comments. CI passes okay.

illia-v (approved):
@joemarshall thanks again!

FYI, Firefox and Chrome tests have been split in separate jobs in the main branch to get faster CI (see #3505), so I made the Node.js job separate while I was resolving conflicts in the last merge commit

PR #3429, opened 2024-07-12
raise if chunked encoding is demanded with http2

zawan-ila:
Relates to https://github.com/urllib3/urllib3/issues/3425


zawan-ila:
@sethmlarson Could you please have a look here?

sethmlarson:
@zawan-ila This PR doesn't match with the issue that you've linked to, I think we should silently ignore the `chunked` parameter and `Transfer-Encoding: chunked` header and send the body without the chunked framing. You'll need to add tests as well for this behavior.

PR #3435, opened 2024-08-01
[1.26] util/ssl: make code resilient to missing hash functions

shashankram:
Backport of https://github.com/urllib3/urllib3/pull/3434
---
In certain environments such as in a FIPS enabled system, certain algorithms such as md5 may be unavailable. Due to the importing of such a module on a system where it is unavailable, urllib will crash and is unusable.

This change makes the code more resilient by deferring the usage of the hash function module unless they are required. If the hash function is unavailable but required, an SSLError is raised indicating the same. If the hash function is unavailable but not required, urllib functions instead of crashing.

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -1,11 +1,11 @@
 from __future__ import absolute_import
 
+import hashlib
 import hmac
 import os
 import sys
 import warnings
 from binascii import hexlify, unhexlify
-from hashlib import md5, sha1, sha256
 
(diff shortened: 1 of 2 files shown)


illia-v (approved):
LGTM, thanks!

shashankram:
@illia-v are the failing runs flakes I can ignore?

illia-v:
> @illia-v are the failing runs flakes I can ignore?

Yes, sorry about them, I'm working on making the tests more stable

PR #3437, opened 2024-08-03
[1.26] Upgrade actions

illia-v:
CI runs in 1.26.x produce many warnings because we use old versions of various actions, this PR updates the actions to the latest versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,9 @@ jobs:
 
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks! LGTM.

Should we consider pinning the actions with commits as in the main branch?

illia-v:
> Should we consider pinning the actions with commits as in the main branch?

Yes, I can create a separate PR

Also, we have to revise the publish workflow for 1.26.x because we didn't update it after the last failed run

PR #3438, opened 2024-08-03
Fix CI after attrs 24.1.0 release

pquentin:
See https://github.com/python-trio/trio/issues/3053 for details. Another option would be to ignore the warning, but we don't use attrs ourselves so don't really care about the version and I'm hoping that trio will quickly fix the issue.


illia-v (approved):
This is something new, thanks for the PR!

CoolCat467:
https://github.com/python-trio/trio/pull/3054 is going to fix this and the version of attrs will no longer need to be specified after a release is made.

CoolCat467:
Fix was released in https://github.com/python-trio/trio/releases/tag/v0.26.1

pquentin:
Thanks, I opened https://github.com/urllib3/urllib3/pull/3440 to remove the attrs pin.

PR #3440, opened 2024-08-05
Remove now unneeded attrs pin

pquentin:
Following the fix in Trio and the subsequent release: https://github.com/python-trio/trio/pull/3054.


pquentin:
Turns out it's still needed, I reported it in https://github.com/python-trio/trio/issues/3053#issuecomment-2268730515.

sethmlarson (approved):

PR #3446, opened 2024-08-10
Fix request and request_encode_url fields docs

angusholder:
Looks like it was a copy/paste mistake when request_encode_body/request_encode_url were documented


pquentin (changes requested):
Good catch, thank you! Would you mind also updating the fields documentation in the request() function?

```
Data to encode and send in the URL or request body, depending on ``method``.
```

That way, the details are in the `request_encode_url` and `request_encode_body` functions, where they're more likely to stay correct.

angusholder:
Okay, I've done `request` now too!

pquentin (approved):
Thanks! LGTM.

PR #3450, opened 2024-08-15
Rely on OpenSSL ALPN support

pquentin:
OpenSSL 1.1.1+ always supports ALPN, but does not support NPN. The Python NPN functions are no-ops with OpenSSL 1.1.1+.

Source: https://github.com/python/cpython/pull/23014

### DIFF
--- a/src/urllib3/util/ssltransport.py
+++ b/src/urllib3/util/ssltransport.py
@@ -193,9 +193,6 @@ class SSLTransport:
     def selected_alpn_protocol(self) -> str | None:
         return self.sslobj.selected_alpn_protocol()
 
-    def selected_npn_protocol(self) -> str | None:
-        return self.sslobj.selected_npn_protocol()
-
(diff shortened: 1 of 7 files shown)


src/urllib3/util/ssltransport.py
```diff
@@ -193,9 +193,6 @@ def cipher(self) -> tuple[str, str, int] | None:
     def selected_alpn_protocol(self) -> str | None:
         return self.sslobj.selected_alpn_protocol()
 
-    def selected_npn_protocol(self) -> str | None:
-        return self.sslobj.selected_npn_protocol()
-
```
pquentin, line 198: You could argue that this is a breaking change, but we don't document SSLTransport so I think that's fine.

pquentin:
Is this worthy of a changelog entry?

sethmlarson (approved):
LGTM, not worth a changelog entry IMO

PR #3451, opened 2024-08-15
Prevent enabling HTTP/2 in HTTP tests

pquentin:
Today, enabling HTTP/2 only affects HTTPS. Allowing `http_version` in HTTP tests gives a false sense of test coverage.

@tcdent Sorry that we did not realize this earlier.


sethmlarson (approved):
Excellent, thanks Quentin! :rocket:

Issue #3323, opened 2024-01-26
Slow test cases on pypy3.9-7.3.15 on Ubuntu 22.04 


ecerulm:
### Subject

Currently the urllib3's test suite on pypy3.9-7.3.14 and pypy3.9-7.3.15 runs unbearably slow, takes more that 1h45m. 

This is the reason why #3308 pypy3.9-7.3.13 is currently pinned in the GitHub Actions CI. 

The issue has been reported upstream to pypy https://github.com/pypy/pypy/issues/4877 and in coverage https://github.com/nedbat/coveragepy/issues/1738. But after some investigatio it's clear the issue is at the pypy side since the running the test without coverage is still unbearably slow. 

The purpose of this issue is to keep some tracking at the urllib3 level even when the solution (IMHO) should come from pypy.

### Environment

Ubuntu 22.04 (on VirtualBox)
pypy3.9-7.3.14 (from pyenv)

### Steps to Reproduce

```shell
#zsh
pyenv install pypy3.9-7.3.14
path=($(pyenv prefix pypy3.9-7.3.14)/bin $path)
nox -s test-pypy3.9 # takes hours
```

It willl take more that 1h45m to get to 80%, each test is slower that the previous one so it's possible that it will take several hours to complete. 

Memory comsumption until 80%  seems stable .  

### Expected Behavior

The test suite should run under 3 minutes like it does when pypy3.9-7.3.13 is used. 

```shell
#zsh
pyenv install pypy3.9-7.3.13
path=($(pyenv prefix pypy3.9-7.3.13)/bin $path)
nox -s test-pypy3.9 # takes 3 in my environment
```

### Actual Behavior

It takes 1h45m to get to 80% test completion 

Each test takes longer that the previous.  After 1h45m each test can take several minutes to complete. 

If I run any of those tests that take several minutes to complete in isolation on the same environment it finishes instantaneously. 

 

# Related

https://github.com/urllib3/urllib3/pull/3309
#3308

mattip:
We just released pypy3.10-v7.3.17 with some improvements in performance, including packaging OpenSSL 3.3.1. Is it any faster in your CI?

ecerulm:
The CI uses  7.3.16  , see https://github.com/urllib3/urllib3/pull/3452

```
 Successfully set up PyPy 7.3.16 with Python (3.10.14)
```

[...]

PR #3308, opened 2024-01-23
Pin to pypy-3.9-v7.3.13 in CI

ecerulm:
CI seems to be continuosly failing on the `test-pypy` under `pypy3.9`. 

Currently `pypy3.9` is mapped to `pypy3.9-v7.3.15` and I believe this particular version of pypy is causing the "coverage" take way longer. 

In this PR I'm pinning the version to `pypy3.9-v7.3.13`, this I think makes the `test-pypy` to complete in under 8 minutes instead of the current +30 minutes which causes the CI job to be cancelled.

Closes #3305

PR #3452, opened 2024-08-16
Unpin pypy version in CI

ecerulm:
https://github.com/pypy/pypy/issues/4877#issuecomment-2293763561

I recently tested main on pypy3.9-7.3.14 and it's not as slow as before https://github.com/urllib3/urllib3/pull/3308  and #3323


pquentin (commented):
Excellent! Thank you. LGTM.

pquentin (approved):

Issue #3454, opened 2024-08-17
optimization request: _encode_invalid_chars is slow


holmanb:
### Context

While benchmarking cloud-init's unit tests, I notice that the largest amount of time is spent, surprisingly, in urllib3's `url.py::_encode_invalid_chars()`.
```
   Ordered by: internal time
   List reduced from 10107 to 3032 due to restriction <0.3>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   236638    5.866    0.000    9.313    0.000 url.py:227(_encode_invalid_chars)
```

### Duplicate

Has the feature been requested before?

I searched and didn't see any requests to optimize this function.

### Contribution

Would you be willing to submit a PR?

I put together a prototype of an idea that passes the unittests and cuts over half of a second off urllib3's unittests for me. It complicates the function a bit, but I'll put up a PR and we can discuss if it is a desirable addition.

_(Help can be provided if you need assistance submitting a PR)_

pquentin:
I'd like to understand cloud-init's use case a bit more. How did you end up with 236k calls to this function? Is this over a long period of time?

sethmlarson:
Closing due to lack of response.

PR #3455, opened 2024-08-17
Optimize _encode_invalid_chars

holmanb:
See https://github.com/urllib3/urllib3/issues/3454 for motivation. I haven't benchmarked it in depth, but presumably this should benefit all urllib3 users - it looks like this cuts around half of a second from urllib3's unit tests.

I've included context for what this accomplishes in the commit message and comments, feel free to let me know if you have any questions or concerns. The basic idea is to eliminate unnecessary allocations by preallocating the bytearray.

Note: I ran out of time to submit a proper PR (sorry). I'll gather some concrete numbers to back up this claim next week.

### DIFF
--- a/changelog/3455.feature.rst
+++ b/changelog/3455.feature.rst
@@ -1,0 +1,1 @@
+Optimized _encode_invalid_chars for faster execution.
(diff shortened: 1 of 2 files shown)


pquentin:
Thanks for the pull request. This is indeed complex, with the opportunity for off-by-one errors in a sensible part of the code (we had a few CVEs around parsing URLs). I would appreciate if you could quantify the gain a bit more precisely. Half a second when running the full test suite is not that much.

holmanb:
After benchmarking this I don't think that this code is a meaningful performance improvement, so I'm going to close this given the complexity of this solution.

PR #3457, opened 2024-08-19
Use pytest-memray limit_memory current_thread_only setting

ecerulm:
Some time ago I requested this functionality to the pytest-memray team https://github.com/bloomberg/pytest-memray/issues/111,

Now that functionality is available since `pytest-memray==1.6.0`.

Enabling `current_thread_only` will give less false positives, as memory in the server threads (which is not part of  the SUT, in this case urllib3) will not be reported.

The need for this setting comes from #3358 where we had to modify a bunch of test cases to properly clean up resources due to false positives.


pquentin (approved):
Thanks for chasing this and following up! LGTM.

emscripten failed (but the check is green!) which explains the coverage issue.

PR #3458, opened 2024-08-23
Add license identifier to project metadata

mschoettle:
License scanning tools (such as the one used by GitLab) rely on the project metadata (made available via the PyPi API) to detect the license for a package.

Currently, `license: null` is returned.

This PR adds the license to the project metadata.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ keywords = ["urllib", "httplib", "threadsafe", "filepost", "http", "https", "ssl
 authors = [
   {name = "Andrey Petrov", email = "andrey.petrov@shazow.net"}
 ]
+license = { text = "MIT" }
 maintainers = [
   {name = "Seth Michael Larson", email="sethmichaellarson@gmail.com"},
   {name = "Quentin Pradet", email="quentin@pradet.me"},


pyproject.toml
```diff
@@ -12,6 +12,7 @@ keywords = ["urllib", "httplib", "threadsafe", "filepost", "http", "https", "ssl
 authors = [
   {name = "Andrey Petrov", email = "andrey.petrov@shazow.net"}
 ]
+license = { text = "MIT" }
```
illia-v, line 15: Thank you for the contribution!

[Packaging docs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license) recommend using a classifier instead of this attribute and we already provide it
> If you are using a standard, well-known license, it is not necessary to use this field. Instead, you should use one of the [classifiers](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#classifiers) starting with `License ::`. (As a general rule, it is a good idea to use a standard, well-known license, both to avoid confusion and because some organizations avoid software whose license is unapproved.)

It's weird that it's not taken into account by the PyPI API, but I'm not against adding this attribute. Will `text = "MIT License"` (as in the docs example and in the existing classifier) work the same as `text = "MIT"`?

mschoettle: Interesting. I came across an issue and follow-up discussion with the intention to make license classifiers more fine-grained (https://github.com/pypi/warehouse/issues/15901) and it seems it turned into a PEP to improve license clarity (https://discuss.python.org/t/pep-639-round-3-improving-license-clarity-with-better-package-metadata/53020).

> Will text = "MIT License" (as in the docs example and in the existing classifier) work the same as text = "MIT"?

I don't think so. Based on my understanding and one case where `license = LGPLv3+` was not detected by the scanning tool it seems that it should be a valid SPDX identifier. In this case that's `MIT` (https://spdx.org/licenses/).

pquentin: I also agree that if we put something here it should be the SPDX identifier.

pquentin: (And to be clear I would also prefer for the PEP to be finalized before adding this.)

pquentin:
Closing until PEP 639 is adopted. Thank you!

Issue #3285, opened 2024-01-17
Create a property for `proxy_is_tunnel`


ghost:
> @abebeos Can you open an issue for creating a property for `proxy_is_tunnel`?

_Originally posted by @sethmlarson in https://github.com/urllib3/urllib3/pull/3283#discussion_r1455851350_

* Maybe `proxy_is_tunneling`, similar to `proxy_is_forwarding`

PR #3459, opened 2024-08-23
Add HTTPSConnection.proxy_is_tunneling

Mr-Sunglasses:
fix: #3285


src/urllib3/connection.py
```diff
@@ -709,7 +717,7 @@ def connect(self) -> None:
 
                 self._tunnel()
                 # Override the host with the one we're requesting data from.
-                server_hostname = self._tunnel_host
+                server_hostname = self._tunnel_host  # type: ignore[assignment]
```
illia-v, line 720: What do you think about casting `self._tunnel_host` to `str` instead of ignoring the type?

```suggestion
                server_hostname = typing.cast(str, self._tunnel_host)
```

illia-v (changes requested):
Please check the suggestion and add [a changelog entry](https://github.com/urllib3/urllib3/tree/main/changelog#readme)

illia-v:
@Mr-Sunglasses thanks for accepting the suggestion, please add a changelog entry

Mr-Sunglasses:
@illia-v I've added the changelog, LMK if anything more to do with this PR.

And thanks a lot for reviewing this PR 🙏,

illia-v (approved):

PR #3462, opened 2024-08-27
[1.26] Backport changes to the publish workflow

illia-v:
Our publish workflow had [failed](https://github.com/urllib3/urllib3/actions/runs/9549560548) last 1.26.19 release because of outdated SLSA generator, we updated the generator in the main branch after that, and we need to backport the change

Here I'm just copying `.github/workflows/publish.yml` from the main branch with one change: disabling publishing to Test PyPI on pushes to 1.26.x
https://github.com/urllib3/urllib3/blob/3da3eb4dc09d58aa63aae4b7d6bdb9ee403a11d1/.github/workflows/publish.yml#L6-L8

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,29 +2,32 @@ name: Publish to PyPI
 
 on:
   push:
+    branches:
+      # We don't publish to Test PyPI on pushes to 1.26.x yet because there is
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
LGTM, thanks @illia-v! :purple_heart:

PR #3466, opened 2024-09-02
Revert "Add TLS settings for HTTP/2"

### LINKED ISSUE #3306: Comply with TLS settings mandated for HTTP/2 in RFC 9113
[RFC 9113](https://www.rfc-editor.org/rfc/rfc9113.html) imposes a number of conditions on TLS settings:

 * Section 9.2: mandates TLS 1.2 or above (can be done with `ssl_minimum_version`)
 * Section 9.2.1/9.2.2: mandates one specific TLS 1.2 cipher suite (TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256) with a specific ECDH curve (though I think the curve choice is specific to servers) and forbids others. http://python-hyper.org/projects/h2/en/stable/negotiating-http2.html lists all allowed cipher suites in practice, but it references RFC 7540 and it's possible allowed cipher suites have changed between the two RFCs.
 * Section 9.2.1: Renegotiation must be disabled (using `ssl.OP_NO_RENEGOTIATION`), meaning at some point a connection will become unusable, at which point we will have to discard it.
 * Section 9.2.1/10.6: forbids the use of TLS compression (`ssl.OP_NO_COMPRESSION` must be set)
 * Section 9.2.3: HTTP/2 clients MUST treat a TLS post-handshake CertificateRequest message as a [connection error](https://www.rfc-editor.org/rfc/rfc9113.html#ConnectionErrorHandler) ([Section 5.4.1](https://www.rfc-editor.org/rfc/rfc9113.html#ConnectionErrorHandler)) of type [PROTOCOL_ERROR](https://www.rfc-editor.org/rfc/rfc9113.html#PROTOCOL_ERROR). Maybe hyper-h2 does this for us? We probably still want to enable post-handshake authentication in case the server is HTTP/1.1.

Most of these will have to be encoded in `create_urllib3_context`, which will need an HTTP/2 mode. It has to be part of the public API to allow users to tweak the context further if they want to, for example to handle a non-conformant server. `HTTP2Connection` will thus need to accept an `ssl_context` parameter at this point.

pquentin:
Reverts urllib3/urllib3#3456, Closes #3306 

Unfortunately, we realized after working on them that we can't really use those new settings. curl also does not implement them. See https://github.com/urllib3/urllib3/issues/3306#issuecomment-2320997012 for details.


illia-v (approved):
Thanks!

PR #3474, opened 2024-09-08
GitHub Actions: Only attempt to publish for upstream

hugovk:
It unsurprisingly fails for forks, so skip it in the same way as `scorecards.yml`:

https://github.com/hugovk/urllib3/actions/runs/10758514893/job/29834070711

(This could also be more tightly scoped to just those steps needing upload permissions, if you also want to test the other steps on forks.)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,6 +15,7 @@ permissions:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Good call, thank you!

PR #3476, opened 2024-09-08
HTTP/2: Add support for multiple streams and response streaming API

### LINKED ISSUE #3294: Move response body handling into HTTP2Response
Currently response handling occurs within the connection itself, but once we have response headers we should be handing off the body processing to the HTTPResponse object. This includes support for decoding gzip and other compressed bodies.

### LINKED ISSUE #3300: Add support for streaming HTTP/2 responses
Support `.read()` and `.stream()` APIs on HTTP2Response

### LINKED ISSUE #3301: Allow HTTP/2 socket to be re-used for future requests

mrdaybird:
Fixes #3300, Fixes #3294 Fixes #3301 

Firstly, this PR is not complete, still I have put it up to get the discussion started. 

This PR tries to achieve the following:
1. Allow multiple streams per `HTTP2Connection`. (solves #3301)
2. Add support for `read` and `stream`. #3300
3. Decoding the response data. #3294 

I have merged these things together into because they are kind of interdependent and gives an idea about the whole picture, like how the API will look like.

Changes proposed by this PR ⇒
1. Adds one new object called `HTTP2Stream`, which manages the state of a stream. (similar to idea begin by @pquentin in #3294)
2. Since a `HTTP2Connection` is now associated with multiple streams, the methods `putrequest`, `endheaders`, `send` and `getresponse` only have meaning in context of a stream. So, I suggest they should not be part of public API. Thus,
    1. `putrequest` is changed to `_putrequest` + a new `stream_id` parameter
    2. `endheaders` is changed to `_endheaders` + a new `stream_id` parameter
    3. `send` is changed to `_send_stream` + a new `stream_id` parameter
    4. `getresponse` is removed, and moved to `HTTP2Stream`
    5. a new `_read_stream` that reads from socket and handles h2 events
 
To give you a little sneak peek, here is a working example of sending two request using single `HTTP2Connection`(HTTP/2 is awesome!) ⇒

```
from urllib3.http2 import inject_into_urllib3
from urllib3.http2.connection import HTTP2Connection

inject_into_urllib3()

conn = HTTP2Connection('www.nghttp2.org', 443)
conn.connect()

stream1 = conn.request("GET", "/httpbin/json")
stream2 = conn.request("GET", "/httpbin/gzip")

res1 = stream1.getresponse()
res2 = stream2.getresponse()

print(res1.data)
print(res2.data) # it also decompresses the gzip!
```

Important thing to note is that 1. `request` returns a `HTTP2Stream` and 2. `getresponse` needs to be called on the `HTTP2Stream` to get a `HTTP2Response`

TODO:
- [x] Support multiple streams
- [x] Add`read` method
- [ ] Add `stream` method
- [ ] Add tests for all of the above
- [ ] Fix old tests

@pquentin @sethmlarson 

Since there are several substantial changes, therefore before moving on with this PR, I wanted to know your opinion of it! 
If you want my explanation for any design/code decisions, please ask!

### DIFF
--- a/src/urllib3/http2/__init__.py
+++ b/src/urllib3/http2/__init__.py
@@ -32,7 +32,7 @@ def inject_into_urllib3() -> None:
     global orig_HTTPSConnection
     orig_HTTPSConnection = urllib3_connection.HTTPSConnection
 
-    HTTPSConnectionPool.ConnectionCls = HTTP2Connection
+    HTTPSConnectionPool.ConnectionCls = HTTP2Connection  # type: ignore[assignment]
     urllib3_connection.HTTPSConnection = HTTP2Connection  # type: ignore[misc]
 
     # TODO: Offer 'http/1.1' as well, but for testing purposes this is handy.
--- a/src/urllib3/http2/connection.py
+++ b/src/urllib3/http2/connection.py
@@ -5,6 +5,8 @@ import re
 import threading
 import types
 import typing
+from enum import Enum
+from io import BytesIO
 
 import h2.config  # type: ignore[import-untyped]
 import h2.connection  # type: ignore[import-untyped]
@@ -14,7 +16,7 @@ from .._base_connection import _TYPE_BODY
 from .._collections import HTTPHeaderDict
 from ..connection import HTTPSConnection, _get_default_user_agent
 from ..exceptions import ConnectionError
-from ..response import BaseHTTPResponse
+from ..response import BaseHTTPResponse, BytesQueueBuffer
 
 orig_HTTPSConnection = HTTPSConnection
 
@@ -87,7 +89,7 @@ class HTTP2Connection(HTTPSConnection):
         self, host: str, port: int | None = None, **kwargs: typing.Any
     ) -> None:
         self._h2_conn = self._new_h2_conn()
-        self._h2_stream: int | None = None
+        self._h2_streams: dict[int, HTTP2Stream] = dict()
         self._headers: list[tuple[bytes, bytes]] = []
 
         if "proxy" in kwargs or "proxy_config" in kwargs:  # Defensive:
@@ -109,13 +111,14 @@ class HTTP2Connection(HTTPSConnection):
             if data_to_send := conn.data_to_send():
                 self.sock.sendall(data_to_send)
 
-    def putrequest(  # type: ignore[override]
+    def _putrequest(
         self,
+        stream_id: int,
         method: str,
         url: str,
         **kwargs: typing.Any,
     ) -> None:
-        """putrequest
+        """_putrequest
         This deviates from the HTTPConnection method signature since we never need to override
         sending accept-encoding headers or the host header.
         """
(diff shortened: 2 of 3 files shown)


pquentin:
Thank you for starting this! I haven't looked at the code yet, only at your original description.

> I have merged these things together into because they are kind of interdependent and gives an idea about the whole picture, like how the API will look like.

We should try to work on as small units of work as possible. If the existing issues don't work well, we can create new ones with a different split. That said, yes, we need to agree on the API itself.

> Adds one new object called `HTTP2Stream`, which manages the state of a stream. (similar to idea begin by @pquentin in #3294)

Before we move forward with this, I'd like to have feedback from @sethmlarson on this HTTP2Stream class.

> Important thing to note is that 1. `request` returns a `HTTP2Stream` and 2. `getresponse` needs to be called on the `HTTP2Stream` to get a `HTTP2Response`

We need to keep the existing APIs. In other words, we need request() to return an HTTP2Response. When created, this response class can take ownership of an existing HTTP2Stream.

mrdaybird:
Thanks for the reply!

> We need to keep the existing APIs. In other words, we need request() to return an HTTP2Response. When created, this response class can take ownership of an existing HTTP2Stream.

The existing `request` method in `HTTPSConnection` does not return anything, but the `getresponse` method returns the HTTP2Response. This doesn't quiet work for HTTP/2 because HTTP/2 can fire up multiple requests but defer the receiving the responses for later. If `request` does not return anything, we have no way of associating a request with its response!

> We should try to work on as small units of work as possible. If the existing issues don't work well, we can create new ones with a different split. 

You are right, I should split up the work. I merged them because it was easier to test things out after everything was in-place. Also, I suggest the following changes to the existing issues:
1. #3301 can be changed to 'Add support for HTTP/2 multiplexing using streams'. The original issue suggested that the connection be reused for multiple request, in HTTP/2 this can be achieved using streams.
2. #3300 suggest adding support for `read` and `stream` methods, but I think when implementing multiplexing (point 1), we have to implement `read`, thus most of the work for this issue can be done in the previous issue. The only thing that is left is `stream` method, which is a wrapper over `read` method.
3. #3294 suggests that the data processing, which I think means receiving from socket + decoding, be moved to HTTP2Response. But the events received over socket will contain event for multiple streams, thus the reading socket + h2 event loop should happen within the connection. Thus, the only part left is decoding the data.

sethmlarson:
Being brief: I was imagining a one-to-one mapping of HTTP2Connection to stream ID and then relying on a global registry of sockets/HTTPStream-like objects so that we can continue using the existing `HTTPConnection.getresponse() -> HTTPResponse` API. Let me know if you need further clarification about my design ideas up to this point.

hold7door:
> Being brief: I was imagining a one-to-one mapping of HTTP2Connection to stream ID and then relying on a global registry of sockets/HTTPStream-like objects so that we can continue using the existing `HTTPConnection.getresponse() -> HTTPResponse` API. Let me know if you need further clarification about my design ideas up to this point.

So, in this approach will there will one request active at a time in a connection. Since, each stream handles one request and its associated response.

sethmlarson:
@hold7door Correct! That's already the case for HTTP/1.1 since we don't support pipe-lining.

mrdaybird:
@sethmlarson I was wondering about how the multiplexing for HTTP/2 will look like in your design? Is it something like, one TCP connection/Socket is associated with multiple HTTP2Connection object. Also, where will the socket data(h2 events) be handled?

sethmlarson:
@mrdaybird Multiplexing is handled by the HTTPSConnectionPool, similar to how you need to use this interface for multiplexing in HTTP/1.1.

illia-v:

PR #3477, opened 2024-09-10
Testing: Install memray on Python 3.13

hugovk:
There's now a memray release with wheels for Python 3.13: https://github.com/bloomberg/memray/pull/658#issuecomment-2338959083

So we can now install it for testing 3.13.

I think this can skip a changelog.


illia-v (approved):
Thanks @hugovk!

PR #3479, opened 2024-09-13
Update accept_encoding=True documentation for Zstandard

naglis:
Currently, `zstd` is not mentioned, even though it [will be added](https://github.com/urllib3/urllib3/blob/2458bfcd3dacdf6c196e98d077fc6bb02a5fc1df/src/urllib3/util/request.py#L31-L36) in the `accept_encoding=True` case if `zstandard` is installed.


illia-v (approved):
LGTM
Thanks @naglis!

PR #3480, opened 2024-09-16
Added pickling support to NewConnectionError and NameResolutionError

chrisoro:
Found out that these two exceptions don't support pickling. Added reduce in similar fashion to other Exceptions (will lose info but you can pickle it...)

### DIFF
--- a/changelog/3480.bugfix.rst
+++ b/changelog/3480.bugfix.rst
@@ -1,0 +1,1 @@
+Added pickling support to ``NewConnectionError`` and ``NameResolutionError``.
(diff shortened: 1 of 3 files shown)


changelog/3480.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added pickling support to NewConnectionError and NameResolutionError.
```
illia-v, line 1:
```suggestion
Added pickling support to ``NewConnectionError`` and ``NameResolutionError``.
```

illia-v (commented):
Thanks! Please add the exceptions to `test/test_exceptions.py::TestPickle::test_exceptions`

illia-v (approved):

PR #3485, opened 2024-10-03
Plug mypy override complaint

### LINKED ISSUE #3483: Please update the headers= type
https://github.com/python/typeshed/pull/12704 was merged in typeshed, that expands the header value types.

urllib3 subclasses HTTP[S]Connection, and needs to make the same change.

dimaqq:
Fixes #3483


sethmlarson:
This is [failing our CI with the added comment](https://github.com/urllib3/urllib3/actions/runs/11155152857/job/31005620485?pr=3485#step:6:18), I don't think we need to ignore here? Closing for now.

dimaqq:
It will be needed when the typeshed change percolates through.

PR #3486, opened 2024-10-05
Correction of the load_default_certs call condition

niking00:
Correction of the `load_default_certs` call condition - there is no need to try to load system certificates if the user specified `cert_reqs='CERT_NONE'`

This change is necessary because certificates in the system may be incorrect, and an error will be caused when initializing the `PoolManager` with the `cert_reqs='CERT_NONE'` parameter, although loading certificates with the `cert_reqs='CERT_NONE'` parameter is not implied by logic.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -955,7 +955,7 @@ def _ssl_wrap_socket_and_match_hostname(
     # custom pyOpenSSL SSLContext objects because they don't support
     # load_default_certs().
     if (
-        resolve_cert_reqs(cert_reqs) != ssl.CERT_NONE
+        context.verify_mode != ssl.CERT_NONE
         and not ca_certs
         and not ca_cert_dir
         and not ca_cert_data


niking00:
Related issue #3519

changelog/3519.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed an issue useless loading system certificates when specifying ``cert_reqs='CERT_NONE'`` in ``HTTPSConnectionPool``
```
sigmavirus24:
```suggestion
Prevent loading of system certificates when ``cert_reqs='CERT_NONE'`` is specified for ``HTTPSConnectionPool``.
```

src/urllib3/connection.py
```diff
@@ -944,11 +944,13 @@ def _ssl_wrap_socket_and_match_hostname(
...
 
-    # Try to load OS default certs if none are given. We need to do the hasattr() check
-    # for custom pyOpenSSL SSLContext objects because they don't support
+    # Try to load OS default certs if none are given and don`t set
+    # cert_reqs='CERT_NONE'. We need to do the hasattr() check for
+    # custom pyOpenSSL SSLContext objects because they don't support
     # load_default_certs().
     if (
-        not ca_certs
+        cert_reqs != ssl.CERT_NONE.name
```
sigmavirus24:
```py
    cert_reqs: None | str | int,
```

So this fixes the `str` case but if someone passes `ssl.CERT_NONE` which is an `enum.IntEnum` that is actually [`ssl.VerifyMode`](https://docs.python.org/3/library/ssl.html#ssl.VerifyMode) then we won't find that.

We should have a function that takes the `cert_reqs` we receive and converts it to `None | ssl.VerifyMode` that way we can say `cert_reqs != ssl.CERT_NONE`

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1752,6 +1752,19 @@ def close(self) -> None:
...
+        context.load_default_certs = mock.Mock()
+        context.options = 0
+
+        with mock.patch("urllib3.util.ssl_.SSLContext", lambda *_, **__: context):
+            with HTTPSConnectionPool(
+                self.host, self.port, cert_reqs="CERT_NONE"
+            ) as pool:
+                with pytest.raises(Exception):
+                    pool.request("GET", "/", timeout=SHORT_TIMEOUT)
+                context.load_default_certs.assert_not_called()
```
sigmavirus24: We should have tests that verify `cert_reqs=ssl.CERT_NONE` as well.

src/urllib3/connection.py
niking00: We already have the function `resolve_cert_reqs`, it will be enough for strict verification of `cert_reqs!= ssl.CERT_NONE`. I was thinking about adding a separate function, but the `resolve_cert_reqs` function is already often used in the codebase, and adding a new function will only confuse.

src/urllib3/connection.py
```diff
@@ -944,11 +944,13 @@ def _ssl_wrap_socket_and_match_hostname(
...
 
-    # Try to load OS default certs if none are given. We need to do the hasattr() check
-    # for custom pyOpenSSL SSLContext objects because they don't support
+    # Try to load OS default certs if none are given and don`t set
+    # cert_reqs='CERT_NONE'. We need to do the hasattr() check for
+    # custom pyOpenSSL SSLContext objects because they don't support
     # load_default_certs().
     if (
-        not ca_certs
+        resolve_cert_reqs(cert_reqs) != ssl.CERT_NONE
```
pquentin: We already have the result of `resolve_cert_reqs()`, let's use it:

```suggestion
        context.verify_mode != ssl.CERT_NONE
```

PR #3489, opened 2024-10-24
Fix HTTP/1.1 debug log

pquentin:
Thanks @nateprewitt for the report.

### DIFF
--- a/changelog/3489.bugfix.rst
+++ b/changelog/3489.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed an issue in debug logs where the HTTP version was rendering as "HTTP/11" instead of "HTTP/1.1".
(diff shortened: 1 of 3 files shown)


sethmlarson (dismissed):
LGTM, thanks!

sethmlarson:
Ah, do we need a changelog for this since it's a bugfix?

Pierre-Sassoulas:
Fun fact, this bug fix broke ``pytest-vcr`` and ``vcrpy``, they now crash on existing cassette with the following error message `'VCRHTTPResponse' object has no attribute 'version_string'`. You'll probably guess what xkcd I'm going to link [here](https://xkcd.com/1172/)

Pierre-Sassoulas:
Didn't look too much into it, but downgrading to urllib < 2.3 works and this was the only occurrence of 'version_string' in the whole diff from 2.2.3..2.3.0, so I suppose it's a reasonable guess that it's the "culprit". (I'm not saying that urllib did anything wrong here, just sharing the 'fun' result of 1 hour of debug 😄)

PR #3492, opened 2024-10-27
Drop support for EOL Python 3.8

illia-v:


pquentin (dismissed):

pquentin (commented):
I'm still seeing references to Python 3.8 using `git grep 3\\.8`.

illia-v:
> I'm still seeing references to Python 3.8 using `git grep 3\\.8`.

There are two checks which could be removed when support for 3.7 was dropped
https://github.com/urllib3/urllib3/blob/8e478e02896e0958dc7d78974d16f8add16bada9/src/urllib3/util/ssl_.py#L347-L352
https://github.com/urllib3/urllib3/blob/81b76087aa91054712a533cb66bc556cc9355c8a/src/urllib3/connection.py#L218-L220

I am not sure whether they were left intentionally, do you know?

pquentin:
No, I just forgot to grep for Python 3.8 when removing 3.7: https://github.com/urllib3/urllib3/pull/3143.

sethmlarson:
Great work yall, as always it's exciting to drop all these conditional features/backports.

illia-v:
> No, I just forgot to grep for Python 3.8 when removing 3.7: #3143.

Thanks for checking!
 
> Also, I just grepped for 3.9, and we can remove the zoneinfo backport in tz_stub.py.

Great find 👍🏻

pquentin (approved):
Thanks! Looks good to me.

PR #3493, opened 2024-10-27
Upgrade macOS from 12 to 15 for CI

illia-v:
GitHub prepares to remove macOS 12 runners


sethmlarson (approved):
Thanks @illia-v, we should also do this on our 1.26.x branch.

PR #3495, opened 2024-10-27
[1.26] Upgrade macOS from 12 to 15 for CI

illia-v:
Backport of #3493


sethmlarson (approved):
LGTM, thanks Illia! :purple_heart:

illia-v:
We need to use x64 (macOS 13 as per [the table](https://github.com/actions/runner-images/tree/main?tab=readme-ov-file#available-images)) instead of arm64 to support older Python releases

And I'll have to check the coverage artifact upload failure

PR #3496, opened 2024-10-27
[1.26] Fix uploading coverage artifacts

illia-v:
Our 1.26 CI [fails](https://github.com/urllib3/urllib3/actions/runs/11542070607/job/32124845927) because [a recent actions/upload-artifact version](https://github.com/actions/upload-artifact/releases/tag/v4.4.0) stopped uploading hidden files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,11 +123,12 @@ jobs:
 
       - name: Upload Coverage
         if: ${{ matrix.nox-session != 'unsupported_python2' }}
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! Looks good to me. Is this what we do in the main branch too?

illia-v:
> Is this what we do in the main branch too?

The action is pinned in the main branch so the error hasn't happened, we'll have to copy this change when we upgrade the action

Issue #3410, opened 2024-06-17
Sunsetting official support for Python 2


sethmlarson:
urllib3 1.x has supported Python 2 for over 4 years since the end of Python 2's security release cycle. Since that time, download numbers indicate that Python 2 makes up [around 2%](https://pypistats.org/packages/urllib3) of our total downloads across all versions.

Today it's quite difficult to actually test Python 2 on our 1.26.x branch due to dependencies and GitHub Actions moving on from Python 2 so long ago. I propose we update the minimum Python version on the 1.26.x branch to be "3" in the packaging metadata and remove any Python 2 CI/testing to avoid further blocking that branch. We won't intentionally break Python 2 support, but in the future a security fix may do so.

We'll also keep App Engine support around, but will be dropping the requirement to test it since it requires Python 2.

pquentin:
> I propose we update the minimum Python version on the 1.26.x branch to be "3" in the packaging metadata

Doing this will effectively block Python 2 users to a specific 1.26.x version - if we don't want to break Python 2 and make sure our fixes reach those users, we can't do that specific part.

> and remove any Python 2 CI/testing to avoid further blocking that branch.

While this seems workable, I don't really trust myself to keep Python 2 support working without CI keeping me honest. The ideal situation as far as I'm concerned would be to at some point completely drop Python 2. (And other Python EOL versions, possibly keeping the latest EOL for longer).

> Since that time, download numbers indicate that Python 2 makes up [around 2%](https://pypistats.org/packages/urllib3) of our total downloads across all versions.

In June as you mentioned we were around 2% across all versions (~1.7% during the week, ~2.1% during the weekend). In October, usage has dropped again, around ~1% (~0.9% during the week, ~1.0% during the weekend). (We've always had different number during the week and the weekend, I suspect hobbyists are more represented during the weekend, and they tend to stick to older versions of Python and urllib3.)

Talking about numbers, here are the current numbers on urllib3 1.26.x. Using the following query on BigQuery:

```sql
SELECT
  REGEXP_EXTRACT(details.python, r"[0-9]+\.[0-9]+") AS python_version,
  COUNT(*) AS num_downloads,
FROM `bigquery-public-data.pypi.file_downloads`
WHERE
  DATE(timestamp)
    BETWEEN DATE_TRUNC(DATE_SUB(CURRENT_DATE(), INTERVAL 1 WEEK), DAY)
    AND CURRENT_DATE()
  AND file.project = 'urllib3'
[...]

PR #3497, opened 2024-10-27
[1.26] Update CI

illia-v:
I combined a few recent CI updates in this PR for 1.26.x


noxfile.py
```diff
@@ -51,7 +51,19 @@ def tests_impl(session, extras="socks,secure,brotli"):
 
 
 @nox.session(
-    python=["2.7", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13", "pypy"]
+    python=[
+        "2.7",
+        "3.6",
+        "3.7",
+        "3.8",
```
pquentin, line 58: Should we remove support for those in future work?

pquentin (approved):

noxfile.py
illia-v: I'm aware of #3410 and don't have a strong opinion on this, but I would prefer not to drop support for older Python versions in 1.26.x unless we encounter significant CI hurdles

pquentin: To be clear I specifically meant 3.6, 3.7 and maybe 3.8, not 2.7. Regarding #3410, I'd like to actually drop Python 2 entirely at some point, not only CI.

PR #3498, opened 2024-10-27
Start testing with Python 3.14

illia-v:
Blocked by https://github.com/pallets/quart/pull/369 at the moment


sethmlarson (approved):
This is really exciting, thank you Illia! :D

jtc1246:
But I tested it today, if I use this Quart version, it cannot run on all the python versions, https://github.com/jtc1246/urllib3/actions/runs/11984164045. But don't use this version (only Quart in dev-requirements.txt), it can only work on lower than 3.14, https://github.com/jtc1246/urllib3/actions/runs/11984277741.

pquentin:
Thanks, the build of the main branch is indeed broken right now, we'll have to fix it.

PR #3500, opened 2024-10-28
Bind test server sockets to IPv6 and IPv4

pquentin:
We lost that feature when migrating from Tornado to Hypercorn. We reimplement it here by overriding `hypercorn.Config.create_sockets`.

This makes Windows tests fast again (from 20 minutes to 4 minutes, and no longer the bottleneck), as on Windows trying to bind to IPv6 would take 2 seconds to fail, making tests much slower. This was quite baffling, but PyInstrument and then some prints in `create_connection` helped find the problem. I documented my thought process in https://fosstodon.org/@quentinpradet/113380164503305815.

I also included a change to print a traceback when the server fails to start, which has been really useful for debugging.

### DIFF
--- a/dummyserver/hypercornserver.py
+++ b/dummyserver/hypercornserver.py
@@ -2,18 +2,82 @@ from __future__ import annotations
 
 import concurrent.futures
 import contextlib
+import errno
 import functools
+import socket
 import sys
 import threading
+import traceback
 import typing
 
 import hypercorn
+import hypercorn.config
 import hypercorn.trio
 import hypercorn.typing
 import trio
 from quart_trio import QuartTrio
 
 from urllib3.util.url import parse_url
+
+
+class Config(hypercorn.Config):
+    def create_sockets(self) -> hypercorn.config.Sockets:
+        assert len(self.bind) == 1
+        secure_sockets, insecure_sockets = [], []
+        if self.ssl_enabled:
+            secure_sockets = self._create_urllib3_sockets(self.bind[0])
+        else:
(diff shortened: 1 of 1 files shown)


illia-v (approved):
Happy to see much faster CI!

dummyserver/hypercornserver.py
```diff
@@ -25,28 +89,35 @@ async def graceful_shutdown(shutdown_event: threading.Event) -> None:
...
+                    shutdown_trigger=functools.partial(
+                        graceful_shutdown, shutdown_event
+                    ),
+                )
             )
-        )
-        ready_event.set()
+            ready_event.set()
+        except Exception:
+            print("Starting server failed", file=sys.stderr)
```
graingert, line 111: Might be neater to use an anyio BlockingPortal using the https://anyio.readthedocs.io/en/stable/api.html#anyio.from_thread.BlockingPortal.start_task method, then startup failures should get propagated to the main thread properly

sethmlarson:
That's really interesting using PyInstrument to help debug this issue, potentially a blog post topic? I wouldn't have thought to reach for this tool!

dummyserver/hypercornserver.py
pquentin: Thanks, interesting suggestion! I'm not sure how to use that, though. (Also for what it's worth we use pure Trio, that no longer uses portal terminology.) Would you mind sharing more details/submitting a PR?

graingert: @pquentin here's the PR https://github.com/urllib3/urllib3/pull/3508

PR #3501, opened 2024-10-28
Remove failing test

pquentin:
This test is never going to pass, so we should remove it.


sethmlarson:
How did it pass before? Can we no longer control whether hypercorn offers HTTP/2?

pquentin:
This test never passed, it was committed like that in https://github.com/urllib3/urllib3/pull/3324. But I just realized that we can control `alpn_protocols` in `hypercorn.Config`, so I can start a dedicated server for this test.

PR #3505, opened 2024-10-29
Run emscripten tests faster in CI

pquentin:
By running Firefox and Chrome in parallel, not collecting thousands of skipped tests, and not printing thousands of line of wget progress.

With this change, all steps run in 5 minutes or less.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     name: ${{ fromJson('{"macos-15":"macOS","windows-latest":"Windows","ubuntu-latest":"Ubuntu","ubuntu-20.04":"Ubuntu 20.04 (OpenSSL 1.1.1)","ubuntu-22.04":"Ubuntu 22.04 (OpenSSL 3.0)"}')[matrix.os] }} ${{ matrix.python-version }} ${{ matrix.nox-session}}
     continue-on-error: ${{ matrix.experimental }}
-    timeout-minutes: 30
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -73,15 +73,19 @@ jobs:
             os: ubuntu-20.04  # CPython 3.9.2 is not available for ubuntu-22.04.
             experimental: false
             nox-session: test-3.9
-          - python-version: "3.11"
+          - python-version: "3.12"
             os: ubuntu-latest
-            nox-session: emscripten
+            nox-session: emscripten(firefox)
```
illia-v, line 78: We have to change the session name in the steps installing browsers
https://github.com/urllib3/urllib3/blob/756f8b8fa85ad1807047c15bea766e1a8f1644ab/.github/workflows/ci.yml#L104-L109

It's interesting that the tests pass without installing browsers at the moment

pquentin: Good catch!

@joemarshall Any idea if this is actually needed or not?

pquentin: Looking at https://github.com/urllib3/urllib3/pull/3427/files, it looks like we have an use case for specifying a version, but then there's a lot of ln and rm happening to actually use it?

That said, to avoid blocking this pull request, I fixed the issue. And then ran into a new Selenium warning introduced today, that I reported at https://github.com/SeleniumHQ/selenium/issues/14622 and added to our warning filters list.

illia-v (approved):
LGTM, thanks!

PR #3508, opened 2024-10-29
Use a blocking portal to collect hypercorn startup failures

graingert:

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,7 +10,8 @@ trustme==1.2.0
 cryptography==43.0.1
 towncrier==23.6.0
 pytest-memray==1.7.0;python_version<"3.14" and sys_platform!="win32" and implementation_name=="cpython"
-trio==0.26.2
+anyio[trio]==4.6.2.post1
+trio==0.27.0
 # https://github.com/pallets/quart/pull/369
 Quart @ git+https://github.com/pallets/quart@67110bf383d8973bce1619e957b4b6ea088ad9f2
 quart-trio==0.11.1
--- a/dummyserver/hypercornserver.py
+++ b/dummyserver/hypercornserver.py
@@ -1,15 +1,14 @@
 from __future__ import annotations
 
-import concurrent.futures
 import contextlib
 import errno
 import functools
 import socket
 import sys
-import threading
-import traceback
 import typing
 
+import anyio.abc
+import anyio.to_thread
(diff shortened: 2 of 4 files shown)


dummyserver/hypercornserver.py
```diff
@@ -80,37 +79,25 @@ def _create_urllib3_sockets(self, bind: str) -> list[socket.socket]:
...
+    shutdown_event: trio.Event,
+    *,
+    task_status: anyio.abc.TaskStatus[list[str]] = anyio.TASK_STATUS_IGNORED,
 ) -> None:
     async with trio.open_nursery() as nursery:
-        try:
-            config.bind = await nursery.start(
+        with trio.fail_after(5):
+            config_bind: list[str]
+            config.bind = config_bind = await nursery.start(
```
graingert, line 92: do we still need to mutate `config.bind` I think it's not used anymore

Mikoboys (approved):

dummyserver/hypercornserver.py
graingert, line 92:
```suggestion
            config_bind = await nursery.start(
```

dev-requirements.txt
```diff
@@ -9,8 +9,9 @@ idna==3.7
 trustme==1.2.0
 cryptography==43.0.1
 towncrier==23.6.0
+anyio[trio]==4.6.2.post1
+trio==0.27.0
 pytest-memray==1.7.0;python_version<"3.14" and sys_platform!="win32" and implementation_name=="cpython"
```
graingert, line 14:
```suggestion
pytest-memray==1.7.0;python_version<"3.14" and sys_platform!="win32" and implementation_name=="cpython"
anyio[trio]==4.6.2.post1
trio==0.27.0
```

graingert:
@pquentin conflicts resolved

pquentin:
Thanks! Can you please remove the changelog entry?

Also, I think I have this before, but is it easy to stick to Trio instead of AnyIO? Or is AnyIO API just better here?

graingert:
> Thanks! Can you please remove the changelog entry?

Done!

> Also, I think I have this before, but is it easy to stick to Trio instead of AnyIO? Or is AnyIO API just better here?

There's no trio equivalent of the blocking portal

pquentin (approved):
Thanks! LGTM if CI still passes.

PR #3512, opened 2024-11-07
Implement context-manager for BaseHTTPResponse

paytonward6:
Implement context-manager for `BaseHTTPResponse` to automatically call `close()` and `release_conn()`.


paytonward6:
Related issue #3513

PR #3516, opened 2024-11-12
issue a ResourceWarning if fetch._ReadStream is not closed correctly

graingert:


src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -121,9 +122,6 @@ def __init__(
         self._is_closed = False
         self.request: EmscriptenRequest | None = request
 
-    def __del__(self) -> None:
```
graingert, line 124: RawIOBase already has a tp_finalize that sets "_finalizing" and calls ".close()"

PR #3518, opened 2024-11-12
Drop unneeded dependency pins and a warning filter

illia-v:


sethmlarson (approved):
Awesome, thank you @illia-v! :purple_heart:

PR #3522, opened 2024-11-24
Add license field to pyproject.toml

s-t-e-v-e-n-k:
While investigating this package on deps.dev[1], I discovered this package is shown as having an unknown license. Given the packaging guidelines, this is likely a bug in deps.dev, but I thought we could work around it with one line in the project metadata.

1: https://deps.dev/pypi/urllib3

### DIFF
--- a/changelog/3522.feature
+++ b/changelog/3522.feature
@@ -1,0 +1,1 @@
+Apply PEP 639 by specifying the license fields in pyproject.toml.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,8 @@ build-backend = "hatchling.build"
 name = "urllib3"
 description = "HTTP library with thread-safe connection pooling, file post, and more."
 readme = "README.md"
+license = "MIT"
+license-files = ["LICENSE.txt"]
 keywords = ["urllib", "httplib", "threadsafe", "filepost", "http", "https", "ssl", "pooling"]
 authors = [
   {name = "Andrey Petrov", email = "andrey.petrov@shazow.net"}
@@ -20,7 +22,6 @@ maintainers = [
 classifiers = [
   "Environment :: Web Environment",
   "Intended Audience :: Developers",
-  "License :: OSI Approved :: MIT License",
   "Operating System :: OS Independent",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3",


pquentin (commented):
Hello, PEP 639 is still not finalized, which is why we rejected https://github.com/urllib3/urllib3/pull/3458 in the past. The Discourse thread does show some progress however.

s-t-e-v-e-n-k:
PEP 639 has been accepted, `Provisional` means that implementations can go ahead. setuptools 75.4, hatchling 1.26.0 and PyPi itself support Metadata-Version as of two weeks ago.

illia-v:
FYI, `License-File: LICENSE.txt` will be added to our wheel `METADATA` by default when hatchling starts using metadata version 2.4 by default (it has been done in a hatchling release, however not published to PyPI yet https://github.com/pypa/hatch/issues/1842)

image


This can be imitated by changing the metadata version in `pyproject.toml`:
```toml
[tool.hatch.build.targets.wheel]
core-metadata-version = "2.4"
```

webknjaz:
Additionally, https://github.com/urllib3/urllib3/blob/181357e/.github%2Fworkflows%2Fpublish.yml#L113 must be bumped before the release can succeed.

pquentin:
OK, I'm happy to follow PEP 639 now, but in addition to the above we should also specify `licenses-files` and remove the trove classifier. See https://github.com/python-attrs/attrs/pull/1377/files for an example.

s-t-e-v-e-n-k:
@pquentin I've rebased this and made the changes you requested.

pquentin (dismissed):
Thanks! LGTM.

pquentin:
Sorry, approved too soon. Can you please add a changelog entry? See https://github.com/urllib3/urllib3/tree/main/changelog for instructions.

s-t-e-v-e-n-k:
> Sorry, approved too soon. Can you please add a changelog entry? See https://github.com/urllib3/urllib3/tree/main/changelog for instructions.

Done!

changelog/3522.feature
```diff
@@ -0,0 +1 @@
+Appy PEP 639 by specifying the license fields in pyproject.toml.
```
pquentin, line 1:
```suggestion
Apply PEP 639 by specifying the license fields in pyproject.toml.
```

pquentin (approved):
Thank you! LGTM.

PR #3530, opened 2024-12-16
Pin dev dependencies with uv pip compile

pquentin:
We regularly get broken builds in CI when our indirect test dependencies get updated. This is challenging to fix.

 * We can't use `pip-compile` because it does not support environment markers like `python_version` or `os_name`, meaning that dependencies computed on Linux with CPython can break on Windows or PyPI, for example.
 * We can't use `uv` wholesale because it's [not compatible with hatch-vcs](https://github.com/astral-sh/uv/issues/7533). Also, some of our maintainers are worried about the current lack of business model and a potential rug pull from Astral.
 * We could use Poetry, but I would need to find out how good it is at resolving dependencies, and as far as I can tell, it would force us to stop using existing packaging standards (that uv follows) to do things the Poetry way instead.

However, one option is to use `uv pip compile,` which does precisely what pip-compile does, but correctly. If it comes to it, we can revert to standard pip-compile and manually fix any environment markers issues. (It's tricky but possible, Trio did it in the past.)


pquentin:
Never mind, `uv pip compile` does not have the capabilities of uv's lock file. Closing.

PR #3531, opened 2024-12-16
Fix PyPI publish with Core metadata 2.4

pquentin:
The latest PyPI push failed because the versions of twine and pkginfo did not support Core metadata 2.4; see https://github.com/urllib3/urllib3/actions/runs/12354192811/job/34475229840.

With twine 6, Core metadata 2.4 is ignored but does not prevent uploads. (This means the new license metadata discussed in #3522 won't be supported now.)


pquentin:
Thanks for the review! This worked: https://github.com/urllib3/urllib3/actions/runs/12358966992/job/34491143732

PR #3532, opened 2024-12-18
adding host details in proxy headers

HarikrishnanTechie:
Added host details in proxy headers so that, request message will have host details

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -1033,6 +1033,9 @@ class HTTPSConnectionPool(HTTPConnectionPool):
         else:
             tunnel_scheme = "http"
 
+        if self._tunnel_host: #Adding host details in proxy headers
+            self.proxy_headers.update({"Host":self._tunnel_host})
+
         conn.set_tunnel(
             scheme=tunnel_scheme,
             host=self._tunnel_host,
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -1033,8 +1033,8 @@ class HTTPSConnectionPool(HTTPConnectionPool):
         else:
             tunnel_scheme = "http"
 
-        if self._tunnel_host: #Adding host details in proxy headers
-            self.proxy_headers.update({"Host":self._tunnel_host})
+        if self._tunnel_host:  # Adding host details in proxy headers
+            self.proxy_headers.update({"Host": f"{self._tunnel_host}:{self.port}"})
 
         conn.set_tunnel(
             scheme=tunnel_scheme,
--- a/changelog/3532.feature.rst
+++ b/changelog/3532.feature.rst
@@ -1,0 +1,1 @@
+Added host details in proxy headers for better debugging and tracing.


src/urllib3/connectionpool.py
```diff
@@ -1033,6 +1033,9 @@ def _prepare_proxy(self, conn: HTTPSConnection) -> None:  # type: ignore[overrid
         else:
             tunnel_scheme = "http"
 
+        if self._tunnel_host:  # Adding host details in proxy headers
```
sigmavirus24, line 1036: We use this unconditionally a little lower. Why are we guarding it here? 

Also, is this just for debugging because this:

a. Breaks tests
b. Seems superfluous and unnecessary

HarikrishnanTechie: Hi @sigmavirus24,

The main enhancement in my code is the inclusion of a Host header in the proxy_headers dictionary when a tunneling host (_tunnel_host) is defined. This change is specifically for cases where a proxy is being used.

Why This Change is Important
Complying with Proxy Protocols:
Including the Host header in the CONNECT request ensures the proxy server can correctly identify the target host for establishing the tunnel. This is especially critical for HTTPS connections where secure communication is required.

Enhancing Proxy Support:
By adding host details dynamically, your code becomes more robust and compatible with environments that rely on proxies. Some proxies may explicitly require the Host header to process requests correctly.

python_verison: 3.10. 
Can we connect to dicuss more about my requirement?

sigmavirus24: This feeds into the stdlib http.client.HTTPConnection and thus isn't necessary https://github.com/python/cpython/blob/180d417e9f9456defd4c5b53cae678c318287921/Lib/http/client.py#L919

Connecting isn't necessary because the tests still fail and you still haven't shown how this fixes a problem or explained what the problem is

HarikrishnanTechie: Problem Statement: In the connect request header, the expected host details (e.g., Host: google.com) is missing. Please find the attachment for the same.
image

Evidence: The screenshot (as described) shows the connect request header missing the Host field. This deviation is causing issues with meeting the application's requirements.

Note:
We are using proxy environment.

sigmavirus24: 3.12 is the first Python to have that logic. You'll need to advocate for the backport of [this commit][] as a bug fix to 3.11, 3.10, and 3.9. 3.8 is EOL. Also if you read that commit closely, **you** are not using `HTTP/1.0` otherwise this wouldn't fail.

Also your screenshot contains clear-text credentials username `alteryx`, password `alt3ryx`. You should change those credentials immediately. 

[this commit]: https://github.com/python/cpython/commit/1a8f862e329c3872a11d4ef8eb85cf353ca2f4d5

HarikrishnanTechie: we are using 3.10 version. Also is there any other alternative to add host details in request headers for connect request?

sethmlarson:
Going to close this one as it appears the discussion has stalled.

PR #3534, opened 2024-12-21
Update .readthedocs.yml addressing a deprecation

illia-v:
Read the Docs sent us a message about a deprecation of projects without explicit Sphinx configuration
https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

We have to set a path to the Sphinx configuration file in .readthedocs.yml explicitly to prevent failing builds in the future

In addition to the required change, I bumped OS and Python versions for the docs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,9 +1,9 @@
 version: 2
 
 build:
-  os: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! LGTM.

PR #3538, opened 2024-12-23
Link to annual reports in sponsors page

pquentin:
Listing all sponsors on that page is too much work, so linking to the reports is a good alternative. Let me know what you think.


sigmavirus24 (approved):
One orthogonal concern I have is: What happens to these posts if something happens to Seth (moves on from urllib3, website goes offline, etc.)? Can we at least "mirror" these to the docs here?

pquentin:
Good idea! Seth, are your source files publicly available?

PR #3539, opened 2024-12-23
Improve release process

pquentin:
The release process is excellent and involves little stress. Still, here are a few things that would help me:

 * Agree on the social media announcement before merging the pull request. Summarizing pull requests you haven't worked on is not always easy, and this is what I spend the most time doing today when releasing a new version.
 * Make the git commands easy to copy/paste to avoid typos.
 * Clarify that pushing the tag will trigger the publish workflow; it does not need to be done manually.
 * Link the exact Tidelift page that needs to be checked
 * Use `CHANGES.rst` as a trigger for CODEOWNERS review since we no longer have a version file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,6 @@
 
 .github/workflows/          @sethmlarson @pquentin @shazow @illia-v
 .github/CODEOWNERS          @sethmlarson @pquentin @shazow @illia-v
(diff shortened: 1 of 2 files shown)


sigmavirus24 (approved):

sethmlarson (approved):
LGTM! Thanks Quentin!

PR #3543, opened 2024-12-28
Pin Quart to 0.20.0 instead of a Git commit

illia-v:
https://github.com/pallets/quart/releases/tag/0.20.0 includes a change we needed for testing with Python 3.14.


sethmlarson (approved):
Thanks Illia!

PR #3545, opened 2024-12-30
Ensure compatibility with httpx>=0.28

carlsmedstad:
Hey, Arch Linux maintainer here :wave:

Version 0.28 of httpx removed support for supplying a path (of string type) to `verify`, only a bool or an SSL context is now supported. See: https://github.com/encode/httpx/releases/tag/0.28.0

Running the test suite with httpx 0.28 will break the dummy server and a such number of tests in `test/with_dummyserver/`.

To resolve this, create an SSL context in the `ProxyApp` init function and supply that to `AsyncClient`, instead of a raw string. This change is backwards compatible, i.e. the test suite will still succeed against the currently pinned version of httpx, 0.25.2.

### DIFF
--- a/dummyserver/asgi_proxy.py
+++ b/dummyserver/asgi_proxy.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ssl
 import typing
 
 import httpx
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):
Thank you!

pquentin:
Thanks! I'm glad that there was a good way to ensure compatibility and that you submitted a pull request, but note that we don't generally guarantee compatibility of test dependencies with versions that are not in https://github.com/urllib3/urllib3/blob/main/dev-requirements.txt

PR #3547, opened 2025-01-09
SingleTLSLayerTestCase: Catch BrokenPipeError

cjwatson:
OpenSSL 3.4.0 returns `ERR_LIB_SYS` in some more situations than it used to.  In the case exercised by
`SingleTLSLayerTestCase.test_close_after_handshake`, https://github.com/python/cpython/pull/127361 (also backported to the 3.12 and 3.13 branches) turns this into `BrokenPipeError`.  It seems reasonable to treat this in the same way as `ConnectionAbortedError` and `ConnectionResetError`.

### DIFF
--- a/test/test_ssltransport.py
+++ b/test/test_ssltransport.py
@@ -122,7 +122,7 @@ class SingleTLSLayerTestCase(SocketDummyServerTestCase):
                         return
                     validate_request(request)
                     ssock.send(sample_response())
-            except (ConnectionAbortedError, ConnectionResetError):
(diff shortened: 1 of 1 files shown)


cjwatson:
I possibly should have mentioned that I noticed this because it's currently causing a build failure in Debian unstable.

cjwatson:
I think this probably doesn't deserve a changelog entry based on what seem to be the usual practices in this repository, but let me know if you disagree and I'll write one.

illia-v (approved):
Thank you @cjwatson!

PR #3549, opened 2025-01-11
The QUERY method is retryable too as idempotent

### LINKED ISSUE #3548: Please add support for the `QUERY` HTTP method
### Context

On 2025-01-07 the [`QUERY` HTTP method, "a safe, idempotent request method that can carry request content." has reached ~the level of a Proposed Standard~  consensus and is on track to become a Proposed Standard](https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/).

The QUERY method solves a problem when GET method is insufficient as it (according to the standard) lacks support for sending the query in the body, leaving only the query part of the URL which is limited.

Therefore it would be great to consider implementing support for QUERY method in urllib3.

### Alternatives

An often-used workaround of using the POST method for such queries has a downside in that it's just the wrong method for this purpose as queries do not cause side-effects and this limits the retryabilty and cacheability of such queries.

### Duplicate

I haven't found a duplicate.

### Contribution

Would you be willing to submit a PR? **Yes**

gdubicki:
according to its description in the Internet-Draft https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/

Fixes #3548

### DIFF
--- a/changelog/3549.feature.rst
+++ b/changelog/3549.feature.rst
@@ -1,0 +1,1 @@
+Added support for the `QUERY HTTP method (Proposed Standard) <https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/>`_. As urllib3 already allows making requests with arbitrary methods, this only means that QUERY is now another retryable by default method.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -182,7 +182,7 @@ class Retry:
 
     #: Default methods to be used for ``allowed_methods``
     DEFAULT_ALLOWED_METHODS = frozenset(
-        ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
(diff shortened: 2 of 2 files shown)


sigmavirus24 (approved):

gdubicki:
Sorry, I was mistaken. It is too early as it's NOT a Proposed Standard yet. See [this comment]([url](https://github.com/urllib3/urllib3/issues/3548#issuecomment-2585735906)) for more info,

gdubicki:
On second thought, if node.js has support for it, why shouldn't urllib3?

sigmavirus24:
Because they're different projects. If another project adds racist slurs to its documentation why does that compel urllib3 to do the same? Another project's decision creates no requirement for us

pquentin:
Closing as decided in https://github.com/urllib3/urllib3/issues/3548#issuecomment-2587744600. Thank you for bringing this to us!

PR #3550, opened 2025-01-15
Lock dev dependencies with uv

illia-v:
uv has fixed the issue which blocked us from adopting it for locking dev dependencies https://github.com/astral-sh/uv/issues/7533#issuecomment-2591516884

Here I've:
* converted all `*requirements.txt` files we have to dependency groups and generated `uv.lock`
* changed install commands in `noxfile.py` according to https://nox.thea.codes/en/stable/cookbook.html#using-a-lockfile
* changed `.readthedocs.yml` following https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv


.pre-commit-config.yaml
```diff
@@ -22,6 +22,11 @@ repos:
       - id: flake8
         additional_dependencies: [flake8-2020]
 
+  - repo: https://github.com/astral-sh/uv-pre-commit
+    rev: 0.5.20
+    hooks:
+      - id: uv-lock
+
```
pquentin, line 29: Does that mean pre-commit will fail every time there's a dependency update? If yes, I'd prefer to stick with periodic manual updates as we do now.

noxfile.py
```diff
@@ -131,7 +136,7 @@ def git_clone(session: nox.Session, git_url: str) -> None:
         session.run("git", "-C", expected_directory, "pull", external=True)
 
 
-@nox.session()
+@nox.session(venv_backend="virtualenv")
```
pquentin, line 139: Why is this needed for botocore but not requests?

noxfile.py
```diff
@@ -210,7 +215,7 @@ def emscripten(session: nox.Session, runner: str) -> None:
             "Node version:",
             session.run("node", "--version", silent=True, external=True),
         )
-    session.install("-r", "emscripten-requirements.txt")
+    session.install("build")
```
pquentin, line 218: Can we move this closer to the line that calls build?

pyproject.toml
```diff
@@ -135,3 +135,60 @@ warn_unused_ignores = true
...
   "ignore-without-code",
 ]
+
+[tool.uv]
+package = true
+
+[tool.uv.sources]
+hypercorn = { git = "https://github.com/urllib3/hypercorn", rev = "urllib3-changes" }
+
+[dependency-groups]
```
pquentin, line 145: Since this is standardized by PEP 735, can we move those just below the `[project.optional-dependencies]` section?

noxfile.py
```diff
@@ -30,13 +32,16 @@ def tests_impl(
...
 
     # Install deps and the package itself.
-    session.install("-r", "dev-requirements.txt")
-    if len(extras) > 0:
-        session.install(f".[{extras}]")
-    else:
-        session.install(".")
+    session.run_install(
+        "uv",
+        "sync",
```
pquentin, line 37: Should we add `--frozen` to avoid touching the lockfile, as mentioned in the nox docs? https://nox.thea.codes/en/stable/cookbook.html#using-a-lockfile

pquentin (commented):
Wow, that was fast. Excellent work, thank you!

.github/workflows/ci.yml
```diff
@@ -111,7 +111,7 @@ jobs:
           allow-prereleases: true
 
       - name: "Install dependencies"
-        run: python -m pip install --upgrade pip setuptools nox
+        run: python -m pip install --upgrade uv nox
```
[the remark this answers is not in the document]
illia-v: I did this to avoid adding a new action, but your suggestion should is fine too 👍🏻

.github/workflows/ci.yml
```diff
@@ -173,7 +173,8 @@ jobs:
           python-version: "3.x"
 
       - name: "Install coverage"
-        run: "python -m pip install -r dev-requirements.txt"
+        # Install the same version of coverage as in the lock file.
+        run: python -m pip install build coverage==$(awk '/^name = "coverage"/ {getline; gsub(/"/, "", $3); print $3}' uv.lock)
```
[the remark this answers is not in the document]
illia-v: That was over-optimization probably 😄

.pre-commit-config.yaml
illia-v: This will fail every time `uv.lock` is out of sync with packages defined in `pyproject.toml`.

Dependabot [doesn't support](https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependabot) uv yet. I dunno if it will continue opening PRs for security issues after requirements are listed in `pyproject.toml`.

noxfile.py
illia-v: botocore job failed with uv

noxfile.py
illia-v: Yes, let's add the flag

.pre-commit-config.yaml
pquentin: This looks good if running pre-commit won't update dependencies unless we modify pyproject.toml. :+1:

(Regarding Dependabot, it's OK; we don't really need Dependabot for our Python development dependencies.)

.github/workflows/ci.yml
```diff
@@ -183,6 +191,7 @@ jobs:
 
       - name: "Combine & check coverage"
         run: |
+          source .venv/bin/activate
```
pquentin, line 194: I wanted to note that `python -m build` can be updated to `uv run -m build`, which avoids activating the virtual environment. I'm not saying we should change it here, though.

noxfile.py
```diff
@@ -246,6 +251,7 @@ def emscripten(session: nox.Session, runner: str) -> None:
             )
 
         dist_dir = pyodide_artifacts_path
+    session.install("build")
     session.run("python", "-m", "build")
```
pquentin, line 255: Here too, `uv run -m build` can be used. But having to define `UV_PROJECT_ENVIRONMENT` makes it even less worth it than above.

pquentin (approved):
Thanks! Happy to merge as-is.

illia-v:
@pquentin I did switch to `uv run -m` there, thanks for the suggestions and review in general

pquentin (dismissed):
Thanks! I have one question left. Still happy to merge as is.

noxfile.py
```diff
@@ -192,9 +194,10 @@ def lint(session: nox.Session) -> None:
 
 @nox.session(python="3.12")
 def pyodideconsole(session: nox.Session) -> None:
+    session.env["UV_PROJECT_ENVIRONMENT"] = session.virtualenv.location
     # build wheel into dist folder
-    session.install("build")
-    session.run("python", "-m", "build")
+    session.run_install("uv", "sync", "--frozen", "--package", "build")
```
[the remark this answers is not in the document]
illia-v: You're correct

pquentin (approved):
Thanks for iterating! LGTM. 🚢

sethmlarson (approved):
The changes look good to me, thank you Illia! :purple_heart:

PR #3551, opened 2025-01-17
Fix nonexistent wheel and CORS errors in the Pyodide console

illia-v:
I noticed that the Pyodide console displays an error right after it's loaded because it has a previous version of urllib3 hard-coded, here I'm making it dynamic

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import re
 import shutil
 import sys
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks! LGTM.

sethmlarson (approved):
LGTM, wish that testing urllib3 with Pyodide was easier, this is quite a bit of indirection!

Issue #3555, opened 2025-01-29
Emscripten bug with partial reads


joemarshall:
On emscripten, if the request is read in chunks, sometimes it fails to read the whole amount. This is because the code in response.py assumes that if a read() call to the underlying request stream doesn't fill the buffer it is given, this must be the last buffer in the stream, which isn't the case sometimes, because the chunks we read are those given by javascript, which may be of arbitrary size.

https://github.com/urllib3/urllib3/blob/37565dc3addfa92ac89b0a3ae7b0134e4f3d6c72/src/urllib3/contrib/emscripten/response.py#L165-L167

I have a fix, which I'll make a PR for in a mo.

joemarshall:
https://github.com/pyodide/pyodide/discussions/5197

PR #3556, opened 2025-01-29
fixed bug in emscripten requests

joemarshall:
Fix for #3555

### DIFF
--- a/src/urllib3/contrib/emscripten/response.py
+++ b/src/urllib3/contrib/emscripten/response.py
@@ -160,26 +160,18 @@ class EmscriptenHttpResponseWrapper(BaseHTTPResponse):
                 # don't cache partial content
                 cache_content = False
                 data = self._response.body.read(amt)
-                if self.length_remaining is not None:
-                    self.length_remaining = max(self.length_remaining - len(data), 0)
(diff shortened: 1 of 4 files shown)


joemarshall:
Should be fixed now.

src/urllib3/contrib/emscripten/response.py
```diff
@@ -164,7 +164,7 @@ def read(
                     self.length_remaining = max(self.length_remaining - len(data), 0)
                 if (self.length_is_certain and self.length_remaining == 0) or len(
                     data
-                ) < amt:
+                ) == 0:
                     # definitely finished reading, close response stream
                     self._response.body.close()
                 return typing.cast(bytes, data)
```
illia-v, line 170: It looks like the `else` block ends with the same code now (lines 177–182), let's bring it out of `if/else`

illia-v (changes requested):
Thanks for fixing this!

joemarshall:
@illia-v 

I think I addressed all the comments here. Sorry for the delay.

illia-v (approved):

illia-v:
The PyPy tests fail in other branches too, the error needs separate investigating

PR #3559, opened 2025-02-06
Bump browser-actions/setup-firefox to fix CI

illia-v:
Our emscripten(firefox) CI job has recently started [failing](https://github.com/urllib3/urllib3/actions/runs/13138651938/job/36660115824#logs) because the newest [Firefox switched](https://blog.nightly.mozilla.org/2024/11/28/announcing-faster-lighter-firefox-downloads-for-linux-with-tar-xz-packaging/) from `.tar.bz2` to `.tar.xz` packaging. We have to use newer `browser-actions/setup-firefox` to fix the error.

BTW, this is not the only patch-version actions upgrade that we need to do. Newer `pypa/gh-action-pypi-publish` uploads correct license info to PyPI, for example. What do you think about reenabling patch-version dependabot upgrades with possible [grouping](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--) of all actions in a single PR?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,7 @@ jobs:
           google-chrome --version
         if: ${{ matrix.nox-session == 'emscripten(chrome)' }}
       - name: "Install Firefox"
-        uses: browser-actions/setup-firefox@233224b712fc07910ded8c15fb95a555c86da76f # v1.5.0
+        uses: browser-actions/setup-firefox@634a60ccd6599686158cf5a570481b4cd30455a2 # v1.5.4
         if: ${{ matrix.nox-session == 'emscripten(firefox)' }}
       - name: "Install node.js"
         uses: actions/setup-node@v4


pquentin (approved):
Thanks! LGTM.

I'm worried about the noise of dependabot - it should not be the main source of PRs. Right now it's at least monthly: https://github.com/dependabot/dependabot-core/issues/6339

illia-v:
> I'm worried about the noise of dependabot - it should not be the main source of PRs. Right now it's at least monthly: [dependabot/dependabot-core#6339](https://github.com/dependabot/dependabot-core/issues/6339)

If we group all actions together in `dependabot.yml`, it can be one PR per month

pquentin:
In practice it's more as we have dependabot enabled for security issues in GitHub workflows too. And urllib3 is not the only repository I maintain. My thinking is influenced by https://davidism.com/disabling-scheduled-dependency-updates/ here. I'm just not sure this adds a lot of value. But we can try if you think it would help!

PR #3561, opened 2025-02-16
TLS 1.3 post-handshake auth fix for kodi - dazn addon

MRLB:
Old if-statemant necessary again, otherwise for example Dazn-Addon in Kodi does not work probably and gives only following as answer:
{'odata.error': {'code': 10000, 'message': {'lang': 'en-US', 'value': 'Playback not allowed'}}}

If-statemant was changed with following commit:
https://github.com/urllib3/urllib3/commit/d7bb83b48c23d8dd429c20a43ac3da74ea6e9df0

(cc @jborean93 )

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -324,7 +324,10 @@ def create_urllib3_context(
     # necessary for conditional client cert authentication with TLS 1.3.
     # The attribute is None for OpenSSL <= 1.1.0 or does not exist when using
     # an SSLContext created by pyOpenSSL.
-    if getattr(context, "post_handshake_auth", None) is not None:
+    if (
+            cert_reqs == ssl.CERT_REQUIRED
+            and getattr(context, "post_handshake_auth", None) is not None
+    ):
         context.post_handshake_auth = True
 
     # The order of the below lines setting verify_mode and check_hostname


jborean93:
There have been a few changes in this section

+ https://github.com/urllib3/urllib3/commit/6a626be4ff623c25270e20db9002705bf4504e4e

This added support for TLS 1.3 post handshake auth with:

```python
if getattr(context, "post_handshake_auth", None) is not None:
```

+ https://github.com/urllib3/urllib3/commit/edc3ddb3d1cbc5871df4a17a53ca53be7b37facc

This added the `cert_reqs == ssl.CERT_REQUIRED` or `sys.version_info >= (3, 7, 4)` check in response to [this CPython bug](https://bugs.python.org/issue37428).

```python
if (cert_reqs == ssl.CERT_REQUIRED or sys.version_info >= (3, 7, 4)) and getattr(
    context, "post_handshake_auth", None
) is not None:
```

The bug was that on Python versions before 3.7.4 (where the backport was added to) setting `context.post_handshake_auth = True` would also implicitly enable cert verification on the client breaking clients who explicitly ignored the server cert.

Note that this check is an `or` check not an `and` meaning that the `cert_reqs = ssl.CERT_REQUIRED` side was only important when running on older Python versions.

+ https://github.com/urllib3/urllib3/commit/ae0668899a144edb06c4d2ca9cec96b760a93e6a

This commit dropped all Python 3.7 logic so that the check now became:

```python
if (
    cert_reqs == ssl.CERT_REQUIRED
    and getattr(context, "post_handshake_auth", None) is not None
):
```

This change is what ended up breaking Ansible and WinRM cert auth as now `post_handshake_auth` was only used if the server cert was checked. This was wrong as the previous check used `or` so that running on Python 3.8+ meant the old `cert_reqs == ssl.CERT_REQUIRED` check done for that CPython bug is no longer needed.

+ https://github.com/urllib3/urllib3/commit/d7bb83b48c23d8dd429c20a43ac3da74ea6e9df0

This was my commit which dropped the `cert_reqs == ssl.CERT_REQUIRED` check:

```python
if getattr(context, "post_handshake_auth", None) is not None:
```

I not sure this PR to revert back to the `cert_reqs == ssl.CERT_REQUIRED` check is correct as it essentially means post handshake auth is only done if the client is configured to check the server's cert which is unrelated.

It sounds like this Dazn Kodi addon, or the server API it is calling, is doing something else here that is causing the failure and needs to either be fixed or this client should use a custom SSLContext with the non-standard logic needed for that service.

jborean93:
https://urllib3.readthedocs.io/en/stable/advanced-usage.html#custom-ssl-contexts shows how to specify a custom `SSLContext` in `urllib3`. If you re using a library on top you'll have to see if they expose a way to pass this through.

If you were going to use `create_urllib3_contex()` you'll have to explicitly disable the `post_handshake_auth`

```python
import ssl

from urllib3.util import create_urllib3_context

ctx = create_urllib3_context()
ctx.load_default_certs()
ctx.post_handshake_auth = False

with PoolManager(ssl_context=ctx) as pool:
    pool.request("GET", "https://www.google.com/")
```

If you need custom behaviour like ignoring cert and hostname validation you can control the context behaviour through various arguments on the function https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.create_urllib3_context.

Issue #3565, opened 2025-02-26
[CI] PyPy tests fail because of unclosed keylogfile.txt


illia-v:
### Subject

[PyPy CI job](https://github.com/urllib3/urllib3/actions/runs/13530170912/job/37810256991) started to fail recently because of the following warning:
```
ResourceWarning: unclosed file <_io.TextIOWrapper name='/tmp/pytest-of-runner/pytest-0/test_sslkeylogfile3/keylogfile.txt' mode='at' encoding='UTF-8'>
```

It's likely that it's related to the "Add `_ssl.keylog_filename` which is useful for debugging ssl problems (https://github.com/pypy/pypy/commit/7c57fa51987752c699d88968ff7e1deb920e4572)" change in PyPy [v7.3.18](https://doc.pypy.org/en/latest/release-v7.3.18.html) released on 2025-02-06.

Although the file is opened in `test_sslkeylogfile`, the warning fails one of the next tests, maybe it's caused by delayed garbage collection like #2924 in the past.

### Environment

GitHub Actions

illia-v:
One of recent runs has been successful https://github.com/urllib3/urllib3/actions/runs/13546009736/job/37857690673

PR #3569, opened 2025-03-04
Add PyPy 3.11 to CI

illia-v:
PyPy includes Python 3.11 for about a month, let's add use it in the CI matrix.

I had to upgrade cryptography to 44.0.2 in dev requirements and add [`check-latest: true`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#check-latest-version) to the `actions/setup-python` step for tests. Without the flag, `actions/setup-python` still downloads v7.3.18 of PyPy instead of the newest v7.3.19 released on February 26. Our tests fail with v7.3.18, see #3565.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,10 @@ jobs:
             os: ubuntu-latest
             experimental: false
             nox-session: test-pypy3.10
+          - python-version: "pypy-3.11"
+            os: ubuntu-latest
+            experimental: false
+            nox-session: test-pypy3.11
           - python-version: "3.x"
           # brotli
             os: ubuntu-latest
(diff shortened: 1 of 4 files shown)


sethmlarson (approved):
LGTM, thanks Illia! Really appreciate the thorough documentation 💜

pquentin:
If I'm not mistaken, it looks like `check-latest: true` adds about ~15 seconds of CI time, making our CI ~10% slower. We could consider removing later or just using it for PyPy.

illia-v:
> If I'm not mistaken, it looks like `check-latest: true` adds about ~15 seconds of CI time, making our CI ~10% slower. We could consider removing later or just using it for PyPy.

Yeah, it slows some jobs, but ones containing up-to-date Python in cache (like, [Ubuntu 22.04 (OpenSSL 3.0) 3.13](https://github.com/urllib3/urllib3/actions/runs/13662540569/job/38196761250#logs) for this PR) still setup Python in one second

PR #3570, opened 2025-03-04
Stop using Ubuntu 20.04 and 22.04 in CI

illia-v:
A couple of our actions use Ubuntu 20.04 which is deprecated as a GitHub Actions runner image and will be fully unsupported by April 1 https://github.com/actions/runner-images/issues/11101.
We cannot run tests with Python 3.9.2 and OpenSSL 1.1.1 in newer Ubuntu releases, so I have to drop the two jobs from the matrix.

The rest of the tests are upgraded to `ubuntu-latest` which is Ubuntu 24.04 as of today.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         os:
           - macos-15
           - windows-latest
-          - ubuntu-latest
+          - ubuntu-24.04
         nox-session: ['']
         include:
           - experimental: false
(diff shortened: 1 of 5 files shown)


pquentin (commented):
Thank you for keeping our CI running!

pquentin:
I removed OpenSSL 3.0 from the required checks.

illia-v:
@pquentin thanks for reviewing this, I've addressed your comments

pquentin (approved):
Thanks! LGTM. Not sure why macOS and Windows are reported as missing when we did not change them.

illia-v:
> Not sure why macOS and Windows are reported as missing when we did not change them.

Me too. I can see that some options are duplicated in the branch protection settings, I've just tried setting alternative variants for Windows and macOS, but it doesn't help

Issue #3567, opened 2025-02-27
ReadTimeoutError doesn't pickle / maintain the str parsing


csm10495:
### Subject

When using urllib3 inside of multiprocessing, python often uses pickle to send exceptions and other data from the child to the parent. If ReadTimeoutError is raised inside a multiprocess, it gets pickle, but loses most of its context in route back to the parent process.

### Environment

Describe your environment.
At least, paste here the output of:

```python
>>> import platform
>>> import ssl
>>> import urllib3
>>>
>>> print("OS", platform.platform())
print("Python", platform.python_version())
print(ssl.OPENSSL_VERSION)
print("urllib3", urllib3.__version__)
OS macOS-15.3.1-x86_64-i386-64bit
>>> print("Python", platform.python_version())
Python 3.10.14
>>> print(ssl.OPENSSL_VERSION)
OpenSSL 1.1.1w  11 Sep 2023
>>> print("urllib3", urllib3.__version__)
urllib3 2.3.0
```

### Steps to Reproduce

A simple and isolated way to reproduce the issue. A code snippet would be great.

```
In [1]: import urllib3
   ...: import pickle

In [2]: ex = urllib3.exceptions.ReadTimeoutError('connection pool', 'my url', 'the message with more context')

In [3]: print(ex)
connection pool: the message with more context

In [4]: print(pickle.loads(pickle.dumps(ex)))
None: None
```

### Expected Behavior

I understand not pickling the connection pool but optimally the error message should at least include the originally given message after pickling.

### Actual Behavior

See the steps to reproduce. The string version of the exception winds up being a confusing: `None: None`

csm10495:
To give a tiny bit more context, my app is actually using requests, but it saves the urllib3 ReadTimeoutError then wraps it like so: https://github.com/psf/requests/blob/main/src/requests/adapters.py#L713

illia-v:
FYI, `ex.url` will be pickled
https://github.com/urllib3/urllib3/blob/65d7b5fdc332f9f51aad2f119fae683eb080919d/src/urllib3/exceptions.py#L48-L50

I guess a passed message was not added to the pickled data because it was not retrievable from an exception object directly, but I see no problem in saving the message to `self._message` in `__init__` for this purpose.

csm10495:
Hey @illia-v i was thinking about this some more. (A long time later since it takes that long for my world to pull in new versions). Now I see something like: 

```
E               requests.exceptions.ReadTimeout: None: oh-no!
[...]

PR #3572, opened 2025-03-06
Add some context previously lost during the pickle process

csm10495:
Update tests to try to stimulate the same changes.

ref: #3567

### DIFF
--- a/changelog/3567.feature.rst
+++ b/changelog/3567.feature.rst
@@ -1,0 +1,2 @@
+Updated exceptions to save and restore more properties
+during the pickle/serialization process.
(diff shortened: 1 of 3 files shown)


changelog/3567.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Various updates to exceptions to save and restore more properties
+during the pickle/serialization process.
```
illia-v, line 2: Please format the message according to https://github.com/urllib3/urllib3/tree/main/changelog#readme. And it should be a feature rather than a bugfix in the file name.

```suggestion
Updated exceptions to save and restore more properties
during the pickle/serialization process.
```

illia-v (commented):
@csm10495 thanks for working on this!

illia-v (approved):

PR #3577, opened 2025-03-28
Set VERIFY flags in create_urllib3_context 

### LINKED ISSUE #3571: create_urllib3_context() should set ssl.VERIFY_X509_PARTIAL_CHAIN and ssl.VERIFY_X509_STRICT
### Context

In [Python 3.13](https://docs.python.org/3/whatsnew/3.13.html) ssl.create_default_context() sets ssl.VERIFY_X509_PARTIAL_CHAIN and ssl.VERIFY_X509_STRICT but create_urllib3_context() does not. This causes confusion when trying to connect to an HTTPS server that uses a certificate that is not RFC 5280 compliant. urllib3 succeeds to connect while other ssl python code does not.

create_urllib3_context() should set ssl.VERIFY_X509_PARTIAL_CHAIN and ssl.VERIFY_X509_STRICT when `sys.version_info >= (3, 13)`

### Alternatives

Base create_urllib3_context() on ssl.create_default_context() to avoid discrepancies like this in the future.

### Duplicate

#712 

### Contribution

Yes

RichieB2B:
Set VERIFY_X509_PARTIAL_CHAIN and VERIFY_X509_STRICT in create_urllib3_context for python 3.13+

fixes #3571

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -101,6 +101,8 @@ try:  # Do we have ssl at all?
         OPENSSL_VERSION_NUMBER,
         PROTOCOL_TLS,
         PROTOCOL_TLS_CLIENT,
+        VERIFY_X509_PARTIAL_CHAIN,
+        VERIFY_X509_STRICT,
         OP_NO_SSLv2,
         OP_NO_SSLv3,
         SSLContext,
@@ -138,6 +140,8 @@ except ImportError:
     OP_NO_SSLv3 = 0x2000000  # type: ignore[assignment]
     PROTOCOL_SSLv23 = PROTOCOL_TLS = 2  # type: ignore[assignment]
     PROTOCOL_TLS_CLIENT = 16  # type: ignore[assignment]
+    VERIFY_X509_PARTIAL_CHAIN = 0x80000  # type: ignore[assignment]
+    VERIFY_X509_STRICT = 0x20  # type: ignore[assignment]
 
 
 _TYPE_PEER_CERT_RET = typing.Union["_TYPE_PEER_CERT_RET_DICT", bytes, None]
@@ -223,6 +227,7 @@ def create_urllib3_context(
     ciphers: str | None = None,
     ssl_minimum_version: int | None = None,
     ssl_maximum_version: int | None = None,
+    verify_flags: int | None = None,
 ) -> ssl.SSLContext:
     """Creates and configures an :class:`ssl.SSLContext` instance for use with urllib3.
 
@@ -247,6 +252,9 @@ def create_urllib3_context(
     :param ciphers:
         Which cipher suites to allow the server to select. Defaults to either system configured
         ciphers if OpenSSL 1.1.1+, otherwise uses a secure default set of ciphers.
+    :param verify_flags:
+        The flags for certificate verification operations. These default to
+        ``ssl.VERIFY_X509_PARTIAL_CHAIN`` and ``ssl.VERIFY_X509_STRICT`` for python 3.13+.
     :returns:
         Constructed SSLContext object with specified options
     :rtype: SSLContext
@@ -319,6 +327,16 @@ def create_urllib3_context(
         options |= OP_NO_TICKET
 
     context.options |= options
+
+    if verify_flags is None:
+        verify_flags = 0
+        # In python 3.13+ ssl.create_default_context() sets VERIFY_X509_PARTIAL_CHAIN
+        # and VERIFY_X509_STRICT so we do the same
+        if sys.version_info >= (3, 13):
+            verify_flags |= VERIFY_X509_PARTIAL_CHAIN
+            verify_flags |= VERIFY_X509_STRICT
+
+    context.verify_flags |= verify_flags
 
     # Enable post-handshake authentication for TLS 1.3, see GH #1634. PHA is
     # necessary for conditional client cert authentication with TLS 1.3.
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -101,7 +101,6 @@ try:  # Do we have ssl at all?
         OPENSSL_VERSION_NUMBER,
         PROTOCOL_TLS,
         PROTOCOL_TLS_CLIENT,
-        VERIFY_X509_PARTIAL_CHAIN,
         VERIFY_X509_STRICT,
         OP_NO_SSLv2,
         OP_NO_SSLv3,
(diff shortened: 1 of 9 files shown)


src/urllib3/util/ssl_.py
```diff
@@ -317,6 +321,11 @@ def create_urllib3_context(
         # there is a risk associated with it being on wire,
         # if the server is not rotating its ticketing keys properly.
         options |= OP_NO_TICKET
+        # In python 3.13+ ssl.create_default_context() sets VERIFY_X509_PARTIAL_CHAIN
+        # and VERIFY_X509_STRICT so we do the same
+        if sys.version_info >= (3, 13):
+            options |= VERIFY_X509_PARTIAL_CHAIN
+            options |= VERIFY_X509_STRICT
```
illia-v, line 328: I believe the flags should be set as [`SSLContext.verify_flags`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.verify_flags), not as options.
I suggest adding a `verify_flags` parameter to `create_urllib3_context` too.

test/test_ssl.py
```diff
@@ -44,6 +45,15 @@ def test_is_ipaddress_true(self, addr: bytes | str) -> None:
...
         assert not ssl_.is_ipaddress(addr)
 
+    def test_create_urllib3_context_options(self):
+        context = ssl_.create_urllib3_context()
+        if sys.version_info >= (3, 13):
+            assert context.options & ssl.VERIFY_X509_PARTIAL_CHAIN
+            assert context.options & ssl.VERIFY_X509_STRICT
+        else:
+            assert context.options | ssl.VERIFY_X509_PARTIAL_CHAIN
+            assert context.options | ssl.VERIFY_X509_STRICT
```
Rogdham, line 55: I think these assert don't check what you indented

```suggestion
            assert not (context.options & ssl.VERIFY_X509_PARTIAL_CHAIN)
            assert not (context.options & ssl.VERIFY_X509_STRICT)
```

RichieB2B: Indeed, thanks!

src/urllib3/util/ssl_.py
RichieB2B: Thanks, I knew something was "off" with that first attempt. :-)

src/urllib3/util/ssl_.py
```diff
@@ -320,6 +328,16 @@ def create_urllib3_context(
...
 
+    if verify_flags is None:
+        verify_flags = 0
+        # In python 3.13+ ssl.create_default_context() sets VERIFY_X509_PARTIAL_CHAIN
+        # and VERIFY_X509_STRICT so we do the same
+        if sys.version_info >= (3, 13):
+            verify_flags |= VERIFY_X509_PARTIAL_CHAIN
+            verify_flags |= VERIFY_X509_STRICT
+
+    context.verify_flags |= verify_flags
```
illia-v, line 339: I can see errors like `AttributeError: 'PyOpenSSLContext' object has no attribute 'verify_flags'` in CI. urllib3 [supports](https://urllib3.readthedocs.io/en/stable/reference/contrib/pyopenssl.html) PyOpenSSL as an alternative TLS backend.

To fix the error, you can take an approach with a setter and a getter similarly to how `options` are managed. 
https://github.com/urllib3/urllib3/blob/3e8f2db735dcaced6a3b777aa1966f40c018af7c/src/urllib3/contrib/pyopenssl.py#L428-L435

It looks like PyOpenSSL [provides](https://stackoverflow.com/a/64968483) a way to set the flags, but I see no public method returning set flags.

test/test_ssl.py
```diff
@@ -63,6 +64,21 @@ def test_create_urllib3_no_context(self) -> None:
...
                 ssl_.create_urllib3_context()
 
+    def test_create_urllib3_context_default_verify_flags(self):
+        context = ssl_.create_urllib3_context()
+        if sys.version_info >= (3, 13):
+            assert context.verify_flags & ssl.VERIFY_X509_PARTIAL_CHAIN
+            assert context.verify_flags & ssl.VERIFY_X509_STRICT
+        else:
+            assert not (context.verify_flags & ssl.VERIFY_X509_PARTIAL_CHAIN)
+            assert not (context.verify_flags & ssl.VERIFY_X509_STRICT)
```
illia-v, line 74: Since we are setting the same flags as Python does, please consider taking an approach similar to `test_default_ssl_context_ssl_min_max_versions` where `verify_flags` value of a context created by Python is compared to one of the urllib3's context. This way we can detect future mismatches earlier.

https://github.com/urllib3/urllib3/blob/3e8f2db735dcaced6a3b777aa1966f40c018af7c/test/with_dummyserver/test_https.py#L1109-L1124

src/urllib3/util/ssl_.py
illia-v, line 339: It seems passing `verify_flags` to `create_urllib3_context` provides no way to set `verify_flags` to 0 because `|=` is used here, right?

illia-v (changes requested):

src/urllib3/util/ssl_.py
RichieB2B: You're right. I took the same approach as `options`. There is a default provided by `SSLContext(PROTOCOL_TLS_CLIENT)` (which is to set `VERIFY_X509_TRUSTED_FIRST`) and we just add flags to it. If a user want to clear the default (unset `VERIFY_X509_TRUSTED_FIRST`) this can only be done by excplicitly setting `verify_flags` after calling `create_urllib3_context`. The same goes for `options`.

test/test_ssl.py
RichieB2B: I took this simple approach which I think will fail for PYOPENSSL? If so, I'm not sure how to fix that.
```
    def test_default_ssl_context_verify_flags(self) -> None:
        ctx = urllib3.util.ssl_.create_urllib3_context()
        ssl_ctx = ssl.create_default_context()
        assert ctx.verify_flags == ssl_ctx.verify_flags
```

src/urllib3/util/ssl_.py
RichieB2B: Do we really need to know (get) the default `verify_flags`? The `options` setter does not bother with this either and just applied the options that are set.

RichieB2B:
Note that this change has downstream effects on `requests`:
https://github.com/urllib3/urllib3/actions/runs/14144861765/job/39630951821

pquentin:
> Note that this change has downstream effects on `requests`: https://github.com/urllib3/urllib3/actions/runs/14144861765/job/39630951821

@sigmavirus24 the certificates introduced in https://github.com/psf/requests/pull/6662 do not pass with `VERIFY_X509_STRICT`. Would it be possible to fix them? An option is to use [trustme](https://trustme.readthedocs.io/en/latest/) (either at runtime or in the Makefile), there's prior art in https://github.com/psf/requests/blob/1764cc938efc3cc9720188dfa6c3852c45211aa0/tests/conftest.py#L37-L46

test/test_ssl.py
RichieB2B: It works for PYOPENSSL when using `ssl.VERIFY_X509_TRUSTED_FIRST` as a default for `verify_flags`.

illia-v (approved):

illia-v:
@pquentin +1 for using trustme for the certificate!

I added a patch from https://github.com/psf/requests/pull/6924 for now to fix the downstream test.

illia-v:
@RichieB2B @Rogdham thanks for working on this!

PR #3589, opened 2025-04-10
Upgrade memray and coverage to fix macOS tests

illia-v:
Upgrading memray seems to fix CI failures which started with macOS 15.4 yesterday.


sethmlarson (commented):
LGTM, thanks Illia.

PR #3593, opened 2025-04-11
Initial WASI/componentize-py support

### LINKED ISSUE #3584: Add support for WASI
### Context

I want to use urllib3 in server-side WebAssembly components using componentize-py.
The implementation plan would be to use the same approach as the emscripten integration and check whether `sys.platform == "wasi"` and if yes, use a custom implementation based on wasi-http host functions.

### Alternatives

Using the library directly with componentize-py fails due to various native libraries missing https://github.com/bytecodealliance/componentize-py/issues/96.

### Duplicate

no

### Contribution

yes

mschuwalow:
resolves #3584 

This pr adds initial support for using urllib3 in a wasi / componentize-py environment.

This is achieved using a similar setup as the emscripten support, by monkeypatching the connection classes when we are in a wasi environment.

The support is not perfect, some things that are normally done by the library such as transfer encoding are done by wasi-http / the host now, so the options will have no effect. It should still lower the barrier of entry for users to use python with wasi significantly, so I think there is value here.


src/urllib3/contrib/wasi/wasi.py
```diff
@@ -0,0 +1,159 @@
...
+
+C_UINT64_MAX = 18_446_744_073_709_551_615
+
+def preload(world_name: str) -> None:
+    importlib.import_module(".imports.outgoing_handler", world_name)
+    importlib.import_module(".imports.types", world_name)
+    importlib.import_module(".imports.streams", world_name)
+    importlib.import_module(".types", world_name)
+
+def send_request(request: WasiRequest, world_name: str) -> WasiResponse:
```
mschuwalow, line 16: This being untyped is a bit annoying, but I don't think there is a good workaround currently.

Hopefully componentize-py eventually allows generating the bindings in a static location so we can have type hints here

src/urllib3/response.py
```diff
@@ -9,7 +9,10 @@
 import sys
 import typing
 import warnings
-import zlib
+try:
+    import zlib
+except ImportError:
```
mschuwalow, line 14: zlib is not currently available in wasi, so I guarded the import for now.
For all other systems there should not be a difference

mschuwalow:
I'll update this to the latest componentize-py version once https://github.com/bytecodealliance/componentize-py/pull/150 is merged.

This will get rid of the `enable_wasi_backend` function / dynamic imports and will allow to have types in the wasi interaction code.

src/wit_world/__init__.py
```diff
@@ -0,0 +1,13 @@
+from typing import TypeVar, Generic, Union, Optional, Protocol, Tuple, List, Any, Self
```
mschuwalow, line 1: This is generated by the wasi-tools/regenerate_bindings.sh script and excluded from packaging. At runtime this will be provided by componentize-py on wasi

pyproject.toml
```diff
@@ -129,6 +133,9 @@ include = [
   "/README.md",
   "/LICENSE.txt",
 ]
+exclude = [
```
mschuwalow, line 136: So the generated code is not packaged

wasi-tools/regenerate_bindings.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+
+rm -rf src/wit_world
```
mschuwalow, line 3: This generates the code in src/wit_world and needs to be rerun when the code in the wasi-tools/wit directory changes

wasi-tools/wit/deps.toml
```diff
@@ -0,0 +1,4 @@
+io = "https://github.com/WebAssembly/wasi-io/archive/v0.2.0.tar.gz"
```
mschuwalow, line 1: The wit files are fetched using the [wit-deps](https://github.com/bytecodealliance/wit-deps) tool.
Everything in wasi-tools/wit/deps/* is generated by it.

rajsite:
@mschuwalow I'm really interested in this support, are there any substantial blockers left?

mschuwalow:
> @mschuwalow I'm really interested in this support, are there any substantial blockers left?

As far as I'm aware this should be working with the latest componentize-py without any known issues (though it was only use for a few smaller demos so far). Will rebase this in the next few days

rajsite:
@mschuwalow out of curiosity, once this merges do you think there would be any changes needed for [requests](https://github.com/psf/requests) to pick it up (my understanding is it's built on urllib3)?

mschuwalow:
> @mschuwalow out of curiosity, once this merges do you think there would be any changes needed for [requests](https://github.com/psf/requests) to pick it up (my understanding is it's built on urllib3)?

I gave requests a very brief try, and it seems to work with the current urllib3 changes as they are. Might need some smaller followups eventually though

crepererum:
> > @mschuwalow out of curiosity, once this merges do you think there would be any changes needed for [requests](https://github.com/psf/requests) to pick it up (my understanding is it's built on urllib3)?
> 
> I gave requests a very brief try, and it seems to work with the current urllib3 changes as they are. Might need some smaller followups eventually though

I've tested it and it seems the response body is always empty (not in `urllib3`, only in combination w/ `requests`). I've tried to fix this here: https://github.com/golemcloud/urllib3/pull/2 .

crepererum:
In case someone else wants to use this: We at InfluxData are using this PR in https://github.com/influxdata/datafusion-udf-wasm . I've rebased it to the latest `urrlib3` release and we are providing a rebased version of this PR, a few minor fixes (which I've also filed as PRs against the golemcloud PR branch), and a wheel [here](https://github.com/influxdata/urllib3/pull/1).

PR #3596, opened 2025-04-19
Delete ci/0001-requests-test-cert-key-usage.patch

illia-v:
The patch is no longer needed because https://github.com/psf/requests/pull/6924 was merged. Applying the patch fails the downstream tests now https://github.com/urllib3/urllib3/actions/runs/14551752825/job/40823765843


illia-v:
The CI error suggests that tests found enabled JSPI in Chrome unexpectedly, it may be because of a Chrome upgrade and should be a separate issue

sigmavirus24 (approved):

sethmlarson (approved):

Issue #3598, opened 2025-04-19
JSPI shipping in Chrome 137 breaks CI


illia-v:
emscripten(chrome) CI jobs have recently started to fail:
https://github.com/urllib3/urllib3/actions/runs/14552169312/job/40824650952
```
FAILED test/contrib/emscripten/test_emscripten.py::test_index[chrome-False] - AssertionError: assert True == False
 +  where True = <function has_jspi at 0x1d3af70>()
 +    where <function has_jspi at 0x1d3af70> = <module 'urllib3.contrib.emscripten.fetch' from '/lib/python3.12/site-packages/urllib3/contrib/emscripten/fetch.py'>.has_jspi
 +      where <module 'urllib3.contrib.emscripten.fetch' from '/lib/python3.12/site-packages/urllib3/contrib/emscripten/fetch.py'> = <module 'urllib3.contrib.emscripten' from '/lib/python3.12/site-packages/urllib3/contrib/emscripten/__init__.py'>.fetch
 +        where <module 'urllib3.contrib.emscripten' from '/lib/python3.12/site-packages/urllib3/contrib/emscripten/__init__.py'> = <module 'urllib3.contrib' from '/lib/python3.12/site-packages/urllib3/contrib/__init__.py'>.emscripten
 +          where <module 'urllib3.contrib' from '/lib/python3.12/site-packages/urllib3/contrib/__init__.py'> = <module 'urllib3' from '/lib/python3.12/site-packages/urllib3/__init__.py'>.contrib
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
```

It's likely to be caused by Chrome 137 shipping JavaScript Promise Integration https://chromestatus.com/feature/5674874568704000.

cc @joemarshall

joemarshall:
Yep, because JSPI is no longer behind a flag, it isn't possible to start a browser without support for it. I've added a flag to disable the JSPI code paths, so that we can test non-jspi code still works, and made a PR. This needs to happen, because there are some circumstances in webassembly where JSPI is not available even on JSPI supporting browsers.

Works on my machine, Just waiting for tests to run on github actions for the PR.

illia-v:
@joemarshall thanks for looking into this. Can we have two jobs to test with Chrome 136 and the current one (137 right now) without forcing non-default behavior?

joemarshall:
[...]

PR #3599, opened 2025-04-22
Fix emscripten CI for Chrome 137+

joemarshall:
Fix to the CI errors occurring in Chrome 137 or later, because JSPI is on by default in those browsers.

https://github.com/urllib3/urllib3/issues/3598


joemarshall:
@illia-v I think this is ready to go now I got rid of the silly test code in fetch.py

illia-v:
@joemarshall thanks for the efforts to fix the CI!
I've reviewed the changes and understand that code of urllib3 (other than tests) is not affected by JSPI becoming default in Chrome. I found some places getting confusing now when an environment can support both JSPI and XHR by default, and I need to spend more time on this next week. Let's use Chrome stable in the meantime

illia-v:
On second thought, I'm ready to merge this with a couple of minor changes. There is still some refactoring I'd like to do on top of this, but it can be done separately

illia-v:
> I don't understand why we need `pyodide.ffi.can_run_sync = lambda: False` twice, but it does fix CI.

I believe one patches tests using web workers and another for tests running in the main thread

PR #3600, opened 2025-04-24
Use stable Chrome in CI for now

illia-v:
Chrome Canary has enabled JSPI by default which breaks our CI #3598

There is a pull request with a possible fix #3599: only tests are changed, package code stays the same.
But I'm not ready to merge it, I want to consider refactoring the tests a bit next week and use the stable Chrome channel in the meantime to get tests passing.


pquentin (approved):
Thanks! LGTM.

PR #3613, opened 2025-05-29
Allow hatch-vcs 0.5 (fix #3612)

### LINKED ISSUE #3612: Support hatch-vcs 0.5.0
### Subject

Build fails with 0.5.0 because it is pinned to 0.4.0

```
        [--:--:--] START   components/python3-urllib3.bst: Running commands
            python -P -mbuild --no-isolation --wheel --outdir /buildstream-build/dist .
            python -P -minstaller /buildstream-build/dist/*.whl --destdir /buildstream-install
        + sh -c -e 'python -P -mbuild --no-isolation --wheel --outdir /buildstream-build/dist .
        '
        * Getting build dependencies for wheel...
        ERROR Missing dependencies:
        	hatch-vcs==0.4.0
        Command 'python -P -mbuild --no-isolation --wheel --outdir /buildstream-build/dist .' failed with exitcode 1
```

### Environment

N/A

### Steps to Reproduce

Build with hatch-vcs 0.5.0

### Expected Behavior

Builds

### Actual Behavior

Does not build

musicinmybrain:
Keep SemVer-bounding `hatch-vcs` as agreed in https://github.com/urllib3/urllib3/pull/3412#discussion_r1644414498, but stop pinning it all the way to the patch release. Fixes https://github.com/urllib3/urllib3/issues/3612.

I’m happy to simply expand this version bound for now, although I would like to suggest that it might not really be necessary to upper-bound the version of `hatch-vcs` after all. There are 79 packages in Fedora that have `hatch-vcs` as a build dependency, and `python-urllib3` is the *only one* that upper-bounds the version of `hatch-vcs`. Furthermore, we found no regressions in any of those packages when we updated from 0.4.0 to 0.5.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 # This file is protected via CODEOWNERS
 
 [build-system]
-requires = ["hatchling>=1.6.0,<2", "hatch-vcs>=0.4.0,<0.6.0"]
+requires = ["hatchling>=1.6.0,<2", "hatch-vcs>=0.4.0,<0.6.0", "setuptools-scm>=8,<9"]
 build-backend = "hatchling.build"
 
 [project]


pquentin (approved):
Thank you! LGTM.

Stories like https://lwn.net/Articles/1020576/ make me reticent to unpin completely, especially since hatch-vcs did not reach 1.0 yet. I've added a commit to pin setuptools-scm, for consistency, as it's way more active than hatch-vcs itself.

musicinmybrain:
> Stories like https://lwn.net/Articles/1020576/ make me reticent to unpin completely, especially since hatch-vcs did not reach 1.0 yet. I've added a commit to pin setuptools-scm, for consistency, as it's way more active than hatch-vcs itself.

Fair enough. Downstream in Fedora, we will (necessarily) keep removing the pins and letting you know whether anything catches fire when we do so. 😄

PR #3621, opened 2025-06-16
Update v2 migration guide to 2.4.0

pquentin:
In preparation of the next release, I've looked at the migration guide which was mostly referring to 2.0 as something that would happen in the future, but it has been released 2 years ago now. Accordingly:

* I now call it 2.x instead of v2.0 to clarify that it's real.
* I updated the important changes to be accurate as of the 2.4.0 release, including the fact that 2.x supports 3.9+
* I reordered the "Common upgrading issues" section to include the actual issues first, then application migration guidance, and package maintenance last. (At this point, libraries have moved on, but applications.)
* I clarified that 1.26.x does not currently receive security fixes.

### DIFF
--- a/docs/v2-migration-guide.rst
+++ b/docs/v2-migration-guide.rst
@@ -1,9 +1,9 @@
 v2.0 Migration Guide
 ====================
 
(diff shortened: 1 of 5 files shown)


illia-v (approved):
Thanks!

PR #3622, opened 2025-06-16
Remove getheader and getheaders methods

pquentin:


sigmavirus24 (dismissed):

sethmlarson (approved):
Thanks Quentin!

pquentin:
Thanks for the reviews! I'll wait a few days to ensure we won't need a 2.5.1 release before merging this. (I also have a conflict to fix.)

PR #3630, opened 2025-06-21
Move Illia Volochii to the top of maintainers

sethmlarson:

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,9 +85,9 @@ Tidelift will coordinate the fix and disclosure with maintainers.
 
 ## Maintainers
 
+- [@illia-v](https://github.com/illia-v) (Illia Volochii)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,9 +85,9 @@ Tidelift will coordinate the fix and disclosure with maintainers.
 
 ## Maintainers
 
+- [@illia-v](https://github.com/illia-v) (Illia Volochii)
```
shazow: Do we want to prefix with `Lead:` or a :star: or something to be more informative for people landing on this list?

To me primarily it was a historically-ordered list of the "urllib3 maintainer family/alumni", but for others it may be a "who is the right person to reach out to?" list, so indicating the current lead may be useful.

sethmlarson: Great idea, I'll use the `Lead:` prefix here.

Issue #3632, opened 2025-06-25
(Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1006)'))). Getting this error when using an HTTPS proxy with urllib3 version 2.5.0


JatinMaheriya:
### Subject
**HTTPS Proxy Fails with SSL Handshake Error in urllib3 v2.5.0 (Works Fine in v1.26.20)**
* I'm using the urllib3 library as a dependency of the requests package. After upgrading to urllib3 version 2.5.0, I encounter the following SSL error when attempting to use an HTTPS proxy:
**Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1006)'))**
* However, the HTTP proxy works correctly with urllib3 v2.5.0.
* Previously, with urllib3 version 1.26.20, both HTTP and HTTPS proxies worked without any issues.

### Environment

* Python: 3.11.5
* urllib3: 2.5.0
* requests: 2.31.0
* OpenSSL: 1.1.1k FIPS  25 Mar 2021
* **OpenSSL (proxy): 1.0.2k-fips  26 Jan 2017**

### Steps to Reproduce

1. Set up the environment as described above.
2. Replace the proxy and URL placeholders in the code snippet below with valid values.
3. Run the following script:
```
import requests

proxies = {
    "http":  "http://username:password@proxy_ip:proxy_port",
    "https": "http://username:password@proxy_ip:proxy_port"
}

url = "https://example.com"

try:
    response = requests.get(url, proxies=proxies)
    print("Status Code:", response.status_code)
    print("Response Body:", response.text)
except Exception as e:
    print("An error occurred:", e)
```

### Expected Behavior

The request should complete successfully when using an HTTPS proxy with the urllib3 v2.5.0, just like it does in urllib3 v1.26.20.

### Actual Behavior

The request fails with the following SSL handshake error:

**(Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1006)')))** 

### Additional Notes
* Downgrading back to urllib3==1.26.20 resolves the issue.

illia-v:
[...]

PR #3633, opened 2025-06-28
urllib3 v2.5.0 HTTPS Proxy SSL Handshake Failure

akshat62:
https://github.com/urllib3/urllib3/issues/3632

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -868,7 +868,7 @@ class HTTPSConnection(HTTPConnection):
             from .util.ssl_ import create_urllib3_context
 
             ssl_context = create_urllib3_context(
-                ssl_version=self.ssl_version,
+                ssl_version=resolve_ssl_version(self.ssl_version),
                 ssl_minimum_version=self.ssl_minimum_version,
                 ssl_maximum_version=self.ssl_maximum_version,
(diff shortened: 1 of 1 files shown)


illia-v:
Hi @akshat62,

Thanks for your contribution to urllib3!

To help me better understand the situation, could you confirm whether your environment is the same as described in #3632, specifically whether you use Python 3.11.5? I believe `create_urllib3_context` sets `verify_flags=0` by default for Python < 3.13.

Regarding the proposed change, I'm a bit hesitant to merge it. The issue seems to stem from a specific proxy server configuration, and making the default context less secure for all users could have broader security implications.

A more targeted solution would be to use a custom context for your proxy via the `proxy_ssl_context` parameter. This would allow you to use a specific SSL context that works for your environment without changing the default behavior for everyone else.

akshat62:
> whether you use Python 3.11.5?

Yes! Using Python 3.11.5

illia-v:
> Yes! Using Python 3.11.5

`create_urllib3_context` sets `verify_flags` to 0 by default for the Python version
https://github.com/urllib3/urllib3/blob/98a60f9c26a4d64cf18c7e004b9adda4303bc28c/src/urllib3/util/ssl_.py#L333-L341

Maybe your issue is unrelated to `verify_flags`

PR #3644, opened 2025-07-18
Prepare Emscripten tests for Chrome 139

illia-v:
This pull request addresses the upcoming change in Chrome 139, which we've already seen in failing CI jobs. It migrates our Emscripten tests to exclusively use the `selenium` fixture, rather than a combination of `selenium` and `selenium_jspi`. Thanks @hoodmane for the suggestion in https://github.com/pyodide/pytest-pyodide/issues/163#issuecomment-3088631178.

Initial attempts at this switch revealed underlying issues. Consequently, this branch evolved to include a series of necessary fixes and related refactorings. Please review the individual commits for detailed changes, I'm describing a few of the more complex ones below.

* ad2dfb53c51aa947c0dbade7497d2db5fab6a0cd Stop using fixtures `selenium_coverage` and `run_from_server` together
This commit stops the combined use of `selenium_coverage` and `run_from_server` fixtures.
Previously, `run_from_server` depended on `selenium_coverage`, and some tests unnecessarily requested both. This combination created conflicts with upcoming monkey patching cleanup. Decoupling them resolves this interference.

* 5f89c1874c1e98d0000ebec1d1903d78643c65ae Share the same code monkey patching JSPI availability
This commit extracts and centralizes the JSPI monkey patching logic into a shared function, now including a crucial cleanup step.
Previously, separate Selenium fixtures for JSPI and non-JSPI paths in Chrome meant no cleanup was needed after monkey patching. With the switch to a single Chrome fixture, any subsequent JSPI-dependent tests would fail if the monkey patch persisted. The added cleanup resolves this, preventing test failures and eliminating code duplication.

* 8316ef0a4204dc179b5f1c5b1a0074f52ce1bed0 Rename `has_jspi` fixture to `prefer_jspi`
The original `has_jspi` fixture provides a boolean value indicating whether a test should use JSPI or XHR. A `False` value doesn't necessarily mean the runtime lacks JSPI support. For instance, the same test against Chrome can have two versions, `test_index[chrome-False]` and `test_index[chrome-True]`, where the boolean reflects `has_jspi`. This has been confusing.
Renaming it to `prefer_jspi` will clarify the fixture's meaning, as it's more easily understood as a characteristic of the test method rather than the runtime itself.

* 22e25c4119ef5585e9689066b02a7143593956dc Deselect and parametrize Emscripten tests properly
This commit refactors Emscripten test management by separating parametrization and deselection into distinct pytest hooks.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,7 @@ jobs:
         if: ${{ matrix.nox-session == 'emscripten(chrome)' }}
         with:
           install-chromedriver: true
+          chrome-version: canary
       - name: Force override system chrome
         run: |
           sudo rm -f /usr/bin/google-chrome
--- a/noxfile.py
+++ b/noxfile.py
@@ -273,7 +273,6 @@ def emscripten(session: nox.Session, runner: str) -> None:
         session,
         extras="",
         pytest_extra_args=[
-            "-x",
             "--runtime",
             f"{runner}-no-host",
             "--dist-dir",
--- a/noxfile.py
+++ b/noxfile.py
@@ -239,7 +239,7 @@ def emscripten(session: nox.Session, runner: str) -> None:
         dist_dir = Path(os.environ["PYODIDE_ROOT"]) / "dist"
     else:
         # we don't have a build tree
-        pyodide_version = "0.27.1"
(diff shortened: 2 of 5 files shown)


pquentin (approved):
Thanks! LGTM.

Issue #3636, opened 2025-07-01
Partial then full read with `decode_content=True` returns partial/truncated data


Grub4K:
### Subject

When using a partial read (`.read(512)`) followed by `.read()`, when `decode_content=True`, only partial content is returned.

### Environment

This should be unrelated to environment.

```
OS Linux-6.8.0-62-generic-x86_64-with-glibc2.39
Python 3.13.0
OpenSSL 3.0.15 3 Sep 2024
urllib3 2.5.0
```

### Steps to Reproduce
Relevant code:
```py
    with urllib3.PoolManager() as pool:
        with pool.request(
            "GET",
            f"http://127.0.0.1:12345{path}",
            redirect=True,
            decode_content=True,
            preload_content=False,
        ) as response:
            prefix = response.read(512)
            remainder = response.read()
```
here `prefix + remainder != server_data`, if the data has a `Content-Encoding`

<details><summary>full reproduction script</summary>

```py
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "urllib3[zstd]==2.5.0",
#     "zstd",
# ]
# ///
import gzip
import http.server
import string
import threading
import zstd

import urllib3


data = 100 * string.printable.encode()


class HTTPTestRequestHandler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path == "/":
            self.send_response(200)
            self.send_header("Content-Length", str(len(data)))
            self.end_headers()
            self.wfile.write(data)
        elif self.path == "/gzip":
            cdata = gzip.compress(data)
            self.send_response(200)
            self.send_header("Content-Encoding", "gzip")
            self.send_header("Content-Length", str(len(cdata)))
            self.end_headers()
            self.wfile.write(cdata)
        elif self.path == "/zstd":
            cdata = zstd.compress(data)
            self.send_response(200)
            self.send_header("Content-Encoding", "zstd")
            self.send_header("Content-Length", str(len(cdata)))
            self.end_headers()
            self.wfile.write(cdata)
        else:
            self.send_response(404)
            self.send_header("Content-Length", "0")
            self.end_headers()


httpd = http.server.ThreadingHTTPServer(("127.0.0.1", 12345), HTTPTestRequestHandler)
thread = threading.Thread(target=httpd.serve_forever)
thread.daemon = True
thread.start()


def test_equal(path: str):
    with urllib3.PoolManager() as pool:
        with pool.request(
            "GET",
[...]

PR #3647, opened 2025-07-31
Fix truncation on mixed read with decode_content=True

alexprabhat99:
Fixes: #3636

### DIFF
--- a/test/test_response.py
+++ b/test/test_response.py
@@ -1560,6 +1560,7 @@ class TestResponse:
             resp.read(16)
         assert "Response may not contain content" in str(excinfo.value)
 
+    @onlyZstd()
     def test_partial_then_full_read_with_decode_content(self) -> None:
         data = 100 * string.printable.encode()
         cdata = zstd_compress(data)


alexprabhat99:
Hi @illia-v 
I'm not sure if the failing check is related to the fix introduced in this PR. Could you please take a look and help verify if it's related? Appreciate your review whenever you get a chance. Thanks!

src/urllib3/response.py
```diff
@@ -986,6 +986,9 @@ def read(
 
         if amt is None:
             data = self._decode(data, decode_content, flush_decoder)
+            if len(self._decoded_buffer) != 0:
+                data = self._decoded_buffer.get_all() + data
```
illia-v, line 990: Let's rely on joining capability of the buffer which may be more efficient, and have a small optimization for the case when `data` is empty

```suggestion
                if data:
                    self._decoded_buffer.put(data)
                data = self._decoded_buffer.get_all()
```

illia-v, line 990: The buffer should stay untouched if `decode_content` is `False`, so the check from lines 996–1001 should be used regardless of the `amt` value

src/urllib3/response.py
```diff
@@ -986,6 +986,9 @@ def read(
...
+
             if cache_content:
```
illia-v, line 992: Perhaps we need to set `cache_content = False` when `read(amt=None, cache_content=True)` follows a partial read

illia-v (changes requested):
@alexprabhat99 thanks for the patch!

The failure happened because of a flaky test, I rerun it and CI is green now

src/urllib3/response.py
Grub4K: That won't quite work. The problem I faced when implementing this is that there are several edge-cases that use read but assume that `read(None)` will always return a complete response. An example where the suggested strategy will fail when not also additionally modified:
```py
resp.read(512)
print(resp.data)
```
The second one will be missing the first 512 bytes since they are not cached if we just silently ignore this.

It might be better to have something like `self._has_uncached_read` similar to `self._has_decoded_content` and `raise` instead of silently returning partial data?

illia-v:
Closing this because #4960 is more complete

PR #3648, opened 2025-08-04
security: update actions for artifact handling only by `id`

GrantBirki:
This pull request updates the `publish.yml` workflow to explicitly use `artifact-ids` instead of the `name` input option when handling artifacts. This is important because artifacts produced by GitHub Actions can be completely overwritten by other workflow runs if they use the same `name`. To avoid potential TOCTOU issues/vulnerabilities where an artifact might be replaced between upload and download, the new `artifact-ids` input allows you to download artifacts by their specific ID rather than by name.

In the context of this repo, this is especially important as the SLSA L3 badge is present on this project, but in order to be truly at SLSA L3, you cannot use `name` for artifacts as it could be replaced between the `build` and `release...` stages of this workflow. This means that another workflow could potentially change the artifact being used by this `publish.yml` job and release a malicious package to `pypi.org`.

---

The pull request where I originally implemented this feature in `actions/download-artifact` does a pretty good job at summarizing why this is important in the context of security in general: https://github.com/actions/download-artifact/pull/401

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,6 +20,7 @@ jobs:
     runs-on: "ubuntu-latest"
     outputs:
       hashes: ${{ steps.hash.outputs.hashes }}
+      artifact-id: ${{ steps.upload-artifact.outputs.artifact-id }}
 
     steps:
       - name: "Checkout repository"
@@ -47,7 +48,8 @@ jobs:
           cd dist && echo "hashes=$(sha256sum * | base64 -w0)" >> $GITHUB_OUTPUT
 
       - name: "Upload dists"
-        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+        uses: actions/upload-artifact@4.6.2 # immutable release
+        id: upload-artifact
         with:
           name: "dist"
           path: "dist/"
@@ -67,9 +69,9 @@ jobs:
 
     steps:
     - name: "Download dists"
-      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+      uses: actions/download-artifact@4.3.0 # immutable release
       with:
-        name: "dist"
+        artifact-ids: ${{ needs.build.outputs.artifact-id }}
         path: "dist/"
 
     - name: "Upload dists to a new GitHub Release Draft"
@@ -100,9 +102,9 @@ jobs:
 
     steps:
     - name: "Download dists"
-      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+      uses: actions/download-artifact@4.3.0 # immutable release
       with:
-        name: "dist"
+        artifact-ids: ${{ needs.build.outputs.artifact-id }}
         path: "dist/"
 
     - name: "Publish dists to Test PyPI"


GrantBirki:
> If your change does not deserve a changelog entry, apply the Skip Changelog GitHub label to your pull request

I do not have the ability to add labels to this PR, but I do not think this change needs a changelog entry so this label should be added.

.github/workflows/publish.yml
```diff
@@ -47,7 +48,8 @@ jobs:
           cd dist && echo "hashes=$(sha256sum * | base64 -w0)" >> $GITHUB_OUTPUT
 
       - name: "Upload dists"
-        uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+        uses: actions/upload-artifact@4.6.2 # immutable release
```
pquentin: Can you please stick to the existing scheme? There may be an argument to switch back to version numbers, but I'd rather do this separately, for all actions, and make sure that the format will be recognized by dependabot.

pquentin (commented):
Good idea, thank you!

.github/workflows/publish.yml
GrantBirki:
```suggestion
        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
```

.github/workflows/publish.yml
```diff
@@ -67,9 +69,9 @@ jobs:
 
     steps:
     - name: "Download dists"
-      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+      uses: actions/download-artifact@4.3.0 # immutable release
```
GrantBirki:
```suggestion
      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
```

.github/workflows/publish.yml
```diff
@@ -100,9 +102,9 @@ jobs:
...
```
GrantBirki:
```suggestion
      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
```

.github/workflows/publish.yml
GrantBirki: @pquentin done! ✅

pquentin (approved):
Thanks! LGTM.

We can address ci.yml later since it's not part of the publishing build and does not impact SLSA levels.

pquentin:
@GrantBirki This didn't work: https://github.com/urllib3/urllib3/actions/runs/16756161327/job/47439256407

Here's what a working run looks like: https://github.com/urllib3/urllib3/actions/runs/16593313383/job/46934035410

.github/workflows/publish.yml
```diff
@@ -100,9 +102,9 @@ jobs:
 
     steps:
     - name: "Download dists"
-      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+      uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
       with:
-        name: "dist"
```
illia-v, line 105: I guess removing this line led to the job failure

GrantBirki: Looking at the logs I can see:  `Starting download of artifact to: /home/runner/work/urllib3/urllib3/dist/dist` ([ref](https://github.com/urllib3/urllib3/actions/runs/16756161327/job/47439256407#step:2:14)) and in the working logs, it looks like this:

```
Starting download of artifact to: /home/runner/work/urllib3/urllib3/dist
```
> [ref](https://github.com/urllib3/urllib3/actions/runs/16593313383/job/46934035410#step:2:13)

So it appears that there is some sort of double nested `dist/dist` action going on when not referencing by `name` 🤔

GrantBirki: Yep that is exactly what is going on! When using a single artifact by name, it resolves the path slightly differently:

https://github.com/actions/download-artifact/blob/448e3f862ab3ef47aa50ff917776823c9946035b/src/download-artifact.ts#L177-L179

So I think the easy fix here would just be to adjust the path used for the artifacts in other steps. I'll open a PR to fix this now.

GrantBirki:
@illia-v @pquentin I have created a follow-up PR that should address the issue with the nested artifact path here: https://github.com/urllib3/urllib3/pull/3650

illia-v:
@GrantBirki thanks for this security hardening and the quick fix!

PR #3655, opened 2025-08-10
Fix redirect handling when an integer is passed to a pool manager

### LINKED ISSUE #3649: Redirects do not work on a PoolManager configured with custom retries
### Subject

This started happening with the 2.5.0 release, as a result of changes made in https://github.com/urllib3/urllib3/commit/f05b1329126d5be6de501f9d1e3e36738bc08857. If you create a `PoolManager` with an explicit number of retries given as an integer, then redirects get disabled. I believe this is not intended.

### Environment

Describe your environment.

```
OS macOS-15.5-arm64-arm-64bit
Python 3.11.10
OpenSSL 3.5.0 8 Apr 2025
urllib3 2.5.0
```

### Steps to Reproduce

Run the following in a Python repl:

```python
import urllib3
http = urllib3.PoolManager(retries=10)
r = http.request("GET", "https://httpbin.org/redirect/1")
```

### Expected Behavior

A response is returned.

### Actual Behavior

The request raises the following exception:

```
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='httpbin.org', port=443): Max retries exceeded with url: https://httpbin.org/redirect/1 (Caused by ResponseError('too many redirects'))
```

### Additional details

The error only occurs when the pool is initialized with an integer number of retries. Both not passing `retries` and passing it set to a `Retry` instance leave redirects enabled and working as expected. Adding a `retries` argument to the `request()` call also works fine, even if an integer is passed as its value.

illia-v:
Fixes #3649

While investigating the issue I found out that `Retry` doesn't need to created with `redirect=False`.
I believe I passed it [previously](https://github.com/urllib3/urllib3/commit/f05b1329126d5be6de501f9d1e3e36738bc08857#diff-2fe80b3f580c0daa9f6a97de561c7fcd92fde02afdfaecdc210b6563817e69bb) to keep `Retry` instances more similar to what an underlying pool would create, but tests prove it's not necessary and leads to the reported bug.

### DIFF
--- a/changelog/3649.bugfix.rst
+++ b/changelog/3649.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed redirect handling in ``urllib3.PoolManager`` when an integer is passed
+for the retries parameter.
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -203,20 +203,18 @@ class PoolManager(RequestMethods):
         **connection_pool_kw: typing.Any,
     ) -> None:
         super().__init__(headers)
+        # PoolManager handles redirects itself in PoolManager.urlopen().
+        # It always passes redirect=False to the underlying connection pool to
+        # suppress per-pool redirect handling. If the user supplied a non-Retry
+        # value (int/bool/etc) for retries and we let the pool normalize it
+        # while redirect=False, the resulting Retry object would have redirect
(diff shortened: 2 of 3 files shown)


pquentin (approved):
Thanks! LGTM. Great catch.

PR #3664, opened 2025-08-26
Emscripten: Fix connection pool when used without an explicit port

hoodmane:
Without this change it inserts `None` as the port. In that case, the following code:
```py
import urllib3
pool = urllib3.HTTPConnectionPool("example.com", maxsize=10, block=True)

print(pool.request("GET", "/"))
```
fails with:
```
TypeError: Failed to parse URL from http://example.com:None/
```
In our tests we always use a test server on localhost with an explicit port.

### DIFF
--- a/changelog/3664.bugfix.rst
+++ b/changelog/3664.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed HTTPConnectionPool when used in Emscripten with no explicit port.
+Fixed ``HTTPConnectionPool`` when used in Emscripten with no explicit port.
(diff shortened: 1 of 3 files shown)


hoodmane:
@illia-v I'm not sure what the best way to test this is because all of the tests use a localhost server with an explicit port. Perhaps mocking `send_request` would be a good approach.

illia-v:
Yes, let's mock `send_request` and check the provided URL

Would you mind adding a [changelog](https://github.com/urllib3/urllib3/tree/main/changelog#readme) entry too?

hoodmane:
Okay added a test and a changelog.

illia-v (approved):
@hoodmane thanks for the fix!

hoodmane:
Thanks for the review @illia-v !

PR #3667, opened 2025-09-08
Changed static annotations of url parameter in __init__ method of RequestError and MaxRetryError class

subhajitsaha01:
The static annotation for the url parameter is changed from `str` to `str| None` of the __init__ methods of the RequestError and the MaxRetryError classes since in one of the test cases a `None` value is being passed to this particular parameter. The links of the sequence of argument data passing is mentioned as follows :

https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/test/test_retry.py#L24
https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/test/test_retry.py#L30
https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/src/urllib3/util/retry.py#L431
https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/src/urllib3/util/retry.py#L519
https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/src/urllib3/exceptions.py#L95
https://github.com/urllib3/urllib3/blob/62c9d1b0a4394216b93153b37f4451be3e94bd61/src/urllib3/exceptions.py#L45

### DIFF
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -42,7 +42,7 @@ class PoolError(HTTPError):
 class RequestError(PoolError):
     """Base exception for PoolErrors that have associated URLs."""
 
-    def __init__(self, pool: ConnectionPool, url: str, message: str) -> None:
+    def __init__(self, pool: ConnectionPool, url: str | None, message: str) -> None:
         self.url = url
         super().__init__(pool, message)
 
@@ -93,7 +93,7 @@ class MaxRetryError(RequestError):
     """
 
     def __init__(
-        self, pool: ConnectionPool, url: str, reason: Exception | None = None
+        self, pool: ConnectionPool, url: str | None, reason: Exception | None = None
     ) -> None:
         self.reason = reason
 


subhajitsaha01:
> @subhajitsaha01 thanks for the PR! I can see that `pool=None` is passed in the same example despite its type hint.
> 
> Also, `pool` can be `None` in non-test cases when exceptions are pickled and unpickled (check `__reduce__` methods in the exceptions file), would you like to update hints in all the exceptions?

Yes @illia-v , I did check the flow and `None` is getting passed inspite of only the static annotation being `ConnectionPool` for the `pool` parameter.

illia-v:
@subhajitsaha01 good, thanks for checking. Is there a reason why you don't update the `pool` annotation too? It seems like the annotations are related.

subhajitsaha01:
> @subhajitsaha01 good, thanks for checking. Is there a reason why you don't update the `pool` annotation too? It seems like the annotations are related.

No there is no specific reasons as such, was just checking on primitive types static annotations data being passed from the test cases but `pool` parameter is of non primitive type. Should I update the PR accordingly?

illia-v (approved):

illia-v:
> Should I update the PR accordingly?

A few exceptions use `pool=None` or `conn=None` in their `__reduce__`, you're welcome to update annotations of the parameters to make it clear that `None` is a possible value.

PR #3673, opened 2025-09-09
Specify JS and GitHub Actions in the CodeQL workflow

illia-v:
A follow-up to #3672.

It looks like CodeQL still scanned only Python and JS files, so I'm trying this change
https://github.com/urllib3/urllib3/actions/runs/17591775500/job/49974059426


illia-v:
It ran the queries for Actions this time, but it's not clear if any file was checked
image

PR #3681, opened 2025-09-22
Monitor GitHub Actions permissions

pquentin:
It will tell us if we can restrict permissions even more.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,6 +11,9 @@ jobs:
     timeout-minutes: 10
 
     steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+        with:
+          config: ${{ vars.PERMISSIONS_CONFIG }}
       - name: "Checkout repository"
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: GitHubSecurityLab/actions-permissions/monitor@v1
+        with:
+          config: ${{ vars.PERMISSIONS_CONFIG }}
       - name: "Checkout repository"
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
         with:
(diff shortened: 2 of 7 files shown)


pquentin:
Thanks for the approval! To be clear I am not planning to merge this, it just gives me useful information to restrict the permissions.

JarLob:
Hi, I have released a new version of `GitHubSecurityLab/actions-permissions` that fixes the `unknown permission` and also fixes running on MacOS 15. You should bump to `37c927c24552caa0ef6040ab0876db729cc12754`.

However I looked at the failing workflows and the bad news are:
1. Windows is not yet supported
2. The monitor action is implemented as MitM proxy and clearly the proxy related tests in the project do not like it.

So you can't use the action in `CI` and `Downstream` workflows.

Heh, maybe I need to rewrite the action to use eBPF instead. But wait... good news are, [StepSecurity implemented the Monitor functionality with eBPF already](https://www.stepsecurity.io/blog/determine-minimum-github-token-permissions-using-ebpf-with-stepsecurity-harden-runner). Try them instead.

JarLob:
Though it looks like `This feature is included in the StepSecurity enterprise tier`.

pquentin:
Hey @JarLob, thanks for checking! I'm not planning on merging this PR. It's too intrusive for a networking library. This was only to see if there were some quick wins. I don't understand why it always comes back as `permissions: {}` even where we upload/download artifacts.

JarLob:
Artifacts are strange beasts. Uploading artifacts doesn't use github token, so no permission is required. Downloading artifacts for public repositories doesn't require permissions. For private - `actions: read` is needed.

Or maybe more precisely `actions: read` is needed when [downloading artifact from a different repository or from a different workflow run](https://github.com/actions/download-artifact?tab=readme-ov-file#download-artifacts-from-other-workflow-runs-or-repositories). Said that, maybe there is a bug in the monitor if the artifacts are downloaded not from the same run, I don't remember comparing the workflow run id.

.github/workflows/lint.yml
```diff
@@ -2,15 +2,20 @@ name: lint
 
 on: [push, pull_request, workflow_dispatch]
 
-permissions:
-  contents: "read"
+permissions: {}
```
illia-v, line 5: I've just found out that empty permissions can break CI in private forks https://github.com/python/cpython/pull/148346

sigmavirus24:
What's the difference between this and zizmor? Zizmor also has an action, right?

pquentin:
@sigmavirus24 This has nothing to do with zizmor, this is supposed to tell us what permissions are used in practice, compared to what we're requesting. But since this doesn't work and we don't want to use `permissions: {}` given @illia-v's comment above, I'm closing this.

PR #3687, opened 2025-10-03
[SSL] Loads SSL certificates with LRU cache

### LINKED ISSUE #3623: ssl_wrap_socket loads SSL certificates into the same context repeatedly
### Subject

During some work with AWS (using boto3/botocore) I noticed a lot of CPU being used by SSLContext.load_verify_locations() inside ssl_wrap_socket(). There is an open issue about this problem in OpenSSL: https://github.com/openssl/openssl/issues/16871

It seems that urllib3 loads the certificates into the context each time ssl_wrap_socket() is called, even if you pass the same SSLContext, which is clearly wasteful and means you hit the OpenSSL issue over and over again.

### Environment

Describe your environment.

Python 3.11.11
OpenSSL 3.2.2 4 Jun 2024 (RHEL 9.6)
urllib3 1.26.19
botocore 1.38.36

### Steps to Reproduce

I don't have a simple way to reproduce this as I needed to profile the code to spot the huigh number of calls to load_verify_locations(). Also, you need to using a lot of different AWS services and regions so there are a high number of connections

### Expected Behavior

The certificates should only be loaded once for a given SSLContext.

I've attached a patch which use functools.cache to avoid loading the certs more than once for a given SSLContext which makes a huge difference in my tests.

[load_verify_locations.patch](https://github.com/user-attachments/files/20755501/load_verify_locations.patch)

### Actual Behavior

The certificates are reloaded into the same SSLContext every time ssl_wrap_socket() is called

intersectCoding:
## Related Issue
Closes #3623.

Thanks for the feedback! I’ve updated the implementation to use @lru_cache(maxsize=128) instead of @cache—this keeps the memoization benefits while enforcing a memory limit.

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
+from __future__ import absolute_import
 
+from functools import lru_cache
 import hashlib
 import hmac
 import os
@@ -221,6 +223,9 @@ def resolve_ssl_version(candidate: None | int | str) -> int:
 
     return candidate
 
+@lru_cache(maxsize=128)
+def load_verify_locations(context, ca_certs, ca_cert_dir, ca_cert_data):
+    context.load_verify_locations(ca_certs, ca_cert_dir, ca_cert_data)
 
 def create_urllib3_context(
     ssl_version: int | None = None,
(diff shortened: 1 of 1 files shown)


illia-v:
Hi @intersectCoding,

This is a duplicate of #3628, I'm sorry that the pull request was not referenced in the issue

PR #3696, opened 2025-10-17
Support Python 3.14t

pquentin:
Now that cffi and cryptography support free-threading, it's possible to test it in CI.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -50,7 +50,11 @@ def tests_impl(
     # Print the Python version, bytesize and free-threading status.
     session.run("python", "--version")
     session.run("python", "-c", "import struct; print(struct.calcsize('P') * 8)")
-    session.run("python", "-c", "import sys; print(getattr(sys, '_is_gil_enabled', lambda: True)())")
+    session.run(
+        "python",
+        "-c",
+        "import sys; print(getattr(sys, '_is_gil_enabled', lambda: True)())",
+    )
     # Print OpenSSL information.
     session.run("python", "-m", "OpenSSL.debug")
 


pquentin:
The lint failures are caused by the dependencies bump. I didn’t investigate yet.

.github/workflows/ci.yml
```diff
@@ -146,13 +146,6 @@ jobs:
           # noxfile.py contains the Pyodide version used currently.
           key: pyodide-downloads-${{ hashFiles('noxfile.py') }}
 
-      # https://github.com/wntrblm/nox/issues/986
-      - name: Create symlink to python.exe to help nox find Python
-        if: runner.os == 'Windows'
-        shell: cmd
-        run: |
-          mklink "%Python3_ROOT_DIR%\python${{ matrix.python-version }}.exe" "%Python3_ROOT_DIR%\python.exe"
```
illia-v, line 154: It was fixed by https://github.com/wntrblm/nox/pull/989, right?

pquentin: Yes, I think so.

pquentin:
> They are caused by the mypy upgrade most likely.

Indeed! I reverted the mypy and h2 upgrades.

> I've noticed a weird mismatch in Python versions used for Ubuntu 3.14t and Windows 3.14t runs:

Good catch! I fixed this by upgrading uv from 0.7 to 0.9. The uv version is a parameter of the setup-uv action, and we don't have any automation to upgrade it. I also disabled memray for free-threading builds as it currently crashes.

Please take another look!

kuraga:
`3.13t`?

PR #3697, opened 2025-10-19
Use backport.zstd instead of zstandard

### LINKED ISSUE #3693: Consider switching from zstandard to backports.zstd
We maintain two decoders for zstd:
* one using the [`zstandard`](https://pypi.org/project/zstandard/) package for Python < 3.14
* another using `compression.zstd` from the standard library for Python >= 3.14

We should consider switching from `zstandard` to the new [`backports.zstd`](https://pypi.org/project/backports.zstd/) package to maintain just one decoder class.

Rogdham:
For Python versions before 3.14, the Zstandard support is currently provided by the `zstandard` library. The API of this library is different from the one in the standard library ([`compression.zstd`](https://docs.python.org/3.14/library/compression.zstd.html)), which results in two different implementations of the `ZstdDecoder` class.

This PR removes the `zstdandard` dependency in favor of the [backport of `compression.zstd`](https://github.com/Rogdham/backports.zstd).

Fixes #3693

_Full disclosure: I'm the author and maintainer of `backports.zstd`, and the maintainer of `pyzstd` (which code was used as a base for the integration into Python). I also helped with PEP-784 and its integration into CPython._

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2,17 +2,30 @@ version = 1
 revision = 2
 requires-python = ">=3.9"
 resolution-markers = [
-    "python_full_version >= '3.10'",
+    "python_full_version >= '3.11'",
+    "python_full_version == '3.10.*'",
     "python_full_version < '3.10'",
 ]
 
 [[package]]
-name = "aiofiles"
-version = "24.1.0"
+name = "accessible-pygments"
+version = "0.0.5"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/0b/03/a88171e277e8caa88a4c77808c20ebb04ba74cc4681bf1e9416c862de237/aiofiles-24.1.0.tar.gz", hash = "sha256:22a075c9e5a3810f0c2e48f3008c94d68c65d763b9b03857924c99e57355166c", size = 30247, upload-time = "2024-06-24T11:02:03.584Z" }
+dependencies = [
+    { name = "pygments" },
+]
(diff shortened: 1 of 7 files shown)


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
pquentin, line 2: Do you need to update your uv version?

pquentin (dismissed):
Thanks! LGTM.

(wow, we repeat a lot of that zstd detection code, maybe this should be refactored in a follow-up PR)

uv.lock
Rogdham: Hmm maybe I am doing something wrong, but running `uv lock` does not make any further changes.

I have updated `uv` to `0.9.4`, which should be the latest version, and still the same.

Feel free to fix if you know how!

changelog/3693.misc.rst
```diff
@@ -0,0 +1 @@
+Provide Zstandard support through the ``backports.zstd`` library for Python before 3.14.
```
illia-v: Let's explicitly state that the `zstd` extra now installs `backports.zstd` package instead of `zstandard`

changelog/3693.misc.rst
```diff
@@ -0,0 +1 @@
+The ``zstd`` extra now installs ``backports.zstd`` instead of ``zstandard`` on Python versions before 3.14.
```
pquentin: Applied the guidelines from https://github.com/urllib3/urllib3/tree/main/changelog (that I often forget!):

```suggestion
Changed the ``zstd`` extra to install ``backports.zstd`` instead of ``zstandard`` on Python 3.13 and before.
```

pquentin (approved):

PR #3704, opened 2025-10-29
Issue 3688 inconsistent exception behavior

### LINKED ISSUE #3688: Inconsistent exception behavior with status-based retries
### Subject

It seems that when the maximum number of retries is exceeded, and the last request fails with a status code in `status_forcelist`, then a `MaxRetryError` is thrown; but if the status code is in `RETRY_AFTER_STATUS_CODES`, then no exception is thrown.

### Environment

Describe your environment.

```python
OS Linux-6.14.0-33-generic-x86_64-with-glibc2.39
Python 3.12.3
OpenSSL 3.0.13 30 Jan 2024
urllib3 2.5.0
```

### Steps to Reproduce

Run this test server:

```python
from http.server import HTTPServer, BaseHTTPRequestHandler

class Handler(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(int(self.path.removeprefix('/')))
        self.send_header('Retry-After', '1')
        self.end_headers()

if __name__ == "__main__":
    server = HTTPServer(('localhost', 8000), Handler)
    server.serve_forever()
```

Then, in parallel:

```
>>> r = urllib3.util.Retry(1, status_forcelist=[502])
>>> urllib3.request("GET", "http://localhost:8000/503", retries=r)
<urllib3.response.HTTPResponse object at 0x7bcb1dbfc730>
>>> urllib3.request("GET", "http://localhost:8000/502", retries=r)
urllib3.exceptions.ResponseError: too many 502 error responses

 [... snipped ...]
urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url: /502 (Caused by ResponseError('too many 502 error responses'))
```

### Expected Behavior

Both calls to `urllib3.request` should raise a `MaxRetryError`. Alternatively, the current behavior should be documented.

### Actual Behavior

When the status is in `status_forcelist`, the exception is thrown. But when it's in `RETRY_AFTER_STATUS_CODES` (and the `Retry-After` header is present), the retry still happens, but no exception is thrown.

Calvin0125:
Updated `Retry.is_retry` so that a `MaxRetryError` is raised if retries are exhausted for the status codes in `RETRY_AFTER_STATUS_CODES` (413, 429, 503). 

This fixes [Issue 3688](https://github.com/urllib3/urllib3/issues/3688) which outlines the inconsistent error behavior. 

Currently the if condition in `Retry.is_retry` checks if `total` is falsy, so if a status is not in the `status_forcelist` and `total` is `0`, `is_retry` returns `False`.  This means that `retries = retries.increment(method, url, response=response, _pool=self)` (line 932 in connectionpool.py) never runs, so the `MaxRetryError` is not raised once retries are exhausted if the status code is not added to the `status_forcelist`. 

The updated code explicitly checks if `total` is `False`, so `is_retry` will return `True` if `total` is `0` but all other conditions are met. This preserves the expected behavior that retries are disabled when `total` is set to `False`. 

This is my first contribution to this codebase so please let me know if there is anything I missed here. Also, this does change behavior, so I'm not sure if it would be considered a 'breaking change' for users of urllib3. Thanks for taking the time to review this!


Calvin0125:
Hi @illia-v, this is now ready for review. Please let me know if you have any questions or concerns.

src/urllib3/util/retry.py
```diff
@@ -384,14 +384,21 @@ def _is_method_retryable(self, method: str) -> bool:
             return False
         return True
 
+    @property
+    def disabled(self) -> bool:
+        """Check if retries are disabled. When self.total is set to False, the expected behavior is
+        that retries will be disabled. The status_forcelist overrides this and will retry even
+        if self.total is False.
+        """
+        return self.total is False
```
sigmavirus24, line 393: What about when total is 0?

Calvin0125: This new property would return `False` if `total` is 0. This preserves the expected behavior that retries are disabled when `total` is set to `False`, but allows `is_retry` to return `True` even if `total` is 0 but all other conditions are met. This allows `retries = retries.increment(method, url, response=response, _pool=self)` (line 932 in connectionpool.py) to be called so that a `MaxRetryError` is raised when retries are exhausted for status codes in `RETRY_AFTER_STATUS_CODES` (413, 429, 503).

sigmavirus24: I'm not sure this is the right thing to do though. I personally expect 0 and false to be equivalent

Calvin0125: Ok that's fair. What if I change `is_retry` back to the way it was and make `status_forcelist` default to `RETRY_AFTER_STATUS_CODES` in `__init__`? This would keep the original behavior if `total` is `False` or 0, and still raise a `MaxRetryError` for the status codes in `RETRY_AFTER_STATUS_CODES`.

Calvin0125: Hey @sigmavirus24, I just wanted to follow up, do you think the above idea would work?

DPS0340:
Drive-by from someone who arrived at this line independently: I was auditing truthiness-vs-identity checks in `retry.py` and landed on exactly the `self.total` in `is_retry()` that this PR changes. Rather than open a duplicate, here is the end-to-end reproduction I had built, in case it is useful evidence for the review.

**`total=None` and `total=0` are documented as opposites** —

```
:param int total:
    Set to ``None`` to remove this constraint and fall back on other counts.
    Set to ``0`` to fail on the first retry.
```

— but the truthiness check collapses them, and `None` is also the value that means *"defer to the per-category counts"*:

```python
Retry(total=None, respect_retry_after_header=True).is_retry("GET", 503, True)  # False
Retry(total=0,    respect_retry_after_header=True).is_retry("GET", 503, True)  # False
Retry(total=1,    respect_retry_after_header=True).is_retry("GET", 503, True)  # True
```

**It reaches real traffic.** Against a local server that returns `503` + `Retry-After: 0` twice and then `200`:

```
Retry(total=None, status=3)  ->  status=503 after 1 request
Retry(total=3)               ->  status=200 after 3 requests
```

So a caller who sets `total=None, status=3` — following the docstring to delegate to the status counter — gets no retry at all on a `Retry-After` response. `status_forcelist` masks it, because that branch returns before reaching this one, which is probably why it has stayed quiet.

Your `not self.disabled` fixes this case as a side effect, since `disabled` is `self.total is False` and `None`/`0` no longer collapse into it. Worth noting because it widens the PR's blast radius slightly beyond the exception-consistency framing in #3688: `total=None` users start seeing `Retry-After` retries they were silently not getting.

I have no stake in this PR and am not proposing changes to it — just handing over the repro so the behaviour change is visible in review rather than discovered later. Happy to open the `total=None` delegation as its own issue if you would rather keep this PR scoped to #3688.

Issue #3707, opened 2025-11-10
Allow ConnectionPool to be pickled, then use that to better serialize exceptions


csm10495:
### Context

I'm trying to make it so that when exceptions for requests (and at the lower level urllib3) come through multiprocessing (via pickle) they maintain as much context as possible. By making ConnectionPool pickle-able, we can save all pool information as needed to properly recreate exception text.

### Alternatives

I started trying to do this via this suggestion: https://github.com/urllib3/urllib3/issues/3567#issuecomment-3487579383 though I kept running into confusing cases where different exception classes have different messages and so on. 

I realized that I'm trying to work around a problem that I believe can be better solved by enhancing serialization capabilities for ConnectionPool (and children).

### Duplicate

Nope. Though this is an extension of https://github.com/urllib3/urllib3/issues/3567

### Contribution

Would you be willing to submit a PR?
Sure.

PR #3708, opened 2025-11-10
Make it so http/https connection pools are pickle-able.

csm10495:
The logic here is that if a pool is pickled, its metadata (like host/port) are saved for reuse on the other side of serialization. The inner pool object itself though isn't saved and specifically set to None, since it is the only piece that can't be properly serialized.

Related to: https://github.com/urllib3/urllib3/issues/3707 and https://github.com/urllib3/urllib3/issues/3567


csm10495:
I'm not sure what the lint is complaining about. I don't think I touched the stuff its complaining about. 😕

PR #3709, opened 2025-11-11
Include host in retry warning messages

### LINKED ISSUE #3583: Retry warning should show host as well as url/path
When a connection pool retries a request, the warning message prints only the url value, usually a relative path like `/simple/virtualenv/` (for connection host `https://pypi.org/`).  Since urllib3 and any calling code are opaquely determining the full url with hostname, it would be useful for debugging for this retry warning to indicate the full path, with something like `"%s/%s" % (self.host, url)`.

https://github.com/urllib3/urllib3/blob/3e8f2db735dcaced6a3b777aa1966f40c018af7c/src/urllib3/connectionpool.py#L868-L870

Is there anything in the host determination logic when the request is actually that precludes being able to succinctly include this information in the warning message at this scope?

mahdighadiriii:
Fixes #3583

When a connection pool retries a request, the warning message now includes the full URL (scheme + host + port + path) instead of just the path. This makes debugging connection issues easier when working with multiple hosts.

The port is included only when it differs from the scheme's default (80 for HTTP, 443 for HTTPS).

## Example
**Before:**
```
Retrying (...) after connection broken by '...': /simple/virtualenv/
```

**After:**
```
Retrying (...) after connection broken by '...': https://pypi.org/simple/virtualenv/
```

## Changes
- Modified `HTTPConnectionPool.urlopen()` to construct full URL for retry warning messages
- Added scheme-aware port handling to omit standard ports (80/443)

## Testing
All existing tests pass

### DIFF
--- a/changelog/3709.bugfix.rst
+++ b/changelog/3709.bugfix.rst
@@ -1,0 +1,1 @@
+Include host in retry warning messages for better debugging.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -873,9 +873,12 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             else:
                 # Standard port, omit it
                 full_url = f"{self.scheme}://{self.host}{url}"
-            
+
             log.warning(
-                "Retrying (%r) after connection broken by '%r': %s", retries, err, full_url
+                "Retrying (%r) after connection broken by '%r': %s",
+                retries,
+                err,
+                full_url,
             )
             return self.urlopen(
                 method,


mahdighadiriii:
The mypy lint failure is unrelated to my changes. I've verified:

My changes only touch `connectionpool.py` and the changelog
Lint passes locally: `uvx nox -s lint` succeeds  
All 38 other CI checks pass including all tests
The mypy error is about an optional `compression` module that my code doesn't touch

The error appears to be a CI environment dependency resolution issue. Could a maintainer please advise?

PR #3711, opened 2025-11-13
Improve speed of `BytesQueueBuffer.get()` by using memoryview

### LINKED ISSUE #3710: Performance issue in `BytesQueueBuffer.get()` byte slicing
### Subject

`BytesQueueBuffer.get()` is slow when downloading gzipped, unchunked data.

By profiling with `line_profiler`, I identified the slowest part in the `BytesQueueBuffer.get()`, specifically the following line:
https://github.com/urllib3/urllib3/blob/b20c8368ec02421ecc4e49d796cd5eff1e04f325/src/urllib3/response.py#L259

### Environment

```
OS Linux-6.12.48-1-MANJARO-x86_64-with-glibc2.42
Python 3.12.5
OpenSSL 3.0.14 4 Jun 2024
urllib3 0.1.dev4309
```

I also confirmed this with python 3.14.

As this is a performance issue, I will put the CPU and memory info just in case:
```
CPU: 13th Gen Intel(R) Core(TM) i5-13500

> free
               total        used        free      shared  buff/cache   available
Mem:           31833       11164       15246        1254        7134       20669
Swap:              0           0           0
```

### Steps to Reproduce

1. Clone https://gist.github.com/phenylshima/0af2d0692db455d2d040805289689fef
2. Install `urllib3==2.5.0`, `fastapi==0.121.1`, and `uvicorn==0.38.0`
3. Start test_server.py
4. Run test_client.py, and check the "Content retrieval time" line.

### Expected Behavior

The two "Content retrieval time" are almost the same.

### Actual Behavior

Content retrieval time is about 9x slower when decode_content=True, compared to taking the whole data with decode_content=False and then unzipping.

```
Response time: 0.001905202865600586 seconds
Content retrieval time: 2.0758426189422607 seconds
Response time: 0.0009431838989257812 seconds
Content retrieval and decompression time: 0.2243025302886963 seconds
```

### Profiling

I think this result contains significant overhead associated with line_profiler, but it shows that the aforementioned line is taking considerably longer time than others.

<details>
<summary>line_profiler result</summary>

```python
Timer unit: 1e-06 s

Total time: 2.37476 s
File: /home/user/0af2d0692db455d2d040805289689fef/.venv/lib/python3.12/site-packages/urllib3/response.py
Function: BytesQueueBuffer.get at line 282

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   282                                               @line_profiler.profile
   283                                               def get(self, n: int) -> bytes:
   284    134218      39514.2      0.3      1.7          if n == 0:
   285                                                       return b""
   286    134218      40398.6      0.3      1.7          elif not self.buffer:
   287                                                       raise RuntimeError("buffer is empty")
   288    134218      38182.3      0.3      1.6          elif n < 0:
   289                                                       raise ValueError("n should be > 0")
   290                                           
   291    134218      37615.7      0.3      1.6          fetched = 0
   292    134218      49670.3      0.4      2.1          ret = io.BytesIO()
   293    134361      39164.8      0.3      1.6          while fetched < n:
   294    134361      38041.0      0.3      1.6              remaining = n - fetched
   295    134361      38083.3      0.3      1.6              chunk = self.buffer.popleft()
   296    134361      38311.4      0.3      1.6              chunk_length = len(chunk)
   297    134361      34412.0      0.3      1.4              if remaining < chunk_length:
   298    134217    1740781.8     13.0     73.3                  left_chunk, right_chunk = chunk[:remaining], chunk[remaining:]
   299    134217      57110.2      0.4      2.4                  ret.write(left_chunk)
   300    134217      39958.4      0.3      1.7                  self.buffer.appendleft(right_chunk)
   301    134217      38834.6      0.3      1.6                  self._size -= remaining
   302    134217      35020.9      0.3      1.5                  break
   303                                                       else:
   304       144         80.0      0.6      0.0                  ret.write(chunk)
   305       144         49.4      0.3      0.0                  self._size -= chunk_length
   306       144         44.4      0.3      0.0              fetched += chunk_length
   307                                           
   308       144         39.9      0.3      0.0              if not self.buffer:
   309         1          0.3      0.3      0.0                  break
   310                                           
   311    134218      69447.3      0.5      2.9          return ret.getvalue()

Total time: 4.33477 s
File: /home/user/0af2d0692db455d2d040805289689fef/.venv/lib/python3.12/site-packages/urllib3/response.py
Function: HTTPResponse.stream at line 1071

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
  1071                                               @line_profiler.profile
  1072                                               def stream(
  1073                                                   self, amt: int | None = 2**16, decode_content: bool | None = None
  1074                                               ) -> typing.Generator[bytes]:
  1075                                                   """
  1076                                                   A generator wrapper for the read() method. A call will block until
  1077                                                   ``amt`` bytes have been read from the connection or until the
  1078                                                   connection is closed.
  1079                                           
  1080                                                   :param amt:
  1081                                                       How much of the content to read. The generator will return up to
  1082                                                       much data per iteration, but may return less. This is particularly
  1083                                                       likely when using compressed data. However, the empty string will
  1084                                                       never be returned.
  1085                                           
  1086                                                   :param decode_content:
  1087                                                       If True, will attempt to decode the body based on the
  1088                                                       'content-encoding' header.
  1089                                                   """
  1090         1          0.9      0.9      0.0          if self.chunked and self.supports_chunked_reads():
  1091                                                       yield from self.read_chunked(amt, decode_content=decode_content)
  1092                                                   else:
  1093    134219     206651.4      1.5      4.8              while not is_fp_closed(self._fp) or len(self._decoded_buffer) > 0:
  1094    134218    4038064.2     30.1     93.2                  data = self.read(amt=amt, decode_content=decode_content)
  1095                                           
  1096    134218      38546.2      0.3      0.9                  if data:
  1097    134218      51505.8      0.4      1.2                      yield data

  2.37 seconds - /home/user/0af2d0692db455d2d040805289689fef/.venv/lib/python3.12/site-packages/urllib3/response.py:282 - BytesQueueBuffer.get
  4.33 seconds - /home/user/0af2d0692db455d2d040805289689fef/.venv/lib/python3.12/site-packages/urllib3/response.py:1071 - HTTPResponse.stream
```

</details>

phenylshima:
Fixes #3710

This pull request improves the speed of `BytesQueueBuffer.get()` by using memoryview, solving the problem described in #3710 .

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,5 @@ updates:
       # upgrade if we run into a bug and need a fix.
       - dependency-name: "*"
         update-types: ["version-update:semver-patch"]
+    cooldown:
+      default-days: 7
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           # `towncrier check` runs `git diff --name-only origin/main...`, which
           # needs a non-shallow clone.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
@@ -98,7 +98,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           fetch-depth: 0 # Needed to fetch the version from git
           persist-credentials: false
@@ -170,7 +170,7 @@ jobs:
     needs: test
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,7 +23,7 @@ jobs:
       security-events: write
     steps:
     - name: "Checkout repository"
-      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
       with:
         persist-credentials: false
 
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           fetch-depth: 0 # Needed to fetch the version from git
           persist-credentials: false
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -22,7 +22,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           persist-credentials: false
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,34 +1,34 @@
 repos:
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.0
+    rev: v3.21.2
     hooks:
       - id: pyupgrade
         args: ["--py39-plus"]
 
-  - repo: https://github.com/psf/black
-    rev: 24.10.0
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 25.11.0
     hooks:
       - id: black
         args: ["--target-version", "py39"]
 
   - repo: https://github.com/PyCQA/isort
-    rev: 5.13.2
+    rev: 7.0.0
     hooks:
       - id: isort
 
   - repo: https://github.com/PyCQA/flake8
-    rev: 7.1.1
+    rev: 7.3.0
     hooks:
       - id: flake8
         additional_dependencies: [flake8-2020]
 
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.7.8
+    rev: 0.9.10
     hooks:
       - id: uv-lock
 
   - repo: https://github.com/zizmorcore/zizmor-pre-commit
-    rev: v1.13.0
+    rev: v1.16.3
     hooks:
       - id: zizmor
 
--- a/dummyserver/testcase.py
+++ b/dummyserver/testcase.py
@@ -314,7 +314,7 @@ class ConnectionMarker:
         orig_request = HTTPConnection.request
 
         def call_and_mark(
-            target: typing.Callable[..., None]
+            target: typing.Callable[..., None],
         ) -> typing.Callable[..., None]:
             def part(
                 self: HTTPConnection, *args: typing.Any, **kwargs: typing.Any
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -228,7 +228,6 @@ def body_to_chunks(
     elif hasattr(body, "read"):
 
         def chunk_readable() -> typing.Iterable[bytes]:
-            nonlocal body, blocksize
             encode = isinstance(body, io.TextIOBase)
             while True:
                 datablock = body.read(blocksize)
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -182,7 +182,7 @@ def requires_network() -> typing.Callable[[_TestFuncT], _TestFuncT]:
         return typing.cast(_TestFuncT, wrapper)
 
     def _decorator_requires_internet(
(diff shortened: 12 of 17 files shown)


sigmavirus24:
Does it make more sense to just always store the byte chunks as memoryviews? My instinct was that would solve this as well. 

We'll need to update the type hints as well as now the deque can have either bytes or memoryview probably

phenylshima:
@sigmavirus24 Thank you for your suggestion! I updated the code to store the data as memoryview.

sigmavirus24:
Is that as performant at least?

phenylshima:
Yes, at least on my machine.

Test code:
```py
from src.urllib3.response import BytesQueueBuffer
import time

buffer = BytesQueueBuffer()
buffer.put(b"x" * 1024 * 1024 * 10)  # 10 MiB
now = time.perf_counter()
while len(buffer) > 5000:
    buffer.get(1024)  # 1 KiB
delta = time.perf_counter() - now

print(delta)
```

The result:
[495aab1](https://github.com/urllib3/urllib3/pull/3711/commits/495aab1bd0e74cc485d7790ba7132802fbfebc3c): 0.0052835239985142834
[19820a4](https://github.com/urllib3/urllib3/pull/3711/commits/19820a42035b7ce266da1f9792f515027a733c41): 0.004722315999970306

The new one looks faster, but it is because the `memoryview(chunk)` is moved to `put` method. I think the overall performance is the same. One thing to note is when we write a big chunk once, and read it only once (either by get or get_all), I expect it to be slightly slower compared to the current main branch (due to running `memoryview(chunk)` and converting it back to `bytes`. If get_all is used, the conversion back to bytes will not happen though).

sigmavirus24:
Worth checking for a performance hit on `.put` then from before your changes to now then, although I don't expect much of one. May as well also do a full end-to-end performance test. Put a bunch of chunks, do a `get_all`, etc.

phenylshima:
I tested in the original repro, and it looked fine.

However, the `put` method got slower (surprising to me):

```py
from src.urllib3.response import BytesQueueBuffer
import time

buffer = BytesQueueBuffer()
now = time.perf_counter()

for _ in range(10 * 1024):  # 10 MiB total
    buffer.put(b"x" * 1024)  # 1 KiB

delta = time.perf_counter() - now
print(f"Time to put data: {delta}s")

while len(buffer) > 0:
    buffer.get(1024)  # 1 KiB

delta = time.perf_counter() - now

print(f"Time to get data: {delta}s")
```

```
> ~/dev/urllib3> main !1 ?1> python t.py
Time to put data: 0.0007599639998261409s
Time to get data: 0.004324924999764335s
> ~/dev/urllib3> main !1 ?1> python t.py
Time to put data: 0.0008291090002785495s
Time to get data: 0.004370479000044725s
> ~/dev/urllib3> main !1 ?1> python t.py
Time to put data: 0.0007629809997524717s
Time to get data: 0.004293366999718273s

> ~/dev/urllib3> @495aab1b ?1> python t.py
Time to put data: 0.0007698970002820715s
Time to get data: 0.004365679000329692s
> ~/dev/urllib3> @495aab1b ?1> python t.py
Time to put data: 0.000990569999885338s
Time to get data: 0.004518313000062335s
> ~/dev/urllib3> @495aab1b ?1> python t.py
Time to put data: 0.0009582410002622055s
Time to get data: 0.004823216000204411s

> ~/dev/urllib3> byte-queue-perf ?1> python t.py
Time to put data: 0.003418835000047693s
Time to get data: 0.007445904999713093s
> ~/dev/urllib3> byte-queue-perf ?1> python t.py
Time to put data: 0.0033251579998250236s
Time to get data: 0.007321049999973184s
> ~/dev/urllib3> byte-queue-perf ?1> python t.py
Time to put data: 0.003460155999619019s
Time to get data: 0.00736400899995715s
```

Maybe because this code is not reveraging anything from bytes->memoryview conversion, as it is just reading and writing the same size?

It might be better to use 495aab1b.

phenylshima:
Now it's like this:
```
>~/dev/urllib3 > byte-queue-perf ?1 > python t.py
Time to put data: 0.0007707550003033248s
Time to get data: 0.004282103000150528s
>~/dev/urllib3 > byte-queue-perf ?1 > python t.py
Time to put data: 0.0007912390001365566s
Time to get data: 0.004329170999881171s
>~/dev/urllib3 > byte-queue-perf ?1 > python t.py
Time to put data: 0.0007524289994762512s
Time to get data: 0.004279935999875306s
```

src/urllib3/response.py
```diff
@@ -231,14 +231,14 @@ class BytesQueueBuffer:
...
 
     def __init__(self) -> None:
-        self.buffer: typing.Deque[memoryview[bytes]] = collections.deque()
+        self.buffer: typing.Deque[bytes | memoryview[bytes]] = collections.deque()
         self._size: int = 0
 
     def __len__(self) -> int:
         return self._size
 
     def put(self, data: bytes | memoryview[bytes]) -> None:
```
sigmavirus24: data will only be bytes then, right?

phenylshima: It can be memoryview. In line 259, the existing chunk is converted to memoryview, and in line 262, the memoryview is written back to the buffer. Therefore, the buffer can have both bytes and memoryview.

The get and get_all methods are written so that, if the memoryview is in the buffer, it will be converted to bytes.

test/test_response.py
```diff
@@ -125,6 +126,45 @@ def test_get_all_memory_usage_single_chunk(self) -> None:
...
+        buffer = BytesQueueBuffer()
+
+        now = time.perf_counter()
+        buffer.put(b"x" * 1024 * 1024 * 10)  # 10 MiB
+        while len(buffer) > 0:
+            get_func(buffer)
+
+        delta = time.perf_counter() - now
+
+        assert delta < 0.5, f"Data put & get time too long: {delta}s"
```
illia-v: The tests that assert on the time delta might be a bit fragile across different environments. Machines with different performance characteristics can cause these timing-based tests to pass or fail inconsistently.

The patched version does clearly improve performance, though: on my machine `TestBytesQueueBuffer.test_performance_single_chunk[get]` shows that it allocates almost half as much memory. Would it make sense to rely on this and add an explicit test that checks the memory allocation instead of the timing-based tests?

src/urllib3/response.py
illia-v: Since line 262 calls `self.buffer.appendleft`, `BytesQueueBuffer.put` is never called with a memory view, right?

phenylshima: Yeah, right. My initial intention was that "maybe allowing memoryview here could benefit someone", but on my second thought, I removed the `memoryview[bytes]`.

test/test_response.py
phenylshima: Thank you for your comment. I agree with the problems with the time delta-based method.
I changed the code to check that there is not additional memory allocation (which is also the cause of this performance issue I guess).

illia-v (approved):
I extended the tests a bit while testing the change locally and fixed a minor issue in the changelog entry

src/urllib3/response.py
illia-v, line 230: I wonder if this comment is still completely correct

test/test_response.py
```diff
@@ -108,15 +108,17 @@ def test_get_all_many(self) -> None:
...
     ) -> None:
         # Allocate 10 1MiB chunks
         buffer = BytesQueueBuffer()
         for i in range(10):
             # This allocates 2MiB, putting the max at around 12MiB. Not sure why.
             buffer.put(bytes(2**20))
 
-        assert len(get_func(buffer)) == 10 * 2**20
+        result = get_func(buffer)
+        assert type(result) is bytes
```
sigmavirus24, line 120: Why not isinstance?

illia-v: I just wanted to ensure that it returns bytes and not something else including a subclass instance which `isinstance` wouln't differenciate

src/urllib3/response.py
phenylshima: This test fails with message "Test was limited to 10.0MiB but allocated 20.0MiB":

```
    @pytest.mark.limit_memory("10.01 MB", current_thread_only=True)
    def test_get_memory_usage_single_chunk(self) -> None:
        buffer = BytesQueueBuffer()
        chunk = bytes(10 * 2**20)  # 10 MiB
        buffer.put(chunk)
        assert len(buffer.get(10 * 2**20)) == 10 * 2**20
```

The write to BytesIO (line 261 and line 266) seems to copy the content.

illia-v: Thanks for checking, we'll fix it separately

pquentin:
Am I misreading the benchmarks? I'm not seeing a speed difference between `main` and the last `byte-queue-perf` benchmark: 0.7ms for put and 4ms for get.

phenylshima:
Thank you for looking at this PR. The current test does not test the execution time, rather the non-existance of extra memory allocation (that led to the overhead in the current main branch).
Changing `buffer.get(2**20)` to `buffer.get(2**10)` or something like that should make the performance difference more visible (I'm not sure if I should commit it, though).

illia-v:
I believe Quentin asked about the benchmarks from https://github.com/urllib3/urllib3/pull/3711#issuecomment-3530483296 and https://github.com/urllib3/urllib3/pull/3711#issuecomment-3533040535 which did not measure the effect of using `memoryview` for chunks that are split.

The benchmarked case was related to a regression that was introduced by this PR initially and has been resolved since then in e898682a2ba0ef1e1205366a48d789f63cf3ec58: memoryview was used even when a chunk was not split.

phenylshima:
Ah I see, thank you for the followup!

pquentin (approved):
Thanks! LGTM. Illia showed me a (private for now) benchmark where this PR makes a drastic difference.

PR #3716, opened 2025-11-19
Bump versions of pre-commit hooks

illia-v:
https://github.com/urllib3/urllib3/actions/runs/19516465216/job/55869409115?pr=3715

There is a warning in the lint output that the isort step may stop working soon:
> Warning:  repo `[https://github.com/PyCQA/isort`](https://github.com/PyCQA/isort%60) uses deprecated stage names (commit, merge-commit, push) which will be removed in a future version.  Hint: often `pre-commit autoupdate --repo [https://github.com/PyCQA/isort`](https://github.com/PyCQA/isort%60) will fix this.  if it does not -- consider reporting an issue to that repo.

Additionally, my editor started reformatting some files with a newer version of black and isort causing noisy git diffs.

So I'm upgrading all pre-commit hooks related to Python here and applying suggested fixes.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
         with:
           # `towncrier check` runs `git diff --name-only origin/main...`, which
           # needs a non-shallow clone.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
       - name: "Checkout repository"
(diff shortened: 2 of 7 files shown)


.pre-commit-config.yaml
```diff
@@ -1,34 +1,34 @@
 repos:
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.19.0
+    rev: v3.21.2
     hooks:
       - id: pyupgrade
         args: ["--py39-plus"]
 
   - repo: https://github.com/psf/black
```
hugovk: This will use the 2x faster mypyc-compiled wheels: https://github.com/psf/black-pre-commit-mirror
```suggestion
  - repo: https://github.com/psf/black-pre-commit-mirror
```

illia-v: Cool, thanks for the suggestion!

sigmavirus24 (approved):

hugovk:
Tip: you may want to enable https://pre-commit.ci/ for this repo to have automated updates of the hooks. It can also update PRs when there are fixes, which can save a bit of extra hassle.

I also stick this at the bottom of the config so you only get autoupdates quarterly instead of the default weekly:

```yaml
ci:
  autoupdate_schedule: quarterly
```

PR #3718, opened 2025-11-20
Add pyOpenSSL and PySocks to mypy checks

pquentin:
This no longer actually upgrades mypy: the version in uv.lock stays the same.

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -381,13 +381,26 @@ class WrappedSocket:
         except OpenSSL.SSL.Error:
             return
 
+    @typing.overload
+    def getpeercert(
+        self, binary_form: typing.Literal[False] = False
+    ) -> dict[str, list[typing.Any]] | None: ...
+
+    @typing.overload
+    def getpeercert(self, binary_form: typing.Literal[True]) -> bytes | None: ...
+
+    @typing.overload
+    def getpeercert(
+        self, binary_form: bool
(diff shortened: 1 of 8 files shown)


pquentin:
I had missed https://github.com/urllib3/urllib3/pull/3715! My version is more comprehensive, but it currently fails due to a dependency conflict. Dropping 3.9 will fix it.

src/urllib3/contrib/pyopenssl.py
```diff
@@ -384,22 +383,22 @@ def _real_close(self) -> None:
 
     def getpeercert(
         self, binary_form: bool = False
-    ) -> dict[str, list[typing.Any]] | None:
+    ) -> dict[str, list[typing.Any]] | bytes | X509 | None:
```
illia-v: What do you think about `@typing.overload` for `binary_form: Literal[True]` and `binary_form: Literal[False]`?

pquentin: Thanks, done. I went back and forth on adding an "unknown bool" case where the value (True or False) is only known at runtime, but ultimately decided to remove it. It would have added those lines:

```python
    @typing.overload
    def getpeercert(
        self, binary_form: bool
    ) -> dict[str, list[typing.Any]] | bytes | None: ...
```

sigmavirus24 (approved):

src/urllib3/contrib/pyopenssl.py
illia-v: Thanks!

illia-v (approved):

PR #3024, opened 2023-05-10
Allow alternative SSL libaries with a warning

pquentin:
Closes https://github.com/urllib3/urllib3/issues/3020

I'll have to run those two tests manually since the test suite does not cover them:

- [x] LibreSSL
- [x] OpenSSL 1.0.2

PR #3720, opened 2025-11-20
Drop Python 3.9

pquentin:


docs/v2-migration-guide.rst
```diff
@@ -48,7 +48,7 @@ Sunsetting urllib3 1.26.x
 urllib3 1.26.x is not currently maintained. urllib3 2.x is the best version of urllib3
 and is widely supported by the larger Python ecosystem. That said, urllib3 1.26.x still
 sees significant download numbers, mainly because the botocore package still requires
-urllib3 1.26.x for Python 3.9 and earlier. If your organization would benefit from the
+urllib3 1.26.x for Python 3.10 and earlier. If your organization would benefit from the
```
illia-v: botocore doesn't require urllib3 1.26.x for Python 3.10

docs/contributing.rst
```diff
@@ -66,7 +66,7 @@ suite::
   [ Nox will create virtualenv if needed, install the specified dependencies, and run the commands in order.]
   .......
   .......
-  nox > Session test-3.9 was successful.
+  nox > Session test-3.10 was successful.
```
illia-v, line 69: test-3.10 is mentioned on the next line

```suggestion
```

src/urllib3/util/ssl_.py
```diff
@@ -361,7 +350,7 @@ def create_urllib3_context(
 
     try:
         context.hostname_checks_common_name = False
-    except AttributeError:  # Defensive: for CPython < 3.9.3; for PyPy < 7.3.8
+    except AttributeError:  # Defensive: for for PyPy < 7.3.8
```
illia-v: PyPy 7.3.8 was a release for 2.7, 3.7, 3.8, and 3.9-beta https://pypy.org/posts/2022/02/pypy-v738-release.html

We don't support any of the Python/PyPy versions anymore, so the check can be dropped

src/urllib3/util/ssl_.py
```diff
@@ -46,9 +46,7 @@ def _is_bpo_43522_fixed(
         # https://foss.heptapod.net/pypy/pypy/-/issues/3129
         return pypy_version_info >= (7, 3, 8)  # type: ignore[operator]
```
illia-v: Same here regarding PyPy 7.3.8 support

illia-v: I assume `_is_bpo_43522_fixed` can be dropped now

montesmarie87-lgtm (approved):

pquentin:
I fully agree with your great review, @illia-v, but I haven't had time to fix my PR yet, sorry.

pquentin:
While tests pass now, coverage is failing because those two lines are no longer covered:

 * https://github.com/urllib3/urllib3/blob/816fcf04528bc0f89672e13398eb813dcc892490/src/urllib3/response.py#L739
 * https://github.com/urllib3/urllib3/blob/816fcf04528bc0f89672e13398eb813dcc892490/src/urllib3/contrib/socks.py#L144-L147

They look like defensive code. Should I just add `Defensive:`?

illia-v:
> They look like defensive code. Should I just add `Defensive:`?

Sound good to me based on #3728 showing that the code is not reached by Python 3.9 too

src/urllib3/util/ssl_.py
```diff
@@ -26,55 +26,6 @@
...
-    # Before fixing OpenSSL issue #14579, the SSL_new() API was not copying hostflags
-    # like X509_CHECK_FLAG_NEVER_CHECK_SUBJECT, which tripped up CPython.
-    # https://github.com/openssl/openssl/issues/14579
-    # This was released in OpenSSL 1.1.1l+ (>=0x101010cf)
-    is_openssl_issue_14579_fixed = openssl_version_number >= 0x101010CF
-
-    return is_openssl and (
-        is_openssl_issue_14579_fixed
-        or _is_bpo_43522_fixed(implementation_name, version_info, pypy_version_info)
-    )
```
illia-v, line 75: I'm not sure this whole method should be deleted because it will drop the support for alternative SSL libraries you added in #3024

pquentin: Right, good catch! I had forgotten about this. However, the Python that comes with macOS Developer Tools is 3.9, which we are dropping here. I don't think there's a common use case with Python 3.10+ and LibreSSL 2.8.3.

```
❯ /usr/bin/python3
Python 3.9.6 (default, Dec  2 2025, 07:27:58)
[Clang 17.0.0 (clang-1700.6.3.2)] on darwin
>>> import ssl
>>> ssl.OPENSSL_VERSION
'LibreSSL 2.8.3'
```

Likewise, supporting OpenSSL 1.0.2 only made sense for Python 3.9 given [PEP 644](https://peps.python.org/pep-0644/).

pquentin: That said, LLVM is considering upgrading the minimum Python version: https://discourse.llvm.org/t/rfc-upgrading-llvm-s-minimum-required-python-version/88605, at which point XCode will ship with a newer one too. So I could be convinced to keep supporting LibreSSL.

illia-v: BTW, I can see that CPython has a CI job for AWS-LC in addition to OpenSSL https://github.com/python/cpython/issues/135401. We may need to check what `_is_has_never_check_common_name_reliable` should return for AWS-LC, separately from this PR of course.

pquentin: LibreSSL isn't supported by Python 3.10+, but builds work again since https://github.com/python/cpython/pull/131128, which was released as part of Python 3.13.4 and 3.14.0. I'm wondering how hard it would be to get a LibreSSL-enabled Python on my laptop.

AWS-LC could be interesting indeed.

illia-v: > I'm wondering how hard it would be to get a LibreSSL-enabled Python on my laptop.

https://gist.github.com/illia-v/1103758861123320a09bd9f599d616f6
These were the commands to build CPython with custom OpenSSL 4 years ago, not sure if they work today

illia-v: After releasing 2.6.3, should we merge this PR with restored `_is_has_never_check_common_name_reliable` but dropped `_is_bpo_43522_fixed` and focus on the SSL issues separately?

illia-v: I'll proceed with the approach from the previous comment this week if you don't mind
So that we can resolve all the useless security alerts for vulnerable dev dependencies for Python 3.9

pquentin: Sure, thank you!

illia-v: > After releasing 2.6.3, should we merge this PR with restored `_is_has_never_check_common_name_reliable` but dropped `_is_bpo_43522_fixed` and focus on the SSL issues separately?

Done

illia-v:
> While tests pass now, coverage is failing because those two lines are no longer covered:

This is now handled based on my comments from https://github.com/urllib3/urllib3/pull/3728

src/urllib3/util/ssl_.py
```diff
@@ -359,10 +320,7 @@ def create_urllib3_context(
         context.check_hostname = False
         context.verify_mode = cert_reqs
 
-    try:
-        context.hostname_checks_common_name = False
-    except AttributeError:  # Defensive: for CPython < 3.9.3; for PyPy < 7.3.8
-        pass
+    context.hostname_checks_common_name = False
```
illia-v, line 315: TODO: check if this works with OpenSSL < 1.1.1l because a comment from this file suggests it may not

illia-v:
These lines suggest that we may still support OpenSSL 1.0.2 with PyOpenSSL 😱 
https://github.com/urllib3/urllib3/blob/fc76bcb911ef484bcfbdd8a27f002379d625d01f/src/urllib3/connection.py#L946-L948

Our docs define PyOpenSSL 16.0.0 as the minimum supported version
https://urllib3.readthedocs.io/en/2.6.3/reference/contrib/pyopenssl.html

src/urllib3/util/ssl_.py
illia-v: https://github.com/python/cpython/issues/87688#issuecomment-1093906646 and the issue in general make me think that any beta/final release of Python 3.10 has no issues with `hostname_checks_common_name` even with OpenSSL 1.1.1 older than 1.1.1l. So maybe we don't need the OpenSSL version check in `_is_has_never_check_common_name_reliable`.

illia-v: Python docs support my assumption
image

illia-v (approved):

PR #3721, opened 2025-11-21
Updated docs to reflect change in optional zstd dependency from `zstandard` to `backports.zstd`

edgarrmondragon:
This was done in https://github.com/urllib3/urllib3/commit/93e6ae22ab1b708ca859cfe7f6f219e8de20e015, but the "Advanced Usage" doc was not updated.


illia-v (commented):
Thanks for catching this!

I've dropped the changelog entry because we already have another related to the changed dependency for the next release

PR #3726, opened 2025-12-03
Fix for issue #3636 around read() after partial read not returning full content

sg3-141-592:
#3636 

This issue was, `read(512)`
- Reads the entire compressed stream from the socket
- Decompresses it
- Returns only the requested 512 bytes
- Stores the rest in `_decoded_buffer`

But when doing `read()` after, it would only try and read from the socket and ignore the buffered decoded data.

Now when `amt=None` we check if `_decoded_buffer` has leftover data

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -658,7 +658,7 @@ class BaseHTTPResponse(io.IOBase):
         return b""
 
     # Compatibility methods for `io` module
-    def readinto(self, b: bytearray) -> int:
+    def readinto(self, b: bytearray | memoryview[int]) -> int:
         temp = self.read(len(b))
         if len(temp) == 0:
             return 0
@@ -821,7 +821,7 @@ class HTTPResponse(BaseHTTPResponse):
     def tell(self) -> int:
         """
         Obtain the number of bytes pulled over the wire so far. May differ from
-        the amount of content returned by :meth:``urllib3.response.HTTPResponse.read``
+        the amount of content returned by :meth:`HTTPResponse.read`
         if bytes are encoded on the wire (e.g, compressed).
         """
         return self._fp_bytes_read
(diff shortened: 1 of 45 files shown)


sg3-141-592:
Sure, we've now got a specific single exception for mixed `decode_content` in `read()`.

This is already tested by existing test `test_response::test_read_with_illegal_mix_decode_toggle`.

And now setting the `cache_content` to false when doing a partial read (can't cache incomplete data)

src/urllib3/response.py
```diff
@@ -1121,17 +1122,31 @@ def read(
...
+
+            # If there's buffered data, combine it with newly decoded data
+            if decode_content and len(self._decoded_buffer) > 0:
+                self._decoded_buffer.put(decoded_data)
+                data = self._decoded_buffer.get_all()
+
+                # We cannot cache this read because it is a mix of buffered data
+                # (from a previous partial read) and the remainder of the stream.
+                # It does not represent the full HTTP body.
+                cache_content = False
```
illia-v, line 1133: This still leaves a room for caching only a part of the content after a partial read because an empty `self._decoded_buffer` does not mean that partial reads haven't occurred, check a suggestion for adding `self._has_uncached_read` in https://github.com/urllib3/urllib3/pull/3647#discussion_r2250139343.

src/urllib3/response.py
```diff
@@ -627,11 +627,6 @@ def _decode(
         Decode the data passed in and potentially flush the decoder.
         """
         if not decode_content:
-            if self._has_decoded_content:
-                raise RuntimeError(
-                    "Calling read(decode_content=False) is not supported after "
-                    "read(decode_content=True) was called."
-                )
```
illia-v, line 634: This check was not redundant, `self._decode` is called in multiple places that still rely on the check

illia-v:
The new tests from this PR no longer reveal the issue on the main branch and the suggested fix is not complete, I'm closing this PR in favor of #4960

PR #3728, opened 2025-12-04
Debug coverage issue when dropping 3.9

pquentin:
This should help me understand the code coverage issue from https://github.com/urllib3/urllib3/pull/3720#issuecomment-3612329364 better.

### DIFF
--- a/src/urllib3/contrib/socks.py
+++ b/src/urllib3/contrib/socks.py
@@ -142,6 +142,7 @@ class SOCKSConnection(HTTPConnection):
                         self, f"Failed to establish a new connection: {error}"
                     )
             else:
+                raise ValueError("debug 2")
                 raise NewConnectionError(
                     self, f"Failed to establish a new connection: {e}"
                 ) from e
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -736,6 +736,7 @@ class HTTPResponse(BaseHTTPResponse):
                     # SSL errors related to framing/MAC get wrapped and reraised here
                     raise SSLError(e) from e
 
+                raise ValueError("debug")
                 raise ReadTimeoutError(self._pool, None, "Read timed out.") from e  # type: ignore[arg-type]
 
             except IncompleteRead as e:


illia-v:
https://github.com/urllib3/urllib3/issues/2386 seems to be related to the socks line

illia-v:
We use different coverage versions for Python 3.9 and next versions, maybe coverage 7.10.7 used for 3.9 detects the line as covered even if it's never touched

src/urllib3/response.py
```diff
@@ -736,6 +736,7 @@ def _error_catcher(self) -> typing.Generator[None]:
                     # SSL errors related to framing/MAC get wrapped and reraised here
                     raise SSLError(e) from e
 
+                raise ValueError("debug")
                 raise ReadTimeoutError(self._pool, None, "Read timed out.") from e  # type: ignore[arg-type]
```
illia-v, line 740: I've researched this case: CPython has been raising `socket.timeout` instead of a generic `SSLError` since 2010 https://github.com/python/cpython/commit/c4df7845143f9afe0d20f4421a41904f3cbb991a
We can drop this unreachable line.

src/urllib3/contrib/socks.py
```diff
@@ -142,6 +142,7 @@ def _new_conn(self) -> socks.socksocket:
                         self, f"Failed to establish a new connection: {error}"
                     )
             else:
+                raise ValueError("debug 2")
                 raise NewConnectionError(
                     self, f"Failed to establish a new connection: {e}"
                 ) from e
```
illia-v, line 148: This is not reached since Python 3.3 when [`socket.error`](https://docs.python.org/3/library/socket.html#socket.error) became an alias of `OSError`.

Found by exploring traceback from one of our tests:
```
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 809, in connect
    negotiate(self, dest_addr, dest_port)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 443, in _negotiate_SOCKS5
    self.proxy_peername, self.proxy_sockname = self._SOCKS5_request(
                                               ~~~~~~~~~~~~~~~~~~~~^
        self, CONNECT, dest_addr)
        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 503, in _SOCKS5_request
    raise SOCKS5AuthError("SOCKS5 authentication failed")
socks.SOCKS5AuthError: SOCKS5 authentication failed

  File "/urllib3/src/urllib3/contrib/socks.py", line 110, in _new_conn
    conn = socks.create_connection(
        (self.host, self.port),
    ...<7 lines>...
        **extra_kw,
    )
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 209, in create_connection
    raise err
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 199, in create_connection
    sock.connect((remote_host, remote_port))
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 47, in wrapper
    return function(*args, **kwargs)
  File "/urllib3/.venv/lib64/python3.14/site-packages/socks.py", line 814, in connect
    raise GeneralProxyError("Socket error", error)
socks.GeneralProxyError: Socket error: SOCKS5 authentication failed
```

Check this PySocks code
https://github.com/Anorov/PySocks/blob/c2fa43cbe1091e799e248e8e4433978916791a8b/socks.py#L806-L820
I guess instances of `socks.ProxyError` subclasses were caught as `ProxyError` before Python 3.3.
But now `socks.ProxyError` is caught as `socket.error`, so it's always converted to `GeneralProxyError` which always has the `socket_err` value that is checked by urllib3 here.

```
>>> socket.error
<class 'OSError'>
>>> issubclass(socks.ProxyError, socket.error)
True
```

illia-v: But I would choose to add `# Defensive:` instead of dropping the unreached lines to be safe in the unlikely event that PySocks fixes the issue.

Alternatively, we can adapt urllib3 to the PySocks bug and make the line reachable by adding something like this before the checks
```python
if isinstance(e, socks.GeneralProxyError) and instance(e.socket_err, socks.ProxyError):
    e = e.socket_err
```
But it's a riskier approach.

PR #3729, opened 2025-12-05
Update ConnectionPool Queue to support gevent monkey_patch

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent
See:

sg3-141-592:
Fix for #3289 

Do late binding of `queue.LifeQueue` via a metaclass that allows it intercept accesses to `QueueCls` and delay the import of `queue` until then.

You can test the behaviour like this. I've not added this as a new test yet, because there's currently no gevent tests and it's a bit different to other tests. Happy to add this as a test if it makes sense in the urllib3 test suite.

``` python
import urllib3
from urllib3.connectionpool import HTTPConnectionPool

print("Before monkey-patch:")
print(f"  QueueCls: {HTTPConnectionPool.QueueCls}")
print(f"  Module: {HTTPConnectionPool.QueueCls.__module__}")

# Now monkey-patch (we only care about queue, not events)
import gevent.monkey
gevent.monkey.patch_queue()

print("\nAfter monkey-patch:")
print(f"  QueueCls: {HTTPConnectionPool.QueueCls}")
print(f"  Module: {HTTPConnectionPool.QueueCls.__module__}")

import queue
print(f"\nWhat queue.LifoQueue points to now:")
print(f"  queue.LifoQueue: {queue.LifoQueue}")
print(f"  Module: {queue.LifoQueue.__module__}")
```


illia-v:
Thanks for the contribution!

> You can test the behaviour like this. I've not added this as a new test yet, because there's currently no gevent tests and it's a bit different to other tests. Happy to add this as a test if it makes sense in the urllib3 test suite.

It would be nice if this is covered by a test. We can consider adding a new test job to our CI suite if needed, we have `test_brotlipy`, for example.

sg3-141-592:
> Thanks for the contribution!
> 
> > You can test the behaviour like this. I've not added this as a new test yet, because there's currently no gevent tests and it's a bit different to other tests. Happy to add this as a test if it makes sense in the urllib3 test suite.
> 
> It would be nice if this is covered by a test. We can consider adding a new test job to our CI suite if needed, we have `test_brotlipy`, for example.

Sure, just added new test for the issue `test/test_connectionpool.py::TestConnectionPoolGevent` and fixed the coverage

test/test_connectionpool.py
```diff
@@ -592,3 +592,65 @@ def test_read_timeout_0_does_not_raise_bad_status_line_error(self) -> None:
...
                     pool._make_request(conn, "", "", timeout=timeout)
+
+
+class TestConnectionPoolGevent:
+
+    def test_queuecls_late_binding_for_gevent(self) -> None:
+        """
+        Test that HTTPConnectionPool uses gevent monkey-patched queues
+        """
+        gevent = pytest.importorskip("gevent", reason="gevent not installed")
```
illia-v, line 603: I don't think gevent is ever installed in CI
Do you want us to help you set up a new CI job?

illia-v: Maybe, if possible, we can apply the gevent patching to all tests in the separate job somehow

sg3-141-592: I've added a new stage to nox `test_gevent` that runs the disabled gevent tests that were missing from the pipeline.

I've had to remove a historic test from this issue https://github.com/urllib3/urllib3/pull/1061 where urllib3 used to block people trying to gevent monkeypatch the queue because of an issue with `six` and `future`. This test isn't relevant anymore.

test/test_connectionpool.py
```diff
@@ -592,3 +592,65 @@ def test_read_timeout_0_does_not_raise_bad_status_line_error(self) -> None:
...
+
+        # HTTPConnectionPool.QueueCls should now resolve to the PATCHED version
+        current_queuecls: typing.Any = HTTPConnectionPool.QueueCls
+
+        # This is what the fix enables: late binding
+        assert (
+            current_queuecls is patched_lifoqueue
+        ), "HTTPConnectionPool.QueueCls should resolve to the patched queue.LifoQueue"
+
+    def test_queuecls_custom_override(self) -> None:
```
illia-v: Do we need any gevent calls in this test function since it's in a test class for gevent?

illia-v, line 636: This test doesn't use gevent, let's move it to `TestConnectionPool`

test/test_connectionpool.py
illia-v: Thanks for adding a CI job!

> I've had to remove a historic test from this issue #1061 where urllib3 used to block people trying to gevent monkeypatch the queue because of an issue with `six` and `future`. This test isn't relevant anymore.

Yes, let's drop the obsolete test.

If I understand correctly the test failed because `TestConnectionPoolGevent` implicitly kept the applied gevent monkey patching for all tests following it, and the previous tests ran without the monkey patching.
We can enable monkey patching for all tests with a custom plugin to be consistent, but the logic of `test_queuecls_late_binding_for_gevent` would need to be changed.

Note, all other tests pass even without the new metaclass with the monkey patching.

```diff
diff --git a/noxfile.py b/noxfile.py
index cc510aa9..2e413927 100644
--- a/noxfile.py
+++ b/noxfile.py
@@ -148,8 +148,8 @@ def test_gevent(session: nox.Session) -> None:
     session.env["UV_PROJECT_ENVIRONMENT"] = session.virtualenv.location
     tests_impl(
         session,
-        extras="socks,brotli,zstd,h2",
         extra_dependencies=["gevent"],
+        pytest_extra_args=["-p", "test.gevent_plugin"],
     )
 
 
diff --git a/test/gevent_plugin.py b/test/gevent_plugin.py
new file mode 100644
index 00000000..3106b83f
--- /dev/null
+++ b/test/gevent_plugin.py
@@ -0,0 +1,16 @@
+"""
+Pytest plugin to apply gevent monkey patching before any test collection.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from gevent import monkey  # type: ignore[import-untyped]
+
+if TYPE_CHECKING:
+    import pytest
+
+
+def pytest_configure(config: pytest.Config) -> None:
+    monkey.patch_queue()

```

illia-v:
I did try to make this approach work, but at some point I realized it was quite difficult to get mypy to accept the type hints.

We ended up choosing the simpler approach from #5057. It may not cover some of the more complex cases where metaclasses really shine, but it should be good enough for most users.

PR #3732, opened 2025-12-08
Restore `getheaders()` and `getheader()`

### LINKED ISSUE #3731: Removal of getheaders changes api, possibly breaking others
### Subject

The recent change to 2.6.0, included this change:

Removed the HTTPResponse.getheaders() method in favor of HTTPResponse.headers. Removed the HTTPResponse.getheader(name, default) method in favor of HTTPResponse.headers.get(name, default). (https://github.com/urllib3/urllib3/issues/3622)

This API change will most likely impact other libraries from easily updating to this version. Its not clear why a method would be removed in favour of using a variable. Couldn't both be supported so it doesn't break users of this object?

illia-v:
Closes #3731

### DIFF
--- a/changelog/3731.feature.rst
+++ b/changelog/3731.feature.rst
@@ -1,0 +1,2 @@
+Restore previously removed ``HTTPResponse.getheaders()`` and
+``HTTPResponse.getheader()`` methods.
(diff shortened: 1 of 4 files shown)


illia-v:
I'm merging this without new deprecations to let dependent libraries skip unnecessary changes

PR #3737, opened 2025-12-10
Document deprecation policy

EpicWink:
As the earliest of:

* the next major version
* after 6 minor versions
* after 30 months

This simply reflects the reality of the (initial, now reverted) removal of `response.getHeader`(`s`). You may want to remove one or two of the options.

I would actually propose a policy of: deprecate for at least 2 years for low-impact changes, and 5 years for high-impact changes.

### DIFF
--- a/docs/deprecation-policy.rst
+++ b/docs/deprecation-policy.rst
@@ -2,22 +2,22 @@
 API stability and deprecation policy
 ====================================
 
-1. ``urllib3`` will not wait until a "major" version release (e.g., 3.0.0,
-   4.0.0, etc.) to make a backwards incompatible change, but we will endeavor
-   to pick a release that we anticipate to be far enough in the future
+* urllib3 will not wait until a "major" version release (e.g., 3.0.0, 4.0.0,
+  etc.) to make a backwards incompatible change, but we will endeavor
+  to pick a release that we anticipate to be far enough in the future
 
-2. A release that is "far enough in the future" may be 4 or more "minor"
-   version releases (e.g., announced when releasing 3.0.1 and "broken" in 3.4.0)
+* A release that is "far enough in the future" may be 4 or more "minor"
+  version releases (e.g., announced when releasing 3.0.1 and "broken" in 3.4.0)
 
-3. urllib3 also reserves the right to push that further out if the release ends
-   up coming up more quickly than anticipated by the maintainers (e.g., if in
-   3.0.1 we announced a breaking change in 3.4.0, and felt we had not given
-   sufficient time prior to releasing 3.3.0, we may change the warning to
-   communicate the breaking change may happen in 3.5.0 or 3.7.0)
+* urllib3 also reserves the right to push that further out if the release ends
+  up coming up more quickly than anticipated by the maintainers (e.g., if in
+  3.0.1 we announced a breaking change in 3.4.0, and felt we had not given
+  sufficient time prior to releasing 3.3.0, we may change the warning to
+  communicate the breaking change may happen in 3.5.0 or 3.7.0)
 
-4. All anticipated breaking changes will be:
+* All anticipated breaking changes will be:
 
-   a. Documented here at https://urllib3.readthedocs.io/
-   b. Included in release notes at the time of decision
-   c. Re-announced the minor release prior to planned breaking changes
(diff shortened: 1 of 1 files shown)


sigmavirus24:
How do you determine impact? By how irresponsible your least responsible user is for not heeding deprecation warnings? Not every code as is open source or even easily searchable? How do you determine the severity of impact in that case? 

Let's be realistic. Deprecations are best effort. I disagree with adding back the recently removed methods, but none of the downstream projects cared enough to address warnings for the last 2.5 years. At some point technical debt needs to be driven down, regardless of who is ignoring years of warnings

EpicWink:
> How do you determine impact?

I don't. It's a subjective decision made by the maintainers. I would imagine a symbol's importance (eg `PoolManager` is more important than `PoolKey.key_ssl_maximum_version`), and its usage in public repositories (ie number of hits in a [GitHub search](https://github.com/search)), would be taken into account.

> By how irresponsible your least responsible user is for not heeding deprecation warnings?

No, I don't think it should be based on users' capability to respond to deprecations. Viewing users who don't respond as "irresponsable" is an incredibly pessimistic view of the world, and I suggest taking a step back and reevaluating what's important here.

I think there are likely reasons for not responding, such as not even knowing about the deprecation warning (docs are a useless way of broadcasting notices, and deprecation warnings are not shown by default when invoking Python - only enabled during testing), or not having the time to support downstream open-source projects.

> Not every code as is open source or even easily searchable.

I think the sample which are the searchable open source projects is enough to get a reasonable estimate of the usage, and it's unrealistic to expect knowledge of inaccessible projects.

> How do you determine the severity of impact in that case?

I don't think the severity needs to be determined: as I said above, it's subjective.

---

@sigmavirus24 do you have any comments on the pull-request's patch as it stands? I strongly believe in documenting (and adhering to) some deprecation policy. I do think you maintainers need to make a choice between "6 minor releases" and "2 years".

sigmavirus24:
Yes, yes, let's focus on the patch, not the topic of uninvited contributions around "governance" after an onslaught of FUD due to a change made after 2.5 years of notice.

Focusing on the patch as it stands:

1. The Changelog is the wrong place for this
2. Release numbers are cheap and calling them "minor" releases implies SemVer (to most people as the vast majority don't even know what PEP 440 is) which we're not following so we _could_ go from 2.6.1 to 2.12.0 in the course of the next year if we had enough reason to do so. I doubt anyone would be comfortable with such a short deprecation period.

I would say the deprecation policy should be documented elsewhere and that it should state clearly:

1. urllib3 will not wait until a "major" version release (e.g., 3.0.0, 4.0.0, etc.) to make a backwards incompatible change, **but** we will endeavor to pick a release that we anticipate to be far enough in the future
2. A release that is "far enough in the future" may be 4 or more "minor" version releases (e.g., announced when releasing 3.0.1 and "broken" in 3.4.0)
3. urllib3 also reserves the right to push that further out if the release ends up coming up more quickly than anticipated by the maintainers (e.g., if in 3.0.1 we announced a breaking change in 3.4.0, and felt we had not given sufficient time prior to releasing 3.3.0, we may change the warning to communicate the breaking change may happen in 3.5.0 or 3.7.0)
4. All anticipated breaking changes will be:
   a. Documented at urllib3.readthedocs.io
   b. Included in release notes at the time of decision
   c. Re-announced the minor release prior to planned breaking changes
   d. Communicated via sponsorship channels


But I'm one person tired of people ignoring warnings in tests and local development and painting a group of people doing their best as villains, even after they reverted the breaking change everyone else had 2.5 years to address.

docs/deprecation-policy.rst
```diff
@@ -0,0 +1,23 @@
+====================================
+API stability and deprecation policy
+====================================
+
+1. ``urllib3`` will not wait until a "major" version release (e.g., 3.0.0,
```
hugovk:
```suggestion
1. urllib3 will not wait until a "major" version release (e.g., 3.0.0,
```

First item can be "urllib3 doesn't follow Semantic Versioning".

(And use bullets, not numbers: https://asrfs.github.io/Tech-Writing-for-Techies/output/html/02-03-lists-instructions.html#_bullet_or_numbered)

PR #3743, opened 2025-12-16
Handle massive values in Retry-After when calculating time to sleep for

meeb:
The method `Retry.parse_retry_after(...)` will accept any integer returned by the remote server in the `Retry-After` header when the server returns a number of seconds to delay for rather than an HTTP-date. This includes massive and invalid values, for example `Retry-After: 9999999999`. This value is passed back to `Retry.sleep_for_retry(...)` which then uses the value directly in `time.sleep(...)`. The type is checked but not the value.

Using values for `time.sleep(...)` that exceed `PyTime_MAX` results in an OverflowError being raised:

```python
>>> time.sleep(9999999999)
  File "<python-input-3>", line 1, in <module>
    time.sleep(9999999999)
    ~~~~~~~~~~^^^^^^^^^^^^
OverflowError: timestamp too large to convert to C PyTime_t
```

This minimal PR adds a sanity check to confirm that the value of the `Retry-After` header is < 2147483647 (2 ** 31 - 1) in value. While `PyTime_MAX` is a `INT64_MAX` this is still a relatively sane upper bound value. If the value exceeds 2147483647 a clearer and more descriptive exception is raised.

This issue was seen in the wild:

https://github.com/meeb/whoisit/issues/60

It's highly unlikely that actually sleeping for decades or centuries is desirable but that's a different issue. I don't believe there's a specified sensible maximum for `Retry-After` in the RFC, perhaps limiting this to `3600` or similar lower value would be sensible?

Currently, even with this PR, a semi-malicious server could add `if ($http_user_agent ~ urllib3) { add_header retry-after 5000000000; }` and lock up some clients that attempt to handle retries properly for over a century (or at least until the process gets killed).

### DIFF
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -180,8 +180,9 @@ class Retry:
         request.
 
     :param int retry_after_max: Number of seconds to allow as the maximum for
-        Retry-After headers. Defaults to 6 hours. Any Retry-After headers larger
-        than this value will be limited to this value.
+        Retry-After headers. Defaults to :attr:`Retry.DEFAULT_RETRY_AFTER_MAX`.
+        Any Retry-After headers larger than this value will be limited to this
+        value.
     """
 
     #: Default methods to be used for ``allowed_methods``


illia-v (commented):
Thank you for the contribution!

We can add a default limit similar to the existing `DEFAULT_BACKOFF_MAX` and a parameter to configure the limit

https://github.com/urllib3/urllib3/blob/83f8643ffb5b7f197457379148e2fa118ab0fcdc/src/urllib3/util/retry.py#L196-L197

https://github.com/urllib3/urllib3/blob/83f8643ffb5b7f197457379148e2fa118ab0fcdc/src/urllib3/util/retry.py#L213

src/urllib3/util/retry.py
```diff
@@ -320,6 +320,10 @@ def parse_retry_after(self, retry_after: str) -> float:
 
         seconds = max(seconds, 0)
 
+        # Check the seconds do not exceed (32bit) PyTime_MAX to avoid an OverflowError
+        if seconds > 2147483647:
```
sigmavirus24: I don't like magic numbers. Can we set it as a `t.Final[int]` at the module level and calculate it as `2**(32-1) - 1` or at least explain that's how it's calculated here. 32bit doesn't mean much to most folks writing python as they won't understand if its signed or unsigned. Alternatively, leave a comment explaining why we "must" set it ourselves here. And use a more humane way of writing the number, e.g., `2_147_483_648`

meeb:
Reasonable suggestions. I've reworked it so the value is configurable and added a basic test in the format of the other tests. The magic number is currently set statically as `2_147_483_648` but with a comment as to why it's that value. I'm not a big fan of magic numbers in Python either, however this is the value used internally by CPython (at least in 32bit builds anyway) that when exceeded triggers the `OverflowError`. If this is just a sanity check feature this is the value it should be.

I don't mind adding an additional `if retry_after > x: retry_after = x` limit feature to stop remote sites from triggering `sleep(...)` for retries over a specific amount of time as a different PR, but that would likely require some discussion on what `x` should be first.

illia-v:
> I don't mind adding an additional `if retry_after > x: retry_after = x` limit feature to stop remote sites from triggering `sleep(...)` for retries over a specific amount of time as a different PR, but that would likely require some discussion on what `x` should be first.

I should clarify that I meant the suggested limit to be used either for capping the header value or for raising an error if it is too big.

I can see that curl limits `Retry-After` to 6 hours, the limit is pretty sensible
https://github.com/curl/curl/blob/6d0ee7b17b0c37c034a78314eda4148104187689/lib/http.c#L3446-L3449
https://github.com/curl/curl/pull/15833

We won't need to check whether it is less than `PyTime_t` in this case.

meeb:
Yep that sounds sensible, thanks for picking `x`! I've removed the `PyTime_t` magic and this now just limits the headers to 6 hours without raising `InvalidHeader` as suggested.

src/urllib3/util/retry.py
```diff
@@ -211,6 +215,7 @@ def __init__(
         status_forcelist: typing.Collection[int] | None = None,
         backoff_factor: float = 0,
         backoff_max: float = DEFAULT_BACKOFF_MAX,
+        retry_after_max: int = DEFAULT_RETRY_AFTER_MAX,
```
illia-v: Please move it to be the last parameter to keep the signature of `__init__` compatible with previous versions

illia-v: Also, we'll need to document the parameter in the Retry docstring, the description will appear in the docs
https://urllib3--3743.org.readthedocs.build/en/3743/reference/urllib3.util.html#urllib3.util.Retry

test/test_retry.py
```diff
@@ -181,6 +181,18 @@ def test_configurable_backoff_max(self) -> None:
         retry = retry.increment(method="GET")
         assert retry.get_backoff_time() == max_backoff
 
+    def test_configurable_retry_after_max(self) -> None:
+        """Configurable retry after is computed correctly"""
+        max_retry_after = Retry.DEFAULT_RETRY_AFTER_MAX
+
+        retry = Retry(retry_after_max=max_retry_after)
```
illia-v: Let's test it without the argument

```suggestion
        retry = Retry()
```

src/urllib3/util/retry.py
```diff
@@ -196,6 +196,10 @@ class Retry:
     #: Default maximum backoff time.
     DEFAULT_BACKOFF_MAX = 120
 
+    #: Default maximum allowed value for Retry-After headers in seconds
+    # This is undocumented in the RFC. Setting to 6 hours matches other popular libraries.
```
illia-v: When the comments are in the current order, the constant doesn't appear in the docs for some reason
https://urllib3--3743.org.readthedocs.build/en/3743/reference/urllib3.util.html#urllib3.util.Retry

```suggestion
    # This is undocumented in the RFC. Setting to 6 hours matches other popular libraries.
    #: Default maximum allowed value for Retry-After headers in seconds
```

changelog/3743.bugfix.rst
```diff
@@ -0,0 +1 @@
+Handle very large values for Retry-After HTTP headers in ``urllib3.utils.Retry.parse_retry_after()`` when calculating how long to sleep between retries.
```
illia-v: Let's be explicit about the new behavior 🙂 


```suggestion
Started treating ``Retry-After`` times greater than 6 hours as 6 hours.
```

meeb:
Thanks for the review. I'll adjust as requested in a couple of days.

meeb:
Your suggested changes have been made, this should be good to go.

PR #3744, opened 2025-12-16
Reinstate warnings for `getheaders` and `getheader` and promote to `FutureWarning`

hugovk:
Opening this as a suggestion and for discussion.

This PR:

* Reinstates the warnings removed in [#3732](https://github.com/urllib3/urllib3/pull/3732)
* Promotes from `DeprecationWarning` to `FutureWarning`
* Sets an arbitrary removal version of 2.7.0, can of course be adjusted (although I recommend advertising a target version)

Part of the problem with the earlier removal (https://github.com/urllib3/urllib3/issues/3731) was, despite years of `DeprecationWarning`, many projects had not noticed and acted upon the warnings.

[`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) is often hidden:

> Base class for warnings about deprecated features when those warnings are intended for other Python developers.
> 
> Ignored by the default warning filters, except in the __main__ module ([PEP 565](https://peps.python.org/pep-0565/)). Enabling the [Python Development Mode](https://docs.python.org/3/library/devmode.html#devmode) shows this warning.

However, one of @sethmlarson's suggestions in https://sethmlarson.dev/deprecations-via-warnings-dont-work-for-python-libraries is for projects to create their own warning class like `Urllib3DeprecationWarning(UserWarning)`, which would be more visible.

Another suggestion is to use [`FutureWarning`](https://docs.python.org/3/library/exceptions.html#FutureWarning):

> Base class for warnings about deprecated features when those warnings are intended for end users of applications that are written in Python.

This would be more visible to users of urllib3, with the hope they will finally take action and update their code.

### DIFF
--- a/changelog/3744.removal.rst
+++ b/changelog/3744.removal.rst
@@ -1,0 +1,1 @@
+Reinstate warnings for ``getheaders`` and ``getheader`` and promote to ``FutureWarning``.


hugovk:
Yep, feel free to close this if you decide to keep the getters after all. This PR was in case you wanted to reinstate them and also to illustrate this method of stepping up deprecations. Using `FutureWarning` with some of the other deprecations sounds like a good idea.

Thanks!

pquentin:
Thanks @hugovk! We should definitely switch to `FutureWarning` for the other warnings. TIL. Closing this PR for now though.

PR #3747, opened 2025-12-18
Move "v2.0 Migration Guide" to the end of the table of contents

illia-v:
It's expected that fewer users need the migration guide nowadays, let's move it from the top of the documentation navbar and put it after the changelog.

image

### DIFF
--- a/changelog/3752.bugfix.rst
+++ b/changelog/3752.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``urllib3.connection.VerifiedHTTPSConnection`` on Emscripten.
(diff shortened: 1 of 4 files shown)


illia-v:
> Maybe that should be v2 too now instead of v2.0?

Yes, that's a great idea!

I updated a documented exception text to the current one too:
https://github.com/urllib3/urllib3/blob/13f0bfd55e4468fe1ea9c6f809d3a87b0f93ebab/src/urllib3/__init__.py#L42-L46

pquentin (approved):
Thanks! LGTM.

PR #3751, opened 2025-12-22
add Proxy Response Header section to advanced-usage.rst

proxymesh:
to mention ProxyHeaderManager


sigmavirus24:
I don't think we need to mention this in our documentation. We're not specifically endorsing this so there's little reason to include it

proxymesh:
I understand. Is there some other location that would be more appropriate to add this?

PR #3752, opened 2025-12-28
Patch VerifiedHTTPSConnection for emscripten

juntyr:
There is always some downstream project relying on outdated names, in this case `botocore`. With this fix it also gets to use the patched https connection for emscripten

### DIFF
--- a/src/urllib3/contrib/emscripten/connection.py
+++ b/src/urllib3/contrib/emscripten/connection.py
@@ -40,7 +40,7 @@ class EmscriptenHTTPConnection:
     is_verified: bool = False
     proxy_is_verified: bool | None = None
 
-    response_class: BaseHTTPResponse = EmscriptenHttpResponseWrapper
+    response_class: type[BaseHTTPResponse] = EmscriptenHttpResponseWrapper
     _response: EmscriptenResponse | None
 
     def __init__(


juntyr:
I also noticed that `EmscriptenHTTPConnection` doesn't have the `response_class` property, another property that `botocore` relies on existing. What should that be set to?

illia-v:
> I also noticed that `EmscriptenHTTPConnection` doesn't have the `response_class` property, another property that `botocore` relies on existing. What should that be set to?

Does it work with `response_class = EmscriptenHttpResponseWrapper`?

juntyr:
> > I also noticed that `EmscriptenHTTPConnection` doesn't have the `response_class` property, another property that `botocore` relies on existing. What should that be set to?
> 
> Does it work with `response_class = EmscriptenHttpResponseWrapper`?



> > I also noticed that `EmscriptenHTTPConnection` doesn't have the `response_class` property, another property that `botocore` relies on existing. What should that be set to?
> 
> Does it work with `response_class = EmscriptenHttpResponseWrapper`?

That sounds correct!

`botocore` doesn't actually do anything with the property other than backing it up and restoring it (so I was able to patch in `None` and the code worked), but given that `response_class` should be the class that is returned by `getresponse`, and we return a `EmscriptenHttpResponseWrapper` there, this makes sense.

PR #3753, opened 2025-12-31
Tests: clarify pooling behavior in test_many_urls

ADarcside:
Adds a brief comment explaining why multiple URLs result in five pooled connections. No functional changes.


ADarcside:
This change only adds a clarifying comment to a test and does not affect user-facing behavior.
Could a maintainer please apply the Skip Changelog label?

PR #3756, opened 2026-01-07
Clarify the mechanism that enables Nagle's algorithm for proxies

akx:
Just a simple little comment change since it isn't immediately obvious why `setdefault(..., [])` would enable Nagle's algorithm (i.e. disable `TCP_NODELAY`).


illia-v (approved):
LGTM, thanks!

PR #3763, opened 2026-01-20
Switch deprecation warnings to `FutureWarning` and reschedule removals

illia-v:
The codebase contains a few deprecation warnings with unmet target versions.
We've recently discovered that warnings using `DeprecationWarning` were not seen by most users.

This PR proposes switching from `DeprecationWarning` to `FutureWarning`, the later should be visible by default. Python will print the first occurrence of matching warnings for each location.

Here's an example of a script using a deprecated function and its output:
```python
from urllib3.fields import format_header_param_rfc2231

def warn():
    print(format_header_param_rfc2231("filename", "test.txt"))
    print(format_header_param_rfc2231("filename", "test.pdf"))

if __name__ == "__main__":
    warn()
    warn()
```

```
$ python format_header.py 
./format_header.py:4: FutureWarning: 'format_header_param_rfc2231' is insecure, deprecated and will be removed in urllib3 v3.0. This is not valid for multipart/form-data header parameters.
  print(format_header_param_rfc2231("filename", "test.txt"))
filename="test.txt"
./format_header.py:5: FutureWarning: 'format_header_param_rfc2231' is insecure, deprecated and will be removed in urllib3 v3.0. This is not valid for multipart/form-data header parameters.
  print(format_header_param_rfc2231("filename", "test.pdf"))
filename="test.pdf"
filename="test.txt"
filename="test.pdf"
```

Related references:
https://sethmlarson.dev/deprecations-via-warnings-dont-work-for-python-libraries
https://github.com/urllib3/urllib3/pull/3744
https://github.com/urllib3/urllib3/issues/3731

### DIFF
--- a/docs/v2-migration-guide.rst
+++ b/docs/v2-migration-guide.rst
@@ -184,9 +184,9 @@ Next you should try installing urllib3 v2.x locally and run your test suite.
 
   $ python -m pip install -U 'urllib3>=2'
 
-Because there are new ``DeprecationWarnings`` you should ensure that you're
+Because there are new ``FutureWarnings`` you should ensure that you're
 able to see those warnings when running your test suite. To do so you can add
-the following to your test setup to ensure even ``DeprecationWarnings`` are
+the following to your test setup to ensure even ``FutureWarnings`` are
 output to the terminal:
 
 .. code-block:: bash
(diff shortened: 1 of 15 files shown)


docs/v2-migration-guide.rst
```diff
@@ -205,17 +205,17 @@ or you can opt-in within your Python code:
   # You can change warning filters according to the filter rules:
   # https://docs.python.org/3/library/warnings.html#warning-filter
   import warnings
-  warnings.filterwarnings("default", category=DeprecationWarning)
+  warnings.filterwarnings("default", category=FutureWarning)
```
pquentin, line 208: Should we use the module parameter to only filter urllib3 warnings? That's not really related to the PR, but the new warnings will be more visible, so it makes sense me to be extra precise here.

illia-v: That sounds interesting, but I tried adding `warnings.filterwarnings("ignore", category=FutureWarning, module="urllib3")` to the example from the PR description, and it didn't silence the warnings. That happened because for warnings with `stacklevel=2`, a module name from the client code has to be set instead of `module="urllib3"`.

PR #3766, opened 2026-01-28
Patch/1.26.20

KiryaDtlv:


illia-v:
@KiryaDtlv appreciate the effort here!

However, we don't accept PRs for old release lines including 1.26.x as they are unsupported (see https://github.com/urllib3/urllib3?tab=security-ov-file#supported-versions). We're focusing all our maintenance efforts on the current 2.x.

PR #3770, opened 2026-02-10
Backport 8864ac407 for CVE-2026-21441

scallaway:
We currently aren't running 2.x in production and therefore rely on the 1.26.x branch.

This backports the fix for CVE-2026-21441 since I believe it's still present in the 1.x branch.


sigmavirus24:
https://github.com/urllib3/urllib3/pull/3766#issuecomment-3814349366

We are no longer supporting 1.26

PR #3771, opened 2026-02-10
Fix decoded buffer handling on full read

swjcpy:
## Summary
- Ensure HTTPResponse.read() returns buffered decoded bytes when amt=None
- Prevent data loss after a partial decoded read followed by full read
- Add regression test covering partial read then full read with decode_content=True

## Testing
- python -m pytest test/test_response.py -k " partial_read_then_read_all_returns_remaining\

### DIFF
--- a/changelog/3675.bugfix.rst
+++ b/changelog/3675.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed a bug where calling ``read(decode_content=True)`` after a partial decoded read
+could return truncated data.


illia-v:
Thanks for the contribution!

FYI, we have issue #3636 tracking the bug, and there are pull requests #3647 and #3726.
The issue is not that simple, and #3726 is the closest to resolving it. Perhaps, I'll help the author resolve the remaining comments if they don't respond soon.

PR #3772, opened 2026-02-11
docs: fix broken link in HTTPResponse.tell docstring

Juneezee:
https://urllib3.readthedocs.io/en/latest/reference/urllib3.response.html#urllib3.response.HTTPResponse.tell

The link to the `read()` method was using incorrect syntax. This PR fixes the formatting so the documentation renders correctly.

Before:

before

After (local preview with `nox -rs docs`):

after


illia-v (approved):
Thanks @Juneezee!

Issue #2206, opened 2021-05-05
Please consider switching SOCKS backend away from PySocks, e.g. to python-socks


mgorny:
### Context

I am packaging urllib3 and its dependencies on Gentoo. In my experience, the [PySocks](https://pypi.org/project/PySocks/) package used by urllib3 right now is in very bad shape.

The package has zero activity since Sep 2019. The tests are currently broken (again, I already had to apply one test fix in the past). By using this package, you're practically placing the burden of fixing it on distribution maintainers.

Could you possibly consider switching to another SOCKS implementation? I've recently had very positive experience with [python-socks](https://pypi.org/project/python-socks/) but anything that's maintained would be an improvement.

### Alternatives

Well, the alternative is to remove SOCKS support entirely for urllib3 in Gentoo entirely. While this isn't probably the most important feature of urllib3, I don't really like the idea of removing the feature.

### Duplicate

Didn't find anything.

### Contribution

I suppose I can try finding some time to give it a shot if you like the idea. However, I'd like to know first if it'd be accepted.

mgorny:
I suppose I could give more context: I'm currently testing Python packages in Gentoo against Python 3.10. However, I am unable to test PySocks because its tests work only with old version of tornado which doesn't work with Python 3.10 anymore.

sethmlarson:
Thanks for bringing this to our attention.  Instead of dropping PySocks has there been any recent attempt to contact @Anorov? I'd prefer going the route of keeping PySocks maintained instead of expecting all their users to migrate.

mgorny:
If you mean beyond the GitHub issues — I don't know. However, he doesn't seem to have any recent public GitHub activity.

That said, I think we're at the point when getting PySocks into really good standing (like, being able to actually run tests without Internet access and bundled precompiled 3proxy) is probably more effort than using a project that's already in good standing.

sigmavirus24:
To be clear "good standing" here only means "good standing" for Linux distros who insist on being able to test on air-gapped machines and not "good standing" in the sense that this project actually (probably) cares about.

sethmlarson:
I've reached out to @Anorov via email.

sethmlarson:
We've not made any headway and I haven't heard back from the maintainer. We should probably consider switching at this point.

azatoth:
[...]

PR #3774, opened 2026-02-15
Switch SOCKS backend from PySocks to python-socks (#2206)

rigens:
Switching SOCKS backend from PySocks to python-socks (issue #2206)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
         uses: browser-actions/setup-firefox@5914774dda97099441f02628f8d46411fcfbd208 # v1.7.0
         if: ${{ matrix.nox-session == 'emscripten(firefox)' }}
       - name: "Install node.js"
-        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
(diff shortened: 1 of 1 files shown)


rigens:
The `downstream (requests)` stage failed because `requests` directly depends on `PySocks` (not `urllib3[socks]`)

https://github.com/psf/requests/blob/da9113c0466850a425e554630a732ec1f7d045fc/pyproject.toml#L53

rigens:
Changelog was skipped because the changes are internal and not user-facing

sigmavirus24:
This is definitely user-facing as it impacts dependencies

pyproject.toml
```diff
@@ -67,7 +67,7 @@ dev = [
     "hypercorn",
     "idna>=3.10",
     "pyopenssl>=25.0.0",
-    "pysocks>=1.7.1",
+    "python-socks>=2.8.0,<3.0.0",
```
sigmavirus24: Why are you capping the version here?

rigens:
@sigmavirus24, changelog was updated

pyproject.toml
rigens: Fixed

sigmavirus24: I didn't ask for a fix.  I asked for an explanation

rigens: The explanation is simple - the next major release of `python-socks` may have breaking changes.
But as I can see, the `urllib3` developers follow the rule that dev dependencies should be more flexible. So I relaxed the `python-socks` dev dependency

PR #3780, opened 2026-02-27
Raise UnrewindableBodyError from rewind_body if seek missing

### LINKED ISSUE #3779: `rewind_body` raises `ValueError` when `seek` missing from body
### Subject

`rewind_body` should raise `UnrewindableBodyError` not `ValueError` when `seek` missing from body.
Current message `ValueError: body_pos must be of type integer, instead it was <class 'int'>` is confusing.

### Environment
```bash
$ uv run --with urllib3 env.py
OS macOS-15.5-arm64-arm-64bit-Mach-O
Python 3.13.1
OpenSSL 3.0.15 3 Sep 2024
urllib3 2.6.4.dev10
```

### Steps to Reproduce

```python
$ cat repro.py
"""Reproduction: urllib3 raises confusing ValueError on retry when body has tell() but no seek().

Run with: uv run --with urllib3 repro.py
"""
import urllib3


class TellableStream:
    """A stream with tell() and read() but no seek()."""

    def __init__(self, data: bytes) -> None:
        self._data = data
        self._pos = 0

    def read(self, n: int = -1) -> bytes:
        if n == -1:
            chunk = self._data[self._pos :]
            self._pos = len(self._data)
        else:
            chunk = self._data[self._pos : self._pos + n]
            self._pos += len(chunk)
        return chunk

    def tell(self) -> int:
        return self._pos


body = TellableStream(b"hello world")

pool = urllib3.HTTPConnectionPool("localhost", port=1)
pool.request(
    "POST",
    "/",
    body=body,
    retries=urllib3.Retry(total=2, connect=2),
    preload_content=True,
    decode_content=True,
)
$ uv run --with urllib3 repro.py
  File "/Users/chromy/Desktop/repro.py", line 31, in <module>
    pool.request(
    ~~~~~~~~~~~~^
        "POST",
        ^^^^^^^
    ...<4 lines>...
        decode_content=True,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/_request_methods.py", line 143, in request
    return self.request_encode_body(
           ~~~~~~~~~~~~~~~~~~~~~~~~^
        method, url, fields=fields, headers=headers, **urlopen_kw
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/_request_methods.py", line 278, in request_encode_body
           ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/connectionpool.py", line 871, in urlopen
    return self.urlopen(
           ~~~~~~~~~~~~^
        method,
        ^^^^^^^
    ...<13 lines>...
        **response_kw,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/connectionpool.py", line 761, in urlopen
    body_pos = set_file_position(body, body_pos)
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/util/request.py", line 150, in set_file_position
    rewind_body(body, pos)
    ~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/chromy/.cache/uv/archive-v0/vRvoZnLVVd6vwhcGJStRk/lib/python3.13/site-packages/urllib3/util/request.py", line 187, in rewind_body
    raise ValueError(
        f"body_pos must be of type integer, instead it was {type(body_pos)}."
    )
ValueError: body_pos must be of type integer, instead it was <class 'int'>.
```

### Expected Behavior

Should raise `UnrewindableBodyError`

### Actual Behavior

Raises:
```
ValueError: body_pos must be of type integer, instead it was <class 'int'>.
```
Which is confusing.

chromy:
Resolves #3779

### DIFF
--- a/changelog/3779.bugfix.rst
+++ b/changelog/3779.bugfix.rst
@@ -1,0 +1,1 @@
+Raised ``UnrewindableBodyError`` instead of ``ValueError`` when retrying a request whose body had ``tell()`` but not ``seek()``.
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -183,6 +183,8 @@ def rewind_body(body: typing.IO[typing.AnyStr], body_pos: _TYPE_BODY_POSITION) -
             "Unable to record file position for rewinding "
             "request body during a redirect/retry."
         )
+    elif body_seek is None:
+        raise UnrewindableBodyError("body does not implement seek.")
     else:
         raise ValueError(
             f"body_pos must be of type integer, instead it was {type(body_pos)}."
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -32,6 +32,7 @@ from urllib3.exceptions import (
     ReadTimeoutError,
     SSLError,
     TimeoutError,
+    UnrewindableBodyError,
 )
 from urllib3.response import HTTPResponse
 from urllib3.util.ssl_match_hostname import CertificateError
@@ -582,6 +583,53 @@ class TestConnectionPool:
         _test(HTTPException)
         _test(SocketError)
         _test(ProtocolError)
+
+    def test_retry_with_body_that_has_tell_but_no_seek(self) -> None:
+        """
+        This is a regression test for issue #3779 [1] where, if
+        seek is missing from body and we retried, we previously raised
+        a confusing error.
+
+        [1] <https://github.com/urllib3/urllib3/issues/3779>
+        """
+
+        class TellableStream:
+            """A stream-like object with tell() and read() but no seek()."""
+
+            def __init__(self, data: bytes) -> None:
+                self._data = data
+                self._pos = 0
+
+            def read(self, n: int = -1) -> bytes:
+                if n == -1:
+                    chunk = self._data[self._pos :]
+                    self._pos = len(self._data)
+                else:
+                    chunk = self._data[self._pos : self._pos + n]
+                    self._pos += len(chunk)
+                return chunk
+
+            def tell(self) -> int:
+                return self._pos
+
+        body = TellableStream(b"hello world")
+
+        with HTTPConnectionPool(host="localhost", maxsize=1) as pool:
+            with patch.object(
+                pool,
+                "_make_request",
+                side_effect=OSError("connection reset"),
+            ):
+                with pytest.raises(
+                    UnrewindableBodyError, match="body does not implement seek"
+                ):
+                    pool.urlopen(
+                        "POST",
+                        "/",
+                        body=body,  # type: ignore[arg-type]
+                        retries=Retry(total=2, allowed_methods=["POST"]),
+                        body_pos=None,
+                    )
 
     def test_read_timeout_0_does_not_raise_bad_status_line_error(self) -> None:
         with HTTPConnectionPool(host="localhost", maxsize=1) as pool:
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -622,6 +622,13 @@ class TestUtil:
         with pytest.raises(UnrewindableBodyError):
             rewind_body(BadSeek(), body_pos=2)
 
+    def test_rewind_body_no_seek(self) -> None:
+        class NoSeek(io.StringIO):
+            seek = None  # type: ignore[assignment]
+
+        with pytest.raises(UnrewindableBodyError):
+            rewind_body(NoSeek(), body_pos=2)
+
     def test_add_stderr_logger(self) -> None:
         handler = add_stderr_logger(level=logging.INFO)  # Don't actually print debug
         logger = logging.getLogger("urllib3")


src/urllib3/util/request.py
```diff
@@ -170,6 +170,11 @@ def rewind_body(body: typing.IO[typing.AnyStr], body_pos: _TYPE_BODY_POSITION) -
     :param int pos:
         Position to seek to in file.
     """
+    if body_pos is _FAILEDTELL:
+        raise UnrewindableBodyError(
+            "Unable to record file position for rewinding "
+            "request body during a redirect/retry."
+        )
```
illia-v: Please justify placing this at the very beginning

changelog/3779.bugfix.rst
```diff
@@ -0,0 +1 @@
+Raise ``UnrewindableBodyError`` instead of ``ValueError`` when retrying a request whose body has ``tell()`` but not ``seek()``.
```
illia-v: We write the entries in the past tense

```suggestion
Raised ``UnrewindableBodyError`` instead of ``ValueError`` when retrying a request whose body had ``tell()`` but not ``seek()``.
```

illia-v (commented):
Thanks for the contribution!

nateprewitt:
This seems totally reasonable to make clearer but I am struggling with the premise a bit.

The inputs we're putting in the test expressly violate the [typing contract](https://github.com/python/cpython/blob/e765696ca0c3421bb5763e4b944d5c1bcca2e84a/Lib/typing.py#L3555-L3565) (and I think this is the first case of this in ~10 years of the code existing). Whether the object can support seek(), it should implement it in some form per [BaseIO](https://docs.python.org/3/library/io.html#io.IOBase.seek). Is there more info on how this was encountered?

src/urllib3/util/request.py
chromy: I was finding the conditional a little hard to follow as I added another case so I ended up moving this one to the top. I've put it back with the existing cases now.

changelog/3779.bugfix.rst
chromy: Thanks!

chromy:
> This seems totally reasonable to make clearer but I am struggling with the premise a bit.
> 
> The inputs we're putting in the test expressly violate the [typing contract](https://github.com/python/cpython/blob/e765696ca0c3421bb5763e4b944d5c1bcca2e84a/Lib/typing.py#L3555-L3565) (and I think this is the first case of this in ~10 years of the code existing). Whether the object can support seek(), it should implement it in some form per [BaseIO](https://docs.python.org/3/library/io.html#io.IOBase.seek). Is there more info on how this was encountered?

Sorry, I thought I had put a link in the initial issue - but I hadn't. 

Background:
- https://github.com/getsentry/objectstore/pull/329
- https://github.com/indygreg/python-zstandard/pull/287
CleanShot 2026-03-11 at 07 19 49@2x

So this arose out the combination of:
- `getsentry/objectstore` passing a stream from `indygreg/python-zstandard` to `urllib3/urllib3`
- `indygreg/python-zstandard` incorrectly implementing the interface
- `urllib3/urllib3` retrying the request and attempting to rewind the body
- `urllib3/urllib3` correctly throwing an error - but that error being quite confusing.

I think the main issue is with `indygreg/python-zstandard`  which should correctly implement the type, I have https://github.com/indygreg/python-zstandard/pull/287 for this. However if I followed correctly the code here already is quite defensive and checks for the presence of `seek()` when that also would be required by the interface - it's just that error that's raised is a bit misleading so I also thought improving the error would be useful.

nateprewitt:
Thanks for the detailed reply, @chromy! That makes a ton of sense. As long as we're trying to fix the issue at the source, that's all I was looking for. Appreciate the contribution!

themavik (commented):
UnrewindableBodyError when seek is missing but tell works matches other rewind_body failures; pool retry test plus test_rewind_body_no_seek are tight. nit: StringIO subclass with seek=None is clever—if typeshed tightens io.StringIO.seek, keep the test explicit about why.

illia-v (approved):

PR #3781, opened 2026-02-27
Add possibility for overriding network address resolution

### LINKED ISSUE #3675: Add possibility to pass a custom `create_connection()` for sockets

NyanKiyoshi:
This allows to override CPython's default [`getaddrinfo()`] function which can be used to customize the resolution of domain names, hostnames, and IP addresses. (closes #3675)

For example, this useful for filtering IP addresses (this being the actual use-case behind the change), as well as the behavior.

Full context: in our instance (#3675) we weren't able to override the network resolution in urllib3 which meant that we had to result to hacks: we were pinning the IP address, e.g., `https://google.com` would become `https://1.2.4.0:443` **before** calling urllib3 which meant that we were only trying 1 IP address and not N addresses, thus this led to two issues:

1. We had a single point of failure (if IP no. 1 is down, then IP no. 2 isn't tried)
2. RFC 8305 (Happy Eyeballs v2) prefers IPv6 over IPv4, which meant resolving `localhost` returned `["::1", "127.0.0.1"]` which means the 1st IP to be tried to connect to is `::1` but some (not all) servers unfortunately only listen on `127.0.0.1` thus leading to being unable to connect to the service (it's easy to mitigate: just put the IP instead of `localhost`, but that's not viable for an OSS project like ours where users have to be told to not put `localhost`). We also cannot change resolution to always return `127.0.0.1` instead of `::1` as it's too case-specific, we prefer having a generic solution.

Note: there is no implementation for emscripten as it uses Node.js thus it's likely not feasible. I initially considered adding such code:

```py
class EmscriptenHTTPConnection:
    ...

    def __init__(..., resolver: Resolver = None):
        ...

        if resolver is not None:
            raise NotImplementedError(
                "Custom resolvers are not supported in emscripten"
            )
```

But I noticed that other parameters are also [already not implemented] thus I followed suite.

[`getaddrinfo()`]: https://docs.python.org/3/library/socket.html#socket.getaddrinfo
[already not implemented]: https://github.com/urllib3/urllib3/blob/fd4dffd2fc544166b76151a2fa3d7b7c0eab540c/src/urllib3/contrib/emscripten/connection.py#L64-L71

### DIFF
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -18,7 +18,7 @@ from .exceptions import (
     URLSchemeUnknown,
 )
 from .response import BaseHTTPResponse
-from .util.connection import _TYPE_SOCKET_OPTIONS
+from .util.connection import _TYPE_SOCKET_OPTIONS, Resolver
 from .util.proxy import connection_requires_http_tunnel
 from .util.retry import Retry
 from .util.timeout import Timeout
@@ -90,6 +90,7 @@ class PoolKey(typing.NamedTuple):
     key_assert_fingerprint: str | None
     key_server_hostname: str | None
     key_blocksize: int | None
+    key_resolver: Resolver | None
 
 
 def _default_key_normalizer(
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -18,8 +18,9 @@ from .exceptions import (
     URLSchemeUnknown,
 )
 from .response import BaseHTTPResponse
-from .util.connection import _TYPE_SOCKET_OPTIONS, Resolver
+from .util.connection import _TYPE_SOCKET_OPTIONS
 from .util.proxy import connection_requires_http_tunnel
+from .util.resolver import Resolver
 from .util.retry import Retry
 from .util.timeout import Timeout
 from .util.url import Url, parse_url
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -91,7 +91,6 @@ class PoolKey(typing.NamedTuple):
     key_assert_fingerprint: str | None
     key_server_hostname: str | None
     key_blocksize: int | None
-    key_resolver: Resolver | None
 
 
 def _default_key_normalizer(
@@ -202,9 +201,12 @@ class PoolManager(RequestMethods):
         self,
         num_pools: int = 10,
         headers: typing.Mapping[str, str] | None = None,
+        resolver: Resolver | None = None,
         **connection_pool_kw: typing.Any,
     ) -> None:
         super().__init__(headers)
+        self.resolver = resolver
+
         # PoolManager handles redirects itself in PoolManager.urlopen().
         # It always passes redirect=False to the underlying connection pool to
         # suppress per-pool redirect handling. If the user supplied a non-Retry
@@ -278,7 +280,7 @@ class PoolManager(RequestMethods):
             for kw in SSL_KEYWORDS:
                 request_context.pop(kw, None)
 
-        return pool_cls(host, port, **request_context)
+        return pool_cls(host, port, resolver=self.resolver, **request_context)
 
     def clear(self) -> None:
         """
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -91,6 +91,7 @@ class PoolKey(typing.NamedTuple):
     key_assert_fingerprint: str | None
     key_server_hostname: str | None
     key_blocksize: int | None
+    key_resolver: Resolver | None
 
 
 def _default_key_normalizer(
@@ -201,12 +202,9 @@ class PoolManager(RequestMethods):
         self,
         num_pools: int = 10,
         headers: typing.Mapping[str, str] | None = None,
-        resolver: Resolver | None = None,
         **connection_pool_kw: typing.Any,
     ) -> None:
         super().__init__(headers)
-        self.resolver = resolver
-
         # PoolManager handles redirects itself in PoolManager.urlopen().
         # It always passes redirect=False to the underlying connection pool to
         # suppress per-pool redirect handling. If the user supplied a non-Retry
@@ -280,7 +278,7 @@ class PoolManager(RequestMethods):
             for kw in SSL_KEYWORDS:
                 request_context.pop(kw, None)
 
-        return pool_cls(host, port, resolver=self.resolver, **request_context)
+        return pool_cls(host, port, **request_context)
 
     def clear(self) -> None:
         """
--- a/src/urllib3/util/__init__.py
+++ b/src/urllib3/util/__init__.py
@@ -1,8 +1,9 @@
 # For backwards compatibility, provide imports that used to be here.
 from __future__ import annotations
 
-from .connection import Resolver, is_connection_dropped
+from .connection import is_connection_dropped
 from .request import SKIP_HEADER, SKIPPABLE_HEADERS, make_headers
+from .resolver import Resolver
 from .response import is_fp_closed
 from .retry import Retry
 from .ssl_ import (
--- a/src/urllib3/util/__init__.py
+++ b/src/urllib3/util/__init__.py
@@ -1,7 +1,7 @@
 # For backwards compatibility, provide imports that used to be here.
 from __future__ import annotations
 
-from .connection import is_connection_dropped
+from .connection import Resolver, is_connection_dropped
 from .request import SKIP_HEADER, SKIPPABLE_HEADERS, make_headers
 from .response import is_fp_closed
 from .retry import Retry
@@ -39,4 +39,5 @@ __all__ = (
     "wait_for_write",
     "SKIP_HEADER",
     "SKIPPABLE_HEADERS",
+    "Resolver",
 )
--- a/src/urllib3/util/__init__.py
+++ b/src/urllib3/util/__init__.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 
 from .connection import is_connection_dropped
 from .request import SKIP_HEADER, SKIPPABLE_HEADERS, make_headers
-from .resolver import Resolver
 from .response import is_fp_closed
 from .retry import Retry
 from .ssl_ import (
@@ -40,5 +39,4 @@ __all__ = (
     "wait_for_write",
     "SKIP_HEADER",
     "SKIPPABLE_HEADERS",
-    "Resolver",
 )
--- a/src/urllib3/util/connection.py
+++ b/src/urllib3/util/connection.py
@@ -2,44 +2,15 @@ from __future__ import annotations
 
 import socket
 import typing
-from typing import Protocol
 
 from ..exceptions import LocationParseError
+from .resolver import Resolver
 from .timeout import _DEFAULT_TIMEOUT, _TYPE_TIMEOUT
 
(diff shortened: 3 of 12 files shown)


src/urllib3/poolmanager.py
```diff
@@ -90,6 +90,7 @@ class PoolKey(typing.NamedTuple):
     key_assert_fingerprint: str | None
     key_server_hostname: str | None
     key_blocksize: int | None
+    key_resolver: Resolver | None
```
sigmavirus24, line 96: Why would the resolver need to be part of the pool key? I see nothing that allows for selection of different resolvers based on any attributes of the request for a pool manager so it shouldn't be necessary to be part of the key

src/urllib3/util/connection.py
```diff
@@ -12,6 +13,35 @@
     from .._base_connection import BaseHTTPConnection
 
 
+class Resolver(Protocol):
```
sigmavirus24: Why not a new `resolver.py` module to keep it nicely contained?

src/urllib3/util/__init__.py
```diff
@@ -39,4 +39,5 @@
     "wait_for_write",
     "SKIP_HEADER",
     "SKIPPABLE_HEADERS",
+    "Resolver",
```
sigmavirus24: You only ever seem to import this from the connection submodule so why export this here. There seems to be no value to exporting it here

src/urllib3/util/connection.py
```diff
@@ -2,6 +2,7 @@
 
 import socket
 import typing
```
sigmavirus24, line 4: Typing is already imported here. Why make this less efficient with `from typing import Protocol `? Did your LLM fail to realize that it didn't need to make things worse?

src/urllib3/util/connection.py
sigmavirus24:
```suggestion
class Resolver(typing Protocol):
```

src/urllib3/util/connection.py
NyanKiyoshi: Nice catch, thanks! The code is proudly human made, likely force of habit 😉 I pushed 70e3045b36e80a269c08c07388dac98fe147bf49 which moves it to `resolver.py`, and removes the `from typing import Protocol` as it looks cleaner too

src/urllib3/util/__init__.py
NyanKiyoshi: This is used for the documentation as there is already [a page for the `util` package](https://github.com/urllib3/urllib3/blob/c5595202baaf77259b6404d455442d68ea33dbaa/docs/reference/urllib3.util.rst?plain=1#L15-L17) - thus allowing to explain what is `Resolver`.

| Before (without `Resolver` in `util/__init__.py`) | After |
| ------ | ----- |
|  Image | Image |

Alternatively, I could add the documentation here: https://github.com/urllib3/urllib3/blob/c5595202baaf77259b6404d455442d68ea33dbaa/src/urllib3/connection.py?1#L82-L105, or even do both: have `Resolver` type's reference still working in Sphinx, and usage information in `HTTPConnection`.

What do you suggest in this instance? Keep it `__init__.py` or you think another way would be better? (if so, what do you suggest?)

src/urllib3/poolmanager.py
NyanKiyoshi: For context, the issue I was trying to solve was this crash:

<details>

```
>>> from urllib3 import PoolManager
>>> manager = PoolManager(resolver=None)
>>> manager.connection_from_url("http://httpbin.org/")
  File "<python-input-2>", line 1, in <module>
    manager.connection_from_url("http://httpbin.org/")
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/urllib3/src/urllib3/poolmanager.py", line 385, in connection_from_url
    return self.connection_from_host(
           ~~~~~~~~~~~~~~~~~~~~~~~~~^
        u.host, port=u.port, scheme=u.scheme, pool_kwargs=pool_kwargs
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/app/urllib3/src/urllib3/poolmanager.py", line 318, in connection_from_host
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/app/urllib3/src/urllib3/poolmanager.py", line 341, in connection_from_context
  File "/app/urllib3/src/urllib3/poolmanager.py", line 148, in _default_key_normalizer
    return key_class(**context)
TypeError: PoolKey.__new__() got an unexpected keyword argument 'key_resolver'
```

</details>

I pushed 59f459945b6dac90a66df8bcec57cf1bcb2867f2 now and I removed the parameter from the `PoolKey` class. The code now looks correct (I'm not familiar with that class thus do let me know what you think)

src/urllib3/util/connection.py
NyanKiyoshi: Good idea, pushed at 70e3045b36e80a269c08c07388dac98fe147bf49

NyanKiyoshi:
@sigmavirus24 just letting you know this pull request is awaiting for your feedback, whenever you have some time to spare. Thanks for your time!

Note: I have been resyncing it periodically against `main`, and performed manual testing to ensure everything is still behaving correctly

src/urllib3/poolmanager.py
```diff
@@ -200,9 +201,12 @@ def __init__(
         self,
         num_pools: int = 10,
         headers: typing.Mapping[str, str] | None = None,
+        resolver: Resolver | None = None,
```
sigmavirus24: This should be covered by the connection pool keyword arguments. Remove it

src/urllib3/poolmanager.py
```diff
@@ -200,9 +201,12 @@ def __init__(
...
         **connection_pool_kw: typing.Any,
     ) -> None:
         super().__init__(headers)
+        self.resolver = resolver
```
sigmavirus24: Again, unnecessary

src/urllib3/poolmanager.py
```diff
@@ -276,7 +280,7 @@ def _new_pool(
             for kw in SSL_KEYWORDS:
                 request_context.pop(kw, None)
 
-        return pool_cls(host, port, **request_context)
+        return pool_cls(host, port, resolver=self.resolver, **request_context)
```
sigmavirus24: Revert this too

src/urllib3/util/__init__.py
sigmavirus24: You can explicitly autoclass the path to the name instead of putting this here. There's little value in this aside from not having to touch a doc file and if someone is doing an import * from this module and already have their own resolver class this may lead to static tooling providing new errors or warnings after they upgrade urllib3 and an influx of bug reports of low quality 

That said, I'm but one person on this project. I don't hold this strongly so if another maintainer feels more strongly, I'm happy to defer to them

src/urllib3/poolmanager.py
NyanKiyoshi: @sigmavirus24 I pushed f66be879d22a4a48e3c3588a8d28bc8b677ead21 for this. However, this requires putting `resolver` inside `PoolKey` as any values passed to `connection_pool_kw` (`PoolManager.__init__(**connection_pool_kw)`) or `connection_from_host(pool_kwargs)` and other `connection_from_[...]` functions are [flowing into the `pool_key_constructor()` function](https://github.com/urllib3/urllib3/blob/8f5c1b46387b4d86aa54c2c526e4930ae58200a7/src/urllib3/poolmanager.py#L339-L342) thus requiring the presence of `key_resolver` inside `PoolKey`

As well, this should be included inside the pool-key as if the value for `resolver` changes between connections then different values should not share the same pool. Given `resolver` affects `HTTPConnection._new_conn()`, two resolvers for the same `(scheme, host, port)` can connect to different addresses (when using different resolvers).

e.g.,:

```python

pm = PoolManager(resolver=my_resolver)
pm.connection_from_url("http://example.com", {"resolver": resolver_a})
pm.connection_from_url("http://example.com", {"resolver": resolver_b})
```

Thus this ensures we are always using the correct resolver.

Issue #3665, opened 2025-08-29
Python 3.13: `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier (_ssl.c:1032)`


benz0li:
### Subject

```python
import requests
requests.get('https://www.google.com')
```

```
urllib3.exceptions.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier (_ssl.c:1032)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
 File "/usr/local/lib/python3.13/site-packages/requests/adapters.py", line 644, in send
 resp = conn.urlopen(
 method=request.method,
 ...<9 lines>...
 chunked=chunked,
 )
 File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 841, in urlopen
 retries = retries.increment(
 method, url, error=new_e, _pool=self, _stacktrace=sys.exc_info()[2]
 )
 File "/usr/local/lib/python3.13/site-packages/urllib3/util/retry.py", line 519, in increment
 raise MaxRetryError(_pool, url, reason) from reason # type: ignore[arg-type]
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='www.google.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier (_ssl.c:1032)')))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
 File "<python-input-1>", line 1, in <module>
 requests.get('https://www.google.com')
 ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/usr/local/lib/python3.13/site-packages/requests/api.py", line 73, in get
 return request("get", url, params=params, **kwargs)
 File "/usr/local/lib/python3.13/site-packages/requests/api.py", line 59, in request
 return session.request(method=method, url=url, **kwargs)
 ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/usr/local/lib/python3.13/site-packages/requests/sessions.py", line 589, in request
[...]

Issue #3614, opened 2025-05-30
Existing kubernetes certificates compatibility broken in urllib with python 3.13


keramblock:
### Subject

Hi, I am not sure if that was intended, but it looks like recent changes in #3577 broke at least some of kubernetes automation in the wild after urllib was updated: https://github.com/kubernetes-client/python/issues/2394

I assume that number will grow as new version will be adopted by others.

### Environment
kubernetes api by major cloud providers(personally I use GKE cluster with cert, that was created about 4 years ago, also someone mentioned EKS)

### Steps to Reproduce
execute `app_deployments = api.list_deployment_for_all_namespaces(label_selector=label)` or any other kubernetes api request to such cluster

### Actual result
```urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='address', port=443): Max retries exceeded with url: /version/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Missing Authority Key Identifier (_ssl.c:1028)')))```

### Expected result
No error at all, or ability to turn on backward compatibility here somehow

Could you please advice me how to fix that, other then setting 'verify_ssl' to False, as was proposed in issue above?

Thanks a lot!

UPD: corresponding issue in GKE tracker: https://issuetracker.google.com/issues/421314618

sigmavirus24:
This is due to a recent change to SSLContext verification flags and OpenSSL behavior changes. You can look at VERIFY_X509 flags on the SSL documentation. 

In reality, the default Kubernetes PKI is bad and the root certificate for the infrastructure should be generated significantly better. Having AKI and SPKI are not remotely new requirements but because this CA solution wasn't really meant for production (but is used that way regardless) this is becoming a problem. It's also plausible these CAs are generated with a negative serial number which isn't expressly forbidden by 5240 but many TLS stacks are starting to reject those during verification

keramblock:
I tried to check CA there myself and it looks like it just isn't there, at least in my case.

sigmavirus24:
Correct. It wasn't created appropriately. OpenSSL now appropriately requires it depending on how you configure verification. As I said, you can customize your SSLContext without disabling verification entirely. Please go read the SSL standard library docs

pquentin:
[...]

PR #3782, opened 2026-03-01
Fix _ssl_wrap_socket_and_match_hostname overriding factory-created context's verify_mode     

dhoepp:
## Summary                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                              
  `_ssl_wrap_socket_and_match_hostname()` in `connection.py` creates an SSL context via                                                                                                                                                                       
  `create_urllib3_context()` (passing `cert_reqs`), then unconditionally overrides it:                                                                                                                                                                        
                                                                                                                                                                                                                                                              
  This is correct for caller-supplied `ssl_context` (the `else` branch). But it also                                                                                                                                                                          
  runs when `default_ssl_context = True`, silently undoing whatever the factory configured.                                                                                                                                                                   
  The factory already received `cert_reqs` and applied it.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                              
  ## Fix                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                              
  Guard with `if not default_ssl_context:` so it only applies to caller-supplied contexts.                                                                                                                                                                    
                                                                                                                                                                                                                                                              
  ## Impact                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                              
  Any patch to `create_urllib3_context` that sets `verify_mode` is silently overridden.                                                                                                                                                                       
  Affects enterprise users behind SSL-intercepting proxies (Zscaler, Cisco Umbrella, etc.)                                                                                                                                                                    
  whose corporate CA certs predate strict RFC 5280 Basic Constraints compliance — they                                                                                                                                                                        
  cannot add their CA to the trust chain and must disable Python-level verification,                                                                                                                                                                          
  which this bug prevents.                        



Related issues #3614 and #3665 describe users hitting similar SSL failures with enterprise and Kubernetes CA certs in Python 3.13. Customizing the SSLContext was suggested as a workaround in both threads, this fix ensures that approach works consistently, as currently any factory-level customization is silently overridden before the connection is made.
                                                                                                                                                                                                                                                              
  nuclearpidgeon's workaround in #3665 (passing a manually-constructed ssl_context directly) works precisely because caller-supplied contexts hit the else branch. This fix extends that capability to factory-created contexts, allowing application-level patches to create_urllib3_context to take effect without requiring every caller to construct and pass their own context.


 ## Historical context  
Tracing the git history suggests this regression was introduced in ea96fde ("Change proxy TLS flow to match destination TLS flow", Jan 2022). Prior to that commit, context was always self.ssl_context, a pre-existing connection-level context with no factory path, so applying cert_reqs unconditionally was correct. When _ssl_wrap_socket_and_match_hostname was extracted and the create_urllib3_context factory branch was added, the override line was transplanted from the old connect() method where context was always a pre-existing self.ssl_context — correct there, but not reconciled with the newly introduced factory branch in the same commit, causing it to also run on factory-created contexts.

  I considered moving the line into the else branch instead, but opted for the if not default_ssl_context: guard to make the intent explicit: the factory owns the configuration of its own context, and the override only applies when the caller supplies their own. Happy to adjust if the maintainers prefer a different approach.




<!---
Thanks for your contribution! ♥️

If this is your first PR to urllib3 please review the Contributing Guide:
https://urllib3.readthedocs.io/en/latest/contributing.html

Adhering to the Contributing Guide means we can review, merge, and release your change faster! :)
--->


illia-v:
Thanks for the contribution @dhoepp! #5093 has fixed this issue and some related ones.

PR #3786, opened 2026-03-03
Fix proxy_ssl_context ignored for HTTPS forwarding proxies (#2577)

### LINKED ISSUE #2577: Proxies should use proxy_ssl_context when connecting via ProxyConfig.use_forwarding_with_https = True

dhoepp:
Found this while debugging SSL issues behind a corporate proxy in #3782. In forwarding mode there's only one TLS handshake and it was always using the destination ssl_context instead of the proxy one. Tunneling already worked because it has a dedicated _connect_tls_proxy() call. 

Fix is a 4-line branch in connect() to pick the right context. Also added ValueError when ssl_context and use_forwarding_for_https=True are both set, per the issue spec.

Fixes #2577.

## Root cause

The tunneling path (`proxy_is_tunneling`) was already handled correctly by `_connect_tls_proxy()`, which explicitly uses `proxy_config.ssl_context`. The forwarding path, however, fell through to the shared `_ssl_wrap_socket_and_match_hostname` call that always passed `self.ssl_context`.

## History

The bug dates to 691679f7 (Sept 2020) when forwarding proxy support was first introduced — the forwarding path was added with `ssl_context=self.ssl_context` and never corrected.    

## Fix

```python
if self.proxy_is_forwarding and self.proxy_config is not None:
    tls_ssl_context = self.proxy_config.ssl_context
else:
    tls_ssl_context = self.ssl_context
```

## Tests

- Removed the `pytest.skip("Test is expected to fail due to urllib3/urllib3#2577")` guard in `test_proxy_https_target_tls_error` — that test now passes directly.
- Added `test_https_proxy_forwarding_uses_proxy_ssl_context` which verifies that an HTTPS forwarding proxy works correctly when only `proxy_ssl_context` is supplied (no `ca_certs` fallback).

### DIFF
--- a/changelog/2577.bugfix.rst
+++ b/changelog/2577.bugfix.rst
@@ -2,4 +2,6 @@ Fixed ``proxy_ssl_context`` being ignored for HTTPS forwarding proxies. When
 ``proxy_is_forwarding`` is ``True``, ``HTTPSConnection.connect()`` now correctly
 uses ``proxy_config.ssl_context`` (derived from ``proxy_ssl_context``) to
 establish the TLS connection to the proxy instead of ``ssl_context`` (which
-is intended for the destination origin).
+is intended for the destination origin). Passing ``ssl_context`` together with
+``use_forwarding_for_https=True`` now raises ``ValueError`` to make this
+separation explicit.
--- a/changelog/2577.bugfix.rst
+++ b/changelog/2577.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed ``proxy_ssl_context`` being ignored for HTTPS forwarding proxies. When
+``proxy_is_forwarding`` is ``True``, ``HTTPSConnection.connect()`` now correctly
+uses ``proxy_config.ssl_context`` (derived from ``proxy_ssl_context``) to
+establish the TLS connection to the proxy instead of ``ssl_context`` (which
+is intended for the destination origin).
--- a/changelog/2577.bugfix.rst
+++ b/changelog/2577.bugfix.rst
@@ -1,4 +1,2 @@
-``proxy_ssl_context`` now works correctly when using ``ProxyManager`` with
-``use_forwarding_for_https=True``. Passing ``ssl_context`` in this configuration
-now emits a ``FutureWarning`` and will raise an error in v3.0; use
-``proxy_ssl_context`` instead.
+Fixed usage of ``proxy_ssl_context`` with ``ProxyManager`` when ``use_forwarding_for_https=True``.
+Passing ``ssl_context`` instead of ``proxy_ssl_context`` for HTTPS proxies in this configuration now emits a ``FutureWarning`` and will raise an error in v3.0.
--- a/changelog/2577.bugfix.rst
+++ b/changelog/2577.bugfix.rst
@@ -1,7 +1,4 @@
-Fixed ``proxy_ssl_context`` being ignored for HTTPS forwarding proxies. When
-``proxy_is_forwarding`` is ``True``, ``HTTPSConnection.connect()`` now correctly
-uses ``proxy_config.ssl_context`` (derived from ``proxy_ssl_context``) to
-establish the TLS connection to the proxy instead of ``ssl_context`` (which
-is intended for the destination origin). Passing ``ssl_context`` together with
-``use_forwarding_for_https=True`` now raises ``ValueError`` to make this
-separation explicit.
+``proxy_ssl_context`` now works correctly when using ``ProxyManager`` with
+``use_forwarding_for_https=True``. Passing ``ssl_context`` in this configuration
+now emits a ``FutureWarning`` and will raise an error in v3.0; use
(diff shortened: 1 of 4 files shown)


src/urllib3/poolmanager.py
```diff
@@ -579,6 +579,15 @@ def __init__(
...
             raise ProxySchemeUnknown(proxy.scheme)
 
+        if (
+            use_forwarding_for_https
+            and connection_pool_kw.get("ssl_context") is not None
+        ):
+            raise ValueError(
+                "ssl_context is not applicable when use_forwarding_for_https=True. "
+                "Use proxy_ssl_context to configure the TLS connection to the proxy."
+            )
```
illia-v, line 595: The issue had been opened before urllib3 2.0 was released, so it made sense to raise a validation error. Nowadays urllib3 2.x is pretty mature, so the immediate error can be too breaking for users who have found `ssl_context` working instead of `proxy_ssl_context`.
Let's use `FutureWarning` mentioning a transition to an error scheduled for v3.0.

dhoepp: Done. swapped `ValueError` for a `FutureWarning` with  `stacklevel=2` and updated the test to use `pytest.warns`. let me know if the message wording needs adjusting. 

All checks passed, also updated the branch against main.

illia-v: Thanks for replacing the error with a warning! However, the change is still breaking because `connect`  does not use `self.ssl_context` when `self.proxy_config.ssl_context` is missing.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -705,23 +724,31 @@ def test_https_proxy_tls_error(
...
+        if use_forwarding_for_https:
+            # ssl_context is not valid with use_forwarding_for_https=True;
+            # proxy_ssl_context must be used instead.
+            with pytest.warns(FutureWarning, match="ssl_context"):
+                proxy_from_url(
+                    proxy_url,
+                    proxy_ssl_context=proxy_ctx,
+                    ssl_context=ctx,
+                    use_forwarding_for_https=use_forwarding_for_https,
+                )
```
illia-v: The proxy is still expected to be usable after the warning, make a check similar to the `use_forwarding_for_https=False` case

changelog/2577.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+Fixed ``proxy_ssl_context`` being ignored for HTTPS forwarding proxies. When
+``proxy_is_forwarding`` is ``True``, ``HTTPSConnection.connect()`` now correctly
+uses ``proxy_config.ssl_context`` (derived from ``proxy_ssl_context``) to
+establish the TLS connection to the proxy instead of ``ssl_context`` (which
+is intended for the destination origin). Passing ``ssl_context`` together with
+``use_forwarding_for_https=True`` now raises ``ValueError`` to make this
+separation explicit.
```
illia-v: Please remove mentions of `ValueError` and internal details, the entry should focus on `proxy_ssl_context` working in the high-level API now

illia-v (changes requested):

test/with_dummyserver/test_proxy_poolmanager.py
dhoepp: Done. Swapped the `ValueError` for a `FutureWarning` with a fallback so the proxy stays usable, and added a test to verify that.

changelog/2577.bugfix.rst
dhoepp: also updated this to be more user facing

test/with_dummyserver/test_proxy_poolmanager.py
dhoepp: All checks green. Had to split `test_proxy_https_target_tls_error` into separate cases. in forwarding mode the proxy handles target TLS so the request succeeds rather than raising, which needed a different assertion.

illia-v:
I've pushed extended tests during the review, I'll revisit the PR tomorrow

illia-v (approved):

illia-v:
@dhoepp thank you for the fix!

There is a $500 bounty for fixing the issue, feel free to submit an expense to [our Open Collective](https://opencollective.com/urllib3) with a link to this pull request.

shazow:
@dhoepp Can you please confirm that it was you who just opened the expense request? Want to make sure that it's not someone else trying to impersonate you. :)

dhoepp:
@shazow yes! Sorry I thought I connected the two a while back

PR #3794, opened 2026-03-16
Fix infinite loop in HTTPResponse.stream when amt=0

### LINKED ISSUE #3793: `HTTPResponse.stream` can fall into infinite loop when `amt` is zero
### Subject

### Steps to Reproduce

```python
import urllib3

for chunk in r.stream(0):
    print(chunk)
```

### Actual Behavior

Run into infinite loop

### Expected Behavior

yama-bun:
### Summary
Fixes an infinite loop in `HTTPResponse.stream()` when `amt=0`. 
As discussed in #3793, `read(0)` returns an empty byte string which previously caused the stream generator to loop indefinitely. 

### Changes
- Added a guard clause to `HTTPResponse.stream` to raise `ValueError` if `amt == 0`.
- Added a changelog entry.

Fixes #3793


yama-bun:
I've updated the comment in the test file from Japanese to English for clarity. Sorry for the oversight!

PR #3796, opened 2026-03-17
Fix late-bound default_socket_options in HTTPConnection

### LINKED ISSUE #3789: HTTPConnection does not respect HTTPConnection.default_socket_options
### Subject

Changing` HTTPConnection.default_socket_options` does nothing.
Creating a new `HTTPConnection` will always use the default value from the class definition and will ignore any changes done to the `HTTPConnection.default_socket_options`  value.

### Environment

Tested on multiple platforms, the bug was introduced in version `2.0.0` and is present in all versions since then
Tested on Ubuntu (bare metal) and alpine linux (docker).
The bug is in urllib3 so should appear on any platform.

### Steps to Reproduce

A very detailed explanation can be seen bellow a bit bellow

### Expected Behavior

When changing `HTTPConnection.default_socket_options`, it is expected that future any connection created after that will use the new defaults.

### Actual Behavior

The new connections ignore `HTTPConnection.default_socket_options`

---

on lines https://github.com/urllib3/urllib3/blob/main/src/urllib3/connection.py#L140-L142 in the `__init__` of the HTTPConnection class, `socket_options` has default value `default_socket_options`.
This value is evaluated when the class itself is created, NOT when an instance of the class is created.
This means that once the class definition is loaded, this value can't be changed.
Changing `HTTPConnection.default_socket_options` does not change the already evaluated default value for `socket_options`

Here is a simple code snippet that demonstrates the issue , show what exactly is broken with a simplified `Test` class and also shows a hacky workaround that still allows to change the defaults.
```
from urllib3.connection import HTTPConnection, HTTPSConnection
import socket

class Test:
    """A simple test class to demonstrate the issue with default arguments being evaluated at function definition time."""

    default_test_conf:str= "default_value"
    test_conf: str | None
    def __init__(self, *, test_conf: None | str = default_test_conf):
        print (f"Initializing Test with test_conf: {test_conf}")
        print (f"Actual default test conf: {Test.default_test_conf}")
        self.test_conf = test_conf

print(f"Demonstrating the issue with HTTPConnection default_socket_options and a workaround by modifying the __kwdefaults__ of HTTPConnection.__init__ and HTTPSConnection.__init__\n")

print ("At this point the default for socket_options is already set")
print(f"HTTPConnection.__init__.__kwdefaults__['socket_options']: {HTTPConnection.__init__.__kwdefaults__['socket_options']}")

# we will now try to modify the defaults
    (socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 1),
    (socket.IPPROTO_TCP, socket.TCP_KEEPINTVL, 5),
    (socket.IPPROTO_TCP, socket.TCP_KEEPCNT, 5),
]

# passing socket_options explicitly should override the default_socket_options, works as expected
con: HTTPConnection = HTTPConnection('example.com', 80, socket_options=HTTPConnection.default_socket_options)
print(f"con.default_socket_options: {HTTPConnection.default_socket_options}")
print(f"con.socket_options: {con.socket_options}")

# when not passing socket_options should use the default_socket_options, does not work as expected
# con2.socket_options has the default socket options, not the modified default_socket_options with keepalive options, which is unexpected
con2: HTTPConnection = HTTPConnection('example.com', 80)
print(f"con2.default_socket_options: {HTTPConnection.default_socket_options}")
print(f"con2.socket_options: {con2.socket_options}")

# the problem is that HTTPConnection.__init__ has socket_options as a default argument, which is evaluated at function definition time, so it does not see the modified default_socket_options when it is defined
# here you can see that the default value for socket_options in HTTPConnection is the original default_socket_options, not the modified one with keepalive options
# the same is true for HTTPSConnection
print(f"HTTPConnection.__init__.__kwdefaults__['socket_options']: {HTTPConnection.__init__.__kwdefaults__['socket_options']}")
print(f"HTTPSConnection.__init__.__kwdefaults__['socket_options']: {HTTPSConnection.__init__.__kwdefaults__['socket_options']}")

# we can show that modifying the "HTTPConnection.__init__.__kwdefaults__['socket_options']" will actually change the default socket options used by HTTPConnection when socket_options is not passed explicitly, which is the workaround for this issue
HTTPConnection.__init__.__kwdefaults__['socket_options'] = HTTPConnection.default_socket_options
HTTPSConnection.__init__.__kwdefaults__['socket_options'] = HTTPSConnection.default_socket_options
con3: HTTPConnection = HTTPConnection('example.com', 80)
print(f"con3.default_socket_options: {HTTPConnection.default_socket_options}")
print(f"con3.socket_options: {con3.socket_options}")

# here is just a very simplified test to show the behavior explained above with a custom class and a default argument that is evaluated at function definition time, which is similar to the socket_options case in HTTPConnection

print("\n\nTesting custom class with default argument evaluated at function definition time:\n")
a = Test()
print(f"a.default_test_conf: {Test.default_test_conf}")
print(f"a.test_conf: {a.test_conf}")

Test.default_test_conf = "new_default_value"
b = Test()
print(f"b.default_test_conf: {Test.default_test_conf}")
print(f"b.test_conf: {b.test_conf}")

print(f"Test.__init__.__kwdefaults__['test_conf']: {Test.__init__.__kwdefaults__['test_conf']}")
Test.__init__.__kwdefaults__["test_conf"] = "another_default_value"
c = Test()
print(f"c.default_test_conf: {Test.default_test_conf}")
print(f"c.test_conf: {c.test_conf}")
```

yama-bun:
### Summary
- Fixes a bug where `HTTPConnection.default_socket_options` changes were ignored because default arguments were bound at function definition time.
- Changed the default value of `socket_options` to `None` in both `HTTPConnection` and `HTTPSConnection`, evaluating it at runtime instead.

### Motivation
In Python, default arguments are evaluated only once at definition time. Since `socket_options` was defaulting to the class-level `default_socket_options` list directly, any subsequent changes to the class variable were not picked up by new connection instances. This fix ensures dynamic evaluation.

### Test plan
- Added `test_HTTPConnection_respects_default_socket_options_change` to verify that modifying the class variable affects new instances.
- Added a similar test for `HTTPSConnection`.
- Verified with `nox -rs test-3.13`.

Closes #3789


yama-bun:
Thank you for the guidance. I attempted several approaches to ensure strict type safety without using Any, but I couldn't resolve the persistent mypy errors regarding the sentinel type and class inheritance.

To avoid leaving a broken PR, I'm closing this for now. I've learned a lot about Python's type system through this challenge. Thanks again!

PR #3797, opened 2026-03-19
fix: prevent infinite loop in HTTPResponse.stream when amt=0

### LINKED ISSUE #3793: `HTTPResponse.stream` can fall into infinite loop when `amt` is zero

nightcityblade:
Fixes #3793

When `amt=0` is passed to `HTTPResponse.stream()`, `read(amt=0)` returns `b""`, which is falsy and never yielded, but the file pointer remains open. This causes the while loop to spin indefinitely.

This adds an early return when `amt` is not None and `<= 0`, consistent with how `read(0)` returns immediately with `b""`.

A test is included to verify the fix.

### DIFF
--- a/changelog/3793.bugfix.rst
+++ b/changelog/3793.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed ``HTTPResponse.stream()`` to handle ``amt=0`` (returns empty bytes immediately) and negative ``amt`` (treats as ``None``, reading all data), matching ``read()`` behavior.
+Fixed ``HTTPResponse.stream()`` to handle ``amt=0``.


themavik (commented):
Early return when `amt <= 0` stops `HTTPResponse.stream(0)` from looping forever; `test_stream_zero_amt` covers it. nit: negative `amt` now yields nothing—if that is reachable from public callers, mirror whatever `read()` does for negatives.

nightcityblade:
Good catch! Updated to match `read()` behavior: negative `amt` is now treated as `None` (read all), while `amt=0` returns immediately. Thanks for the review.

src/urllib3/response.py
```diff
@@ -1238,6 +1238,13 @@ def stream(
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
+        if amt is not None and amt < 0:
+            # Negative numbers should be treated the same as None (read all),
+            # matching the behavior of read().
+            amt = None
```
illia-v: I don't think adding these lines is necessary because negative numbers have been already treated correctly in `read_chunked` and `read`

src/urllib3/response.py
```diff
@@ -1238,6 +1238,13 @@ def stream(
...
+        elif amt == 0:
+            return
```
illia-v: https://github.com/urllib3/urllib3/pull/198#issuecomment-19162605 suggests that there was a decision to never return an empty string, but I guess this doesn't count because the `return` just interrups a generator, it doesn't return an empty string

illia-v (commented):
We'll need a [proper](https://github.com/urllib3/urllib3/tree/main/changelog#readme) changelog entry if we choose this as a fix.

nightcityblade:
Added the changelog entry in `changelog/3793.bugfix.rst` as requested. Thanks for the pointer!

illia-v:
@nightcityblade please check other comments

nightcityblade:
Thanks for the review! You're right — negative numbers are already handled correctly in `read_chunked` and `read`, so the extra guard was unnecessary. Removed it in the latest push. Only keeping the `amt=0` early return to prevent the infinite loop.

changelog/3793.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ``HTTPResponse.stream()`` to handle ``amt=0`` (returns empty bytes immediately) and negative ``amt`` (treats as ``None``, reading all data), matching ``read()`` behavior.
```
illia-v: Need to change this:
* returns empty bytes immediately - incorrect, it doesn't return empty bytes
* negative ``amt`` - no changes related to this


```suggestion
Fixed ``HTTPResponse.stream()`` to handle ``amt=0``.
```

Issue #3798, opened 2026-03-19
_TYPE_BODY should include Iterable[str]


nateprewitt:
### Context

This came up while working on psf/requests#7272.

The `_TYPE_BODY` type alias currently allows `str` and `bytes` independently, and `Iterable[bytes]` for streaming bodies:

https://github.com/urllib3/urllib3/blob/0248277dd7ac0239204889ca991353ad3e3a1ddc/src/urllib3/_base_connection.py#L9

But `Iterable[str]` is handled correctly at runtime through two separate code paths and imo should be included in the type. This code path has been used by Requests for essentially its entire existence.

`body_to_chunks()` passes any iterable through as chunks:

https://github.com/urllib3/urllib3/blob/0248277dd7ac0239204889ca991353ad3e3a1ddc/src/urllib3/util/request.py#L248-L252

Then if it's a `str`, connection.py encodes each chunk to UTF-8 before sending:

https://github.com/urllib3/urllib3/blob/0248277dd7ac0239204889ca991353ad3e3a1ddc/src/urllib3/connection.py#L504-L510

What confirmed this was likely an oversight rather than intentional is urllib3 already documents this behavior. Both [HTTPConnectionPool.urlopen()](https://github.com/urllib3/urllib3/blob/0248277dd7ac0239204889ca991353ad3e3a1ddc/src/urllib3/connectionpool.py#L407) and [_make_request()](https://github.com/urllib3/urllib3/blob/0248277dd7ac0239204889ca991353ad3e3a1ddc/src/urllib3/connectionpool.py#L635) describe body as:

> an iterable of :class:`str`/:class:`bytes`, or a file-like object.

That documentation was added in 2020, I believe it's worked for the majority of urllib3's existence too. We even have tests exercising this behavior that are currently suppressed for type mismatches.

### Reproduction

```python
import urllib3

def string_body():
    yield "hello "
    yield "world"
   
pool = urllib3.HTTPConnectionPool("httpbin.org", 80)
pool.urlopen("POST", "/post", body=string_body())
```

### Suggested fix

```python
_TYPE_BODY = Union[bytes, IO[Any], Iterable[bytes | str], str]
```

Or `Iterable[bytes] | Iterable[str]` if you prefer to keep them separate.

### Contribution

[...]

PR #3799, opened 2026-03-19
Add Iterable[str] to input body type

nateprewitt:
This PR addresses #3798 by adding `Iterable[str]` to the accepted input types for bodies. This was the behavior documented prior to types being added (and is still the documented behavior), so this change is largely cosmetic.


sigmavirus24 (approved):
:+1: From me, but I'm not about to merge this without @illia-v at least taking a peek

PR #3801, opened 2026-03-25
Fix SOCKS proxy auth failure with percent-encoded userinfo

### LINKED ISSUE #3785: SOCKS proxy URIs with percent-encoded userinfo do not work
### Subject

A SOCKS proxy URI userinfo subcomponent containing percent encoded values (e.g. `socks5h://user:%70ass@host:1080`) will not have the username and password be percent decoded or unquoted before they're passed to the proxy server for authentication.

### Environment

```
OS: Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.36
Python: 3.13.0
urllib3: 2.6.3
```

### Steps to Reproduce

I ran into this problem with a password containing a delimiter character that needed to be percent encoded to have the proxy URI correctly parsed.  Despite fixing the encoding, authentication failed.  As a synthetic test, one can percent encode any username or password character from a case that works and it will fail.

```python
import urllib3
p = 'socks5h://user:pass@host:1080'
spm = urllib3.contrib.socks.SOCKSProxyManager(proxy_url=p)
spm.request('GET', 'https://httpbin.org/ip').data
# suppose this works, now replace a character in the username or password with the percent encoded equivalent:
p = 'socks5h://user:%70ass@host:1080'
spm = urllib3.contrib.socks.SOCKSProxyManager(proxy_url=p)
spm.request('GET', 'https://httpbin.org/ip').data
# this should fail with: socks.SOCKS5AuthError: SOCKS5 authentication failed
```

### Expected Behavior

Pass percent decoded or unquoted username and password values to the proxy for authentication.

### Actual Behavior

Percent encoded usernames and passwords are presented to the proxy causing authentication failures.

### Additional Notes

While exploring this issue I convinced myself that it's reasonable for the parse_uri function and the Uri object to not hold percent decoded values, because they only break the URI down as far as subcomponents.  I would think differently if, like urllib's urlsplit function, there were accessors for username and password directly.

I believe the problem lies in the [SOCKSProxyManager](https://github.com/urllib3/urllib3/blob/2.6.3/src/urllib3/contrib/socks.py#L196-L198):

```python
        parsed = parse_url(proxy_url)

        if username is None and password is None and parsed.auth is not None:
            split = parsed.auth.split(":")
            if len(split) == 2:
                username, password = split
```

After the username and password no longer have the semantics of the URI userinfo subcomponent, they should be percent decoded or unquoted for use in the rest of the application.  The ready example is "how does cURL handle this?" which is [here](https://github.com/curl/curl/blob/curl-8_18_0/lib/setopt.c#L2006-L2028):

```c
    /* URL decode the components */
    if(!result && u) {
      Curl_safefree(s->str[STRING_PROXYUSERNAME]);
      result = Curl_urldecode(u, 0, &s->str[STRING_PROXYUSERNAME], NULL,
                              REJECT_ZERO);
    }
    if(!result && p) {
      Curl_safefree(s->str[STRING_PROXYPASSWORD]);
      result = Curl_urldecode(p, 0, &s->str[STRING_PROXYPASSWORD], NULL,
                              REJECT_ZERO);
    }
```

I also tested using requests v2.32.5 which works, and uses urllib3, [but manually configures the SOCKSProxyManager username and password](https://github.com/psf/requests/blob/v2.32.5/src/requests/adapters.py#L256-L266) instead of relying on parsing the proxy_url:

```python
        elif proxy.lower().startswith("socks"):
            username, password = get_auth_from_url(proxy)
            manager = self.proxy_manager[proxy] = SOCKSProxyManager(
                proxy,
                username=username,
                password=password,
                num_pools=self._pool_connections,
                maxsize=self._pool_maxsize,
                block=self._pool_block,
                **proxy_kwargs,
            )
```

The [get_auth_from_url](https://github.com/psf/requests/blob/v2.32.5/src/requests/utils.py#L1017) utility performs the percent decode here.

Krishnachaitanyakc:
## Summary

Fixes #3785.

SOCKS proxy URIs with percent-encoded userinfo (e.g. `socks5h://user:%70ass@host:1080`) fail authentication because the username and password extracted from `parsed.auth` are not percent-decoded before being sent to the proxy server.

- Added `urllib.parse.unquote()` calls to percent-decode the username and password after extracting them from the proxy URL's auth component in `SOCKSProxyManager.__init__`.
- Added an integration test `test_socks_with_percent_encoded_auth_in_url` that verifies SOCKS5 authentication succeeds when credentials are percent-encoded in the proxy URI.
- Added a towncrier changelog fragment.

This is consistent with how cURL handles proxy userinfo (it calls `Curl_urldecode` on the extracted username and password) and how the `requests` library works around this by manually calling `unquote` via `get_auth_from_url`.

## Test plan

- [x] New test `test_socks_with_percent_encoded_auth_in_url` passes (uses `socks5://us%65r:%70ass@host:port` and verifies authentication succeeds with decoded `user:pass`).
- [x] All existing SOCKS proxy tests continue to pass (20/20).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,7 +117,7 @@ jobs:
           check-latest: true
 
       - name: "Install uv"
-        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
         with:
           version: "0.9.5"
 
@@ -186,7 +186,7 @@ jobs:
           python-version: "3.x"
 
       - name: "Install uv"
-        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
         with:
           version: "0.7.12"
 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -28,13 +28,13 @@ jobs:
         persist-credentials: false
 
     - name: "Run CodeQL init"
-      uses: github/codeql-action/init@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4.31.0
+      uses: github/codeql-action/init@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
       with:
         config-file: "./.github/codeql.yml"
         languages: "python,javascript-typescript,actions"
 
     - name: "Run CodeQL autobuild"
-      uses: github/codeql-action/autobuild@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4.31.0
+      uses: github/codeql-action/autobuild@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
 
     - name: "Run CodeQL analyze"
-      uses: github/codeql-action/analyze@4e94bd11f71e507f7f87df81788dff88d1dacbfb # v4.31.0
+      uses: github/codeql-action/analyze@c10b8064de6f491fea524254123dbe5e09572f13 # v4.35.1
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -26,7 +26,7 @@ jobs:
           python-version: "3.x"
 
       - name: "Install uv"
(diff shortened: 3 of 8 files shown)


Krishnachaitanyakc:
@pquentin @illia-v can you please review these changes?

illia-v:
Let's see if #4948 can be a more universal solution

Krishnachaitanyakc:
> Let's see if #3801 can be a more universal solution

@illia-v I'm sorry, what do you mean?

illia-v:
> > Let's see if #3801 can be a more universal solution
> 
> @illia-v I'm sorry, what do you mean?

Sorry, I meant #4948, it makes a similar change to `parse_url`

PR #4948, opened 2026-03-30
fix(parse_url): decode percent-encoded auth credentials

### LINKED ISSUE #4945: parse_url doesn't %-decode user/password
### Subject

As far as I understand[^rfc], reserved characters need to %-encoded in URLs.

[^rfc]: [RFC 3986 section 3.2.1.](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1): `userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )`

So, if you have user "u{er" with password "pa}}word" for example.com, you would encode it as `http://u%7Ber:pass%7D%7Dword@example.com`. However, parse_url doesn't decode those parts:

```py
>>> import urllib3
>>> urllib3.util.parse_url("http://u%7Ber:pass%7D%7Dword@example.com").auth
'u%7Ber:pass%7D%7Dword'
```

Even worse, it encodes them for invalid URLs if you try to work around it:

```py
>>> urllib3.util.parse_url("http://u{er:pass}}word@example.com").auth
'u%7Ber:pass%7D%7Dword'
```

Therefore, if you use it to parse an URL pass `url.auth` into `make_headers(basic_auth=url.auth)`, you get a 401 and a downtime (ask me how I know).

### Environment

```
OS macOS-26.4-x86_64-i386-64bit-Mach-O
Python 3.14.3
OpenSSL 3.5.5 27 Jan 2026
urllib3 2.6.3
```

### Steps to Reproduce

See above.

### Expected Behavior

`urllib3.util.parse_url("http://u%7Ber:pass%7D%7Dword@example.com").auth` should return `u{er:pa}}word`.

`yarl` and `httpx.URL` get this right.

### Actual Behavior

It returns `u%7Ber:pass%7D%7Dword`

This behavior is shared by the stdlib and ada_url.

guoyangzhen:
Fixes #4945

## Problem

`parse_url()` does not decode percent-encoded characters in the auth (userinfo) component of URLs. When a URL contains `http://u%7Ber:pass%7D%7Dword@example.com`, the `auth` property returns `u%7Ber:pass%7D%7Dword` instead of the decoded `u{er:pass}}word`.

This causes authentication failures when the parsed auth is passed to `make_headers(basic_auth=url.auth)`, as the server expects the actual credentials, not their percent-encoded form.

The behavior is inconsistent with other popular HTTP libraries like `yarl` and `httpx` which correctly decode auth components.

## Fix

- Added `from urllib.parse import unquote` import
- Changed the auth processing to decode percent-encoded characters using `unquote()` instead of re-encoding invalid characters with `_encode_invalid_chars()`

## Tests

- Added `test_parse_url_decodes_percent_encoded_auth` with 5 test cases covering:
  - Normal auth (no encoding needed)
  - Curly braces in auth (`%7B`, `%7D`)
  - `@` and `#` characters in auth (`%40`, `%23`)
  - Simple auth string
  - No auth present

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import re
 import typing
+from urllib.parse import unquote
 
 from ..exceptions import LocationParseError
 from .util import to_str
@@ -421,8 +422,8 @@ def parse_url(url: str) -> Url:
             auth, _, host_port = authority.rpartition("@")
             auth = auth or None
             host, port = _HOST_PORT_RE.match(host_port).groups()  # type: ignore[union-attr]
-            if auth and normalize_uri:
-                auth = _encode_invalid_chars(auth, _USERINFO_CHARS)
+            if auth:
+                auth = unquote(auth)
             if port == "":
                 port = None
         else:
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -306,6 +306,25 @@ class TestUtil:
     ]
 
     @pytest.mark.parametrize(
+        "url, expected_auth",
+        [
+            ("http://user:pass@example.com", "user:pass"),
+            ("http://u%7Ber:pass%7D%7Dword@example.com", "u{er:pass}}word"),
+            ("http://user%40name:pass%23word@example.com", "user@name:pass#word"),
+            ("http://simple@example.com", "simple"),
+            ("http://example.com", None),
+        ],
+    )
+    def test_parse_url_decodes_percent_encoded_auth(
+        self, url: str, expected_auth: str | None
+    ) -> None:
+        """Assert parse_url decodes percent-encoded auth credentials.
+
+        Regression test for #4945.
+        """
+        assert parse_url(url).auth == expected_auth
+
+    @pytest.mark.parametrize(
         "url, expected_url",
         chain(parse_url_host_map, non_round_tripping_parse_url_host_map),
     )


src/urllib3/util/url.py
```diff
@@ -421,8 +422,8 @@ def parse_url(url: str) -> Url:
             auth, _, host_port = authority.rpartition("@")
             auth = auth or None
             host, port = _HOST_PORT_RE.match(host_port).groups()  # type: ignore[union-attr]
-            if auth and normalize_uri:
-                auth = _encode_invalid_chars(auth, _USERINFO_CHARS)
+            if auth:
+                auth = unquote(auth)
```
illia-v, line 426: This breaks `test_control_characters_are_percent_encoded` added in #1647 as you can see in the CI output

A change like the following can keep the current tests passing
```suggestion
                auth = unquote(auth)
                if normalize_uri:
                    auth = _encode_invalid_chars(auth, _USERINFO_CHARS)
```

But the following two new cases will fail because the encoded characters are not present in `_USERINFO_CHARS` from https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1
```
FAILED test/test_util.py::TestUtil::test_parse_url_decodes_percent_encoded_auth[http://u%7Ber:pass%7D%7Dword@example.com-u{er:pass}}word] - AssertionError: assert 'u%7Ber:pass%7D%7Dword' == 'u{er:pass}}word'
FAILED test/test_util.py::TestUtil::test_parse_url_decodes_percent_encoded_auth[http://user%40name:pass%23word@example.com-user@name:pass#word] - AssertionError: assert 'user%40name:pass%23word' == 'user@name:pass#word'
```

opello:
I would pose the following questions:

Should the Url.auth property, which appears to represent the RFC 3986 userinfo subcomponent, be presented with percent encoded values decoded?  This seems to obviously present a problem with username or password values that contain colons, preventing the user from accurately identifying either the username or password.

Should username and password properties be added to Url, that would be unquoted, to allow easier consumption of the underlying information with less ambiguity?

illia-v:
We added properties providing decoded auth in #5055.
`Url.auth` is not intended to provide the decoded auth.

Thanks for the contribution!

PR #4953, opened 2026-04-01
Handle null body response in Emscripten

### LINKED ISSUE #3723: `JsNull` not handled in emscripten response body handler.
### Subject

Under Pyodide, if `streaming=True` in `send_jspi_request` then the line `body_stream_js = response_js.body.getReader()` [1] may fail with `AttributeError: 'JsNull' object has no attribute 'getReader'`. This is because `response_js.body` may be `JsNull`, if the fetch response is `null` [2]. I presume the fix should be to modify the guard [3], to check against both `None` and `JsNull`.

[1] https://github.com/urllib3/urllib3/blob/6e0e96c76fedec21a7189342f59cd39a1d8e7086/src/urllib3/contrib/emscripten/fetch.py#L610
[2] https://pyodide.org/en/stable/usage/type-conversions.html#javascript-to-python
[3] https://github.com/urllib3/urllib3/blob/6e0e96c76fedec21a7189342f59cd39a1d8e7086/src/urllib3/contrib/emscripten/fetch.py#L608

### Environment

Running under Pyodide in Deno, with SSL not loaded
```
OS Emscripten-4.0.9-wasm32-32bit
Python 3.13.2
urllib3 2.5.0
```

### Steps to Reproduce

This script causes the error on the `requests.post` line. The script is not runnable as the endpoint is private, requiring an `access_token`, `device_id` and `script_policy_id`.

```python
import requests

access_token = ...
device_id = ...
script_policy_id = ...

headers = {
    "Authorization": f"Bearer {access_token}",
    "Content-Type": "application/json"
}

d_url = f"https://graph.microsoft.com/beta/deviceManagement/managedDevices('{device_id}')/initiateOnDemandProactiveRemediation"
body = {
    "scriptPolicyId": script_policy_id
}
resp = requests.post(d_url, headers=headers, json=body, timeout=120)
```

### Expected Behavior

The script should not throw an exception when processing the response body.

### Actual Behavior

The script fails with
```
File "/lib/python313.zip/_pyodide/_base.py", line 597, in eval_code_async
await CodeRunner(
...<9 lines>...
.run_async(globals, locals)
File "/lib/python313.zip/_pyodide/_base.py", line 411, in run_async
coroutine = eval(self.code, globals, locals)
File "<exec>", line 48, in <module>
File "<exec>", line 40, in TriggerPlaybook
File "/lib/python3.13/site-packages/requests/api.py", line 115, in post
return request("post", url, data=data, json=json, **kwargs)
File "/lib/python3.13/site-packages/requests/api.py", line 59, in request
~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lib/python3.13/site-packages/requests/sessions.py", line 589, in request
File "/lib/python3.13/site-packages/requests/sessions.py", line 703, in send
File "/lib/python3.13/site-packages/requests/adapters.py", line 667, in send
resp = conn.urlopen(
method=request.method,
...<9 lines>...
chunked=chunked,
)
File "/lib/python3.13/site-packages/urllib3/connectionpool.py", line 787, in urlopen
response = self._make_request(
conn,
...<10 lines>...
**response_kw,
)
File "/lib/python3.13/site-packages/urllib3/connectionpool.py", line 493, in _make_request
conn.request(
~~~~~~~~~~~~^
method,
^^^^^^^
...<6 lines>...
enforce_content_length=enforce_content_length,
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
)
^
File "/lib/python3.13/site-packages/urllib3/contrib/emscripten/connection.py", line 116, in request
self._response = send_streaming_request(request)
~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
File "/lib/python3.13/site-packages/urllib3/contrib/emscripten/fetch.py", line 453, in send_streaming_request
return send_jspi_request(request, True)
File "/lib/python3.13/site-packages/urllib3/contrib/emscripten/fetch.py", line 610, in send_jspi_request
body_stream_js = response_js.body.getReader()
^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'JsNull' object has no attribute 'getReader'
```

# Workaround
Currently I have worked around the issue by monkey patching `send_jspi_request` as follows
```python
import requests
import urllib3.contrib.emscripten.fetch as fetch

original_send_jspi_request = fetch.send_jspi_request

def new_send_jspi_request(request, streaming: bool):
    streaming = False
    return original_send_jspi_request(request, streaming)

fetch.send_jspi_request = new_send_jspi_request
```

juntyr:
JS response body may be null, for instance for HEAD requests, and this was not yet handled in all cases.

cc @hoodmane @joemarshall

### DIFF
--- a/src/urllib3/contrib/emscripten/fetch.py
+++ b/src/urllib3/contrib/emscripten/fetch.py
@@ -47,6 +47,7 @@ from pyodide.ffi import (  # type: ignore[import-not-found]
     JsArray,
     JsException,
     JsProxy,
+    jsnull,
     to_js,
 )
 
@@ -528,7 +529,9 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
 
         headers = dict(Parser().parsestr(js_xhr.getAllResponseHeaders()))
 
-        if not is_in_browser_main_thread():
+        if js_xhr.response is jsnull:
+            body = b""
+        elif not is_in_browser_main_thread():
             body = js_xhr.response.to_py().tobytes()
         else:
             body = js_xhr.response.encode("ISO-8859-15")
@@ -596,19 +599,20 @@ def send_jspi_request(
         else:
             headers[str(iter_value_js.value[0])] = str(iter_value_js.value[1])
     status_code = response_js.status
-    body: bytes | io.RawIOBase = b""
+    body: bytes | io.RawIOBase
 
     response = EmscriptenResponse(
         status_code=status_code, headers=headers, body=b"", request=request
     )
-    if streaming:
+    if response_js.body is jsnull:
+        body = b""
+    elif streaming:
         # get via inputstream
-        if response_js.body is not None:
-            # get a reader from the fetch response
-            body_stream_js = response_js.body.getReader()
-            body = _JSPIReadStream(
-                body_stream_js, timeout, request, response, js_abort_controller
-            )
+        # get a reader from the fetch response
+        body_stream_js = response_js.body.getReader()
+        body = _JSPIReadStream(
+            body_stream_js, timeout, request, response, js_abort_controller
+        )
     else:
         # get directly via arraybuffer
         # n.b. this is another async JavaScript call.
@@ -618,7 +622,7 @@ def send_jspi_request(
             js_abort_controller,
             request=request,
             response=response,
-        ).to_py()
+        ).to_bytes()
     response.body = body
     return response
 
--- a/changelog/4953.bugfix.rst
+++ b/changelog/4953.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed no response body (e.g. from HEAD requests) handling on Emscripten.
--- a/src/urllib3/contrib/emscripten/emscripten_fetch_worker.js
+++ b/src/urllib3/contrib/emscripten/emscripten_fetch_worker.js
@@ -22,6 +22,16 @@ self.addEventListener("message", async function (event) {
     if (!value || curOffset >= value.length) {
       // read another buffer if required
       try {
+        if (reader === null) {
+          // no response body - clear connection and return
+          connections.delete(connectionID);
+          Atomics.store(intBuffer, 0, Status.SUCCESS_EOF);
+          Atomics.notify(intBuffer, 0);
+          // finished reading successfully
+          // return from event handler
+          return;
+        }
+
         let readResponse = await reader.read();
 
         if (readResponse.done) {
@@ -85,7 +95,7 @@ self.addEventListener("message", async function (event) {
       intBuffer[1] = written;
       // make a connection
       connections.set(connectionID, {
-        reader: response.body.getReader(),
+        reader: response.body ? response.body.getReader() : null,
         intBuffer: intBuffer,
         byteBuffer: byteBuffer,
         value: undefined,
--- a/test/contrib/emscripten/test_emscripten.py
+++ b/test/contrib/emscripten/test_emscripten.py
@@ -435,6 +435,67 @@ def test_specific_method(
     )
 
 
+def test_null_body_non_streaming(
+    selenium_coverage: typing.Any,
+    testserver_http: PyodideServerInfo,
+) -> None:
+    """A 204 makes ``fetch`` set ``response.body`` to JavaScript ``null``.
+
+    Pyodide converts that to ``JsNull``, which is *not* ``None``, so an
+    ``is not None`` guard lets it through and the reader call then fails with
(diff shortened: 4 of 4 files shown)


src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -596,19 +599,20 @@ def send_jspi_request(
...
 
     response = EmscriptenResponse(
         status_code=status_code, headers=headers, body=b"", request=request
     )
-    if streaming:
+    if response_js.body is jsnull:
+        body = b""
+    elif streaming:
         # get via inputstream
-        if response_js.body is not None:
```
juntyr, line 606: @hoodmane The body here is still a JSProxy, right?, and from what I can tell, `pyodide.ffi.jsnull` is not `None`

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -618,7 +622,7 @@ def send_jspi_request(
             js_abort_controller,
             request=request,
             response=response,
-        ).to_py()
```
juntyr, line 621: @hoodmane `to_py` returns `memory` but we want `bytes` here

illia-v:
Is this related to https://github.com/urllib3/urllib3/issues/3723?

hoodmane:
Yes

hoodmane:
@juntyr can you add tests to https://github.com/urllib3/urllib3/blob/main/test/contrib/emscripten/test_emscripten.py?

src/urllib3/contrib/emscripten/fetch.py
hoodmane: Yeah `if response_js.body is jsnull:` is the correct check for null. It is indeed not `None`.

hoodmane:
Changes make sense to me other than the missing test coverage.

juntyr:
@hoodmane I unfortunately do not currently have the time to push this PR to completion but am happy if somebody else can take over

DPS0340:
@juntyr said in April they were out of time to finish this, and @hoodmane's only outstanding request was test coverage. That is a small enough gap that I wrote it rather than let the PR keep sitting — **the tests are on a branch off this PR's head**, so you can cherry-pick one commit and nothing else about the PR changes:

```
https://github.com/DPS0340/urllib3/tree/tests-for-4953
git fetch https://github.com/DPS0340/urllib3 tests-for-4953 && git cherry-pick d783d92f
```

I'm not opening a competing PR; this is yours if you want it, and I'm happy to just hand it over.

**Two tests, because the null body reaches a different branch on each path:**

- `test_null_body_non_streaming` — goes through `arrayBuffer()`.
- `test_null_body_streaming` (webworker, both JSPI and non-JSPI via `prefer_jspi`) — reads `response.body`, which is `_JSPIReadStream` under JSPI and the worker's `getReader()` otherwise. That `getReader()` on `JsNull` is the exact `AttributeError` in #3723, so covering only the non-streaming path would miss the reported failure.

**One decision worth flagging: they use a 204, not HEAD**, even though the changelog here says "e.g. from HEAD requests". I started with HEAD and checked it against the dummyserver instead of assuming:

```
HEAD  /                              -> 200  body=b'Dummy server!'
GET   /status?status=204 NO CONTENT  -> 204  body=b''
```

The route answers `HEAD /` with a body, so a HEAD-based test would not reliably reach the null-body branch. The fetch spec gives a 204 a null body **by status**, independent of what the server sends, which makes it the deterministic trigger. `/status` already accepts an arbitrary status code, so this needs no new server endpoint.

The fix itself I have not touched — the worker-side `reader === null` guard and the `is jsnull` checks are as you wrote them.

I cannot run these locally (they need the Pyodide/selenium harness), so they are unverified in that sense. `black`, `isort` and `flake8` are clean on the file, and `mypy` reports nothing on the added lines.

juntyr:
@DPS0340 That was very kind of you, thank you <3

src/urllib3/contrib/emscripten/fetch.py
```diff
@@ -528,7 +529,9 @@ def send_request(request: EmscriptenRequest) -> EmscriptenResponse:
 
         headers = dict(Parser().parsestr(js_xhr.getAllResponseHeaders()))
 
-        if not is_in_browser_main_thread():
+        if js_xhr.response is jsnull:
+            body = b""
```
hoodmane, line 533: I guess we still need coverage for this line.

PR #4960, opened 2026-04-11
Fix full read with `amt=None` after a partial read

### LINKED ISSUE #3636: Partial then full read with `decode_content=True` returns partial/truncated data

illia-v:
Fixes #3636

A partial read could leave some data inside `_decoded_buffer`, this buffered data was ignored during the next full read with `amt=None`.

### DIFF
--- a/changelog/3636.bugfix.rst
+++ b/changelog/3636.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed a bug where ``HTTPResponse.read(amt=None)`` ignored decompressed data
+buffered from previous partial reads. Also prevented ``HTTPResponse.read()``
+from caching only a chunk of data after a partial read when
+``cache_content=True``.
--- a/changelog/3636.bugfix.rst
+++ b/changelog/3636.bugfix.rst
@@ -1,4 +1,2 @@
-Fixed a bug where ``HTTPResponse.read(amt=None)`` ignored decompressed data
-buffered from previous partial reads. Also prevented ``HTTPResponse.read()``
-from caching only a chunk of data after a partial read when
-``cache_content=True``.
+Fixed a bug where ``HTTPResponse.read(amt=None)`` was ignoring decompressed data
+buffered from previous partial reads.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -485,7 +485,6 @@ class BaseHTTPResponse(io.IOBase):
         self.reason = reason
         self.decode_content = decode_content
         self._has_decoded_content = False
-        self._uncached_read_occured = False
         self._request_url: str | None = request_url
         self.retries = retries
 
@@ -1090,7 +1089,6 @@ class HTTPResponse(BaseHTTPResponse):
             amt = None
         elif amt is not None:
             cache_content = False
-            self._uncached_read_occured = True
 
             if self._decoder and self._decoder.has_unconsumed_tail:
                 decoded_data = self._decode(
@@ -1122,7 +1120,7 @@ class HTTPResponse(BaseHTTPResponse):
                 self._decoded_buffer.put(data)
                 data = self._decoded_buffer.get_all()
 
-            if cache_content and not self._uncached_read_occured:
+            if cache_content:
                 self._body = data
         else:
             # do not waste memory on buffer when not decoding
@@ -1175,7 +1173,6 @@ class HTTPResponse(BaseHTTPResponse):
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
-        self._uncached_read_occured = True
         if decode_content is None:
             decode_content = self.decode_content
         if amt and amt < 0:
@@ -1385,7 +1382,6 @@ class HTTPResponse(BaseHTTPResponse):
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
-        self._uncached_read_occured = True
         self._init_decoder()
         # FIXME: Rewrite this method and make it a class with a better structured logic.
         if not self.chunked:
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -485,6 +485,7 @@ class BaseHTTPResponse(io.IOBase):
         self.reason = reason
         self.decode_content = decode_content
         self._has_decoded_content = False
+        self._uncached_read_occured = False
         self._request_url: str | None = request_url
         self.retries = retries
 
@@ -1089,6 +1090,7 @@ class HTTPResponse(BaseHTTPResponse):
             amt = None
         elif amt is not None:
             cache_content = False
+            self._uncached_read_occured = True
 
             if self._decoder and self._decoder.has_unconsumed_tail:
                 decoded_data = self._decode(
@@ -1114,7 +1116,13 @@ class HTTPResponse(BaseHTTPResponse):
 
         if amt is None:
             data = self._decode(data, decode_content, flush_decoder)
-            if cache_content:
+            # It's possible that there is buffered decoded data after a
+            # partial read.
+            if decode_content and len(self._decoded_buffer) > 0:
+                self._decoded_buffer.put(data)
+                data = self._decoded_buffer.get_all()
+
+            if cache_content and not self._uncached_read_occured:
                 self._body = data
         else:
             # do not waste memory on buffer when not decoding
@@ -1167,6 +1175,7 @@ class HTTPResponse(BaseHTTPResponse):
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
+        self._uncached_read_occured = True
         if decode_content is None:
             decode_content = self.decode_content
         if amt and amt < 0:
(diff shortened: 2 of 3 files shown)


test/test_response.py
```diff
@@ -1969,8 +2039,8 @@ def pop_current_chunk(self, amt: int = -1, till_crlf: bool = False) -> bytes:
                 self.cur_chunk = self.cur_chunk[amt:]
             return chunk_part
 
-    def readline(self) -> bytes:
-        return self.pop_current_chunk(till_crlf=True)
+    def readline(self, amt: int = -1) -> bytes:
+        return self.pop_current_chunk(amt, till_crlf=amt < 0)
```
illia-v, line 2043: This was changed to fix
```
self = <http.client.HTTPResponse object at 0x7ca9336e2b30>

    def _read_next_chunk_size(self):
        # Read the next chunk size from the file
>       line = self.fp.readline(_MAXLINE + 1)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: MockChunkedEncodingResponse.readline() takes 1 positional argument but 2 were given

/usr/lib64/python3.14/http/client.py:544: TypeError
```

changelog/3636.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Fixed a bug where ``HTTPResponse.read(amt=None)`` ignored decompressed data
+buffered from previous partial reads. Also prevented ``HTTPResponse.read()``
+from caching only a chunk of data after a partial read when
+``cache_content=True``.
```
pquentin, line 4: From the changelog alone, I don't understand whether "Also prevented ..." is a separate problem (a bad thing) or a separate fix (a good thing).

Also, if it's really separate, should it be a different PR?

changelog/3636.bugfix.rst
```diff
@@ -0,0 +1,4 @@
+Fixed a bug where ``HTTPResponse.read(amt=None)`` ignored decompressed data
```
pquentin, line 1:
```suggestion
Fixed a bug where ``HTTPResponse.read(amt=None)`` was ignoring decompressed data
```

src/urllib3/response.py
```diff
@@ -1089,6 +1090,7 @@ def read(
             amt = None
         elif amt is not None:
             cache_content = False
+            self._uncached_read_occured = True
```
pquentin, line 1093: Note: at this point, the uncached read hasn't occurred yet, so the variable name is misleading. But I don't have a better name to offer.

src/urllib3/response.py
```diff
@@ -1167,6 +1175,7 @@ def read1(
             If True, will attempt to decode the body based on the
             'content-encoding' header.
         """
+        self._uncached_read_occured = True
```
pquentin, line 1178: Why isn't this closer to the `_raw_read()` call? Because if there's already data to decode, we still want that set?

src/urllib3/response.py
```diff
@@ -1376,6 +1385,7 @@ def read_chunked(
...
```
pquentin, line 1388: Same question as read1

illia-v:
@pquentin thanks for the review, the suggestions totally make sense. I moved the part related to caching to a separate PR #4967.

test/test_response.py
```diff
@@ -898,6 +898,35 @@ def test_read_multi_decoding_too_many_links(self) -> None:
...
+        )
+        assert r.read(3) == b"foo"
+        # Force putting some decoded data in the buffer as the buffer
+        # is normally empty unless decoder has issues like not
+        # respecting `max_length` https://github.com/google/brotli/issues/1396
+        middle_part = r._decode(
+            r._raw_read(), decode_content=True, flush_decoder=False, max_length=3
+        )
+        assert middle_part == b"bar"
+        r._decoded_buffer.put(middle_part)
```
sethmlarson, line 915: Is there a way we can trigger this behavior without manipulating private APIs? Such as reading a single byte that expands into multiple? I'm thinking using `.compress(b"x" * 1000)` and then calling `.read(1)`, won't that buffer?

sethmlarson (approved):
One comment, otherwise LGTM. Thanks Illia!

test/test_response.py
illia-v: No, it won't buffer because decompressors do not put excessive bytes into `_decoded_buffer` normally, only the Brotli library does as mentioned in the comment.
I hope Brotli fixes this at some point, but after that and before we drop `_decoded_buffer`, I'd still like the test to check that data in `_decoded_buffer` is not ignored.

PR #4962, opened 2026-04-13
Fix proxy_ssl_context mutation when target uses cert_reqs='CERT_NONE'

### LINKED ISSUE #4961: proxy_ssl_context is mutated when target uses cert_reqs='CERT_NONE'
## Summary

When using `ProxyManager` with an HTTPS proxy and `cert_reqs='CERT_NONE'` for the target, urllib3 mutates the caller-provided `proxy_ssl_context` by overwriting its `verify_mode` to `CERT_NONE`. The mutation is permanent - it persists on the context object across all subsequent connections made through that `ProxyManager`.

## Reproduction

```python
import ssl
import urllib3
from urllib3.util.ssl_ import create_urllib3_context

proxy_ssl_context.load_verify_locations("/path/to/proxy-ca.pem")
proxy_ssl_context.check_hostname = False
proxy_ssl_context.verify_mode = ssl.CERT_REQUIRED

print(f"Before: {proxy_ssl_context.verify_mode}")  # CERT_REQUIRED (2)

pm = urllib3.ProxyManager(
    "https://proxy.example.com:8443",
    cert_reqs="CERT_NONE",
)
pm.request("GET", "https://target.example.com/")

print(f"After: {proxy_ssl_context.verify_mode}")  # CERT_NONE (0) - mutated!
```

## Root Cause

Two issues in `connection.py`:

1. **`_connect_tls_proxy()`** passes `self.cert_reqs` (the *target's* cert_reqs) to `_ssl_wrap_socket_and_match_hostname()` for the proxy TLS handshake. When the target has `cert_reqs='CERT_NONE'`, this value leaks into the proxy path.

2. **`_ssl_wrap_socket_and_match_hostname()`** unconditionally sets `context.verify_mode = resolve_cert_reqs(cert_reqs)` on line 937, even when `context` is a user-provided `ssl_context` rather than an internally-created one. Since `ProxyConfig` holds a reference to the user's context (not a copy), the mutation sticks.

## Expected Behavior

A user-provided `proxy_ssl_context` should never be modified by urllib3. The proxy TLS handshake should use the proxy context's own `verify_mode`, not the target connection's `cert_reqs`.

## Environment

- urllib3 `main` branch (confirmed on v2.6.4.dev20)
- Python 3.13+ (likely affects all supported versions)

yonatanmgr:
_connect_tls_proxy() was passing the target connection's cert_reqs to _ssl_wrap_socket_and_match_hostname(), which then unconditionally overwrote verify_mode on the user-provided proxy_ssl_context. This caused the proxy context to permanently inherit the target's CERT_NONE setting.

Two changes:
- _connect_tls_proxy() now derives cert_reqs from the proxy context's own verify_mode when a proxy_ssl_context is provided.
- _ssl_wrap_socket_and_match_hostname() no longer mutates verify_mode on user-provided ssl_context objects.

Fixes #4961

### DIFF
--- a/changelog/4961.bugfix.rst
+++ b/changelog/4961.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed ``proxy_ssl_context`` being mutated when ``cert_reqs='CERT_NONE'`` was
+set for the target connection. The target's ``cert_reqs`` no longer leaks into
+the proxy TLS handshake, and user-provided ``ssl_context`` objects passed to
+``_ssl_wrap_socket_and_match_hostname()`` are no longer modified in place.
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -865,9 +865,19 @@ class HTTPSConnection(HTTPConnection):
         # `_connect_tls_proxy` is called when self._tunnel_host is truthy.
         proxy_config = typing.cast(ProxyConfig, self.proxy_config)
         ssl_context = proxy_config.ssl_context
+
+        # When a proxy_ssl_context is provided, derive cert_reqs from the
+        # proxy's own context rather than using the target connection's
+        # cert_reqs. Otherwise the target's CERT_NONE can leak into the
+        # proxy handshake, and the caller's context object gets mutated.
+        if ssl_context is not None:
+            proxy_cert_reqs = ssl_context.verify_mode
+        else:
+            proxy_cert_reqs = None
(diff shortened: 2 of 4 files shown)


illia-v:
Thanks for the contribution @yonatanmgr! https://github.com/urllib3/urllib3/pull/5093 has fixed this issue and some related ones.

PR #4967, opened 2026-04-19
Prevent content caching after a partial read

illia-v:
#4960 fixes `read(amt=None)` calls after a partial read. This PR addresses a related issue where only a portion of content could be cached.

### DIFF
--- a/src/urllib3/contrib/emscripten/response.py
+++ b/src/urllib3/contrib/emscripten/response.py
@@ -36,6 +36,7 @@ class EmscriptenHttpResponseWrapper(BaseHTTPResponse):
     ):
         self._pool = None  # set by pool class
         self._body = None
+        self._uncached_read_occured = False
         self._response = internal_response
         self._url = url
         self._connection = connection
@@ -160,10 +161,13 @@ class EmscriptenHttpResponseWrapper(BaseHTTPResponse):
                 # don't cache partial content
                 cache_content = False
                 data = self._response.body.read(amt)
+                self._uncached_read_occured = True
             else:  # read all we can (and cache it)
                 data = self._response.body.read()
-                if cache_content:
+                if cache_content and not self._uncached_read_occured:
                     self._body = data
+                else:
+                    self._uncached_read_occured = True
             if self.length_remaining is not None:
                 self.length_remaining = max(self.length_remaining - len(data), 0)
             if len(data) == 0 or (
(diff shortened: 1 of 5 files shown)


src/urllib3/contrib/emscripten/response.py
```diff
@@ -160,10 +161,13 @@ def read(
                 # don't cache partial content
                 cache_content = False
                 data = self._response.body.read(amt)
+                self._uncached_read_occured = True
```
sethmlarson: Do we want to call it `uncached_read`? We've been using "partial read" language elsewhere, maybe `_partial_read_occurred`? (also nit: "occurred")

src/urllib3/response.py
```diff
@@ -1411,6 +1415,7 @@ def read_chunked(
                     chunk = b""
                 else:
                     self._update_chunk_length()
+                    self._uncached_read_occured = True
```
sethmlarson: Is this needed, or is it set from a different `.read()` call?

test/test_response.py
```diff
@@ -219,6 +219,47 @@ def test_cache_content_preload_false(self) -> None:
         assert r._body == b"foo"  # type: ignore[comparison-overlap]
         assert r.data == b"foo"
 
+    def test_cache_content_with_explicit_read_call(self) -> None:
+        fp = BytesIO(b"foo")
+        r = HTTPResponse(fp, preload_content=False)
+        assert r.read(cache_content=True) == b"foo"
```
sethmlarson: Can we test with `.read(None, cache_content=True)` too?

sethmlarson (dismissed):
Few comments, otherwise LGTM!

src/urllib3/response.py
illia-v: This `read_chunked` code path uses `http.client` code via `self._handle_chunk` instead of our `read` method

src/urllib3/contrib/emscripten/response.py
illia-v: Thanks for spotting the typo!

Currently `self._uncached_read_occured = True` is set not only after partial reads, but also after any read where content was not cached. Calling it partial would be a bit misleading.

test/test_response.py
illia-v: Done, I added `@pytest.mark.parametrize("read_args", ((), (None,), (-1,)))` same as we have for a couple of other tests

sethmlarson (approved):

PR #4974, opened 2026-04-26
Avoid infinite loop in `HTTPResponse.read_chunked` when `amt` is zero

### LINKED ISSUE #3793: `HTTPResponse.stream` can fall into infinite loop when `amt` is zero

Cycloctane:
#3797 didn't cover the `HTTPResponse.read_chunked`. `read_chunked(0)` still runs into infinite loop.

Fixes #3793


illia-v (approved):
Good catch, LGTM, thanks!

Issue #1362, opened 2018-04-17
Malformed `Set-Cookie` header containing CRLF is ignored


paultiplady:
Raising here after reporting in `requests`: https://github.com/requests/requests/issues/4592#issuecomment-382004884

When receiving a malformed Set-Cookie header containing a CRLF, the `\r\n` character is incorrectly included in the header value, causing `requests` to include it in the generated `Cookie` header.

# Expected Result

Per RFC https://tools.ietf.org/html/rfc7230#section-3.2.4, "A user agent that receives an obs-fold in a response message that is not within a message/http container MUST replace each received obs-fold with one or more SP octets prior to interpreting the field value."

Or, just raise an error on the invalid HTTP header.

# Actual Result

An invalid Set-Cookie value was returned containing a CRLF character.

# Reproduction Steps

Requests maintainers came up with a quick repro using that library in https://github.com/requests/requests/issues/4592#issuecomment-381857896.

Here's the bogus cookie (extracted from the raw HTTP text):

```
Set-Cookie: ___utmvbtouVBFmB=gZg
    XbNOjalT: Lte; path=/; Max-Age=900
```
Which gives:

```
Cookie: ___utmvbtouVBFmB=gZg\r\n    XbNOjalT: Lte; <other-cookies>
```

njsmith:
I'm surprised this wasn't fixed by https://github.com/urllib3/urllib3/pull/1318. Or was it, but requests is using an older version?

nateprewitt:
That's a good callout @njsmith, it looks like this may be fixed, but urllib3 hasn't cut a release in almost a year. I don't have time to move the original repro to urllib3 tonight but can take a look tomorrow.

SethMichaelLarson:
Thanks all for the investigation! Can someone confirm whether this still occurs on master? If it's already fixed I'll close this issue as duplicate. :tada:

nateprewitt:
Went ahead and tested. I can't get this to work in Requests with the current urllib3 master, but that wasn't making sense. I made a [urllib3-only repro](https://gist.github.com/nateprewitt/d4df91be2f24cec7d7d531a82f7aa019) and it appears to be fixed for the newline case. I'll dig into it on the Requests end and see if we're botching something.

urllib3 doesn't however handle the other case I posed in the first test of whitespace between a header name and the colon.

Same section of the RFC:
> No whitespace is allowed between the header field-name and colon.  In
   the past, differences in the handling of such whitespace have led to
[...]

PR #4982, opened 2026-04-30
Reject whitespace in HTTP/1.1 header names

webhop123:
Fixes #1362

This tightens the HTTP/1.1 request path so header field names have to be valid tokens before they are written to the request. That catches names with spaces or tabs from normal request headers and tunnel headers instead of letting them reach the wire.

I left the header collection behavior alone so this only affects outbound serialization.

Tested with:
- python -m pytest test\test_connection.py test\test_http2_connection.py
- python -m pre_commit run black --files src\urllib3\connection.py test\test_connection.py
- python -m pre_commit run isort --files src\urllib3\connection.py test\test_connection.py
- python -m pre_commit run flake8 --files src\urllib3\connection.py test\test_connection.py

### DIFF
--- a/changelog/1362.bugfix.rst
+++ b/changelog/1362.bugfix.rst
@@ -1,0 +1,1 @@
+Reject outbound HTTP/1.1 header names containing whitespace including tabs and newlines.
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -45,6 +45,7 @@ from ._version import __version__
 from .exceptions import (
     ConnectTimeoutError,
     HeaderParsingError,
+    InvalidHeader,
     NameResolutionError,
     NewConnectionError,
     ProxyError,
@@ -79,6 +80,16 @@ RECENT_DATE = datetime.date(2025, 1, 1)
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 
 
+def _validate_http1_header_name(header: str | bytes) -> None:
+    try:
+        header_name = to_str(header, "ascii")
+    except UnicodeDecodeError as e:
+        raise InvalidHeader(f"Invalid header name {header!r}") from e
+
+    if not header_name or _CONTAINS_CONTROL_CHAR_RE.search(header_name):
+        raise InvalidHeader(f"Invalid header name {header!r}")
+
+
 class HTTPConnection(_HTTPConnection):
     """
     Based on :class:`http.client.HTTPConnection` but provides an extra constructor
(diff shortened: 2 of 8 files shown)


webhop123:
Looks like the remaining red job is unrelated to this change. PyPy is failing while setting up the dummy Hypercorn server with `trio.TrioInternalError` from `anyio.from_thread.start_blocking_portal(backend=trio)`, before it gets to this patch. I do not have permission to rerun the job from the fork, but the rest of the matrix, coverage, lint, docs, and downstream checks are green.

changelog/1362.bugfix.rst
```diff
@@ -0,0 +1 @@
+Rejected outbound HTTP/1.1 header names containing whitespace.
```
sigmavirus24:
```suggestion
Reject outbound HTTP/1.1 header names containing whitespace including tabs and newlines.
```

src/urllib3/connection.py
```diff
@@ -79,6 +79,16 @@ class BaseSSLError(BaseException):  # type: ignore[no-redef]
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 
 
+def _validate_http1_header_name(header: str | bytes) -> None:
+    try:
+        header_name = to_str(header, "ascii")
+    except UnicodeDecodeError as e:
+        raise ValueError(f"Invalid header name {header!r}") from e
```
sigmavirus24: Please use a different exception class

sigmavirus24 (changes requested):
We need a way to allow folks deliberately using these characters for security testing to bypass this otherwise as soon as we release this week have a slee of bugs accusing us of being horrible people

webhop123:
Thanks, updated this to use `InvalidHeader` and added `assert_header_names=False` as an escape hatch for folks intentionally sending malformed headers. Also applied the changelog wording.

sigmavirus24:
I wouldn't go about this by having a boolean parameter as someone somewhere is going to start documenting you need to set this to False, some LLM will be trained on that garbage and then everyone relying on an LLM not to give them slop will get sloppier urllib3 have. There needs to be an escape hatch but your LLM won't come up with it for you as it hasn't understood how this project does it historically.

webhop123:
> I wouldn't go about this by having a boolean parameter as someone somewhere is going to start documenting you need to set this to False, some LLM will be trained on that garbage and then everyone relying on an LLM not to give them slop will get sloppier urllib3 have. There needs to be an escape hatch but your LLM won't come up with it for you as it hasn't understood how this project does it historically.

I’ll remove assert_header_names from the public connection/pool kwargs and rework
the escape hatch to be explicit/advanced instead. Is the custom ConnectionCls path
the historical pattern you had in mind here?

PR #4983, opened 2026-05-04
Avoid repeated SSL verify location loads

### LINKED ISSUE #3623: ssl_wrap_socket loads SSL certificates into the same context repeatedly

rubinovitz:
Fixes #3623.

When callers pass the same `SSLContext` and the same CA certificate locations to `ssl_wrap_socket`, urllib3 currently calls `SSLContext.load_verify_locations()` on every wrap. This can repeatedly reload the same certificates into the same context and is expensive for high-volume clients.

This caches loaded verify locations per context using a `WeakKeyDictionary`, so the same `(ca_certs, ca_cert_dir, ca_cert_data)` tuple is only loaded once for a given context. Custom context implementations that cannot be weak-referenced fall back to the previous behavior.

Validation:
- repo-local virtualenv in disposable checkout
- `PYTHONPATH=src .venv/bin/python -m pytest test/test_util.py -q` -> 326 passed, 6 skipped
- `PYTHONPATH=src .venv/bin/python -m pytest test/test_ssl.py test/test_util.py -q` -> 371 passed, 6 skipped
- Tenfold Docker eval for urllib3 #3623: 0.003423s

### DIFF
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -8,6 +8,7 @@ import sys
 import typing
 import warnings
 from binascii import unhexlify
+from weakref import WeakKeyDictionary
 
 from ..exceptions import ProxySchemeUnsupported, SSLError
 from .url import _BRACELESS_IPV6_ADDRZ_RE, _IPV4_RE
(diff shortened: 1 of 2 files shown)


sigmavirus24 (commented):
I know what happens, so I'm not asking for my benefit... 

Some people use this with private trust bundles. Those may change. What happens if these files change while a process using urllib3 is running with this in memory cache?

PR #4989, opened 2026-05-08
Extract test dependency group, deduplicate dev and mypy

### LINKED ISSUE #3594: Possible improvements to dependency groups
### Context

There are two things I notice about the current dependency groups.

----

First, there is a lot of repetition between `dev` and `mypy`:

https://github.com/urllib3/urllib3/blob/910f0314998aa17050383f3daeef6fc0fef8af47/pyproject.toml#L58-L94

I am not sure exactly what to say that the common dependencies represent, but if we could figure out something sensible to call them, we could represent them in their own dependency group:

```toml
some_other_group = [
    "anyio[trio]>=4.8.0",
    "cryptography>=44.0.2",
    "httpx>=0.28.1",
    "hypercorn>=0.15.0",  # dev lacks the version bound; both are affected by tool.uv.sources.hypercorn
    "idna>=3.10",
    "pytest>=8.3.4",
    "quart>=0.20.0",
    "quart-trio>=0.12.0",
    "trio>=0.27.0",
    "trustme>=1.2.1",
]
```

and then use [Dependency Group Includes](https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-group-include):

```toml
dev = [
    "build>=1.2.2.post1",
    "coverage>=7.8.0",
    "h2>=4.1.0",
    "pyopenssl>=25.0.0",
    "pysocks>=1.7.1",
    "pytest-memray==1.7.0 ; python_full_version < '3.14' and implementation_name == 'cpython' and sys_platform != 'win32'",
    "pytest-socket>=0.7.0",
    "pytest-timeout>=2.3.1",
    "towncrier>=24.8.0",
    {include-group = "some_other_group"},
]
mypy = [
    "mypy>=1.14.1",
    "types-requests>=2.32.0.20241016",
    "nox>=2024.10.9",
    "zstandard>=0.23.0",
    {include-group = "some_other_group"},
]
```

----

Second, in Fedora we really benefit from dependency groups because we can use them to generate build-time requirements for running the tests rather than manually enumerating them; see discussion in https://src.fedoraproject.org/rpms/python-urllib3/pull-request/44.

However, we really just want the dependencies for running the tests, without profilers (like `pytest-memray`), coverage analysis (`coverage`), or tools for maintainer tasks (`build`, `towncrier`). It would be really helpful if there were a dependency group (`test`, presumably) that *just* listed test dependencies. Even if you decided not to split out coverage and profiling dependencies, and even if we still had to patch out a couple of unwanted dependencies, this would be a much more useful starting point than the `dev` group is.

If you decide to do this, the “dependency group includes” technique demonstrated above would make it easy to include a new `test` group in the `dev` group, so `dev` could still install “everything.”

### Alternatives

For the first suggestion, repetition of dependency specifications seems to be more of an inconvenience for upstream maintainers than it is for downstream packagers and users. The alternative is the *status quo*, and it’s generally adequate.

For the second suggestion, one alternative for the Fedora package is to manually list `BuildRequires` for testing, as we did prior to 2.4.0. This is acceptable; it requires some care to make sure that the manual list remains synchronized with the actual upstream dependencies. The other alternative is to patch the contents of the `dev` dependency group – using `sed` or a TOML-specific patching tool – before using it to generate `BuildRequires`, as currently proposed in https://src.fedoraproject.org/rpms/python-urllib3/pull-request/44. This is viable, but the less patching we can do, the less trouble it is, and the fewer errors we are likely to make.

### Duplicate

I did not find any similar requests.

### Contribution

Yes, I would be willing to submit a PR if a rough consensus on design and nomenclature emerges.

xronocode:
Closes #3594.

## Summary

Introduces a dedicated `test` dependency group containing only what is needed to run the test suite — no profiling (`coverage`, `pytest-memray`) and no maintainer tooling (`build`, `towncrier`). Downstream packagers (e.g. Fedora, see linked PR in the issue) can install just this group to generate build-time test requirements without patching the contents of `dev`.

`dev-base` now layers on top of `test` and adds the maintainer/profiling tools, so `dev` and `dev-min-pyopenssl` retain the same final dependency closure as before. `mypy` also includes `test` instead of repeating its contents, eliminating the duplication called out in the issue.

This implements @pquentin's suggestion in https://github.com/urllib3/urllib3/issues/3594#issuecomment-2997196389:

> Sure, that sounds good, we can add a new group called `test` and include it in `dev` and `mypy`.

## Equivalence

I verified via `uv export --frozen --only-group <name> --no-hashes` that the resolved package set is **byte-for-byte identical** for `dev`, `dev-base`, and `dev-min-pyopenssl` before vs. after.

The `mypy` group additionally pulls `pysocks`, `pytest-socket`, and `pytest-timeout` (because they now come in via the shared `test` group). `nox -s mypy` still passes with no changes — the existing `# type: ignore[import-untyped]` annotations on the SOCKS imports cover them. Happy to split `mypy` back out from `test` if you'd prefer to keep its environment leaner.

## On `idna`

@pquentin asked in the issue whether `idna` is still needed. I checked: it is.

- `src/urllib3/util/url.py:333` does a conditional `import idna` (try/except `ImportError`); without the package, `_idna_encode` raises `LocationParseError`.
- `test/test_util.py` exercises both paths: `test_create_connection_with_valid_idna_labels` requires the package to be present, and `class TestUtilWithoutIdna` (line 1093) uses an `ImportBlocker("idna")` to verify the fallback when it's missing.

The other places that mention `idna` in the source (`connection.py`, `util/connection.py`) use Python's stdlib `encodings.idna` codec via `host.encode("idna")` and don't need the package. So `idna` stays in `dev`/`mypy`.

## Test plan

- [x] `nox -s mypy` — passes (83 files, no issues)
- [x] `nox -s lint` — passes
- [x] `uv lock --check` — clean after the refactor
- [x] `uv export` diff for `dev`/`dev-base`/`dev-min-pyopenssl`: identical closures
- [x] `pre-commit run` on touched files — passes
- [x] `towncrier check` — fragment found and renders correctly
- [ ] Full CI on push

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,22 +56,29 @@ h2 = [
 ]
 
 [dependency-groups]
-dev-base = [
+# Bare dependencies needed to run the test suite. Kept free of profiling
+# (`coverage`, `pytest-memray`) and maintainer tooling (`build`, `towncrier`)
+# so downstream packagers can install just this group when generating
+# build-time test requirements (see issue #3594).
+test = [
     "anyio[trio]>=4.8.0",
-    "build>=1.2.2.post1",
-    "coverage>=7.8.0",
     "h2>=4.1.0",
     "httpx>=0.28.1",
-    "hypercorn",
+    "hypercorn>=0.15.0",
     "pysocks>=1.7.1",
     "pytest>=8.3.4",
-    "pytest-memray==1.8.0 ; python_full_version < '3.15' and implementation_name == 'cpython' and sys_platform != 'win32'",
     "pytest-socket>=0.7.0",
     "pytest-timeout>=2.3.1",
     "quart>=0.20.0",
     "quart-trio>=0.12.0",
-    "towncrier>=24.8.0",
     "trio>=0.27.0",
+]
+dev-base = [
+    {include-group = "test"},
+    "build>=1.2.2.post1",
+    "coverage>=7.8.0",
+    "pytest-memray==1.8.0 ; python_full_version < '3.15' and implementation_name == 'cpython' and sys_platform != 'win32'",
+    "towncrier>=24.8.0",
 ]
 dev = [
     {include-group = "dev-base"},
@@ -94,17 +101,11 @@ dev-min-pyopenssl = [
     "trustme==0.9.0",
 ]
 mypy = [
-    "anyio[trio]>=4.8.0",
+    {include-group = "test"},
     "backports.zstd>=1.0.0; python_version<'3.14'",
     "cryptography>=44.0.2",
-    "httpx>=0.28.1",
-    "hypercorn>=0.15.0",
     "idna>=3.10",
     "mypy>=1.14.1",
-    "pytest>=8.3.4",
-    "quart>=0.20.0",
-    "quart-trio>=0.12.0",
-    "trio>=0.27.0",
     "trustme>=1.2.1",
     "types-requests>=2.32.0.20241016",
     "nox>=2024.10.9",
--- a/changelog/3594.feature.rst
+++ b/changelog/3594.feature.rst
@@ -1,0 +1,1 @@
+Added a ``test`` dependency group containing the minimum dependencies needed to run the test suite, intended for downstream packagers. The ``dev-base`` and ``mypy`` groups now include this new group via ``include-group``, removing duplication.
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -56,23 +56,30 @@ h2 = [
...
+# so downstream packagers can install just this group when generating
+# build-time test requirements (see issue #3594).
+test = [
     "anyio[trio]>=4.8.0",
-    "build>=1.2.2.post1",
-    "coverage>=7.8.0",
     "h2>=4.1.0",
     "httpx>=0.28.1",
-    "hypercorn",
+    "hypercorn>=0.15.0",
```
pquentin, line 67: The >= is a bit misleading, since we require the commits in urllib3/hypercorn to pass the test suite. It currently lines up roughly with 0.15, indeed. Happy to keep this if it helps you.

pquentin (approved):
Thanks! LGTM.

hroncok:
Thank you.

PR #4992, opened 2026-05-09
Resolve connection pool queue class lazily

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent

grantf04:
## Summary
- resolve the default `queue.LifoQueue` class when each connection pool is constructed instead of storing an import-time reference
- keep `ConnectionPool.QueueCls` as an explicit override hook for subclasses
- add regression coverage for late monkey-patching and explicit queue-class overrides

Closes #3289.

## Testing
- `.venv/bin/python -m pytest test/test_queue_monkeypatch.py -q`
- `.venv/bin/python -m pytest test/test_connectionpool.py test/test_queue_monkeypatch.py -q`
- `.venv/bin/python -m mypy src/urllib3/connectionpool.py test/test_queue_monkeypatch.py`
- `.venv/bin/python -m nox -rs lint`
- `git diff --check`

### DIFF
--- a/changelog/3289.bugfix.rst
+++ b/changelog/3289.bugfix.rst
@@ -1,0 +1,1 @@
+Changed connection pools to resolve the default ``queue.LifoQueue`` class when a pool is constructed, allowing late monkey-patching to affect newly-created pools.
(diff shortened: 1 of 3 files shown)


illia-v:
Closing this because #3729 showed a better solution

PR #4998, opened 2026-05-10
Remove unnecessary response decoder flushing

### LINKED ISSUE #2799: Do we flush the decoder when reaching EOF in partial reads?

webhop123:
﻿Fixes #2799

## Summary
- remove response decoder output flushing now that Jython is no longer supported
- keep zstd end-of-stream validation through a finish hook so incomplete zstd data still raises DecodeError
- add regression coverage ensuring read(), read1(), and read_chunked() do not flush decoders

## Testing
- python -m pytest test/test_response.py
- python -m pytest test/with_dummyserver/test_connectionpool.py -k "decode or chunked or gzip"
- python -m pre_commit run pyupgrade --files src/urllib3/response.py test/test_response.py
- python -m pre_commit run black --files src/urllib3/response.py test/test_response.py
- python -m pre_commit run isort --files src/urllib3/response.py test/test_response.py
- python -m pre_commit run flake8 --files src/urllib3/response.py test/test_response.py

### DIFF
--- a/changelog/2799.bugfix.rst
+++ b/changelog/2799.bugfix.rst
@@ -1,0 +1,1 @@
+Removed unnecessary response decoder flushing during decompression.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -59,8 +59,8 @@ class ContentDecoder:
     def has_unconsumed_tail(self) -> bool:
         raise NotImplementedError()
 
-    def flush(self) -> bytes:
-        raise NotImplementedError()
+    def finish(self) -> None:
+        return None
 
 
 class DeflateDecoder(ContentDecoder):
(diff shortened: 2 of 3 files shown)


illia-v (commented):
In my opinion, this is an unjustified change that could be breaking, at least for projects that implement `ContentDecoder` in custom decoders with flushing logic.

illia-v:
The issue has been closed https://github.com/urllib3/urllib3/issues/2799#issuecomment-4528894234

PR #5003, opened 2026-05-11
fix: ProxyManager now honors HTTPConnection.default_socket_options

### LINKED ISSUE #2936: ProxyManager doesn't honor HTTPConnection.default_socket_options

agent-leo:
## Summary

Fixes #2936.

When using , proxy connection pools previously defaulted  to an empty list. This silently discarded any user-modified  (e.g. ), making it impossible to apply global socket options to proxy connections without explicitly passing  to every  instance.

## Changes

- Proxy connection pools now inherit  from the connection class, minus . This preserves the original intent of enabling Nagle's algorithm for proxies (to avoid packet fragmentation) while still honoring custom socket options.
- Added regression test .

## Test Plan

All 396 tests in , , and  pass.

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -6,6 +6,7 @@ import queue
 import sys
 import typing
 import warnings
+import socket
 import weakref
 from socket import timeout as SocketTimeout
 from types import TracebackType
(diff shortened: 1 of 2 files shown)


illia-v:
Closing this as the issue has been addressed in #5054.

We have updated our documentation to clarify that the `socket_options` parameter should be used instead of modifying `HTTPConnection.default_socket_options` directly. Wrapper libraries built on top of urllib3 generally expose a way to pass these options down to the connection level.

PR #5005, opened 2026-05-12
Fix queue class resolution with monkey-patched queue modules

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent

360john360:
This fixes #3289 by resolving the default connection-pool queue class at pool construction time instead of always using the `QueueCls` reference captured when `urllib3.connectionpool` is imported.

Explicit custom `QueueCls` subclasses are still honored, but the default path now uses the current `queue.LifoQueue`, which allows gevent-style monkey-patching to work as expected.

Tested with:

- `python -m pytest test/test_queue_monkeypatch.py test/test_connectionpool.py::TestConnectionPool::test_pool_size test/test_connectionpool.py::TestConnectionPool::test_put_conn_when_pool_is_full_nonblocking test/test_connectionpool.py::TestConnectionPool::test_put_conn_when_pool_is_full_blocking`
- `python -m pytest test/test_connectionpool.py test/test_queue_monkeypatch.py`
- `python -m black --check --target-version py310 src/urllib3/connectionpool.py test/test_queue_monkeypatch.py`
- `python -m mypy --ignore-missing-imports --no-warn-unused-ignores test/test_queue_monkeypatch.py`

### DIFF
--- a/changelog/3289.bugfix.rst
+++ b/changelog/3289.bugfix.rst
@@ -1,0 +1,1 @@
+Resolved connection pool queue classes at construction time so monkey-patched queue implementations are honored.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -61,6 +61,7 @@ if typing.TYPE_CHECKING:
 log = logging.getLogger(__name__)
 
 _TYPE_TIMEOUT = typing.Union[Timeout, float, _TYPE_DEFAULT, None]
+_DEFAULT_QUEUE_CLASS = queue.LifoQueue
 
 
 # Pool objects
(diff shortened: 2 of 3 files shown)


360john360:
CI is mostly green now. The remaining failure is Windows 3.13 in `test_ssl_custom_validation_failure_terminates` (`The socket was not terminated` after 0.5s), which looks unrelated to this change. The new queue test and neighboring connection pool tests passed in that job, and lint/downstream are green.

I tried rerunning the failed jobs, but I do not have permission to do that from the fork.

illia-v:
Closing this because #3729 showed a better solution

PR #5006, opened 2026-05-12
Support bytes values in HTTPHeaderDict

### LINKED ISSUE #3072: HttpHeaderDict does not support bytes object

NguyenCong2k:
## Summary
- decode bytes header values with latin-1 before storing them in HTTPHeaderDict
- allow bytes values in mapping/iterable header sources
- add regression coverage for setitem, add, and constructor paths

Fixes #3072

## Testing
```powershell
python -m pytest test\test_collections.py -q
```

### DIFF
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -15,7 +15,7 @@ if typing.TYPE_CHECKING:
     class HasGettableStringKeys(Protocol):
         def keys(self) -> typing.Iterator[str]: ...
 
-        def __getitem__(self, key: str) -> str: ...
+        def __getitem__(self, key: str) -> str | bytes: ...
 
 
 __all__ = ["RecentlyUsedContainer", "HTTPHeaderDict"]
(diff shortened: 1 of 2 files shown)


NguyenCong2k:
Thanks for the review. I pushed an update addressing all three points:

- updated the casts in `ensure_can_construct_http_header_dict()` to match `str | bytes` values
- widened `__init__` and `extend` keyword value annotations to `str | bytes`
- renamed the constructor regression test for clarity

Tested with:

```powershell
python -m pytest test\test_collections.py -q

PR #5010, opened 2026-05-12
Add retry_after_max_strict to raise on excessive Retry-After

### LINKED ISSUE #1338: Retry.get_retry_after does't respect time-out

theluckystrike:
## Summary

When `retry_after_max_strict=True`, a `MaxRetryAfterWaitError` is raised instead of silently capping `Retry-After` values that exceed `retry_after_max`. This gives callers programmatic control over handling excessive server retry delays.

### Background

PR #3743 added `retry_after_max` with silent capping behavior. This is great for most use cases, but some callers need to **know** when a server requests an unreasonable wait — for example, to fail fast, alert, or switch to a different endpoint.

`retry_after_max_strict` is an opt-in boolean (default `False`) that raises `MaxRetryAfterWaitError` instead of capping. Zero backward compatibility risk since the default preserves existing behavior.

### Changes

| File | Change |
|---|---|
| `src/urllib3/exceptions.py` | Add `MaxRetryAfterWaitError(HTTPError)` with `retry_after` and `max_wait` attributes |
| `src/urllib3/util/retry.py` | Add `retry_after_max_strict: bool = False` to `__init__()`, `new()`, and `parse_retry_after()` |
| `test/test_retry.py` | 5 test cases: raises, within-limit passthrough, default caps, sleep raises, propagation via `new()` |
| `changelog/1338.feature.rst` | Changelog entry |

### Usage

```python
from urllib3.exceptions import MaxRetryAfterWaitError

# Raise if server asks for more than 60 seconds
retry = Retry(retry_after_max=60, retry_after_max_strict=True)

try:
    response = http.request("GET", url, retries=retry)
except MaxRetryAfterWaitError as e:
    print(f"Server wants {e.retry_after}s, our max is {e.max_wait}s")
```

Fixes #1338

## Test plan

- [x] All 67 tests pass in `test/test_retry.py` (62 existing + 5 new)
- [ ] CI passes (format, lint, full test matrix)

### DIFF
--- a/changelog/1338.feature.rst
+++ b/changelog/1338.feature.rst
@@ -1,0 +1,1 @@
+Added ``retry_after_max_strict`` parameter to :class:`~urllib3.util.retry.Retry`. When set to ``True``, a :exc:`~urllib3.exceptions.MaxRetryAfterWaitError` is raised instead of silently capping ``Retry-After`` values that exceed ``retry_after_max``.
--- a/src/urllib3/exceptions.py
+++ b/src/urllib3/exceptions.py
@@ -331,5 +331,23 @@ class HeaderParsingError(HTTPError):
         super().__init__(message)
 
 
+class MaxRetryAfterWaitError(HTTPError):
+    """Raised when a Retry-After header exceeds the configured maximum wait time
+    and ``retry_after_max_strict`` is enabled.
+
+    :param float retry_after: The Retry-After value from the server, in seconds.
+    :param int max_wait: The configured maximum wait time, in seconds.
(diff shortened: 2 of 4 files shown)


theluckystrike:
Hi! Just checking in — all CI checks are passing. Happy to make any adjustments if needed. Thanks for your time!

theluckystrike:
Friendly follow-up — all 37 CI checks remain green across Python 3.10–3.15, Ubuntu/macOS/Windows. Happy to address any feedback. cc @sethmlarson

PR #5013, opened 2026-05-13
Reject response headers with whitespace before colon

JoeyXi:
## Summary
- reject response headers whose field name contains trailing whitespace before the colon
- preserve the existing warning-only behavior for other header parsing defects
- add unit and socket-level regression coverage

Fixes #1362

## Tests
- python -m pytest test\test_util.py test\with_dummyserver\test_socketlevel.py -k header_parsing_or_related_cases

### DIFF
--- a/changelog/1362.bugfix.rst
+++ b/changelog/1362.bugfix.rst
@@ -1,0 +1,1 @@
+Rejected malformed response headers that included whitespace before the colon in a field name.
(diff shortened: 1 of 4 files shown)


JoeyXi:
Closing this PR to avoid duplicating the active work on #1362. After re-checking the current issue thread, #5017 is already linked as the PR expected to close the issue and more directly addresses the active obs-fold response-header value scope. This PR was focused on the whitespace-before-colon response-header-name path, so I do not want to keep it open as a competing or confusing approach. Thanks for maintaining urllib3.

JoeyXi:
Withdrawn in favor of the more relevant active work on #5017.

PR #5014, opened 2026-05-13
Add cleartext HTTP/2 h2c support

naosu616-debug:
Closes #3303.

## Summary

This adds explicit h2c support to the experimental HTTP/2 adapter.

Related: #5015 splits out the HTTP/2 socket reuse fix that this change depends
on. If that lands first, I can rebase this PR so the remaining diff stays
focused on h2c.

The existing `inject_into_urllib3()` behavior is unchanged: HTTPS pools still use
HTTP/2 and HTTP pools stay on HTTP/1.1 by default. Cleartext HTTP/2 has to be
enabled explicitly:

- `inject_into_urllib3(h2c=True)`
- `inject_into_urllib3(h2c="prior_knowledge")`
- `inject_into_urllib3(h2c="upgrade")`

The prior-knowledge path uses a cleartext `HTTPConnection` socket and starts the
HTTP/2 connection immediately. The upgrade path sends an HTTP/1.1
`Upgrade: h2c` request, waits for `101 Switching Protocols`, and then continues
the request as HTTP/2 on stream 1.

The cleartext connection also uses `:scheme` `http` and port `80` as the default
authority port.

Prior-knowledge h2c now opens the cleartext socket before sending HTTP/2 frames
through an `HTTPConnectionPool`. HTTP/2 connections also avoid the HTTP/1.1
dropped-connection probe, since a readable HTTP/2 socket can just have control
frames waiting and should still be reusable.

I also reset h2c upgrade state when an upgrade falls back to an HTTP/1.1
response or when the connection is closed, so a pooled connection does not carry
stale HTTP/2 state into a later request.

## Tests

Added coverage for:

- prior-knowledge cleartext requests;
- HTTP/1.1 upgrade request headers;
- switching to HTTP/2 after a `101` response;
- `HTTPConnectionPool` with `h2c=True` reusing one connection for two requests;
- `HTTPConnectionPool` with `h2c="upgrade"` reusing one connection for two requests;
- opt-in injection and restoration behavior;
- repeated injection and upgrade-state reset behavior.

I also added a changelog fragment.

## Local checks

- `uv run pytest test/test_http2_connection.py test/test_compatibility.py -q`
- `uv run pytest test/with_dummyserver/test_https.py -q -k "http2 or alpn or simple"`
- `uv run pytest test/with_dummyserver/test_connectionpool.py -q -k "http2"`
[...]

PR #5015, opened 2026-05-13
Keep HTTP/2 sockets reusable with pending control frames

### LINKED ISSUE #3301: Allow HTTP/2 socket to be re-used for future requests

naosu616-debug:
Closes #3301.

## Summary

HTTP/2 sockets can be readable even when the connection is still healthy. For
example, the peer may have sent a control frame after the response stream ended.
The HTTP/1.1 dropped-connection probe treats any readable socket as disconnected,
which can cause a pooled `HTTP2Connection` to be closed instead of reused.

This is also split out from #5014 so the socket reuse behavior can be reviewed
independently from h2c support.

This overrides `HTTP2Connection.is_connected` so the pool keeps an open HTTP/2
socket available for future requests. The existing request model is unchanged:
urllib3 still has one active request per pooled connection here, and this does
not add HTTP/2 multiplexing.

## Tests

Added a socketpair regression test that:

- sends a complete HTTP/2 response;
- leaves a pending HTTP/2 control frame on the socket;
- returns the connection to an `HTTPSConnectionPool`;
- checks that the same connection is reused without closing the socket;
- sends a second request on the reused HTTP/2 connection and reads the response.

## Local checks

- `uv run pytest test/test_http2_connection.py test/test_compatibility.py -q`
- `uv run pytest test/with_dummyserver/test_https.py -q -k "http2"`
- `uv run pytest test/with_dummyserver/test_connectionpool.py -q -k "http2"`
- `uv run --group mypy mypy src/urllib3/http2/connection.py test/test_http2_connection.py test/test_compatibility.py`
- `uv run --with pre-commit pre-commit run --files changelog/3301.bugfix.rst src/urllib3/http2/connection.py test/test_http2_connection.py`
- `uv run towncrier check --compare-with origin/main`

### DIFF
--- a/changelog/3301.bugfix.rst
+++ b/changelog/3301.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed HTTP/2 connection reuse when control frames are waiting on the socket.
--- a/src/urllib3/http2/connection.py
+++ b/src/urllib3/http2/connection.py
@@ -109,6 +109,12 @@ class HTTP2Connection(HTTPSConnection):
             if data_to_send := conn.data_to_send():
                 self.sock.sendall(data_to_send)
 
+    @property
+    def is_connected(self) -> bool:
+        # Readable HTTP/2 sockets may only have control frames waiting, so the
+        # HTTP/1.1 dropped-connection probe is not reliable for HTTP/2.
+        return self.sock is not None
+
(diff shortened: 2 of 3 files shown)


illia-v:

PR #5016, opened 2026-05-13
Use seek and tell for file body length

### LINKED ISSUE #3689: File-like object: Implement seek() and tell() for length
Implement: `src/urllib3/util/request.py`
```
File-like object, TODO: use seek() and tell() for length?
```

Wheezily:
Fixes #3689.

This uses `tell()` and `seek()` to detect the remaining length for binary file-like request bodies when the current position can be restored safely. Text streams and non-seekable bodies still keep an unknown length, so they continue using chunked framing.

Validation:
- `python -m pytest test/test_util.py::TestUtil::test_body_to_chunks_uses_seek_tell_for_binary_file_length test/test_util.py::TestUtil::test_body_to_chunks_keeps_unknown_length_for_text_file test/test_util.py::TestUtil::test_body_to_chunks_keeps_unknown_length_without_seek_tell test/test_util.py::TestUtil::test_body_to_chunks_keeps_unknown_length_when_end_seek_fails test/with_dummyserver/test_socketlevel.py::TestContentFraming::test_chunked_not_specified test/with_dummyserver/test_socketlevel.py::TestContentFraming::test_chunked_specified`
- `python -m towncrier check --compare-with origin/main`
- `python -m compileall src/urllib3/util/request.py test/test_util.py test/with_dummyserver/test_socketlevel.py`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/changelog/3689.feature.rst
+++ b/changelog/3689.feature.rst
@@ -1,0 +1,1 @@
+Added ``Content-Length`` detection for seekable binary file-like request bodies.
--- a/src/urllib3/util/request.py
+++ b/src/urllib3/util/request.py
@@ -194,6 +194,44 @@ class ChunksAndContentLength(typing.NamedTuple):
     content_length: int | None
 
 
+def _get_body_file_position_length(body: typing.Any) -> int | None:
+    body_tell = getattr(body, "tell", None)
+    body_seek = getattr(body, "seek", None)
+
+    if body_tell is None or body_seek is None or isinstance(body, io.TextIOBase):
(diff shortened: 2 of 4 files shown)


Wheezily:
Force-pushed a follow-up to fix the CI issues from the previous run. The socket-level bytes-io test was still waiting for chunked framing, but seekable file-like bodies now use Content-Length, so I updated that expectation. I also added unit coverage for the helper edge cases that coverage reported. Current GitHub checks are green.

PR #5019, opened 2026-05-14
HTTPResponse: drain_conn in 64KiB chunks to avoid overloading memory

shazow:
Should be a safe/conservative change. Let me know if 128KiB doesn't feel right or if we'd like any additional testing for it.

One question I had: I noticed in a few places we're doing chunked reads, and have another hardcoded value here in the read1 implementation from 3+ years ago:

https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L1224-L1232

And another default value to 64 KiB here:
https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L1238-L1240

Any chance we want to normalize them to one constant?

### DIFF
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -562,7 +562,7 @@ class BaseHTTPResponse(io.IOBase):
         self._retries = retries
 
     def stream(
-        self, amt: int | None = 2**16, decode_content: bool | None = None
+        self, amt: int | None = _READ_CHUNK_SIZE, decode_content: bool | None = None
     ) -> typing.Iterator[bytes]:
         raise NotImplementedError()
 
@@ -1240,7 +1240,7 @@ class HTTPResponse(BaseHTTPResponse):
         return self._decoded_buffer.get(amt)
 
     def stream(
-        self, amt: int | None = 2**16, decode_content: bool | None = None
+        self, amt: int | None = _READ_CHUNK_SIZE, decode_content: bool | None = None
     ) -> typing.Generator[bytes]:
         """
         A generator wrapper for the read() method. A call will block until
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -50,8 +50,8 @@ if typing.TYPE_CHECKING:
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


sigmavirus24 (dismissed):

illia-v (commented):
Thanks @shazow!
Let's add a changelog entry about optimizing memory usage in `drain_conn` for large responses.

> One question I had: I noticed in a few places we're doing chunked reads, and have another hardcoded value here in the read1 implementation from 3+ years ago:
> 
> https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L1224-L1232

This does additional reads until at least some data can be decompressed to avoid returning an empty bytes, the amount should be as small as possible.

> And another default value to 64 KiB here:
> https://github.com/urllib3/urllib3/blob/main/src/urllib3/response.py#L1238-L1240

I think using the same 64 KiB constant for `stream` and `drain_conn` is a good idea 👍🏻

shazow:
@illia-v Thanks for the review! Updated stream amt default to use the new constant and added a changelog entry. Please take another look.

shazow:
Er wait, you asked for 64 KiB, I'll swap that, one moment.

shazow:
Ok should be good to go. :)

illia-v (dismissed):
LGTM

src/urllib3/response.py
```diff
@@ -50,6 +50,9 @@
 
 log = logging.getLogger(__name__)
 
+# Read in 64 KiB chunks
+_READ_CHUNK_SIZE = 1 << 16
```
pquentin: nit: I find the `**` syntax that we were using until now clearer.

```suggestion
_READ_CHUNK_SIZE = 2 ** 16
```

pquentin (dismissed):
Thanks! LGTM.

src/urllib3/response.py
shazow: I'm indifferent, I just picked the same syntax cpython uses: https://github.com/python/cpython/blob/main/Lib/http/client.py#L116

pquentin (approved):

illia-v (approved):

PR #5020, opened 2026-05-14
Unfold obsolete folded response header values

adityawrk:
Addresses the original obs-fold response-header case discussed in #1362.

Python's `http.client` preserves obsolete folded response header lines in
`HTTPMessage` values. This change replaces each RFC 9112
`OWS CRLF RWS` sequence with a single space before urllib3 or downstream
consumers interpret the value.

The normalization:

- keeps urllib3's `HTTPHeaderDict` and the original `HTTPMessage` synchronized,
  preventing cookie consumers such as Requests from re-emitting raw CRLF;
- preserves header names, ordering, duplicate fields, and parser state;
- uses a common-case fast path and a linear-time pattern for long field values.

The socket regression covers the original folded `Set-Cookie`, duplicate
cookies, every SP/HTAB combination around CRLF, mixed and repeated whitespace,
multiple folds, and ordinary headers. Focused unit coverage also verifies
adjacent folds and preservation of the original message's order, duplicates,
payload, policy, and defects.

This intentionally does not address the separate whitespace-before-colon/header
name validation scope discussed on #1362.

Validation performed:

- full local suite: 2,279 passed, 333 skipped, 4 xfailed;
- focused strict nox sessions on Python 3.12, 3.13, and 3.14;
- pyupgrade, Black, isort, and flake8;
- towncrier changelog validation;
- source distribution and wheel builds;
- end-to-end Requests cookie extraction and subsequent request serialization.

### DIFF
--- a/changelog/1362.bugfix.rst
+++ b/changelog/1362.bugfix.rst
@@ -1,0 +1,1 @@
+Unfolded obsolete folded response header values before exposing them through response headers.
--- a/changelog/1362.bugfix.rst
+++ b/changelog/1362.bugfix.rst
@@ -1,1 +1,3 @@
-Unfolded obsolete folded response header values before exposing them through response headers.
+Fixed response header handling to replace obsolete folded header lines
+(`obs-fold`) with spaces in accordance with RFC 9112, preventing raw
+CRLF sequences from appearing in header values such as ``Set-Cookie``.
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -77,6 +77,11 @@ port_by_scheme = {"http": 80, "https": 443}
 RECENT_DATE = datetime.date(2025, 1, 1)
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
+_OBSOLETE_FOLD_RE = re.compile(r"\r\n[ \t]+")
+
+
+def _normalize_header_value(value: str) -> str:
+    return _OBSOLETE_FOLD_RE.sub(" ", value)
 
 
 class HTTPConnection(_HTTPConnection):
@@ -616,7 +621,10 @@ class HTTPConnection(_HTTPConnection):
                 exc_info=True,
             )
 
-        headers = HTTPHeaderDict(httplib_response.msg.items())
+        headers = HTTPHeaderDict(
+            (name, _normalize_header_value(value))
+            for name, value in httplib_response.msg.items()
+        )
 
         response = HTTPResponse(
             body=httplib_response,
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -77,11 +77,37 @@ port_by_scheme = {"http": 80, "https": 443}
 RECENT_DATE = datetime.date(2025, 1, 1)
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
-_OBSOLETE_FOLD_RE = re.compile(r"\r\n[ \t]+")
+# Starting the optional OWS match at the beginning of a whitespace run avoids
+# quadratic backtracking for long header values.
+_OBSOLETE_FOLD_RE = re.compile(r"(?:(?<![ \t])[ \t]+)?\r\n[ \t]+")
 
 
 def _normalize_header_value(value: str) -> str:
+    if "\r\n" not in value:
+        return value
     return _OBSOLETE_FOLD_RE.sub(" ", value)
+
+
+def _normalize_header_values(
+    message: http.client.HTTPMessage,
+) -> list[tuple[str, str]]:
+    header_items = message.items()
+    headers_changed = False
+    for index, (name, value) in enumerate(header_items):
+        normalized_value = _normalize_header_value(value)
+        if normalized_value != value:
+            header_items[index] = (name, normalized_value)
+            headers_changed = True
+
+    if headers_changed:
+        # Keep the original message in sync for downstream consumers such as
+        # cookie jars while preserving its identity and header ordering.
+        for name, _ in header_items:
+            del message[name]
+        for name, value in header_items:
+            message[name] = value
+
+    return header_items
 
 
 class HTTPConnection(_HTTPConnection):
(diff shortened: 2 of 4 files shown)


adityawrk:
Small scope clarification after re-reading #1362 and the newer related PRs:

This PR is intentionally limited to the original obs-folded response-header value case from the issue body. It does not attempt to fix the later/current whitespace-before-colon header-name scope, which #5013 is already addressing.

I also noticed #5017 covers a very similar obs-fold normalization path and was opened earlier. I am happy to close this in favor of #5017, or keep it open only if maintainers prefer this smaller diff/test shape.

adityawrk:
Thanks @sethmlarson, I appreciate you reviewing the competing approaches and keeping this PR open.

src/urllib3/connection.py
```diff
@@ -77,6 +77,11 @@ class BaseSSLError(BaseException):  # type: ignore[no-redef]
 RECENT_DATE = datetime.date(2025, 1, 1)
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
+_OBSOLETE_FOLD_RE = re.compile(r"\r\n[ \t]+")
```
illia-v: Unlike obsoleted RFC 7230, RFC 9112 defines `obs-fold = OWS CRLF RWS`
```suggestion
_OBSOLETE_FOLD_RE = re.compile(r"[ \t]*\r\n[ \t]+")
```

test/with_dummyserver/test_socketlevel.py
```diff
@@ -88,6 +88,32 @@ def multicookie_response_handler(listener: socket.socket) -> None:
...
+            sock = listener.accept()[0]
+
+            buf = b""
+            while not buf.endswith(b"\r\n\r\n"):
+                buf += sock.recv(65536)
+
+            sock.send(
+                b"HTTP/1.1 200 OK\r\n"
+                b"Set-Cookie: ___utmvbtouVBFmB=gZg\r\n"
+                b"    XbNOjalT: Lte; path=/; Max-Age=900\r\n"
```
illia-v: Please add other headers to the test to check for all combinations of tabs and spaces before and after CRLF

src/urllib3/connection.py
```diff
@@ -584,7 +589,10 @@ def getresponse(  # type: ignore[override]
                 exc_info=True,
             )
 
-        headers = HTTPHeaderDict(httplib_response.msg.items())
+        headers = HTTPHeaderDict(
+            (name, _normalize_header_value(value))
```
illia-v: This path runs for every response, so I think it's worth keeping the common case as cheap as possible.

Could we add a fast path before doing the regex substitution, for example by checking whether `\r` is present in the value first? That would avoid the extra regex work for normal header values while still handling obs-fold when needed.

changelog/1362.bugfix.rst
```diff
@@ -0,0 +1 @@
+Unfolded obsolete folded response header values before exposing them through response headers.
```
illia-v:
```suggestion
Fixed response header handling to replace obsolete folded header lines
(`obs-fold`) with spaces in accordance with RFC 9112, preventing raw
CRLF sequences from appearing in header values such as ``Set-Cookie``.
```

src/urllib3/connection.py
adityawrk: Updated to consume RFC 9112 OWS before CRLF as well as RWS after it. I used the output-equivalent linear-time expression (?:(?<![ \t])[ \t]+)?\r\n[ \t]+ instead of the literal leading-star form: the latter repeatedly rescans long whitespace runs and took about 2.8 seconds on a 65 KB adversarial field locally, versus about 1.4 ms for this form. Exhaustive small-input and randomized equivalence checks passed, including adjacent folds, and the new tests cover mixed and repeated OWS/RWS.

test/with_dummyserver/test_socketlevel.py
adityawrk: Expanded the socket regression in one response to cover SP/SP, SP/HTAB, HTAB/SP, and HTAB/HTAB around CRLF, plus mixed whitespace, repeated and adjacent folds, duplicate Set-Cookie fields, and an ordinary unchanged header. The assertions check exact one-space results in both urllib3 headers and the original HTTPMessage, including duplicate order.

src/urllib3/connection.py
adityawrk: Added the fast path before regex substitution. It checks for CRLF specifically, so normal headers and bare-CR-only values return immediately. The fold expression is also linear-time for long whitespace runs, avoiding a worst case introduced by the literal leading-star pattern.

changelog/1362.bugfix.rst
adityawrk: Applied the suggested RFC 9112-focused changelog wording verbatim.

adityawrk:
Pushed `00162b20`, rebased onto current `main`, and addressed all four review
threads.

Two additional issues surfaced during end-to-end and adversarial validation:

1. Normalizing only `HTTPHeaderDict` left the original #1362 Requests
   reproduction intact because Requests extracts cookies from
   `response._original_response.msg`. The implementation now updates that same
   `HTTPMessage` in place only when a fold is found, preserving its identity,
   parser state, global header order, and duplicates.
2. The literal `[ \t]*\r\n[ \t]+` suggestion has quadratic behavior on long
   whitespace runs. The committed expression is substitution-equivalent but
   linear-time; exhaustive small-input and randomized equivalence checks include
   adjacent and repeated folds.

Validation completed locally:

- full suite: 2,279 passed, 333 skipped, 4 xfailed;
- strict focused nox runs on Python 3.12, 3.13, and 3.14;
- real Requests session: normalized cookie extraction and clean subsequent
  `Cookie` request header;
- pyupgrade, Black, isort, flake8, towncrier validation, and package builds;
- three independent final reviews with no remaining actionable findings.

The refreshed GitHub CI matrix is now running. Once it is green, these threads
should be ready for re-review.

adityawrk:
@illia-v All requested updates are now in `00162b20`, the four addressed review threads are resolved, and all 38 checks are green (including downstream Requests and botocore). When you have a chance, could you please re-review? Thank you.

illia-v (approved):

illia-v:
@adityawrk please confirm if https://opencollective.com/urllib3/expenses/340721 was created by you

adityawrk:
@illia-v - Yes, it was created by me

adityawrk:
@illia-v - Thanks for your generosity. Excited to contribute more to urllib. :)

PR #5029, opened 2026-05-17
Make `is_ipaddress` detect non-standard forms of IPv4 addresses

illia-v:
@christos-spearbit highlighted that urllib3 can connect to hosts represented as non-standard IPv4 addresses via `socket.connect` (e.g., `urllib3.request("GET", "http://2130706433:8000/")`), but our `is_ipaddress` function does not detect such forms so some logic of the library that relies on `is_ipaddress` may be missed.

I update the `_IPV4_RE` regex to match the forms too.

FYI, non-standard IPv4 forms are not allowed inside IPv4-mapped IPv6 addresses like `::ffff:192.168.1.1`, so urllib3's logic for them is aligned with RFCs.

### DIFF
--- a/changelog/5029.bugfix.rst
+++ b/changelog/5029.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed ``is_ipaddress()`` to detect non-standard IPv4 forms accepted by
+``socket.connect``, such as hex (``0x7f000001``), octal (``0177.0.0.1``), and
+decimal integers (``2130706433``), ensuring SSL certificate verification uses
+the correct mode for these addresses.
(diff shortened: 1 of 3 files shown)


shazow:
+0

Is it desirable to pass through ambiguous formats straight to socket.connect?

I don't feel strongly here, but just wanted to note that the stdlib [`ipaddress.ip_address(address)`](https://docs.python.org/3/library/ipaddress.html) also rejects these kinds of addresses.

Hypothetically, I can imagine some years from now a CVE where something is relying on filtering out some specific addresses is getting bypassed by rewriting those addresses with non-standard variants. 🤪

shazow (approved):

illia-v:
> Is it desirable to pass through ambiguous formats straight to socket.connect?

I just don't feel like we need to make a breaking change - disallow such forms of IPv4. But if we keep support for them, the function has to aligned.

> a CVE where something is relying on filtering out some specific addresses is getting bypassed by rewriting those addresses with non-standard variants. 🤪

My next change will be related to this 😄. I want to update the security policy to mention that vulnerability reports about SSRF will not be accepted because urllib3 does not provide the functionality.
I know that @sethmlarson usually suggests relying on allow lists instead of block lists as a best practice.

pquentin (approved):
Thanks! LGTM.

PR #5034, opened 2026-05-19
Reject invalid response header folding

aigrouphobinh:

PR #5036, opened 2026-05-19
Reject folded response headers

nexicturbo:
### Summary
- reject HTTP responses whose parsed header values still contain folded CR/LF continuations
- add a response-header helper plus unit coverage
- add socket-level coverage for a folded `Set-Cookie` response header
- keep the fix narrow: invalid folded response headers are rejected instead of silently sanitizing or accepting them

### Testing
- `uv run --group test pytest test/test_util.py -k 'assert_header_parsing or assert_no_response_header_folding'`
- `uv run --group test pytest test/with_dummyserver/test_socketlevel.py -k 'folded_response_header_is_rejected or BrokenHeaders'`
- `uv run --group test pytest test/test_util.py test/with_dummyserver/test_socketlevel.py`
- `uv run --group mypy mypy src/urllib3/util/response.py src/urllib3/connection.py`
- GitHub CI is now fully green on the PR: package, changelog, CodeQL, lint, downstream requests/botocore, docs, coverage, and the full Python/platform matrix all report success.

/claim #5034

### DIFF
--- a/changelog/5034.bugfix.rst
+++ b/changelog/5034.bugfix.rst
@@ -1,0 +1,1 @@
+Rejected HTTP responses containing folded header values.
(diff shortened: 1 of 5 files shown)


illia-v:
Fixed by merged #5020

PR #5046, opened 2026-05-21
Fix empty allowed_methods retry handling

### LINKED ISSUE #5044: Empty allowed_methods container retries on all methods
### Subject

```python
```

### Environment

```
Python 3.13.13
OpenSSL 3.5.6 7 Apr 2026
urllib3 0.1.dev4378
```

### Steps to Reproduce

```python
```

### Expected Behavior

1. Warn on empty container
2. Error on empty container

```diff
    def _is_method_retryable(self, method: str) -> bool:
        """Checks if a given HTTP method should be retried upon, depending if
        it is included in the allowed_methods
        """
-       if self.allowed_methods and method.upper() not in self.allowed_methods:
+       if (
+           self.allowed_methods is not None
+           and method.upper() not in self.allowed_methods
+       ):
            return False
        return True
```

or simply:

```diff
    def _is_method_retryable(self, method: str) -> bool:
        """Checks if a given HTTP method should be retried upon, depending if
        it is included in the allowed_methods
        """
-       if self.allowed_methods and method.upper() not in self.allowed_methods:
-           return False
-       return True
+       return self.allowed_methods is None or method in self.allowed_methods
```

```python

            warnings.warn(
                UserWarning,
                stacklevel=2,
            )
```

### Actual Behavior

puneetdixit200:
## Summary
- distinguish `allowed_methods=None` from an explicitly empty container
- make empty `allowed_methods` retry no methods instead of all methods
- add regression coverage and a changelog entry

Fixes #5044

## Tests
- `python -m pytest test/test_retry.py::TestRetry::test_allowed_methods_with_status_forcelist`
- `python -m pytest test/test_retry.py`
- `python -m ruff format --check src/urllib3/util/retry.py test/test_retry.py`
- `python -m ruff check src/urllib3/util/retry.py test/test_retry.py`
- `git diff --check`

### DIFF
--- a/changelog/5044.bugfix.rst
+++ b/changelog/5044.bugfix.rst
@@ -1,0 +1,1 @@
+Treated an empty ``Retry.allowed_methods`` container as retrying no methods instead of all methods.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -404,7 +404,10 @@ class Retry:
         """Checks if a given HTTP method should be retried upon, depending if
         it is included in the allowed_methods
         """
-        if self.allowed_methods and method.upper() not in self.allowed_methods:
+        if (
+            self.allowed_methods is not None
+            and method.upper() not in self.allowed_methods
+        ):
             return False
         return True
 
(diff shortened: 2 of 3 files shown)


textbook:
Personally I consider it bad form to raise a PR for changes someone else suggested, particularly without waiting for community consensus on whether that's the right and complete solution.

PR #5048, opened 2026-05-22
Resolve connection pool queue class lazily

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent

Catnap7:
This PR resolves #3289 by deferring resolution of the default connection-pool queue class until a pool is constructed.

Previously `ConnectionPool.QueueCls` captured `queue.LifoQueue` at import time. If a runtime such as gevent patched the `queue` module after urllib3 was imported, the default pool still used the original class reference. The change keeps custom `QueueCls` subclasses working as before, but when the class is still the default it uses the current `queue.LifoQueue` at pool creation time.

Validation:
- `.venv\Scripts\python.exe -m pytest test/test_queue_monkeypatch.py`
- `.venv\Scripts\python.exe -m pytest test/test_connectionpool.py test/test_queue_monkeypatch.py`
- `.venv\Scripts\python.exe -m mypy src\urllib3\connectionpool.py`
- `git diff --check`

Fixes #3289

### DIFF
--- a/changelog/3289.bugfix.rst
+++ b/changelog/3289.bugfix.rst
@@ -1,0 +1,1 @@
+Resolved the default connection pool queue class at pool construction time so queue monkey patches applied after importing urllib3 can be honored while preserving custom ``QueueCls`` subclasses.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -76,7 +76,13 @@ class ConnectionPool:
     """
 
     scheme: str | None = None
-    QueueCls = queue.LifoQueue
+    QueueCls: type[queue.LifoQueue[typing.Any]] = queue.LifoQueue
+
+    def _get_queue_cls(self) -> type[queue.LifoQueue[typing.Any]]:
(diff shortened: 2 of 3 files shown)


Catnap7:
Note on the current red Ubuntu 3.14/3.15 CI jobs: both are failing in `TestProxyManager.test_tunnel_sets_http_11_alpn`, which appears unrelated to this pool queue change.

The captured TLS ClientHello contains ALPN `http/1.1`, but the test asserts `b"h2" not in self.buf` over the entire raw buffer; in both failures, `h2` appears in random handshake/key-share bytes rather than as the ALPN protocol. A previous run also hit the same failure on macOS 3.10 before passing on retry. I don't have permission to rerun the upstream jobs directly.

The local targeted validation for this PR passes:
- `.venv\Scripts\python.exe -m pytest test/test_connectionpool.py test/test_queue_monkeypatch.py`
- `.venv\Scripts\python.exe -m mypy src\urllib3\connectionpool.py test\test_queue_monkeypatch.py`

PR #5049, opened 2026-05-22
Treat empty retry allowed_methods as no methods

### LINKED ISSUE #5044: Empty allowed_methods container retries on all methods

puneetdixit200:
## Summary
- distinguish `allowed_methods=None` from an explicitly empty collection
- make an empty `allowed_methods` collection retry no methods
- add regression coverage for status retries with `allowed_methods=set()`

## Verification
- `python -m pytest test/test_retry.py::TestRetry::test_allowed_methods_with_status_forcelist -q` (failed before implementation)
- `python -m pytest test/test_retry.py -q`
- `python -m mypy src/urllib3/util/retry.py`
- `python -m compileall -q src test/test_retry.py`
- `git diff --check`
- `python -m pytest test -q` reached 1977 passed / 304 skipped / 4 xfailed, with one unrelated `TestHTTPSProxyVerification::test_https_proxy_ipv4_san` unclosed-socket ResourceWarning on Python 3.14.5; rerunning that test alone passed.

Fixes #5044.

AI-assisted: yes. I used AI assistance to prepare this patch and reviewed/tested the resulting code locally.

### DIFF
--- a/changelog/5044.bugfix.rst
+++ b/changelog/5044.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed ``Retry(allowed_methods=...)`` to distinguish ``None`` from an empty
+collection, so an explicitly empty collection retries no request methods.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -404,9 +404,9 @@ class Retry:
         """Checks if a given HTTP method should be retried upon, depending if
         it is included in the allowed_methods
         """
-        if self.allowed_methods and method.upper() not in self.allowed_methods:
-            return False
-        return True
+        if self.allowed_methods is None:
+            return True
+        return method.upper() in self.allowed_methods
 
     def is_retry(
         self, method: str, status_code: int, has_retry_after: bool = False
--- a/test/test_retry.py
+++ b/test/test_retry.py
@@ -263,10 +263,14 @@ class TestRetry:
         assert not retry.is_retry("GET", status_code=418)
 
     def test_allowed_methods_with_status_forcelist(self) -> None:
-        # Falsey allowed_methods means to retry on any method.
+        # None allowed_methods means to retry on any method.
         retry = Retry(status_forcelist=[500], allowed_methods=None)
         assert retry.is_retry("GET", status_code=500)
         assert retry.is_retry("POST", status_code=500)
+
+        retry = Retry(status_forcelist=[500], allowed_methods=set())
+        assert not retry.is_retry("GET", status_code=500)
+        assert not retry.is_retry("POST", status_code=500)
 
         # Criteria of allowed_methods and status_forcelist are ANDed.
         retry = Retry(status_forcelist=[500], allowed_methods=["POST"])


puneetdixit200:
Pushed `7b012f4c` rebased onto current `main` (`e63ea111`). Local verification passed after the rebase:

- `uv run --frozen --group test python -m pytest test/test_retry.py::TestRetry::test_allowed_methods_with_status_forcelist -q` (1 passed)
- `uv run --frozen --group test python -m pytest test/test_retry.py -q` (62 passed)
- `uv run --frozen --group mypy mypy src/urllib3/util/retry.py`
- `uv run --frozen --group test python -m compileall -q src test/test_retry.py`
- `git diff --check upstream/main...HEAD`

The previous red jobs I sampled were failing before project tests during checkout/action setup, so this should give CI a fresh run on the current branch head.

puneetdixit200:
Rebased this PR onto current `main` in `7b012f4`.

The previous red checks were failing before the test code ran: one checkout path failed while fetching the PR merge ref, and lint failed while downloading the `setup-uv` action archive. After the rebase push, the new run is past those setup failures; `package`, `changelog entry`, `lint`, docs, and `Ubuntu 3.13 test_min_pyopenssl` are already passing, with the rest of the matrix still running.

Local verification before pushing:
- `nox --reuse-existing-virtualenvs --sessions test-3.13 -- test/test_retry.py::TestRetry::test_allowed_methods_with_status_forcelist` - passed
- `nox --reuse-existing-virtualenvs --sessions lint` - passed
- `git diff --check` - passed

puneetdixit200:
Follow-up on the fresh CI run: the old setup/checkout failures are gone after the rebase. Current sampled status is 35 successful checks, with coverage and botocore still running, and one macOS 3.10 failure in `test/with_dummyserver/test_connectionpool.py::TestConnectionPoolTimeouts::test_timeout_float`.

That failing test is outside this PR retry diff (`src/urllib3/util/retry.py`, `test/test_retry.py`, and the changelog). The log shows `ConnectionRefusedError` / `NewConnectionError` to a localhost dummyserver port. I tried to rerun failed jobs with `gh run rerun 26546829463 --failed`, but GitHub denied it because this account does not have admin rights on the repository.

puneetdixit200:
Follow-up on the new run after the rebase:

- The checkout/setup failures from the previous run are resolved.
- 36 checks are passing now, including `lint`, `package`, `changelog entry`, docs, coverage, Ubuntu matrix jobs, Windows jobs that completed, macOS 3.11/3.12/3.13/3.14/3.14t/3.15, and downstream `requests`.
- The only current failure is `macOS 3.10`, where the full test suite reached `1995 passed` and then failed `test/with_dummyserver/test_connectionpool.py::TestConnectionPoolTimeouts::test_timeout_float` with `HTTPConnection(host='localhost', port=49660): Failed to establish a new connection: [Errno 61] Connection refused`. That path is outside this PR's retry/changelog/test_retry diff, so this looks like an unrelated dummy-server/runner flake.
- I tried to rerun failed jobs, but GitHub rejected it because rerun requires repository admin rights.

puneetdixit200:
Refreshed the branch to current head `7b012f4c2460f73c1a106b7bf3fb7ea966f7a53b` and rechecked the current CI.

Local verification:

```text
.nox/test-3-13/bin/python -m pytest test/test_retry.py -q
# 62 passed

git diff --check upstream/main...HEAD
# passed
```

The previous all-red checkout failures are gone. At the time of this check, all completed current-head jobs are passing except `macOS 3.10`, which failed in an unrelated dummyserver timeout test:

```text
test/with_dummyserver/test_connectionpool.py::TestConnectionPoolTimeouts::test_timeout_float
ConnectionRefusedError: [Errno 61] Connection refused
```

The PR diff is limited to `changelog/5044.bugfix.rst`, `src/urllib3/util/retry.py`, and `test/test_retry.py`, so I do not see a branch-scoped code change for that macOS dummyserver failure. `gh run rerun 26546829463 --job 78200380506` reports that the job cannot be rerun from my permissions. Could someone with repo permissions rerun the `macOS 3.10` job if needed? `downstream (botocore)` was still in progress when I checked.

illia-v:
Please check https://github.com/urllib3/urllib3/issues/5044#issuecomment-5552655479, this PR breaks compatibility.

PR #5055, opened 2026-05-24
Add auth decoding properties for URL class

### LINKED ISSUE #4945: parse_url doesn't %-decode user/password

illia-v:
Resolves #4945


opello:
It seems like this is going a long way to avoid creating `Url.auth_username` and `Url.auth_password` and/or `Url.auth_username_decoded` and `Url.auth_password_decoded` to adopt the unambiguous naming convention.  I'm curious what the use case is for a colon-joined, decoded pair?  It presents the opportunity for an ambiguous parse later if either value contains a colon.

[urllib.parse.urlsplit](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit) returns a named tuple that uses `username` and `password` keys that are specifically not unquoted (%-expanded).  I think that both unquoting and requiring the user to unquote are reasonable options so long as the documentation is clear.

illia-v:
@opello new properties can be used directly in the current functionality of the library where the credentials have to be decoded:
- `auth_decoded_joined` for arguments of `make_headers`
- `auth_decoded` for credentials of SOCKS proxies

> I'm curious what the use case is for a colon-joined, decoded pair? It presents the opportunity for an ambiguous parse later if either value contains a colon.

This has been documented in the docstring. The case with a colon inside a username is a known limitation of HTTP basic authorization.

opello:
I see, that makes sense for the internal library uses, especially when make_headers takes care of the base64 encoding.  But that strikes me as a fairly narrow use case.  I'm probably just wrong about my base assumptions here.

Thanks for the response.

src/urllib3/util/url.py
```diff
@@ -115,6 +120,41 @@ def __new__(  # type: ignore[no-untyped-def]
...
+        If ``auth`` is ``None``, this returns ``None``.
+
+        It is not checked if the username or password contains a colon,
+        so this property may not be suitable for all use cases.
+        """
+        username, password = self.auth_decoded
+        if username is None:
+            return None
+        if password is None:
+            return username
```
pquentin: Should we keep the colon when the password is empty? While the latest RFC isn't clear about this, https://stackoverflow.com/questions/14247637/can-the-password-in-http-basic-be-optional suggests that we should keep the colon in this case.

src/urllib3/util/url.py
```diff
@@ -115,6 +120,41 @@ def __new__(  # type: ignore[no-untyped-def]
...
+        User information with %-escapes decoded, as a string prepared
+        for encoding into an 'authorization: basic ...' header.
+
+        This is a convenience property that joins the username and
+        password with a colon, if both are present.
+        If only the username is present, it returns just the username.
+        If ``auth`` is ``None``, this returns ``None``.
+
+        It is not checked if the username or password contains a colon,
+        so this property may not be suitable for all use cases.
```
pquentin: I think colons are OK in passwords: https://datatracker.ietf.org/doc/html/rfc7617#page-5

src/urllib3/util/url.py
```diff
@@ -115,6 +120,41 @@ def __new__(  # type: ignore[no-untyped-def]
...
+        """
+        username, password = self.auth_decoded
+        if username is None:
+            return None
+        if password is None:
+            return username
+        return f"{username}:{password}"
```
pquentin: Note that `unquote()` used in `auth_decoded` [defaults to UTF-8](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.unquote), while [urllib3 enforces Latin-1 for Basic Authorization headers](https://github.com/urllib3/urllib3/blob/12ef275e392164e2ae13ea92c9b873da3cbdcdd4/src/urllib3/util/request.py#L126-L129) in `make_headers`. The spec for Basic Authorization only requires an ASCII-compatible encoding, so both Latin-1 and UTF-8 work.

I don't expect this to be a big issue in practice, since using non-ASCII characters in a password invites trouble. Still, I'm wondering if we should call `unquote(encoding="Latin-1")` for consistency. The issue is that users with Unicode passwords would be stuck, while the current solution can work if they build headers themselves.

pquentin (commented):
Thanks! I have a few questions around edge cases.

src/urllib3/util/url.py
illia-v: That's a very good point!

[RFC 3986 2.5. Identifying Data](https://datatracker.ietf.org/doc/html/rfc3986#section-2.5) suggests that UTF-8 should be expected in percent-encoded octets:
> When a new URI scheme defines a component that represents textual data consisting of characters from the Universal Character Set [UCS], the data should first be encoded as octets according to the UTF-8 character encoding [STD63]; then only those octets that do not correspond to characters in the unreserved set should be percent-encoded.  For example, the character A would be represented as "A", the character LATIN CAPITAL LETTER A WITH GRAVE would be represented as "%C3%80", and the character KATAKANA LETTER A would be represented as "%E3%82%A2".

But I can see the issue with the further `basic_auth.encode('latin-1')).decode()`.
Should we consider using UTF-8 for the basic authentication like [browsers do](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Authentication#character_encoding_of_http_authentication)?

illia-v: This paragraphs support the idea that UTF-8 can be used for the basic authentication. ~~And Latin-1 seems not to qualify because it is not a standard Unicode Encoding Form.~~

[RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617) "'Basic' HTTP Authentication Scheme"
> For the password, recipients MUST support all characters defined in the "OpaqueString" profile defined in [Section 4.2 of [RFC7613]](https://datatracker.ietf.org/doc/html/rfc7613#section-4.2).

[RFC 8265](https://datatracker.ietf.org/doc/html/rfc8265#section-4) "PRECIS: Usernames and Passwords" that obsoleted RFC 7613
> This document specifies that a password is a string of Unicode code points [Unicode] that is conformant to the OpaqueString profile (specified below) of the PRECIS FreeformClass defined in Section 4.3 of [RFC8264] and expressed in a standard Unicode Encoding Form (such as UTF-8 [RFC3629]).

illia-v: RFC 7617 does not prohibit ISO-8859-1

> In practice, most implementations chose either a locale-specific encoding such as ISO-8859-1 ([ISO-8859-1]), or UTF-8 ([RFC3629]).  For backwards compatibility reasons, this specification continues to leave the default encoding undefined, as long as it is compatible with US-ASCII (mapping any US-ASCII character to a single octet matching the US-ASCII character code).

src/urllib3/util/url.py
illia-v: Yes, let's keep the colon 👍🏻

src/urllib3/util/url.py
illia-v: This was a great point!

I believe we should still treat URL auth as UTF-8 and `make_headers` should accept a desired encoding while preserving `latin-1` as the default one for backward compatibility. Please check https://github.com/urllib3/urllib3/pull/5092 adding the encoding parameters to `make_headers`.

pquentin (approved):
Thank you for iterating! LGTM.

illia-v:
@pquentin thanks for the reviews!

PR #5057, opened 2026-05-24
Resolve default pool queue lazily

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent

ImAurum11:
This fixes #3289 by keeping `QueueCls` as the override point, but resolving the default queue class when a pool is created. If `queue.LifoQueue` has been monkey-patched after urllib3 was imported, new pools now pick up the patched class. Explicit `QueueCls` overrides still win.

I kept this narrower than #3729: no metaclass, no gevent-specific CI job. The regression test patches `queue.LifoQueue` directly, which is the behavior urllib3 needs to honor.

Verification
uv run --group test pytest test\test_connectionpool.py test\test_queue_monkeypatch.py
uv run --group mypy mypy src\urllib3\connectionpool.py test\test_connectionpool.py

Closes #3289.

### DIFF
--- a/changelog/3289.bugfix.rst
+++ b/changelog/3289.bugfix.rst
@@ -1,3 +1,5 @@
-Resolved the default connection pool queue class at pool creation time so
-monkey-patched ``queue.LifoQueue`` implementations can be used unless
-``QueueCls`` is explicitly overridden.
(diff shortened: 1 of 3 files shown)


sethmlarson (approved):
LGTM

PR #5060, opened 2026-05-25
Preserve invalid timeout errors in blocking pools

### LINKED ISSUE #5059: Confusing FullPoolError raised when passing a string timeout
### Subject

In general, urllib3 validates that timeouts are ints or floats, and raises a meaningful exception when this isn't the case. Example:

```python
r = urllib3.request("GET", "http://localhost:9200/", timeout="1")
```

Which raises:

```
ValueError: Timeout value connect was 1, but it must be an int, float or None.
```
But we have found one instance where instead we are seeing an obscure seemingly unrelated exception:

```python
pool = urllib3.HTTPConnectionPool('localhost', 9200, block=True)
pool.urlopen('GET', "/", timeout="1")
```

Which raises:

```
urllib3.exceptions.FullPoolError: HTTPConnectionPool(host='localhost', port=9200): Pool reached maximum size and no more connections are allowed.
```

### Environment

OS macOS-26.5-arm64-arm-64bit-Mach-O
Python 3.14.3
OpenSSL 3.5.5 27 Jan 2026
urllib3 2.6.3

### Steps to Reproduce

```python
import urllib3

pool = urllib3.HTTPConnectionPool('localhost', 9200, block=True)
pool.urlopen('GET', "/", timeout="1")
```

### Expected Behavior

A `ValueError` is raised.

### Actual Behavior

A `FullPoolError` is raised.

Mirochill:
## Summary

- validate per-request `timeout` before acquiring a pooled connection
- avoid returning a placeholder connection to a full blocking pool when timeout validation raises
- add a regression for string timeout values being reported as `ValueError` instead of `FullPoolError`
- add the `5059.bugfix.rst` changelog fragment

Fixes #5059.

## Validation

- Not run locally
- `git diff --check`
- `git diff --cached --check`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/changelog/5059.bugfix.rst
+++ b/changelog/5059.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed ``HTTPConnectionPool.urlopen()`` masking invalid per-request timeout
+values with ``FullPoolError`` in blocking pools.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -761,9 +761,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         # for future rewinds in the event of a redirect/retry.
         body_pos = set_file_position(body, body_pos)
 
+        # Validate per-request timeouts before acquiring a connection. If this
+        # raises, no connection has been taken from the pool and the cleanup
+        # block below must not try to return one.
+        timeout_obj = self._get_timeout(timeout)
+
         try:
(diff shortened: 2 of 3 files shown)


pquentin:
@Mirochill Sorry, we ended up adopting the approach in this PR (rejecting non-float values rather than converting strings to float), but I only noticed now, after merging mine.

(By the way, `git diff --check` is not validation)

PR #5070, opened 2026-05-31
Treat malformed folded response headers as parse defects

francosmit94-web:
## Summary

This adds regression coverage for #1362 and treats malformed folded response headers as parse defects when a folded continuation line appears to introduce a new header field.

The change is intentionally narrow: it only flags folded continuation lines that look like header field names, rather than treating all folded header values as invalid.

Fixes #1362.

## Verification

- TestBrokenHeaders: 4 passed
- Adjacent header parsing tests: 3 passed
- util response parsing tests: 4 passed
- socket-level suite excluding one unrelated SSL timing test: 116 passed, 9 skipped, 1 deselected

A full socket-level run previously produced an unrelated failure in `TestSSL.test_ssl_custom_validation_failure_terminates`; the new regression test and adjacent header parsing tests passed.

### DIFF
--- a/src/urllib3/util/response.py
+++ b/src/urllib3/util/response.py
@@ -1,9 +1,16 @@
 from __future__ import annotations
 
 import http.client as httplib
-from email.errors import MultipartInvariantViolationDefect, StartBoundaryNotFoundDefect
+import re
+from email.errors import (
+    MessageDefect,
+    MultipartInvariantViolationDefect,
+    StartBoundaryNotFoundDefect,
+)
 
 from ..exceptions import HeaderParsingError
+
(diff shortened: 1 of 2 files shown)


francosmit94-web:
This adds regression coverage for #1362 and treats malformed folded
  response headers as parse defects when a folded continuation line appears
  to introduce a new header field.

  The change is intentionally narrow: it only flags folded continuation
  lines that look like header field names, rather than treating all folded
  header values as invalid.

  Fixes #1362.

  ## Verification

  - TestBrokenHeaders: 4 passed
  - Adjacent header parsing tests: 3 passed
  - util response parsing tests: 4 passed
  - socket-level suite excluding one unrelated SSL timing test: 116 passed,
  9 skipped, 1 deselected

  A full socket-level run previously produced an unrelated failure in
  `TestSSL.test_ssl_custom_validation_failure_terminates`; the new
  regression test and adjacent header parsing tests passed.

PR #5071, opened 2026-06-01
Fix port-zero handling

illia-v:
@christos-spearbit noticed that previously, when a new request passed `port=0`, pool selection treated that value as if no port had been provided at all, so it fell back to the scheme default (80 for HTTP or 443 for HTTPS) and reused the existing default-port pool instead of creating or selecting a distinct port-zero pool.

This change preserves port 0 as a distinct one in URL parsing, pool selection, and proxy configuration.


sigmavirus24 (commented):
Why do we want to allow a pool for port 0? It's reserved and primarily reserved due to how POSIX works when trying to get the next available port number for binding a listener... So why do we want/need to handle this?

illia-v:
@sigmavirus24 that's a good question!

My primary goal is to ensure that if a user explicitly requests port 0, urllib3 does not implicitly fall back to port 80 or 443.

While browsers like Firefox and Chrome block requests to port 0, curl will attempt the connection (as detailed in [Daniel's blog post](https://daniel.haxx.se/blog/2014/10/25/pretending-port-zero-is-a-normal-one/)).

Personally, I think it's best to let the underlying operating system decide how to handle a port 0 request rather than blocking it at the library level. However, I'm indifferent to the exact behavior as long as we stop defaulting it to port 80/443.

sigmavirus24 (approved):

PR #5079, opened 2026-06-04
Strip fragments from absolute request targets

Ninja3047:
urllib3 could send URL fragments in absolute-form request targets for HTTPConnectionPool absolute URLs/absolute redirects and proxied HTTP requests, potentially leaking client-side fragment data to origin servers or proxies.

Relevant excerpts from the RFCs
> [3.2.2. ](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2.2)[absolute-form](https://datatracker.ietf.org/doc/html/rfc9112#name-absolute-form)
When making a request to a proxy, other than a CONNECT or server-wide OPTIONS request (as detailed below), a client MUST send the target URI in "absolute-form" as the request-target.
>  absolute-form  = absolute-URI

> [4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3).  Absolute URI
>   Some protocol elements allow only the absolute form of a URI without
>   a fragment identifier.  For example, defining a base URI for later
>   use by relative references calls for an absolute-URI syntax rule that
>   does not allow a fragment.
>   absolute-URI  = scheme ":" hier-part [ "?" query ]

### DIFF
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -622,3 +622,60 @@ class TestConnectionPool:
             # Verify the URL isn't mangled
             actual_url = mock_request.call_args[0][2]
             assert actual_url == path
+
+    @pytest.mark.parametrize(
+        ("target_url", "expected_url"),
+        [
+            pytest.param(
+                "http://localhost/path#marker=value",
+                "http://localhost/path",
+                id="fragment-after-path",
+            ),
+            pytest.param(
+                "http://localhost/path?x=1#marker=value",
+                "http://localhost/path?x=1",
+                id="fragment-after-query",
+            ),
+            pytest.param(
+                "http://localhost/#marker=value",
+                "http://localhost/",
+                id="root-path-with-fragment",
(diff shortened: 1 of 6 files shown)


test/test_connectionpool.py
```diff
@@ -622,3 +622,60 @@ def test_paths_arent_parsed_as_urls(self, path: str) -> None:
             # Verify the URL isn't mangled
             actual_url = mock_request.call_args[0][2]
             assert actual_url == path
+
+    @pytest.mark.parametrize(
+        ("target_url", "expected_url"),
+        [
+            pytest.param(
+                "http://localhost/path#access_token=secret",
```
sethmlarson: These test cases make it seem like this stripping is a security feature, but I don't think that's true. Can we change the test cases to not have a security angle to avoid confusing future maintainers?

illia-v (approved):

illia-v:
LGTM, thanks!

PR #5082, opened 2026-06-08
Reject CRLF and null in multipart field header values

netliomax25-code:
1. `make_multipart` writes `content_type`, `content_location` and the `content_disposition` prefix into the part headers without checking them, while `name` and `filename` are already neutralized by `format_multipart_header_param`.
2. A CR, LF or NUL in any of those three values splits the part-header block and injects extra headers into the request body. A common trigger is an attacker-supplied upload Content-Type forwarded through the `(filename, data, content_type)` tuple API.

Added a check in `make_multipart` that rejects those characters with `ValueError`, keeping the three structured values consistent with the existing name/filename handling.

Repro before the change:

    fields = [("file", ("a.txt", b"DATA", "text/plain\r\nX-Injected: 1"))]
    body, _ = encode_multipart_formdata(fields)
    # body contains: Content-Type: text/plain\r\nX-Injected: 1\r\n

### DIFF
--- a/changelog/5082.bugfix.rst
+++ b/changelog/5082.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``RequestField.make_multipart()`` to reject carriage return, line feed and null characters in the ``content_type``, ``content_location`` and ``content_disposition`` values, preventing header injection into multipart request bodies.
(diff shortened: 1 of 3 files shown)


netliomax25-code:
any update?

PR #5083, opened 2026-06-09
validate Content-Length and chunk-size before int conversion

netliomax25-code:
1. Content-Length is converted with `int()`, which also accepts `+5`, `1_000` and unicode digits that RFC 9112 section 6.3 (`1*DIGIT`) forbids.
2. chunk-size is converted with `int(line, 16)`, which also accepts a `0x` prefix, signs and underscores that RFC 9112 section 7.1 (`1*HEXDIG`) forbids.
3. Either case lets a server or intermediary frame a response that urllib3 reads with a different length than a stricter peer.

Validated both fields against a strict pattern in `_init_length` and `_update_chunk_length` before the conversion. Optional whitespace around comma-separated Content-Length values and chunk extensions still parse as before.

### DIFF
--- a/changelog/5083.bugfix.rst
+++ b/changelog/5083.bugfix.rst
@@ -1,0 +1,1 @@
+Tightened response framing parsing to reject ``Content-Length`` and chunked transfer ``chunk-size`` values that are not valid per RFC 9112 (for example ``Content-Length: 1_000`` or a ``0x``-prefixed chunk size), which ``int()`` would otherwise accept.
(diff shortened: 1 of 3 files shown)


netliomax25-code:
Pushed a fix for the coverage job. The new digit-only check already rejects a negative Content-Length, so the old `if length < 0` branch was unreachable and was holding coverage at 99%. Dropped it; the `-10` case still returns None through the ValueError path.

netliomax25-code:
any update?

PR #5085, opened 2026-06-10
reject control characters and spaces in URL host in parse_url

netliomax25-code:
1. `parse_url` percent-encodes control characters in the userinfo, path, query, and fragment components (`test_control_characters_are_percent_encoded` pins this), but the host was returned verbatim, so a host containing a control character or space, including CR and LF, passed through the parser unchanged. `_REG_NAME_PAT` only excludes `[ ] % : / ? #`, so the bytes `0x00-0x20` and `0x7f` survive.
2. Reject a host containing any of those characters with `LocationParseError` in `parse_url`, the single parsing chokepoint that `is_same_host` and the proxy/pool paths also go through. The character set matches `http.client._contains_disallowed_url_pchar_re`, keeping the parser at least as strict as the stdlib transport the host is handed to, and keeping host-based caller checks from receiving a malformed host.

Validation: `parse_url("http://exa\r\nmple.com/")` previously returned `host=exa
mple.com`; it now raises `LocationParseError`. Added `test_control_characters_in_host_are_rejected` over the same character range as the existing percent-encoding test (fails before the change, passes after). Existing url and pool tests still pass.


avinashkamat48 (commented):
The changelog fragment is named `5084.bugfix.rst`, but this PR is #5085 and #5084 is a different open PR. That can make the generated changelog attribution confusing and may collide if #5084 also needs a fragment. Please rename this to `5085.bugfix.rst` or the relevant issue number.

netliomax25-code:
Good catch. There's no separate issue for this one, and 5084 is actually an unrelated PR, so per the changelog README I renamed the fragment to 5085.bugfix.rst (the PR number). Pushed.

netliomax25-code:
gentle ping

illia-v:
Thanks for the contribution!

Coincidentally, more comprehensive changes addressing the issues raised here had already been developed privately during [[Patch the Planet](https://trailofbits.com/patch-the-planet/)](https://trailofbits.com/patch-the-planet/) before this PR was opened. They were merged in #5095.

PR #5089, opened 2026-06-20
Use lazy queue creation for default connection pools

### LINKED ISSUE #3289: Make custom queue class feature work well with gevent

shaiananvari8:
Fixes #3289.

## Summary
- Default `ConnectionPool.QueueCls` to `None` so the pool constructor resolves `queue.LifoQueue` lazily after any runtime patching.
- Add `_make_queue()` to keep the default queue path late-bound while preserving explicit custom `QueueCls` overrides.
- Add regression coverage for late-bound default queues and custom queue classes.
- Add the bugfix changelog fragment.

## Validation
- `.venv\Scripts\python.exe -m pytest test\test_connectionpool.py -k queue_class` - 2 passed
- `.venv\Scripts\python.exe -m pytest test\test_connectionpool.py` - 88 passed
- `git diff --check` - passed

The issue carries the $100 OpenCollective bounty label; happy to coordinate payout details after acceptance.

### DIFF
--- a/changelog/3289.bugfix.rst
+++ b/changelog/3289.bugfix.rst
@@ -1,0 +1,2 @@
+Used the monkey-patched ``queue.LifoQueue`` when creating default connection pools,
+improving compatibility with gevent while still respecting custom queue classes.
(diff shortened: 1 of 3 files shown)


illia-v:
Superseded by #3289

PR #5091, opened 2026-06-24
Port "Reject CR/LF in HTTP tunnel request headers"

illia-v:
This ports fixes from https://github.com/python/cpython/issues/146211 to our copies of the code for older Python versions.


illia-v:
I decided to enable the patch for Python 3.11.9-3.11.15 and 3.12.3-3.12.13 too as this costs us nothing, makes more users secure, and reduces complexity of conditions in the test file.

sigmavirus24 (approved):

PR #5092, opened 2026-06-26
Make Basic auth encoding explicit in `make_headers`

illia-v:
Based on #5055

Allows passing explicit encoding for `basic_auth` and `proxy_basic_auth`.
This lets callers explicitly choose UTF-8 for Basic authentication credentials when needed, while avoiding a breaking switch from `latin-1` by default.

Clean diff: https://github.com/illia-v/urllib3/compare/fix-4945...make_headers-encoding
Original discussion: https://github.com/urllib3/urllib3/pull/5055#r3440708947


pquentin (dismissed):
Thanks! LGTM.

Thank you for providing the clean diff.

PR #5093, opened 2026-06-26
HTTPSConnection: Prefer proxy verify policy for the proxy connection

### LINKED ISSUE #3472: Cannot use `proxy_ssl_context` with `CERT_VERIFIED` and `ssl_context` with `CERT_NONE`
### Subject

When specifying a `proxy_ssl_context` for an HTTPS proxy with `verify_mode=CERT_VERIFIED`, but the `ssl_context` has `verify_mode=CERT_NONE`, the proxy manager fails to create a connection. This is for a scenario where the proxy https connection should be verified but the proxied connection is not.

It appears that the `verify_mode` from the proxied connection `ssl_context` is leaking when wrapping the proxy connection with the `proxy_ssl_context` 
https://github.com/urllib3/urllib3/blob/cc42860721836febf3fb6ebb485ed27d7f80122d/src/urllib3/connection.py#L795-L822

In the above, `self.cert_reqs` is set to `CERT_NONE` from the `ssl_context` and `_ssl_wrap_socket_and_match_hostname` applies that over the `proxy_ssl_context`:
https://github.com/urllib3/urllib3/blob/cc42860721836febf3fb6ebb485ed27d7f80122d/src/urllib3/connection.py#L871

### Environment

Describe your environment.

```
OS Linux-6.10.5-1-default-x86_64-with-glibc2.40
Python 3.12.3
OpenSSL 3.1.4 24 Oct 2023
urllib3 2.2.2
```

### Steps to Reproduce

A simple and isolated way to reproduce the issue. A code snippet would be great.

I created this test in `test/with_dummyserver/test_proxy_poolmanager.py` `TestHTTPProxyManager` which reproduces the issue

```py
    def test_proxy_https_target_skip_verify(self) -> None:
        proxy_url = self.https_proxy_url
        proxy_ctx = ssl.create_default_context()
        proxy_ctx.load_verify_locations(DEFAULT_CA)
        ctx = ssl.create_default_context()
        ctx.check_hostname = False
        ctx.verify_mode = ssl.CERT_NONE

        with proxy_from_url(
            proxy_url,
            proxy_ssl_context=proxy_ctx,
            ssl_context=ctx,
        ) as proxy:
            r = proxy.request("GET", f"{self.https_url}/")
            assert r.status == 200
```

### Expected Behavior

What should happen.

Not fail. We want to be able to verify the certificate of the proxy connection, but skip verification of the proxied connection.

### Actual Behavior

What happens instead.
You may attach logs, packet captures, etc.

```
============================= test session starts ==============================
collecting ... collected 1 item

test/with_dummyserver/test_proxy_poolmanager.py::TestHTTPProxyManager::test_proxy_https_target_skip_verify 

========================= 1 failed, 1 warning in 0.35s =========================
FAILED [100%]
test/with_dummyserver/test_proxy_poolmanager.py:603 (TestHTTPProxyManager.test_proxy_https_target_skip_verify)
self = <test.with_dummyserver.test_proxy_poolmanager.TestHTTPProxyManager object at 0x7fe2861d6150>

    def test_proxy_https_target_skip_verify(self) -> None:
        proxy_url = self.https_proxy_url
        proxy_ctx = ssl.create_default_context()
        proxy_ctx.load_verify_locations(DEFAULT_CA)
        ctx = ssl.create_default_context()
        ctx.check_hostname = False
        ctx.verify_mode = ssl.CERT_NONE
    
        with proxy_from_url(
            proxy_url,
            proxy_ssl_context=proxy_ctx,
            ssl_context=ctx,
        ) as proxy:
>           r = proxy.request("GET", f"{self.https_url}/")

test/with_dummyserver/test_proxy_poolmanager.py:617: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
venv/lib/python3.12/site-packages/urllib3/_request_methods.py:136: in request
venv/lib/python3.12/site-packages/urllib3/_request_methods.py:183: in request_encode_url
venv/lib/python3.12/site-packages/urllib3/poolmanager.py:633: in urlopen
    return super().urlopen(method, url, redirect=redirect, **kw)
venv/lib/python3.12/site-packages/urllib3/poolmanager.py:443: in urlopen
venv/lib/python3.12/site-packages/urllib3/connectionpool.py:775: in urlopen
    self._prepare_proxy(conn)
venv/lib/python3.12/site-packages/urllib3/connectionpool.py:1044: in _prepare_proxy
    conn.connect()
venv/lib/python3.12/site-packages/urllib3/connection.py:624: in connect
    self.sock = sock = self._connect_tls_proxy(self.host, sock)
venv/lib/python3.12/site-packages/urllib3/connection.py:698: in _connect_tls_proxy
    sock_and_verified = _ssl_wrap_socket_and_match_hostname(
venv/lib/python3.12/site-packages/urllib3/connection.py:767: in _ssl_wrap_socket_and_match_hostname
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ssl.SSLContext object at 0x7fe285524550>
value = <VerifyMode.CERT_NONE: 0>

    @verify_mode.setter
    def verify_mode(self, value):
>       super(SSLContext, SSLContext).verify_mode.__set__(self, value)
E       ValueError: Cannot set verify_mode to CERT_NONE when check_hostname is enabled.

/usr/lib64/python3.12/ssl.py:680: ValueError
```

### LINKED ISSUE #4961: proxy_ssl_context is mutated when target uses cert_reqs='CERT_NONE'

shazow:
Fixed ``HTTPSConnection`` overriding the ``proxy_config``'s cert policy with the target request's SSL context.

Including a ``proxy_config`` will now properly inherit ``verify_mode`` for the proxy connection and values for ``ca_*`` ``ssl_*`` are ignored.

### DIFF
--- a/changelog/5093.bugfix.rst
+++ b/changelog/5093.bugfix.rst
@@ -1,2 +1,2 @@
 Fixed ``HTTPSConnection`` overriding the ``proxy_config``'s cert policy with the target request's SSL context.
-Including a ``proxy_config`` will now properly inherit ``verify_mode`` for the proxy connection and values for ``ca_certs``, ``ca_cert_dir``, and ``ca_cert_data`` will be ignored.
+Including a ``proxy_config`` will now properly inherit ``verify_mode`` for the proxy connection and values for ``ca_*`` ``ssl_*`` are ignored.
--- a/changelog/5093.bugfix.rst
+++ b/changelog/5093.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed ``HTTPSConnection`` overriding the ``proxy_config``'s cert policy with the target request's SSL context.
+Including a ``proxy_config`` will now properly inherit ``verify_mode`` for the proxy connection and values for ``ca_certs``, ``ca_cert_dir``, and ``ca_cert_data`` will be ignored.
--- a/changelog/5093.bugfix.rst
+++ b/changelog/5093.bugfix.rst
@@ -1,2 +1,11 @@
-Fixed ``HTTPSConnection`` overriding the ``proxy_config``'s cert policy with the target request's SSL context.
-Including a ``proxy_config`` will now properly inherit ``verify_mode`` for the proxy connection and values for ``ca_*`` ``ssl_*`` are ignored.
+Fixed ``HTTPSConnection.connect()`` overriding
+``ProxyConfig.ssl_context``'s certificate policy and proxy identity checks with
+the target connection's TLS settings when forwarding through an HTTPS proxy.
+``HTTPSConnection._connect_tls_proxy()`` now uses
+``ProxyConfig.ssl_context``, ``ProxyConfig.assert_hostname``, and
+``ProxyConfig.assert_fingerprint`` for the proxy handshake, while retaining the
+existing fallback to ``HTTPSConnection.ssl_context``, ``cert_reqs``, ``ca_*``,
+and ``ssl_*`` when ``ProxyConfig.ssl_context`` is ``None``.
+``HTTPSConnection`` no longer applies target SNI, assertions, or client
+credentials to forwarding proxy handshakes and continues to use its
+``ssl_context`` as a fallback when an HTTPS proxy forwards an HTTP target.
--- a/changelog/5093.bugfix.rst
+++ b/changelog/5093.bugfix.rst
@@ -1,11 +1,7 @@
 Fixed ``HTTPSConnection.connect()`` overriding
 ``ProxyConfig.ssl_context``'s certificate policy and proxy identity checks with
 the target connection's TLS settings when forwarding through an HTTPS proxy.
-``HTTPSConnection._connect_tls_proxy()`` now uses
-``ProxyConfig.ssl_context``, ``ProxyConfig.assert_hostname``, and
-``ProxyConfig.assert_fingerprint`` for the proxy handshake, while retaining the
-existing fallback to ``HTTPSConnection.ssl_context``, ``cert_reqs``, ``ca_*``,
-and ``ssl_*`` when ``ProxyConfig.ssl_context`` is ``None``.
+
 ``HTTPSConnection`` no longer applies target SNI, assertions, or client
 credentials to forwarding proxy handshakes and continues to use its
 ``ssl_context`` as a fallback when an HTTPS proxy forwards an HTTP target.
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -915,15 +915,29 @@ class HTTPSConnection(HTTPConnection):
         # `_connect_tls_proxy` is called when self._tunnel_host is truthy.
         proxy_config = typing.cast(ProxyConfig, self.proxy_config)
         ssl_context = proxy_config.ssl_context
+
+        if ssl_context is not None:
+            # Prefer the proxy's cert policy for the proxy connection
+            cert_reqs = ssl_context.verify_mode
+            ca_certs = None
+            ca_cert_dir = None
+            ca_cert_data = None
+        else:
+            # Otherwise we inherit the pool's cert policies
+            cert_reqs = self.cert_reqs
+            ca_certs = self.ca_certs
+            ca_cert_dir = self.ca_cert_dir
+            ca_cert_data = self.ca_cert_data
+
         sock_and_verified = _ssl_wrap_socket_and_match_hostname(
             sock,
-            cert_reqs=self.cert_reqs,
+            cert_reqs=cert_reqs,
             ssl_version=self.ssl_version,
             ssl_minimum_version=self.ssl_minimum_version,
             ssl_maximum_version=self.ssl_maximum_version,
-            ca_certs=self.ca_certs,
-            ca_cert_dir=self.ca_cert_dir,
-            ca_cert_data=self.ca_cert_data,
+            ca_certs=ca_certs,
+            ca_cert_dir=ca_cert_dir,
+            ca_cert_data=ca_cert_data,
             server_hostname=hostname,
             ssl_context=ssl_context,
             assert_hostname=proxy_config.assert_hostname,
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -838,14 +838,12 @@ class HTTPSConnection(HTTPConnection):
             # fall back to using the connection's SSL context until
             # urllib3 v3.0. Appropriate warning is emitted in
             # ``ProxyManager.__init__``.
+            wrapped_socket: ssl.SSLSocket | SSLTransport
             if self.proxy_is_forwarding and self.proxy_config is not None:
-                self.sock = sock = self._connect_tls_proxy(self.host, sock)
-                sock_and_verified = _WrappedAndVerifiedSocket(
-                    socket=sock,
-                    is_verified=self.proxy_is_verified is True,
-                )
+                wrapped_socket = self._connect_tls_proxy(self.host, sock)
+                is_verified = self.proxy_is_verified is True
             else:
-                sock_and_verified = _ssl_wrap_socket_and_match_hostname(
+                wrapped_socket, is_verified = _ssl_wrap_socket_and_match_hostname(
                     sock=sock,
                     cert_reqs=self.cert_reqs,
                     ssl_version=self.ssl_version,
@@ -863,7 +861,7 @@ class HTTPSConnection(HTTPConnection):
                     assert_hostname=self.assert_hostname,
                     assert_fingerprint=self.assert_fingerprint,
                 )
-            self.sock = sock_and_verified.socket
+            self.sock = wrapped_socket
 
         # If an error occurs during connection/handshake we may need to release
         # our lock so another connection can probe the origin.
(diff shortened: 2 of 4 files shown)


src/urllib3/connection.py
```diff
@@ -883,15 +883,29 @@ def _connect_tls_proxy(self, hostname: str, sock: socket.socket) -> ssl.SSLSocke
...
+            ca_cert_dir = self.ca_cert_dir
+            ca_cert_data = self.ca_cert_data
+
         sock_and_verified = _ssl_wrap_socket_and_match_hostname(
             sock,
-            cert_reqs=self.cert_reqs,
+            cert_reqs=cert_reqs,
             ssl_version=self.ssl_version,
             ssl_minimum_version=self.ssl_minimum_version,
             ssl_maximum_version=self.ssl_maximum_version,
```
illia-v: I can see these values are ignored by `_ssl_wrap_socket_and_match_hostname` unless it has to create a new context.
Let's set the values to `None` if `ssl_context is not None`. This should not affect functionality, but will bring more clarity to the method.

test/with_dummyserver/test_proxy_poolmanager.py
```diff
@@ -921,10 +921,11 @@ def test_https_proxy_assert_fingerprint_md5(
         destination_url = f"https://{server.host}:{server.port}"
 
         proxy_ctx = urllib3.util.ssl_.create_urllib3_context(verify_flags=0)
+        proxy_ctx.load_verify_locations(proxy.ca_certs)
         proxy_fingerprint = self._get_proxy_fingerprint_md5(proxy.ca_certs)
         with proxy_from_url(
             proxy_url,
-            ca_certs=proxy.ca_certs,
+            ca_certs=server.ca_certs,
```
shazow, line 971: I think these test fixes make sense but I'm not 100% sure, please double check?

Before we were overriding proxy certs with server target certs anyway, so it passed. Now we don't do that anymore, so we pass proxy policy via proxy_ctx and ca_certs becomes the target server cert policy.

Feels like this was a footgun that we're fixing, but might be breaking in some edge cases? Let me know if I should expand about this in the changelog.

src/urllib3/connection.py
shazow: Hmm I got a [flakey failure under macOS](https://github.com/urllib3/urllib3/actions/runs/28329505922/job/83925123632?pr=5093) that I thought was related (it was checking  to this but re-running passed.

```
FAILED test/with_dummyserver/test_socketlevel.py::TestProxyManager::test_tunnel_sets_http_11_alpn - assert b'h2' not in ...
```

shazow: This time Windows flaked on the same test. Is this a known flakey test? https://github.com/urllib3/urllib3/actions/runs/28329993127/job/83926409713?pr=5093

src/urllib3/connection.py
```diff
@@ -807,28 +807,30 @@ def connect(self) -> None:
             # urllib3 v3.0. Appropriate warning is emitted in
             # ``ProxyManager.__init__``.
             if self.proxy_is_forwarding and self.proxy_config is not None:
-                ssl_context = self.proxy_config.ssl_context
+                self.sock = sock = self._connect_tls_proxy(self.host, sock)
+                sock_and_verified = _WrappedAndVerifiedSocket(
+                    socket=sock,
+                    is_verified=self.proxy_is_verified is True,
+                )
```
illia-v: Should we make the private `self._connect_tls_proxy` return `sock_and_verified` instead of `sock_and_verified.socket` to avoid creating `_WrappedAndVerifiedSocket` again here?

Alternatively, we can make `sock_and_verified` unneeded in this method by saving local variables and using them instead of `sock_and_verified.is_verified` and `sock_and_verified.socket`.

changelog/5093.bugfix.rst
```diff
@@ -0,0 +1,11 @@
+Fixed ``HTTPSConnection.connect()`` overriding
+``ProxyConfig.ssl_context``'s certificate policy and proxy identity checks with
+the target connection's TLS settings when forwarding through an HTTPS proxy.
+``HTTPSConnection._connect_tls_proxy()`` now uses
+``ProxyConfig.ssl_context``, ``ProxyConfig.assert_hostname``, and
+``ProxyConfig.assert_fingerprint`` for the proxy handshake, while retaining the
+existing fallback to ``HTTPSConnection.ssl_context``, ``cert_reqs``, ``ca_*``,
+and ``ssl_*`` when ``ProxyConfig.ssl_context`` is ``None``.
```
illia-v: Let's avoid documenting details of the private method?

illia-v (commented):
LGTM overall 👍🏻

DPS0340:
Drive-by: the `lint` failure here is two `mypy` errors, and both are one-line annotations. Posting the diff rather than a suggestion box since they are in two different files.

```
src/urllib3/connection.py:901: error: Incompatible types in assignment
    (expression has type "SSLContext | None", variable has type "SSLContext")  [assignment]
test/with_dummyserver/test_proxy_poolmanager.py:833: error: Item "None" of
    "ProxyConfig | None" has no attribute "ssl_context"  [union-attr]
```

**First one** — in `_connect_tls_proxy`, `ssl_context` is inferred as `SSLContext` from the `if` branch, then the `else` branch can assign `None`:

```python
ssl_context = self.ssl_context if self.proxy_is_forwarding else None
```

`cert_reqs` right below it already declares its union up front, so the fix matches the surrounding style:

```diff
+        ssl_context: ssl.SSLContext | None
         cert_reqs: int | str | None
         if proxy_ssl_context is not None:
```

**Second one** — `proxy_config` is `ProxyConfig | None`, and `proxy_from_url(...)` always sets it, so narrowing is enough:

```diff
         assert proxy.proxy_ssl_context is ssl_context
+        assert proxy.proxy_config is not None
         assert proxy.proxy_config.ssl_context is None
```

With both applied on your branch (`f1ea8b14`):

```
$ mypy -p urllib3    # connection.py: no errors
$ mypy -p test       # test_proxy_poolmanager.py: no errors
```

Two things I checked so this is not misleading:

- **The `coverage` failure is separate** and I have not looked at it; this only addresses `lint`.
- `black --check` also flags `connection.py`, but the diff is at lines 270/312 (`(version, code, message)` → `version, code, message`), which is **pre-existing and unrelated to this PR** — my local black is simply a different version than CI's. I left it alone.

Happy to open this as a PR against your branch if that is easier than applying it by hand.

src/urllib3/connection.py
shazow: Tried both, the former version felt like it was warping the intent of _WrappedAndVerifiedSocket too much, so went with the latter.

changelog/5093.bugfix.rst
shazow: Yeaaa I was too tunnelvisioned about documenting literal changes, removed that one.

I looked at several changelog examples to try and find the appropriate tone we want there, but maybe I should rephrase to be more consumer-centric?

illia-v (approved):
Looks good to me! Thanks for the great work, @shazow!

PR #5094, opened 2026-06-28
Don't parameterize tests using non-Collection iterables

cjwatson:
This is deprecated as of pytest 9.1; see
https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method.


illia-v (approved):
Thanks for reporting this and preparing a patch!

PR #5095, opened 2026-06-29
Reject invalid URL hosts

Ninja3047:
Fixed URL parsing to more strictly enforce RFC 3986 host syntax, rejecting
invalid host input such as raw spaces and control characters, malformed
percent-encodings, and percent-encoded control characters in HTTP(S) hosts and
IPv6 zone identifiers, including proxy CONNECT tunnel targets. Host
normalization now also follows RFC 3986 normalization rules for percent-encoded
octets by decoding percent-encoded unreserved characters and uppercasing the
hexadecimal case of retained percent-encoded octets.

---

William Tan from Trail of Bits in collaboration with OpenAI.

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import re
 import typing
+from functools import partial
 
 from ..exceptions import LocationParseError
 from .util import to_str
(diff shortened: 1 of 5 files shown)


src/urllib3/util/url.py
```diff
@@ -457,43 +503,48 @@ def parse_url(url: str) -> Url:
...
-        scheme, authority, path, query, fragment = _URI_RE.match(url).groups()  # type: ignore[union-attr]
-        normalize_uri = scheme is None or scheme.lower() in _NORMALIZABLE_SCHEMES
-
-        if scheme:
-            scheme = scheme.lower()
-
-        if authority:
-            auth, _, host_port = authority.rpartition("@")
-            auth = auth or None
-            host, port = _HOST_PORT_RE.match(host_port).groups()  # type: ignore[union-attr]
```
illia-v, line 470: For transparency, this could have previously raised an `AttributeError`. This PR replaces the line to harden validation, removing the need for the old error handling.

illia-v (approved):
LGTM! Thanks, William

PR #5097, opened 2026-07-01
Deal with ssl.PROTOCOL_TLSv1 removal from Python built with OpenSSL 4+

hroncok:
https://github.com/python/cpython/commit/3364e7e62fa24d0e19133fb0f90b1c24ef1110c5
removed ssl.PROTOCOL_TLSv1 when python is built with OpenSSL 4+.

We are hitting this error in Fedora (where we already updated OpenSSL):

    ImportError while loading conftest '/builddir/build/BUILD/python-urllib3-2.7.0-build/urllib3-2.7.0/test/conftest.py'.
    test/__init__.py:42: in <module>
        import urllib3.contrib.pyopenssl as pyopenssl
    ../BUILDROOT/usr/lib/python3.15/site-packages/urllib3/contrib/pyopenssl.py:72: in <module>
        ssl.PROTOCOL_TLSv1: OpenSSL.SSL.TLSv1_METHOD,
        ^^^^^^^^^^^^^^^^^^
    E   AttributeError: module 'ssl' has no attribute 'PROTOCOL_TLSv1'. Did you mean '.PROTOCOL_TLS' instead of '.PROTOCOL_TLSv1'?

Used LLM to figure out what to use as a replacement protocol in tests.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -68,8 +68,10 @@ __all__ = ["inject_into_urllib3", "extract_from_urllib3"]
 _openssl_versions: dict[int, int] = {
     util.ssl_.PROTOCOL_TLS: OpenSSL.SSL.SSLv23_METHOD,  # type: ignore[attr-defined]
     util.ssl_.PROTOCOL_TLS_CLIENT: OpenSSL.SSL.SSLv23_METHOD,  # type: ignore[attr-defined]
-    ssl.PROTOCOL_TLSv1: OpenSSL.SSL.TLSv1_METHOD,
 }
+
+if hasattr(ssl, "PROTOCOL_TLSv1") and hasattr(OpenSSL.SSL, "TLSv1_METHOD"):
+    _openssl_versions[ssl.PROTOCOL_TLSv1] = OpenSSL.SSL.TLSv1_METHOD
 
 if hasattr(ssl, "PROTOCOL_TLSv1_1") and hasattr(OpenSSL.SSL, "TLSv1_1_METHOD"):
     _openssl_versions[ssl.PROTOCOL_TLSv1_1] = OpenSSL.SSL.TLSv1_1_METHOD
--- a/changelog/5097.bugfix.rst
+++ b/changelog/5097.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed an ``AttributeError`` on Python built with OpenSSL 4+, where
+``ssl.PROTOCOL_TLSv1`` no longer exists.
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -176,19 +176,21 @@ class TestSSL:
 
         context.set_ciphers.assert_not_called()
 
+    # PROTOCOL_TLS_SERVER is used as a stand-in for any non-default ssl_version.
+    # PROTOCOL_TLSv1/TLSv1_2 are unavailable when Python is built with OpenSSL 4+.
     @pytest.mark.parametrize(
         "kwargs",
         [
             {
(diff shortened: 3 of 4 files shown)


illia-v (commented):
Thanks for reporting this!

Please change the tests to use `PROTOCOL_TLSv1_2` instead of `PROTOCOL_TLSv1`.

This should satisfy the following check catching exact TLS versions
https://github.com/urllib3/urllib3/blob/c420e267acb739917bee96e021ad9a2d27de8aae/src/urllib3/util/ssl_.py#L222-L223

hroncok:
I am not sure I follow. That still gets:

    AttributeError: module 'ssl' has no attribute 'PROTOCOL_TLSv1_2'

DPS0340:
Drive-by: the only red check here is `changelog entry`, and it is a missing file rather than anything wrong with the change. Since you are packaging this downstream rather than working in this repo day to day, here is the exact shape it wants.

Create `changelog/5097.bugfix.rst` containing one past-tense sentence aimed at users:

```rst
Fixed an ``AttributeError`` on Python built with OpenSSL 4+, where
``ssl.PROTOCOL_TLSv1`` no longer exists.
```

Naming is `<ISSUE-OR-PR>.<TYPE>.rst`; `changelog/README.rst` says to use the issue number if there is one, and the PR number otherwise, which is `5097` here. Types are the five towncrier defaults — `bugfix` is the right one for this.

There is also an escape hatch if the maintainers would rather not spend a line on it: the workflow passes when the **`Skip Changelog`** label is applied. That is a maintainer action, not something you can set yourself, but worth knowing it exists.

For what it is worth this looks like a real downstream blocker rather than a nicety — Fedora already ships OpenSSL 4, and the `ImportError` in your description breaks `conftest` at collection time, so the whole test suite fails to run, not just the TLSv1 cases.

Not my PR and I have no opinion on the fix itself; just clearing the one gate that is showing red.

hroncok:
This passes the CI but only with openssl < 4.

illia-v:
> I am not sure I follow. That still gets:
> 
> ```
> AttributeError: module 'ssl' has no attribute 'PROTOCOL_TLSv1_2'
> ```

Oh, I didn't know that `ssl.PROTOCOL_TLSv1_1` and `ssl.PROTOCOL_TLSv1_2` were removed too. I can see this in https://github.com/python/cpython/pull/146217 now.

illia-v (approved):

illia-v:
I've checked that the following imports still succeed in a Fedora 45 container:
https://github.com/urllib3/urllib3/blob/627636551e0e0159996b0f28dd21a60372cc5b10/src/urllib3/util/ssl_.py#L56-L71

These ones should start failing, but they have already been handled grancefuly:
https://github.com/urllib3/urllib3/blob/627636551e0e0159996b0f28dd21a60372cc5b10/src/urllib3/util/ssl_.py#L82-L90

src/urllib3/contrib/pyopenssl.py
```diff
@@ -70,9 +68,11 @@ class UnsupportedExtension(Exception):  # type: ignore[no-redef]
 _openssl_versions: dict[int, int] = {
     util.ssl_.PROTOCOL_TLS: OpenSSL.SSL.SSLv23_METHOD,  # type: ignore[attr-defined]
     util.ssl_.PROTOCOL_TLS_CLIENT: OpenSSL.SSL.SSLv23_METHOD,  # type: ignore[attr-defined]
```
xnox, line 70: Shouldn't the above two - also be TLS_METHOD?

hroncok: Why?

PR #5098, opened 2026-07-03
Type pool_classes_by_scheme as a mapping to pool callables

### LINKED ISSUE #3554: typing: unable to override `pool_classes_by_scheme` with a `partial`d class
### Subject

incredibly specific unfortunately -- https://github.com/getsentry/sentry/blob/00f7a931ef9cb9efc31658a89d85eed19a52155e/src/sentry/net/http.py#L138-L141

we have to patch all the way down to do ip address filtering in requests / urllib3 -- to override the connection creation mechanism

as a result we override pool classes and it appears the typing for this is a little too strict (`dict[str, type[HTTPConnectionPool]]`)

### Environment

Describe your environment.

```pycon
>>> import platform
>>> import ssl
>>> import urllib3
>>> 
>>> print("OS", platform.platform())
OS macOS-15.2-arm64-arm-64bit-Mach-O
>>> print("Python", platform.python_version())
Python 3.13.1
>>> print(ssl.OPENSSL_VERSION)
OpenSSL 3.0.15 3 Sep 2024
>>> print("urllib3", urllib3.__version__)
urllib3 2.2.2
>>> ^D
```

### Steps to Reproduce

I don't have a good minimal reproduction but the error from mypy is as follows:

```
src/sentry/net/http.py:138: error: Dict entry 0 has incompatible type "str": "partial[SafeHTTPConnectionPool]"; expected "str": "type"  [dict-item]
src/sentry/net/http.py:139: error: Dict entry 1 has incompatible type "str": "partial[SafeHTTPSConnectionPool]"; expected "str": "type"  [dict-item]
Found 2 errors in 1 file (checked 6465 source files)
```

### Expected Behavior

ideally no errors

### Actual Behavior

see above

___

an aside -- if I adjust the annotation slightly I can get this to pass:

```diff
-pool_classes_by_scheme = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
+pool_classes_by_scheme: dict[str, typing.Callable[..., HttpConnectionPool]] = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
```

this ~slightly reduces type safety in that a call to the value of this will no longer have its arguments checked agains the shape of `HTTPConnection` -- but it does allow more flexibility for what values can be in the dictionary (while still preserving the return value shape)

I'll probably just `# type: ignore` for now though since that's the least friction

apoorvdarshan:
Closes #3554

## Problem

`pool_classes_by_scheme` is typed as `dict[str, type[HTTPConnectionPool]]`, so overriding it with a callable that returns a pool — e.g. a `functools.partial`, as [Sentry does](https://github.com/getsentry/sentry/blob/00f7a931ef9cb9efc31658a89d85eed19a52155e/src/sentry/net/http.py#L138-L141) to customize connection creation for IP filtering — fails type checking, even though it works at runtime:

```python
pm.pool_classes_by_scheme = {"http": functools.partial(HTTPConnectionPool), ...}
# error: Dict entry 0 has incompatible type "str": "partial[HTTPConnectionPool]";
#        expected "str": "type"  [dict-item]
```

## Fix

Type the values as `Callable[..., HTTPConnectionPool]` (via a `_TYPE_POOL_CLASSES_BY_SCHEME` alias). A pool *class* is already a valid `Callable[..., HTTPConnectionPool]`, so this is a safe widening — every existing usage still type-checks, and `_new_pool` continues to call the value as `pool_cls(host, port, **request_context)`. The local annotation in `_new_pool` is updated to match.

## Verification

- Added `test_pool_classes_by_scheme_accepts_callable` to `test/test_poolmanager.py`. It is primarily a typing regression test (mypy runs over the `test` package): the `partial` assignment produces the `[dict-item]` error on `main` and type-checks with this change, and the test also exercises the assignment at runtime.
- `mypy --strict` on a minimal reproduction: clean with the fix, errors without it.
- `test_poolmanager.py`, `test_proxymanager.py`, `test_connectionpool.py`: 146 passed. `ruff check` clean; the diff touches only the changed lines.

## Disclosure

Prepared with the assistance of an AI tool (Claude Code); I reproduced the issue, reviewed and verified the change and test results, and will respond to review feedback personally.

### DIFF
--- a/changelog/3554.bugfix.rst
+++ b/changelog/3554.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the type hint of ``PoolManager.pool_classes_by_scheme`` to accept any callable returning a connection pool (for example a ``functools.partial``), not only a pool class.
--- a/src/urllib3/poolmanager.py
+++ b/src/urllib3/poolmanager.py
@@ -156,7 +156,17 @@ key_fn_by_scheme = {
     "https": functools.partial(_default_key_normalizer, PoolKey),
 }
 
-pool_classes_by_scheme = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
+#: A scheme may map to any callable returning an :class:`HTTPConnectionPool`,
+#: not just a pool class, so that e.g. a ``functools.partial`` can be used to
+#: customize pool creation.
+_TYPE_POOL_CLASSES_BY_SCHEME = typing.Dict[
+    str, typing.Callable[..., HTTPConnectionPool]
+]
+
+pool_classes_by_scheme: _TYPE_POOL_CLASSES_BY_SCHEME = {
+    "http": HTTPConnectionPool,
+    "https": HTTPSConnectionPool,
+}
 
 
 class PoolManager(RequestMethods):
@@ -256,7 +266,8 @@ class PoolManager(RequestMethods):
         connection pools handed out by :meth:`connection_from_url` and
         companion methods. It is intended to be overridden for customization.
         """
-        pool_cls: type[HTTPConnectionPool] = self.pool_classes_by_scheme[scheme]
+        pool_cls: typing.Callable[..., HTTPConnectionPool]
+        pool_cls = self.pool_classes_by_scheme[scheme]
         if request_context is None:
             request_context = self.connection_pool_kw.copy()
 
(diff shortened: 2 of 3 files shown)


apoorvdarshan:
Applied the pyupgrade and Black rewrites reported by the lint job. All configured pre-commit hooks pass for the changed files, and test/test_poolmanager.py passes (39 tests).

apoorvdarshan:
The fresh lint run revealed an existing duck-typed custom pool in test_proxymanager. I widened the map values to callable factories returning Any and cast at the typed API boundary, preserving both partial factories and the supported duck-typed pool. Full mypy passes across 83 source files; pool/proxy tests pass (52 tests), and all changed-file hooks pass.

PR #5101, opened 2026-07-06
Preserve explicit port 0 in connection_from_url and HTTP/2 authority

nishat-06:
connection_from_url() and the HTTP/2 request authority build the port with a truthiness check, so an explicit :0 in the URL gets dropped and replaced by the default 80 or 443. This is the same handling #5071 corrected for URL parsing, pool selection, and proxy configuration, but these two call sites were left using `port or <default>`. Switching them to an explicit None check keeps an explicit port 0.

### DIFF
--- a/src/urllib3/http2/connection.py
+++ b/src/urllib3/http2/connection.py
@@ -127,10 +127,11 @@ class HTTP2Connection(HTTPSConnection):
         self._request_url = url or "/"
         self._validate_path(url)  # type: ignore[attr-defined]
 
+        port = self.port if self.port is not None else 443
         if ":" in self.host:
-            authority = f"[{self.host}]:{self.port or 443}"
+            authority = f"[{self.host}]:{port}"
         else:
-            authority = f"{self.host}:{self.port or 443}"
+            authority = f"{self.host}:{port}"
 
         self._headers.append((b":scheme", b"https"))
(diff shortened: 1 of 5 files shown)


src/urllib3/http2/connection.py
```diff
@@ -127,10 +127,11 @@ def putrequest(  # type: ignore[override]
         self._request_url = url or "/"
         self._validate_path(url)  # type: ignore[attr-defined]
 
+        port = self.port if self.port is not None else 443
         if ":" in self.host:
-            authority = f"[{self.host}]:{self.port or 443}"
+            authority = f"[{self.host}]:{port}"
         else:
-            authority = f"{self.host}:{self.port or 443}"
+            authority = f"{self.host}:{port}"
```
illia-v, line 134: Please add a test covering these lines

illia-v (commented):
Thanks!

This is a good follow-up to a recent #5071.

src/urllib3/http2/connection.py
nishat-06: Done. Added test_request_authority_port_zero in test/test_http2_connection.py, it builds the request with port 0 and asserts the :authority header comes out as example.com:0. It fails against the old self.port or 443 (gives :443) and passes with the None check.

illia-v (approved):

PR #5109, opened 2026-07-08
Do not re-apply unchanged options to a used PyOpenSSL context (#5107)

### LINKED ISSUE #5107: ValueError when reusing an ssl_context with PyOpenSSL backend
### Subject

When `urllib3.contrib.pyopenssl.inject_into_urllib3()` is used and a custom `ssl_context` is provided to `HTTPSConnectionPool`, creating a second connection from the pool raises:

```
ValueError: Context has already been used to create a Connection, it cannot be mutated again
```

This occurs because the context is unconditonally changed for every new connection established in the pool.

https://github.com/urllib3/urllib3/blob/4e5ce638723118dc41f218cd6715f91a9f321908/src/urllib3/connection.py#L987

And in recent pyOpenSSL release (26.2.0+), however, modifying a `PyOpenSSLContext` after it has already spawned a connection is prohibited on the OpenSSL wrapper level (https://github.com/pyca/pyopenssl/pull/1429).

Is there a recommended approach or potential fix within urllib3 to address this compatibility issue?

The same code (below) runs fine without PyOpenSSL injection.

### Environment

OS macOS/Linux
Python 3.11/3.12/3.13
OpenSSL 3.6.2 7 Apr 2026
urllib3 2.7.0

### Steps to Reproduce

- Set up a fresh virtual environment and install the following packages:
```
pip install pyopenssl==26.2.0 urllib3 certifi idna
```

- Run the script below
```python
import ssl
import platform
import urllib3
import certifi


print(ssl.OPENSSL_VERSION)


# 1. Enable PyOpenSSL backend

# 2. Create a shared SSLContext
ssl_context = urllib3.util.create_urllib3_context()
pool = urllib3.HTTPSConnectionPool(
    "storage.googleapis.com",
    ssl_context=ssl_context,
    maxsize=1,
    cert_reqs="CERT_REQUIRED",
    ca_certs=certifi.where()
)

# 3. First request (creates Connection #1 on ssl_context)
r1 = pool.request("GET", "/", headers={"Connection": "close"})

# 4. Second request (creates Connection #2 on ssl_context)
r2 = pool.request("GET", "/")
print("pyOpenSSL version:", urllib3.contrib.pyopenssl.OpenSSL.__version__)
print("Request 1 status:", r1.status)
print("Request 2 status:", r2.status)
```

### Expected Behavior
Status code 400 is expected here since we are using a dummy url.
```
Request 1 status: 400
Request 2 status: 400
```

### Actual Behavior

```
  File "./test.py", line 31, in <module>
    r2 = pool.request("GET", "/")
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/_request_methods.py", line 135, in request
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/_request_methods.py", line 182, in request_encode_url
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 788, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 464, in _make_request
    self._validate_conn(conn)
  File "./venv/lib/python3.12/site-packages/urllib3/connectionpool.py", line 1106, in _validate_conn
    conn.connect()
  File "./venv/lib/python3.12/site-packages/urllib3/connection.py", line 796, in connect
    sock_and_verified = _ssl_wrap_socket_and_match_hostname(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/connection.py", line 937, in _ssl_wrap_socket_and_match_hostname
    ^^^^^^^^^^^^^^^^^^^
  File "./venv/lib/python3.12/site-packages/urllib3/contrib/pyopenssl.py", line 452, in verify_mode
    self._ctx.set_verify(_stdlib_to_openssl_verify[value], _verify_callback)
  File "./venv/lib/python3.12/site-packages/OpenSSL/SSL.py", line 869, in inner
    raise ValueError(
ValueError: Context has already been used to create a Connection, it cannot be mutated again
```

apoorvdarshan:
## Summary

Fixes #5107. With the `pyopenssl` contrib module and a custom `ssl_context`, creating a **second** connection from a pool raised:

```
ValueError: Context has already been used to create a Connection, it cannot be mutated again
```

(Reported as the root cause of an Apache Beam pipeline failure.)

## Root cause

`PyOpenSSLContext.options`, `minimum_version` and `maximum_version` setters each call `_set_ctx_options()` → `self._ctx.set_options(...)`. A connection pool re-applies the same `ssl_context` settings for **every** new connection, but once the context has been used to create an `OpenSSL.SSL.Connection` (in `wrap_socket`), OpenSSL forbids any further mutation — so the second connection's (unchanged) re-assignment raised.

## Fix

Skip the mutation when the value is unchanged, so re-applying the same settings is a no-op. A genuine change on a not-yet-used context still applies as before.

## Verification

- Added `test_reapplying_unchanged_options_after_use` (a unit test): it reproduces the exact `ValueError` on `main` and passes with this change.
- `TestPyOpenSSLHelpers` unit tests pass (5). The `pyopenssl` integration tests that need a live TLS dummyserver fail identically with and without this change in my environment (unrelated).
- `ruff check` / `ruff format --check` clean. Added `changelog/5107.bugfix.rst`.

---

*Disclosure: prepared with AI assistance; reviewed and verified locally.*

### DIFF
--- a/changelog/5107.bugfix.rst
+++ b/changelog/5107.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``ValueError`` (``Context has already been used to create a Connection``) when reusing a ``ssl_context`` across multiple connections with the ``pyopenssl`` contrib module. Re-applying unchanged ``options``, ``minimum_version`` and ``maximum_version`` on the wrapped context is now a no-op.
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -438,6 +438,11 @@ class PyOpenSSLContext:
 
     @options.setter
     def options(self, value: int) -> None:
+        if value == self._options:
+            # Re-applying the same value would call ``set_options`` again, which
+            # raises once the context has been used to create a connection. See
+            # https://github.com/urllib3/urllib3/issues/5107.
+            return
         self._options = value
         self._set_ctx_options()
 
@@ -547,6 +552,8 @@ class PyOpenSSLContext:
 
     @minimum_version.setter
     def minimum_version(self, minimum_version: int) -> None:
+        if minimum_version == self._minimum_version:
+            return
         self._minimum_version = minimum_version
         self._set_ctx_options()
 
@@ -556,6 +563,8 @@ class PyOpenSSLContext:
 
     @maximum_version.setter
     def maximum_version(self, maximum_version: int) -> None:
+        if maximum_version == self._maximum_version:
+            return
         self._maximum_version = maximum_version
         self._set_ctx_options()
 
--- a/test/contrib/test_pyopenssl.py
+++ b/test/contrib/test_pyopenssl.py
@@ -85,6 +85,34 @@ class TestPyOpenSSLHelpers:
 
         assert _dnsname_to_stdlib(name) == expected_result
 
+    def test_reapplying_unchanged_options_after_use(self) -> None:
+        """Re-applying unchanged TLS settings must not fail after the context
+        has been used to create a connection (see GH-5107).
+
+        A connection pool re-applies the same ``ssl_context`` settings for every
+        new connection, but OpenSSL forbids mutating a context once it has been
+        used, so the no-op assignments must be skipped.
+        """
+        import ssl
+
+        import OpenSSL.SSL
+
+        from urllib3.contrib.pyopenssl import PyOpenSSLContext
+
+        ctx = PyOpenSSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        ctx.minimum_version = ssl.TLSVersion.TLSv1_2
+        ctx.maximum_version = ssl.TLSVersion.TLSv1_3
+        ctx.options |= ssl.OP_NO_COMPRESSION
+
+        # Using the context to create a connection freezes it.
+        OpenSSL.SSL.Connection(ctx._ctx)
+
+        # Re-applying the current values (as a pool does per connection) is a
+        # no-op and must not raise.
+        ctx.minimum_version = ctx.minimum_version
+        ctx.maximum_version = ctx.maximum_version
+        ctx.options = ctx.options
+
     @mock.patch("urllib3.contrib.pyopenssl.log.warning")
     def test_get_subj_alt_name(self, mock_warning: mock.MagicMock) -> None:
         """


DPS0340:
Drive-by triage, since this has been sitting three weeks with one red check and no reviews: **the Windows 3.14 failure is unrelated to this PR.** Posting the evidence so an empty-looking red mark is not what keeps it stalled.

There are **zero failed tests** in that run. The exit code comes from a single teardown error:

```
____ ERROR at teardown of TestSSL.test_ssl_failed_fingerprint_verification ____
  File "dummyserver\socketserver.py", line 143, in run
  File "test\with_dummyserver\test_socketlevel.py", line 1639, in socket_handler
    ssl_sock.send(b"HTTP/1.1 200 OK\r\n" ...)
  File "ssl.py", line 1232, in send
ConnectionAbortedError: [Errno 10053] An established connection was aborted
                        by the software in your host machine
```

Three things place it outside this PR:

1. **It is in a dummyserver background thread**, not the test body — the server tried to write a response after the client had already gone away.
2. **`WSAECONNABORTED` (10053) is the classic Windows race** for that shutdown ordering; on Linux the same sequence is a silent `EPIPE` that the thread swallows.
3. **The failing path is `test/with_dummyserver/test_socketlevel.py`**, while this PR touches only:
   ```
   changelog/5107.bugfix.rst
   src/urllib3/contrib/pyopenssl.py
   test/contrib/test_pyopenssl.py
   ```
   The pyopenssl tests in that same run all passed — I checked the log rather than assuming, and `test_create_urllib3_context_*`, `test_assert_fingerprint_*` are green.

So the 37 green checks are the real signal here and a re-run should clear it.

Worth adding for whoever picks this up: **#5107 is not a niche report.** It broke an Apache Beam pipeline (apache/beam#39235), and pyOpenSSL 26.2.0+ made the mutation a hard error rather than a quiet one, so the surface widens as that release propagates.

Not my PR and I am not proposing changes to it — just handing over the log analysis so the red mark can be discounted with evidence instead of a guess.

illia-v:
I don't see how this fixes #5107 because the reported error happened while `context.verify_mode` was changed.

PR #5112, opened 2026-07-10
cap chunk-size and trailer line length in read_chunked

nishat-06:
On a chunked response, urllib3's own reader takes the chunk-size line and the trailer lines without the length bound that http.client applies:

    # HTTPResponse._update_chunk_length / read_chunked trailer loop
    line = self._fp.fp.readline()      # unbounded; stdlib uses readline(_MAXLINE + 1)

A malicious server can then make `.stream()`/`.read_chunked()` buffer an arbitrarily large single line into memory before any length check. This applies the same `_MAXLINE + 1` / `LineTooLong` guard that #5091 added to the tunnel reader to both readline sites in the chunked path.


nishat-06:
gentle ping

PR #5115, opened 2026-07-10
Document retry backoff guidance

### LINKED ISSUE #3677: Document good defaults for retry: backoff factor
### Context

Default value for backoff factor in Retry class is `0`. This value may be deadly for a real-world service that's being called with a client code.

I think the defaults should try to be as close to “best practices” as possible. They say that backoff factor for not-lightweight queries should be 0.5-1 second.

Maybe `backoff_factor=0.7` and `backoff_jitter=0.7` would be good defaults?

Background: my carefree retries config in GH Actions probably made Plausible to [restrict their export endpoint](https://github.com/plausible/analytics/pull/5089).

### Alternatives

None.

### Duplicate

No issue found.

### Contribution

Yes.

yashkuceriya:
Closes #3677.

## Summary

- Documents why production retry policies should configure a non-zero `Retry.backoff_factor`.
- Adds a concrete status-retry example using `status_forcelist` with backoff.
- Adds a changelog fragment for the user-facing documentation update.

## Why

Issue #3677 asks whether urllib3 should change the default backoff behavior. A maintainer noted that changing the default may be surprising, but that the documentation should make the recommended production configuration clearer. This keeps runtime behavior unchanged and surfaces the guidance in the user guide where retry examples are introduced.

## Validation

- `git diff --check`
- `UV_CACHE_DIR=/private/tmp/uv-cache PATH="/Users/yash/Library/Python/3.14/bin:$PATH" python3 -m nox -s docs`

### DIFF
--- a/changelog/3677.misc.rst
+++ b/changelog/3677.misc.rst
@@ -1,0 +1,1 @@
+Documented configuring ``Retry.backoff_factor`` for production retry policies.
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -585,14 +585,13 @@ To disable redirects but keep the retrying logic, specify ``redirect=False``:
 For more granular control you can use a :class:`~util.retry.Retry` instance.
 This class allows you far greater control of how requests are retried.
 
-When enabling retries in production, configure a non-zero ``backoff_factor``
-to avoid retrying repeatedly without any delay. urllib3 will sleep between
-retry attempts using an exponential backoff. For example,
-``backoff_factor=0.1`` sleeps for ``0.0s``, ``0.2s``, ``0.4s``, ``0.8s``,
-and so on between retries. urllib3 also respects the ``Retry-After`` response
-header by default when present for supported retry responses.
+When enabling retries against production services, consider configuring a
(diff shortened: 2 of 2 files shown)


sigmavirus24 (commented):
Given this is already documented elsewhere, why should we accept this contribution? There seems little additional value here

yashkuceriya:
Thanks for checking this. That's fair — the API reference already documents the mechanics of `backoff_factor`, so I narrowed the wording here to avoid repeating the formula.

The value I was aiming for is discoverability in the User Guide: issue #3677 calls out production retry policies where `backoff_factor=0` can cause back-to-back retries, and the current User Guide retry examples introduce `Retry` without flagging backoff at the point users are likely to copy a configuration. The updated version keeps the default behavior unchanged, points readers back to `Retry` for the detailed calculation, and adds a practical status-retry example for 429/5xx responses.

If you still feel the API reference is sufficient, I'm happy to close this rather than add noise.

opq0pqo (approved):

opq0pqo (approved):

PR #5117, opened 2026-07-13
Raise when iterating closed responses

### LINKED ISSUE #1780: iterating a closed response improperly produces data
Consider the following script:
```
import urllib3

http = urllib3.PoolManager()
resp.close()
for d in resp:
    print(repr(d))
```
With urllib3 1.25.7, this program prints `b''`. With urllib3 1.24.3, one sees:
```
  File "example.py", line 6, in <module>
    for d in resp:
ValueError: I/O operation on closed file.
```
The latter is in line with what I expect.

Sanjays2402:
Fixes #1780.

Iterating an already-closed `HTTPResponse` now raises the standard `ValueError: I/O operation on closed file.` instead of silently appearing empty. The regression test fails without the source change; `test/test_response.py` and the full suite pass locally (2231 passed, 333 skipped, 45 xfailed).

Prepared with AI assistance.

### DIFF
--- a/changelog/1780.bugfix.rst
+++ b/changelog/1780.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed iteration over a closed ``HTTPResponse`` to raise ``ValueError`` instead
+of silently producing no data.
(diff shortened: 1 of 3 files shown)


sigmavirus24:
The AI didn't assist you very much. The discussion shows that a very different approach is appropriate here

PR #5119, opened 2026-07-15
reject CR/LF and NUL in multipart field headers

nishat-06:
RequestField only escapes the name and filename params through format_multipart_header_param, so the Content-Type/Content-Location values set by make_multipart and any caller-supplied header names/values are written into the multipart/form-data body unchecked. A CR/LF in an attacker-influenced value, such as a forwarded upload Content-Type, then injects extra part headers or whole parts into the request. render_headers now rejects CR, LF, and NUL in the header names and values it emits, matching the escaping already applied to name and filename.

### DIFF
--- a/changelog/5119.bugfix.rst
+++ b/changelog/5119.bugfix.rst
@@ -1,0 +1,1 @@
+Rejected carriage return, line feed, and NUL characters in ``multipart/form-data`` field header names and values so an untrusted value (such as a forwarded upload ``Content-Type``) can no longer inject extra part headers into the request body.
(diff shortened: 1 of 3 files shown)


sethmlarson:
We assume that the `Content-Type` is under control of the program, because most programs will hard-code this value.

nishat-06:
Makes sense. If those values are program-controlled then there's no untrusted input to guard against here, and the same reasoning covers the caller-supplied headers dict. Fair call, I'll leave it.

PR #5120, opened 2026-07-15
docs: correct description of Retry.RETRY_AFTER_STATUS_CODES

### LINKED ISSUE #3546: Misleading description for RETRY_AFTER_STATUS_CODES
In [the docs](https://urllib3.readthedocs.io/en/2.3.0/reference/urllib3.util.html#urllib3.util.Retry.RETRY_AFTER_STATUS_CODES), it says:

> `RETRY_AFTER_STATUS_CODES = frozenset({413, 429, 503})`
> Default status codes to be used for `status_forcelist`

But this doesn't seem correct. `status_forcelist` defaults to `None`. Looking at the code, there doesn't seem to be any connection between `RETRY_AFTER_STATUS_CODES` and `status_forcelist`.

pctablet505:
Fixes #3546

The docstring for Retry.RETRY_AFTER_STATUS_CODES incorrectly stated that these codes were defaults for status_forcelist. In practice, they are the status codes that trigger a retry when a Retry-After header is present and respect_retry_after_header is enabled.


sethmlarson (approved):

src/urllib3/util/retry.py
```diff
@@ -198,7 +198,8 @@ class Retry:
         ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
     )
 
-    #: Default status codes to be used for ``status_forcelist``
+    #: Default status codes that trigger a retry when a Retry-After header is
+    #: present and :attr:`Retry.respect_retry_after_header` is enabled.
```
illia-v, line 202: `nox -s docs` fails with a warning after this change

```
src/urllib3/util/retry.py:docstring of urllib3.util.Retry.RETRY_AFTER_STATUS_CODES:1: WARNING: py:attr reference target not found: Retry.respect_retry_after_header [ref.attr]
```

PR #5122, opened 2026-07-17
reject ambiguous Content-Length/Transfer-Encoding request framing

nishat-06:
`HTTPConnection.request()` picks one framing mode but never rejects a caller supplying both, so `headers={"Content-Length": "5", "Transfer-Encoding": "chunked"}` sends the body Content-Length-framed while still emitting the caller's `Transfer-Encoding: chunked` verbatim, and `chunked=True` alongside a caller-supplied `Content-Length` chunk-encodes the body while still emitting the `Content-Length`. Both put two framing headers on the wire, which RFC 9112 Section 6.3 tells senders not to do, and it matters when an application forwards client-supplied headers through urllib3 (a proxy or gateway), since a front-end honouring one header and a back-end honouring the other disagree on where the body ends. http.client has the same shape and doesn't reject it either, so the check has to live here, and it raises before `putrequest()` so nothing goes out; HTTP/2 already drops `transfer-encoding` and the emscripten path leaves framing to the browser, so this is the only affected path.

### DIFF
--- a/src/urllib3/connection.py
+++ b/src/urllib3/connection.py
@@ -493,6 +493,14 @@ class HTTPConnection(_HTTPConnection):
         if headers is None:
             headers = {}
         header_keys = frozenset(to_str(k.lower()) for k in headers)
+        if "content-length" in header_keys and (
+            chunked or "transfer-encoding" in header_keys
+        ):
+            raise ValueError(
+                "Content-Length and Transfer-Encoding can't both be set, "
+                "the framing of the request would be ambiguous "
+                "(RFC 9112 Section 6.3)"
+            )
         skip_accept_encoding = "accept-encoding" in header_keys
         skip_host = "host" in header_keys
         self.putrequest(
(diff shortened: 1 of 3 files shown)


src/urllib3/connection.py
```diff
@@ -493,6 +493,14 @@ def request(  # type: ignore[override]
         if headers is None:
             headers = {}
         header_keys = frozenset(to_str(k.lower()) for k in headers)
+        if "content-length" in header_keys and (
```
Sanjays2402, line 496: This still accepts `Transfer-Encoding: gzip`; below, any TE header flips `chunked=True`, so the body is chunk-framed even though the header never advertises `chunked`. Should this reject a TE value whose final coding isn’t `chunked` while tightening request framing?

nishat-06: Good catch, that's real. Confirmed it on the wire with `Transfer-Encoding: gzip` and a `b"hello"` body:

```
POST / HTTP/1.1
Host: 127.0.0.1:58811
Transfer-Encoding: gzip

5
hello
```

The `elif "transfer-encoding" in header_keys: chunked = True` branch treats any TE value as chunked, so the body ends up chunk-framed while the header advertises a coding that isn't chunked.

I'd rather keep it out of this PR though. It's a different failure mode: this PR is about two framing headers going out at once (RFC 9112 6.3, the CL.TE/TE.CL disagreement), whereas that one is a single header whose value doesn't match the framing actually used. Fixing it also means deciding what to do with legitimate values like `gzip, chunked`, where the final coding is chunked and the current behaviour is arguably right, so it needs its own tests and changelog. Happy to do it as a follow-up if that sounds reasonable.

nishat-06:
any update?

PR #5126, opened 2026-07-19
Compare redirect origin against the request URL, not the pool

nishat-06:
PoolManager.urlopen() decides what to strip before following a redirect with conn.is_same_host(redirect_location), but for a plain-HTTP request forwarded through an HTTP proxy ProxyManager.connection_from_host() hands back the proxy's pool, so the comparison is against the proxy rather than the request host: remove_headers_on_redirect fires on legitimate same-host redirects, and is skipped when the server points Location at the proxy's own origin. The same recursion passes the already-populated headers dict back through ProxyManager._set_proxy_headers(), where the Host derived from the previous URL wins over the one derived from the new location, so a cross-host redirect goes out as `GET http://b.example/second HTTP/1.1` with `Host: a.example` and the request target and Host name different origins. Comparing the parsed request URL against the parsed redirect target fixes both directions of the stripping decision, and Host is dropped when the origin changes so _set_proxy_headers() recomputes it for the new location.

Added a socket-level regression covering the three cases through a forwarding proxy (cross-host, same-host, and a redirect back to the proxy's origin); all three fail on main.


nishat-06:
any update?

src/urllib3/poolmanager.py
```diff
@@ -476,15 +496,16 @@ def urlopen(  # type: ignore[override]
...
-            redirect_location
-        ):
+        # Strip headers marked as unsafe to forward to the redirected location,
+        # along with Host, which ProxyManager._set_proxy_headers() derives from
+        # the request URL and so has to be recomputed for the new location.
+        # 'conn' is the proxy's pool when forwarding through an HTTP proxy, so
+        # compare the request URL itself rather than asking the pool.
+        if not _same_origin(u, parse_url(redirect_location)):
+            headers_to_remove = retries.remove_headers_on_redirect | {"host"}
             new_headers = kw["headers"].copy()
```
illia-v, line 506: `kw["headers"]` can be `None` leading to a redirect error here

src/urllib3/poolmanager.py
```diff
@@ -161,6 +161,26 @@ def _default_key_normalizer(
 pool_classes_by_scheme = {"http": HTTPConnectionPool, "https": HTTPSConnectionPool}
 
 
+def _same_origin(url: Url, other: Url) -> bool:
```
illia-v, line 164: Could we extract the comparison from `HTTPConnectionPool.is_same_host` into a shared helper and use it from both places?

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1168,6 +1168,74 @@ def echo_socket_handler(listener: socket.socket) -> None:
...
+        ],
+    )
+    def test_redirect_rewrites_host_and_strips_credentials(
+        self,
+        redirect_target: str | None,
+        expected_host: bytes | None,
+        expected_authorization: bool,
+    ) -> None:
+        if redirect_target is None:
+            redirect_target = f"http://{self.host}:{self.port}/second"
```
illia-v, line 1192: `self.port` is assigned by `_start_server()` below, so it's possible to get an attribute error if the `redirect_target is None` variant is run alone.

illia-v (changes requested):
Thanks for reporting this!

PR #5140, opened 2026-08-01
Add an unique extra attribute to retrying warning (so pip can rewrite it)

### LINKED ISSUE #2580: Allow tweaking the presentation style of logged errors
### Context

**What are you trying to do?**

I've been trying to improve pip's output (hopefully, you've noticed!) and our networking-related output seems like a really good candidate to try to improve. This includes various kinds of errors originating from urllib3 and especially the presentation when there's going to be a retry.

```
WARNING: Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x10439a3a0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known')': /simple/foooooooooooooo/
```

**How do you expect to do it?**

Would it be OK to change that into multiple log messages, at debug and warning levels? That way, the logging presentation would look like:

```
WARNING: Connection broken to /simple/foooooooooooooo/
WARNING: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known
DEBUG: NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x10439a3a0>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known')
WARNING: Will retry connection (3 remaining retries)
DEBUG: Retry(total=3, connect=None, read=None, redirect=None, status=None)
```

While a "regular" output would be:

```
WARNING: Connection broken to /simple/foooooooooooooo/
WARNING: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known
WARNING: Will retry connection (3 remaining retries)
```

I mean, honestly, an ideal solution would probably look like some way to hook into pip's new error message presentation:

```
warning: will-retry-network-failure

× Retrying connection to https://pypi.org/simple/foooooooooooooo/
╰─> Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known

    Remaining retries: 4

note: This is an issue with the network connection.
```

**Is it something you currently cannot do?**
Nope.

**Is this related to an existing issue/problem?**
Nope.

### Alternatives

**Can you achieve the same result doing it in an alternative way?**
I looked, and the only thing I could think of is a hacky formatter for urllib3. 🤷🏽 

**Is the alternative considerable?**
🤷🏽 

### Duplicate

**Has the feature been requested before?**
Not that I could find.

### Contribution

**Would you be willing to submit a PR?**
Depends on what level of complexity is involved in implementing this, and how aggressively we scope-creep on this. :)

ichard26:
This is intended for pip so it can identify and rewrite this warning to be user-friendly. This is a hack to emulate structured logging for one specific warning, but it's the least bad hack still feasible. See #2580.

Let me know if you want a test and where to add it. (I can figure out the other details.) Also, I'd recommend not putting this in the changelog unless you want other people to rely on this as well (which I doubt). On pip's end, I'll make sure that we have unit/integration tests to ensure this works for us and to notify us if something breaks. 

<details>
<summary>In terms of what this allows pip to do, here's an early screenshot from my corresponding pip patch</summary>

image

and before:

image
</details>

### DIFF
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -866,7 +866,16 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         if not conn:
             # Try again
             log.warning(
-                "Retrying (%r) after connection broken by '%r': %s", retries, err, url
+                "Retrying (%r) after connection broken by '%r': %s",
+                retries,
+                err,
+                url,
+                # Provide an unique attribute with the host needed by pip to
+                # rewrite this warning. Ideally, we'd go with a better solution,
+                # but backwards compatibility and other constraints make those
+                # unfeasible, so this is the least bad option.
+                # See also: https://github.com/urllib3/urllib3/issues/2580.
+                extra={"__urllib3-retry-warning": {"host": self.host}},
             )
             return self.urlopen(
                 method,


ichard26:
It'd be a bit more robust if we also provided the retry object (to read the remaining retry count) and the causing error in the extra dictionary as well. Depends on whether it's likely these warnings will change in the future. LMK!

sigmavirus24:
Including more things from day 0 means we can't change/remove them in the future. Without knowing the future it might be more problematic. 

I'm also looking above the current diff and I see that err is assigned from the original `e` instead of `new_e`, which I'm not certain is right. I'm on my phone so hard to verify the current behavior is 100% correct or if pip might want potentially the underlying exception or our wrapped error. I also have an instinct (but again I haven't verified) that some exceptions hold onto things that might lead it to look like a memory leak if it isn't deref'd because of passing it to the logging library. 

tl;dr I think we should start with a smaller footprint for now until we sort out other questions.

sigmavirus24 (approved):

sigmavirus24:
I approved but I defer to @illia-v if they want tests around logs which personally I'm not find of but I'm not strongly against. Also not enthusiastic about nothing this in the change log so feeling like we should be able to skip that check but again not a strong opinion

illia-v (approved):
I'm fine with this change as-is. I'd prefer to skip the changelog entry, since this is intended as an implementation detail for pip rather than something we want to advertise as a public interface.

ichard26:
Great, thank you @sigmavirus24 and @illia-v!

(Curious question, when is the next urllib3 release planned? I don't mean to pressure you -- take the time you need/follow whatever schedule you have -- just wondering when I can expect to file the corresponding PR for pip.)

illia-v:
> (Curious question, when is the next urllib3 release planned? I don't mean to pressure you -- take the time you need/follow whatever schedule you have -- just wondering when I can expect to file the corresponding PR for pip.)

I'd like to have a new release in August. We'll need to wait for CVE IDs, though, which may delay the release by a week or two.

PR #5143, opened 2026-08-03
Reject invalid ASCII characters in URL hosts

### LINKED ISSUE #1963: urllib3.util.url.parse_url doesn't handle Windows domain name separator in user part of the URL
One of our python application uses `requests`[1] package for HTTP invocations within the application. This `requests` package internally uses `urllib3` package for connection management. Things work fine in most of our normal use cases.
However, we have a specific setup where environment level proxies have been setup for HTTP interaction. We are on Linux and the `http_proxy` and `https_proxy` environment variables have been set appropriately. `requests` package picks up these environment variables (and we do want those to be picked up) and passes it along to the `urllib3` package during connection establishment. While doing so, the call ends up in `urllib3.util.url.parse_url` where the value that's set for the `http_proxy` environment variable gets passed to that function.
In one of our setups, the value is of the form `http://engr\john:pass@192.168.11.2:1234`. Notice the "user" part in that URL. It's `engr\john` which represents a Windows style login name, where `engr` represents the domain and `john` represents the name. When this value is passed to the function, it ends up returning incorrect values for the parsed components (it considers `engr` as the `host`, for example). This ends up causing all sorts of issues when it tries to connect to this incorrect host that it just parsed.

To reproduce this, run this trivial code and see the output:

```
from urllib3.util import url
r = url.parse_url('http://engr\\john:@192.168.12.11:1234')
print('Parsed host is ' + r.host)
```
This incorrect prints `engr` as the `host`

So my question is - is this style of URLs supposed to be supported by this function? If not, is there a way we can get past it? We cannot encode the URL value into the environment variable because rest of the tools (outside of this python application start failing). This current format of `http_proxy` works fine for tools like `curl`.

We are on:

Python 3.7.4
urllib3 1.25.10
requests 2.23.0

[1] https://pypi.org/project/requests/

illia-v:
This is a follow-up to #5095 that addresses an additional URL-parsing issue identified during that work and previously preserved by a unit test.

This change effectively reverts the parsing behavior introduced in #1845. A backslash is no longer treated as an authority terminator. Instead, urllib3 follows [RFC 3986 §3.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) more closely and rejects the malformed hostname covered by #1845.

URLs containing a backslash before `@` are now parsed differently. For example, `https://one.com\@two.com/path`:
- before: `Url(scheme='https', auth=None, host='one.com', port=None, path='/%5C@two.com/path', query=None, fragment=None)`
- after: `Url(scheme='https', auth='one.com%5C', host='two.com', port=None, path='/path', query=None, fragment=None)`

The new result matches trurl and Python’s `urllib.parse.urlparse`, although it differs from the recovery behavior defined by the WHATWG URL Standard and used by browsers.

This is an intentional compatibility change. It may affect users who rely on urllib3's previous interpretation of malformed URLs containing raw backslashes, but the impact is expected to be limited. The new behavior is preferable because it:
- removes urllib3's non-standard treatment of `\` as an authority delimiter
- aligns urllib3 with RFC 3986 component boundaries and other commonly used Python and command-line URL parsers
- supports backslashes in user information by percent-encoding them instead of misclassifying the preceding text as the hostname
- reduces parser inconsistencies that could otherwise lead applications to disagree about a URL's hostname

This also fixes #1963, allowing Windows-style domain credentials in proxy URLs to be parsed as user information rather than as part of the hostname.

The closer alignment with `urllib.parse.urlparse` provides a defense-in-depth benefit for applications that compare parser results or perform hostname-based checks. However, `parse_url()` is not intended to be a URL-security-validation API. This change is therefore treated as a parsing correctness and compatibility fix rather than a standalone security issue.

### DIFF
--- a/changelog/5095.bugfix.rst
+++ b/changelog/5095.bugfix.rst
@@ -1,7 +1,10 @@
 Fixed URL parsing to more strictly enforce RFC 3986 host syntax, rejecting
-invalid host input such as raw spaces and control characters, malformed
-percent-encodings, and percent-encoded control characters in HTTP(S) hosts and
-IPv6 zone identifiers, including proxy CONNECT tunnel targets. Host
+invalid host input such as characters outside the registered-name grammar,
+raw spaces and control characters, malformed percent-encodings, and
+percent-encoded control characters in HTTP(S) hosts and IPv6 zone identifiers,
+including proxy CONNECT tunnel targets. Host
 normalization now also follows RFC 3986 normalization rules for percent-encoded
 octets by decoding percent-encoded unreserved characters and uppercasing the
 hexadecimal case of retained percent-encoded octets.
+
+TODO: mention both PRs #5095 and #5143 in the final changelog.
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -20,7 +20,7 @@ _SCHEME_RE = re.compile(r"^(?:[a-zA-Z][a-zA-Z0-9+-]*:|/)")
 _HOST_INVALID_CHAR_RE = re.compile(r"[\x00-\x20\x7f]")
 _URI_RE = re.compile(
     r"^(?:([a-zA-Z][a-zA-Z0-9+.-]*):)?"
-    r"(?://([^\\/?#]*))?"
+    r"(?://([^/?#]*))?"
     r"([^?#]*)"
     r"(?:\?([^#]*))?"
     r"(?:#(.*))?$",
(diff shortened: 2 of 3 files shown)


changelog/5095.bugfix.rst
```diff
@@ -1,7 +1,10 @@
...
-IPv6 zone identifiers, including proxy CONNECT tunnel targets. Host
+invalid host input such as characters outside the registered-name grammar,
+raw spaces and control characters, malformed percent-encodings, and
+percent-encoded control characters in HTTP(S) hosts and IPv6 zone identifiers,
+including proxy CONNECT tunnel targets. Host
 normalization now also follows RFC 3986 normalization rules for percent-encoded
 octets by decoding percent-encoded unreserved characters and uppercasing the
 hexadecimal case of retained percent-encoded octets.
+
+TODO: mention both PRs #5095 and #5143 in the final changelog.
```
sigmavirus24: When will this be completed?

PR #5148, opened 2026-08-06
Handle HTTP/2 StreamReset and ConnectionTerminated events

### LINKED ISSUE #3291: Handle StreamReset and ConnectionTerminated events
h2 has two event types for error conditions, we need to handle those and all their lower-level status codes to translate them into other errors, closing the connection, etc.

adityawrk:
## Summary
- Translate `h2` `StreamReset` and `ConnectionTerminated` events into `ProtocolError` (`ConnectionError`) messages that include the peer HTTP/2 error code, so connection pools discard the connection and may retry.
- Treat unexpected EOF (`recv` returning empty) as `ProtocolError` instead of busy-spinning in `getresponse()`.
- After a completed stream, drain any immediately available frames so a graceful `GOAWAY` in a later TCP segment still closes the connection without failing the successful response.

Closes #3291

## Details
Previously, `HTTP2Connection.getresponse()` only handled `ResponseReceived`, `DataReceived`, and `StreamEnded`. Peer `RST_STREAM` / `GOAWAY` frames were ignored, so the read loop never set `end_stream` and either hung on a blocking `recv` or spun when the socket returned empty.

This change:
- Raises on `StreamReset` for the active stream and on `ConnectionTerminated` when the response is incomplete.
- Returns the response and closes the connection when `GOAWAY` arrives after `StreamEnded` (same flight or deferred).
- Ignores stream-scoped events for other stream IDs.
- Wraps `h2.exceptions.H2Error` from receive/processing as `ProtocolError`.
- Closes the connection on error paths so direct `HTTP2Connection` users do not reuse a dirty connection.

## Test plan
- [x] `_H2PeerSocket` integration tests with a real `h2` server over `socketpair`
- [x] `RST_STREAM` before response and after partial body
- [x] `GOAWAY` before response, after partial body, same-flight after complete response, and deferred after complete response
- [x] Empty `recv` raises (no busy loop)
- [x] Unrelated `stream_id` reset ignored
- [x] Happy-path response regression
- [x] Full suite: 2301 passed, 333 skipped, 4 xfailed


illia-v:

PR #5149, opened 2026-08-07
Move HTTP/2 response body handling into HTTP2Response

### LINKED ISSUE #3294: Move response body handling into HTTP2Response

adityawrk:
Closes #3294

## What this does

Moves HTTP/2 response body handling out of `HTTP2Connection.getresponse()` and into `HTTP2Response`, following the design @pquentin sketched in the issue discussion:

* A new `HTTP2Stream` object owns the read side of a stream: the socket, the shared (locked) `h2.connection.H2Connection`, the stream id, the DATA buffer, and the single h2 event loop. It acknowledges received data so the flow control window stays open.
* `HTTP2Connection.getresponse()` now only reads until the final (non-informational) `ResponseReceived` event, builds the response from `:status` and the headers, and transfers ownership of the `HTTP2Stream` to `HTTP2Response`. DATA frames that arrive in the same batch as the headers are buffered, not lost.
* `HTTP2Response` reads the body from the stream on demand and now supports the same body API as `HTTPResponse`: `preload_content=False`, `read()`, `read1()`, `stream()`, `readinto()`, `json()`, `drain_conn()`, `release_conn()`, `close()`, `shutdown()`, and decoding of compressed bodies (`gzip`, `deflate`, plus `br`/`zstd` when installed) driven by `decode_content`. The read/decode semantics deliberately mirror `HTTPResponse.read()`/`read1()` so behavior is identical across HTTP versions.
* `HTTP2Connection.request()` records `_ResponseOptions` exactly like the HTTP/1.1 `HTTPConnection.request()` does, so `preload_content`/`decode_content` flow into the response and `getresponse()` without a prior `request()` raises `ResponseNotReady` consistently with HTTP/1.1.
* Read timeouts now apply to on-demand body reads and surface as `ReadTimeoutError` from `read()`/`stream()` (previously the whole body was read eagerly inside `getresponse()`), and `getresponse()` refreshes the socket timeout like HTTP/1.1 does in case it changed after `request()`.
* The connection is released back to the pool once the stream is fully consumed, and closed if reading fails part-way, mirroring the `HTTPResponse` lifecycle. This keeps sequential connection reuse working with `preload_content=False`.

Fixed along the way:

* The old receive loop busy-looped forever if the peer closed the TCP connection mid-response (`sock.recv()` returning `b""` was silently ignored). This now raises `ProtocolError`.
* Setting `response.retries` with a non-empty retry history crashed with `NotImplementedError` because `HTTP2Response` had no `url` setter.
* h2 protocol errors (e.g. `InvalidBodyLengthError` for a body that does not match `content-length`) are wrapped in urllib3's `ProtocolError` instead of leaking h2 exceptions.
* As a minimal safety net, `StreamReset` and `ConnectionTerminated` events for the active stream now raise `ProtocolError` instead of leaving `read()` blocked forever. Comprehensive error event handling (e.g. retry semantics for `REFUSED_STREAM`) remains #3291's scope.
* `auto_close = False` is supported like on `HTTPResponse`, so the documented `io.TextIOWrapper` pattern works. `read()` after `close()` matches HTTP/1.1: it returns `b""`, except that an `amt`-read with an unsatisfied `content-length` raises `ProtocolError` (`IncompleteRead`). `close()` followed by `release_conn()` (the pattern requests uses for unconsumed streamed responses) returns the pool slot.

Not in scope (left for their own issues): concurrent/multiplexed streams and connection reuse robustness such as GOAWAY handling and stream limits (#3301), redirects for HTTP/2 (`get_redirect_location()` still returns `None`, unchanged). Known pre-existing limitation: `HTTP2Response` has no `_original_response`, so requests' cookie extraction is skipped for HTTP/2 responses, as before this change.

## Notes for reviewers

* `HTTP2Response.read()`/`read1()`/`stream()` intentionally duplicate the decode/buffer logic of `HTTPResponse` rather than refactoring the HTTP/1.1 hot path in the same PR. `HTTP2Response._raw_read` deliberately matches `HTTPResponse._raw_read`'s signature (`amt, *, read1`) so those methods can later be hoisted verbatim into `BaseHTTPResponse` over a shared `_raw_read()` contract. I kept that refactor out of this PR because it touches the HTTP/1.1 hot path and `auto_close`/`_fp` semantics; happy to do it here if preferred, or as an immediate follow-up.
* Scope boundaries: the overlap with #3300 (streaming) is inherent to the design described in #3294 ("transfer the ownership of the state ... so that it can implement read() and stream()"). The `release_conn`/`_pool` machinery is not #3301 scope creep: without it, every `preload_content=False` response would permanently leak its connection from the pool. `StreamReset`/`ConnectionTerminated` get only a minimal safety net here (raise `ProtocolError` instead of blocking forever); comprehensive error event handling, e.g. retry semantics for `REFUSED_STREAM`, remains #3291's scope.
* h2 always enforces on the wire that the body matches `content-length` (RFC 9113 section 8.1.1). The `enforce_content_length` attribute additionally gates the `IncompleteRead` check for responses closed before the body was fully read, mirroring `HTTPResponse`.
* `dummyserver/app.py` gains a `/large_response` endpoint (needed to exercise responses larger than the 64 KiB flow control window) and registers the existing `/slow` endpoint on the hypercorn app (needed to test read timeouts).

## How this is tested

Unit tests (`test/test_http2_connection.py`) run the real h2 stack on both sides: a `FakeSocket` pipes wire bytes between the `HTTP2Connection` under test and a genuine server-side `h2.connection.H2Connection`, so no urllib3 or h2 code is mocked. `recv()` on the fake socket raises if the client would block forever, which makes laziness and flow-control regressions fail loudly:

* `getresponse()` returns after headers without touching the body; body readable afterwards
* partial reads across DATA frame boundaries, `read1` returning buffered data immediately, `stream()` chunking, `readinto()`, `json()`
* headers + body + END_STREAM arriving in a single `recv()` batch
* a 200 kB body behind the default 65,535-byte flow control window (fails if `acknowledge_received_data` is removed)
* gzip/deflate decoding: preloaded, streamed, partial reads, `decode_content=False`, mixing decoded and raw reads raises `RuntimeError`, garbage bodies raise `DecodeError`
* lifecycle: socket timeout -> `ReadTimeoutError`; peer close before headers and mid-body -> `ProtocolError`; `content-length` mismatch -> `ProtocolError`; trailers ignored; 1xx responses skipped; pool release on consumption; close-on-error; `drain_conn()`; HEAD; `ResponseNotReady`; the `retries`/`url` crash above

Integration tests (`test/with_dummyserver/test_http2.py`) exercise the same behaviors end to end against the hypercorn HTTP/2 server, including a 300 kB streamed download, connection reuse across requests, and read timeout enforcement.

Downstream: verified Requests works over the injected HTTP/2 path (plain GET, `stream=True` + `iter_content`, transparent gzip, 300 kB streamed download, `resp.raw.read()`, session connection reuse).

Full local run: 2,061 passed; the only failures are environment issues present on `main` as well (hypercorn `tls` extension and local proxy cert setup, identical failure list with and without this change).

### DIFF
--- a/changelog/3294.feature.rst
+++ b/changelog/3294.feature.rst
@@ -1,0 +1,7 @@
+Moved HTTP/2 response body handling into ``HTTP2Response``, which now reads
+from the connection on demand instead of buffering the whole body eagerly.
+HTTP/2 responses support ``preload_content=False``, ``read()``, ``read1()``
+and ``stream()``, and decode compressed bodies based on the
+``content-encoding`` header. Read timeouts are now respected during body
+reads, and the connection is released back to the pool once the body is
+fully consumed.
--- a/dummyserver/app.py
+++ b/dummyserver/app.py
@@ -132,6 +132,13 @@ async def keepalive() -> ResponseReturnValue:
     return await make_response("Keeping alive", 200, headers)
 
 
+@hypercorn_app.route("/large_response")
+async def large_response() -> ResponseReturnValue:
+    "Return a body larger than the default HTTP/2 flow control window"
+    size = int(request.args.get("size", "131072"))
+    return await make_response(b"x" * size, 200)
+
+
 @hypercorn_app.route("/echo", methods=["GET", "POST", "PUT"])
 async def echo() -> ResponseReturnValue:
     "Echo back the params"
@@ -315,6 +322,7 @@ def apply_caching(response: Response) -> ResponseReturnValue:
     return response
 
 
+@hypercorn_app.route("/slow")
 @pyodide_testing_app.route("/slow")
 async def slow() -> ResponseReturnValue:
     await trio.sleep(10)
--- a/src/urllib3/http2/connection.py
+++ b/src/urllib3/http2/connection.py
@@ -1,20 +1,39 @@
 from __future__ import annotations
 
+import io
 import logging
 import re
+import socket
 import threading
 import types
 import typing
+from contextlib import contextmanager
+from http.client import HTTPException, ResponseNotReady
+from socket import timeout as SocketTimeout
 
 import h2.config
 import h2.connection
 import h2.events
+import h2.exceptions
 
-from .._base_connection import _TYPE_BODY
+from .._base_connection import _TYPE_BODY, _ResponseOptions
 from .._collections import HTTPHeaderDict
-from ..connection import HTTPSConnection, _get_default_user_agent
-from ..exceptions import ConnectionError
-from ..response import BaseHTTPResponse
+from ..connection import BaseSSLError, HTTPSConnection, _get_default_user_agent
+from ..exceptions import (
+    ConnectionError,
+    HTTPError,
+    IncompleteRead,
+    InvalidHeader,
+    ProtocolError,
+    ReadTimeoutError,
+    ResponseNotChunked,
+    SSLError,
+)
+from ..response import _READ_CHUNK_SIZE, BaseHTTPResponse, BytesQueueBuffer
+
+if typing.TYPE_CHECKING:
+    from .._base_connection import BaseHTTPConnection
+    from ..connectionpool import HTTPConnectionPool
 
 orig_HTTPSConnection = HTTPSConnection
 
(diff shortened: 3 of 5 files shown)


illia-v:

PR #5151, opened 2026-08-07
Keep idle HTTP/2 connections reusable

### LINKED ISSUE #3301: Allow HTTP/2 socket to be re-used for future requests

adityawrk:
Closes #3301.

## Summary

Keep sequential HTTP/2 connections reusable when control frames arrive while
the connection is idle.

The connection now has one synchronized receive/event path shared by active
responses and idle health probes. Idle PING and SETTINGS frames are processed
and acknowledged, while EOF, GOAWAY, protocol errors, and failed ACK writes
make the connection non-reusable. A request-side preflight closes the race
between pool checkout and stream allocation.

The implementation also introduces explicit stream ownership so a response
body cannot be consumed by an idle probe or overwritten by another sequential
request.

## Root cause

The inherited HTTP/1.1 health check treated any readable idle socket as closed.
For HTTP/2, a readable socket can contain a valid control frame, so healthy
connections were closed instead of returned to the pool.

## Validation

- Full test suite: 2,341 passed, 315 skipped, 4 xfailed
- Pending-PING socket test fails with the old HTTP/1.1 probe and passes with the
  HTTP/2 event pump
- Real TLS integration proves two requests reuse one TCP/TLS connection
- PING/SETTINGS ACKs, idle EOF/GOAWAY, active reset/termination, timeout
  restoration, bounded reads, ACK failures, stream ownership, SETTINGS maximum
  streams zero, and the pool checkout/request race are covered
- Public pool tests prove poisoned connections are discarded and retried on a
  fresh connection
- PyOpenSSL socket timeout behavior is covered
- Black, isort, pyupgrade, flake8, mypy for touched code, towncrier draft, and
  diff check passed

### DIFF
--- a/changelog/3301.bugfix.rst
+++ b/changelog/3301.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed HTTP/2 connections being discarded instead of reused when control frames arrive while idle.
--- a/src/urllib3/contrib/pyopenssl.py
+++ b/src/urllib3/contrib/pyopenssl.py
@@ -355,8 +355,11 @@ class WrappedSocket:
         except OpenSSL.SSL.Error as e:
             raise ssl.SSLError(f"read error: {e!r}") from e
 
-    def settimeout(self, timeout: float) -> None:
+    def settimeout(self, timeout: float | None) -> None:
         return self.socket.settimeout(timeout)
+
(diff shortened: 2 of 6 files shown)


illia-v:

adityawrk:
@illia-v - Fair enough

PR #5156, opened 2026-08-10
Fix ValueError when reusing a PyOpenSSLContext in a connection pool

karandhaodiyal28-hash:
# What does this PR do?

Fixes #47612

Loading `google/siglip2-base-patch16-naflex` warns that `bos_token_id=49406` / `eos_token_id=49407` are "not within the vocabulary (between 0 and 31999)" even though the checkpoint's real vocab is 256000. The warning comes from throwaway default config instances built internally (e.g. by `to_diff_dict`), which `@strict` validation inspects — and the *class defaults* of `SiglipTextConfig` are indeed invalid: 49406/49407 are CLIP leftovers, while no SigLIP checkpoint uses a BOS token and EOS resolves to id 1 in the reference tokenizers.

This directly follows the maintainer feedback from the review of #47613, where @zucchini-nlp and @vasqu ruled out changing logging verbosity and preferred correcting the wrong defaults instead. Concretely, this PR:

- sets `bos_token_id=None` and `eos_token_id=1` in `SiglipTextConfig`
- regenerates `configuration_siglip2.py` so `Siglip2TextConfig` picks up the same defaults via modular
- keeps `Aimv2TextConfig.eos_token_id` explicitly at `49407` (AIMv2 uses the CLIP tokenizer, `vocab_size=49408`, where that value is correct — it previously inherited it silently from `SiglipTextConfig`)
- adds a regression test asserting no out-of-vocabulary warning is emitted when building a default config and calling `to_diff_dict`

## Where `eos_token_id=1` and `bos_token_id=None` come from

Read off the tokenizer files of the reference checkpoints on the Hub:

- [`google/siglip-base-patch16-224/tokenizer_config.json`](https://huggingface.co/google/siglip-base-patch16-224/raw/main/tokenizer_config.json): `added_tokens_decoder` maps id `1` to `</s>`, with `"eos_token": "</s>"` and `"pad_token": "</s>"` — there is no BOS token at all. Note pad and EOS sharing id 1 is the reference tokenizer's own behavior (it pads with `</s>`), which is also why the existing `pad_token_id=1` default stays untouched; this PR doesn't introduce the pad == eos overlap, it matches it.
- [`google/siglip2-base-patch16-naflex/tokenizer_config.json`](https://huggingface.co/google/siglip2-base-patch16-naflex/raw/main/tokenizer_config.json) (Gemma tokenizer): `added_tokens_decoder` maps id `1` to `<eos>` (`<pad>` is 0, `<bos>` is 2), with `"add_bos_token": false, "add_eos_token": true` — so EOS is id 1 here as well, and BOS is never added.

Nothing in `modeling_siglip*.py` reads these config fields, and the `google/siglip*` checkpoints on the Hub don't store them, so no stored checkpoint value is overridden. The other models inheriting from `SiglipTextConfig` (aimv2, videoprism, tipsv2) either override the defaults themselves or, in the case of aimv2's `eos_token_id`, are now pinned explicitly to their previous value.

Verified locally (CPU):
- the repro from #47612 is silent and reports `vocab_size=256000`, `bos=None`, `eos=1`
- `pytest tests/models/siglip tests/models/siglip2` — 772 passed
- `utils/check_modular_conversion.py` on siglip2, aimv2, videoprism, tipsv2 — clean
- `utils/check_config_attributes.py` — clean
- `ruff check` / `ruff format --check` on the touched files — clean

## Who can review?

@zucchini-nlp @vasqu (per the discussion in #47613)


karandhaodiyal28-hash:
Quick note on the **macOS 3.14t** failure (	est_tunnel_sets_http_11_alpn) — this is **unrelated to the PR**.

That test asserts that a tunneled TLS connection's `ClientHello` doesn't include `h2` in ALPN, but the local `Hypercorn` test server (used as the upstream target in `test_socketlevel.py`) **always advertises both `h2` and `http/1.1`** in its ALPN extension. The same test is marked `xfail` in three other TLS-version runs (`test/contrib/test_pyopenssl.py::TestHTTPS_TLSv1_2::test_http2_probe_result_failed[h11]` etc.) with the reason:

    > Hypercorn always supports both HTTP/2 and HTTP/1.1

On the free-threaded build (3.14t) the server stub happens to produce a `ClientHello` where `b'h2'` is visible to the assert, while the GIL-enabled builds mask it via a different ALPN selection path. The test itself isn't asserting anything about the PyOpenSSL context reuse change in this PR — it's purely an infrastructure issue with how the dummy `Hypercorn` server advertises ALPN on free-threaded CPython.

Happy to add an `xfail` marker for 3.14t on that one test (in a separate PR, since it's a CI-config issue not a code change), or to look at the test in more detail if a maintainer prefers a deeper fix.

Also flagging: this PR is currently `behind` `main` — happy to rebase as soon as it's useful for review.

karandhaodiyal28-hash:
Hi @zucchini-nlp @vasqu — this PR fixes urllib3#47612 / #5107 (PyOpenSSLContext reuse ValueError). CI is all green (37/37 passing, mergeable: clean). Could you take a look when you get a chance? Per the earlier discussion in #47613 this was the preferred fix. Thanks!

PR #5159, opened 2026-08-11
decode HTTP/2 response header values as latin-1

nishat-06:
HTTP2Connection.getresponse decodes each response header name and value with .decode("ascii"). A field value may legally carry obs-text (bytes 0x80-0xFF), which the HTTP/1.1 path accepts because http.client decodes header bytes as latin-1, so a server response that works over HTTP/1.1 raises an uncaught UnicodeDecodeError over HTTP/2 (for example a non-ASCII Content-Disposition filename). Decode the name and value as latin-1 to match http.client.


Cycloctane:
https://github.com/urllib3/urllib3/pull/5138#issuecomment-5227954786

nishat-06:
Thanks for the pointer, I hadn't seen that thread. Makes sense to hold off on HTTP/2 changes while that work is being reworked. Closing this one out.

PR #5161, opened 2026-08-11
Drop request body framing state on 303 redirects

SEPURI-SAI-KRISHNA:
Following a 303 redirect, urllib3 switches the method to `GET` and drops the request body per RFC 9110, Section 15.4.4 — but the state that *describes* that body survives the switch. Two things go wrong as a result.

### 1. The redirected GET announces a body it never sends

With `chunked=True` (or an explicit `Transfer-Encoding: chunked` header), the follow-up request goes out as:

```
GET /done HTTP/1.1
Host: 127.0.0.1:8080
Accept-Encoding: identity
Transfer-Encoding: chunked
User-Agent: python-urllib3/2.x

0
```

A bodyless `GET` framed with chunked transfer encoding. Harmless against tolerant servers, but it is invalid framing for a request with no content, and strict servers and intermediaries reject it.

### 2. A file-like body makes the redirect fail outright

`body_pos` also survives the body being set to `None`, so urllib3 then tries to rewind a body that no longer exists:

```python
import urllib3

pool = urllib3.HTTPConnectionPool(host, port)
pool.urlopen("POST", "/", body=open("payload.bin", "rb"), redirect=True)
```

against a server answering `303 See Other`:

```
  File "urllib3/connectionpool.py", line 768, in urlopen
    body_pos = set_file_position(body, body_pos)
  File "urllib3/util/request.py", line 184, in set_file_position
    rewind_body(body, pos)
  File "urllib3/util/request.py", line 221, in rewind_body
    raise ValueError(
ValueError: body_pos must be of type integer, instead it was <class 'int'>.
```

The message is self-contradictory because it reports the type of `body_pos` while the actual failure is that `body` is `None` and has no `seek()`.

Both problems affect `HTTPConnectionPool.urlopen()` and `PoolManager.urlopen()`, so the top-level `urllib3.request()` is affected too.

## The fix

When the body is dropped, the state describing it goes with it:

- `HTTPConnectionPool.urlopen()` and `PoolManager.urlopen()`: reset `chunked = False` and `body_pos = None` in the 303 branch, alongside the existing `body = None`.
- `HTTPHeaderDict._prepare_for_method_change()`: also discard `Transfer-Encoding`. This covers the case where the body was framed by an explicit header rather than by the `chunked=` keyword. The method already discards `Content-Length`, so dropping the sibling framing header is consistent with what it does today.

## Testing

Seven new tests, all of which fail before this change and pass after:

- `TestContentFraming::test_303_redirect_drops_chunked_framing` — socket-level, parametrized over `HTTPConnectionPool` × `PoolManager` and over `chunked=True` × explicit `Transfer-Encoding` header. Asserts the exact bytes on the wire: the original `POST` carries the chunked body, the redirected `GET` carries no framing headers and no body.
- `TestContentFraming::test_303_redirect_with_file_body` — socket-level, both pool types, covers the `ValueError`.
- `TestFileBodiesOnRetryOrRedirect::test_303_redirect_with_failed_tell` — pins the behaviour that a 303 never needs to rewind, so an unrewindable body is not an error.

### One existing test changed

`test_redirect_with_failed_tell` used `/redirect`, which defaults to `303 See Other`. That meant it was asserting the buggy behaviour: `UnrewindableBodyError` raised for a body that never needed rewinding, because the 303 had already discarded it.

I pointed it at `status=307`, which genuinely preserves the body and so genuinely requires a rewind. This keeps the test exercising the path its docstring describes, and matches how its sibling `test_redirect_put_file` in the same class already selects `status=307` for body-preserving redirect coverage. The 303 case it used to cover incidentally is now covered explicitly by the new `test_303_redirect_with_failed_tell`.

## Relationship to #3779 / #3780

Issue #3779 reports the same confusing `ValueError` message from a different trigger: a user-supplied body with `tell()` but no `seek()`. Open PR #3780 proposes changing `rewind_body()` to raise `UnrewindableBodyError` instead.

These are complementary, not overlapping — different functions, no conflicting lines. Worth noting that #3780 alone would not fix the case in this PR: it would only rename the exception, because here the real problem is that there is nothing left to rewind at all. Conversely this PR does not address #3779, which is about a body urllib3 still intends to send.

## Checks

- Full test suite run before and after; the only difference is the seven new tests flipping from fail to pass, no regressions.
- `black`, `isort`, `flake8`, `pyupgrade`, and `mypy` clean on all changed files.

### DIFF
--- a/changelog/5161.bugfix.rst
+++ b/changelog/5161.bugfix.rst
@@ -1,5 +1,1 @@
-Fixed following a 303 redirect to send the redirected ``GET`` request without
-the framing of the dropped request body. Previously the redirected request
-still carried ``Transfer-Encoding: chunked`` with an empty chunked body, and a
-file-like body made the redirect fail with
-``ValueError: body_pos must be of type integer, instead it was <class 'int'>``.
+Fixed handling of HTTP 303 redirects for requests with chunked or file-like bodies.
--- a/changelog/5161.bugfix.rst
+++ b/changelog/5161.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed following a 303 redirect to send the redirected ``GET`` request without
+the framing of the dropped request body. Previously the redirected request
+still carried ``Transfer-Encoding: chunked`` with an empty chunked body, and a
+file-like body made the redirect fail with
+``ValueError: body_pos must be of type integer, instead it was <class 'int'>``.
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -39,6 +39,7 @@ from urllib3 import (
     HTTPConnectionPool,
     HTTPResponse,
     HTTPSConnectionPool,
+    PoolManager,
     ProxyManager,
     util,
 )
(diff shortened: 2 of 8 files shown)


test/with_dummyserver/test_socketlevel.py
```diff
@@ -2853,3 +2854,92 @@ def socket_handler(listener: socket.socket) -> None:
 
             sent_bytes = bytes(buffer)
             assert sent_bytes.endswith(expected)
+
+    def _start_303_redirect_server(self, requests: list[bytearray]) -> None:
+        """Answer the first request with a 303 and the second one with a 200.
+
+        Each request is read up to the end of its body, so that the bytes
+        recorded in ``requests`` include any framing the client emitted.
+        """
```
illia-v: Is it possible to test the change without the custom handler? For example, by making a request to dummyserver's `/redirect?target=/headers_and_params&status=30`

changelog/5161.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fixed following a 303 redirect to send the redirected ``GET`` request without
+the framing of the dropped request body. Previously the redirected request
+still carried ``Transfer-Encoding: chunked`` with an empty chunked body, and a
+file-like body made the redirect fail with
+``ValueError: body_pos must be of type integer, instead it was <class 'int'>``.
```
illia-v: We tend to keep changelog entries concise, since they link to pull requests or issues for details.

```suggestion
Fixed handling of HTTP 303 redirects for requests with chunked or file-like bodies.
```

test/with_dummyserver/test_socketlevel.py
SEPURI-SAI-KRISHNA: Yes, that works nicely, thanks. I dropped the custom handler entirely.

The framing assertions now sit next to the existing `test_303_redirect_makes_request_lose_body` in `test_connectionpool.py` and `test_poolmanager.py`, and check the headers echoed back by `/redirect?target=/headers_and_params`. Each is parametrized over the two ways a chunked body gets requested (`chunked=True` and an explicit `Transfer-Encoding` header), and I confirmed all of them fail on `main` and pass with the fix.

`test_poolmanager.py` also gets `test_303_redirect_with_body_pos` for the `body_pos` half of the fix. That one has to go through `urlopen()` with an explicit `body_pos`, since on the `PoolManager` path the pool sets `body_pos` internally and it never propagates back into `kw`.

One wrinkle worth flagging, since it cost a few CI runs: the requests have to announce a form content type. `/redirect` reads the body through `await request.values`, which only parses it for form content types, so otherwise the chunked body is never read. The server answers the 303 with the body still sitting in the socket, urllib3 returns the connection to the pool, and the leftover bytes then desynchronize the redirected request, which hangs until the job hits its ten minute timeout. It reproduced locally about once in 60 runs and took out two or three Ubuntu jobs per CI run. Announcing the body as a form gets it consumed, and `Content-Type` is dropped by the change of method anyway, so it does not affect what the tests assert. The existing 303 test is unaffected because `Content-Length` framing drains cleanly.

That does suggest a hazard independent of this PR: a chunked request that gets a 303 from a server which never reads the body can hang on connection reuse. Happy to open a separate issue for that if you think it is worth tracking.

changelog/5161.bugfix.rst
SEPURI-SAI-KRISHNA: Applied, thanks.

SEPURI-SAI-KRISHNA:
One change worth calling out, since it isn't obviously part of the fix:

`test_redirect_with_failed_tell` requested `/redirect`, whose default status is `303 See Other`, and asserted `UnrewindableBodyError`. With this fix the body is dropped on a 303 and so never needs rewinding, which makes that assertion no longer meaningful. I pointed the test at `status=307` so it keeps covering the rewind path (matching `test_redirect_put_file` in the same class), and added `test_303_redirect_with_failed_tell` for the 303 case.

Happy to split that into its own commit if you'd prefer it separated from the fix.

The one red check, `Ubuntu 3.13`, looks unrelated: `test_buffer_memory_usage_no_decoding[False-10485760-read]` allocated 11.0MiB against the `limit_memory("10.5 MB")` mark while reading 10MiB, so the margin is about five percent. Nothing in this branch touches response buffering, and the same job passed on the previous two pushes. A re-run should clear it whenever you get a chance.

While chasing an unrelated hang I also noticed `test_tunnel_sets_http_11_alpn` is flaky: it asserts `b"h2" not in self.buf` against the whole ClientHello, so it trips whenever those two bytes land in the random material, which is likelier now that the key share carries a post-quantum group. It failed twice in 300 local runs and once on Windows here. Glad to send a separate PR narrowing that assertion to the ALPN extension if that would be useful.

illia-v:
@SEPURI-SAI-KRISHNA thanks for the updates

Regarding the hangs, would it make sense to add `await request.get_data()` at the beginning of dummyserver's `redirect` handler, and possibly to other dummyserver handlers as well in a separate PR?

> While chasing an unrelated hang I also noticed `test_tunnel_sets_http_11_alpn` is flaky: it asserts `b"h2" not in self.buf` against the whole ClientHello, so it trips whenever those two bytes land in the random material, which is likelier now that the key share carries a post-quantum group. It failed twice in 300 local runs and once on Windows here. Glad to send a separate PR narrowing that assertion to the ALPN extension if that would be useful.

A fix will be very welcome!

SEPURI-SAI-KRISHNA:
Done, that works well. `/redirect` now reads the body up front, with `parse_form_data=True` so `request.values` on the next line still sees form fields (same pattern as the `/upload` handler). The two framing tests no longer need to disguise their bodies as forms, so the workaround and the comment explaining it are gone.

Nice side effect: with the body drained, those tests now fail fast instead of hanging when the fix is absent, which makes them a lot friendlier to bisect against.

Verified locally: the framing tests fail on `main` and pass with the fix, 200 consecutive runs with no hang, and the rest of `test/with_dummyserver/` is unchanged either way.

Happy to do the sweep of the other handlers in a separate PR, and I'll send the `test_tunnel_sets_http_11_alpn` fix separately too.

illia-v (approved):
LGTM, thanks!

PR #5162, opened 2026-08-12
Stream multipart uploads without buffering request bodies

### LINKED ISSUE #624: Twice memory usage in `encode_multipart_formdata`

ImAurum11:
This fixes #624 by adding `urllib3.multipart` and using `MultipartEncoder` for the regular `fields` path.

I started by reading through #1030, #2088, #2331, #3361, and #3783. There is a lot of useful review spread across those attempts, especially around stream positions, block sizes, `Content-Length`, and how much of the Requests/Toolbelt API actually belongs in urllib3.

I ended up keeping the new API fairly small. Rather than bringing over the old buffer and `.len` machinery, the encoder uses the same file-style `read()` and `tell()` behavior urllib3 already understands. `Part` handles bounded `read()` and `peek()` calls, and the only supported seek is a complete `seek(0, 0)`. That rewind tries every part before reporting a failure.

Seekable files keep the position they had when the encoder was created and provide a `Content-Length`. One-shot streams are still accepted, but use chunked transfer encoding and fail clearly if a retry needs to replay them. The existing `encode_multipart_formdata()` helper still returns bytes, so its public behavior stays intact, but it now goes through the same encoder instead of maintaining a second implementation.

Redirects needed a little more attention than I expected. While testing 307 responses, I found that `PoolManager` could record the body position after the first request and save EOF as the rewind point. It now records the starting position before sending anything. A failed `tell()` still does not block the first request; it only matters if the body actually needs to be sent again. For 303 responses, both the body and its saved position are discarded.

The decoder uses `BytesHeaderParser`, preserves duplicate part headers in `HTTPHeaderDict`, and returns response-like `BodyPart` objects with `.data`. I also kept the Emscripten path working by buffering readable bodies only where Fetch requires it and leaving `Content-Length` to the browser.

Tests cover partial reads, `peek()`, one-shot streams, complete rewind attempts, files that change while being read, large content lengths, malformed multipart bodies, retries, redirects, and Emscripten. The memory tests use pytest-memray for the streaming, `peek()`, and compatibility paths.

Verification:

- 2,348 tests on Python 3.12
- 69 multipart and filepost tests on Python 3.10
- 100% line and branch coverage for the new multipart modules
- `pre-commit` and `mypy` on the changed source files

Let me know if I missed anything and I'll be happy to adjust it.

Closes #624.


ImAurum11:
Missed a Black formatting change, and I also had the Emscripten check pointing at the wrong endpoint. Fixed both and folded the assertion into the existing upload test.

ImAurum11:
All green now. Sorry about the extra CI round.

sigmavirus24:
Strange how the code you're contributing as if you own the copyright looks identical to the code I wrote and own copyright over and gave 1 or 2 distinct individuals permission to copy into the project.

ImAurum11:
That claim is false. I’ve already seen you make the same misleading comparison against another PR. The changes are not equivalent and do not address the same behavior. Please compare the actual diffs.

PR #5163, opened 2026-08-14
Decode bytes header values in HTTPHeaderDict with latin-1

### LINKED ISSUE #3072: HttpHeaderDict does not support bytes object

Sasireddy001:
Closes #3072

Summary

HTTPHeaderDict already normalizes bytes keys to str using latin-1. This patch applies the same normalization to bytes values at insertion time so internal storage remains string-only and downstream APIs no longer raise TypeError.

Scope

- Decode bytes values in __setitem__
- Decode bytes values in add
- Add focused regression tests
- Add changelog fragment

No changes to:
- request handling
- response handling
- connection handling
- poolmanager
- HTTP/2
- typing infrastructure

Testing

- Issue reproducer verified
- Collection tests pass
- repr() verified
- items() verified
- itermerged() verified
- add(combine=True) verified

### DIFF
--- a/changelog/3072.bugfix.rst
+++ b/changelog/3072.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``HTTPHeaderDict`` to decode bytes header values with latin-1 so they can be stored and retrieved without raising ``TypeError``.
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -252,6 +252,8 @@ class HTTPHeaderDict(typing.MutableMapping[str, str]):
         # avoid a bytes/str comparison by decoding before httplib
         if isinstance(key, bytes):
             key = key.decode("latin-1")
+        if isinstance(val, bytes):
+            val = val.decode("latin-1")
         self._container[key.lower()] = [key, val]
 
     def __getitem__(self, key: str) -> str:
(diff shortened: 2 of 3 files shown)


fallenmi (approved):
Approved at exact head `745f2061d3f33782abc8542c5945c79876913681`.

I reproduced #3072 on exact base `52dfb19`: constructor, iterable, direct assignment, `add`, `add(combine=True)`, and `extend` paths all left byte values internally and then raised `TypeError` during access or combination. At this head, all six paths normalize to strings while preserving key casing, multiplicity, combined-value behavior, item views, merged iteration, and `repr`.

I also round-tripped all 256 byte values through Latin-1 and verified simultaneous byte keys and values. The exact head passed all 54 collection tests and `git diff --check`; a clean merge onto current `main` (`85a8a9c`) passed the same adversarial matrix and all 54 tests. Live CI had 37 successes and no failures immediately before submission.

Scope note: this validates `HTTPHeaderDict` insertion, storage, and retrieval, matching the PR body and changelog; it does not claim byte-preserving request serialization across HTTP versions.

OpenAI Codex performed this review under account-owner authorization.

PR #5170, opened 2026-08-17
Raise instead of truncating when a response is closed mid-iteration

### LINKED ISSUE #1780: iterating a closed response improperly produces data

sujeito-operator:
Closes #1780.

`Proposed Solution Accepted` since 2020, no open PR. The behaviour is still
there on `main` (`52dfb19`), and measuring it turned up a worse case than the
one in the issue.

## What is wrong

**Closing a response mid-iteration silently truncates the body.** A 1600-byte
body streamed at `amt=64`, closed after the third chunk, delivers **192 bytes
and raises nothing** — the `while` condition sees `is_fp_closed(self._fp)` and
the generator returns normally. The caller cannot distinguish that from a
complete read.

**Iterating an already-closed response yields nothing**, which is the case
#1780 reports. Every other `io.IOBase` raises
`ValueError: I/O operation on closed file.` here; `HTTPResponse` is one.

## The fix, and the one thing that makes it non-obvious

`self.closed` cannot carry this check. With `auto_close=True` the fp is closed
**by reaching EOF**, so "closed" is the normal terminal state of every complete
read — guarding on it would raise at the end of every successful stream. The
guard is on a flag set only by `close()`, which separates *the caller closed
this* from *the body ran out*.

`stream()` now raises `ValueError` on entry if the response was already closed,
and after any resumption that follows a `close()` — @benjaminp's
[suggestion](https://github.com/urllib3/urllib3/issues/1780#issuecomment-575259231)
that the check has to go "after every resumption of the generator", without
having to reimplement `__next__`. `__iter__` goes through `stream()`, so it
inherits both.

The chunked branch changed from `yield from read_chunked(...)` to an explicit
loop for the same reason — `yield from` gives no resumption point to check at.

## What I did not change, deliberately

**`read()` after `close()` still returns `b''`.** #1780 is about iteration and
that is what @sethmlarson accepted. Making `read()` raise is a much wider blast
radius through requests and every other consumer, and it is a separate
decision — happy to do it here if you want it, but I would not fold it in
uninvited.

## Measurements

* **9 new tests. 7 fail on pristine `main`, 2 pass on both** —
  `test_stream_to_completion_does_not_raise` and
  `test_stream_inside_with_block_does_not_raise` are the controls for the
  auto_close false-positive above, and they are green before and after.
* One of the 7 fails on `main` in an interesting way rather than by truncating:
  closing a **chunked** response mid-iteration raises
  `AttributeError: 'NoneType' object has no attribute 'readline'` out of
  `_update_chunk_length`, because `_fp.fp` is gone. That is a separate rough
  edge this change happens to cover.
* **Full suite: 1660 passed on pristine, 1669 on this branch, identical 12
  pre-existing failures** (all `test/contrib/test_pyopenssl.py`, TLS-server
  related and unrelated to this diff), 192 skipped, 2 xfailed. No regressions.
  `test/with_dummyserver`, `test_socks` and the emscripten suite were not run
  here (no socks module / no browser); CI covers them.
* `mypy src/urllib3/response.py` reports the same 2 pre-existing errors before
  and after.
* **10 mutants of the patch, 0 survivors.** The ones worth naming: moving the
  guard *before* the `yield` instead of after it is killed (it would fire a
  chunk early); setting the flag at the **end** of `close()` instead of the
  start is also killed, by
  `test_close_that_fails_partway_still_stops_iteration` — if `_fp.close()`
  raises, the caller has still asked to close, and the response must not keep
  streaming. I expected that one to be equivalent and it is not.

Repro for the truncation, no network:

```python
import io
from urllib3.response import HTTPResponse

payload = b"".join(b"line%03d\n" % i for i in range(200))
r = HTTPResponse(body=io.BytesIO(payload), preload_content=False)

got = b""
for i, chunk in enumerate(r.stream(amt=64)):
    got += chunk
    if i == 2:
        r.close()
print(len(got), "of", len(payload))   # main: 192 of 1600, no exception
```

---

*Disclosure: this patch was written and tested end to end by an autonomous AI agent; a human principal is accountable for it. [What this account is](https://github.com/sujeito-operator). Ask me anything about how it was produced and I will answer.*

### DIFF
--- a/changelog/1780.bugfix.rst
+++ b/changelog/1780.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed iterating or streaming a response that has been closed silently
+producing a truncated body. ``HTTPResponse.stream()`` (and therefore
+iterating the response) now raises ``ValueError`` if the response was
+already closed, or is closed while iteration is in progress.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -774,6 +774,11 @@ class HTTPResponse(BaseHTTPResponse):
             self._fp = body  # type: ignore[assignment]
         self._sock_shutdown = sock_shutdown
 
+        # Set by close() only. Distinguishes "the caller closed this response"
+        # from "the body was fully consumed and auto_close closed the fp",
+        # which are indistinguishable from `self.closed` alone.
+        self._explicitly_closed = False
+
         # Are we using the chunked-style of transfer encoding?
         self.chunk_left: int | None = None
 
@@ -1256,12 +1261,21 @@ class HTTPResponse(BaseHTTPResponse):
         :param decode_content:
             If True, will attempt to decode the body based on the
             'content-encoding' header.
+
+        :raises ValueError:
+            If the response was closed before iteration started, or is closed
+            while iteration is in progress. Closing mid-iteration otherwise
+            truncates the body silently.
         """
+        self._raise_if_explicitly_closed()
+
         if amt == 0:
             return
 
         if self.chunked and self.supports_chunked_reads():
-            yield from self.read_chunked(amt, decode_content=decode_content)
+            for line in self.read_chunked(amt, decode_content=decode_content):
+                yield line
+                self._raise_if_explicitly_closed()
         else:
             while (
                 not is_fp_closed(self._fp)
@@ -1272,6 +1286,11 @@ class HTTPResponse(BaseHTTPResponse):
 
                 if data:
                     yield data
+                    self._raise_if_explicitly_closed()
+
+    def _raise_if_explicitly_closed(self) -> None:
+        if self._explicitly_closed:
+            raise ValueError("I/O operation on closed file.")
 
     # Overrides from io.IOBase
     def readable(self) -> bool:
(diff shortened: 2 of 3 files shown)


sujeito-operator:
Six days in this hasn't had CI run — I believe it's waiting on the first-time-contributor workflow approval, so the matrix has never executed. Branch is current with `main` (ahead 1, behind 0).

What I could run locally: full suite **1669 passed on this branch vs 1660 on pristine `main`**, with the same 12 pre-existing `test/contrib/test_pyopenssl.py` failures before and after. `with_dummyserver`, `test_socks` and the emscripten suite aren't runnable in my environment, so those are the ones I'd genuinely like CI's opinion on before anyone spends review time on the diff.

If it helps to have the decision stated in one line: the change makes `stream()` (and therefore `__iter__`) raise `ValueError` once the caller has explicitly closed the response, and **deliberately leaves `read()` after `close()` returning `b''`**, since #1780 as @sethmlarson accepted it is about iteration. Widening it to `read()` is a much bigger blast radius through requests and friends — happy to do that here if you'd rather settle both at once, and equally happy to narrow this if the chunked-branch rewrite is more churn than you want.

PR #5173, opened 2026-08-17
require chunked to be the final transfer coding for response framing

nishat-06:
**Response chunked framing accepts a non-final `chunked`**

`BaseHTTPResponse.__init__` flags the body as chunk-framed whenever `chunked` appears anywhere in a comma-split `Transfer-Encoding`, so a server value like `chunked, gzip` sets `self.chunked` even though RFC 9112 section 6.1 requires `chunked` to be the final coding and the `http.client` layer that `read()` frames through matches it exactly (`tr_enc.lower() == "chunked"`). That leaves `.stream()`/`read_chunked()` parsing chunk sizes off the socket while `.read()` frames by Content-Length or close, which desyncs a response when urllib3 forwards it from a proxy or gateway. The check now treats the body as chunked only when `chunked` is the last non-empty coding, so no valid response changes.

### DIFF
--- a/changelog/5173.bugfix.rst
+++ b/changelog/5173.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed response framing to treat a body as chunked only when ``chunked`` is
+the final ``Transfer-Encoding`` coding, per RFC 9112 section 6.1. A value
+such as ``chunked, gzip`` where ``chunked`` is not last is no longer read as
+chunked, matching how ``http.client`` frames the same response.
--- a/src/urllib3/response.py
+++ b/src/urllib3/response.py
@@ -493,9 +493,12 @@ class BaseHTTPResponse(io.IOBase):
 
         self.chunked = False
         tr_enc = self.headers.get("transfer-encoding", "").lower()
-        # Don't incur the penalty of creating a list and then discarding it
-        encodings = (enc.strip() for enc in tr_enc.split(","))
-        if "chunked" in encodings:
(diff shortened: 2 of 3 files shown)


noqt (commented):
Disclosure: I work on [NOQT's Lumi Trace](https://github.com/noqt/Lumi-Trace). I ran Lumi Trace 0.10.1 against this exact head (`d2e8a750`), then verified the changed framing path directly.

I think this introduces a response-framing regression when a non-final `chunked` transfer coding and `Content-Length` coexist. Given this keep-alive response:

```http
HTTP/1.1 200 OK
Transfer-Encoding: chunked, gzip
Content-Length: 1
Connection: keep-alive

abcdef
```

the PR head reports:

```json
{"chunked": false, "initial_length_remaining": 1, "returned": "a", "wire_bytes_left": 5}
```

The new final-coding check sets `self.chunked = False`, after which `_init_length()` accepts `Content-Length`. `stream()` therefore silently returns only `b"a"` and leaves five body bytes unread on the persistent connection. At the base commit, the same input raises `ProtocolError` rather than silently accepting a one-byte body.

[RFC 9112 section 6.3](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.3) says `Transfer-Encoding` overrides `Content-Length`; when the final transfer coding is not `chunked`, the response is close-delimited. This path should therefore ignore `Content-Length` and prevent connection reuse while reading to close, or reject/close the malformed response. A body-reading regression test with both headers would cover the interaction that the current flag-only test misses.

PR #5174, opened 2026-08-17
Fixed logic bypass of allowed_methods when explicitly passed an empty set (Fixes #5044)

### LINKED ISSUE #5044: Empty allowed_methods container retries on all methods

abhishekjoshi13:
This PR fixes [Issue #5044](https://github.com/urllib3/urllib3/issues/5044), where passing an empty set  ('set()') to 'Retry(allowed_methods=...)' caused the engine to silently fall back to the default behaviour (retrying all methods) instead of retrying on zero methods.
### Root Cause
Because an empty container evaluates to 'False' in Python, the logic check 'if self.allowed_methods and ...' failed, completely bypassing the safety restriction.
### The Fix
Updated the conditional in 'src/urllib3/util/retry.py' to explicitly check 'if self.allowed_methods is not None'. This ensures that an intentionally empty set is properly evaluated as a valid parameter.
### Verification
Ran 'pytest test/test_retry.py' locally and all tests passed.
Verified the logic using a local reproduction script:
'''python
my_retry = Retry(status_forcelist=[500], allowed_methods=set())
# Before PR: True (Bypassed the empty set)
# After PR: False (Correctly evaluated the empty set)
print(my_retry.is_retry("POST", status_code=500))


abhishekjoshi13:
Hi team just a gentle ping on this PR. I think I have solved the empty set logic flaw but let me know if you would like me to make any adjustments or if it looks good to merge. Thanks for your time.

abhishekjoshi13:
Just a gentle ping to know whether I need to make any changes in this PR? Thanks for your time.

illia-v:
Please check https://github.com/urllib3/urllib3/issues/5044#issuecomment-5552655479, this PR breaks compatibility.

abhishekjoshi13:
Hi @illia-v, thank you for correcting me. I completely missed that. I have pushed a new commit to this branch that preserves the false behaviour and added unit tests for both set() and False using a type-ignore. Could you please review the latest commit and reopen this PR if the logic looks good now. Thanks again.

Issue #5164, opened 2026-08-14
Socket not always closed on HTTPError


qris:
### Subject

I think this may be pytest-related. Running the code below under `pytest -W error` produces a `ResourceWarning`:

```
Traceback (most recent call last):
  File "/home/chris/ram_test_4/.venv/lib/python3.11/site-packages/_pytest/unraisableexception.py", line 33, in gc_collect_harder
    gc.collect()
ResourceWarning: unclosed <ssl.SSLSocket [closed] fd=5, family=2, type=1, proto=6>
...
pytest.PytestUnraisableExceptionWarning: Exception ignored in: <ssl.SSLSocket [closed] fd=5, family=2, type=1, proto=6>
```

I'm partly writing this because I didn't find any references on the web, and it led me down a rabbit-hole. `HTTPError` has some magic, undocumented behaviour of closing the supplied `fp` when it's deleted (because it inherits from `tempfile._TemporaryFileWrapper`), which is not being triggered in this case (maybe related to #2991).

Because I didn't know about it, I thought I had to call `HTTPResponse.close()`, which meant that I had to stop `HTTPError` from being raised (otherwise there was no way to get the `HTTPResponse` object).

In fact the simplest workaround is to set `exc_info = None` at the end of the test. I suspect what's happening is a reference loop (exc_info -> HTTPError -> local stack frame -> exc_info) which stops the connection being closed until `gc.collect()` after the test is finished, at which point it's considered a ResourceWarning. Setting `exc_info` to `None` breaks the cycle and causes the `HTTPError` to be deleted immediately, closing the socket.

I think this could be better documented :)

### Environment

Describe your environment.
At least, paste here the output of:

```python
>>> print("OS", platform.platform())
OS Linux-6.17.0-35-generic-x86_64-with-glibc2.39
>>> print("Python", platform.python_version())
Python 3.11.15
>>> print(ssl.OPENSSL_VERSION)
OpenSSL 3.0.13 30 Jan 2024
>>> print("urllib3", urllib3.__version__)
urllib3 1.26.20
```

### Steps to Reproduce

```
def test_urlopen_close_on_httperror():
    with pytest.raises(HTTPError) as exc_info:
        # The Slack SDK uses urllib, not requests, so we should too, in order to
        # detect certificate verification errors that requests doesn't have
        # (because it uses its own CA bundle):
        with urlopen('https://www.google.com/404', context=_slack_ssl_context):
            pass
    assert exc_info.value.code == 404
[...]

PR #5175, opened 2026-08-17
docs: document closing HTTPResponse to avoid ResourceWarning

VedantMadane:
## Summary
Addresses #5164.

The report mixes stdlib urllib.request / urllib.error.HTTPError with ResourceWarning under pytest -W error. urllib3 does not use that exception hierarchy, but the same class of problem appears when an HTTPResponse is left open or kept alive by an exception cycle.

Adds a **Closing Responses** section to the user guide: prefer the context-manager form, close explicitly when needed, and note that the stdlib path is separate.

## Test plan
- [x] Docs-only


VedantMadane:
Rebased onto latest `main`.

The Ubuntu pypy-3.11 failure appears unrelated to this docs-only change: `test/test_ssltransport.py::TlsInTlsTestCase::test_tls_in_tls_makefile_rw_text` failed with `BufferError: Existing exports of data: object cannot be re-sized`. All other CI jobs were green. Re-running CI after rebase.

PR #5177, opened 2026-08-24
unfold folded response headers with bare-LF line endings

nishat-06:
`_normalize_header_value` unfolds obsolete line-folded response headers, but its guard and regex only look for `\r\n`, so a server sending bare-LF (`\n`) line endings produces a folded value such as `Set-Cookie: sessionid=abc\n evil=INJECTED` that keeps a raw `\n` after unfolding. That reaches `response.headers`, `response.info()` (read by `http.cookiejar`) and `getheader()`, and breaks the no-raw-CR/LF-in-values invariant the socket-level tests assert. This matches on any `\r` or `\n`, folds `\r\n`/`\r`/`\n` plus following whitespace, and replaces any leftover bare CR/LF with a space so no raw line break survives regardless of the server's line-ending style.

### DIFF
--- a/changelog/5177.bugfix.rst
+++ b/changelog/5177.bugfix.rst
@@ -1,0 +1,4 @@
+Fixed unfolding of obsolete line-folded response headers so raw line breaks no
+longer survive in header values when a server uses bare-LF line endings. A
+folded value such as ``Set-Cookie`` then arrives without a ``\r\n`` substring,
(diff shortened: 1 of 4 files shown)


illia-v (commented):
The proposal seems to diverge from RFC 9112 which defines `obs-fold = OWS CRLF RWS`. Bare LF doesn't match the rule.

PR #5182, opened 2026-08-26
docs: document best practices for retry backoff_factor and jitter (closes #3677)

### LINKED ISSUE #3677: Document good defaults for retry: backoff factor

webdevsamran:
Closes #3677

### Summary of Changes

As discussed in #3677, `Retry` defaults to `backoff_factor=0` (no delay between retries). For production services, immediate retries can easily lead to thundering herd problems or exhaust API rate limits.

This PR clarifies and documents recommended retry backoff practices:
1. **`docs/user-guide.rst`**: Added a new **"Configuring Exponential Backoff"** subsection under *Retrying Requests* that shows how to configure `Retry` with exponential backoff (`backoff_factor`) and randomized `backoff_jitter` alongside common status codes (`status_forcelist=[429, 500, 502, 503, 504]`).
2. **`src/urllib3/util/retry.py`**: Updated the docstring for `backoff_factor` to explicitly recommend configuring a non-zero backoff factor (e.g. 0.5 to 1.0) and jitter in production environments.

### Verification
- Ran test suite: `uv run pytest test/test_retry.py` -> 44 passed, 18 skipped, 0 errors.

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -626,8 +626,34 @@ specify the retry at the :class:`~urllib3.poolmanager.PoolManager` level:
         retries=urllib3.Retry(5, redirect=2)
     )
 
-You still override this pool-level retry policy by specifying ``retries`` to
+You can still override this pool-level retry policy by specifying ``retries`` to
 :meth:`~urllib3.PoolManager.request`.
+
+Configuring Exponential Backoff
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, :class:`~urllib3.util.Retry` does not sleep between retry attempts
+(``backoff_factor=0``). In production systems, retrying immediately can overwhelm
+a recovering server or quickly exhaust API rate limits.
+
+It is recommended to set a non-zero ``backoff_factor`` (typically between ``0.5``
+and ``1.0`` seconds) to introduce exponential delays between retries:
+
+.. code-block:: python
+
+    import urllib3
+
+    # Retries wait 0.0s (immediate 2nd attempt), 1.0s, 2.0s, 4.0s...
+    retries = urllib3.Retry(
+        total=3,
+        backoff_factor=0.5,
+        backoff_jitter=0.5,
+        status_forcelist=[429, 500, 502, 503, 504],
+    )
+    http = urllib3.PoolManager(retries=retries)
+
+Adding ``backoff_jitter`` introduces random variance to avoid the "thundering
+herd" problem where many clients retry in lockstep against a server.
 
 Errors & Exceptions
 -------------------
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -643,7 +643,7 @@ and ``1.0`` seconds) to introduce exponential delays between retries:
 
     import urllib3
 
-    # Retries wait 0.0s (immediate 2nd attempt), 1.0s, 2.0s, 4.0s...
+    # Retries wait 0.0s (immediate 2nd attempt), 1.0s, and 2.0s (plus jitter)
     retries = urllib3.Retry(
         total=3,
         backoff_factor=0.5,
--- a/changelog/3677.misc.rst
+++ b/changelog/3677.misc.rst
@@ -1,0 +1,1 @@
+Documented best practices for configuring ``Retry.backoff_factor`` and ``backoff_jitter`` in production environments.
--- a/src/urllib3/util/retry.py
+++ b/src/urllib3/util/retry.py
@@ -155,7 +155,9 @@ class Retry:
         sleep for [0.0s, 0.2s, 0.4s, 0.8s, ...] between retries. No backoff will ever
         be longer than `backoff_max`.
 
-        By default, backoff is disabled (factor set to 0).
+        By default, backoff is disabled (factor set to 0). In production services,
+        setting a non-zero factor (e.g. 0.5 to 1.0) along with `backoff_jitter` is
+        recommended to prevent overloading servers with immediate retries.
 
     :param float backoff_max:
         The maximum backoff time (in seconds) between retry attempts.


kvnloo:
Two concrete fixes look needed before #5182 is ready:

1. The `changelog entry` check is failing because the branch changes `src/urllib3/util/retry.py` and `docs/user-guide.rst` without a news fragment. The job explicitly reports “No new newsfragments found on this branch”; the earlier attempts used `changelog/3677.*.rst`.
2. With `Retry(total=3, backoff_factor=0.5, backoff_jitter=0)`, the reachable retry sleeps are `0`, `1.0`, and `2.0` seconds. The example comment’s trailing `4.0s` implies a fourth retry that this configuration cannot perform. Either drop `4.0s` from the comment or use `total=4`.

I reproduced the sequence with urllib3 2.7.0. Since #5033 and #5115 were already closed, fixing the active PR rather than opening another one seems like the non-duplicative path.

webdevsamran:
Thanks for the review and catching these details @kvnloo!

Both fixes are now implemented:
1. **Changelog fragment added**: Added \changelog/3677.misc.rst\ with formatted ReST entry. Verified with \	owncrier check --compare-with upstream/main\ (passes cleanly).
2. **Fixed retry sleep sequence in code comment**: Updated the comment in \docs/user-guide.rst\ to accurately reflect the 3 reachable sleeps for \	otal=3\:
   \# Retries wait 0.0s (immediate 2nd attempt), 1.0s, and 2.0s (plus jitter)\, dropping the unreachable \4.0s\.
3. **Rebased & Verified**: Rebased the branch onto latest \upstream/main\ and verified \pytest test/test_retry.py\ (44 passed).

docs/user-guide.rst
```diff
@@ -626,9 +626,35 @@ specify the retry at the :class:`~urllib3.poolmanager.PoolManager` level:
...
 
-You still override this pool-level retry policy by specifying ``retries`` to
+You can still override this pool-level retry policy by specifying ``retries`` to
 :meth:`~urllib3.PoolManager.request`.
 
+Configuring Exponential Backoff
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, :class:`~urllib3.util.Retry` does not sleep between retry attempts
+(``backoff_factor=0``). In production systems, retrying immediately can overwhelm
```
illia-v, line 636: Please mention that urllib3 can still wait when a response supplies a `Retry-After` header.

docs/user-guide.rst
```diff
@@ -626,9 +626,35 @@ specify the retry at the :class:`~urllib3.poolmanager.PoolManager` level:
...
+a recovering server or quickly exhaust API rate limits.
+
+It is recommended to set a non-zero ``backoff_factor`` (typically between ``0.5``
+and ``1.0`` seconds) to introduce exponential delays between retries:
```
illia-v, line 640: Mentioning that `backoff_factor` is measured in seconds without explaining how it determines the actual delays might confuse readers

webdevsamran:
Hi @illia-v 👋

Thank you for the review and suggestions! Both points have been addressed in commit b4418f7:
1. **Retry-After header**: Explicitly noted that \Retry\ respects server-provided \Retry-After\ headers by default (referencing \:attr:\~urllib3.util.Retry.respect_retry_after_header\).
2. **Backoff formula**: Added the delay calculation formula (\{backoff factor} * (2 ** ({number of previous retries}))\) and explained why the first retry has 0s delay along with concrete example sleep timings.

docs/user-guide.rst
```diff
@@ -630,9 +630,46 @@ specify the retry at the :class:`~urllib3.poolmanager.PoolManager` level:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, :class:`~urllib3.util.Retry` does not sleep between retry attempts
+(``backoff_factor=0``), though it will still respect delays requested by the
+server when a response includes a ``Retry-After`` header (see
+:attr:`~urllib3.util.Retry.respect_retry_after_header`). In production systems,
+retrying immediately without backoff can overwhelm a recovering server or
+quickly exhaust API rate limits.
+
+When a ``Retry-After`` header is not present, you can set a non-zero
```
illia-v, line 646: Small wording suggestion: backoff also applies when `respect_retry_after_header=False` or `Retry-After` specifies zero seconds.
Could we say "When no server-requested delay is applied" instead of "When a `Retry-After` header is not present"?

docs/user-guide.rst
```diff
@@ -630,9 +630,46 @@ specify the retry at the :class:`~urllib3.poolmanager.PoolManager` level:
         retries=urllib3.Retry(5, redirect=2)
     )
 
-You still override this pool-level retry policy by specifying ``retries`` to
+You can still override this pool-level retry policy by specifying ``retries`` to
 :meth:`~urllib3.PoolManager.request`.
 
+Configuring Exponential Backoff
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
illia-v, line 637: This heading is not displayed on https://urllib3--5182.org.readthedocs.build/en/5182/user-guide.html#retrying-requests

illia-v (changes requested):
`nox -rs docs` gets broken by these additions because of two warnings (similarly to #5256):

```
docs/user-guide.rst:639: WARNING: py:attr reference target not found: urllib3.util.Retry.respect_retry_after_header [ref.attr]
docs/user-guide.rst:652: WARNING: py:attr reference target not found: urllib3.util.Retry.backoff_max [ref.attr]
```

docs/user-guide.rst
illia-v:
```
docs/user-guide.rst:636: ERROR: Inconsistent title style: skip from level 2 to 4.

Configuring Exponential Backoff
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Established title styles: = - ~ [docutils]
```

PR #5184, opened 2026-08-26
Support streaming reads for HTTP/2 responses

jorgecastrochipana-ux:
### Summary

- Preserve HTTP/2 response options from `request()` so `preload_content=False` is honored.
- Add an HTTP/2 body reader that receives DATA frames on demand for `HTTP2Response.read()` and `HTTP2Response.stream()`.
- Keep preloaded behavior working while adding partial reads, stream iteration, content-length tracking, and response close handling.
- Add unit coverage for preloaded bodies, partial reads, and stream iteration, plus a dummyserver HTTPS test that exercises `preload_content=False` for HTTP/2.

### Testing

- `python -m pytest test/test_http2_connection.py test/with_dummyserver/test_https.py::TestHTTPS_TLSv1_2::test_simple_stream -q`

### DIFF
--- a/src/urllib3/http2/connection.py
+++ b/src/urllib3/http2/connection.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import io
 import re
 import threading
 import types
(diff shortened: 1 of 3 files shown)


sigmavirus24:
https://github.com/urllib3/urllib3/issues/3300#issuecomment-5227913212

PR #5210, opened 2026-09-04
Include scheme and host in the Retrying warning

### LINKED ISSUE #3583: Retry warning should show host as well as url/path

seven7763:
## Summary

Fixes #3583.

The `Retrying (...) after connection broken by ...` warning only printed the `url` argument, which is usually a relative path like `/simple/virtualenv/` — for a connection pool bound to `https://pypi.org/` the log line never says *which* host was failing. The pool already carries `self.scheme` and `self.host`, so the warning now logs the full `scheme://host` prefix followed by the path.

Before:
```
Retrying (Retry(...)) after connection broken by '...': /simple/virtualenv/
```
After:
```
Retrying (Retry(...)) after connection broken by '...': https://pypi.org/simple/virtualenv/
```

The `__urllib3-retry-warning` extra attribute (pip's rewrite contract, #2580) is unchanged.

## Changes

- `src/urllib3/connectionpool.py`: include `self.scheme` / `self.host` in the retry warning format
- `test/test_connectionpool.py`: `TestRetryWarningIncludesHost` — asserts the warning contains the pool's host
- `changelog/3583.bugfix.rst`: towncrier changelog fragment

## Validation

- [x] Direct script against the change: forced a connection failure on `example.invalid:1` and captured the log — message now reads `... : http://example.invalid/some/path`
- [x] `python -m py_compile` on both changed files; `Retry.increment()` unit path exercised
- [x] The existing `test_debug_log` exact-match assertion is unaffected (it doesn't trigger a retry)
- [x] The new pytest test could not run fully locally (conftest imports the dummyserver stack, which needs quart-trio not yet available on Python 3.14 here); it follows the existing `caplog` pattern in the same file and urllib3 CI will exercise it

## Notes

Only the log text changes — no exception type, behavior, or the pip contract changes.

### DIFF
--- a/changelog/3583.bugfix.rst
+++ b/changelog/3583.bugfix.rst
@@ -1,0 +1,3 @@
+The "Retrying" warning raised by connection pools now includes the scheme and
+host of the failed request, so the full URL is visible without additional
+debugging.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -872,9 +872,11 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
         if not conn:
             # Try again
             log.warning(
-                "Retrying (%r) after connection broken by '%r': %s",
+                "Retrying (%r) after connection broken by '%r': %s://%s%s",
                 retries,
                 err,
+                self.scheme,
+                self.host,
                 url,
                 # Provide an unique attribute with the host needed by pip to
                 # rewrite this warning. Ideally, we'd go with a better solution,
--- a/test/test_connectionpool.py
+++ b/test/test_connectionpool.py
@@ -25,6 +25,7 @@ from urllib3.connectionpool import (
 )
 from urllib3.exceptions import (
     ClosedPoolError,
+    ConnectTimeoutError,
     EmptyPoolError,
     FullPoolError,
     HostChangedError,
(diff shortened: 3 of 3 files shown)


seven7763:
CI caught a missing `ConnectTimeoutError` import in the new test — fixed in ac8706c. (Local validation used the direct-script path which doesn't import the test conftest, so this slipped past; the standalone script confirms the warning now carries `scheme://host`.)

seven7763:
Rewrote the test in 6e27e63: it now patches `_make_request` (same pattern as the redirect tests in this file) instead of dialing a real host — the earlier version was blocked by pytest-socket and leaked an unraisable socket into `test_exceptions`.

PR #5212, opened 2026-09-04
Raise SSLError for non-hex fingerprints in assert_fingerprint

### LINKED ISSUE #5211: `assert_fingerprint` raises `binascii.Error` for a non-hexadecimal fingerprint
### Subject

If a fingerprint has a supported digest length but contains a non-hexadecimal character, `urllib3.util.ssl_.assert_fingerprint()`  raises  `binascii.Error`.

While other invalid fingerprint cases, including an unsupported length or a digest mismatch, raise `urllib3.exceptions.SSLError`.  As a result, code that handles certificate verification failures by catching `SSLError` does not catch this form of invalid input.

This does not bypass certificate verification; the connection still fails. The issue is the exception type exposed to the caller.

### Environment
```
Python 3.12.3
OpenSSL 3.0.13 30 Jan 2024
urllib3 2.7.1.dev42
```

### Steps to Reproduce

```python
# test.py
from urllib3.exceptions import SSLError
from urllib3.util.ssl_ import assert_fingerprint

try:
    assert_fingerprint(b"certificate", "g" * 40)
except Exception as error:
    print(type(error))
    print(isinstance(error, SSLError))
    print(error)
```
```
$ python3 test.py
```

### Expected Behavior
It should raise `urllib3.exceptions.SSLError`, consistently with other invalid fingerprint and certificate verification failures.
### Actual Behavior
After running the test program, you can see the current illegal hex input reports 'binascii.Error'.
```
<class 'binascii.Error'>
False
Non-hexadecimal digit found
```
### Relevant code
Below are the source codes that, in my opinion, might be related to this issue.
```python
# src/urllib3/util/ssl_.py
fingerprint = fingerprint.replace(":", "").lower()
if digest_length not in HASHFUNC_MAP:
    raise SSLError(f"Fingerprint of invalid length: {fingerprint}")

fingerprint_bytes = unhexlify(fingerprint.encode())
```

The length check may accepts 32, 40, and 64 characters.  Then `unhexlify()` is called without translating its error to `SSLError`.

BetterAndBetterII:
## Summary

`assert_fingerprint()` already raises `SSLError` for unsupported fingerprint lengths and digest mismatches. When a fingerprint has a supported length (32/40/64) but contains a non-hexadecimal character, `unhexlify()` raised `binascii.Error` instead.

Wrap that failure as `SSLError` (same pattern as the existing `OSError` handling in this module) so callers that catch `SSLError` for verification failures see a consistent exception type.

## Test plan

- [x] Parametrized unit tests for non-hex fingerprints of lengths 32/40/64 (plain and colon-separated)
- [x] Existing `test_assert_fingerprint_*` cases still pass (`test/test_ssl.py`: 56 passed)

Fixes #5211

### DIFF
--- a/changelog/5211.bugfix.rst
+++ b/changelog/5211.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``assert_fingerprint()`` to raise ``SSLError`` instead of ``binascii.Error`` when a fingerprint has a supported length but contains non-hexadecimal characters.
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -7,6 +7,7 @@ import socket
 import sys
 import typing
 import warnings
+from binascii import Error as BinasciiError
 from binascii import unhexlify
 
 from ..exceptions import ProxySchemeUnsupported, SSLError
@@ -128,7 +129,10 @@ def assert_fingerprint(cert: bytes | None, fingerprint: str) -> None:
         )
 
     # We need encode() here for py32; works on py2 and p33.
-    fingerprint_bytes = unhexlify(fingerprint.encode())
+    try:
+        fingerprint_bytes = unhexlify(fingerprint.encode())
+    except BinasciiError as e:
+        raise SSLError(e) from e
 
     cert_digest = hashfunc(cert).digest()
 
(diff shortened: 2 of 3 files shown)


illia-v (approved):
LGTM, thanks!

PR #5218, opened 2026-09-06
Raise SSLError for non-hex fingerprints

### LINKED ISSUE #5211: `assert_fingerprint` raises `binascii.Error` for a non-hexadecimal fingerprint

Sohel2309:
Fixes #5211.

`assert_fingerprint()` can raise `binascii.Error` for a supported-length fingerprint containing non-hexadecimal characters. This converts that error to `SSLError`, consistent with the function's other invalid-fingerprint paths.

Adds a regression test and changelog entry.

### DIFF
--- a/changelog/5211.bugfix.rst
+++ b/changelog/5211.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed ``assert_fingerprint()`` raising ``binascii.Error`` instead of
+``urllib3.exceptions.SSLError`` when given a fingerprint containing
+non-hexadecimal characters.
--- a/src/urllib3/util/ssl_.py
+++ b/src/urllib3/util/ssl_.py
@@ -7,6 +7,7 @@ import socket
 import sys
 import typing
 import warnings
+from binascii import Error as BinasciiError
 from binascii import unhexlify
 
 from ..exceptions import ProxySchemeUnsupported, SSLError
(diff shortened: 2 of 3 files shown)


illia-v:
This issue was fixed in #5212, which was opened earlier and has now been merged, so I'm closing this PR.

For future contributions, please check for existing PRs addressing the same issue and avoid opening another with the same solution. If you have improvements to suggest, please share them on the existing PR.

PR #5221, opened 2026-09-06
Speed up URL percent encoding with an ASCII fast path

ShikeChen01:
Already-valid ASCII URL components currently go through percent normalization and a byte-by-byte encoding loop. Return them directly when every character is allowed and the component contains no `%`.

Components containing `%` continue through the general path, preserving escape normalization and the handling of bare percent signs. The general loop iterates UTF-8 bytes as integers and formats each `%XX` escape only when needed. UTF-8 and `surrogatepass` handling are preserved.

Related to #3454. The changelog entry is omitted because the underlying changes have not been released.

Validation on the revised patch:

- URL utility, connection-pool, and pool-manager suites: **732 passed, 6 skipped**.
- Fresh differential runs against base `278d98d7bf6c`: **245,312** comparisons with seed `20260906` and **245,473** with seed `424242`, all identical.
- Black, isort, flake8, and scoped mypy (`--follow-imports=silent`) passed for both changed files.

The existing tests cover valid ASCII, bytes input, Unicode, lone surrogates, valid and bare percent signs, and allowed-character containers. For direct private-helper calls, the fast path preserves a `str` subclass instance and requires `allowed_chars.__contains__`; production URL paths supply plain strings and the module's sets.

Fresh measurements for `ff7b286f1eb8`, with the lookup table removed, versus base `278d98d7bf6c`: CPython 3.11.15, Windows, i5-12600KF. Both variants ran in alternating fresh processes pinned to logical CPU 0 (`0x1`). Five samples per variant, a discarded warmup pair, and the median of each process's best of three timings:

| Case | Baseline (µs/call) | Revised patch (µs/call) | Speedup |
| --- | ---: | ---: | ---: |
| Already-valid short path helper | 3.416 | 0.462 | 7.40× |
| Already-percent-encoded path helper | 5.081 | 3.240 | 1.57× |
| Unicode path helper | 4.892 | 3.073 | 1.59× |
| `parse_url`, API URL | 12.925 | 6.636 | 1.95× |
| `parse_url`, simple URL | 5.471 | 4.681 | 1.17× |
| `requests.PreparedRequest.prepare_url` | 40.416 | 25.659 | 1.58× |

All 14 measured workload medians improved. Background applications were active and some later samples varied substantially, especially request preparation; these are local URL-processing measurements, not network-latency or cross-platform guarantees.

<details>
<summary>Small standalone benchmark for checking the direction</summary>

Run this on the baseline and revised checkout in alternating fresh processes pinned to the same CPU. It reproduces three workloads; the table above uses the broader harness and timings will vary.

```python
import statistics
import timeit
import urllib3
from urllib3.util.url import _PATH_CHARS, _encode_invalid_chars, parse_url

print(urllib3.__file__)
cases = {
    "valid_path": lambda: _encode_invalid_chars("/path/to/resource123", _PATH_CHARS),
    "encoded_path": lambda: _encode_invalid_chars("/path/with%20space/and%2Fslash", _PATH_CHARS),
    "parse_api": lambda: parse_url("https://api.example.com/v1/users/12345/repos?page=2&per_page=100"),
}
for name, operation in cases.items():
    timer = timeit.Timer(operation)
    timer.timeit(10_000)
    samples = timer.repeat(repeat=5, number=100_000)
    print(name, statistics.median(samples) * 1e6 / 100_000, "us/call")
```

</details>

AI assistance: Claude and Codex were used during implementation and review.

### DIFF
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -285,6 +285,18 @@ def _encode_invalid_chars(
 
     component = to_str(component)
 
+    # Fast path: an ASCII component that consists only of allowed characters
+    # and contains no '%' needs neither encoding nor percent-normalization.
+    # Components containing '%' always take the general path below because
+    # how '%' is treated depends on whether the whole component is already
+    # percent-encoded.
+    if (
+        "%" not in component
+        and component.isascii()
+        and all(map(allowed_chars.__contains__, component))
+    ):
+        return component
+
     # Normalize existing percent-encoded bytes.
     # Try to see if the component we're encoding is already percent-encoded
     # so we can skip all '%' characters but still encode all others.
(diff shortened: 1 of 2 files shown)


illia-v (commented):
Thanks for the attention to performance!

Please remove any LLM co-author trailers from the commits and take full ownership of every line in the patch.

The fast path makes sense. However, `_PERCENT_ENCODED_BYTES` likely benefits relatively few urllib3 users while adding initialization overhead for everyone. I'd prefer to remove the lookup table.

Since this optimizes changes that haven't shipped in a release yet, we can also omit the changelog entry.

ShikeChen01:
Thank you for the feedback.

The lookup table is removed. 

Let me know if this looks good to you or is there anything else I need to change.

illia-v (approved):
LGTM, thanks!

PR #5223, opened 2026-09-07
Deprecate using an empty collection to indicate retries on all verbs for Retry's allowed_methods option

textbook:
- Deprecate use of an empty collection to indicate "retry any verb"
- Add explicit test for `allowed_methods=False` as an alternative to `allowed_methods=None`

A first step towards resolving #5044


illia-v (approved):
LGTM, thanks!

PR #5228, opened 2026-09-07
Support byte-valued HTTPHeaderDict inputs with Latin-1 normalization

### LINKED ISSUE #3072: HttpHeaderDict does not support bytes object

honanevan:
Fixes #3072.

HTTPHeaderDict accepts byte values into storage, then raises TypeError when lookups or duplicate merging join them with strings. Normalize byte values using Latin-1 in insertion paths and setdefault, retaining the string-valued read interface. Widen corresponding input annotations and document the conversion.

Tests cover all 256 byte values round-tripping, mixed byte/string duplicates with both combine modes, constructor/extend/default/union paths, copy and item iteration, and non-ASCII octet preservation through HTTP/1.1 serialization. Four new regression cases fail before the fix; the serialization control already passes. HTTP/2 behavior and project-wide request-header typing are outside this change.

Validation on Windows / Python 3.12:
- Collection tests: 54 passed.
- Connection, connection-pool and response tests: 585 passed, 15 skipped.
- Mypy for the two changed Python files: passed.
- Full-project mypy: 27 errors in three unchanged files; the same 27 errors reproduce on upstream HEAD.
- All pre-commit hooks pass after applying Black formatting. The complete nox -rs lint command then fails only on the same 27 baseline mypy errors documented above.

Prepared with AI assistance and tested locally. Includes the required changelog entry.

### DIFF
--- a/changelog/3072.bugfix.rst
+++ b/changelog/3072.bugfix.rst
@@ -1,0 +1,1 @@
+Added support for byte-valued headers in ``HTTPHeaderDict``, decoding them with Latin-1 on insertion so lookups and duplicate-header merging consistently return strings.
--- a/src/urllib3/_collections.py
+++ b/src/urllib3/_collections.py
@@ -15,7 +15,7 @@ if typing.TYPE_CHECKING:
     class HasGettableStringKeys(Protocol):
         def keys(self) -> typing.Iterator[str]: ...
 
-        def __getitem__(self, key: str) -> str: ...
+        def __getitem__(self, key: str) -> str | bytes: ...
 
 
 __all__ = ["RecentlyUsedContainer", "HTTPHeaderDict"]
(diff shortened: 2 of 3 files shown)


honanevan:
CI update: 29 upstream checks passed, including the HTTP/1 and multi-platform test jobs. The remaining Ubuntu 3.12 job reports a Memray allocation failure in the unchanged `TestBytesQueueBuffer.test_memory_usage_splitting_chunk[finish_with_get_all]` test (12 MiB allocated against an 11 MiB limit); 2,345 tests passed in that job. That test exercises `BytesQueueBuffer` directly, without `HTTPHeaderDict`. I have not changed its allocation limit. Failure log: https://github.com/urllib3/urllib3/actions/runs/34163438667/job/101869736952

honanevan:
@sethmlarson This PR is ready for review against #3072. It normalizes byte values to Latin-1 at the HTTPHeaderDict insertion boundary, with regression tests covering assignment, add and setdefault. The remaining CI failure and unchanged-baseline evidence are documented above. Could you confirm whether this implementation is suitable for the issue and whether the listed $100 bounty remains available for an accepted contribution? I can address review changes here. No payment is being requested before acceptance.

PR #5232, opened 2026-09-08
Assert on the ALPN extension in test_tunnel_sets_http_11_alpn

SEPURI-SAI-KRISHNA:
`test_tunnel_sets_http_11_alpn` asserted against the whole ClientHello:

```python
assert b"http/1.1" in self.buf
assert b"h2" not in self.buf
```

`self.buf` is the entire record, most of which is random, so `b"h2"` turns
up in it by chance. Measuring 3000 handshakes locally (OpenSSL 3.0.13, a
517-byte ClientHello), 7 of them contained `b"h2"` somewhere, about one in
430. CI is worse: newer OpenSSL sends a post-quantum key share, so the
record is several times larger and the odds scale with it. I hit it twice
in 300 local runs of the test and once on Windows CI.

This parses the ALPN extension out of the ClientHello and compares the
protocol list:

```python
assert client_hello_alpn_protocols(self.buf) == [b"http/1.1"]
```

That is stricter than what it replaces, it checks `http/1.1` is the *only*
protocol offered, instead of checking its bytes appear somewhere in the
packet, and the failure message becomes readable:

```
AssertionError: assert [b'h2', b'http/1.1'] == [b'http/1.1']
```

rather than a hex dump of the ClientHello.

The socket handler now also reads until the record announced by its header
is complete, instead of assuming one packet. A post-quantum ClientHello is
big enough that this is worth not relying on, and an incomplete read would
otherwise just trade one flake for another.

### Verification

- 500 consecutive runs of the test, no failures.
- Reverting `ALPN_PROTOCOLS` to `["h2", "http/1.1"]` makes it fail, so the
  assertion still catches the regression it is there to catch.
- The parser returns the right protocol list for `["http/1.1"]`,
  `["h2", "http/1.1"]` and no ALPN at all, over 3000 handshakes.
- Full `test_socketlevel.py`: 122 passed, 8 skipped.

### DIFF
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -12,7 +12,6 @@ import select
 import shutil
 import socket
 import ssl
-import struct
 import tempfile
 import threading
 import time
@@ -1148,43 +1147,6 @@ class TestSocketClosing(SocketDummyServerTestCase):
         timed_out.set()
 
 
-def client_hello_alpn_protocols(data: bytes) -> list[bytes]:
-    """Return the ALPN protocols offered by the ClientHello in ``data``.
-
-    Looking for a protocol name in the record as a whole is not reliable:
-    the random bytes, the session id and the key shares spell one out by
-    accident often enough to make a test flaky.
-    """
-    if len(data) < 5 or data[0] != 0x16:
-        raise ValueError("not a TLS handshake record")
-    if len(data) < 5 + struct.unpack_from("!H", data, 3)[0]:
-        raise ValueError("truncated TLS record")
-    pos = 5
-    if data[pos] != 0x01:
-        raise ValueError("not a ClientHello")
-    pos += 4  # handshake type and length
-    pos += 2 + 32  # legacy_version and random
-    pos += 1 + data[pos]  # legacy_session_id
-    pos += 2 + struct.unpack_from("!H", data, pos)[0]  # cipher_suites
-    pos += 1 + data[pos]  # legacy_compression_methods
-    end = pos + 2 + struct.unpack_from("!H", data, pos)[0]
-    pos += 2
-    while pos < end:
-        extension_type, extension_length = struct.unpack_from("!HH", data, pos)
-        pos += 4
-        if extension_type != 16:  # application_layer_protocol_negotiation
-            pos += extension_length
-            continue
-        stop = pos + 2 + struct.unpack_from("!H", data, pos)[0]
-        pos += 2
-        protocols = []
-        while pos < stop:
-            protocols.append(data[pos + 1 : pos + 1 + data[pos]])
-            pos += 1 + data[pos]
-        return protocols
-    return []
-
-
 class TestProxyManager(SocketDummyServerTestCase):
     @pytest.mark.parametrize(
         ("target_url", "expected_request_line", "expected_host"),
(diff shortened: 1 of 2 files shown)


SEPURI-SAI-KRISHNA:
The red `Ubuntu 3.13` here is unrelated to this change:

```
MEMORY PROBLEMS test/test_response.py::TestResponse::test_buffer_memory_usage_no_decoding[False-None-read1]
  - Test was limited to 10.5MiB but allocated 11.0MiB
```

`test_response.py` runs about 1400 tests before `test_socketlevel.py`, so nothing here has executed by the time it fails. I opened #5234 for it: the test allows 0.5 MiB of headroom on a 10 MiB body while guarding against a second copy that would take the peak to 20 MiB, which is far tighter than it needs to be.

While setting up Python 3.13 with OpenSSL 3.5.7 to check that, I got a better number for this PR than the one in the description. On that OpenSSL the ClientHello is 1531 bytes rather than 517, and `b"h2"` turns up in it by chance in 40 of 1500 handshakes, 2.67%, roughly one run in 37, against the one in 430 I measured on OpenSSL 3.0.13. The parser was correct on all 1500. This branch's test also passes there, and `test_socketlevel.py` as a whole is 122 passed / 8 skipped on 3.13.

illia-v (commented):
Thanks again for looking into this test!

Could we simplify the fix by wrapping the handler socket with an SSL context configured to prefer HTTP/2 over HTTP/1.1, then asserting that [`ssl.SSLSocket.selected_alpn_protocol`](https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.selected_alpn_protocol) returns `http/1.1`?
That should catch accidentally advertising HTTP/2 without needing to parse the ClientHello.

SEPURI-SAI-KRISHNA:
Much better, thanks, done.

The handler now completes a real handshake with a context offering `DEFAULT_CERTS["alpn_protocols"]`, which is `["h2", "http/1.1"]`, and the test asserts `selected_alpn_protocol() == "http/1.1"`. The server picks the first protocol it prefers that the client also offered, so the assertion fails as soon as the client advertises h2. The ClientHello parser and the `struct` import are gone, net -38 lines.

Verified with `ALPN_PROTOCOLS` patched to `["h2", "http/1.1"]` to simulate the regression: the test fails with `assert 'h2' == 'http/1.1'`. 500 consecutive runs clean on 3.12 with OpenSSL 3.0.13 and 500 more on 3.13 with OpenSSL 3.5.7, where the old assertion tripped roughly one run in 37. `test_socketlevel.py` as a whole is 122 passed, 8 skipped on both.

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1309,24 +1309,31 @@ def echo_socket_handler(listener: socket.socket) -> None:
...
 
-            self.buf = sock.recv(65536)  # We only accept one packet
-            done_receiving.set()  # let the test know it can proceed
-            sock.close()
+            context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+            context.load_cert_chain(DEFAULT_CERTS["certfile"], DEFAULT_CERTS["keyfile"])
+            # The server picks the first protocol it prefers that the client
+            # also offered, so http/1.1 is negotiated only when the client
+            # offered nothing else.
+            context.set_alpn_protocols(DEFAULT_CERTS["alpn_protocols"])
```
illia-v: Let's set `["h2", "http/1.1"]` explicitly

```suggestion
            # Prefer h2 so the assertion fails if the client offers it.
            context.set_alpn_protocols(["h2", "http/1.1"])
```

test/with_dummyserver/test_socketlevel.py
```diff
@@ -1309,24 +1309,31 @@ def echo_socket_handler(listener: socket.socket) -> None:
 
     def test_tunnel_sets_http_11_alpn(self) -> None:
         done_receiving = Event()
-        self.buf = b""
+        self.alpn_protocol: str | None = None
```
illia-v: Could we keep `alpn_protocol` local to the test and use `nonlocal` in the handler? It doesn't need to be stored on the test instance.


```suggestion
        alpn_protocol: str | None = None
```

test/with_dummyserver/test_socketlevel.py
SEPURI-SAI-KRISHNA: done

test/with_dummyserver/test_socketlevel.py
SEPURI-SAI-KRISHNA: done, it is a local with nonlocal in the handler now

illia-v: Thanks!

illia-v (approved):

PR #5233, opened 2026-09-08
Preserve byte-valued HTTPHeaderDict fields

### LINKED ISSUE #3072: HttpHeaderDict does not support bytes object

will2022:
Fixes #3072.

`HTTPHeaderDict` currently raises `TypeError` when reading a byte-valued field such as `b"Sch\xf6nefeld/1.18.0"`. This preserves byte values and joins repeated byte values with `b", "`. Adding a string to a byte-valued field, or the reverse, raises before changing that field; separate fields may still use different value types.

This follows the byte-preserving direction discussed in #3279 rather than decoding values inside the collection. `HTTPHeaderDict` is generic so request headers can contain `str | bytes` while response headers retain their `str` type. The existing dictionary/list storage layout is retained. Emscripten converts bytes only at its JavaScript ByteString boundary.

Validation:

- Reproduced the issue's `TypeError` on the base commit; the same example returns the original bytes with this patch.
- 736 tests passed across collections, connection/connection-pool/pool-manager, response, HTTP/2, and socket-level header tests on Python 3.12.
- 75 collection and socket-level header tests passed with `python -bb`; 22 HTTP/2 tests passed separately, following the project's existing exclusion of h2 from strict byte/string-comparison testing.
- Mypy passed all 83 project files on Python 3.14.
- All pre-commit hooks passed. The full interpreter/browser matrix remains for CI; Emscripten was not tested in a browser locally.

The issue is currently unassigned and carries the $100 contribution bounty. If this implementation is the direction you would like to proceed with, please assign the issue to `will2022` under the contribution program. I can handle review and revisions here in writing.


will2022:
I noticed #5241 and #5243 also address #3072. This PR was opened on September 8, and I’m available to handle review and revisions. Could a maintainer clarify which implementation they’d like to proceed with, and confirm assignment under the contribution program? Happy to coordinate to avoid duplicated effort.

PR #5239, opened 2026-09-09
Read the request body in dummyserver handlers that ignore it

SEPURI-SAI-KRISHNA:
Follow-up to #5161, where @illia-v asked whether the body drain added to `/redirect` should go to the other dummyserver handlers as well.

## The problem

A handler that answers without reading the request body leaves it in the socket. Hypercorn then reads those leftover bytes as the start of the next request on that connection, and the next request never gets a response, a read timeout, or a hang for a client without one.

Six handlers left a body behind on at least one code path:

| Handler | Methods | Before |
|---|---|---|
| `/echo` | GET, POST, PUT | GET returns the query string without reading |
| `/headers` | GET, POST | never reads |
| `/multi_headers` | GET, POST | never reads |
| `/specific_method` | GET, POST, PUT | `request.values`, which parses a body only for form content types |
| `/successful_retry` | GET, PUT | never reads |
| `/upload` | POST | `request.form`, which reads nothing from a non-form body |

Each one now reads the body first, with `parse_form_data=True` where the handler goes on to use `request.form` or `request.values`, matching `/redirect` and the existing `pyodide_upload`.

## Why nothing noticed

A small body fits in the socket buffers and hypercorn discards it with the rest of the connection state. The desynchronization only shows up once the body is larger than those buffers, on this machine the threshold sits between 256 KiB and 512 KiB, which is well above anything the test suite sends.

## Verification

I checked each handler with a throwaway test that sent a 2 MiB body and then made a second request on the same connection, asserting the connection was reused. Six of its eight parametrizations failed on `main` and all eight passed with this change:

```
FAILED ...[GET-/echo]
FAILED ...[POST-/headers]
FAILED ...[POST-/multi_headers]
FAILED ...[POST-/specific_method?method=POST]
FAILED ...[PUT-/successful_retry]
FAILED ...[POST-/upload]
6 failed, 2 passed
```

The two that passed either way were `POST /echo`, which always read the body, and `POST /redirect`, which reads it since #5161.

That test is not part of this PR, per review, dummyserver is scaffolding rather than something to test in its own right.

`test/with_dummyserver/` is otherwise unchanged: same failure set and same pass count before and after.

## What I did not do

**A blanket `before_request` hook.** It is the obvious generalization, and I tried it first:

```python
@hypercorn_app.before_request
@pyodide_testing_app.before_request
async def read_request_body() -> None:
    await request.get_data(parse_form_data=True)
```

It breaks four tests. `parse_form_data=True` consumes a multipart body into `request.form`, so `/echo` then echoes back an empty body and `test_post_with_multipart` fails. Dropping `parse_form_data` instead breaks every handler that reads `request.form` or `request.values`, which is the original `/redirect` bug. The per-handler call lets each one ask for what it actually needs.

**`OPTIONS` paths.** `/echo_json` and `pyodide_upload` return early for `OPTIONS` without reading the body, but a CORS preflight does not carry one, and on the hypercorn app `/echo_json` is registered for `POST` only.

### DIFF
--- a/test/with_dummyserver/test_connectionpool.py
+++ b/test/with_dummyserver/test_connectionpool.py
@@ -677,6 +677,42 @@ class TestConnectionPool(HypercornDummyServerTestCase):
                 b"--boundary--\r\n",
             ]
 
+    @pytest.mark.parametrize(
+        "method,url",
+        [
+            ("GET", "/echo"),
+            ("POST", "/echo"),
+            ("POST", "/headers"),
+            ("POST", "/multi_headers"),
+            ("POST", "/redirect?target=/"),
+            ("POST", "/specific_method?method=POST"),
+            ("PUT", "/successful_retry"),
+            ("POST", "/upload"),
+        ],
+    )
+    def test_handler_reads_the_request_body(self, method: str, url: str) -> None:
+        """A body the handler has no use for still has to leave the socket.
+
+        Anything left behind is read as the start of the next request made on
+        that connection, which then never gets a response. The body has to be
+        larger than the socket buffers for that to happen.
+        """
+        with HTTPConnectionPool(self.host, self.port, maxsize=1) as pool:
+            pool.request(
+                method,
+                url,
+                body=b"x" * (2 * 1024 * 1024),
+                headers={"test-name": "test_handler_reads_the_request_body"},
+                redirect=False,
+                retries=False,
+            )
+
+            # A generous timeout, so that a handler that stops reading the
+            # body fails the test instead of hanging it.
+            response = pool.request("GET", "/", timeout=10, retries=False)
+            assert response.status == 200
+            assert pool.num_connections == 1, "the connection was not reused"
+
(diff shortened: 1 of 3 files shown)


test/with_dummyserver/test_connectionpool.py
illia-v, line 1: Could we remove this test?
I don't think we need dedicated tests for dummyserver itself; it's already exercised by urllib3's tests.

SEPURI-SAI-KRISHNA: Removed.

For the record, since it is the reason the test existed: six of its eight parametrizations failed without the handler changes, so the sweep is not a no-op. The failure only appears once the body is larger than the socket buffers, though, which is exactly why urllib3's own tests never caught it and why the handlers had been leaking bodies unnoticed. Happy to bring it back if you would rather have that pinned, but I take the point that dummyserver is scaffolding rather than something to test in its own right.

illia-v (approved):

Issue #1999, opened 2020-09-25
Construct [Proxy-]Authorization headers from username:password section of URLs


sethmlarson:
We already support this behavior for SOCKS proxies, we should support it in general.

Currently we simply drop this information on the floor for both URLs and HTTP[S] proxy URLs. Might be nice to add support for this in v2. We can error out if both `Authorization` or `Proxy-Authorization` are given and they don't match for the same request to be extra safe.

hramezani:
@sethmlarson I have some questions:

> We already support this behavior for SOCKS proxies, we should support it in general.

Do you mean username/password extraction from URL here:

https://github.com/urllib3/urllib3/blob/3a815144ba9822b2981d3300833ae553d6d22a78/src/urllib3/contrib/socks.py#L173-L178

and add them to `connection_pool_kw["_socks_options"]`?

> Currently we simply drop this information on the floor for both URLs and HTTP[S] proxy URLs. Might be nice to add support for this in v2.

Do we want to implement it for `ProxyManager`? 

> We can error out if both Authorization or Proxy-Authorization are given and they don't match for the same request to be extra safe.

Does it mean that we have to extract username/password from url and check them with headers in `__init__`?

sigmavirus24:
@sethmlarson so I'm not exactly a fan of just doing things for users without their permission. I get this is probably the "smart" thing to do, but i tmakes it hard to factor/maintain later and will probably end up embedding this in the wrong place.

Thinking of v2, I'm wondering if there needs to be some kind of way of specifying authentication strategies (yes, plural) with a high-ish level object. If you're familiar with requests, there's a way of creating a subclass of AuthBase and building logic there which allows you to (for a session or for an individual call) specify a single "Auth" class to handle that request and any redirects that are followed (depending on security context).

There's also a probably not ideal hack towards something like this for Requests: https://toolbelt.readthedocs.io/en/latest/authentication.html but I do think this begs a few questions:

- Do we want to give people "strategies" to provide/handle authentication such that one of the _default_ strategies might be to parse from the URL if present?
- Do we want people to have multiple strategies (if one fails, fall back on the others)?
- Do we want to give people the ability to register strategies per domain?

To be clear, i think we could provide a strategy manager without forcing per-domain to start with. I will say, I don't recall having received bug reports about the hacks I wrote in the Montreal airport but I don't know if that means no one uses it.

abebeos:
@CCLDArjun , this has now a bounty, will you continue with your PR #2870?

[...]

PR #5245, opened 2026-09-09
Construct Basic authentication from request and proxy URL userinfo

stantheman0128:
Fixes #1999.

Request URL userinfo now generates Basic Authorization, and HTTP(S) proxy URL userinfo generates Proxy-Authorization. Credentials are percent-decoded using the existing URL API and encoded with the existing Latin-1 Basic-auth convention. Userinfo is removed from request targets and stored proxy URLs. Matching explicit authentication headers are accepted; conflicting values fail before sending the request, without including credentials in the error.

Caller header mappings remain unchanged. Redirect handling uses the target origin rather than the proxy connection's host, preserving authentication for same-origin forwarding and removing it on cross-origin redirects under the default Retry policy. Redirect log messages omit userinfo. Forwarded proxy headers are merged case-insensitively so matching credentials aren't duplicated.

Regression coverage includes:
- Ordinary, percent-encoded, username-only, Latin-1 and unsupported-encoding credentials; explicit matches/conflicts and caller mapping reuse.
- Actual HTTP request bytes for direct pools, PoolManager and forwarding proxies, with no userinfo in the request target.
- Same-origin and cross-origin redirects, both directly and through a forwarding proxy.
- A real verified TLS connection through HTTP CONNECT: proxy credentials occur only in CONNECT and origin credentials only in the tunneled request.
- Credential-free redirect logs and conflict errors, plus matching byte-valued headers.

Local verification on Windows/Python 3.13.3:
- Original manager/proxy-manager baseline: 52 passed; initial new regressions: 8 failed, 1 passed before implementation.
- Final 25 specialized unit/socket/TLS cases passed. The expanded proxy suite also passed 116 tests before the last two CONNECT cases were added.
- Final full `test/` suite with the project socket restrictions: **2327 passed, 329 skipped, 45 xfailed**.
- The four changed source modules report 885 statements, two missed (both unchanged lines), 99% combined statement coverage. The complete merged platform coverage remains for hosted CI.
- Linux-target mypy passed 85 files. Changed Python files passed Black (py310), isort, flake8 and pyupgrade; `git diff --check` passed.

Hosted verification now reports 37 successful checks and one failed Ubuntu 3.13 job. [Requests and botocore downstream tests](https://github.com/urllib3/urllib3/actions/runs/34345866889) pass. [Combined coverage from the available 28 artifacts](https://github.com/urllib3/urllib3/actions/runs/34345866881/job/102449184560) reports 3,949 statements, zero missed, 100%. The remaining matrix failure is documented in [this diagnosis and rerun request](https://github.com/urllib3/urllib3/pull/5245#issuecomment-5601216595): an unchanged test finds `h2` inside ClientHello random bytes even though the actual ALPN list is only `http/1.1`. A maintainer rerun is still needed; this is not a claim that the full matrix passed. CONNECT validation also checks that a matching Proxy-Authorization supplied in request headers is removed from the tunneled origin request, and a conflicting value is rejected before connecting.

Implemented and self-reviewed with OpenAI Codex; no independent human review is claimed.

### DIFF
--- a/changelog/1999.feature.rst
+++ b/changelog/1999.feature.rst
@@ -1,0 +1,3 @@
+Added Basic authentication from request and HTTP(S) proxy URL userinfo,
+with percent decoding, conflicting-header checks, and removal of credentials
+from request targets and stored proxy URLs.
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -657,3 +657,29 @@ standard logger interface to change the log level for urllib3's logger:
 .. code-block:: python
 
     logging.getLogger("urllib3").setLevel(logging.WARNING)
+
+
+Authentication in URLs
+----------------------
+
+Request URLs may contain ``username:password`` userinfo::
+
+    http.request("GET", "https://username:password@example.com/private")
+
+urllib3 percent-decodes the userinfo and generates a Basic ``Authorization``
+header using Latin-1, matching :func:`urllib3.util.make_headers`. A username
+without a password is treated as having an empty password. Credentials that
+cannot be represented in Latin-1 raise ``ValueError``. To choose another
+encoding or authentication scheme, omit URL userinfo and supply an explicit
+header instead.
+
+The userinfo is removed from the request target. Headers supplied by the
+caller are not mutated: matching authentication headers are accepted, while
+conflicting values raise ``ValueError`` before the request is sent. Generated
+origin authentication follows the configured redirect-header removal policy;
+by default it is removed when the target origin changes.
+
+HTTP and HTTPS proxy URLs accept userinfo in the same way, generating
+``Proxy-Authorization`` and removing userinfo from the stored proxy URL.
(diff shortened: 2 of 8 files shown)


stantheman0128:
The only failing matrix job appears to be a false positive in the unchanged `test_tunnel_sets_http_11_alpn` assertion.

I parsed the ClientHello shown in the [Ubuntu 3.13 failure](https://github.com/urllib3/urllib3/actions/runs/34345866881/job/102447198778): the bytes `h2` occur at offset 19, inside the ClientHello random field (offsets 11–42). The actual ALPN extension contains only `http/1.1`. The test currently searches the entire packet for `h2`, so those random bytes trigger the assertion. The test file is identical to the base commit; no test changes were made here.

The contributor-side CLI reports that this job cannot be rerun. Could a maintainer rerun the failed Ubuntu 3.13 job and its coverage dependency when convenient? I have left the code and assertions unchanged.

PR #5247, opened 2026-09-10
Test pickling of `InvalidChunkLength`

brianhelba:
`InvalidChunkLength` is not covered by the `TestPickle` matrix, and it is a natural candidate for it: the exception carries a whole `HTTPResponse` object in its `args`, which is exactly the kind of payload that can break pickling when an exception crosses a process boundary (e.g. multiprocessing or concurrent.futures workers).

The test builds a response that has been partially read, so that `partial`, `expected`, and `length` all hold non-default values, and asserts that each of them survives a pickle round trip.

### DIFF
--- a/test/test_exceptions.py
+++ b/test/test_exceptions.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import io
 import pickle
 import socket
 from email.errors import MessageDefect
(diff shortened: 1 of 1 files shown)


illia-v (commented):
Could we assert that non-default `length`, `partial`, and `expected` survive pickling and unpickling?
At the moment this only checks that pickling and unpickling succeed and preserve the exception type.

brianhelba:
@illia-v Done.

illia-v (approved):
Thanks!

PR #5250, opened 2026-09-10
Classify TLS handshake timeouts as connect timeouts

### LINKED ISSUE #1366: Low-level ssl handshake timeout exception treats as ReadTimeoutError
First I've post the issue in the requests requests/requests#4590 project, but then found out that this is only urllib3's feature.
In short, urllib3 treats ssl handshake timeout as ReadTimeoutError, what can be little unexpected.
System information and stack trace you can find in requests/requests#4590

### LINKED ISSUE #2591: socket.timeout from _validate_conn() is re-raised as ReadTimeoutError
### Subject

I am experimenting with different cases of broken connections in order to figure out how to handle the exceptions from the "requests" package, and found a case where a broken connection causes requests `session.post()` to raise `urllib3.exceptions.MaxRetryError` with text saying this was caused by:

ReadTimeoutError("HTTPSConnectionPool(host='localhost', port=5989): Read timed out. (read timeout=10)")

Since this was a broken connection test case (a Docker container handling the targeted port, and the container was started but paused), I was wondering why a read timeout was surfaced when I expected a connection error or connection timeout.

Digging into the urllib3 code revealed that the code in `HTTPConnectionPool._make_request()` calls `self._validate_conn()` which raises a `socket.timeout`. That exception is handled in `self._raise_timeout()` which re-raises it as `ReadTimeoutError`.

The current code (from urllib3 version 1.26.9) is:

```
    def _make_request(
        self, conn, method, url, timeout=_Default, chunked=False, **httplib_request_kw
    ):
    . . .
        try:
            self._validate_conn(conn)
        except (SocketTimeout, BaseSSLError) as e:
            # Py2 raises this as a BaseSSLError, Py3 raises it as socket timeout.
            self._raise_timeout(err=e, url=url, timeout_value=conn.timeout)
            raise
```

```
    def _raise_timeout(self, err, url, timeout_value):
        if isinstance(err, SocketTimeout):
            raise ReadTimeoutError(
                self, url, "Read timed out. (read timeout=%s)" % timeout_value
            )
        . . .
```

It seems to me that re-raising `socket.timeout` as `ReadTimeoutError` is not appropriate in the case where `self._validate_conn()` times out, because that is when a connection is established and validated, not when a response is read.

### Environment

OS: macOS-11.6.4-x86_64-i386-64bit
Python: 3.9.10
urllib3: 1.26.9

### Steps to Reproduce

requests `session.post()` on "https://localhost:{port}" where the port is handled by a Docker container on the local system, and where the Docker container is started and then paused.

### Expected Behavior

I think the exception re-raised in `_raise_timeout()` should depend on the context: If the connection is still being created, it should be raised as a ConnectTimeoutError, and if reading the response times out, it should be re-raised as `ReadTimeoutError`. Not sure about the case where sending the request times out.

### Actual Behavior

See above.

viknith:
Closes #2591.
Closes #1366.

## Summary

TLS handshakes run while the socket is configured with the connection timeout. A socket timeout raised during HTTPS connection validation was incorrectly converted to `ReadTimeoutError`, making callers treat a connection-establishment failure like a response-read timeout.

This change:

- raises `ConnectTimeoutError` for `SocketTimeout` during connection validation and TLS handshake;
- preserves existing handling for other SSL errors and response-read timeouts;
- uses distinct connect and read timeout values in the regression test; and
- adds the required Towncrier news fragment.

## Validation

- Focused TLS regression: 1 passed
- Affected socket, connection-pool, retry, and exception suites: 330 passed, 8 skipped
- Full Python 3.13 nox suite: 2,357 passed, 288 skipped, 45 expected failures
- `nox -rs lint`: passed, including pyupgrade, Black, isort, Flake8, uv-lock, zizmor, Prettier, ESLint, and mypy
- Towncrier draft: passed
- Wheel and source distribution builds: passed

### DIFF
--- a/changelog/2591.bugfix.rst
+++ b/changelog/2591.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed TLS handshake timeouts being incorrectly raised as ``ReadTimeoutError`` instead of ``ConnectTimeoutError``.
--- a/src/urllib3/connectionpool.py
+++ b/src/urllib3/connectionpool.py
@@ -26,6 +26,7 @@ from .connection import (
 from .connection import port_by_scheme as port_by_scheme
 from .exceptions import (
     ClosedPoolError,
+    ConnectTimeoutError,
     EmptyPoolError,
     FullPoolError,
     HostChangedError,
@@ -468,7 +469,13 @@ class HTTPConnectionPool(ConnectionPool, RequestMethods):
             # Trigger any extra validation we need to do.
             try:
                 self._validate_conn(conn)
-            except (SocketTimeout, BaseSSLError) as e:
+            except SocketTimeout as e:
+                raise ConnectTimeoutError(
+                    conn,
+                    f"Connection to {self.host} timed out. "
+                    f"(connect timeout={conn.timeout})",
+                ) from e
+            except BaseSSLError as e:
                 self._raise_timeout(err=e, url=url, timeout_value=conn.timeout)
                 raise
 
--- a/test/with_dummyserver/test_socketlevel.py
+++ b/test/with_dummyserver/test_socketlevel.py
@@ -46,6 +46,7 @@ from urllib3._collections import HTTPHeaderDict
 from urllib3.connection import HTTPConnection, _get_default_user_agent
 from urllib3.connectionpool import _url_from_pool
 from urllib3.exceptions import (
+    ConnectTimeoutError,
     InsecureRequestWarning,
     MaxRetryError,
     ProtocolError,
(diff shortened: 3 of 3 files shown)


viknith:
Complete newb to this repository, we are hitting an issue where we want to retry ConnectTimeouts but because they are not being raised as ConnectTimeouts our system is not retrying because we wouldn't want to retry ReadTimeouts and risk leaving resources hanging that were created on a prior execution, so I drafted this change to resolve it, feel free to leave thoughts or let me know if a similar change was attempted in the past that wasn't approved and why.

PR #5253, opened 2026-09-12
Share security policy between GitHub and Read the Docs

illia-v:
I noticed that README uses a relative link to our security policy, which can break when rendered outside the repository (e.g., urllib3's PyPI page)
https://github.com/urllib3/urllib3/blob/5f2a6a843d0100d1351c3f94d58581ca98d17267/README.md#L91

So I added the security policy to docs and linked it from README. Additionally, the policy is now included in the source distribution and PyPI links.
I moved the file to the project root where GitHub should still discover it for https://github.com/urllib3/urllib3?tab=security-ov-file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,9 @@ To report a security vulnerability, please use the
 [Tidelift security contact](https://tidelift.com/security).
 Tidelift will coordinate the fix and disclosure with maintainers.
 
-See the repository [Security Policy](.github/SECURITY.md) for more details.
+See the repository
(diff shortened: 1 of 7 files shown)


illia-v:
https://urllib3--5253.org.readthedocs.build/en/5253/security.html

shazow (approved):

sigmavirus24 (approved):

illia-v:
@shazow @sigmavirus24 thanks for the review!

PR #5103, opened 2026-07-06
Replace deprecated pyOpenSSL `X509.get_subject` and `Context.set_passwd_cb` methods

illia-v:
Fixes #5099 by replacing use of deprecated `X509.get_subject` and `Context.set_passwd_cb`.

Previously emitted warnings:
- `DeprecationWarning: X509.get_subject is deprecated. You should use cryptography's X.509 APIs instead.`
- `DeprecationWarning: Context.set_passwd_cb is deprecated. You should decrypt and load your private key yourself, with cryptography's key loading APIs, and then use Context.use_privatekey instead.`

PR #5255, opened 2026-09-15
Fix loading unencrypted client keys with a password in pyOpenSSL

illia-v:
While checking for unintended breaking changes before the next release, I found that #5103 introduced a small regression when `load_cert_chain()` is called with an unencrypted client key and a non-`None` password (including an empty string or empty bytes).

Previously, the unused password was ignored. After #5103, this configuration raised `ssl.SSLError` "Unable to load certificate chain: TypeError('Password was given but private key is not encrypted.')".

This restores the previous behavior, matching the standard SSL backend. Incorrect passwords for encrypted keys remain rejected.

I didn't add a changelog entry because this fixes an unreleased bug.


src/urllib3/contrib/pyopenssl.py
```diff
@@ -512,7 +512,15 @@ def load_cert_chain(
                 # versions because set_passwd_cb() became deprecated in 26.3.0.
                 if int(OpenSSL.__version__.split(".")[0]) >= 26:
                     with open(keyfile or certfile, "rb") as key_file:
-                        private_key = load_pem_private_key(key_file.read(), password)
+                        key_data = key_file.read()
+                    try:
+                        private_key = load_pem_private_key(key_data, password)
+                    except TypeError:
+                        # Match SSLContext by ignoring passwords supplied
```
sethmlarson, line 519: Do we want to avoid trying again if `password` is already `None` and reraise?

sethmlarson (approved):
Only one comment, LGTM, nice catch on this.

src/urllib3/contrib/pyopenssl.py
illia-v: `password` should never be `None` on these lines because the branch is guarded by `if password is not None`

PR #5258, opened 2026-09-15
Skip `test_read_chunked_with_trailing_data_does_not_hang` for brotlicffi

illia-v:


sethmlarson (approved):
Let's open a template issue (title only) so this doesn't get forgotten, too.

PR #5260, opened 2026-09-17
Fix several issues with IPv6 and IPv6 zone IDs

Ninja3047:
* Fixed a regression where unbracketed scoped IPv6 addresses passed to `HTTPConnectionPool` or `HTTPSConnectionPool` were rejected or incorrectly normalized.
* Fixed a regression where repeated normalization of IPv6 zone identifiers in `ProxyManager` caused the forwarded request target to disagree with the `Host` header.
* Fixed a bug where IPv6 zone identifiers were incorrectly lowercased in connection-pool keys and CONNECT targets.

Regressions introduced by https://github.com/urllib3/urllib3/pull/5095

IPv6 zone ids are really complicated and interact in annoying ways with existing RFCs so some of these things we can choose to ignore.

Here are some of the relevant RFCs + Drafts:
 * https://www.rfc-editor.org/info/rfc3986/ - Uniform Resource Identifier (URI): Generic Syntax
 * https://www.rfc-editor.org/info/rfc6874/ - Representing IPv6 Zone Identifiers in Address Literals and Uniform Resource Identifiers
 * https://www.rfc-editor.org/info/rfc9844/ - Entering IPv6 Zone Identifiers in User Interfaces
 * https://datatracker.ietf.org/doc/html/draft-ietf-6man-rfc6874bis-09 - Representing IPv6 Zone Identifiers in Address Literals and Uniform Resource Identifiers
 * https://datatracker.ietf.org/doc/html/draft-schinazi-httpbis-link-local-uri-bcp-03 - Best Practices for Link-Local Connectivity in URI-Based Protocols

Technically some of the changes here may not follow the RFCs strictly since they are ambiguous, underspecified, or conflict. Some of these wider issues are discussed in some of the RFCs/drafts linked above.


src/urllib3/util/url.py
```diff
@@ -382,11 +384,11 @@ def _normalize_host(host: str | None, scheme: str | None) -> str | None:
...
                         zone_id = zone_id[1:]
                     zone_id = _PERCENT_RE.sub(
-                        partial(_normalize_zone_id_percent_encoding, error_host=host),
+                        partial(_normalize_host_percent_encoding, error_host=host),
                         zone_id,
                     )
-                    zone_id = _encode_invalid_chars(zone_id, _UNRESERVED_CHARS)
-                    return f"{host[:start].lower()}%{zone_id}{host[end:]}"
+                    # Keep the zone separator encoded across repeated URL parsing.
+                    return f"{host[:start].lower()}%25{zone_id}{host[end:]}"
```
illia-v, line 391: Requests rebuilds its URL using `parse_url().host`, then strips brackets through `urllib.parse` before calling `connection_from_host()`. The newly retained `%25` is consequently interpreted as part of the socket's zone identifier. Requesting `http://[fe80::1%2525]/` with Requests selected scope 25 previously, this PR selects scope 2525.

illia-v (commented):
This PR seems to address both v2.8.0 regressions and pre-existing IPv6 issues, but it also introduces a regression in Requests. Resolving that compatibility issue while retaining all these changes seems likely to add even more complexity. I'd suggest focusing this PR on the v2.8.0 regressions, adding a test covering the affected Requests behavior, and addressing the older issues separately.

src/urllib3/util/url.py
Ninja3047: should be fixed now

src/urllib3/util/url.py
illia-v, line 1: For the next 2.8.1 patch release, I'd prefer the minimal diff so it's clear exactly which changes fix the regressions. Including the larger refactoring makes those changes harder to understand, so I'd suggest handling it separately.

For this file, I think the required change can be limited to:

```diff
diff --git a/src/urllib3/util/url.py b/src/urllib3/util/url.py
--- a/src/urllib3/util/url.py
+++ b/src/urllib3/util/url.py
@@ -391,6 +391,9 @@
                     return host.lower()
             elif not _IPV4_RE.match(host):
                 if "%" in host:
+                    if ":" in host and host.isascii() and _IPV6_RE.match(host.partition("%")[0]):
+                        # Native IPv6 zones contain literal percent signs.
+                        return host.lower()
                     host = _HOST_PERCENT_RE.sub(_normalize_host_percent_encoding, host)
                 return to_str(
                     b".".join([_idna_encode(label) for label in host.split(".")]),
```

src/urllib3/poolmanager.py
```diff
@@ -454,7 +454,9 @@ def urlopen(  # type: ignore[override]
             kw["headers"] = self.headers
 
         if self._proxy_requires_url_absolute_form(u):
-            response = conn.urlopen(method, u._replace(fragment=None).url, **kw)
+            # The connection pool normalizes the URL. Rebuilding it here would
+            # decode IPv6 zone identifiers a second time.
+            response = conn.urlopen(method, url.split("#", 1)[0], **kw)
```
illia-v, line 459: Could you also mention in the comment that the fragment is stripped here, while URL normalization is left to the connection pool?

PR #5261, opened 2026-09-17
Make trustme a test dependency

musicinmybrain:
Make `trustme` a test dependency since it’s imported unconditionally in `test/conftest.py`, and is therefore required in order to run any tests at all:

https://github.com/urllib3/urllib3/blob/b1d30ab61fe0db8f11092805e8c5ac43e091064a/test/conftest.py#L10

Since the `test` dependency group is included directly or indirectly in all the other dependency groups, use a minimum version bound that’s compatible with the version of `trustme` pinned in `dev-min-pyopenssl`:

https://github.com/urllib3/urllib3/blob/b1d30ab61fe0db8f11092805e8c5ac43e091064a/pyproject.toml#L102


illia-v (commented):
Thanks for reporting this!

We could introduce another base dependency group without `trustme` and reuse it across the other groups. The `test` group could then depend on `trustme>=1.2.1`.

PR #5262, opened 2026-09-17
dummyserver leaks sockets when a port bind is retried

SEPURI-SAI-KRISHNA:
### What happened

Some random tests on Ubuntu CI ERROR with:

```text
ExceptionGroup: multiple unraisable exception warnings (2 sub-exceptions)
```

The two warnings are always:

```text
ResourceWarning: unclosed <socket family=10 (AF_INET6) laddr=('::1', PORT)>
ResourceWarning: unclosed <socket family=2 (AF_INET) laddr=('0.0.0.0', 0)>
```

Both come from `dummyserver/hypercornserver.py`, in `_retry_create_urllib3_sockets`. Which test gets the ERROR is random. It happened in 12 different tests across 13 jobs.

### Why it happens

`_retry_create_urllib3_sockets` retries up to 10 times when the bind fails with `EADDRINUSE`.

Inside `_create_urllib3_sockets`, we go through the results of `getaddrinfo("localhost")` and create a socket for each one. We set the socket options, bind the first socket to port 0, and it gets a random port. Each later socket is then bound to that port. The problem is that if one of those binds fails, we don't close the sockets that were created during that attempt.

In the CI Ubuntu runners, `localhost` resolves to `::1` first and then `127.0.0.1`. So in the failing case:

1. the `::1` socket gets port `P` from the kernel
2. we then try to bind `127.0.0.1:P`
3. the IPv4 bind gets `EADDRINUSE`
4. the retry starts
5. the socket list from the failed attempt is lost
6. the `::1` socket is never closed
7. the IPv4 socket whose bind failed is never closed either

CPython later cleans up these sockets and raises the two `ResourceWarning`s. Since `pyproject.toml` has `filterwarnings = ["error"]`, pytest reports these warnings as an ERROR.

### How I reproduced it

I first reproduced it with real sockets and a real kernel `EADDRINUSE`. I had a blocker listening on `127.0.0.1` at the port that `::1` had just received. Locally, `localhost` only resolves to `127.0.0.1`, so I only had to emulate the `::1`-then-`127.0.0.1` resolver order. That produced the same two unclosed sockets as CI.

I then reproduced it end-to-end through a real pytest run in `test/with_dummyserver/test_connection.py`. The test injects one port collision during dummyserver startup, and a plugin records whether the collision actually happened.

With main, each of 5 runs gave:

```text
29 passed, 1 error
2 unclosed sockets
```

The error was the same CI `ExceptionGroup`, with the same two sub-exceptions.

With the fix, each of 5 runs gave:

```text
30 passed, 0 errors
0 unclosed sockets
```

Across all 10 runs, `collision_injected=True` every time. Each run had one failed bind with `EADDRINUSE`, followed by a successful retry.

Separately, I checked the sockets returned after a successful retry. It still returns two open sockets, `::1` and `127.0.0.1`, and both use the same port.

### CI

I searched the CI history from June 15 through September 16, 2026. The failures themselves happened in 13 Ubuntu jobs between July 1 and September 9. Five of those jobs are proven flaky because the same job also passed on the same commit. Two of the jobs were on main. The rate was about 1 in 393 Ubuntu CPython test jobs. There were no new occurrences from September 10 through September 16.

CI example:
- https://github.com/urllib3/urllib3/actions/runs/31429999901/job/93590820201

Proven flaky jobs:
- https://github.com/urllib3/urllib3/actions/runs/30850679098/job/91809657780
- https://github.com/urllib3/urllib3/actions/runs/33895538431/job/101097170836
- https://github.com/urllib3/urllib3/actions/runs/33972110476/job/101322331392
- https://github.com/urllib3/urllib3/actions/runs/34273101910/job/102219395742

As a separate check, exactly the same 13 jobs contain `Retrying binding to ...`.

### The fix

The loop is wrapped in `try / except BaseException`. If anything goes wrong, all sockets created so far are closed and the exception is re-raised.

Each socket is also added to the list immediately after `socket.socket(...)`. This means the socket whose `bind()` fails is also closed. The retry behaviour is unchanged.

### Checks

The full test suite passes both before and after the change: `2396 passed, 309 skipped, 4 xfailed, 0 failures`.

Black, isort, flake8 and pyupgrade are clean with the pinned pre-commit versions. Mypy is clean when targeting Python 3.13 and 3.14, which matches the latest Python used by the CI lint job. When targeting Python 3.12, mypy reports:

```text
src/urllib3/connection.py:146: error: Final can be only used as an outermost qualifier in a variable annotation  [valid-type]
```

The same error is present on unmodified upstream main, so it is unrelated to this change.

### Related PRs

While looking into this, I found that the same change is already present in two open, unreviewed PRs:

- #4999 - "Construct Authorization and Proxy-Authorization headers from URL credentials" - last updated May 20 
- #5042 - "Update request header typing for bytes" - last updated May 23

In both PRs, this change is bundled with unrelated work. There are no reviews or comments about this change, and neither PR mentions the leak in its description. There is also no existing issue about this leak. This PR keeps the change focused and includes the CI evidence and the end-to-end reproduction.

### Why there is no test

Illia said on #5239:

> "I don't think we need dedicated tests for dummyserver itself"

So I did not add a dedicated dummyserver test.

### A few things that may come up

<details>
<summary><b>Isn't this already in #4999/#5042?</b></summary>

Yes. The same change is already included in both PRs, but it is bundled with unrelated work and neither PR has been reviewed. This PR is focused on this issue and includes the CI evidence and reproduction.
</details>

<details>
<summary><b>Why <code>BaseException</code> instead of <code>OSError</code>?</b></summary>

It also cleans up on things such as `KeyboardInterrupt`. The exception is always re-raised. This is also the choice used in #4999/#5042.
</details>

<details>
<summary><b>Why only Ubuntu?</b></summary>

I don't know. The retry message only appears in the Ubuntu logs. One caveat is that the message is only visible when a failing test prints its captured output.
</details>

<details>
<summary><b>How often does this happen?</b></summary>

About 1 in 393 Ubuntu CPython jobs.
</details>

<details>
<summary><b>Does the fix change behaviour when there is no bind failure?</b></summary>

The full test suite passes both before and after the change. The suite starts the dummyserver many times, including cases where there is no bind failure. The code on that path is also nearly the same. The change only adds the `try` wrapper and moves where the socket is appended to the list.
</details>

<details>
<summary><b>Why is there no test?</b></summary>

This follows Illia's guidance on #5239.
</details>

<details>
<summary><b>The comment says IPv4 comes first. Is that still true?</b></summary>

The comment at lines 39-41 is stale. The code follows the resolver order, and CI binds `::1` first. This PR does not change the comment.
</details>

### One unrelated observation

The retry loop also silently retries on `OSError`s other than `EADDRINUSE`. It eventually raises `failed to bind socket` without the original error. This is not part of this PR.


illia-v (approved):
LGTM, thanks!

PR #5265, opened 2026-09-18
Reject blocksize <= 0 to avoid silently dropping file bodies

### LINKED ISSUE #5248: `HTTPConnection(blocksize=0)` silently drops file-like request bodies
### Subject

When `blocksize=0` and the request body is a file-like object,  I discovered the request sent by urllib3 does not contain any data blocks, but the header still declares Transfer-Encoding: chunked. At this point, neither the client nor the server reported any errors, and the data was silently discarded.

### Environment

```
Python 3.12.3
OpenSSL 3.0.13 30 Jan 2024
urllib3 2.7.1.dev58
```

### Steps to Reproduce
```python
import socket
import tempfile
import threading
from pathlib import Path
from urllib3.connection import HTTPConnection
PAYLOAD = b"SECRET_PAYLOAD_C3_BLOCKSIZE_ZERO"
captured = []
srv = socket.socket()
srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
srv.bind(("127.0.0.1", 0))
srv.listen(4)
port = srv.getsockname()[1]
def serve():
    for _ in range(2):
        c, _ = srv.accept()
        c.settimeout(1)
        buf = b""
        try:
            while True:
                chunk = c.recv(65536)
                if not chunk:
                    break
                buf += chunk
                if b"\r\n\r\n" in buf:
                    c.settimeout(0.2)
        except OSError:
            pass
        captured.append(buf)
        c.sendall(b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\nConnection: close\r\n\r\n")
        c.close()
threading.Thread(target=serve, daemon=True).start()
path = Path(tempfile.mkdtemp()) / "upload.bin"
path.write_bytes(PAYLOAD)
def post(blocksize):
    conn = HTTPConnection("127.0.0.1", port, blocksize=blocksize, timeout=5)
    with path.open("rb") as fh:
        conn.request("POST", "/upload", body=fh)
    conn.getresponse().read()
    conn.close()
post(16384)
post(0)
print("default:", captured[0])
print("zero:   ", captured[1])
```

### Expected Behavior

Either reject `blocksize <= 0`, or treat `0` as “read all” (like `-1`). A file body must not disappear without an exception.

### Actual Behavior
Captured on the wire against a local HTTP server, same on-disk file both times.
```
default: b'POST /upload HTTP/1.1\r\nHost: 127.0.0.1:46047\r\nAccept-Encoding: identity\r\nTransfer-Encoding: chunked\r\nUser-Agent: python-urllib3/2.7.1.dev58\r\n\r\n20\r\nSECRET_PAYLOAD_C3_BLOCKSIZE_ZERO\r\n0\r\n\r\n'
zero:    b'POST /upload HTTP/1.1\r\nHost: 127.0.0.1:46047\r\nAccept-Encoding: identity\r\nTransfer-Encoding: chunked\r\nUser-Agent: python-urllib3/2.7.1.dev58\r\n\r\n0\r\n\r\n'
```

Pitchfork-and-Torch:
## Summary
`HTTPConnection(blocksize=0)` with a file-like body sent `Transfer-Encoding: chunked` and only the terminating `0\r\n\r\n` chunk because `file.read(0)` always returns `b\"\"` in Python. The upload disappeared with no client error.

Raise `ValueError` when `blocksize <= 0` in:
- `HTTPConnection.__init__`
- `body_to_chunks`
- HTTP/2 file-like send path

Fixes #5248.

## Test plan
- [x] Unit tests for `body_to_chunks(blocksize=0)` and `HTTPConnection(..., blocksize=0)`
- [x] Manual wire capture repro from the issue now raises instead of sending an empty body


illia-v:
This PR does not comply with our contributing policy because another pull request addressing the same issue was already open.

Please see our contributing guidelines: https://github.com/urllib3/urllib3/tree/2.8.0#contributing

PR #5266, opened 2026-09-18
Fix flaky SSL socket termination test on Windows

SEPURI-SAI-KRISHNA:
## What happened

`test_ssl_custom_validation_failure_terminates` intermittently fails on Windows with:

```text
AssertionError: The socket was not terminated
assert False
where False = wait(0.5)
```

The same test is also run through the pyopenssl copy in `test/contrib/test_pyopenssl.py`.

There was no thread exception in any of the failures.

I checked CI from Jun 15 to Sep 17 2026. There were 15 failures, all on Windows, between Jun 24 and Sep 15. Eight are proven flaky because the same job passed on the same commit, and three happened on main.

The failure occurred on Python 3.10, 3.11, 3.12, 3.14, 3.14t, and 3.15. There were no occurrences on 3.13. This is about 1 failure in 217 Windows test jobs.

The authors of #5249 and #5222 also hit the same failure on their PRs.

The eight proven flaky jobs are:

https://github.com/urllib3/urllib3/actions/runs/28471462074/job/84384922745

https://github.com/urllib3/urllib3/actions/runs/30303029543/job/90100365977

https://github.com/urllib3/urllib3/actions/runs/32984841906/job/98249021464

https://github.com/urllib3/urllib3/actions/runs/33557044583/job/100020310500

https://github.com/urllib3/urllib3/actions/runs/33895542413/job/101097183633

https://github.com/urllib3/urllib3/actions/runs/34835918191/job/103949597749

https://github.com/urllib3/urllib3/actions/runs/35006288030/job/104506744557

https://github.com/urllib3/urllib3/actions/runs/33557044583/job/100020310404

## Why it happens

I found two ways to produce the exact CI failure by injecting the conditions into the real test on main.

First, the server can get `ConnectionResetError` during the TLS handshake instead of receiving the client's TLS alert. The handler catches the exception and returns without setting `server_closed`. The final `wait()` then fails regardless of how long it waits.

Second, the server can handle the TLS alert more than 0.5 seconds after the client has finished. `LONG_TIMEOUT` is 0.5 seconds on CI, so the wait times out.

I cannot tell from the CI logs which case Windows is hitting.

I couldn't reproduce the first case on Linux in 300 runs per test copy, even with a real TCP reset from the client. The alert that arrived before the reset was still delivered.

Since all observed CI failures are on Windows, the first case seems more likely, but that is only an inference.

There is also a report of Windows dropping received data when an RST arrives:

https://github.com/objeck/objeck-lang/issues/669

There is a separate report of Windows SSL tests intermittently getting `ConnectionResetError` because closing a socket with unread data can send a reset:

https://github.com/python-trio/trio/issues/1293

I have not independently verified either report.

## The change

The PR changes only the test in `test/with_dummyserver/test_socketlevel.py`, plus the changelog entry.

When the server gets `ConnectionResetError`, it now sets `server_closed` before returning. A reset from the client means the connection has been terminated.

The timeout is unchanged and remains `server_closed.wait(LONG_TIMEOUT)`.

Nothing else in the test changes. `is_closed_socket`, the `SSLError` branch, and the client side are untouched.

The pyopenssl copy gets the same fix because it imports `TestSSL` as `TestSocketSSL`.

The reset case is covered by setting the event before returning. The late alert case is not covered by this change. If the server handles the alert more than 0.5 seconds late, the test still fails as it does on main.

## Verification

I ran the two test copies on Python 3.12 five times and Python 3.14t twice for each case.

| Case                  | Main                     | With the fix             |
| --------------------- | ------------------------ | ------------------------ |
| Nothing injected      | pass                     | pass                     |
| Reset                 | FAIL with the CI message | pass                     |
| 0.6 second late alert | FAIL with the CI message | FAIL with the CI message |
| Real TCP RST          | pass                     | pass                     |

I also ran 300 runs per test copy with normal close and real TCP RST. All runs passed.

The full `test_socketlevel.py` and `test_pyopenssl.py` tests give:

```text
382 passed, 153 skipped, 2 xfailed
```

The same result as main.

Black, isort, flake8, and pyupgrade are clean. Mypy with Python 3.13 and 3.14 reports no issues.

## One separate issue

There is another problem with the termination check that I did not change here.

`is_closed_socket()` is always true because the TLS wrap detaches the original socket. This means the test passes even if the client never closes its socket.

I verified this by suppressing the client's `close()`. The test still passes on main and with this fix.

This is separate from the flaky failure, so I left it unchanged. I have a tested rewrite for this check and can open it as a follow up if the maintainers want it.


test/with_dummyserver/test_socketlevel.py
```diff
@@ -1979,7 +1981,7 @@ def socket_handler(listener: socket.socket) -> None:
         ) as pool:
             with pytest.raises(SSLError):
                 pool.request("GET", "/", retries=False, timeout=LONG_TIMEOUT)
-        assert server_closed.wait(LONG_TIMEOUT), "The socket was not terminated"
+        assert server_closed.wait(5), "The socket was not terminated"
```
illia-v: Is the timeout increase based on an evidence that CI has exceeded the current timeout, or is it a precaution against a possible delay?

SEPURI-SAI-KRISHNA: The 5 second wait is a precaution. I don't have evidence from CI that the 0.5 second timeout was actually exceeded.

I checked the raw logs for all 8 proven flaky jobs. The failing test took between 0.518 and 0.529 seconds in the call phase. Seven of those failures have a rerun of the same job, where the same test copy took between 0.010 and 0.026 seconds. The eighth is proven flaky by a different run on the same commit.

So the failures are spending almost exactly the full 0.5 seconds in `wait()`.

That still does not tell us why the event was not set. I reproduced both cases locally with `CI=true`:

```text id="fr4y4v"
injected reset       -> call ~0.51 s
injected late alert  -> call ~0.51 s
```

Both produce the same failure and timing because `wait()` expires in either case. The CI logs do not record whether the server eventually set the event.

I also checked whether delays above 0.5 seconds happen on these Windows runners. They do. In both failing and passing jobs, a small number of unrelated tests took more than 0.3 seconds longer than their median, with some delays above 1 second and a few close to 3 seconds. This shows that delays of this size are possible, but not that one happened in this test.

So the two changes cover different cases. Setting `server_closed` on `ConnectionResetError` fixes the case where the handler returns without setting the event. The 5 second wait is only a precaution for the case where the server is late.

The reset case still looks more likely because all 15 failures were on Windows, while I couldn't reproduce it on Linux even with a real TCP reset in 300 runs per test copy. But the CI logs cannot tell us which case actually happened.

I'm fine with dropping the 5 second wait if you prefer to keep the change limited to the reset handling. The tradeoff is that a late alert would still produce the same failure if that turns out to be what Windows is hitting.

illia-v: Let's drop the increased timeout for now.

SEPURI-SAI-KRISHNA: done

illia-v: Thanks!

illia-v (approved):

PR #5269, opened 2026-09-20
Document upload buffering and `blocksize` tuning

illia-v:
Document how to stream file uploads and tune `blocksize` to reduce read/send overhead, including chunking behavior and memory tradeoffs.
A known case is boto/botocore#3612, where increasing `blocksize` improved large S3 upload throughput with urllib3 v2.


nateprewitt (approved):
This looks good, thanks @illia-v!

@hssyoo, not sure if you have anything else you think would be generally relevant from your work around this.

illia-v:
@nateprewitt thanks for the review!

